Issue #37, opened 2013-08-15
Missing Repository APIs


wlynch:
Might as well make this one large issue and cross things off as we go instead of making 1 issue per API. There are still a few APIs missing for the [repository service](http://developer.github.com/v3/repos/):
- [x] Repository
- [x] Collaborators
- [x] `Comments` : @wlynch92
- [x] `Commits`: @ktoso 
- [x] Contents
- [x] Forks
- [x] Keys: @willnorris
- [x] Hooks
- [x] Merging
- [x] Releases
- [x] Statistics
- [x] Statuses

I'll edit this post as the APIs are assigned/implemented. If any one wants to help in implementing any of these APIs feel free to comment saying which you'll be working on.

PR #49, opened 2013-08-25
Repo commits

ktoso:
The long awaited commit with `repo/commits` API.
The API is different than `git/commit` in having more detailed information about the commit itself - for example lists of files, and authors etc.

The tests are a bit verbose. I've tested against the example jsons from the github api page, but that saved me from overlooking a field or two when updating the structs. 

I've also moved this code to the new style that we've started using in the lib recently (pointers everywhere).
I skipped the URL fields as we did before, though not sure if maybe for the "compare commits" case the URL may be helpful - they're for patches etc.

Looking forward to your comments guys :)

Implements one of the APIs mentioned in #37

### DIFF
--- a/github/repos_commits_test.go
+++ b/github/repos_commits_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"reflect"
 	"testing"
+	"time"
 )
 
 func TestRepositoriesService_ListCommits(t *testing.T) {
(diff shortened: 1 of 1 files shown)


github/repos_commits.go
```diff
@@ -0,0 +1,143 @@
...
+// RepositoryCommit represents a commit in a repo.
+// Note that it's wrapping a Commit, so author/committer information is in two places,
+// but contain different details about them: in RepositoryCommit "github details", in Commit - "git details".
+type RepositoryCommit struct {
+	SHA       *string  `json:"sha,omitempty"`
+	Commit    *Commit  `json:"commit,omitempty"`
+	Author    *User    `json:"author,omitempty"`
+	Committer *User    `json:"committer,omitempty"`
+	Parents   []Commit `json:"parents,omitempty"`
+	Message   *string  `json:"message,omitempty"`
```
shurcooL: I suspect `Message` field was added here erroneously. It's should only be inside `Commit`, not `RepositoryCommit`.

From what I can tell, it's not actually ever present in the GitHub API responses. Does anyone have counter-examples? If not, I'll make a PR to remove it.

gmlewis: SGTM. Thanks, @shurcooL.

shurcooL: While making that PR, I ran into another issue and submitted a fix at #491. That needs to be reviewed and merged first, @gmlewis.

PR #63, opened 2013-10-11
add support for basic Releases methods

tbruyelle:
Includes Repositories.ListReleases, Repositories.GetRelease, Repositories.CreateRelease and Repostories.DeleteRelease.

I'll attempt to code the remaining methods about assets (http://developer.github.com/v3/repos/releases) in following commits.

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -6,8 +6,8 @@
 package github
 
 import (
-	"io"
 	"fmt"
+	"io"
 )
 
 // RepositoryRelease represents a GitHub release in a repository.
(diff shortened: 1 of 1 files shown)


github/repos_releases.go
```diff
@@ -0,0 +1,242 @@
...
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	req.Header.Add("Accept", mimeReleasePreview)
+
+	assets := new([]ReleaseAsset)
+	resp, err := s.client.Do(req, assets)
+	if err != nil {
+		return nil, resp, nil
```
shurcooL: Is this an unintended error? If `err` is not `nil`, it returns `nil` error.

If this is intended, we should really add a comment here noting that, because otherwise it looks like a typo.

github/repos_releases.go
```diff
@@ -0,0 +1,242 @@
...
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	req.Header.Add("Accept", mimeReleasePreview)
+
+	asset := new(ReleaseAsset)
+	resp, err := s.client.Do(req, asset)
+	if err != nil {
+		return nil, resp, nil
```
shurcooL: Same here, returning `nil` error when `err != nil` seems incorrect.

PR #68, opened 2013-11-07
Support both named and numeric ID schemes

sqs:
For context: https://github.com/google/go-github/pull/60

This PR is a prototype for supporting both named (`owner/repo`) and numeric ID schemes. It only includes modifications to 3 repository methods for now, as we figure out the right way to implement this.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -51,6 +51,34 @@ func (r Repository) String() string {
 	return Stringify(r)
 }
 
+// RepositorySpec specifies a repository.
+type RepositorySpec interface {
+	Path() string
+}
+
+// RepositoryName identifies a repository by its owner and name.
+type RepositoryName struct {
+	// Owner is the user login of the repository's owner.
+	Owner string
+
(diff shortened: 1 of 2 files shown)


phuna:
Hi, I'm getting into the same issue when a repository is renamed then github API will return 404 for it if using named URL. Is there any plan to resolve conflict & merge this PR yet?

Thank you

willnorris:
closing this, since we're long past the ability to make this change and @shurcooL just added new methods to handle ID based fetching in #329

shurcooL:
I have to admit, I did not see the actual content of this PR until just now. I mistakenly thought this was the same change as #60.

Now that I've seen it, I completely agree with https://github.com/google/go-github/pull/68#issuecomment-28015311. I like it a lot too.

Both luckily and unfortunately, it is too late to much such a change now, so nothing would've changed even if I had seen it. But I'll definitely remember this idea for reference if I'm ever designing API/client libraries in the future.

Issue #152, opened 2014-11-20
Make it easer to detect rate limit errors


ImJasonH:
The GitHub API [returns rate limit information in HTTP response headers](https://developer.github.com/v3/#rate-limiting) for every request.

It might be useful if the GitHub client (optionally) inspected those values and proactively avoided hitting rate limits by blocking a subsequent request until it thinks it will have quota, then [checking that it does indeed have quota](https://developer.github.com/v3/rate_limit/), before making the request.

This would probably be prohibitively hard to get right 100% of the time, but as long as it left some slush quota in case of race conditions, it could be really useful for clients to automatically obey their rate limit without having to design for it.

willnorris:
As discussed in #153, automatically throttling the application is not the right behavior for this library.  However, we could make it a little easier for applications to detect rate limit errors (basically a 403 response with zero rate.Remaining) so they can respond appropriately.

shurcooL:
> As discussed in #153, automatically throttling the application is not the right behavior for this library. However, we could make it a little easier for applications to detect rate limit errors (basically a 403 response with zero rate.Remaining) so they can respond appropriately.

Agreed.

> As you mention, the key is just leaving some wiggle room in case of race conditions.

Now that #255 has been done, this should be very doable.

> You want to take a stab at this?

I'll take this. PR soon.

It'll be 2 PRs. First one just to add a new error type for rate limit exceeded situations and returning it when appropriate. The second PR will be an optimization (with no change to library behavior) to avoid doing unnecessary networking calls when it's known that the rate limit has been exhausted and not enough time has yet passed for it to be reset.

PR #153, opened 2014-11-21
Add an option to automatically obey rate limits

ImJasonH:
This is done by checking whether the number of remaining requests we
have according to previous rate limit header responses is >5% of the
total number of requests per hour we are allowed, and if the client is
below that number, it will wait a short amount of time until issuing the
API request.

This is not fully threadsafe and may produce surprising results if two
identically-configured clients are making requests at the same time,
consuming the same quota pool. The 5% soft threshold is intended to
alleviate some of this, but it can still be fooled by a determined user.

This is only intended as a convenience to users, not as a guarantee that
rate limits won't be hit while using this option.

This also adds a test that checks that requests against a
rate-limit-enforcing server succeed with the option on and fail with it
off.

This attempts to address #152

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -77,6 +77,13 @@ type Client struct {
 	Repositories  *RepositoriesService
 	Search        *SearchService
 	Users         *UsersService
+
+	// If true, calls to Do will only result in an API request if there is
+	// deemed to be room under the rate limit that the request would
+	// succeed. If there is no room under the rate limit, requests will
+	// block until there is quota.
+	ObeyRateLimit bool
+	wait          func(time.Duration) // Called when rate-limit waiting happens
 }
 
 // ListOptions specifies the optional parameters to various List methods that
@@ -138,6 +145,7 @@ func NewClient(httpClient *http.Client) *Client {
 	c.Repositories = &RepositoriesService{client: c}
 	c.Search = &SearchService{client: c}
 	c.Users = &UsersService{client: c}
+	c.wait = time.Sleep
 	return c
 }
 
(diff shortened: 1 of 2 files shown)


willnorris:
(sorry for the stupid-long delay on getting around to review this :disappointed:)

I'm not sure that sleeping is always the right thing to do in the event that a rate limit has been hit.  Rather, I think the library should construct and return a 403 response (the same that GitHub would return if you're over the limit).  The application can then decide to either sleep until the limit expires, or do something else.

So that makes the changes to the library much simpler:  we already record the most recent Rate in [Client](http://godoc.org/github.com/google/go-github/github#Client), so we could just update `Do()` to consult that first, and return a constructed 403 response if `Remaining` is 0.  The only catch is that `client.Rate` isn't a pointer, so it will initialize with a zero Remaining value... we'd also need to make sure that `client.Rate.Limit` is non-zero, otherwise the very first call would fail.

ImJasonH:
Ha, no worries, I actually figured out it wasn't the right way to go and pretty much abandoned the idea. :)

The main problem is that GitHub's rate limiting is done in a, uh, suboptimal way. Instead of having a gradually refilling bucket of quota, it simply sets a reset time an hour after your first request and says you get all of your quota back at that time. So in a lot of cases these requests will pile up for an ~hour, which is a really long time to wait without any feedback.

As such, I think it's likely the developer will want to have more control/visibility into the process than just allowing `go-github` to handle things transparently. It would be possible to write such a thing as a custom wrapper around `go-github` if you really wanted to, but that would be up to each developer.

Your proposal to proactively self-rate-limit when we think we'll be over does make some sense, but I'd worry that you'd end up in cases where the server would have let you make the request but for whatever you self-limit. As Walter Sobchak reminds us, "the Supreme Court has roundly rejected prior restraint." :)

willnorris:
agreed.  Though I do think go-github could make it a little easier for an application to detect that it is a rate limit error versus some other kind of error.  I'll leave #152 open, and rename it to capture that.

PR #157, opened 2014-12-10
Add Git.DeleteTag()

tcnksm:


willnorris:
So this is currently possible using [`GitService.DeleteRef`](https://github.com/google/go-github/blob/0839c734c3ff947840fa6bb0acb74e780fe624f1/github/git_refs.go#L153), using something like:

```go
client.Git.DeleteRef("owner", "repo", "tag/v1.0")
```

I *guess* we could create a convenience wrapper around that for tags, but it seems unnecessary.  Even if we did, I'd probably have it call `DeleteRef`, like:

```go
func (s *GitService) DeleteTag(owner string, repo string, tag string) (*Response, error) {
    return s.DeleteRef(owner, repo, "tag/"+tag)
}
```

willnorris:
closing, based on no recent discussion and the fact that this can be done with DeleteRef (see above)

PR #158, opened 2014-12-10
Updates PushEvent and PushEventCommit structs

stevedomin:
It looks like the struct for `PushEvent` and `PushEventCommit` got out of date with the actual Github API and were missing some fields.

This PR adds these fields back.


willnorris:
hmm, something is definitely weird here.  Search for `PushEvent` in the response at https://api.github.com/repos/google/go-github/events.  They certainly seem to still follow the older format, not what the documentation says.

willnorris:
I remember that there's something special about push events, where webhooks include an older payload format that they didn't want to change which would break people... maybe that accounts for the difference.  It's been a while since I've looked at all of this, so I don't quite remember what was what.

stevedomin:
My bad for not reading the docs properly, they indeed say that the API and Webhook payload differ.

> Note: the example payload below is from a webhook delivery. The Event API PushEvent payload will differ.

Not sure what's the best way to deal with that, will have a think about it.

PR #161, opened 2015-01-12
Add CommentCount to Commit.

shurcooL:
This integer exposes the number of comments that have been posted for that commit on GitHub.

It is documented at https://developer.github.com/v3/repos/commits/.

This is needed for https://github.com/shurcooL/Go-Package-Store/pull/30.

P.S. I think @GitHub messed up the API a little by putting this field inside of what we refer to as [`Commit`](http://godoc.org/github.com/google/go-github/github#Commit) in this package, rather than [`RepositoryCommit`](http://godoc.org/github.com/google/go-github/github#RepositoryCommit), since technically comments are a property of github commits rather than git commits. But what can I do. :/ I've carefully tested to ensure this works.

### DIFF
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -12,14 +12,15 @@ import (
 
 // Commit represents a GitHub commit.
 type Commit struct {
-	SHA       *string       `json:"sha,omitempty"`
-	Author    *CommitAuthor `json:"author,omitempty"`
-	Committer *CommitAuthor `json:"committer,omitempty"`
-	Message   *string       `json:"message,omitempty"`
-	Tree      *Tree         `json:"tree,omitempty"`
-	Parents   []Commit      `json:"parents,omitempty"`
-	Stats     *CommitStats  `json:"stats,omitempty"`
-	URL       *string       `json:"url,omitempty"`
+	SHA          *string       `json:"sha,omitempty"`
+	Author       *CommitAuthor `json:"author,omitempty"`
+	Committer    *CommitAuthor `json:"committer,omitempty"`
+	Message      *string       `json:"message,omitempty"`
+	Tree         *Tree         `json:"tree,omitempty"`
+	Parents      []Commit      `json:"parents,omitempty"`
+	Stats        *CommitStats  `json:"stats,omitempty"`
+	URL          *string       `json:"url,omitempty"`
+	CommentCount *int          `json:"comment_count,omitempty"`
 }
 
 func (c Commit) String() string {


shurcooL:
> P.S. I think @GitHub messed up the API a little by putting this field inside of what we refer to as [`Commit`](http://godoc.org/github.com/google/go-github/github#Commit) in this package, rather than [`RepositoryCommit`](http://godoc.org/github.com/google/go-github/github#RepositoryCommit), since technically comments are a property of github commits rather than git commits. But what can I do. :/ I've carefully tested to ensure this works.

Hmm, maybe not?

```Go
// Commit represents a GitHub commit.
type Commit struct {
```

I thought it was supposed to represent a git commit. Well, I'm not sure.

willnorris:
Hmm... maybe our comment on the Commit struct is confusing, since I think you're right that it's supposed to represent the Git data, not the GitHub data for the commit.

But I think they put it in the right place... I think you're patching the wrong one.  The methods documented at https://developer.github.com/v3/repos/commits/ correspond to our methods in [repos_commits.go](https://github.com/google/go-github/blob/master/github/repos_commits.go).  So I think you want to be adding this onto `RepositoryCommit`, right?  Or am I looking at the wrong thing perhaps?

willnorris:
doh, nevermind.  Now I'm looking closer and I think you're right... looks like they put this in the wrong place.

willnorris:
@pengwynn this certainly seems odd.  Why isn't `comment_count` at the same level as `comments_url`... the upper level that contains GitHub commit data, rather than the git commit data?  Is it too late for that to be corrected? (I'm not really sure when this was introduced into the API)

shurcooL:
Thanks for getting the GitHub team involved. If they can fix this upstream, that'll be better as it will keep the GitHub API vey nice and clean.

I think it can be done by adding the same field in the right place, and deprecating/eventually removing the second copy in the wrong place.

If that's done, I'll update this PR to get the value from the more appropriate place.

pengwynn:
Admittedly it would have been cleaner to put that count up one level, but it's too late to change it now. We'll have to live with it I'm afraid.

willnorris:
or add it to the upper level, document it that way, and then just leave the other one for backwards compatibility?

shurcooL:
@pengwynn, any additional thoughts on @willnorris's suggestion above?

shurcooL:
Ping @pengwynn, can you let us know if there will be any attempted improvements to the GitHub API on this point? If yes, what time frame should I expect? If not, I think we can proceed with merging this PR.

pengwynn:
@shurcooL I wouldn't wait on an API change for this one.

willnorris:
added field comment and merged as 1e55bf6be814270d7cc31a60868c4bd89e4e68c0

shurcooL:
Thanks!

PR #163, opened 2015-02-05
Update links to current location of oauth2 library

kjk:


willnorris:
So that's actually a different oauth2 library entirely, intended to fully replace goauth2.  I'm all for switching over to the new library, but it means we need to make sure and update all of the examples as well, not just the URLs.  I just hadn't gotten around to doing it yet.

kjk:
my bad, I thought it was just a repo move

kjk:
For what it's worth: I've used golang.org/x/oauth2 in a small program and it worked with go-github

willnorris:
oh yeah, I have no idea it will work fine.  I just haven't gotten around to updating things.

PR #170, opened 2015-03-12
Clean client_id and client_secret from error.

### LINKED ISSUE #169: OAuth token leakage in error messages
OAuth tokens are leaked in Go errors. The errors contain the full URL which contains `client_id` and `client_secret` and can be leaked to a user or easily passed on to someone who shouldn't see it.

Shouldn't it be sanitized from the URL - basically just remove the `client_id` and `client_secret` fields and their respective values from the error string - to prevent those from being leaked to an end-user?

**Before:** `Get https://api.github.com/user/:username/repos?sort=created&client_id=test1234&client_secret=1234test: 404 Not Found []`.

**After:** `Get https://api.github.com/user/:username/repos?sort=created: 404 Not Found []`.

I have a basic solution that I'm about to commit and submit as a pull request. It simply removes those fields and their values from the error text. It satisfies the before/after clause detailed above.

AustinDizzy:
The client_id and client_secret URL variables, containing sensitive app information, are leaked to users if the Go error is returned to them. To prevent this, sanitizeErr cleans the fields from the URL in the error message, making it safe to expose to users. This resolves #169.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -333,8 +333,23 @@ type ErrorResponse struct {
 
 func (r *ErrorResponse) Error() string {
 	return fmt.Sprintf("%v %v: %d %v %+v",
-		r.Response.Request.Method, r.Response.Request.URL,
+		r.Response.Request.Method, sanitizeURL(r.Response.Request.URL),
 		r.Response.StatusCode, r.Message, r.Errors)
+}
+
+// sanitizeURL redacts the client_id and client_secret tokens from the URL which
+// may be exposed to the user, specifically in the ErrorResponse error message.
+func sanitizeURL(uri *url.URL) *url.URL {
+	if uri == nil {
+		return nil
+	}
+	params := uri.Query()
+	if len(params.Get("client_secret")) > 0 || len(params.Get("client_id")) > 0 {
+		params.Set("client_id", "REDACTED")
+		params.Set("client_secret", "REDACTED")
+		uri.RawQuery = params.Encode()
+	}
+	return uri
 }
 
 /*


github/github.go
```diff
@@ -200,6 +201,17 @@ func (c *Client) NewUploadRequest(urlStr string, reader io.Reader, size int64, m
 	return req, nil
 }
 
+// sanitizeErr removes client_id and client_secret variables from URL in
+// net/http error, making it safe to expose.
+func sanitizeErr(err error) error {
+	if err == nil {
+		return nil
+	}
+	re := regexp.MustCompile(`client_(id|secret)=([^&|\s|:]*)(&)?`)
```
parkr: Why recompile this regexp every time?

AustinDizzy: @parkr You're right, my mistake. The commit I just pushed places the regex outside the func so that it's only compiled once at compile time.

parkr:
Cool.

One thing I might consider is simply redacting the values here. Rails does this with passwords. The output just replaces the actual value with `[redacted]`. The advantage of this is that at least you know you are sending the client ID and client secret up. Otherwise an authorized request looks the same as an unauthorized request.

willnorris:
okay, so here's a brain dump of some ideas on this, some with conflicting approaches of how this could be handled.

#### credentials shouldn't be passed in query params

As I commented on #169, the root problem here is that credentials are being passed as query parameters at all.  This is a problem with the GitHub API itself, so is beyond our immediate control.  I've pulled in a few engineers that work on the GitHub API in hopes that they may be able to fix this upstream.  This is of course my preferred solution, as it addresses the real problem.  We'll have to wait and see what the GitHub folks say on that.

#### client id and secret are added by UnauthenticatedRateLimitedTransport

One thing to consider is that `client_id` and `client_secret` are only ever appended to the URL if you are using the [UnauthenticatedRateLimitedTransport](https://github.com/google/go-github/blob/7ea4ee6d222607c11ea86e99a6f6723beeae785d/github/github.go#L481).  I wonder if it would be possible to redact the values at the [very end of RoundTrip](https://github.com/google/go-github/blob/7ea4ee6d222607c11ea86e99a6f6723beeae785d/github/github.go#L515).  I'm not sure if that would potentially cause other problems, like some code retrying the original request or something?

#### sanitizeErr hides the original error type

I'm not crazy about the fact that `sanitizeErr` collapses all errors down to a simple string.  Note that many errors returned are actually of type [ErrorResponse](https://github.com/google/go-github/blob/7ea4ee6d222607c11ea86e99a6f6723beeae785d/github/github.go#L328), which provides access to some additional error info.  We likely just want to sanitize the URL string that is used in that type's [Error method](https://github.com/google/go-github/blob/7ea4ee6d222607c11ea86e99a6f6723beeae785d/github/github.go#L334).  That won't cover 100% of the possible errors returned (notably, those returned by `NewRequest` mentioned below), but I think it's a better place to start and see how far it gets us.

#### patch only addresses errors in NewRequest

The example you give in #169 is a 404 response, which only happens as a result of calling [Do](https://github.com/google/go-github/blob/7ea4ee6d222607c11ea86e99a6f6723beeae785d/github/github.go#L294).  You're patch here only sanitizes errors inside `NewRequest` and `NewUploadRequest`, which are far less likely to actually return errors (basically, only if you provide an invalid URL I think).  The place we should more likely be sanitizing is inside of `Do`.  Moving the sanitization to either the unauthenticated transport or ErrorResponse (see above) may take care of this, but it's worth noting anyway.

#### redact values rather than strip params

+1 to @parkr's suggestion of redacting the values rather than stripping params altogether.

AustinDizzy:
I believe it would be sufficient to sanitize the error message at the [Error method](https://github.com/google/go-github/blob/7ea4ee6d222607c11ea86e99a6f6723beeae785d/github/github.go#L334), as that is the error message that's returned from various API calls that would be reflected to the user. Even if the error is caused by net/http, [CheckResponse](https://github.com/google/go-github/blob/master/github/github.go#L371) would still return the ErrorResposne which contains the sanitized URL.

And as far as redacting the parameters, it seems reasonable as to why we'd want to keep the fields but redact their values. I'll revert my commits shortly and redact the values at ErrorResponse's Error method instead of stripping them at NewRequest. It'll also be easier to redact the params if we do it in the Error method because we'll have access to the URL object right there from the http.Response.

In the end, the best solution to this would be to just get those sensitive values out of the URL.

AustinDizzy:
I've pushed the new code that redacts the fields if they're present instead of removing them. Makes it really easy since the ErrorResponse gives access to the Reqest which gives a pointer to url.URL. I had the field values replaced with `REDACTED` (as opposed to `[redacted]`) since net/url adds the fields as a URL-encrypted value which made the actual value `%5Bredacted%5D`. I guess if everyone favors `[redacted]` more, I can simply replace the URL-encrypted values with their correctly mapped characters [before setting the RawQuery](https://github.com/AustinDizzy/go-github/commit/a63440d20cf2de334145bf45d09320e5afd89df2#diff-9bd6c2575adb220521a096df92e2d502R350).

willnorris:
rebased onto master and merged as c9bf21a1b7aad75370c3d7d92a78642f0b4594a6.

I added a test for this, and in doing so noticed that we don't actually need to redact the `client_id` parameter, since it's not expected to be secret (for example, it's exposed to all users during an OAuth flow).  I've gone and ahead and removed that, so we're just redacting the `client_secret` now.

Thanks for your work on this!

PR #171, opened 2015-03-19
fix sample options

dmnlk:
RepositoryListByOrgOptions doesn't have `Sort` param


willnorris:
thanks for the fix!  rebased onto master and tweaked language; merged as 31ecb2cc3f0d2eb77551e39001c0063326ed103f

PR #182, opened 2015-04-28
Two bug fixes and one enhancement

xogeny:
The `git_url` and `html_url` tags in the `RepositoryContent` struct were wrong.  I've fixed tha
and added one additional field that is in the API but wasn't provided here, `download_url`.  The
`download_url` link is important because it is not subject to the same download limitations
that access through the API is.  So it provides a way to get at large files.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -19,16 +19,17 @@ import (
 
 // RepositoryContent represents a file or directory in a github repository.
 type RepositoryContent struct {
-	Type     *string `json:"type,omitempty"`
-	Encoding *string `json:"encoding,omitempty"`
-	Size     *int    `json:"size,omitempty"`
-	Name     *string `json:"name,omitempty"`
-	Path     *string `json:"path,omitempty"`
-	Content  *string `json:"content,omitempty"`
-	SHA      *string `json:"sha,omitempty"`
-	URL      *string `json:"url,omitempty"`
-	GitURL   *string `json:"git_url,omitempty"`
-	HTMLURL  *string `json:"html_url,omitempty"`
+	Type        *string `json:"type,omitempty"`
+	Encoding    *string `json:"encoding,omitempty"`
+	Size        *int    `json:"size,omitempty"`
+	Name        *string `json:"name,omitempty"`
+	Path        *string `json:"path,omitempty"`
+	Content     *string `json:"content,omitempty"`
+	SHA         *string `json:"sha,omitempty"`
+	URL         *string `json:"url,omitempty"`
+	GitURL      *string `json:"git_url,omitempty"`
+	HTMLURL     *string `json:"html_url,omitempty"`
+	DownloadURL *string `json:"download_url,omitempty"`
 }
 
 // RepositoryContentResponse holds the parsed response from CreateFile, UpdateFile, and DeleteFile.
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -27,8 +27,8 @@ type RepositoryContent struct {
 	Content  *string `json:"content,omitempty"`
 	SHA      *string `json:"sha,omitempty"`
 	URL      *string `json:"url,omitempty"`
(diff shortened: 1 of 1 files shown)


xogeny:
I submitted the agreement right after I submitted the PR.  I suspect your reply and my submission just crossed.  Also, the fact that the commits came from one account and the pull request from another could be an issue.  But it looks like you base things on commit emails.  So hopefully it is ok...?

xogeny:
I signed it!

willnorris:
Having the pull request and commit authors differ does confuse things a little, but we can make it work.  However, I don't see a CLA on file for the email address in these commits (mic***@gmail.com).

willnorris:
rebased onto current master and edit commit message to fit 50 chars. Merged as 17225f1b849c58d2f2035b5d47e9e22d71838a0f and 88853b4d83ec313aa748c10791f243f59f67c6af.  Thanks!

xogeny:
You are welcome.  And thanks for the super quick turnaround on this.

BTW, this isn't quite the end of the story here.  I consider this kind of a quick fix.  If you actually try and get the `RepositoryContent` for a specific file and it is over 1Mb, you'll get an error.  I added the `download_url` specifically to allow me to download the file without the 1Mb limit (since it downloads via `raw.githubusercontent.com`).  However, to get that `download_url`, you actually have to fetch the parent directory and then find the child with the matching path and then do an `http.Get` for that URL.  In other words, what should be a one step process turns into at least a 3 step process.

Part of the problem is the GitHub API because it won't let you fetch just the `download_url` for a file (at least not that I can see) without also fetching the `content` field.  The problem is you can't get either of these if `content` is larger than 1Mb.  But you could work around this by adding another method to `RepositoriesService` service along the lines of `DownloadContent` that always fetches the parent directory and finds the appropriate download URL and then calls `http.Get` and returns an `io.ReadCloser` to the caller.

The question is...would you accept a pull request for such functionality?  Or do you think that should be the users responsibility?

willnorris:
ugh.  Do you know if using the `application/vnd.github.VERSION.raw` custom media type makes a difference with respect to the 1 MB limit?

willnorris:
Assuming using the custom media type is still limited to 1 Mb, I think a `DownloadContent` method as you describe it sounds fine.  We'd definitely want to document it pretty clearly, and return an appropriate error if it is called for anything that doesn't have a download URL (like a directory or symlink).

xogeny:
Interesting question.  So [here is an example of a link that demonstrates the problem](https://api.github.com/repos/modelica-3rdparty/BondLib/contents/BondLib%202.3/package.mo?ref=df7a40fe612617da22e27d39edfa4b27d65f23d0) (this is precisely the link that `go-github` is doing a `GET` on behind the scenes).

You can see that if you fetch it with a browser (which will use this, for example, as the `Accept` header: `text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8`).  You get an error.  If you change the `Accept` header to explicitly `application/json`, you get the same error (no surprise).  If you change the `Accept` header to `application/vnd.github.VERSION.raw` you get the exact same error.  So my point is that the custom media type doesn't seem to matter.  The GitHub documentation doesn't actually say how you use these media types.  I **assume** it is via an `Accept` header (if they did it some other way, that would be crazy).

So based on your most recent response and my examination of the issue, it seems you are open to another pull request for `DownloadContent`.  I'll try to whip something up before I move on to something else.

willnorris:
yeah, I tried a few different files and can confirm that it looks like the 1 Mb limit applies regardless of content type.  And yes, you're definitely expected pass that as an Accept header.  Compare:

    curl -i "https://api.github.com/repos/modelica-3rdparty/BondLib/contents/.gitignore"

vs

    curl -H "Accept: application/vnd.github.VERSION.raw" -i "https://api.github.com/repos/modelica-3rdparty/BondLib/contents/.gitignore"

PR #183, opened 2015-04-28
This adds an additional method to RepositoriesService to download a file

xogeny:
This is a followup to #182.  This provides a consistent way to download files regardless of whether they are over 1Mb in size.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -13,8 +13,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
+	"path"
 )
 
 // RepositoryContent represents a file or directory in a github repository.
@@ -89,6 +91,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
 		return nil, resp, err
 	}
 	return readme, resp, err
+}
+
+// DownloadContents can be used to download files.  It circumvents the normal
+// GitHub API (which has a file size limit) and uses the `download_url` link
+// provided by the API.  That URL provides a direct link to the raw data
+// associated with that version of the file and is not subject to the same
+// limits as the GitHub API based requests.
+//
+// If path references a file or a symlink and the call succeeds, it
+// returns an instance of io.ReadCloser that provides the contents of
+// the file.  Otherwise, it returns an error.
+func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt *RepositoryContentGetOptions) (io.ReadCloser, error) {
+	dir := path.Dir(filepath)
+	fname := path.Base(filepath)
+	_, dcon, _, err := s.GetContents(owner, repo, dir, opt)
+	if err != nil {
+		return nil, err
+	}
+	for _, con := range dcon {
+		if *con.Name == fname {
+			if con.DownloadURL == nil || *con.DownloadURL == "" {
+				return nil, fmt.Errorf("No download link found for %s", filepath)
+			}
+			resp, err := http.Get(*con.DownloadURL)
+			if err != nil {
+				return nil, fmt.Errorf("Error downloading %s from %s: %v", filepath,
+					*con.DownloadURL, err)
+			}
+			return resp.Body, nil
+		}
+	}
+	return nil, fmt.Errorf("No file named %s found in %s", fname, dir)
 }
 
 // GetContents can return either the metadata and content of a single file
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -109,7 +109,7 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 			if contents.DownloadURL == nil || *contents.DownloadURL == "" {
 				return nil, fmt.Errorf("No download link found for %s", filepath)
 			}
-			resp, err := http.Get(*contents.DownloadURL)
+			resp, err := s.client.client.Get(*contents.DownloadURL)
 			if err != nil {
 				return nil, err
 			}
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -104,25 +104,24 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
 // the file.  Otherwise, it returns an error.
 func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt *RepositoryContentGetOptions) (io.ReadCloser, error) {
 	dir := path.Dir(filepath)
-	fname := path.Base(filepath)
-	_, dcon, _, err := s.GetContents(owner, repo, dir, opt)
+	filename := path.Base(filepath)
+	_, dirContents, _, err := s.GetContents(owner, repo, dir, opt)
 	if err != nil {
 		return nil, err
 	}
-	for _, con := range dcon {
-		if *con.Name == fname {
-			if con.DownloadURL == nil || *con.DownloadURL == "" {
+	for _, contents := range dirContents {
+		if *contents.Name == filename {
+			if contents.DownloadURL == nil || *contents.DownloadURL == "" {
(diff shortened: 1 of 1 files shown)


xogeny:
Looks like I'm confusing your `cla/google` check again with my dual identity.  Sorry.  I should have submitted this as my other self.  Let me know if you want me to re-submit as the other user.

xogeny:
Is this submission OK or would you like me to change it in some way?

github/repos_contents.go
```diff
@@ -91,6 +93,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
 	return readme, resp, err
 }
 
+// DownloadContents can be used to download files.  It circumvents the normal
```
willnorris: I don't think the details of exactly how this is implemented is important for the public documentation here.  Maybe just something like:

> DownloadContents returns an io.ReadCloser that reads the contents of the specified file.  This function will work with files of any size, as opposed to GetContents which is limited to 1 Mb files.  It is the caller's responsibility to close the ReadCloser.

github/repos_contents.go
```diff
@@ -91,6 +93,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
...
+// GitHub API (which has a file size limit) and uses the `download_url` link
+// provided by the API.  That URL provides a direct link to the raw data
+// associated with that version of the file and is not subject to the same
+// limits as the GitHub API based requests.
+//
+// If path references a file or a symlink and the call succeeds, it
+// returns an instance of io.ReadCloser that provides the contents of
+// the file.  Otherwise, it returns an error.
+func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt *RepositoryContentGetOptions) (io.ReadCloser, error) {
```
willnorris: s/filepath/path/ for consistency with `GetContents`

github/repos_contents.go
```diff
@@ -91,6 +93,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
...
+	dir := path.Dir(filepath)
+	fname := path.Base(filepath)
```
willnorris: s/fname/filename/

github/repos_contents.go
```diff
@@ -91,6 +93,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
...
+	_, dcon, _, err := s.GetContents(owner, repo, dir, opt)
```
willnorris: s/dcon/dirContents/

github/repos_contents.go
```diff
@@ -91,6 +93,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
...
+		return nil, err
+	}
+	for _, con := range dcon {
+		if *con.Name == fname {
+			if con.DownloadURL == nil || *con.DownloadURL == "" {
+				return nil, fmt.Errorf("No download link found for %s", filepath)
+			}
+			resp, err := http.Get(*con.DownloadURL)
+			if err != nil {
+				return nil, fmt.Errorf("Error downloading %s from %s: %v", filepath,
```
willnorris: should be able to just `return nil, err` here.  I don't think we need to construct a new error.

github/repos_contents.go
```diff
@@ -91,6 +93,38 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
...
+// returns an instance of io.ReadCloser that provides the contents of
+// the file.  Otherwise, it returns an error.
+func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt *RepositoryContentGetOptions) (io.ReadCloser, error) {
+	dir := path.Dir(filepath)
+	fname := path.Base(filepath)
+	_, dcon, _, err := s.GetContents(owner, repo, dir, opt)
+	if err != nil {
+		return nil, err
+	}
+	for _, con := range dcon {
```
willnorris: s/con/contents/

willnorris:
Overall this looks great, thanks!  I left a few comments, mostly just relating to docs and more explicit var names.

willnorris:
(and thanks for the ping @parkr. somehow I managed to stop watching this repo, which makes no sense at all :angry: )

github/repos_contents.go
xogeny: Yes, that is what I did at first.  But if you do that, you shadow the `path` import which is required to extract the file name and parent directory.  So there are three options here:

   1. Change the argument to `path` and `import` the `path` package under a different name.

   2. Change all other occurrences to of `path` as arguments to something else so they are all consistent with each other.

   3. Change just the argument to this function.

I thought 1 would be confusing since it would effectively rename (in the scope of this function) a standard package to something else.  I chose 3 because, in fact, this really must be the path to a file (whereas in other cases it could be either a directory or a filename).

If you want me to use 1 or 2 or if you have another options, let me know.

xogeny:
I made all but one of the suggested changes.  I requested a bit more guidance on what you want in the case of the `path` argument (see above).

github/repos_contents.go
willnorris: no, you're totally right.  I overlooked that it would shadow the `path` import.

willnorris:
what do you think about simplifying the function docs a bit?

xogeny:
@willnorris Sorry, forgot to adjust the docs.  Should be done now.  I just went with your wording
which was was fine for me.

willnorris:
looks great, thanks!  We'll definitely want some tests on this as well.  I'm happy to add those myself, or you can take a stab if you want.

xogeny:
I considered the tests.  But I wasn't quite sure what your guidelines were for "mocking" GitHub in this case.  I'm happy to help, but it wasn't clear to me what you would be looking for there.

willnorris:
Most (or maybe all?) of our tests work with a single request-response pair, and then just have multiple tests methods to test different responses.  That certainly makes for more boilerplate code than I would typically care for, but for most of our tests it isn't too bad and allows us to have very clear error messages.  In hindsight, had I known about [gomock](https://code.google.com/p/gomock/) at the time I started on this, I might have done things differently, I don't know.

The tests for this will be slightly more complicated than most of ours since in the success case, you actually need to register two handlers on different URLs, but the same pattern we use elsewhere should work.  I'll go ahead and take a first pass at it.

github/repos_contents.go
```diff
@@ -91,6 +93,32 @@ func (s *RepositoriesService) GetReadme(owner, repo string, opt *RepositoryConte
...
+	_, dirContents, _, err := s.GetContents(owner, repo, dir, opt)
+	if err != nil {
+		return nil, err
+	}
+	for _, contents := range dirContents {
+		if *contents.Name == filename {
+			if contents.DownloadURL == nil || *contents.DownloadURL == "" {
+				return nil, fmt.Errorf("No download link found for %s", filepath)
+			}
+			resp, err := http.Get(*contents.DownloadURL)
```
willnorris: oops, I completely overlooked this.  We can't just call http.Get and assume that http.DefaultClient is the right client to use.  For example, this will fail on App Engine which requires that you use their [urlfetch](https://cloud.google.com/appengine/docs/go/urlfetch/) library.  This is why [NewClient](https://github.com/google/go-github/blob/master/github/github.go#L118) takes an `http.Client`.  So instead, this should be:

```go
resp, err := s.client.client.Get(*contents.DownloadURL)
```

so that this call will always use the internal `http.Client` inside the `github.Client`.

xogeny:
Just updated PR to use `s.client.client`.

willnorris:
awesome.  I just rebased onto master, squashed commits down and merged as c9d6108f1456ea668b9543a40e63a0e88175e0e5.  You can see how I did the tests in 6d44ada0e0e255873723fc4a143a754c4ed1247a if you're curious.

xogeny:
Very nice.  Thanks!

PR #193, opened 2015-04-30
licenses: add support for the OSS licenses API provided by GitHub.

### LINKED ISSUE #185: Add support for Licenses API
Announcement: https://developer.github.com/changes/2015-03-09-licenses-api/
Docs: https://developer.github.com/v3/licenses/

parkr:
Fixes #185.

- [x] Implement & test `IssuesService.List`
- [x] Implement & test `IssuesService.Get`
- [x] Ensure the preview header is being used and has TODO comments.
- [x] Implement for `Repository` & `Repositories.Get`

/cc @willnorris

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -37,6 +37,8 @@ const (
 
 	// Media Type values to access preview APIs
 
+	// https://developer.github.com/changes/2015-03-09-licenses-api/
+	mediaTypeLicensesPreview = "application/vnd.github.drax-preview+json"
 )
 
 // A Client manages communication with the GitHub API.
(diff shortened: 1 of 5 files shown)


willnorris:
we should also extend [Repository](https://github.com/google/go-github/blob/508dcfd4c05dc9873c7b731c5154b40a4f65671f/github/repos.go#L19) and [Repositories.Get](https://github.com/google/go-github/blob/508dcfd4c05dc9873c7b731c5154b40a4f65671f/github/repos.go#L251) to include license support.

parkr:
@willnorris I considered that. We have to add the preview Accept header in order to achieve this. Should we default to the preview?

willnorris:
yeah, in this case I don't think there's any harm to just always using the preview type.

parkr:
@willnorris  Pushed as a separate commit to keep things sane.

parkr:
@willnorris Let me know if there's anything else you need from me here.

willnorris:
looks good, thanks!

PR #198, opened 2015-05-06
use *[]string for DeploymentRequest.RequiredContexts

victorcoder:
There's no way of serializing/requesting an empty required_contexts that
is needed for bypassing the commit status checks for the ref.

It's a breaking change but will allow to serialize an empty array and
omit empty value as expected.

Example: http://play.golang.org/p/6Jsh8L7AfO


victorcoder:
I signed it!

willnorris:
thanks!

Issue #188, opened 2015-04-29
Add support for updated starring API


willnorris:
Announcement: https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/

rubenvereecken:
I'd like to tackle this one to get a feel for the project. I've already tracked the location of the code currently responsible for listing a user's starred repositories.

What would be the best way to support the new attributes (currently just `starred_at`)? GitHub allows a custom `Accept` header to request a different format. My guess would be to have this behavior controlled by the `ActivityListStarredOptions`, but the problem is that the returned object is now not an array of repositories anymore. Maybe an extra function `ListStarredWithDates` or similar?

willnorris:
yeah, this should be a good issue to get started with.  Most of the time with preview APIs like this, we just enable them by default if we can and don't bother adding an option for it.  In this particular case, it seems kinda silly that they put this behind a custom media type at all, since all it seems to be doing is adding a single field in the response to a single method.

Take a look at how this was done for the licenses API, and follow the same approach: https://github.com/google/go-github/search?q=mediaTypeLicensesPreview

 - define a new var with the new media type
 - update the relevant methods to set that as the accept header on requests (and add a TODO comment to remove it when it's no longer needed)
 - update tests to check for the presence of that media type

willnorris:
fixed in #199

PR #199, opened 2015-05-20
Support for updated starring API

rubenvereecken:
Supports the updated starring API as mentioned in issue #188.

There is now a new struct called StarredRepository which holds a pointer to a Repository and the new `starred_at` timestamp. This was the cleanest way I found to go about this

For now there is a special media type to request the new API. I've added a TODO for it to be deleted when the new API fully launches.

### DIFF
--- a/github/activity_star.go
+++ b/github/activity_star.go
@@ -7,6 +7,12 @@ package github
 
 import "fmt"
 
+// StarredRepository is returned by ListStarred.
+type StarredRepository struct {
+	StarredAt  *Timestamp  `json:"starred_at,omitempty"`
+	Repository *Repository `json:"repository,omitempty"`
+}
+
 // ListStargazers lists people who have starred the specified repo.
 //
 // GitHub API Docs: https://developer.github.com/v3/activity/starring/#list-stargazers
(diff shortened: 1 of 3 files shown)


github/activity_star.go
```diff
@@ -7,6 +7,11 @@ package github
 
 import "fmt"
 
+type StarredRepository struct {
```
willnorris: needs godoc

github/activity_star_test.go
```diff
@@ -62,12 +63,13 @@ func TestActivityService_ListStarred_specifiedUser(t *testing.T) {
...
 	mux.HandleFunc("/users/u/starred", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeStarringPreview)
 		testFormValues(t, r, values{
 			"sort":      "created",
 			"direction": "asc",
 			"page":      "2",
 		})
-		fmt.Fprint(w, `[{"id":2}]`)
+		fmt.Fprint(w, `[{"repository":{"id":2}}]`)
```
willnorris: go ahead and add a timestamp to the test case

willnorris:
ah, now I understand why they did this as an API preview.  I thought they were just adding a timestamp to an existing struct... I hadn't realized that they entirely moved the repository.  That certainly means this is a breaking change, but I don't think this gets used very often, so I don't think it will be too bad.

rubenvereecken:
Added a little godoc to the StarredRepository and included timestamps in the two relevant tests. All passes.

It's indeed a bigger change than expected but it's just for the one method. Glad you think no harm's done.

willnorris:
looks good, thanks!

PR #200, opened 2015-05-22
Ability to list organization repos

captncraig:
With recent changes to github api it is possible to list organization and user repos in the same list. A special header is required to activate this until the feature is fully rolled out.

https://developer.github.com/changes/2015-01-07-prepare-for-organization-permissions-changes/

This PR allows users to opt-in to this functionality during the migration period.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -122,6 +122,11 @@ type RepositoryListOptions struct {
 	// Default is "asc" when sort is "full_name", otherwise default is "desc".
 	Direction string `url:"direction,omitempty"`
 
+	// Include orginization repositories the user has access to.
+	// This will become the default behavior in the future, but is opt-in for now.
+	// See https://developer.github.com/changes/2015-01-07-prepare-for-organization-permissions-changes/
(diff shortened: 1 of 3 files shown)


github/repos.go
```diff
@@ -146,6 +151,10 @@ func (s *RepositoriesService) List(user string, opt *RepositoryListOptions) ([]R
 		return nil, nil, err
 	}
 
+	if opt != nil && opt.IncludeOrg {
+		req.Header.Set("Accept", "application/vnd.github.moondragon+json")
```
willnorris: would you mind pulling this value out into a const alongside [similar values](https://github.com/google/go-github/blob/4da7e49b0b4273d0becd418b24a2f49f5c56bd66/github/github.go#L38-L44).  Also, go ahead and update the test to pass true, and test for the presence of the header using `testHeader`.

willnorris:
rebased onto master and merged as 34bb95064e99bce6015a98c366c4cd74712cebeb.  thanks!

Issue #91, opened 2014-04-08
Make it easier to create commit statuses from pull requests


tchap:
This just saves some code. The thing is that in the pull request object, there is a field called `statuses_url`, which can be directly used to create new commit statuses. The current API implements `CreateStatus`, which internally generates the same URL, the user just has to take care of getting the right ref, which is, however, already supplied in `statuses_url`. So it could make sense to implement `PostStatus` or so, which would take the whole URL path, e.g. `octocat/Hello-World/statuses/6dcb09b5b57875f334f61aebed695e2e4193db5e`. So the API could be `PostStatus(url string, status *RepoStatus)`. What do you think?

I will implement this anyway since I need it. Perhaps just let me know what do you think about the method name...

PR #202, opened 2015-05-25
Feature add get issue by url

BRMatt:
Several parts of the Github API expose hypermedia links to other API resources. Webhook payloads also make heavy use of URLs to link to to resources affected by the webhook's event.

This PR refactors a few methods to allow consumers to fetch issues and their comments by URL (mainly because I need these specific methods for a project I'm working on). If you'd like this done consistently across the project I'd be happy to help out, but I'm not sure whether it's best to have one mega PR that is hard to grok, or several smaller ones.

I've filled out a CLA.

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -208,7 +208,15 @@ func (s *IssuesService) ListByRepo(owner string, repo string, opt *IssueListByRe
 // GitHub API docs: http://developer.github.com/v3/issues/#get-a-single-issue
 func (s *IssuesService) Get(owner string, repo string, number int) (*Issue, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/issues/%d", owner, repo, number)
-	req, err := s.client.NewRequest("GET", u, nil)
+
+	return s.GetByURL(u)
+}
+
+// Get a single issue by its API url.
+//
+// Intended for following hypermedia links exposed in the API.
+func (s *IssuesService) GetByURL(url string) (*Issue, *Response, error) {
+	req, err := s.client.NewRequest("GET", url, nil)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/issues.go
+++ b/github/issues.go
@@ -214,7 +214,7 @@ func (s *IssuesService) Get(owner string, repo string, number int) (*Issue, *Res
 
 // Get a single issue by its API url.
 //
-// Intended for following hypermedia links exposed in the API.
+// Intended to be used for following hypermedia links exposed in the API.
 func (s *IssuesService) GetByURL(url string) (*Issue, *Response, error) {
 	req, err := s.client.NewRequest("GET", url, nil)
 	if err != nil {
--- a/github/issues_comments.go
+++ b/github/issues_comments.go
@@ -52,7 +52,15 @@ func (s *IssuesService) ListComments(owner string, repo string, number int, opt 
 	} else {
 		u = fmt.Sprintf("repos/%v/%v/issues/%d/comments", owner, repo, number)
 	}
-	u, err := addOptions(u, opt)
+
+	return s.ListCommentsByURL(u, opt)
+}
+
+// ListCommentsByURL lists all issue comments at the specified URL.
+//
+// Intended to be used for following hypermedia links in the API.
+func (s *IssuesService) ListCommentsByURL(url string, opt *IssueListCommentsOptions) ([]IssueComment, *Response, error) {
+	u, err := addOptions(url, opt)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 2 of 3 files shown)


willnorris:
> Several parts of the Github API expose hypermedia links to other API resources

Actually, pretty much ALL of the API exposes hypermedia links... it is very much designed that way.  However, go-github intentionally exposes a more RPC-style interface (see my comment at https://github.com/google/go-github/issues/91#issuecomment-39858455).  So I'd like to keep from littering the entire library with *ByURL methods if we can keep from it.

In the case that you're running into, is following the hyperlinks the only way to get at the data you're trying to access?  For example, you need to be able to fetch an issue by URL.  Does the data you working with not also have the fields needed to call [Issues.Get](https://godoc.org/github.com/google/go-github/github#IssuesService.Get) (the owner, repo name, and issue ID)?  Or was following they hyperlink just a matter of convenience?

One thing to note is that go-github is built in two layers... the lower layer that handles building requests and sending them to GitHub (client.NewRequest and client.Do), and then the higher service layer that wraps that to handle various operations.  That lower layer is intentionally exposed so that users of the library can manually construct and send requests when the service layer doesn't do what they need.  So my first inclination is to say that you should just use that directly, and this doesn't need to be added to the library itself.

That said, I think we could add one additional method to the lower layer to make things like this a little simpler.  Something that combines NewRequest and Do, like:

```go
func (c *Client) GetByURL(url string, v interface{}) (*Response, error) {
	req, err := c.NewRequest("GET", url, nil)
	if err != nil {
		return nil, err
	}

	return c.Do(req, v)
}
```

That does require ever-so-slightly more work to call it, since you need to create a struct to receive the response, but it's not too bad.  So instead of your:

```go
issue, resp, err := client.Issues.GetByURL(url)
```

You would have:

```go
issue := new(github.Issue)
resp, err := client.GetByURL(url, issue)
```

That would also allow you to do something like the following where you don't necessarily know what type to expect in the response:

```go
value := new(json.RawMessage)
resp, err := client.GetByURL(url, value)
```

though I'm not sure how often that would actually be useful in practice.

What do you think?

BRMatt:
Wow, thanks for the rapid & in depth response! :heart: 

> Actually, pretty much ALL of the API exposes hypermedia links... it is very much designed that way.

For sure, you're absolutely right. I guess I was trying to avoid coming in with a "the API is designed for this, api client is wrong for not doing this" mindset.

> However, go-github intentionally exposes a more RPC-style interface (see my comment at #91 (comment))

Thanks, that's very helpful! I'd had a quick look through previous issues to see if there were any with hypermedia-sounding titles, but missed that one. Having read your rationale I can see where you're coming from - adding tonnes of `xByURL` methods would certainly make the API needlessly noisy.

> In the case that you're running into, is following the hyperlinks the only way to get at the data you're trying to access? 

This is for a small bot that fetches a list of all pull requests within an organisation, checks the issue comments to see whether the PR has been reviewed, and assigns appropriate labels to the PR (e.g. "wip", "awaiting-review", "approved"). There's also a webhook component that listens for pings from Github and re-scans the issue.

My initial problem was that there didn't seem to be an easy way to get from an issue returned by `ListByOrg` to the associated comments for that issue. The `Issue` struct doesn't contain any information about the repository, and the Github documentation [suggests that no relevant information is returned](https://developer.github.com/v3/issues/#list-issues). I ended up using a regex to extract the owner and repo from the issue's url. :see_no_evil: 

I had a chat with GH support about this and they mentioned that the ~real~ payload actually includes a `comments_url` field, so my reasoning was I could use this URL to easily get the comments for the issue (but looking at the problem again, `go-github` doesn't actually expose the `comments_url`, so this PR wouldn't have done me a lot of good. I guess that's what happens when you come back to a problem a few weeks later and only implement half of what you were thinking of when you last touched it!).

I've just had another look at the actual json payload returned by the `/orgs/:org/issues` endpoint and it seems that that endpoint actually includes a `repository` payload with each issue that I could use to load the issue's comments, but unfortunately the `Issue` struct doesn't include that field.

> That lower layer is intentionally exposed so that users of the library can manually construct and send requests when the service layer doesn't do what they need

Ohhhh that's neat, I didn't appreciate that when looking through the package! When I saw that `addOptions` is a private function I assumed the end users weren't supposed to use the underlying API themselves.

> That said, I think we could add one additional method to the lower layer to make things like this a little simpler

That definitely sounds like something that would make life easier! Given that the `repository` payload is only available on certain (one?) endpoint I guess it makes sense to keep it out of the struct definition, but I could use the underlying API to get at it in this project. :smile: 

As an aside, one possible problem with adding this helper is that people could expect analogous `POST`/`PUT`/`PATCH` functions, but I guess it may not actually be a problem in 90% of cases.

In light of all this, it seems that this PR isn't appropriate any more - I'll go ahead and close it.

Thanks again for your help!

BRMatt:
Forgot to ask, would you like me to add the `GetByURL` func to the client?

willnorris:
I'm fine with adding a Repository field to Issue, just add a comment that says that it's only populated in some API responses.

Additionally, feel free to add `client.GetByURL` if that would be useful for you.  I've long thought about doing something like this to better handle pagination links (the way we currently do it is specifically recommended against)

PR #206, opened 2015-06-03
Feature expose request helpers

BRMatt:
Follows on from [discussion in #202](https://github.com/google/go-github/pull/202#issuecomment-105300075).

The main difference is that GetByURL also accepts a ListOptions struct. This was done to allow consumers to re-use the existing Options structs without interpolating raw values into urls.

I've also changed the `Response` struct to expose links that were defined in the header, which makes it easier for consumers to paginate through results. The implementation isn't very clean, but I don't get a huge amount of time to work on this and wanted to get your feedback before running down a rabbit hole!

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -322,6 +322,29 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	return response, err
 }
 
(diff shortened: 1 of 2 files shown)


gmlewis:
Closing year-old PR as inactive.

PR #209, opened 2015-06-05
Adding DownloadReleaseAsset

### LINKED ISSUE #207: Dowloading release assets
I'm trying to use `go-github` to download assets from a release. Here's the code I have:
```go
// Based on https://developer.github.com/v3/repos/releases/#get-a-single-release-asset
func githubDownloadReleaseAsset(asset *github.ReleaseAsset, out io.Writer) error {
	gh := githubClient() // Initializes *github.Client

	req, err := gh.NewRequest("GET", *asset.URL, nil)
	req.Header.Set("Accept", "application/octet-stream")
	if err != nil {
		return err
	}

	_, err = gh.Do(req, out)
	return err
}
```
It seems the GitHub API redirects to a pre-signed URL for an S3 bucket, but I'm getting a `400` error, and I can't quite figure out why. Any help would be much appreciated.

Bowbaq:
Fixes #207

Adds `DownloadReleaseAsset` to `RepositoriesService`

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -8,6 +8,7 @@ package github
 import (
 	"errors"
 	"fmt"
+	"io"
 	"mime"
 	"os"
 	"path/filepath"
@@ -209,6 +210,29 @@ func (s *RepositoriesService) GetReleaseAsset(owner, repo string, id int) (*Rele
 		return nil, resp, nil
 	}
 	return asset, resp, err
+}
+
+// DownloadReleaseAsset downloads a release asset.
+//
+// DownloadReleaseAsset returns an io.ReadCloser that reads the contents of the
+// specified release asset. It is the caller's responsibility to close the ReadCloser.
+//
+// GitHub API docs : http://developer.github.com/v3/repos/releases/#get-a-single-release-asset
+func (s *RepositoriesService) DownloadReleaseAsset(owner, repo string, id int) (io.ReadCloser, error) {
+	u := fmt.Sprintf("repos/%s/%s/releases/assets/%d", owner, repo, id)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", defaultMediaType)
+
+	resp, err := s.client.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.Body, nil
 }
 
 // EditReleaseAsset edits a repository release asset.
--- a/github/repos_releases_test.go
+++ b/github/repos_releases_test.go
@@ -6,8 +6,10 @@
 package github
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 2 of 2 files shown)


Bowbaq:
I signed it!

github/repos_releases.go
```diff
@@ -211,6 +213,54 @@ func (s *RepositoriesService) GetReleaseAsset(owner, repo string, id int) (*Rele
...
+	if s.client.client.Transport == nil {
+		resp, err = http.DefaultTransport.RoundTrip(req)
+	} else {
+		resp, err = s.client.client.Transport.RoundTrip(req)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	// GitHub API streamed the asset directly
```
Bowbaq: I wasn't able to get GitHub's API to exhibit that behavior, so this is based on my interpretation of the documentation

zachgersh:
Hey @Bowbaq - :+1: on the pull.

You've got a small typo `DowloadReleaseAsset` instead of `DownloadReleaseAsset`

Bowbaq:
@zachgersh fixed the typo, good catch

jtarchie:
Can this be accepted? We are on a branch that relies on this feature. We'd like to use latest changes from the official repo though.

parkr:
LGTM.

github/repos_releases.go
willnorris: Go's default HTTP client will transparently follow redirects (up to 10 hops), so you don't need to differentiate between streamed directly and redirected assets... from the clients perspective, they will all look like they are streamed directly.  Instead of digging into the `s.client.client.Transport` above, just call `s.client.client.Get` as is done in [DownloadContents](https://github.com/google/go-github/blob/9420d0f48acb838ccf8a7f8dde9eaee19af71132/github/repos_contents.go#L112).

willnorris: err, ignore that point about using `s.client.client.Get`... I think it can be simplified even further... lemme poke around a little more (I haven't touched this code in a while)

willnorris: okay, you do need to call `s.client.client.Get`.  I was initially thinking about passing a `io.Writer` to `s.client.Do`, but that won't work because it tries to read and copy the entire response body, which isn't what we want here.

github/repos_releases.go
```diff
@@ -211,6 +212,29 @@ func (s *RepositoriesService) GetReleaseAsset(owner, repo string, id int) (*Rele
...
+func (s *RepositoriesService) DownloadReleaseAsset(owner, repo string, id int) (io.ReadCloser, error) {
+	u := fmt.Sprintf("repos/%s/%s/releases/assets/%d", owner, repo, id)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", defaultMediaType)
+
+	resp, err := s.client.client.Do(req)
```
Bowbaq: @willnorris ended up using `s.client.client.Do` because we need the custom `Accept` header

willnorris: ah yeah... I forgot about the custom header.

willnorris:
rebased onto master and merged as 25b0481acbe9f2123f2b39c3235b156cff6296cd.  Sorry about the crazy long time it took to get this reviewed.

PR #210, opened 2015-06-12
Added 'Before' field to NotificationListOptions

erwinvaneyk:
Before is a missing field in the NotificationListOptions, as it is a documented parameter according to https://developer.github.com/v3/activity/notifications/#list-your-notifications


erwinvaneyk:
I signed it!

willnorris:
thanks, rebased onto master and merged as c801b653c20d36a7cb3cf219daf3fe170f0078cf

PR #211, opened 2015-06-12
Fixed method name

diezcami:
Changed ListEventsRecievedByUser to ListEventsReceivedByUser.
Just a minor typo fix, but I figured it'd be better not to have it in the long run when more people are using the package.


willnorris:
thanks!  rebased onto master and merged as 1bbee11acfa2fba812381a79bbc70960f2c72a32

Issue #191, opened 2015-04-29
Add support for Feeds API


willnorris:
huh, not sure how I never noticed this one before... I'm not entirely sure when it was added.

Docs: https://developer.github.com/v3/activity/feeds/

Notably, this method requires basic auth, so we'll need to have a recommended way for clients to do that, whether we actually provide a `BasicAuthTransport` ourselves or not.  (see also: https://github.com/google/go-github/issues/119#issuecomment-62276682)

PR #216, opened 2015-06-23
Implementation of Feeds API

saromanov:
Hi!
I found that [Feeds API](https://developer.github.com/v3/activity/feeds/) is missing and just implemented it. The main problem was, what to do with basic auth and at the end, i decided to do as with oauth2 example, i.e make client outside go-github.
```go
type Transport struct {
	Username string
	Password string
}
func(t* Transport) RoundTrip(r *http.Request)(*http.Response, error) {
	r.SetBasicAuth(t.Username, t.Password)
	return http.DefaultTransport.RoundTrip(r)
}

func main() {
  client := &http.Client{Transport: &Transport{"username", "password"}}
  cl := github.NewClient(client)
  res, _, _ := cl.Feeds.ListFeeds()
}

```
I don't know, probably this is not best solution, but i don't see any reasons to put basic auth to go-github.
What do you think?

### DIFF
--- a/github/feeds.go
+++ b/github/feeds.go
@@ -1,0 +1,61 @@
+package github
+
+//FeedsService handles communication with the feeds related
+//methods of the GitHub API.
+//Note. client should provide basic basic auth.
+//
+//Github API docs: https://developer.github.com/v3/activity/feeds/
+type FeedsService struct {
+	client *Client
+}
+
+//Feeds represents the all the feeds available to the authenticated user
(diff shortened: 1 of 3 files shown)


bithavoc:
I like this :+1:

willnorris:
Looks like this got fixed in #315.  Apparently it got reimplemented by someone else because this PR didn't mention #191, so it didn't show up as a pending pull request on that issue.

PR #221, opened 2015-07-20
TokenSource provider for the basic auth authorizations api

mcqueenorama:
This provides a TokenSource to I can use the api documented here:

https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization

Here's a gist showing sample usage:

https://gist.github.com/mcqueenorama/c63f0c2aa332558515d5


mcqueenorama:
I signed the CLA just now.

mcqueenorama:
Trying to get the CLA to go through.  I signed it.

mcqueenorama:
Added my email address to my commits, but its not on the first several.  Shall I close this and redo it?

github/basic_auth_token.go
```diff
@@ -0,0 +1,199 @@
...
+//	https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization
+//
+type Creds struct {
+	Username       string
+	Password       string
+	PostBodyReader io.Reader
+	Config         *oauth2.Config
+}
+
+// Token set username/password and postbody and do nasic auth
```
wallclockbuilder: +// Token set username/password and postbody and do ~~nasic~~ basic auth

mcqueenorama:
How about merging this thing?

mcqueenorama:
Can you guys merge this?

willnorris:
Really sorry for the delayed reply.  There are a few issues with the pull request as written:

From a high level, I don't think the basic auth stuff should be buried just inside the authorizations call, since basic auth can actually be used throughout the API.  It seems much cleaner to treat it exactly the same as we do OAuth authentication today... you simply provide an `http.Client` to `NewClient` that is capable of adding the necessary auth headers.  See more on #258 and 8f3e741 which added a `BasicAuthTransport` (which includes support for two-factor auth).

So that was just the BasicAuth transport.  The thing I hadn't gotten to was actually implementing all of the OAuth Authorizations API methods.  BasicAuth support was previously blocking that, but it should be easily doable now.  I've got a patch started which I'll try and get submitted soon.

I've got one or two more comments about this PR in particular, which are mostly for future reference at this point, but I'll go ahead and point those out individually.

github/basic_auth_token.go
```diff
@@ -0,0 +1,199 @@
...
+	}, nil
+
+}
+
+// retrieveTokenBasicAuth - Do Basic auth via username/password not client_id/client_secret
+//
+// POST body for the auth call comes from the caller as a Reader
+func retrieveTokenBasicAuth(TokenURL, Username, Password string, postBodyReader io.Reader) (*oauth2.Token, error) {
+
+	hc := http.Client{}
```
willnorris: this is a problem.  All API calls have to be made with the client that is provided to `NewClient`.  That's because a zero-value `http.Client` won't work in certain environments, notable App Engine as well as Google's prod environment.

github/basic_auth_token.go
```diff
@@ -0,0 +1,199 @@
...
+//
+type Creds struct {
+	Username       string
+	Password       string
+	PostBodyReader io.Reader
+	Config         *oauth2.Config
+}
+
+// Token set username/password and postbody and do basic auth
+func (ba Creds) Token() (tk *oauth2.Token, err error) {
```
willnorris: Having a `TokenSource` that transparently does the token exchange using Basic Auth is certainly clever, and might be something we want to support.  But if we do, we should first implement all of the [OAuth Authorizations](https://developer.github.com/v3/oauth_authorizations) endpoints as a regular API service like everything else in go-github, and then add a thin wrapper around that that does the necessary calls.

willnorris:
I'm going to go ahead and close the pull request since we now have pretty good support for BasicAuth and the OAuth Authorizations endpoints.  

The one kinda interesting thing this pull request did have which isn't in go-github is the TokenSource that transparently does the token exchange using BasicAuth.  I'd be open to discussing that a little further if anyone feels really strongly about it, but I have concerns about how well that would work in practice with 2fa.  It would also need to be rewritten to use the current BasicAuth and OAuth endpoint methods.  So if someone wants to attempt that, feel free to file a new issue to discuss the approach.

PR #227, opened 2015-08-04
Add date to the User struct, it's sometimes used for committer and author dates

brendandburns:
e.g. https://api.github.com/repos/googlecloudplatform/kubernetes/commits/7ad83079263ea73e022936887e67f267e41fd211

### DIFF
--- a/github/users.go
+++ b/github/users.go
@@ -43,6 +43,8 @@ type User struct {
 	DiskUsage         *int       `json:"disk_usage,omitempty"`
 	Collaborators     *int       `json:"collaborators,omitempty"`
 	Plan              *Plan      `json:"plan,omitempty"`
+        // Date sometimes occurs when users take action like author a commit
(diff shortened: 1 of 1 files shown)


brendandburns:
An alternate fix would be to change `RepositoryCommit.Comitter` to be a `CommitAuthor` rather than a `User`, let me know which one you prefer.

brendandburns:
or, even have `PullRequest.ListCommits(...)` return a `[]Commit` instead of `[]RepositoryCommit`

willnorris:
Are you sure you don't mean to be fetching the `RepositoryCommit.Commit.Committer`, which is a `CommitAuthor`?  As opposed to the `RepositoryCommit.Committer`, which is a `User`.

brendandburns:
Ah, you are correct!  sorry for the noise!

PR #229, opened 2015-08-15
update repos example to use RateLimits

yichengq:
The RateLimit() func is deprecated, and should not appear in the
official example.

Ref: http://godoc.org/github.com/google/go-github/github#Client.RateLimit


willnorris:
thanks!  rebased onto master and merged as 3ec5f1a72fe1c26a8a7abeefcabbd2e9f6153d99

Issue #232, opened 2015-09-11
Add more more location-specific fields to the PullRequestComment struct


ojarjur:
Pull request comments have multiple fields that are important for the location of a comment but that are currently omitted from that struct:

    * diff_hunk: This is important as the position field is actually the position *within-the-hunk* rather than within the file.
    * original_commit_id: This is important, as "commit_id" tracks the pull request, so after additional changes are pushed to the pull-request, the "commit_id" changes and no longer points to the commit at which the comment was made.
    * original_position: This is similar to original_commit_id, and is important as the "position" field gets cleared out after the pull request is updated (since the comment is no longer at the current commit).

willnorris:
fixed in 08bb599b461f73841736d7269029562423ae9037

PR #233, opened 2015-09-11
Add comment location fields to the PullRequestComment struct.

ojarjur:
This change adds the fields "DiffHunk", "OriginalPosition", and "OriginalCommitId" to the PullRequestComment struct. These fields provide important information about the location of a comment.

See issue #232 for more details.


willnorris:
squashed and merged as 08bb599b461f73841736d7269029562423ae9037

PR #235, opened 2015-09-18
Fix incorrect URLs for milestone APIs

difro:
API URLs starting with "/" does not play well with Github Enterprise,
whose API endpoint looks like "http(s)://hostname/api/v3/".

This fixes incorrect URLs in miletones APIs.


willnorris:
ah, good catch.  thanks!  rebased onto master and merged as 1ee11e5972575f291558c46d04daca566b2845cd

Issue #238, opened 2015-10-03
add ListOptions to ListWatched 


bradrydzewski:
I'd like the ability to provide `ListOptions` when getting the list of watched repositories:

```diff
-func (s *ActivityService) ListWatched(user string)
+func (s *ActivityService) ListWatched(user string, opt *ListOptions)
```

I can have a pull request available next week

willnorris:
sounds good to me

sona-tar:
:+1:

willnorris:
fixed in  c57bb02

PR #240, opened 2015-10-17
Add ListOptions to ListWatched

sona-tar:
Supoprt #238

### DIFF
--- a/github/activity_watching.go
+++ b/github/activity_watching.go
@@ -50,13 +50,14 @@ func (s *ActivityService) ListWatchers(owner, repo string, opt *ListOptions) ([]
 // the empty string will fetch watched repos for the authenticated user.
 //
 // GitHub API Docs: https://developer.github.com/v3/activity/watching/#list-repositories-being-watched
-func (s *ActivityService) ListWatched(user string) ([]Repository, *Response, error) {
+func (s *ActivityService) ListWatched(user string, opt *ListOptions) ([]Repository, *Response, error) {
 	var u string
 	if user != "" {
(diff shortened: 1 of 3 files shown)


github/activity_watching.go
```diff
@@ -50,13 +50,14 @@ func (s *ActivityService) ListWatchers(owner, repo string, opt *ListOptions) ([]
...
 // GitHub API Docs: https://developer.github.com/v3/activity/watching/#list-repositories-being-watched
-func (s *ActivityService) ListWatched(user string) ([]Repository, *Response, error) {
+func (s *ActivityService) ListWatched(user string, opt *ListOptions) ([]Repository, *Response, error) {
 	var u string
 	if user != "" {
 		u = fmt.Sprintf("users/%v/subscriptions", user)
 	} else {
 		u = "user/subscriptions"
 	}
+	u, err := addOptions(u, opt)
```
parkr: Does this handle the `nil` case for `opt`? Could the tests be amended to ensure this?

sona-tar:
I fixed it!

sona-tar:
@willnorris  ping.
 Can we get feedback on this PR?

github/activity_watching.go
willnorris: yes, `addOptions` handles nil values.  I see now that we don't any tests for `addOptions`, but that would be the place to test this, not in every method that happens to call it (which is hundreds).

willnorris:
made above changes I mentioned and merged as c57bb02.  Thanks!

Issue #119, opened 2014-06-11
authorizations API support


rgarcia:
https://developer.github.com/v3/oauth_authorizations/

One tricky part is that it only accepts basic auth, so I think it'd require users of the library to implement a simple RoundTripper that added this header (and then use that as the HTTP client transport for go-github).

PR #241, opened 2015-10-18
Support OAuth Authorization API.

sona-tar:
#119 

I'm writing tests for OAuth Authrization API.
Please wait for next weekend.

Sample Code
```go
package main                                                                          
                                                                                      
import (                                                                              
    "flag"                                                                            
    "fmt"                                                                             
    "os"                                                                              
)                                                                                     
                                                                                      
import (                                                                              
    "net/http"                                                                        
)                                                                                     
                                                                                      
type Transport struct {                                                               
    Username string                                                                   
    Password string                                                                   
}                                                                                     
                                                                                      
func (t Transport) RoundTrip(req *http.Request) (*http.Response, error) {             
    req.SetBasicAuth(t.Username, t.Password)                                          
    return http.DefaultTransport.RoundTrip(req)                                       
}                                                                                     
                                                                                      
func (t *Transport) Client() *http.Client {                                           
    return &http.Client{Transport: t}                                                 
}                                                                                     
                                                                                      
const Version string = "0.0.1"                                                        
                                                                                      
func main() {                                                                         
    var user *string = flag.String("user", "", "Github user name")                    
    var pass *string = flag.String("password", "", "Github password")                 
                                                                                      
    flag.Parse()                                                                      
                                                                                      
    if len(*user) == 0 || len(*pass) == 0 {                                           
        flag.Usage()                                                                  
        os.Exit(1)                                                                    
    }                                                                                 
                                                                                      
    // Basic Auth                                                                     
    t := &Transport{                                                                  
        Username: *user,                                                              
        Password: *pass,                                                              
    }                                                                                 
    client := github.NewClient(t.Client())                                            
                                                                                      
    // Get                                                                            
    opt := &github.ListOptions{PerPage: 100}                                          
    auths, _, err := client.OAuth.ListAuthorizations(opt)                             
    if err != nil {                                                                   
        fmt.Printf("Error : client.OAuth.ListAuthorizations %s\n", err.Error())       
        os.Exit(0)                                                                    
    }                                                                                 
                                                                                      
    for _, auth := range auths {                                                      
        fmt.Printf("ID : %v\n", *auth.ID)                                             
        a, _, err := client.OAuth.GetSingleAuthorization(*auth.ID)                    
        fmt.Printf("ID : %v\n", *a.ID)                                                
        if err != nil {                                                               
            fmt.Printf("Error : GetSingleAuthorization %s\n", err.Error())            
            os.Exit(1)                                                                
        }                                                                             
    }                                                                                 
    auth_req := &github.AuthorizationRequest {                                   
        Scopes: &[]string{"public_repo"},                                        
        Note:   github.String("go-github tset"),                                 
    }                                                                            
    new_auth, _, err := client.OAuth.CreateAuthorization(auth_req)               
    if err != nil {                                                              
        fmt.Printf("Error : client.OAuth.CreateAuthorization %s\n", err.Error()) 
        os.Exit(1)                                                               
    }                                                                            
    fmt.Printf("ID : %v\n", *new_auth.ID)                                        
}                                                                             
```

```sh
go run gauth.go -user YOUR_USER -password YOUR_PASS
```

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -71,6 +71,7 @@ type Client struct {
 	Rate Rate
 
 	// Services used for talking to different parts of the GitHub API.
+	OAuth         *OAuthService
 	Activity      *ActivityService
 	Gists         *GistsService
 	Git           *GitService
@@ -133,6 +134,7 @@ func NewClient(httpClient *http.Client) *Client {
 	uploadURL, _ := url.Parse(uploadBaseURL)
 
 	c := &Client{client: httpClient, BaseURL: baseURL, UserAgent: userAgent, UploadURL: uploadURL}
+	c.OAuth = &OAuthService{client: c}
 	c.Activity = &ActivityService{client: c}
 	c.Gists = &GistsService{client: c}
 	c.Git = &GitService{client: c}
@@ -323,6 +325,7 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 			err = json.NewDecoder(resp.Body).Decode(v)
 		}
 	}
+
 	return response, err
 }
 
--- a/github/oauth.go
+++ b/github/oauth.go
@@ -1,0 +1,280 @@
+// Copyright 2015 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"fmt"
+)
+
+// OAuthService handles communication with the OAuth related
+// methods of the GitHub API.
+// You can only access this API via Basic Authentication using your username and password, not tokens.
+//
+// GitHub API docs: http://developer.github.com/v3/oauth_authorizations/
+type OAuthService struct {
+	client *Client
+}
+
+// Authorization represents a GitHub authorized infomaration
+type Authorization struct {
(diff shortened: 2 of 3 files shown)


sona-tar:
I finished tests.
Please review.

willnorris:
So I'll be merging this in just a moment.  Since it took so long for me to get to reviewing this, I just made a number of changes myself, rather than leave comments.  Most were relatively minor stylistic things:

 - rename to OAuthService to AuthorizationsService 
 - rename oauth.go file to authorizations.go
 - combined GetOrCreate and GetOrCreateFingerprint into single method and rename to GetOrCreateForApp to distinguish it from the Get method which is getting an authorization by ID rather than by application client ID.
 - removed RevokeAll method, as it's not actually documented in the GitHub API (maybe it used to be?)
 - updated several methods that don't actually return an Authorization (Delete, Revoke)
 - simplified values in tests (having long strings doesn't actually test anything more than a single character does)
 - renamed Update method to Edit for consistency with the rest of the library
 - several small changes like variable renames for consistency with the rest of the library
 
If it looks like I missed something or messed something up, certainly let me know.

sona-tar:
:+1:

PR #247, opened 2015-11-20
github: document and add support for all IssueEvent types

rsc:
I processed every IssueEvent from the entire history of the golang/go repository and these are all the ones I found. So I have some belief that they might be a complete set. :-)


rsc:
Re CLA, I should be whitelisted as part of Google Inc. Thanks.

willnorris:
lgtm, thanks.

PR #248, opened 2015-11-24
Fix default comment for Direction in issue lists

eparis:
Both:
https://developer.github.com/v3/issues/#list-issues
https://developer.github.com/v3/issues/#list-issues-for-a-repository

State that the default is `desc` for direction. (As does my results). So
update the comment.


willnorris:
oops, thanks for the fix!  cherry-picked and merged as 3eb9d5d

PR #250, opened 2015-11-27
Do not try to unmarshall response when http response code is 204

uzzz:
I'm not completely sure if it's correct semantics in my PR, but I had such bug: some GH API endpoints (contributors collection endpoint in my case) returns 204 No Content when there's no content, which leads to an EOF error. Probably it would be better to not to try unmarshall data to avoid error. 

Code to reproduce:

```go
package main

import (
	"fmt"

)

func main() {
	_, _, err := client.Repositories.ListContributors("cloudfoundry", "em-posix-spawn", nil)
	if err != nil {
		fmt.Println("An error has occured:", err.Error())
	}
}
```

```
$ go run reproduce.go
An error has occured: EOF
```

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -316,7 +316,7 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		return response, err
 	}
 
-	if v != nil {
+	if v != nil && resp.StatusCode != http.StatusNoContent {
 		if w, ok := v.(io.Writer); ok {
 			io.Copy(w, resp.Body)
 		} else {
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -430,6 +430,24 @@ func TestDo_rateLimit_errorResponse(t *testing.T) {
 	}
 }
 
+func TestDo_noContent(t *testing.T) {
+	setup()
+	defer teardown()
+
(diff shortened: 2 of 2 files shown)


parkr:
LGTM.

willnorris:
so the real problem here isn't just when it's a 204 response, but rather when there is an empty response body, right?  For example, if I change the test case to return a `202 Accepted`, it still causes an error, albeit a different one than the 204 response:

    github_test.go:447: Do returned unexpected error: invalid character 'N' looking for beginning of value

So I think a better fix would be not to specifically look at the status code, but rather to verify that the body is non-empty, check the error returned from Decode, or something similar.  I'll play around with it and see what I can come up with.

willnorris:
err, I just noticed that you're using `http.Error` to return the status, so you're not actually returning an empty response body in the test... in your case, you're returning the body "No Content".  So that explains the `invalid character 'N'` message I was getting.

Though it still seems that this method is failing on 204 responses like you originally reported.

willnorris:
okay, I change this to specifically check for `io.EOF` errors coming from the json decoding call.  Merged as c9c9f3beeac50d26ec5440b8da68efeb9506d25e.  Thanks!

Issue #251, opened 2015-12-17
Add url prop to Hook struct


selfcontained:
When retrieving repo/org hooks, either by id or a list, the github api returns a `url` property.  Currently the `Hook` struct doesn't map this to a `URL` prop.  This would be really useful to expose, wondering if there is a reason it isn't currently?  I'm happy to open a PR to add this if not.

willnorris:
fixed in 59f41ca

PR #252, opened 2015-12-18
adding URL property to Hook struct

selfcontained:
This PR adds the `URL` property to the `Hook` struct to map the existing `url` property returned by the Github API for listing hooks or retrieving them by id.

Addresses #251

### DIFF
--- a/github/repos_hooks.go
+++ b/github/repos_hooks.go
@@ -70,6 +70,7 @@ type Hook struct {
 	CreatedAt *time.Time             `json:"created_at,omitempty"`
 	UpdatedAt *time.Time             `json:"updated_at,omitempty"`
 	Name      *string                `json:"name,omitempty"`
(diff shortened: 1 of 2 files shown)


parkr:
You'll want to add a test for this, I believe. Otherwise, :+1: .

selfcontained:
Wasn't sure what the precedence was for testing individual props.  It doesn't look like there are tests around any of them specifically, so I added checks for `URL` into the existing hook list/get tests.  Lemme know if there's an alternate approach you'd prefer.

willnorris:
As Brad notes, we don't have great test coverage for every struct field.  In a number of places, we have a single test that does that, but for the rest of the tests we use minimally populated structs.  The integration tests also provide coverage against the live API for a handful of types.

Anyway, cherry-picked the initial commit as 59f41ca3f75660d005283021d0c00bd0dd0da756. Thanks.

PR #255, opened 2015-12-29
Fix data race surrounding Client.Rate.

### LINKED ISSUE #253: github.Client is not safe for concurrent use; is that intended or a bug?
From https://godoc.org/net/http:

> Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

https://godoc.org/github.com/google/go-github/github#Client makes no such claims, which is why I'm unsure if this is a bug or intended behavior.

However, given that it looks and feels a lot like `http.Client`, I thought it might also support concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

That's not the case. If you look at the source of [`github.Client.Do`](https://github.com/google/go-github/blob/44b1ede22d71e26fa5d7af7a50204f7e6daadfc8/github/github.go#L310):

```
c.Rate = response.Rate
```

It writes to `c.Rate` without any mutexes, which is not safe and will cause data races:

```
WARNING: DATA RACE
Write by goroutine 93:
  github.com/google/go-github/github.(*Client).Do()
      .../src/github.com/google/go-github/github/github.go:310 +0x16b
```

Is that intended, or is this a bug?

If I want to make concurrent calls to GitHub API via multiple goroutines, should I create multiple `github.Client`s?

shurcooL:
Previously, concurrent calls to `Client.Do` would have data races writing `Client.Rate`. Protect it with a mutex.

Resolves #253.

I've been using/testing it for the last week or two, and I haven't spotted any other issues with concurrency.

Technically, this is a ever-so-slightly breaking API change, since `Rate` was previously an exported struct field, but now it's a method, so callers will need to change `client.Rate` to `client.Rate()`. I don't see any other way around this. Aside from exporting the mutex and keeping `Rate` as is, but that would require the user to know about the mutex and lock it when reading `Rate`. That seems very unfriendly API.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -17,6 +17,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/go-querystring/query"
@@ -64,11 +65,8 @@ type Client struct {
 	// User agent used when communicating with the GitHub API.
 	UserAgent string
 
-	// Rate specifies the current rate limit for the client as determined by the
-	// most recent API call.  If the client is used in a multi-user application,
-	// this rate may not always be up-to-date.  Call RateLimits() to check the
-	// current rate.
-	Rate Rate
+	rateMu sync.Mutex
+	rate   Rate
 
 	// Services used for talking to different parts of the GitHub API.
 	Activity      *ActivityService
@@ -292,6 +290,17 @@ func (r *Response) populateRate() {
 	}
 }
 
+// Rate specifies the current rate limit for the client as determined by the
+// most recent API call.  If the client is used in a multi-user application,
+// this rate may not always be up-to-date.  Call RateLimits() to check the
+// current rate.
+func (c *Client) Rate() Rate {
+	c.rateMu.Lock()
+	rate := c.rate
+	c.rateMu.Unlock()
+	return rate
+}
+
 // Do sends an API request and returns the API response.  The API response is
 // JSON decoded and stored in the value pointed to by v, or returned as an
 // error if an API error has occurred.  If v implements the io.Writer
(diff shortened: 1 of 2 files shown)


shurcooL:
So far, it's just a simple patch for `client.Rate` without any documentation change, because I wanted to ask you first if you'd want to add something like this?

```diff
 // A Client manages communication with the GitHub API.
+//
+// Clients are safe for concurrent use by multiple goroutines, but a single Client
+// shouldn't be used to make calls on behalf of different users.
 type Client struct {
```

Also, due to the nature of this PR, it should be reviewed more carefully than a simple GitHub API coverage addition.

github/github.go
```diff
@@ -64,11 +65,8 @@ type Client struct {
 	// User agent used when communicating with the GitHub API.
 	UserAgent string
 
-	// Rate specifies the current rate limit for the client as determined by the
-	// most recent API call.  If the client is used in a multi-user application,
-	// this rate may not always be up-to-date.  Call RateLimits() to check the
-	// current rate.
-	Rate Rate
+	rateMu sync.Mutex
```
willnorris: any reason not to use a `sync.RWMutex` here so that Rate() can just perform a read lock?

shurcooL: Yes, I chose `sync.Mutex` over `sync.RWMutex` here on purpose, because I estimated that it'd be simpler and faster, and hence more appropriate.

I've seen some arguments and articles demonstrating that the more complex read-write locks are often slower than the simpler mutexes for basic use cases:

- https://blogs.oracle.com/roch/entry/beware_of_the_performance_of
- http://preshing.com/20111124/always-use-a-lightweight-mutex/
- http://askldjd.com/2011/03/06/performance-comparison-on-reader-writer-locks/

I was convinced to at least use `sync.Mutex` by default, and upgrade to `sync.RWMutex` after doing profiling and ensuring it's actually faster and better.

If you prefer I change this to `sync.RWMutex`, I can do that.

willnorris: no, this is fine.

Though I would certainly be curious to see benchmarks of the two, not just for this library but in general.  I just wonder how much difference it really makes in practice.

jabley:
:1:

Thank you, I was just running https://github.com/jabley/project-status and hitting the race condition. Quick update and sorted.

PR #259, opened 2016-01-02
Add initial support for protected branches

bjoernhaeuser:
For more information about the preview see here:
https://developer.github.com/changes/2015-11-11-protected-branches-api/

This adds only retrieving the information and not setting it. Setting
will come in a later commit (if that one here is okay).

I tried to follow the naming guidelines as much as possible and
hopefully I made it right the first time. :-)


willnorris:
yep, this looks great.  Thanks!

PR #261, opened 2016-01-05
Add "SuspendedAt" field to User struct.

griffin-stewie:
There is "suspended_at" key in "/users/:username" response  when I use GitHub Enterprise. so I added "SuspendedAt" field to User struct same as `CreatedAt` & `UpdatedAt`.


willnorris:
cherry-picked as 6387a9448dec3a903271745e7baddd14e0ba90c9. thanks

PR #273, opened 2016-02-01
add RepositoriesService.ListByUser

husio:
Implementation of listing all user repositories as described by
https://developer.github.com/v3/repos/#list-user-repositories


husio:
I signed it!

willnorris:
perhaps it's not the best named, but this is what [`RepositoriesService.List`](https://godoc.org/github.com/google/go-github/github#RepositoriesService.List) is.

husio:
You're right, I missed it.

PR #276, opened 2016-02-11
Minor Rate, RateLimits documentation fixes.

shurcooL:
- Rate is now a method, not a field.
- Mention RateLimits rather than RateLimit since the latter is deprecated.
- Use Go convention for marking deprecated identifiers. See https://github.com/golang/blog/commit/257114af91a0defb1fc2c16c7f4ae2429b8a4e0f.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -75,7 +75,7 @@ type Client struct {
 	UserAgent string
 
 	rateMu sync.Mutex
-	rate   Rate
+	rate   Rate // Rate limit for the client as determined by the most recent API call.
 
 	// Services used for talking to different parts of the GitHub API.
 	Activity      *ActivityService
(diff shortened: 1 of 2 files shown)


github/github.go
```diff
@@ -487,7 +487,7 @@ func (r RateLimits) String() string {
 	return Stringify(r)
 }
 
-// RateLimit is deprecated.  Use RateLimits instead.
+// RateLimit is deprecated.  Deprecated: Use RateLimits instead.
```
shurcooL: Maybe I could remove the first sentence altogether? I know `golint` would complain, but it's not very helpful.

https://godoc.org/image/jpeg#Reader does that, it just says "Deprecated: Reader is deprecated.".

willnorris: yeah, according to golang/blog@257114a it has to appear at the beginning of the paragraph.

shurcooL: All right, I've updated it to be more like https://godoc.org/image/jpeg#Reader. It follows the second form as described in https://github.com/golang/go/issues/10909#issuecomment-136492606:

> ```
> 2. long form 1, used with entities lacking docs
> // Deprecated: T is deprecated.
> type T struct {
>      // Deprecated: for backward compatibility FieldC is still populated,
>      // but new code should use FieldD instead.
>      FieldC int32
> 
>      FieldD int64
> }
> ```

PTAL.

PR #277, opened 2016-02-11
Add RateLimitError type, detect and return it when appropriate.

### LINKED ISSUE #152: Make it easer to detect rate limit errors

shurcooL:
Fixes #152.

Add tests for `RateLimitError`.

Refactor `populateRate` into a more general `parseRate`.

Here's a sample error returned when hitting the rate limit:

```
GET https://api.github.com/users/shurcooL: 403 API rate limit exceeded for xxx.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.); rate reset in 15m8.193650147s
```

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -434,6 +434,9 @@ func (e *Error) Error() string {
 // the 200 range.  API error responses are expected to have either no response
 // body, or a JSON response body that maps to ErrorResponse.  Any other
 // response body will be silently ignored.
+//
+// The error type will be *RateLimitError for rate limit exceeded errors,
+// and *TwoFactorAuthError for two-factor authentication errors.
 func CheckResponse(r *http.Response) error {
 	if c := r.StatusCode; 200 <= c && c <= 299 {
 		return nil
--- a/github/github.go
+++ b/github/github.go
@@ -236,7 +236,7 @@ type Response struct {
 func newResponse(r *http.Response) *Response {
 	response := &Response{Response: r}
 	response.populatePageValues()
-	response.populateRate()
+	response.Rate = parseRate(r)
 	return response
 }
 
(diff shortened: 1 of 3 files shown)


github/github.go
```diff
@@ -427,10 +443,18 @@ func CheckResponse(r *http.Response) error {
 	if err == nil && data != nil {
```
shurcooL: Maybe it's worth expanding documentation of `CheckError` to mention both `TwoFactorAuthError ` and `RateLimitError` error types?

github/github.go
```diff
@@ -427,10 +443,18 @@ func CheckResponse(r *http.Response) error {
...
 		json.Unmarshal(data, errorResponse)
 	}
-	if r.StatusCode == http.StatusUnauthorized && strings.HasPrefix(r.Header.Get(headerOTP), "required") {
+	switch {
+	case r.StatusCode == http.StatusUnauthorized && strings.HasPrefix(r.Header.Get(headerOTP), "required"):
 		return (*TwoFactorAuthError)(errorResponse)
+	case r.StatusCode == http.StatusForbidden && r.Header.Get(headerRateRemaining) == "0" && strings.HasPrefix(errorResponse.Message, "API rate limit exceeded for "):
+		return &RateLimitError{
+			Rate:     parseRate(r),
```
shurcooL: Technically, it's possible to avoid having to parse rate from `*http.Response` here. It's already done previously in `Do` when it does:

```Go
response := newResponse(resp) // response.Rate is parsed and available here.

c.rateMu.Lock()
c.rate = response.Rate
c.rateMu.Unlock()

err = CheckResponse(resp) // CheckResponse is not given the Rate, so it has to parse it from resp a second time...
```

The problem is that `func CheckResponse(r *http.Response) error` is exported and its signature does not accept the rate.

However, it might be possible to create an unexported `func checkResponseWithRate(r *http.Response, rate Rate) error` and use that in `Do` instead.

github/github.go
willnorris: yeah, probably so.

github/github.go
willnorris: I'm actually fine either way.  It's unlikely the extra call will have a measurable difference (though I guess we could benchmark if we really wanted to be sure :) )

shurcooL: I'll keep it as is, because `CheckResponse` is exported and can't easily be modified, and I don't think it's worth adding a second version just to optimize this really minor calculation (at the expense of adding code complexity). Rate limit errors should not occur in "hot loops" where each millisecond saved is beneficial.

github/github.go
shurcooL: Done in 76866e1ff66da4ce270bc18c95c53946109759bc. PTAL.

willnorris:
rebased onto current master and merged as c3f5683ad1dfe8c2c5aad4a7bbe536cbdc4a1009 and 39d75606a4cbec14d6f9f9d4e295dad4e3711682.  thanks!

PR #278, opened 2016-02-12
support 'since' pagination

flyinprogrammer:
Support parsing 'since' from Link header, which enables us to actually paginate through listing Users. This is especially helpful when working with Github Enterprise.

https://developer.github.com/v3/users/#get-all-users

Also, this is kind of a hack since we're overloading the idea of a 'Page' but it does work well with code like this:

```go

opt := &github.UserListOptions{
	Since: 0,
}

// get all pages of results
var allUsers []github.User
for {
	users, resp, err := client.Users.ListAll(opt)
	if err != nil {
		return err
	}
	allUsers = append(allUsers, users...)
	if resp.NextPage == 0 {
		break
	}
	opt.Since = resp.NextPage
}
```

Also I did not add tests - if we want tests, and we want this code, I can write tests.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -263,8 +263,13 @@ func (r *Response) populatePageValues() {
 				continue
 			}
 			page := url.Query().Get("page")
-			if page == "" {
+			since := url.Query().Get("since")
+			if page == "" && since == "" {
 				continue
+			}
+
+			if since != "" {
+				page = since
 			}
 
 			for _, segment := range segments[1:] {


flyinprogrammer:
I signed it!

willnorris:
The problem with parsing `since` as an int and overloading `page` in this way is that `since` is sometimes a timestamp.  For example, when [listing commits on a repo](https://developer.github.com/v3/repos/commits/#list-commits-on-a-repository), [listing notifications](https://developer.github.com/v3/activity/notifications/#list-your-notifications), and probably others.  I don't actually know off the top of my head if those use the rel=next links in the response, but if they did, then this code would have issues (looks like it would just fail silently).

The "right" approach (and I thought we had a long standing issue open for this, but I can't seem to find it), is that we shouldn't be trying to parse values out of the links in the response headers.  We should just be treating those URLs as opaque and fetching them directly to get the next page of results.  This gets a little messy with knowing exactly what to unmarshal the response as, but it's certainly doable.  Though I know this doesn't help you right now.

In the short term, maybe we should just document how to paginate the "list all users" responses?  Because the `since` parameter in that case is defined as the last user ID that you saw, you actually don't even need the rel=next link from the response.  Just take the ID of the last user in the previous page, and use it to populate the `since` value.

PR #281, opened 2016-02-13
issues: add Repository

pcarrier:


pcarrier:
I signed it!

willnorris:
thanks!

shurcooL:
When is this field set? Can you provide some examples or references?

I've chcked https://developer.github.com/v3/issues/ and I don't see `"repository"` anywhere.

I do see `"repository_url"`, but that's a string. Did you mean to add that instead?

willnorris:
That's actually a good question.  I had the same thought, but forgot to double check it before merging this in.  When is this actually set?  Maybe on a webhook event I suspect?

willnorris:
No, even on issue events, the repository is sent alongside the issue, not inside of it: https://developer.github.com/v3/activity/events/types/#issuesevent

pcarrier:
It is sent on `/issues` and `/orgs/$org/issues`.

willnorris:
ah yes, I see it there now (docs at https://developer.github.com/v3/issues/#list-issues don't show repository field, but I just tested and verified it). Thanks for confirming.

shurcooL:
Thanks for confirming.

PR #282, opened 2016-02-14
Do URL path escape composing the URL for GetContents

daviddengcn:
When any of owner/repo/path contains a "#" an escaping is needed.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -119,6 +119,11 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 	return nil, fmt.Errorf("No file named %s found in %s", filename, dir)
 }
 
+func escapeUrlPath(path string) string {
+	u := url.URL{Path: path}
+	return u.EscapedPath()
+}
+
 // GetContents can return either the metadata and content of a single file
 // (when path references a file) or the metadata of all the files and/or
 // subdirectories of a directory (when path references a directory). To make it
@@ -129,7 +134,7 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 // GitHub API docs: http://developer.github.com/v3/repos/contents/#get-contents
 func (s *RepositoriesService) GetContents(owner, repo, path string, opt *RepositoryContentGetOptions) (fileContent *RepositoryContent,
 	directoryContent []*RepositoryContent, resp *Response, err error) {
-	u := fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, path)
+	u := escapeUrlPath(fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, path))
 	u, err = addOptions(u, opt)
 	if err != nil {
 		return nil, nil, nil, err
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -119,7 +119,7 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 	return nil, fmt.Errorf("No file named %s found in %s", filename, dir)
 }
 
-func escapeUrlPath(path string) string {
+func escapeURLPath(path string) string {
 	u := url.URL{Path: path}
 	return u.EscapedPath()
 }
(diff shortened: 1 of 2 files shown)


willnorris:
GitHub won't let owner or repo have '#' in them.  That's certainly valid in filenames though, so I definitely see that it would need to be escaped there.  Are there other methods that take a path like this that would also need to be escaped.  Do you have an existing repository that has a file or directory with '#' in its name that I can test a few things on?

github/repos_contents.go
```diff
@@ -119,6 +119,11 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 	return nil, fmt.Errorf("No file named %s found in %s", filename, dir)
 }
 
+func escapeUrlPath(path string) string {
```
shurcooL: Minor, `golint` would say this needs to be named `escapeURLPath`, since Go style uses MixedCaps style.

daviddengcn:
Thanks for the review. I found this problem when my crawler was crawling the following package:

https://github.com/MattBlack85/Golearning/tree/master/golang_tour

github/repos_contents.go
```diff
@@ -119,6 +120,10 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 	return nil, fmt.Errorf("No file named %s found in %s", filename, dir)
 }
 
+func escapeURLPath(path string) string {
+	return template.URLQueryEscaper(path)
```
willnorris: so this is nice in that it works with go1.4.  Internally, `html/template.URLQueryEscaper` just calls `text/template.URLQueryEscaper` which in turn calls `net/url.QueryEscape`, so let's just use that directly here.

I'm also a little bit concerned about *over* escaping... url.QueryEscape escapes [all of these characters](https://github.com/golang/go/blob/aa22c42d704f9c6d9d8e65f972ac819678629ff2/src/net/url/url.go#L123) which is probably more than we need.  At least escaping "/" is not harmful (to wit: https://api.github.com/repos/MattBlack85/Golearning/contents/golang_tour%2Fexcercise.tour%2325.go).  How would you feel about trying to create a test file that uses all of the characters listed above (or at least as many as your OS will allow) to verify that GitHub unescapes them all properly (I feel fairly certain it does, but it'd be nice to confirm).

It's also interesting to note that the GitHub API itself does not escape the '#' in URLs it outputs (e.g: the response from https://api.github.com/repositories/12052373/contents/golang_tour). /cc @pengwynn @izuzak

daviddengcn: Yeah, it does escape more than necessary. Fortunately, those over escaping works well.

I've change the file name to "p#?%/中.go" which contains most common chars needing escapses.

Maybe, someone can file a bug to Github API.

willnorris: Updating the test is certainly good, though I was actually referring to testing how the GitHub API actually behaves.  I'll try creating a test repo to verify.

(the two folks I cc'ed above work on the API)

willnorris: okay, seems to work well on a live URLs: 

 - https://api.github.com/repos/google-test/wjn1/contents/p%23%3F%25%2F%E4%B8%AD.go
 - https://api.github.com/repos/google-test/wjn1/contents/%24%26%2B%2C%3A%3B%3D%3F%40.txt

Several of the escapes are unnecessary, but none seem to cause any problems.

willnorris:
So I ended up simplifying this a little... calling `url.QueryEscape` directly, and only calling it on the passed-in filepath, not the full constructed API path.  The response value in the test also wasn't really serving any purpose, so I removed that.  

Rebased, squashed and merged as d3fb171e8bc79b0c7222da1b7d84b45b9d300395.  Thanks for adding this!

github/repos_contents.go
izuzak: > It's also interesting to note that the GitHub API itself does not escape the '#' in URLs it outputs (e.g: the response from https://api.github.com/repositories/12052373/contents/golang_tour).

Thanks so much for the ping, @willnorris :bow: -- that _is_ interesting. I'll open an internal issue so that the team can take a closer look.

izuzak: > It's also interesting to note that the GitHub API itself does not escape the '#' in URLs it outputs (e.g: the response from https://api.github.com/repositories/12052373/contents/golang_tour). 

@willnorris This should now be fixed, but let us know if you're still noticing this anywhere. Thanks again for the report! 💙

PR #283, opened 2016-02-16
go-github: move event types into event_types.go

gmlewis:
Change-Id: I9df89a4fc8d2213e5caf17599f752167386b98f3


github/event_types.go
```diff
@@ -0,0 +1,106 @@
...
+	Email *string `json:"email,omitempty"`
+}
+
+//PullRequestEvent represents the payload delivered by PullRequestEvent webhook
+type PullRequestEvent struct {
+	Action      *string      `json:"action,omitempty"`
+	Number      *int         `json:"number,omitempty"`
+	PullRequest *PullRequest `json:"pull_request,omitempty"`
+
+	// The following fields are shared by Webhook events and the Events API.
```
willnorris: actually, it's the opposite of this, right?  The following fields are *only* populated in Webhook events.

github/event_types.go
```diff
@@ -0,0 +1,106 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
```
willnorris: maybe add a file-level comment that these event types are shared between the Events API and used as Webhook payloads.

github/event_types.go
gmlewis: Yes indeed.  Sorry about that.
Done.

github/event_types.go
gmlewis: Done.

PR #287, opened 2016-02-17
go-github: add CreateEvent

gmlewis:
Change-Id: I013dcbfb58ec10756c7b4e4bae6fe79b5e7c0062

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -11,12 +11,25 @@ package github
 //
 // GitHub API docs: http://developer.github.com/v3/activity/events/types/#pushevent
 type PushEvent struct {
-	PushID  *int                 `json:"push_id,omitempty"`
-	Head    *string              `json:"head,omitempty"`
-	Ref     *string              `json:"ref,omitempty"`
-	Size    *int                 `json:"size,omitempty"`
-	Commits []PushEventCommit    `json:"commits,omitempty"`
-	Repo    *PushEventRepository `json:"repository,omitempty"`
+	PushID       *int                 `json:"push_id,omitempty"`
+	Head         *string              `json:"head,omitempty"`
+	Ref          *string              `json:"ref,omitempty"`
+	Size         *int                 `json:"size,omitempty"`
+	Commits      []PushEventCommit    `json:"commits,omitempty"`
+	Repo         *PushEventRepository `json:"repository,omitempty"`
+	Before       *string              `json:"before,omitempty"`
+	DistinctSize *int                 `json:"distinct_size,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	After      *string          `json:"after,omitempty"`
+	Created    *bool            `json:"created,omitempty"`
+	Deleted    *bool            `json:"deleted,omitempty"`
+	Forced     *bool            `json:"forced,omitempty"`
+	BaseRef    *string          `json:"base_ref,omitempty"`
+	Compare    *string          `json:"compare,omitempty"`
+	HeadCommit *PushEventCommit `json:"head_commit,omitempty"`
+	Pusher     *User            `json:"pusher,omitempty"`
+	Sender     *User            `json:"sender,omitempty"`
 }
 
 func (p PushEvent) String() string {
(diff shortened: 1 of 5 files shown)


willnorris:
no need to break all of these into individual commits... just add them all in one

github/event_types.go
```diff
@@ -7,16 +7,35 @@
 
 package github
 
+// Pusher represents the name and email address of a GitHub pusher.
+type Pusher struct {
```
willnorris: rather than create a new type, let's simply use a `User` instead.  Even though the GitHub API docs only document including the name and email fields, I have a feeling that this really is supposed to identify a GitHub user.

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+}
+
+// CreateEvent represents the payload delivered by CreateEvent webhook.
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#createevent
+type CreateEvent struct {
+	Ref          *string `json:"ref,omitempty"`
+	RefType      *string `json:"ref_type,omitempty"`
```
willnorris: document possible values.  Documentation style we've used for enums like this is:

```go
// Possible values are: foo, bar, baz.
```

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+}
+
+// IssuesEvent represents the payload delivered by IssuesEvent webhook.
+//
+// This webhook gets fired when an issue is assigned, unassigned, labeled,
+// unlabeled, opened, closed, or reopened.
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#issuesevent
+type IssuesEvent struct {
+	Action   *string `json:"action,omitempty"`
```
willnorris: document possible values

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+	// The following fields are only populated by Webhook events.
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+}
+
+// MemberEvent represents the payload delivered by MemberEvent webhook.
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#memberevent
+type MemberEvent struct {
+	Action *string `json:"action,omitempty"`
```
willnorris: document possible values (even though there's only one, "added")

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+	// The following fields are only populated by Webhook events.
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+}
+
+// MembershipEvent represents the payload delivered by MembershipEvent webhook.
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#membershipevent
+type MembershipEvent struct {
+	Action *string `json:"action,omitempty"`
```
willnorris: Possible values are: "added", "removed".

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+	Scope  *string `json:"scope,omitempty"`
```
willnorris: Possible values are: "team".

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+	// The following fields are only populated by Webhook events.
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+}
+
+// PullRequestReviewCommentEvent represents the payload delivered by PullRequestReviewCommentEvent webhook.
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#pullrequestreviewcommentevent
+type PullRequestReviewCommentEvent struct {
+	Action      *string             `json:"action,omitempty"`
```
willnorris: grab docs from GitHub developer site.  They include a lot more than just possible values for this one which we'll definitely want to capture.

github/event_types.go
```diff
@@ -106,3 +125,243 @@ type IssueCommentEvent struct {
...
+
+// PublicEvent represents the payload delivered by PublicEvent webhook.
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#publicevent
+type PublicEvent struct {
+	// The following fields are only populated by Webhook events.
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+}
+
```
willnorris: this isn't actually here, but could you go ahead and add a doc for the PullRequestEvent.action field?  The GitHub API docs include a lot more than just possible values for this one which we should definitely capture.

gmlewis: OK, I believe I have filled in the interesting parts of the GitHub documentation.
PTAL when you get a chance, and I will start working on running 'fields.go' and the integration tests.
Let me know if you would like me to move the three pre-existing *Event structs so that they are all in sorted order.  I left them at the top of the file to reduce churn, but am happy to organize them alphabetically if you think that is better.

willnorris:
looks great, thanks!  I just went ahead and sorted the types.

PR #294, opened 2016-02-23
go-github: add utilities for processing payload messages

gmlewis:
Change-Id: I1c80b8eb7816efc9b83e6b8be956430c68809d5e

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -1,0 +1,119 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This file provides functions for validating payloads from GitHub Webhooks.
+// GitHub docs: https://developer.github.com/webhooks/securing/#validating-payloads-from-github
+
+package github
+
+import (
+	"crypto/hmac"
+	"crypto/sha1"
+	"crypto/sha256"
+	"crypto/sha512"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"hash"
+	"io/ioutil"
+	"net/http"
+	"strings"
+)
+
+const (
+	// sha1Prefix is the prefix used by GitHub before the HMAC hexdigest.
+	sha1Prefix = "sha1"
+	// sha256Prefix and sha512Prefix are provided for future compatibility.
+	sha256Prefix = "sha256"
+	sha512Prefix = "sha512"
+	// signatureHeader is the GitHub header key used to pass the HMAC hexdigest.
+	signatureHeader = "X-Hub-Signature"
+)
+
+// genMAC generates the HMAC signature for a message provided the secret key
+// and hashFunc.
+func genMAC(message, key []byte, hashFunc func() hash.Hash) []byte {
+	mac := hmac.New(hashFunc, key)
+	mac.Write(message)
+	return mac.Sum(nil)
+}
+
+// checkMAC reports whether messageMAC is a valid HMAC tag for message.
+func checkMAC(message, messageMAC, key []byte, hashFunc func() hash.Hash) bool {
+	expectedMAC := genMAC(message, key, hashFunc)
+	return hmac.Equal(messageMAC, expectedMAC)
+}
+
+// messageMAC returns the hex-decoded HMAC tag from the signature and its
+// corresponding hash function.
+func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
+	if signature == "" {
+		return nil, nil, errors.New("missing signature")
+	}
+	sigParts := strings.SplitN(signature, "=", 2)
+	if len(sigParts) != 2 {
+		return nil, nil, fmt.Errorf("error parsing signature %q", signature)
+	}
+
+	var hashFunc func() hash.Hash
+	switch sigParts[0] {
+	case sha1Prefix:
+		hashFunc = sha1.New
+	case sha256Prefix:
+		hashFunc = sha256.New
+	case sha512Prefix:
+		hashFunc = sha512.New
+	default:
+		return nil, nil, fmt.Errorf("unknown hash type prefix: %q", sigParts[0])
+	}
+
+	buf, err := hex.DecodeString(sigParts[1])
+	if err != nil {
+		return nil, nil, fmt.Errorf("error decoding signature %q: %v", signature, err)
+	}
+	return buf, hashFunc, nil
+}
+
+// ValidatePayload validates an incoming GitHub Webhook event request
+// and returns the (JSON) payload.
+// secretKey is the GitHub Webhook secret message.
+//
+// Example usage:
+//
+//     func (s *GitHubEventMonitor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+//       payload, err := github.ValidatePayload(r, s.webhookSecretKey)
+//       if err != nil { ... }
+//       // Process payload...
+//     }
+//
+func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err error) {
+	payload, err = ioutil.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	sig := r.Header.Get(signatureHeader)
+	if err := validateSignature(sig, payload, secretKey); err != nil {
+		return nil, err
+	}
+	return payload, nil
+}
+
+// validateSignature validates the signature for the given payload.
+// signature is the GitHub hash signature delivered in the X-Hub-Signature header.
+// payload is the JSON payload sent by GitHub Webhooks.
+// secretKey is the GitHub Webhook secret message.
+//
+// GitHub docs: https://developer.github.com/webhooks/securing/#validating-payloads-from-github
+func validateSignature(signature string, payload, secretKey []byte) error {
+	messageMAC, hashFunc, err := messageMAC(signature)
+	if err != nil {
+		return err
+	}
+	if !checkMAC(payload, messageMAC, secretKey, hashFunc) {
+		return errors.New("payload signature check failed")
+	}
+	return nil
+}
--- a/github/messages_test.go
+++ b/github/messages_test.go
@@ -1,0 +1,81 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"bytes"
+	"net/http"
+	"testing"
+)
+
+func TestValidatePayload(t *testing.T) {
+	const defaultBody = `{"yo":true}` // All tests below use the default request body and signature.
+	const defaultSignature = "sha1=126f2c800419c60137ce748d7672e77b65cf16d6"
+	secretKey := []byte("0123456789abcdef")
+	tests := []struct {
+		signature   string
+		eventID     string
+		event       string
+		wantEventID string
+		wantEvent   string
+		wantPayload string
+	}{
+		// The following tests generate expected errors:
+		{},                         // Missing signature
+		{signature: "yo"},          // Missing signature prefix
+		{signature: "sha1=yo"},     // Signature not hex string
+		{signature: "sha1=012345"}, // Invalid signature
+		// The following tests expect err=nil:
+		{
+			signature:   defaultSignature,
+			eventID:     "dead-beef",
+			event:       "ping",
+			wantEventID: "dead-beef",
+			wantEvent:   "ping",
+			wantPayload: defaultBody,
+		},
+		{
+			signature:   defaultSignature,
+			event:       "ping",
+			wantEvent:   "ping",
+			wantPayload: defaultBody,
+		},
+		{
+			signature:   "sha256=b1f8020f5b4cd42042f807dd939015c4a418bc1ff7f604dd55b0a19b5d953d9b",
+			event:       "ping",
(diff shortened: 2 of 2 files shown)


messages/messages.go
```diff
@@ -0,0 +1,126 @@
...
+	// signatureHeader is the Github header key used to pass the HMAC hexdigest.
+	signatureHeader = "X-Hub-Signature"
+	// deliveryHeader is the Gihub header key used to pass the globally-unique Github Event ID.
+	deliveryHeader = "X-GitHub-Delivery"
+)
+
+// GenSignature generates a signature for the given payload.
+// hashType possible values are "sha1", "sha256", "sha512".
+// If hashType is empty or invalid, "sha1" is used.
+func GenSignature(payload, secretKey []byte, hashType string) string {
```
willnorris: Do we really need this function?  I can't think of a case where a user of the go-github library should ever need to generate a signature.  I know that you're using it in tests below, but I actually think we should use hardcoded, known signature values for tests.

messages/messages.go
```diff
@@ -0,0 +1,126 @@
...
+// If hashFunc is nil, sha1.New is used.
+func checkMAC(message, messageMAC, key []byte, hashFunc func() hash.Hash) bool {
+	expectedMAC := genMAC(message, key, hashFunc)
+	return hmac.Equal(messageMAC, expectedMAC)
+}
+
+// EventPayload validates an incoming Github Webhook event request
+// and returns the event ID, event type, and the (JSON) payload.
+// secretKey is the GitHub Webhook secret message.
+func EventPayload(r *http.Request, secretKey []byte) (eventID, event string, payload []byte, err error) {
```
willnorris: I probably wouldn't return eventID here.  It's not relevant for validating the message signature, and can very easily be parsed off the request by the caller if they want it.

messages/messages.go
```diff
@@ -0,0 +1,126 @@
...
+// ValidateSignature validates the signature for the given payload.
+// sig is the GitHub hash signature delivered in the X-Hub-Signature header.
+// payload is the JSON payload sent by GitHub Webhooks.
+// secretKey is the GitHub Webhook secret message.
+// GitHub docs: https://developer.github.com/webhooks/securing/#validating-payloads-from-github
+func ValidateSignature(sig string, payload, secretKey []byte) error {
+	if sig == "" {
+		return errors.New("missing signature")
+	}
+	var sigLen int
```
willnorris: sigLen is a slightly confusing name, since this isn't actually the length of the signature, it's the length of the prefix.  An alternate option (which might be easier to read, I'm not sure?) would be to:

```go
sigParts := strings.SplitN(sig, "=", 2)
...
```

messages/messages.go
```diff
@@ -0,0 +1,126 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package messages provides functions for validating payloads from GitHub Webhooks.
+// GitHub docs: https://developer.github.com/webhooks/securing/#validating-payloads-from-github
+package messages
```
willnorris: depending on how much we can simplify the number of exported methods, we might be able to just move this into the main github package.

gmlewis: SGTM.
I'll move them to `github/messages*.go`.

messages/messages.go
gmlewis: SGTM.  I'll remove it.

messages/messages.go
gmlewis: SGTM.

messages/messages.go
gmlewis: Much nicer.  Thanks.

gmlewis:
Files updated. PTAL.

parkr:
:shipit: would _love_ to use this.

parkr:
@willnorris Any update here? 🙏

github/messages.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+const (
+	// sha1Prefix is the prefix used by Github before the HMAC hexdigest.
+	sha1Prefix = "sha1"
+	// sha256Prefix and sha512Prefix are provided for future compatibility.
+	sha256Prefix = "sha256"
+	sha512Prefix = "sha512"
+	// eventHeader is the Github header key used to pass the event type.
+	eventHeader = "X-Github-Event"
```
parkr: I believe this should be `X-GitHub-Event`, with the capital `H` in `GitHub`. 😄

shurcooL: The same change should probably be applied to the comments too; the canonical brand name spelling is "GitHub", not "Github". But if that's a wider problem than just in this PR, it might be acceptable to address it in a different PR.

gmlewis: @parkr - It turns out that canonical headers don't work that way. You can verify this yourself using `curl` or `bat`.
Here is an example response I just got from the GitHub API:

```
X-Github-Request-Id : 68840050:7F91:291081F8:5755B423
X-Github-Media-Type : github.v3
```

I tried searching for the `RFC` that defines this but came up empty.  I'm sure it is out there somewhere.

As for the comments correcting the brand name spelling... yes, that could all be cleaned up by someone.

gmlewis: It turns out that with all the code I removed from the PR, that those header constants were no longer needed anyway, so I removed them.

gmlewis:
@shurcooL  - when you get a chance, could you PTAL?

github/messages.go
```diff
@@ -0,0 +1,113 @@
...
+	buf, err := hex.DecodeString(sigParts[1])
+	if err != nil {
+		return nil, nil, fmt.Errorf("error decoding signature %q: %v", signature, err)
+	}
+	return buf, hashFunc, nil
+}
+
+// ValidatedPayload validates an incoming GitHub Webhook event request
+// and returns the (JSON) payload.
+// secretKey is the GitHub Webhook secret message.
```
shurcooL: I don't have a lot of context on how `ValidatedPayload` is meant to be used (perhaps it's a good idea to provide an example). So I have the following question.

> The client must close the response body when finished with it

Looking at the implementation, I see that ValidatedPayload reads the request body. It does not close the request body.

Is the user expected to close the request body after calling `ValidatedPayload`?

Also, `ValidatedPayload` is not documented in a way that makes it clear it will read the request body. Perhaps it can be inferred from "validates the request, returns JSON payload". But maybe it should be explicitly stated that it reads the request body and doesn't close it, if it's indeed the user's responsibility to close it?

Or should it close the request body itself?

github/messages_test.go
```diff
@@ -0,0 +1,81 @@
...
+			req.Header.Set(signatureHeader, test.signature)
+		}
+
+		got, err := ValidatedPayload(req, secretKey)
+		if err != nil {
+			if test.wantPayload != "" {
+				t.Errorf("ValidatedPayload(%#v): err = %v, want nil", test, err)
+			}
+			continue
+		}
```
shurcooL: Related to my comment above, this test code is not valid because it results in the request body not being closed after being read.

shurcooL:
I've spotted one issue with the request body and/or documentation. Aside from that, I'm not seeing any other issues. I think if we resolve that, this will have my LGTM.

github/messages.go
```diff
@@ -0,0 +1,113 @@
...
+func ValidateSignature(signature string, payload, secretKey []byte) error {
+	messageMAC, hashFunc, err := messageMAC(signature)
+	if err != nil {
+		return err
+	}
+	if !checkMAC(payload, messageMAC, secretKey, hashFunc) {
+		return errors.New("payload signature check failed")
+	}
+	return nil
+}
```
shurcooL: Also, what's the difference between `ValidatedPayload` and `ValidateSignature`? Is it intentional that one uses adjective "validated" and the other uses verb "validate"? It seems unintentional, since they are both phrased as "X validates ...":

```
// ValidatedPayload validates ...
// ValidateSignature validates ...
```

The more exported symbols there are, the harder the API is to learn and to use (users will need to learn both, and figure out which one is more appropriate for them to use). Is it beneficial to have both of them exported, or can we make one of them unexported?

`ValidatedPayload` looks to be a simple wrapper on top of `ValidateSignature`.

github/messages.go
```diff
@@ -0,0 +1,113 @@
...
+	// signatureHeader is the GitHub header key used to pass the HMAC hexdigest.
+	signatureHeader = "X-Hub-Signature"
+)
+
+// genMAC generates the HMAC signature for a message provided the secret key and hashFunc.
+// If hashFunc is nil, sha1.New is used.
+func genMAC(message, key []byte, hashFunc func() hash.Hash) []byte {
+	if hashFunc == nil {
+		hashFunc = sha1.New
+	}
```
shurcooL: Style, since `checkMAC` is the only caller of `genMAC`, I think `genMAC` can be simplified by factoring the "If hashFunc is nil, sha1.New is used." behavior out to `checkMAC`. I think this is simpler, isn't it?

```Go
// genMAC generates the HMAC signature for a message provided the secret key and hashFunc.
func genMAC(message, key []byte, hashFunc func() hash.Hash) []byte {
	mac := hmac.New(hashFunc, key)
	mac.Write(message)
	return mac.Sum(nil)
}

// checkMAC reports whether messageMAC is a valid HMAC tag for message.
// If hashFunc is nil, sha1.New is used.
func checkMAC(message, messageMAC, key []byte, hashFunc func() hash.Hash) bool {
	if hashFunc == nil {
		hashFunc = sha1.New
	}
	expectedMAC := genMAC(message, key, hashFunc)
	return hmac.Equal(messageMAC, expectedMAC)
}
```

shurcooL: Actually, even `checkMAC` is only called in one place, so you can factor out the default hash func logic even higher, into `ValidateSignature` itself.

That way, neither `genMAC` nor `checkMAC` need to worry about handling nil hash func.

gmlewis: OK, SGTM.  I'll move the nil has func up to `ValidateSignature`.

github/messages.go
gmlewis: Sorry, obviously some more comments are needed.  `ValidatePayload` will be used by a server and GitHub will send events to it. A server does not close the response body... only clients do that, and in this case, GitHub is the client.

How about if I add the comment:
```go
// ValidatePayload is called by the server and therefore does not need to close the request body.
```

github/messages.go
gmlewis: SGTM. I will stop exporting `validateSignature` and rename the other to `ValidatePayload`.

github/messages_test.go
gmlewis: I believe this test is OK because we are not going through the `http` package or server testing framework, but just calling the function directly and `bytes.Buffer` has no `Close` function.

github/messages.go
willnorris: but here we're talking about a request body, not a response body.  Isn't it the server's job to close that when finished reading?  I'm not sure how strict that needs to be with requests.

github/messages.go
gmlewis: It turns out that `messageMAC` always returns a valid `hashFunc` (if there are no errors) so the `nil` check is not needed at all.  :smile:

gmlewis:
@shurcooL - I believe I addressed your comments. PTAL.

github/messages.go
gmlewis: In https://godoc.corp.google.com/pkg/net/http it says:

> // For server requests the Request Body is always non-nil
> // but will return EOF immediately when no body is present.
> // The Server will close the request body. The ServeHTTP
> // Handler does not need to.

So I don't think it needs closing.  Do you agree?

willnorris: yep, I agree. That sounds pretty clear.

github/messages_test.go
shurcooL: Ok, I agree. What I said was wrong, I misunderstood the context in which this is being done. I thought it was an outbound client request being simulated, not an inbound server request.

github/messages.go
shurcooL: Ah, I completely missed that this is being done on inbound server requests. Right, this is for webhooks, so GitHub is making the request to user code, and this func is used to validate GitHub's incoming request.

> How about if I add the comment:
>
> ```Go
> // ValidatePayload is called by the server and therefore does not need to close the request body.
> ```

Do you think this comment is helpful and clear? I'm not sure. It almost raises more questions than answers, like "by what server"? It could be the HTTP server that the user is running, or GitHub's server.

If you think it's clear, then feel free to leave it as is.

Otherwise, I think it's completely fine to just remove it. The `net/http` package covers this pretty well, so I don't think it's needed to repeat it. As long as people realize that `ValidatePayload` is used to validate requests coming from GitHub. To that end, if issue #361 is going to be resolved by adding a short example of `ValidatePayload` being used somewhere, then that's even more reason to leave this out. I think seeing `ValidatePayload` used in context of `http.HandleFunc` makes it very clear:

```Go
http.HandleFunc("/hook", func(w http.ResponseWriter, r *http.Request) {
    if !github.VerifyPayload(r, os.Getenv("GH_WEBHOOK_SIGNATURE")) {
        http.Error(w, "invalid signature", http.StatusForbidden)
        return
    }
    // Proces payload ...
})
```

This is really nitpicky so feel free to ignore.

gmlewis: SGTM. I've removed the unclear text and replaced with an example to address #361.

gmlewis:
Cool, thanks for your reviews, @shurcooL - I appreciate it!
Submitting.

PR #295, opened 2016-02-24
add additional fields to Milestone struct

jonasi:
Additional fields (html_url, labels_url, id, closed_at) pulled from
https://developer.github.com/v3/issues/milestones/


willnorris:
cool, thanks!

PR #298, opened 2016-03-03
go-github: add support for commit reference SHA-1 method

### LINKED ISSUE #297: Add support for commit reference SHA-1 method
Announcement: https://developer.github.com/changes/2016-02-24-commit-reference-sha-api/

Docs: https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference

gmlewis:
Fixes #297.

Change-Id: Ib17c43b368a0d0ee1a33690c17ad7c0f4dab55e5

### DIFF
--- a/github/repos_commits.go
+++ b/github/repos_commits.go
@@ -6,6 +6,7 @@
 package github
 
 import (
+	"bytes"
 	"fmt"
 	"time"
 )
@@ -144,6 +145,33 @@ func (s *RepositoriesService) GetCommit(owner, repo, sha string) (*RepositoryCom
 	}
 
 	return commit, resp, err
+}
+
+// GetCommitSHA1 gets the SHA-1 of a commit reference.  If a last-known SHA1 is
+// supplied and no new commits have occurred, a 304 Unmodified response is returned.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference
+func (s *RepositoriesService) GetCommitSHA1(owner, repo, ref, lastSHA string) (string, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/commits/%v", owner, repo, ref)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return "", nil, err
+	}
+	if lastSHA != "" {
+		req.Header.Set("If-None-Match", `"`+lastSHA+`"`)
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeCommitReferenceSHAPreview)
+
+	var buf bytes.Buffer
+	resp, err := s.client.Do(req, &buf)
+	if err != nil {
+		return "", resp, err
+	}
+
+	return buf.String(), resp, err
 }
 
 // CompareCommits compares a range of commits with each other.
--- a/github/repos_commits_test.go
+++ b/github/repos_commits_test.go
@@ -119,6 +119,47 @@ func TestRepositoriesService_GetCommit(t *testing.T) {
 	}
 }
 
+func TestRepositoriesService_GetCommitSHA1(t *testing.T) {
+	setup()
+	defer teardown()
+	const sha1 = "01234abcde"
+
+	mux.HandleFunc("/repos/o/r/commits/master", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeCommitReferenceSHAPreview)
+
+		fmt.Fprintf(w, sha1)
+	})
+
+	got, _, err := client.Repositories.GetCommitSHA1("o", "r", "master", "")
+	if err != nil {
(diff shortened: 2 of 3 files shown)


gmlewis:
Note that while testing this function live against GitHub, I can't seem to get the `If-None-Match` feature to work, although it appears to work with `curl`, so I'm obviously doing something wrong.

Here is my test code:

```go
func getSHA(client *github.Client) {
	owner := "google-test"
	repo := "my-test-repo"
	ref := "master"

	sha1, _, err := client.Repositories.GetCommitSHA1(owner, repo, ref, "")
	if err != nil {
		log.Printf("error getting commit SHA1 %q: %v", repo, err)
	}
	fmt.Printf("sha1 = %v\n", sha1)

	v, _, err := client.Repositories.GetCommitSHA1(owner, repo, ref, sha1)
	if err == nil {
		log.Printf("error getting commit SHA1 %v/%v err = nil, wanted 304 Unmodified", repo, ref)
	}
	fmt.Printf("v = %v, err = %v\n", v, err)
}
```

and here is the output:
```
sha1 = bae5635861a8f568801549cfe5441fe4239c2c04
2016/03/03 18:22:28 error getting commit SHA1 my-test-repo/master err = nil, wanted 304 Unmodified
v = bae5635861a8f568801549cfe5441fe4239c2c04, err = <nil>
```

shurcooL:
I might be missing something, but why do you expect an error to be non-nil?

```Go
if err == nil {
    log.Printf("error getting commit SHA1 %v/%v err = nil, wanted 304 Unmodified", repo, ref)
}
```

I would expect that a nil error is okay. What I'd want to check is that `resp.Status` == http.StatusNotModified`.

In other words, I'd expect something like the following:

```Go
v, resp, err := client.Repositories.GetCommitSHA1(owner, repo, ref, sha1)
if err != nil {
    log.Printf("error getting commit SHA1: %v\n", err)
}
if resp.Status != http.StatusNotModified {
    log.Printf("error getting commit SHA1 %v/%v resp.Status = %v, wanted 304 Unmodified", repo, ref, resp.StatusCode)
}
fmt.Printf("v = %v, err = %v\n", v, err)
```

@gmlewis, in case you're not familiar with how `net/http` client treats status codes, you can see the docs for https://godoc.org/net/http#Get:

> An error is returned if there were too many redirects or if there was an HTTP protocol error. A non-2xx response doesn't cause an error.

Please ignore if I'm off base.

gmlewis:
Ah, thanks, @shurcooL !  I'll try that out.

gmlewis:
That was it, @shurcooL ... thanks again!
It turns out that I was also missing the double quotes around the If-None-Match string (which are critical).
The live test now works.
PTAL.

github/repos_commits_test.go
```diff
@@ -119,6 +119,47 @@ func TestRepositoriesService_GetCommit(t *testing.T) {
...
+	if got != want {
+		t.Errorf("Repositories.GetCommitSHA1 = %v, want %v", got, want)
+	}
+
+	mux.HandleFunc("/repos/o/r/commits/tag", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeCommitReferenceSHAPreview)
+		testHeader(t, r, "If-None-Match", `"`+sha1+`"`)
+
+		http.Error(w, "Unmodified", http.StatusNotModified)
```
willnorris: this isn't an error, so just use `w.WriteHeader(http.StatusNotModified)`

github/repos_commits.go
```diff
@@ -146,6 +147,35 @@ func (s *RepositoriesService) GetCommit(owner, repo, sha string) (*RepositoryCom
 	return commit, resp, err
 }
 
+// GetCommitSHA1 gets the SHA-1 of a commit reference.
+// If a last-known SHA1 is supplied and no new commits have occurred, a
+// 304 Unmodified response is returned.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference
+// See also: https://developer.github.com/changes/2016-02-24-commit-reference-sha-api/
```
willnorris: no need for this link since the GitHub docs already link to the announcement blog post.

github/repos_commits.go
```diff
@@ -146,6 +147,35 @@ func (s *RepositoriesService) GetCommit(owner, repo, sha string) (*RepositoryCom
 	return commit, resp, err
 }
 
+// GetCommitSHA1 gets the SHA-1 of a commit reference.
+// If a last-known SHA1 is supplied and no new commits have occurred, a
```
willnorris: If the intent is for this to be a separate paragraph, I think it would need a blank line before it.  If not, why have it separated from the line above?

github/repos_commits_test.go
gmlewis: Done.

github/repos_commits.go
gmlewis: Done.

github/repos_commits.go
gmlewis: Done.

Issue #291, opened 2016-02-19
Add support for Source Import API


willnorris:
Announcement: https://developer.github.com/changes/2016-02-19-source-import-preview-api/

Docs: https://developer.github.com/v3/migration/source_imports/

I'll take this myself, as we've been using it for quite a while internally at Google to power https://code.google.com/export-to-github/ and have a working implementation.

/cc @imjasonh in case you care to take it

ImJasonH:
All yours. :)

PR #307, opened 2016-03-18
go-github: add support for Migrations API

### LINKED ISSUE #292: Add support for Migrations API
As far as I know, this actually hasn't been announced, I just happened to stumble on the docs based on the new Source Import API.

Docs: https://developer.github.com/v3/migration/migrations/

Some more info can be found at https://help.github.com/enterprise/2.4/admin/guides/migrations/exporting-the-github-com-organization-s-repositories/.  I'm still not 100% sure what this does.  Anyone want to experiment and find out? :)

gmlewis:
The Migrations API is the GitHub equivalent of Google TakeOut.
It allows you to create an archive of one or more repositories
that you can download and later use for import purposes.

See: https://help.github.com/enterprise/2.4/admin/guides/migrations/exporting-the-github-com-organization-s-repositories/
for more information.

Fixes #292.

Change-Id: Id20624acb4f16a2aade9143e723c8eb85ec75402

### DIFF
--- a/github/migrations.go
+++ b/github/migrations.go
@@ -1,0 +1,225 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+// MigrationService provides access to the migration related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/migration/
+type MigrationService struct {
+	client *Client
+}
+
+// Migration represents a GitHub migration (archival).
+type Migration struct {
+	ID   *int    `json:"id,omitempty"`
+	GUID *string `json:"guid,omitempty"`
+	// State is the current state of a migration.
+	// Possible values are:
+	//     "pending" which means the migration hasn't started yet,
+	//     "exporting" which means the migration is in progress,
+	//     "exported" which means the migration finished successfully, or
+	//     "failed" which means the migration failed.
+	State *string `json:"state,omitempty"`
+	// LockRepositories indicates whether repositories are locked (to prevent
+	// manipulation) while migrating data.
+	LockRepositories *bool `json:"lock_repositories,omitempty"`
+	// ExcludeAttachments indicates whether attachments should be excluded from
+	// the migration (to reduce migration archive file size).
+	ExcludeAttachments *bool         `json:"exclude_attachments,omitempty"`
+	URL                *string       `json:"url,omitempty"`
+	CreatedAt          *string       `json:"created_at,omitempty"`
+	UpdatedAt          *string       `json:"updated_at,omitempty"`
+	Repositories       []*Repository `json:"repositories,omitempty"`
+}
+
+func (m Migration) String() string {
+	return Stringify(m)
+}
+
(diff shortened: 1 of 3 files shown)


github/orgs_migrations.go
```diff
@@ -0,0 +1,247 @@
...
+	// the migration (to reduce migration archive file size).
+	ExcludeAttachments *bool `json:"exclude_attachments,omitempty"`
+}
+
+// StartMigration starts the generation of a migration archive.
+// repos is a slice of repository names to migrate.
+//
+// GitHub API docs: https://developer.github.com/v3/migration/migrations/#start-a-migration
+func (s *OrganizationsService) StartMigration(org string, repos []string, opt *MigrationOptions) (*Migration, *Response, error) {
+	if org == "" {
```
willnorris: We don't do these kinds of empty string checks anywhere else in the library.  I agree that they're nice, since it allows returning an error before the API is ever called, but I just worry a little about inconsistency with the rest of the library.

github/orgs_migrations.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMigrationsPreview)
+
+	// Disable the redirect mechanism because AWS fails if the GitHub auth token is provided.
+	var buf bytes.Buffer
+	var copyErr error
+	saveRedirect := s.client.client.CheckRedirect
+	s.client.client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
```
willnorris: This makes the client no longer thread-safe.  Isn't there somewhere else that we deal with AWS resources like this?  Maybe as part of the releases API?  How do we handle it there?

github/orgs_migrations.go
```diff
@@ -0,0 +1,247 @@
...
+	// ExcludeAttachments indicates whether attachments should be excluded from
+	// the migration (to reduce migration archive file size).
+	ExcludeAttachments *bool `json:"exclude_attachments,omitempty"`
+}
+
+// StartMigration starts the generation of a migration archive.
+// repos is a slice of repository names to migrate.
+//
+// GitHub API docs: https://developer.github.com/v3/migration/migrations/#start-a-migration
+func (s *OrganizationsService) StartMigration(org string, repos []string, opt *MigrationOptions) (*Migration, *Response, error) {
```
willnorris: Maybe move all these methods into a new `MigrationService`?  That way, we could add the source import methods as well (#291).

github/orgs_migrations.go
gmlewis: OK, removed checks.

github/orgs_migrations.go
gmlewis: Done. I also changed the filenames.

github/orgs_migrations.go
gmlewis: I looked around but couldn't find any other similar situations.
I could protect this call with a mutex to make it thread-safe.
What do you think?

gmlewis:
Updated to return only a URL for the redirect.
PTAL.

github/migrations.go
```diff
@@ -0,0 +1,228 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+)
+
+const defaultTimeout = 30 * time.Second
```
willnorris: I don't think this is needed anymore

PR #309, opened 2016-03-21
Correct path escaping

### LINKED ISSUE #308: Getting contents of files from folders with spaces doesn't work.
If the directory name includes a space character it gets encoded into `+` and the request will fail with 404 Not found.
It looks like there is a problem that was introduced by https://github.com/google/go-github/commit/d3fb171e8bc79b0c7222da1b7d84b45b9d300395
Turns out escaping paths and escaping queries are note quite the same as illustrated by: http://play.golang.org/p/5dg6gaiLiv.
PR coming.

pzurek:
Use the correct method to escape the url path.
Fixes #308.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -130,8 +130,8 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 func (s *RepositoriesService) GetContents(owner, repo, path string, opt *RepositoryContentGetOptions) (fileContent *RepositoryContent, directoryContent []*RepositoryContent, resp *Response, err error) {
 	// escape characters not allowed in URL path.  This actually escapes a
 	// lot more, but seems to be harmless.
-	urlPath := url.URL{Path: path}
-	escapedPath := urlPath.EscapedPath()
+	tmpURL := url.URL{Path: path}
+	escapedPath := tmpURL.EscapedPath()
 	u := fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, escapedPath)
 	u, err = addOptions(u, opt)
 	if err != nil {
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -130,7 +130,8 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 func (s *RepositoriesService) GetContents(owner, repo, path string, opt *RepositoryContentGetOptions) (fileContent *RepositoryContent, directoryContent []*RepositoryContent, resp *Response, err error) {
 	// escape characters not allowed in URL path.  This actually escapes a
 	// lot more, but seems to be harmless.
-	escapedPath := url.QueryEscape(path)
+	urlPath := url.URL{Path: path}
+	escapedPath := urlPath.EscapedPath()
 	u := fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, escapedPath)
 	u, err = addOptions(u, opt)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pzurek:
Looks like this breaks 1.4 compatibility. ¯\_(ツ)_/¯

shurcooL:
Right, `EscapedPath` was introduced in 1.5.

If it were up to me, I'd drop 1.4 support since it's more than 1 major version behind current stable release of Go.

However, if 1.4 support is going to be kept, you could consider relying on this:

> The String and RequestURI methods use EscapedPath to construct their results. In general, code should call EscapedPath instead of reading u.RawPath directly.

_(Source: https://godoc.org/net/url#URL.EscapedPath)_

So, perhaps you can use either `RequestURI` or `String` methods. But I'm not sure that's great, it will compile in 1.4 through tip, but 1.4 behavior might be different since its `RequestURI` and `String` do not use `EscapedPath`.

Another option might be to vendor/copy the code of `EscapedPath` into a unexported helper in go-github here.

Or just not do anything at all. This PR doesn't seem to close any known/verified issue, does it?

github/repos_contents.go
```diff
@@ -130,7 +130,8 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 func (s *RepositoriesService) GetContents(owner, repo, path string, opt *RepositoryContentGetOptions) (fileContent *RepositoryContent, directoryContent []*RepositoryContent, resp *Response, err error) {
 	// escape characters not allowed in URL path.  This actually escapes a
 	// lot more, but seems to be harmless.
```
shurcooL: With the change below, is this comment still accurate/relevant or should it be updated?

shurcooL:
> Or just not do anything at all. This PR doesn't seem to close any known/verified issue, does it?

Oh, I just noticed this PR is made to fix #308. It'd be a good idea to mention that in the PR description and git commit message, otherwise it's not clear.

github/repos_contents.go
willnorris: yeah, this comment can be removed.

willnorris:
I'm fine dropping go1.4 support.  But please update `README.md` and `.travis.yml` to reflect that.

pzurek:
@shurcooL Good point! I'll test if using string allows us to keep 1.4 support. It makes sense if it's as simple as that.

willnorris:
whichever implementation we go with, please also add a unit test case to demonstrate this problem.

pzurek:
@willnorris Would something as above suffice?

willnorris:
oh, I thought the problem was with a literal `+`?  Maybe extend that test to include that as well?  So something like `"some directory+name/file.go"` or whatever?

willnorris:
the other thing I'd recommend is to try this with the live GitHub API (similar to what I did in https://github.com/google-test/wjn1).  At some point, we might need to add a specific integration test for these kinds of issues, since it relies on how the GitHub API does escaping, and the only way to adequately test that is to hit the live API.

github/repos_contents.go
```diff
@@ -128,9 +128,8 @@ func (s *RepositoriesService) DownloadContents(owner, repo, filepath string, opt
 //
 // GitHub API docs: http://developer.github.com/v3/repos/contents/#get-contents
 func (s *RepositoriesService) GetContents(owner, repo, path string, opt *RepositoryContentGetOptions) (fileContent *RepositoryContent, directoryContent []*RepositoryContent, resp *Response, err error) {
-	// escape characters not allowed in URL path.  This actually escapes a
-	// lot more, but seems to be harmless.
-	escapedPath := url.QueryEscape(path)
+	tmpURL := url.URL{Path: path}
+	escapedPath := tmpURL.String()
```
shurcooL: Style suggestion. No need for `tmpURL`, just do:

```Go
escapedPath := (&url.URL{Path: path}).String()
```

shurcooL:
> @shurcooL Good point! I'll test if using string allows us to keep 1.4 support. It makes sense if it's as simple as that.

Have you tried both `String()` and `RequestURI()`? I've checked, and `RequestURI()` is also available in Go 1.4. Do they both work, or only `String()`? If both, which do you think is better to use here and why?

pzurek:
My preference was `String()` for being less enigmatic so I started with this and it worked.
Actually my first pre-PR version of this fix worked with `String()` and I only changed to `EscapedPath()` later. Looks like now I made a full circle.

willnorris:
LGTM

gmlewis:
Squashed commits into single commit and merged.

PR #312, opened 2016-03-21
change DownloadReleaseAsset API to additionally return a URL

### LINKED ISSUE #246: DownloadReleaseAsset fails.
Hi, 

We are trying to a download a release jar with DownloadReleaseAsset. 
It fails on this: 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidArgument</Code><Message>Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified</Message><ArgumentName>Authorization</ArgumentName><ArgumentValue>Bearer{token}</ArgumentValue><RequestId> {req-id} </RequestId><HostId> {host} </HostId></Error>
```

Could you point us in the right direction? 
Thank you.

gmlewis:
If a redirect to AWS occurs, then the redirect will be canceled and
the redirect Location will be returned in the string and the
io.ReadCloser will be nil. At this point, the user would need to
perform the http.Get on the returned URL in a different http.Client
(without the GitHub authentication transport).

Fixes #246.

Change-Id: I4fe2dfc9cd0ce81934a07f11a6296e71c7dd51a0

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -10,8 +10,10 @@ import (
 	"fmt"
 	"io"
 	"mime"
+	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 // RepositoryRelease represents a GitHub release in a repository.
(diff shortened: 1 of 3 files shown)


gmlewis:
Testing now... please hold off on review... I apologize... I meant to request this PR after testing was completed.

gmlewis:
Successfully tested DownloadReleaseAsset with AWS redirect. String URL is now returned.

github/repos_releases.go
```diff
@@ -213,27 +215,43 @@ func (s *RepositoriesService) GetReleaseAsset(owner, repo string, id int) (*Rele
...
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
-		return nil, err
+		return nil, "", err
 	}
 	req.Header.Set("Accept", defaultMediaType)
 
+	s.client.clientMu.Lock()
+	defer s.client.clientMu.Unlock()
```
willnorris: I think we should move the clientMu.Unlock() into the deferred func below, so that there is no chance that the mutex has been unlocked before the client.CheckRedirect is reset:

```go
defer func() { 
    s.client.client.CheckRedirect = saveRedirect
    s.client.clientMu.Unlock()
}()
```

github/repos_releases.go
```diff
@@ -213,27 +215,43 @@ func (s *RepositoriesService) GetReleaseAsset(owner, repo string, id int) (*Rele
...
+	s.client.client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		loc = req.URL.String()
+		return errors.New("disable redirect")
+	}
+	defer func() { s.client.client.CheckRedirect = saveRedirect }()
+
 	resp, err := s.client.client.Do(req)
 	if err != nil {
-		return nil, err
+		if !strings.Contains(err.Error(), "disable redirect") {
```
willnorris: this seems like a fragile check (even though the string is setup right above).  Would it be too heavy weight to add a custom (non-exported) error type for this?  We already know that we're going to have to use this same technique for the Migration API, so maybe it's worth doing?

willnorris: okay maybe not "fragile", just inelegant :)

github/repos_releases.go
gmlewis: I believe that there is currently no chance that the mutex will be unlocked before client.CheckRedirect is reset due to the guarantees of the Go language spec: https://golang.org/ref/spec#Defer_statements

github/repos_releases.go
gmlewis: SGTM.  Will do.

github/repos_releases.go
willnorris: ah, fair point.

github/repos_releases.go
gmlewis: Due to the http package adding its own information to the error, I can't check for equivalence in an elegant way.

For example, if I change the above to look like this:

```go
	s.client.clientMu.Lock()
	defer s.client.clientMu.Unlock()

	killRedirErr := errors.New("disable redirect")

	var loc string
	saveRedirect := s.client.client.CheckRedirect
	s.client.client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
		loc = req.URL.String()
		return killRedirErr
	}
	defer func() { s.client.client.CheckRedirect = saveRedirect }()

	resp, err := s.client.client.Do(req)
	if err != nil {
		if err != killRedirErr {
			return nil, "", err
		}
		return nil, loc, nil
	}

	return resp.Body, "", nil
```

...then the unit test fails because the returned error is not `disable redirect`, but instead it is `Get /yo: disable redirect`.
So it appears that some amount of string manipulation is necessary (like checking the suffix of the returned error message against the text of `killRedirErr.Error()`) unless you know of a better way to do this.

shurcooL: > So it appears that some amount of string manipulation is necessary (like checking the suffix of the returned error message against the text of `killRedirErr.Error()`) unless you know of a better way to do this.

I agree with Will that doing string matching is inelegant, but unfortunately, I don't know of a better way either.

willnorris: oh well, it was worth a shot.

willnorris:
:shipit:

PR #313, opened 2016-03-22
Add support for 451 status code

### LINKED ISSUE #306: Add support for 451 status code
Announcement: https://developer.github.com/changes/2016-03-17-the-451-status-code-is-now-supported/

go-github passes along the http response object, which will already include the 451 status unchanged.  The only thing that's different is that the response now includes a new "block" field.  We could either expand the existing Error type with this field and document that it's only populated on certain types of errors, or created a new Error type entirely.  At the moment, I'm leaning toward the former, but could probably be convinced otherwise if someone has a strong opinion.

gmlewis:
Fixes #306.

Change-Id: I3b42c946f8255e165c3bbd83b1566675d21fb0e8

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -366,6 +366,13 @@ type ErrorResponse struct {
 	Response *http.Response // HTTP response that caused this error
 	Message  string         `json:"message"` // error message
 	Errors   []Error        `json:"errors"`  // more detail on individual errors
+	// Block is only populated on certain types of errors such as code 451.
+	// See https://developer.github.com/changes/2016-03-17-the-451-status-code-is-now-supported/
+	// for more information.
+	Block *struct {
+		Reason    string     `json:"reason,omitempty"`
+		CreatedAt *Timestamp `json:"created_at,omitempty"`
+	} `json:"block,omitempty"`
 }
 
 func (r *ErrorResponse) Error() string {
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -520,7 +520,8 @@ func TestCheckResponse(t *testing.T) {
 		Request:    &http.Request{},
 		StatusCode: http.StatusBadRequest,
 		Body: ioutil.NopCloser(strings.NewReader(`{"message":"m",
-			"errors": [{"resource": "r", "field": "f", "code": "c"}]}`)),
+			"errors": [{"resource": "r", "field": "f", "code": "c"}],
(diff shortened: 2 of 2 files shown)


github/github_test.go
```diff
@@ -538,6 +538,42 @@ func TestCheckResponse(t *testing.T) {
 	}
 }
 
+func TestCheckResponse_withBlock(t *testing.T) {
```
willnorris: maybe just add a "block" to the existing `TestCheckResponse`?  All it's really doing is verifying that the JSON is unmarshaled correctly, so splitting these into separate tests doesn't gain much.

github/github_test.go
```diff
@@ -538,6 +538,42 @@ func TestCheckResponse(t *testing.T) {
...
+			CreatedAt: &Timestamp{time.Date(2016, 3, 17, 15, 39, 46, 0, time.UTC)},
+		},
+	}
+	if err.Block == nil && err.Block.CreatedAt == nil {
+		t.Errorf("CreatedAt = nil, want %v", want.Block.CreatedAt)
+	} else if !want.Block.CreatedAt.Equal(*err.Block.CreatedAt) {
+		t.Errorf("CreatedAt = %v, want %v", err.Block.CreatedAt, want.Block.CreatedAt)
+	}
+	// Clear times for DeepEqual to succeed.
+	err.Block.CreatedAt = nil
```
willnorris: I'm surprised this is necessary.  DeepEqual fails even for time values that are identical?

gmlewis: Thanks, Will!  (For others listening in... Will realized that `+00:00` is not the same at `time.UTC` and it should be replaced by `Z` instead.)

github/github_test.go
gmlewis: Done.

github/github.go
```diff
@@ -357,6 +357,14 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	return response, err
 }
 
+// Block represents more detailed error information such as is returned
+// for a code 451. See https://developer.github.com/changes/2016-03-17-the-451-status-code-is-now-supported/
+// for more information.
+type Block struct {
```
willnorris: I worry about created a new top-level, exported type just for this very obscure kind of error.  How bout declaring this as an anonymous inner struct on the ErrorResponse.  Something like http://play.golang.org/p/tmW_cl2pGr

willnorris:
:shipit:

PR #315, opened 2016-03-23
Add support for Feeds API

### LINKED ISSUE #191: Add support for Feeds API

gmlewis:
Fixes #191.

Change-Id: I2fbcca9913cd20df75e9b5a5272a50879025d526

### DIFF
--- a/github/activity.go
+++ b/github/activity.go
@@ -12,3 +12,58 @@ package github
 type ActivityService struct {
 	client *Client
 }
+
+// FeedLink represents a link to a related resource.
+type FeedLink struct {
+	HRef *string `json:"href,omitempty"`
+	Type *string `json:"type,omitempty"`
+}
+
+// Feeds represents timeline resources in Atom format.
+type Feeds struct {
+	TimelineURL                 *string  `json:"timeline_url,omitempty"`
+	UserURL                     *string  `json:"user_url,omitempty"`
+	CurrentUserPublicURL        *string  `json:"current_user_public_url,omitempty"`
+	CurrentUserURL              *string  `json:"current_user_url,omitempty"`
+	CurrentUserActorURL         *string  `json:"current_user_actor_url,omitempty"`
+	CurrentUserOrganizationURL  *string  `json:"current_user_organization_url,omitempty"`
+	CurrentUserOrganizationURLs []string `json:"current_user_organization_urls,omitempty"`
+	Links                       *struct {
+		Timeline                 *FeedLink  `json:"timeline,omitempty"`
+		User                     *FeedLink  `json:"user,omitempty"`
+		CurrentUserPublic        *FeedLink  `json:"current_user_public,omitempty"`
+		CurrentUser              *FeedLink  `json:"current_user,omitempty"`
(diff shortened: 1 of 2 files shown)


github/feeds.go
```diff
@@ -0,0 +1,69 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+// FeedsService provides access to the feeds related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/activity/feeds/
+type FeedsService struct {
```
willnorris: This is part of the "Activity" portion of the GitHub docs, so this would go on the existing `ActivityService`, rather than creating a new feeds service.

github/feeds.go
```diff
@@ -0,0 +1,69 @@
...
+	client *Client
+}
+
+// AtomRef represents an Atom reference.
+type AtomRef struct {
```
willnorris: AtomRef isn't really the best name for this.  While yes, they are being used to a reference an atom feed in this context, what GitHub is *really* doing here is using [HAL](http://stateless.co/hal_specification.html) style links.  So really just calling this `Link` would be appropriate.  However, in the interest of not using such a generic term for such a specific use case, maybe `FeedLink` is better.

github/feeds.go
gmlewis: Done.  It's interesting that their docs place this in /activity/feeds/ but the actual RESTful API uses just /feeds.

github/feeds.go
gmlewis: Done.  PTAL.

github/activity.go
```diff
@@ -12,3 +12,59 @@ package github
 type ActivityService struct {
 	client *Client
 }
+
+// FeedLink represents a HAL-style link.
```
willnorris: sorry to nitpick here :grin:

This is really only a "HAL-style link" when it is serialized as part of a "_links" JSON object.  Which this is when it's part of a `Feeds.Links`.  But for the purposes of documenting what the type itself is, outside the context of how it may be used in some cases, it's more accurate to simply say that it's a "link to a related resource".

An interesting spec to read on this is [rfc5988](https://tools.ietf.org/html/rfc5988) which defines links outside of how they may be serialized in section 3.  It does then go own to define the Link HTTP header in section 5, but that's really because it existed in a previous version of HTTP and got dropped because there wasn't a solid use case for it at the time.

gmlewis: No worries!  I'm learning a bunch here... very interesting.
Should I leave the hal_specification.html link, move it, or just remove it entirely?

willnorris: just remove it entirely.  It's not really relevant for users of this library.

gmlewis: OK, done.

Issue #319, opened 2016-03-30
Repository stargazers timestamps


ArtIsResistance:
There is no support for ["List Stargazers" method](https://developer.github.com/v3/activity/starring/#list-stargazers) with alternative response (with timestamps). 

There was a similar issue [#188](https://github.com/google/go-github/issues/188) with pull-request [#199](https://github.com/google/go-github/pull/199) that implements ["List Repositories Being Starred"](https://developer.github.com/v3/activity/starring/#list-repositories-being-starred) only, so i am going to implement this feature in issued method.

PR #320, opened 2016-03-30
Add support for Activities.ListStargazers response with timestamp.

ArtIsResistance:
Add custom header to request to get an alternate response with star creation timestamp.

Solves #319


willnorris:
thanks!

PR #323, opened 2016-03-30
Documentation and Example of GetReadme 

### LINKED ISSUE #322: Improve godocs based on my experiences
I'd like to open a PR with some more clarification around RepositoryContent functions. I misunderstood when I needed to call Decode() and I still confused on when someone might call String(). If others agree, I'd like to capture a few more notes in the comments so godocs captures it. 

Thoughts?

Thanks again for the help on Slack @willnorris :+1:

mjbrender:
I had a tough time figuring out how to view a readme through go-github. Adding a couple more notes on documentation for godoc to pick up and I added an example of doing so.

Fixes #322


github/repos_contents.go
```diff
@@ -72,7 +73,8 @@ func (r *RepositoryContent) Decode() ([]byte, error) {
 	return o, nil
 }
 
-// GetReadme gets the Readme file for the repository.
+// GetReadme gets the Readme file for the repository, which is stored as base64.
```
willnorris: Rather than all the docs various, would it not be sufficient to just document the Content field of RepositoryContent to say that it may be encoded, and if so, then use `Decode` to decode it?

I don't really want to document that Readme is always stored as base64 because I don't actually know that that's true.  Otherwise, why would GitHub bother returning an "encoding" value?

examples/getreadme/main.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
willnorris: could you change this to be an inline example, similar to what I did in 32a84d13d2c44f6a093a89da754d8a57a399c90f?

github/repos_contents.go
mjbrender: Great point - I'll address this.

examples/getreadme/main.go
mjbrender: Glad to, though I don't know if I follow the definition of inline. Would it also go `github/repos_test.go`? Happy to have it land where you'd like it to live.

github/repos_contents.go
mjbrender: I'm thinking what could be said safely is that it's retrieving encoded content. For instance: `// GetReadme gets the encoded readme file for the repository.`

Or was part of the point is you don't want to assume that it will remain encoded?

mjbrender: As I type it up, I think it's great to have a note on RepositoryContent, though I want to give people a heads up as they look at each command. I could go either way, but I err on sharing wherever it hurt as a user.

examples/getreadme/main.go
willnorris: close. actually in `github/repos_contents_test.go`, right after the existing `TestRepositoriesService_GetReadme` test.  So we would just add a `ExampleRepositoriesService_GetReadme` there

github/repos_contents.go
willnorris: These methods clearly return a `RepositoryContent` struct, so wouldn't it be reasonable to expect a user to look at that type to then know what to do with it?

Nowhere else in go-github (that I can recall), does a method document how to use the returned type.  Instead, the type itself documents what its fields are, special cases about when they may or may not be present, etc.  I think the one exception may be where we mention that the caller is responsible for closing an `io.ReadCloser` similar to how the `net/http` documents that, but I may be misremembering that.

examples/getreadme/main.go
mjbrender: @willnorris working on this, but I'm not familiar with the convention of an example in test. From [reading up here](https://blog.golang.org/examples), it looks like I may need to change the package to `github_test` in order to import and run this line: `client := github.NewClient(nil)`. If I do that, calls to `setup`, `teardown` and `RepositoryContent` are undefined. 

What's the right move here? Outputs below: 

```
### After changing the package to github_test ###
# github.com/google/go-github/github_test
./repos_contents_test.go:14: undefined: setup
./repos_contents_test.go:15: undefined: teardown
./repos_contents_test.go:16: undefined: RepositoryContent
./repos_contents_test.go:28: undefined: setup
./repos_contents_test.go:29: undefined: teardown
./repos_contents_test.go:30: undefined: RepositoryContent
./repos_contents_test.go:38: undefined: setup
./repos_contents_test.go:39: undefined: teardown
./repos_contents_test.go:40: undefined: mux in mux.HandleFunc
./repos_contents_test.go:41: undefined: testMethod
./repos_contents_test.go:41: too many errors
FAIL	github.com/google/go-github/github [build failed]

#### After importing and naming the package back to github ####
import cycle not allowed in test
package github.com/google/go-github/github (test)
	imports github.com/google/go-github/github

FAIL	github.com/google/go-github/github [setup failed]
```

mjbrender: Got it! I use `setup` & `teardown` now that I figured them out :ok_hand:

mjbrender:
I think we're on track now @willnorris. The edits we landed on are mostly in https://github.com/google/go-github/pull/323/commits/02f6480341ae88c85a61f502a305e0dc702518ed. The only addition is a description for `String()` from 962a0f11ab8740fc3d1a9be31b268b8deff32768. You're welcome to cherrypick (based on what I read in the repo) or lmk if you'd like some rebasing done on my end. I didn't want to lose your comments on past commits until you looked it over. Cheers :+1:

willnorris:
With the new `GetContents` method in #331, this gets a lot simpler since callers should now always use that and not think at all about whether the content is encoded or not.  Given that, I moved the docs down to the Content field and update the Example to use GetContents and simplified it a bit.

PR #331, opened 2016-04-08
add RepoContent.GetContent to replace Decode

willnorris:
This came up when reviewing #323.  I think the best approach here is to have a single method that callers are instructed to use, regardless of the content encoding... they shouldn't have to think about that.

/cc @mjbrender

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -61,6 +61,8 @@ func (r RepositoryContent) String() string {
 }
 
 // Decode decodes the file content if it is base64 encoded.
+//
+// Deprecated: Use GetContent instead.
 func (r *RepositoryContent) Decode() ([]byte, error) {
 	if *r.Encoding != "base64" {
 		return nil, errors.New("cannot decode non-base64")
@@ -70,6 +72,27 @@ func (r *RepositoryContent) Decode() ([]byte, error) {
 		return nil, err
 	}
 	return o, nil
+}
+
+// GetContent returns the content of r, decoding it if necessary.
+func (r *RepositoryContent) GetContent() (string, error) {
+	var encoding string
+	if r.Encoding != nil {
+		encoding = *r.Encoding
+	}
+
+	switch encoding {
+	case "base64":
+		c, err := base64.StdEncoding.DecodeString(*r.Content)
+		return string(c), err
+	case "":
+		if r.Content == nil {
+			return "", nil
+		}
+		return *r.Content, nil
+	default:
+		return "", fmt.Errorf("unsupported content encoding: %v", encoding)
+	}
 }
 
 // GetReadme gets the Readme file for the repository.
--- a/github/repos_contents_test.go
+++ b/github/repos_contents_test.go
@@ -8,27 +8,91 @@ import (
 	"testing"
 )
 
-func TestDecode(t *testing.T) {
-	setup()
-	defer teardown()
-	r := RepositoryContent{Encoding: String("base64"), Content: String("aGVsbG8=")}
-	o, err := r.Decode()
-	if err != nil {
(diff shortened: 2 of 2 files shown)


github/repos_contents.go
```diff
@@ -72,6 +74,28 @@ func (r *RepositoryContent) Decode() ([]byte, error) {
...
+	switch encoding {
+	case "base64":
+		c, err := base64.StdEncoding.DecodeString(*r.Content)
+		return string(c), err
+	case "":
+		var s string
+		if r.Content != nil {
+			s = *r.Content
+		}
+		return s, nil
```
gmlewis: I'm concerned that this does an unnecessary string copy.
What about:

```go
if r.Content == nil {
    return "", nil
}
return *r.Content, nil
```

github/repos_contents.go
```diff
@@ -72,6 +74,28 @@ func (r *RepositoryContent) Decode() ([]byte, error) {
 	return o, nil
 }
 
+// GetContent returns the content of r, decoding it if necessary.
+func (r *RepositoryContent) GetContent() (string, error) {
```
gmlewis: What do you think about gradually switching over to Go's naming policies as new APIs are added?
So for example, this would be `Content` instead of `GetContent`?
https://golang.org/doc/effective_go.html#Getters
If you want to stay consistent with the rest of `go-github`, that is fine with me too.
"I saw an opportunity and I seized it." :smiley:

willnorris: I would have, except that there is an exported field named Content.

github/repos_contents.go
willnorris: ah, good call.  Done.

github/repos_contents_test.go
```diff
@@ -8,27 +8,91 @@ import (
...
+		r := RepositoryContent{Encoding: tt.encoding, Content: tt.content}
+		got, err := r.GetContent()
+		if err != nil && !tt.wantErr {
+			t.Errorf("RepositoryContent(%q, %q) returned unexpected error: %v", tt.encoding, tt.content, err)
+		}
+		if err == nil && tt.wantErr {
+			t.Errorf("RepositoryContent(%q, %q) did not return unexpected error", tt.encoding, tt.content)
+		}
+		if want := tt.want; got != want {
+			t.Errorf("RepositoryContent.Decode returned %+v, want %+v", got, want)
```
gmlewis: s/Decode/GetContent/

github/repos_contents_test.go
```diff
@@ -8,27 +8,91 @@ import (
...
 }
 
-func TestDecodeBadEncoding(t *testing.T) {
-	setup()
-	defer teardown()
-	r := RepositoryContent{Encoding: String("bad")}
-	_, err := r.Decode()
-	if err == nil {
-		t.Errorf("Should fail to decode non-base64")
+func TestRepositoryContent_GetContents(t *testing.T) {
```
gmlewis: s/GetContents/GetContent/

gmlewis:
LGTM

github/repos_contents_test.go
willnorris: done.

github/repos_contents_test.go
willnorris: done

shurcooL:
How come PRs are showing up with 0 commits lately?

willnorris:
uh, that's weird.  It had a commit earlier when Glenn reviewed it.  I did delete the branch on my fork, but that shouldn't have made a difference.  GitHub bug maybe?

/cc @fabianperez

mjbrender:
Thanks for these improvements and for keeping me in mind @willnorris. 👍

PR #332, opened 2016-04-09
Add RepositoriesService.GetByID method.

shurcooL:
Note: GetByID uses the undocumented GitHub API endpoint `/repositories/:id`.

Include license preview Accept header.

Helps #329.

This commit is based on https://github.com/sourcegraph/go-github/commit/da394ef1f53de34e9c3e5782ca27589df32b426c which has been tested in production. It has been amended by me to include [Licences API](https://developer.github.com/changes/2015-03-09-licenses-api/) header, improved documentation and commit message, with appropriate updates to the test.

I've tested it locally with the Licences API header to make sure it works, and I've confirmed that it did.

PR #333, opened 2016-04-09
Add UsersService.GetByID method.

### LINKED ISSUE #329: Proposal: Add GetByID methods to UsersService, RepositoriesService, with caveat.
I'd like to hear your thoughts on the following proposal.

I know this project is Go client library for accessing the GitHub API. It tries to stay very current with GitHub API changes. It implements support for new APIs early, even when they're still in preview period and may result in changes before preview period ends.

As a result, this Go package tries to maintain a stable API, but when breaking GitHub API changes happen that necessitate otherwise, it follows suit.

I propose considering adding two methods, but read below for details:

```Go
// GetByID fetches a user.
func (s *UsersService) GetByID(id int) (*User, *Response, error) { ... }

// GetByID fetches a repository.
func (s *RepositoriesService) GetByID(id int) (*Repository, *Response, error) { ... }
```

### Motivation

Right now, `UsersService` exposes two methods for getting users, one is by `login`, and the other gets the authenticated user.

It's possible to change usernames on github.com, so `login` is not necessary a very permanent way to look up the same user. As far as I know, IDs can not ever be changed.

### Caveat

Both of these methods are not definitely described as part of the GitHub API on https://developer.github.com/v3/ from what I can tell.

Because of that, simply providing these methods would be completely unacceptable, because there is a higher than usual chance they may change or go away in the future.

However, I consider them to be quite close in nature to the new API endpoints that are a part of a preview period (that require setting specific media type Accept header values to opt in). So, if these methods are provided with a comment saying:

```Go
// Note: GetByID uses the undocumented GitHub API endpoint /user/:id.
```

Then it might be acceptable, since users of the API can see if the benefits of using this method (for their needs) is worth accepting a risk of potentially needing to update code if there are changes in the future.

However, both API endpoints do work, and have worked for at least a year without any changes, from what I know.

E.g.:

https://api.github.com/users/octocat - get user with username "octocat", it says user ID is 583231
https://api.github.com/user/583231 - gets the same user by ID

Also note, it's possible to get a user by ID in a following way (this uses fully documented APIs only, however it returns a smaller set of user details, but a second API call to get user by login can be made if neccessary):

```Go
// gitHubUsersGetByID fetches a GitHub user based on their userID.
func gitHubUsersGetByID(gh *github.Client, userID int) (*github.User, *github.Response, error) {
	req, err := gh.NewRequest("GET", fmt.Sprintf("/users?since=%v&per_page=1", userID-1), nil)
	if err != nil {
		return nil, nil, err
	}
	var users []github.User
	resp, err := gh.Do(req, &users)
	if err != nil {
		return nil, resp, err
	}
	if len(users) != 1 {
		return nil, resp, fmt.Errorf("expected 1 user, got %v users", len(users))
	}
	if users[0].ID == nil {
		return nil, resp, fmt.Errorf("got user with nil user ID: %#v", users[0])
	}
	if *users[0].ID != userID {
		return nil, resp, fmt.Errorf("expected user ID %v, got user ID %v", userID, *users[0].ID)
	}
	return &users[0], resp, nil
}
```

### Implementation details

See the following commits for implementation details.

`UsersService.GetByID` - https://github.com/sourcegraph/go-github/commit/e4f201ff9b2966d65e34b342d19a386d5c35624f

`RepositoriesService.GetByID` - https://github.com/sourcegraph/go-github/commit/da394ef1f53de34e9c3e5782ca27589df32b426c

If this proposal is accepted, I can clean them up and make PRs.

What do you think?

shurcooL:
Note: GetByID uses the undocumented GitHub API endpoint `/user/:id`.

Resolves #329 (assuming #332 is also merged).

This commit is based on sourcegraph@e4f201ff9b2966d65e34b342d19a386d5c35624f which has been tested in production. It has been amended by me to add details to the commit message only.

@sqs, can you please confirm you're okay with this commit being contributed to Google?

### DIFF
--- a/github/users.go
+++ b/github/users.go
@@ -95,6 +95,25 @@ func (s *UsersService) Get(user string) (*User, *Response, error) {
 	return uResp, resp, err
 }
 
+// GetByID fetches a user.
+//
+// Note: GetByID uses the undocumented GitHub API endpoint /user/:id.
+func (s *UsersService) GetByID(id int) (*User, *Response, error) {
+	u := fmt.Sprintf("user/%d", id)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	user := new(User)
+	resp, err := s.client.Do(req, user)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return user, resp, err
+}
+
 // Edit the authenticated user.
 //
 // GitHub API docs: http://developer.github.com/v3/users/#update-the-authenticated-user
--- a/github/users_test.go
+++ b/github/users_test.go
@@ -100,6 +100,26 @@ func TestUsersService_Get_invalidUser(t *testing.T) {
 	testURLParseError(t, err)
 }
 
+func TestUsersService_GetByID(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/user/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, `{"id":1}`)
+	})
+
+	user, _, err := client.Users.GetByID(1)
+	if err != nil {
+		t.Errorf("Users.GetByID returned error: %v", err)
+	}
+
+	want := &User{ID: Int(1)}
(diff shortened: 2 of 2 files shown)


sqs:
I confirm.

PR #337, opened 2016-04-13
add support for git signing API methods

### LINKED ISSUE #334: Add support for git signing API methods
Announcement: https://developer.github.com/changes/2016-04-04-git-signing-api-preview/

gmlewis:
Successfully tested using new "gpg_key" scopes.

Fixes #334.

Change-Id: I9b10ae5e7679f5196778d9eb3ded6168d7facfc3

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -72,6 +72,9 @@ const (
 
 	// https://developer.github.com/changes/2016-04-01-squash-api-preview/
 	mediaTypeSquashPreview = "application/vnd.github.polaris-preview+json"
+
+	// https://developer.github.com/changes/2016-04-04-git-signing-api-preview/
+	mediaTypeGitSigningPreview = "application/vnd.github.cryptographer-preview+json"
 )
 
 // A Client manages communication with the GitHub API.
--- a/github/users_gpg_keys_test.go
+++ b/github/users_gpg_keys_test.go
@@ -1,0 +1,110 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+func TestUsersService_ListGPGKeys(t *testing.T) {
+	setup()
+	defer teardown()
+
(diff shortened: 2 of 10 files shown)


gmlewis:
@shurcooL - when you get a chance, can you please review this PR?

github/github.go
```diff
@@ -72,6 +72,9 @@ const (
 
 	// https://developer.github.com/changes/2016-04-01-squash-api-preview/
 	mediaTypeSquashPreview = "application/vnd.github.polaris-preview+json"
+
+	// https://developer.github.com/changes/2016-04-04-git-signing-api-preview/
+	mediaTypeGitSigningPreview = "application/vnd.github.cryptographer-preview"
```
shurcooL: Shouldn't you include the `+json` suffix here?

As far as I understand, we want to use the default mediaTypeV3 value of `application/vnd.github.v3+json`, except to replace the version `v3` with `cryptographer-preview`.

See https://developer.github.com/v3/media/ for explanation of +json part of media type.

github/users_gpg_keys_test.go
```diff
@@ -0,0 +1,110 @@
...
+mQINBFcEd9kBEACo54TDbGhKlXKWMvJgecEUKPPcv7XdnpKdGb3LRw5MvFwT0V0f
+...
+=tqfb
+-----END PGP PUBLIC KEY BLOCK-----`
+
+	mux.HandleFunc("/user/gpg_keys", func(w http.ResponseWriter, r *http.Request) {
+		gpgKey := &struct {
+			ArmoredPublicKey *string `json:"armored_public_key,omitempty"`
+		}{}
+		json.NewDecoder(r.Body).Decode(gpgKey)
```
shurcooL: Minor style nitpick. Since you're just creating a blank `gpgKey`, I would write the above as:

```Go
var gpgKey struct {
	ArmoredPublicKey *string `json:"armored_public_key,omitempty"`
}
json.NewDecoder(r.Body).Decode(&gpgKey)
```

I think that's a little easier to read and see that `gpgKey` is just a zero value. It's up to you if you want to take this suggestion.

shurcooL:
Left some minor comments above. Not seeing any other issues. LGTM.

github/github.go
gmlewis: Done.  Update will be pushed momentarily.

github/users_gpg_keys_test.go
gmlewis: Done.  Update will be pushed momentarily.

PR #342, opened 2016-04-21
add (AuthorizationsService) CreateImpersonation/DeleteImpersonation

vetinari:


vetinari:
short usage for those:

```go
func runAsUser(client *github.Client, login string, job func(*github.Client) error) error {
    auth, _, err := client.Authorizations.CreateImpersonation(login, &github.AuthorizationRequest{Scopes: []github.Scope{"write:public_key"}})
    if err != nil {
        return fmt.Errorf("failed to get impersonation token for %s: %s\n", login, err)
    }
    defer func() {
        if _, err := client.Authorizations.DeleteImpersonation(login); err != nil {
            fmt.Printf("ERROR: Failed to delete impersonation token for %s: %s\n", login, err)
        }
    }()
    ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: *(auth.Token)})
    uclient := github.NewClient(tc)
    uclient.BaseURL = GitHubBase
    return job(uclient)
}   
// fetch all users and for each user fetch all keys to "uk":
...
   switch len(uk) {
        case 0:
            fmt.Printf("INFO: User %s does not have any keys\n", login)
            continue
        case 1:
            if keys[login].Key != *(uk[0].Key) {
                err := runAsUser(client, login, func(cl *github.Client) error {
                    if _, err := cl.Users.DeleteKey(*(uk[0].ID)); err != nil {
                        fmt.Errorf("failed to delete key %d from %s: %s\n", *(uk[0].ID), login, err)
                    }
                    sshKey := keys[login].Key
                    mailAddr := keys[login].Mail
                    if _, _, err := ul.Users.CreateKey(&github.Key{Key: &sshKey, Title: &mailAddr}); err != nil {
                        return fmt.Errorf("failed to create key for %s: %s\n", login, err)
                    }
                    return nil
                })
                if err != nil {
                    log.Printf("ERROR: while updating key for %s: %s\n", login, err)
                }
            }
 ...
```

github/authorizations.go
```diff
@@ -318,3 +318,39 @@ func (s *AuthorizationsService) Revoke(clientID string, token string) (*Response
...
+	resp, err := s.client.Do(req, a)
+	if err != nil {
+		return nil, resp, err
+	}
+	return a, resp, err
+}
+
+// Delete an impersonation OAuth token.
+//
+// Revokes (all?) impersonation token(s).
```
shurcooL: I can see how the GitHub API docs are unclear, prompting the question mark.

> Delete an impersonation OAuth token

But is it possible to find out by testing whether it really deletes all, or a single one, and avoid the question mark in the docs here?

vetinari:
I'll give it a try tomorrow

github/authorizations.go
vetinari: I deletes just one, because there can be only one:
creating a second impersonation token will return 401s for attempts with the first one while the second will work

github/authorizations.go
```diff
@@ -318,3 +318,39 @@ func (s *AuthorizationsService) Revoke(clientID string, token string) (*Response
...
+	resp, err := s.client.Do(req, a)
+	if err != nil {
+		return nil, resp, err
+	}
+	return a, resp, err
+}
+
+// Delete an impersonation OAuth token.
+//
+// Revokes an impersonation token. NOTE: there can be only one.
```
shurcooL: > Delete an impersonation OAuth token.
>
> Revokes an impersonation token.

No need for the second sentence, it just repeats what the first one already said.

github/authorizations.go
```diff
@@ -318,3 +318,39 @@ func (s *AuthorizationsService) Revoke(clientID string, token string) (*Response
 
 	return s.client.Do(req, nil)
 }
+
+// Create an impersonation OAuth token.
+//
+// This required admin permissions. With the returned Authorization.Token you can e.g. create or
```
shurcooL: `required` -> `requires`.

github/authorizations.go
```diff
@@ -318,3 +318,39 @@ func (s *AuthorizationsService) Revoke(clientID string, token string) (*Response
...
+// delete a users ssh public key. NOTE: creating a new token automatically revokes an existing one.
```
shurcooL: `users` -> `user's`?

`ssh public key` should be `public SSH key` according to https://github.com/google/go-github/blob/0e3941cd1f14d0322165504974eeefab7e1ddc6d/github/users_keys.go#L10.

gmlewis:
@vetinari - can you please add unit tests for these two new functions, following the style of the other unit tests in this repo?

gmlewis:
@shurcooL - I'm thinking of leaving this open for one more week to give @vetinari a chance to write the unit tests... at that point if he has not had time, and you have given your LGTM to this PR, I will check this in and write up some unit tests for the functions myself in a separate PR.  Does that SGTY?

gmlewis:
Incorporated in https://github.com/google/go-github/commit/07a1059646109986eb08a3f57126f59e1d19f049

PR #344, opened 2016-04-21
Correctly handle API errors when downloading Release Assets

### LINKED ISSUE #343: DownloadReleaseAsset swallows API errors & returns io.ReadCloser of error payload
If the API request to download a release asset fails, no `error` is returned but the `io.ReadCloser` of the error JSON payload is:

```go
package main

import (
	"fmt"
	"io"
	"os"

)

func main() {

	rc, url, err := client.Repositories.DownloadReleaseAsset("not-a", "real-repo", 0)
	if err != nil {
		fmt.Println("error:", err)
	}

	if rc != nil {
		fmt.Print("rc: ")
		io.Copy(os.Stdout, rc)
		fmt.Print("\n")
		rc.Close()
	}

	if url != "" {
		fmt.Println("url:", url)
	}
}
```

Would expect:

```
→ go run main.go
error: Not Found
```

But what is returned:

```
→ go run main.go
rc: {"message":"Not Found","documentation_url":"https://developer.github.com/v3"}
```

rodaine:
If `RepositoriesService.DownloadReleaseAsset` encounters an error from the GitHub API (e.g., 404 Not Found), The error would not be returned by the method and instead the error payload would be delivered as the `io.ReadCloser`. This patch resolves this with `CheckResponse`.

Fixes #343

/cc @willnorris (thanks for the help!)

### DIFF
--- a/github/repos_releases_test.go
+++ b/github/repos_releases_test.go
@@ -253,6 +253,32 @@ func TestRepositoriesService_DownloadReleaseAsset_Redirect(t *testing.T) {
 	}
 }
 
+func TestRepositoriesService_DownloadReleaseAsset_APIError(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/releases/assets/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", defaultMediaType)
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprint(w, `{"message":"Not Found","documentation_url":"https://developer.github.com/v3"}`)
+	})
+
+	resp, loc, err := client.Repositories.DownloadReleaseAsset("o", "r", 1)
+	if err == nil {
+		t.Error("Repositories.DownloadReleaseAsset did not return an error")
+	}
+
+	if resp != nil {
+		resp.Close()
+		t.Error("Repositories.DownloadReleaseAsset returned stream, want nil")
+	}
+
+	if loc != "" {
+		t.Error(`Repositories.DownloadReleaseAsset returned "%s", want empty ""`, loc)
+	}
+}
+
 func TestRepositoriesService_EditReleaseAsset(t *testing.T) {
 	setup()
 	defer teardown()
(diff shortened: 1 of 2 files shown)


gmlewis:
LGTM. @rodaine - thank you for fixing this... it looks like I missed this case.
@googlebot - did you do your thing?

rodaine:
No sweat! 😅  If you're referring to the CLA, I think I got the ✅

willnorris:
👍

parkr:
@rodaine Did you check the actual return value of the example error or is that out of scope?

rodaine:
@parkr in the test? Since I'm handing it off to CheckResponse, that's somewhat beyond the scope of the test. Being consistent with the rest of the methods that use [`CheckResponse`](https://github.com/google/go-github/blob/master/github/github.go#L460), I figured bubbling that up to the consumer and allowing them to take action on it is probably best. All 3XX responses _should_ be captured by the [`CheckRedirect`](https://github.com/google/go-github/blob/master/github/repos_releases.go#L240) callback on the `http.Client`, so what remains are actual errors.

github/repos_releases_test.go
```diff
@@ -253,6 +253,32 @@ func TestRepositoriesService_DownloadReleaseAsset_Redirect(t *testing.T) {
...
+		t.Error("Repositories.DownloadReleaseAsset did not return an error")
+	}
+
+	if resp != nil {
+		resp.Close()
+		t.Error("Repositories.DownloadReleaseAsset returned stream, want nil")
+	}
+
+	if loc != "" {
+		t.Error(`Repositories.DownloadReleaseAsset returned "%s", want empty ""`, loc)
```
shurcooL: This should be `t.Errorf`, not `t.Error`.

@willnorris, will you take an improved `.travis.yml` that would run vet as part of CI? It would've caught this issue.

willnorris: Certainly

rodaine: My bad! Good catch.

shurcooL: I'll fix both with 1 PR (it's a good way to verify that the CI changes detect the issue; first commit will fail CI, the second will succeed). /cc @rodaine

PR #345, opened 2016-04-22
Improve test error message.

shurcooL:
Print the actual value that doesn't match.

Minor followup fix for c3f5683ad1dfe8c2c5aad4a7bbe536cbdc4a1009 (#277).


willnorris:
looks good, thanks!

PR #346, opened 2016-04-22
Travis: Run gofmt -s, go vet, go test -race, add Go 1.6.

shurcooL:
Also make tip a fast-finish allowed failure. That way, if CI fails on tip due to a temporary issue with tip, it will not break build status. However, it's still possible to see tip build status by looking at CI details page.

Expect CI to fail on this PR because there are known go vet issues, those will be fixed in next commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,22 @@
+sudo: false
 language: go
 go:
- - 1.4
- - 1.5
- - tip
-install: go get -v ./github
-script: go test -v ./github
+  - 1.5.4
+  - 1.6.2
+  - tip
+matrix:
+  include:
+    - go: 1.4.3
+      script:
+        - go get -t -v ./...
+        - go test -v -race ./...
+  allow_failures:
+    - go: tip
+  fast_finish: true
+install:
+  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
+script:
+  - go get -t -v ./...
+  - diff -u <(echo -n) <(gofmt -d -s .)
+  - go tool vet .
+  - go test -v -race ./...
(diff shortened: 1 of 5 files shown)


shurcooL:
As expected, CI is now failing because of the go vet issue.

https://travis-ci.org/google/go-github/builds/124940342

```bash
$ go tool vet .
github/repos_releases_test.go:278: possible formatting directive in Error call

The command "go tool vet ." exited with 1.
```

However, I'm gonna have to do something about Go 1.4. It's failing because of:

```bash
$ go tool vet .
go tool: no such tool "vet"; to install:
	go get golang.org/x/tools/cmd/vet
```

shurcooL:
Hmm, vet has been removed altogether from `golang.org/x/tools/cmd/vet` because Go only supports 2 point versions back (current and previous).

I'll just have CI skip running vet for 1.4, that seems reasonable.

shurcooL:
All done, this is ready from my side for review and merging. /cc @willnorris

gmlewis:
LGTM. Awaiting approval from @willnorris

.travis.yml
```diff
@@ -1,7 +1,22 @@
...
+  allow_failures:
+    - go: tip
+  fast_finish: true
+install:
+  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
+script:
+  - go get -t -v ./...
+  - diff -u <(echo -n) <(gofmt -d -s .)
+  - go tool vet .
+  - go test -v -race ./...
```
willnorris: it looks like two files in tests/integration are missing the "integration" build constraint, so expanding this call to "./..." picks those up.  Could you add that build constraint to `tests/integration/authorizations_test.go` and `tests/integration/github_test.go` as part of this?

PR #347, opened 2016-04-22
Predict *RateLimitError, return immediately without network call (9700x speedup when rate limit exceeded).

shurcooL:
This (minor) optimization is possible because the client keeps track of the rate limits from previous API calls, and GitHub's rate limiting model is known in advance, and is predictable.

In situations where it's possible to predict that the network API call will result in a `*RateLimitError`, simply return it immediately instead of making the network API call to GitHub's servers. This is both faster for the client (by around 9700 times in my benchmarks) because it avoids the network round-trip latency, and better for GitHub's servers that won't need to needlessly reject network requests.

Note that this optimization only kicks in when the client is already exceeding the rate limit and it hasn't been reset yet. It has no effect during most normal usage when rate limit is not exceeded.

### Benchmarks

I wrote a small benchmark (it assumes you've already exhausted the GitHub API rate limit before you begin):

```Go
func BenchmarkClientDo_rateLimitError(b *testing.B) {
	gh := github.NewClient(nil)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_, _, err := gh.Repositories.Get("shurcooL", "vfsgen")
		if _, ok := err.(*github.RateLimitError); !ok {
			log.Println("no *github.RateLimitError error, but expected it")
		}
	}
}
```

Results before this PR (with a latency to api.github.com of around 85 ms):

```
$ go test -bench=. -benchmem
testing: warning: no tests to run
PASS
BenchmarkClientDo_rateLimitError-8	      20	  97057828 ns/op	   55922 B/op	     126 allocs/op
ok  		2.609s
```

Results after this PR:

```
go test -bench=. -benchmem
testing: warning: no tests to run
PASS
BenchmarkClientDo_rateLimitError-8	  200000	     10005 ns/op	    2763 B/op	      34 allocs/op
ok  		3.576s
```

In this limited scenario, it's a speedup by 9700 times.

Helps #152 and #153.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -333,9 +333,15 @@ func (c *Client) Rate() Rate {
 // JSON decoded and stored in the value pointed to by v, or returned as an
 // error if an API error has occurred.  If v implements the io.Writer
 // interface, the raw response body will be written to v, without attempting to
-// first decode it.
+// first decode it.  If rate limit is exceeded and reset time is in the future,
+// Do returns *RateLimitError immediately without making a network API call.
 func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	rateLimitCategory := category(req.URL.Path)
+
+	// If we've hit rate limit, don't make further requests before Reset time.
+	if err := c.checkRateLimitBeforeDo(req, rateLimitCategory); err != nil {
+		return nil, err
+	}
 
 	resp, err := c.client.Do(req)
 	if err != nil {
@@ -374,6 +380,33 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	}
 
 	return response, err
+}
+
+// checkRateLimitBeforeDo does not make any network calls, but uses existing knowledge from
+// current client state in order to quickly check if *RateLimitError can be immediately returned
+// from Client.Do, and if so, returns it so that Client.Do can skip making a network API call unneccessarily.
+// Otherwise it returns nil, and Client.Do should proceed normally.
+func (c *Client) checkRateLimitBeforeDo(req *http.Request, rateLimitCategory rateLimitCategory) error {
+	c.rateMu.Lock()
+	rate := c.rateLimits[rateLimitCategory]
+	c.rateMu.Unlock()
+	if !rate.Reset.Time.IsZero() && rate.Remaining == 0 && time.Now().Before(rate.Reset.Time) {
+		// Create a fake response.
+		resp := &http.Response{
+			Status:     http.StatusText(http.StatusForbidden),
+			StatusCode: http.StatusForbidden,
+			Request:    req,
+			Header:     make(http.Header),
+			Body:       ioutil.NopCloser(strings.NewReader("")),
+		}
+		return &RateLimitError{
+			Rate:     rate,
+			Response: resp,
+			Message:  fmt.Sprintf("API rate limit of %v still exceeded until %v, not making remote request.", rate.Limit, rate.Reset.Time),
+		}
+	}
+
+	return nil
 }
 
 /*
--- a/github/github.go
+++ b/github/github.go
@@ -86,8 +86,9 @@ type Client struct {
 	// User agent used when communicating with the GitHub API.
 	UserAgent string
 
-	rateMu sync.Mutex
-	rate   Rate // Rate limit for the client as determined by the most recent API call.
+	rateMu     sync.Mutex
+	rateLimits [categories]Rate // Rate limits for the client as determined by the most recent API calls.
+	mostRecent rateLimitCategory
 
 	// Services used for talking to different parts of the GitHub API.
 	Activity       *ActivityService
@@ -323,7 +324,7 @@ func parseRate(r *http.Response) Rate {
 // current rate.
 func (c *Client) Rate() Rate {
 	c.rateMu.Lock()
-	rate := c.rate
+	rate := c.rateLimits[c.mostRecent]
 	c.rateMu.Unlock()
 	return rate
 }
@@ -334,6 +335,8 @@ func (c *Client) Rate() Rate {
 // interface, the raw response body will be written to v, without attempting to
 // first decode it.
 func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
+	rateLimitCategory := category(req.URL.Path)
+
 	resp, err := c.client.Do(req)
 	if err != nil {
 		return nil, err
@@ -348,7 +351,8 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	response := newResponse(resp)
 
 	c.rateMu.Lock()
-	c.rate = response.Rate
+	c.rateLimits[rateLimitCategory] = response.Rate
+	c.mostRecent = rateLimitCategory
 	c.rateMu.Unlock()
 
 	err = CheckResponse(resp)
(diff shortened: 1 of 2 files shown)


github/github.go
```diff
@@ -542,6 +569,25 @@ func (r RateLimits) String() string {
...
+	coreCategory rateLimitCategory = iota
+	searchCategory
+
+	categories // An array of this length will be able to contain all rate limit categories.
+)
+
+// category returns the rate limit category of the endpoint, determined by Request.URL.Path.
+func category(path string) rateLimitCategory {
+	switch {
+	default:
```
gmlewis: Curious... why do you list `default:` first? I don't remember seeing it that way before.

github/github.go
```diff
@@ -332,8 +333,31 @@ func (c *Client) Rate() Rate {
 // JSON decoded and stored in the value pointed to by v, or returned as an
 // error if an API error has occurred.  If v implements the io.Writer
 // interface, the raw response body will be written to v, without attempting to
-// first decode it.
+// first decode it.  If rate limit is exceeded and reset time is in the future,
+// Do returns *RateLimitError immediately without making a network API call.
 func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
+	rateLimitCategory := category(req.URL.Path)
```
gmlewis: What do you think about pulling this entire section (~22 lines) into its own separate function call?

As it stands, I think it draws attention away from someone looking at `Do` to find out what it does... but if it were in a separate function like `checkRateLimit` (or whatever you think is better) then I think this section would be much easier to read.

github/github.go
```diff
@@ -567,6 +613,17 @@ func (c *Client) RateLimits() (*RateLimits, *Response, error) {
 		return nil, nil, err
 	}
 
+	if response.Resources != nil {
+		c.rateMu.Lock()
+		if response.Resources.Core != nil {
```
gmlewis: Just curious... did you add any unit test for this section?  I was interested in seeing how these `.Core` and `.Search` responses were returned.

github/github.go
shurcooL: It's a style I've iterated towards after using `switch` statements more for multiplexing across multiple branching decisions. I prioritize the order of cases, returning `coreCategory` and `searchCategory`, to be more consistent with how they're defined, over the conditions that lead to those return values.

I read it as:

> For all paths, return `coreCategory`, except when it has `/search/` prefix, then return `searchCategory`.

I personally like it because I think the cases themselves are more important than the conditions to trigger them, but I know this is less common. If people feel this is worse, I can revert to a more traditional style of sorting by conditions (putting `default` last).

github/github.go
shurcooL: That sounds like a good idea, I'll do it. It'll look something like this:

```Go
if err := c.checkRateLimit(req); err != nil {
    return nil, err
}
```

github/github.go
shurcooL: I haven't yet. Should I?

> I was interested in seeing how these `.Core` and `.Search` responses were returned.

What do you mean by "how they were returned"?

gmlewis: Yeah, that sounds weird... not sure what I meant by that... I guess it would just be nice to see a test case for one or both of these.  Your call.

github/github.go
gmlewis: For other people reading, I want to note that the order of case expressions in Go is important... they are evaluated left-to-right and top-to-bottom and the first one that matches wins (except for `default` which wins when no others match).
https://golang.org/ref/spec#Switch_statements

Keeping it this way is fine with me.

github/github.go
shurcooL: Done in faca548.

github/github.go
shurcooL: Done in faca548.

shurcooL:
@gmlewis, I've addressed your current feedback, PTAL.

gmlewis:
Thanks, @shurcooL! LGTM 👍 
Awaiting review by @willnorris

shurcooL:
@willnorris Should I take the longer time to review to mean you're not as confident about this change and not sure how to respond to it, or have you simply not gotten around to it yet?

This change, at a high level, is very much in line with what you discussed/proposed in these two comments:

https://github.com/google/go-github/issues/152#issuecomment-70887913
https://github.com/google/go-github/pull/153#issuecomment-70884748

I think the high level design is solid.

About the implementation, I actually took a lot of time to come up with how to achieve the desired behavior, because I knew there were more than 1 sets of rate limits, depending on the route. But I think the final solution I came up with is pretty decent and works. But of course I welcome any potential suggestions for how to make the implementation even cleaner.

Let me know if I can do anything to help here.

willnorris:
just a case of me being out of town when the PR came in, and then busy with other things.  Looking at it now.

github/github.go
```diff
@@ -323,7 +324,7 @@ func parseRate(r *http.Response) Rate {
 // current rate.
 func (c *Client) Rate() Rate {
 	c.rateMu.Lock()
-	rate := c.rate
+	rate := c.rateLimits[c.mostRecent]
```
willnorris: this seems potentially problematic, since if a single client is being used for both search and non-search calls, the "most recent category" could get switched out from under you by another go routine before you have a chance to call Rate.  What if Rate always returns the core rate (like it did before), and we provide a different mechanism to access to other rates? Or alternatively, we just instruct callers to pull that rate off the response? checkRateLimitBeforeDo would still handle both categories as it does currently.

willnorris: actually yeah, we don't really need to provide another mechanism... we already instruct users to call client.RateLimits for the most up to date values.  That should be sufficient, no?

shurcooL: > if a single client is being used for both search and non-search calls, the "most recent category" could get switched out from under you by another go routine before you have a chance to call Rate.

If I'm not mistaken, that cannot happen. Both `rateLimits` and `mostRecent` are in a critical section protected by `rateMu` mutex. That mutex is very important to prevent this problem. Is that not the case?

shurcooL: Oh, I think I see what you meant. You weren't referring to a race condition in the code, but a higher-level "logical" race condition. E.g., a client makes a core category API call and then calls `Rate` expecting to see Rate limit for that category, but may instead get rate limit for the search category.

I agree that's potentially bad and should be considered. I'll think about this and get back to you later.

One observation is that each API call already returns a `response` which contains the rate limit from that very API call. So there's really no need for client to call `Rate`.

shurcooL: > What if Rate always returns the core rate (like it did before)

Now that I think about it more, are you sure that's what `Rate` currently does? Doesn't it already always return the latest rate limit from either of two categories, whichever was called last? In other words, this PR might be currently preserving previous behavior.

If you're sure that the current `Rate` only contains core category rate limits, can you show me where that happens and what prevents a call to search from updating rate?

It looks to me that search API calls would update it also. See here:

- https://github.com/google/go-github/blob/81ea1e5cb324404cd87c66d2a1050c7c31a2121b/github/github.go#L350-L352
- https://github.com/google/go-github/blob/81ea1e5cb324404cd87c66d2a1050c7c31a2121b/github/search.go#L157

willnorris: yep, looks like you're right... search is already updating the client rate, which leads to this same problem.  Maybe we should just deprecate this method entirely since it's likely to cause problems, and just instruct people to use the rate limit on the response, or to call client.RateLimits?  What do you think?

shurcooL: > Maybe we should just deprecate this method entirely since it's likely to cause problems, and just instruct people to use the rate limit on the response, or to call client.RateLimits? What do you think?

I am absolutely in favor of doing that, given those 2 better alternatives exist, and the current behavior is pretty unreliable (it can give you core or search rate limit, effectively randomly).

willnorris: Sounds good.  Could you go ahead and update the doc for this method then as part of this PR?

shurcooL: Done in 13c2701463e4b4a6b09952cdd50cd0d82b1e5351.

willnorris:
looks great, thanks!

shurcooL:
Awesome, thanks! It's not every day one gets a chance to submit a PR with a 9700x perf improvement. :P

PR #349, opened 2016-04-25
add more payloads to event.Payload

gmlewis:
Change-Id: I733568c73e1bc2cd748c5c5cd3d16654c7dd7828

### DIFF
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -27,12 +27,54 @@ func (e Event) String() string {
 	return Stringify(e)
 }
 
-// Payload returns the parsed event payload. For recognized event types
-// (PushEvent), a value of the corresponding struct type will be returned.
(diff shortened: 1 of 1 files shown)


github/activity_events.go
```diff
@@ -31,8 +31,50 @@ func (e Event) String() string {
 // (PushEvent), a value of the corresponding struct type will be returned.
```
shurcooL: > For recognized event types (PushEvent), a value of the corresponding struct type will be returned.

This change makes that comment out of date. There's no longer just the single recognized event type `PushEvent`.

gmlewis: Thanks, @shurcooL! I'll update it.

gmlewis: Done.

PR #350, opened 2016-05-04
Fix Int helper documentation and simplify code.

shurcooL:
This change makes it accurate and consistent with others.

/cc @willnorris

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -689,24 +689,12 @@ func cloneRequest(r *http.Request) *http.Request {
 
 // Bool is a helper routine that allocates a new bool value
 // to store v and returns a pointer to it.
-func Bool(v bool) *bool {
-	p := new(bool)
-	*p = v
-	return p
-}
+func Bool(v bool) *bool { return &v }
 
 // Int is a helper routine that allocates a new int value
 // to store v and returns a pointer to it.
-func Int(v int) *int {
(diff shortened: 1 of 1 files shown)


gmlewis:
I believe that both `Int` and `String` could be simplified as:
`return &v`
I've been wanting to benchmark the difference just out of curiosity.
But no big deal either way... the changes here LGTM.
Awaiting approval from @willnorris.

shurcooL:
@gmlewis Agree. `Bool` as well.

> I've been wanting to benchmark the difference just out of curiosity.

Why not... satisfy our curiosity? :)

```
$ go test -bench=. -test.benchmem
testing: warning: no tests to run
PASS
BenchmarkOldInt-8	100000000	        22.7 ns/op	       8 B/op	       1 allocs/op
BenchmarkNewInt-8	100000000	        22.1 ns/op	       8 B/op	       1 allocs/op
ok  	github.com/shurcooL/play/31	4.552s
```

Source: https://play.golang.org/p/bgBxkyyuiE.

Looks like the compiler produces same code for both variants.

Still, the new variant is better style and simpler, so it's worth doing.

gmlewis:
Very interesting!  Thanks for performing the experiment, @shurcooL!
I'm surprised that they are so similar.
👍

PR #352, opened 2016-05-05
Add support for the Source Import API

### LINKED ISSUE #291: Add support for Source Import API

### LINKED ISSUE #351: Add support for enhancements to Source Import API
announcement: https://developer.github.com/changes/2016-05-03-source-import-api-enhancements-for-working-with-git-lfs/

willnorris:
Fixes #291
Fixes #351

### DIFF
--- a/github/migrations_source_import.go
+++ b/github/migrations_source_import.go
@@ -1,0 +1,326 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import "fmt"
+
+// Import represents a repository import request.
+type Import struct {
+	// The URL of the originating repository.
+	VCSURL *string `json:"vcs_url,omitempty"`
+	// The originating VCS type. Can be one of 'subversion', 'git',
+	// 'mercurial', or 'tfvc'. Without this parameter, the import job will
+	// take additional time to detect the VCS type before beginning the
+	// import. This detection step will be reflected in the response.
+	VCS *string `json:"vcs,omitempty"`
+	// VCSUsername and VCSPassword are only used for StartImport calls that
+	// are importing a password-protected repository.
+	VCSUsername *string `json:"vcs_username,omitempty"`
+	VCSPassword *string `json:"vcs_password,omitempty"`
+	// For a tfvc import, the name of the project that is being imported.
+	TFVCProject *string `json:"tfvc_project,omitempty"`
+
+	// LFS related fields that may be preset in the Import Progress response
+
+	// Describes whether the import has been opted in or out of using Git
+	// LFS. The value can be 'opt_in', 'opt_out', or 'undecided' if no
+	// action has been taken.
+	UseLFS *string `json:"use_lfs,omitempty"`
+	// Describes whether files larger than 100MB were found during the
+	// importing step.
+	HasLargeFiles *bool `json:"has_large_files,omitempty"`
+	// The total size in gigabytes of files larger than 100MB found in the
+	// originating repository.
+	LargeFilesSize *int `json:"large_files_size,omitempty"`
+	// The total number of files larger than 100MB found in the originating
+	// repository. To see a list of these files, call LargeFiles.
+	LargeFilesCount *int `json:"large_files_count,omitempty"`
+
+	// Identifies the current status of an import.  An import that does not
+	// have errors will progress through these steps:
+	//
+	//     detecting - the "detection" step of the import is in progress
+	//         because the request did not include a VCS parameter. The
+	//         import is identifying the type of source control present at
+	//         the URL.
+	//     importing - the "raw" step of the import is in progress. This is
+	//         where commit data is fetched from the original repository.
+	//         The import progress response will include CommitCount (the
+	//         total number of raw commits that will be imported) and
+	//         Percent (0 - 100, the current progress through the import).
+	//     mapping - the "rewrite" step of the import is in progress. This
+	//         is where SVN branches are converted to Git branches, and
+	//         where author updates are applied. The import progress
+	//         response does not include progress information.
+	//     pushing - the "push" step of the import is in progress. This is
+	//         where the importer updates the repository on GitHub. The
+	//         import progress response will include PushPercent, which is
+	//         the percent value reported by git push when it is "Writing
+	//         objects".
+	//     complete - the import is complete, and the repository is ready
+	//         on GitHub.
+	//
+	// If there are problems, you will see one of these in the status field:
+	//
+	//     auth_failed - the import requires authentication in order to
+	//         connect to the original repository. Make an UpdateImport
+	//         request, and include VCSUsername and VCSPassword.
+	//     error - the import encountered an error. The import progress
+	//         response will include the FailedStep and an error message.
+	//         Contact GitHub support for more information.
+	//     detection_needs_auth - the importer requires authentication for
+	//         the originating repository to continue detection. Make an
+	//         UpdatImport request, and include VCSUsername and
+	//         VCSPassword.
+	//     detection_found_nothing - the importer didn't recognize any
+	//         source control at the URL.
+	//     detection_found_multiple - the importer found several projects
+	//         or repositories at the provided URL. When this is the case,
+	//         the Import Progress response will also include a
+	//         ProjectChoices field with the possible project choices as
+	//         values. Make an UpdateImport request, and include VCS and
+	//         (if applicable) TFVCProject.
+	Status        *string `json:"status,omitempty"`
+	CommitCount   *int    `json:"commit_count,omitempty"`
+	StatusText    *string `json:"status_text,omitempty"`
+	AuthorsCount  *int    `json:"authors_count,omitempty"`
+	Percent       *int    `json:"percent,omitempty"`
+	PushPercent   *int    `json:"push_percent,omitempty"`
+	URL           *string `json:"url,omitempty"`
+	HTMLURL       *string `json:"html_url,omitempty"`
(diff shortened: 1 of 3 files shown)


github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	// LFS. The value can be 'opt_in', 'opt_out', or 'undecided' if no
+	// action has been taken.
+	UseLFS *string `json:"use_lfs,omitempty"`
+	// Describes whether files larger than 100MB were found during the
+	// importing step.
+	HasLargeFiles *bool `json:"has_large_files,omitempty"`
+	// The total size in gigabytes of files larger than 100MB found in the
+	// originating repository.
+	LargeFilesSize *int `json:"large_files_size,omitempty"`
+	// The total number of of files larger than 100MB found in the
```
gmlewis: s/of of//

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	// originating repository. To see a list of these files, make a "Get
+	// Large Files" request.
+	LargeFilesCount *int `json:"large_files_count,omitempty"`
+
+	// Identifies the current status of an import.  An import that does not
+	// have errors will progress through these steps:
+	//
+	//     detecting - the "detection" step of the import is in progress
+	//         because the request did not include a vcs parameter. The
```
gmlewis: s/vcs/VCS/

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+func (i Import) String() string {
+	return Stringify(i)
+}
+
+// SourceImportAuthor identifies an author imported from a source repository.
+type SourceImportAuthor struct {
+	ID         *int    `json:"id,omitempty"`
+	RemoteID   *string `json:"remote_id,omitempty"`
```
gmlewis: You might want to add a comment pointing to the original JSON GitHub docs.
Is it possible that `RemoteID` is an `*int`?
I remember being surprised before by the GitHub API where I expected a `string` but got an `int`.

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+
+func (a SourceImportAuthor) String() string {
+	return Stringify(a)
+}
+
+// LargeFile identifies a file larger than 100MB found during a repository import.
+type LargeFile struct {
+	RefName *string `json:"ref_name,omitempty"`
+	Path    *string `json:"path,omitempty"`
+	OID     *string `json:"oid,omitempty"`
```
gmlewis: Is it possible that `OID` is an `*int`?
Maybe a URL do the GitHub docs here too would be good in the comment.

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	u := fmt.Sprintf("repos/%v/%v/import", owner, repo)
+	req, err := s.client.NewRequest("PUT", u, in)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypeImportPreview)
+
+	out := new(Import)
```
gmlewis: For posterity, I think that `out := &Import{}` is cleaner and easier to read, but I understand that this is more consistent with the rest of the repo.

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	ImportPercent *float64 `json:"import_percent,omitempty"`
+	URL           *string  `json:"url,omitempty"`
+	HTMLURL       *string  `json:"html_url,omitempty"`
+	AuthorsURL    *string  `json:"authors_url,omitempty"`
+	RepositoryURL *string  `json:"repository_url,omitempty"`
+	Message       *string  `json:"message,omitempty"`
+	FailedStep    *string  `json:"failed_step,omitempty"`
+}
+
+func (i Import) String() string {
```
gmlewis: Just for idiomatic Go, all `String` methods should have a comment... but for consistency with the rest of the repo, you can leave it.  Your call.

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return out, resp, err
+}
+
+// UpdateImport initiates a repository import.
+//
+// GitHub API docs: https://developer.github.com/v3/migration/source_imports/#start-an-import
```
gmlewis: https://developer.github.com/v3/migration/source_imports/#update-existing-import

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	u := fmt.Sprintf("repos/%v/%v/import/authors", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypeImportPreview)
+
+	authors := new([]SourceImportAuthor)
```
gmlewis: authors := []SourceImportAuthor{}

github/migrations_source_import.go
```diff
@@ -0,0 +1,313 @@
...
+	u := fmt.Sprintf("repos/%v/%v/import/large_files", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypeImportPreview)
+
+	files := new([]LargeFile)
```
gmlewis: ditto

github/migrations_source_import_test.go
```diff
@@ -0,0 +1,225 @@
...
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeImportPreview)
+		fmt.Fprint(w, `{"status":"complete"}`)
+	})
+
+	got, _, err := client.Migrations.ImportProgress("o", "r")
+	if err != nil {
+		t.Errorf("ImportProgress returned error: %v", err)
+	}
+	want := &Import{Status: String("complete")}
```
gmlewis: nit:
`if want := &...; !reflect...`

gmlewis:
LGTM
:+1:

github/migrations_source_import.go
shurcooL: Does https://github.com/golang/go/wiki/CodeReviewComments have a suggestion on which is the preferred style to use?

I don't see it mentioned there...

I don't have a strong preference of one way over the other, but I would like it if there was a suggested idiomatic way. I do like `new(Import)` slightly better than `&Import{}` as a way to initialize an _empty_ `*Import`.

shurcooL: Effective Go sorta sends an unclear message. On the one hand, in https://golang.org/doc/effective_go.html#allocation_new, it selects `new(SyncedBuffer)` over `&SyncedBuffer{}`, but in https://golang.org/doc/effective_go.html#composite_literals it says:

> The expressions `new(File)` and `&File{}` are equivalent.

shurcooL: Actually, disregard what I said above. This method follows the same style as all others in this package, and IMO it's out of scope to discuss changes to style in this PR.

E.g., see https://github.com/google/go-github/blob/842c551fdeae14c97c04ef490f601ae4d849a00c/github/repos.go#L271-L277.

If there are style changes to be be discussed/made, I think it's best to discuss that elsewhere and apply for the entire package.

github/migrations_source_import.go
shurcooL: Doing that would be inconsistent with https://github.com/google/go-github/blob/842c551fdeae14c97c04ef490f601ae4d849a00c/github/repos.go#L152-L158.

github/migrations_source_import.go
willnorris: done

github/migrations_source_import.go
willnorris: yep, you're absolutely right.  I had just copied the JSON-oriented docs, and forgot to go back and rename methods and fields to use the Go naming.  done.

github/migrations_source_import.go
willnorris: it's definitely a string:

```json
{
    "remote_id": "nobody@fc7da526-431c-80fe-3c8c-c148ff18d7ef"
}
```

github/migrations_source_import.go
willnorris: also definitely a string:

```json
{
    "oid": "d3d9446802a44259755d38e6d163e820",
}
```

github/migrations_source_import.go
willnorris: agreed... leaving as-is for now for consistency.

github/migrations_source_import.go
willnorris: String is one method that golint doesn't complain about having a godoc because it is so common.

github/migrations_source_import.go
willnorris: done

github/migrations_source_import.go
willnorris: agreed on leaving as-is for consistency

github/migrations_source_import_test.go
willnorris: I typically only do that style if the want value is really simple, like a primitive type.  The struct here does have only a single field, but it's still a bit more complex such that I like leaving it on separate lines, particularly in case we want or need to expand the fields it tests for.

Issue #324, opened 2016-04-03
Add support for commit squashing


willnorris:
Announcement: https://developer.github.com/changes/2016-04-01-squash-api-preview/
API Docs: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

Unfortunately, this will almost certainly require a breaking change to https://godoc.org/github.com/google/go-github/github#PullRequestsService.Merge

gsquire:
I can work on this. Would providing another function such that the existing merge function doesn't change be beneficial?

willnorris:
I think a breaking change is okay here.

PR #353, opened 2016-05-07
Add squash feature to pull requests

gsquire:
Issue #324

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -248,19 +248,30 @@ type PullRequestMergeResult struct {
 	Message *string `json:"message,omitempty"`
 }
 
+// PullRequestOptions lets you define how a pull request will be merged.
+type PullRequestOptions struct {
+	Squash bool
+}
+
 type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
+	Squash        *bool   `json:"squash,omitempty"`
 }
 
 // Merge a pull request (Merge Button™).
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-buttontrade
-func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string) (*PullRequestMergeResult, *Response, error) {
+func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string, options *PullRequestOptions) (*PullRequestMergeResult, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/merge", owner, repo, number)
 
-	req, err := s.client.NewRequest("PUT", u, &pullRequestMergeRequest{
-		CommitMessage: &commitMessage,
-	})
+	pullRequestBody := &pullRequestMergeRequest{CommitMessage: &commitMessage}
+	if options != nil {
+		pullRequestBody.Squash = &options.Squash
+	}
+	req, err := s.client.NewRequest("PUT", u, pullRequestBody)
+
+	// TODO: This header will be unnecessary when the API is no longer in preview.
+	req.Header.Set("Accept", mediaTypeSquashPreview)
 
 	if err != nil {
 		return nil, nil, err
--- a/github/pulls_test.go
+++ b/github/pulls_test.go
@@ -341,6 +341,7 @@ func TestPullRequestsService_Merge(t *testing.T) {
 
 	mux.HandleFunc("/repos/o/r/pulls/1/merge", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "PUT")
+		testHeader(t, r, "Accept", mediaTypeSquashPreview)
 		fmt.Fprint(w, `
 			{
 			  "sha": "6dcb09b5b57875f334f61aebed695e2e4193db5e",
@@ -349,7 +350,8 @@ func TestPullRequestsService_Merge(t *testing.T) {
 			}`)
 	})
 
-	merge, _, err := client.PullRequests.Merge("o", "r", 1, "merging pull request")
+	options := &PullRequestOptions{Squash: true}
+	merge, _, err := client.PullRequests.Merge("o", "r", 1, "merging pull request", options)
 	if err != nil {
 		t.Errorf("PullRequests.Merge returned error: %v", err)
 	}
--- a/github/github.go
+++ b/github/github.go
@@ -72,6 +72,9 @@ const (
 
 	// https://developer.github.com/changes/2016-05-12-reactions-api-preview/
 	mediaTypeReactionsPreview = "application/vnd.github.squirrel-girl-preview"
+
+	// https://developer.github.com/changes/2016-04-01-squash-api-preview/
+	mediaTypeSquashPreview = "application/vnd.github.polaris-preview+json"
 )
 
 // A Client manages communication with the GitHub API.


github/pulls_test.go
```diff
@@ -349,7 +349,7 @@ func TestPullRequestsService_Merge(t *testing.T) {
 			}`)
 	})
 
-	merge, _, err := client.PullRequests.Merge("o", "r", 1, "merging pull request")
+	merge, _, err := client.PullRequests.Merge("o", "r", 1, "merging pull request", false)
```
gmlewis: You probably want to add:
`testHeader(t, r, "Accept", mediaTypeSquashPreview)`
right after the `testMethod` above like is done in `repos_test.go` line 22

github/pulls.go
```diff
@@ -250,17 +250,20 @@ type PullRequestMergeResult struct {
...
 // GitHub API docs: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-buttontrade
-func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string) (*PullRequestMergeResult, *Response, error) {
+func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string, squash bool) (*PullRequestMergeResult, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/merge", owner, repo, number)
 
 	req, err := s.client.NewRequest("PUT", u, &pullRequestMergeRequest{
 		CommitMessage: &commitMessage,
+		Squash:        &squash,
 	})
+	req.Header.Add("Accept", mediaTypeSquashPreview)
```
gmlewis: You probably want to add a `TODO` comment above this line:
`// TODO: remove custom Accept header when this API fully launches.`

github/pulls.go
```diff
@@ -250,17 +250,20 @@ type PullRequestMergeResult struct {
 
 type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
+	Squash        *bool   `json:"squash"`
```
gmlewis: I think you can safely add `omitempty` to this:
```go
Squash        *bool   `json:"squash,omitempty"`
```

gsquire:
Hi @gmlewis I added what you commented on. I ended up adding a new function to check for headers with multiple keys since the API method will have more than one.

github/pulls.go
```diff
@@ -250,17 +250,21 @@ type PullRequestMergeResult struct {
...
 type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
+	Squash        *bool   `json:"squash,omitempty"`
 }
 
 // Merge a pull request (Merge Button™).
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-buttontrade
-func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string) (*PullRequestMergeResult, *Response, error) {
+func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string, squash bool) (*PullRequestMergeResult, *Response, error) {
```
willnorris: probably put this into an Options struct, so we don't have to make another breaking change if they add more options in the future.

gsquire:
Fixed the base conflicts and rebased onto master.

github/pulls_test.go
```diff
@@ -349,7 +350,8 @@ func TestPullRequestsService_Merge(t *testing.T) {
 			}`)
 	})
 
-	merge, _, err := client.PullRequests.Merge("o", "r", 1, "merging pull request")
+	options := &PullRequestOptions{Squash: false}
```
gmlewis: nit: how about making this option `true` for the test... otherwise the option is unnecessary.

gmlewis:
LGTM :+1: 
Awaiting approval from @willnorris

gsquire:
Rebased onto master again.

github/github_test.go
```diff
@@ -105,6 +105,16 @@ func testHeader(t *testing.T, r *http.Request, header string, want string) {
...
 
+func testCanonicalHeader(t *testing.T, r *http.Request, header string, want string) {
+	canonical := http.CanonicalHeaderKey(header)
+	for _, h := range r.Header[canonical] {
+		if h == want {
+			return
+		}
+	}
+	t.Errorf("Multiple header get, wanted %s but not found for header %s", want, header)
+}
```
shurcooL: Can you explain why `testCanonicalHeader` is needed, instead of using `testHeader`?

It looks like its implementation differs from `testHeader` in that it checks all values for the `header` key rather than just the first like `testHeader` does, is that right? Is that neccessary? Does it make sense to add a new method rather than modify `testHeader` to check all header values?

gsquire: Good question.

The reason I added this was because I found when testing, that I needed to check for multiple accept headers. So I found out the preferred way to test that is through the canonical form on the Go docs for [Header](https://golang.org/pkg/net/http/#Header).

shurcooL: Gotcha. So it sounds like you do want to test all header values, not just the first one, at least for this particular test.

I wonder how it was that all other tests got away with checking just the first header.

Well, since this is needed, then you should keep this func. While it's possible to change `testHeader` to test for all header values rather than the first, that would make all other tests that use it less strict than they are now, which may not be desired. Unless @willnorris has some comments on this.

Unfortunately, `testHeader` is not documented, so it's not obvious if its intention was to check all header values or only the first. I don't feel comfortable suggesting you change it without knowing that.

But if you keep `testCanonicalHeader`, a suggestion I'll make is 2 things:

1. rename it to something more descriptive, like `testHeaderAny` or `testHeaderMultiple`, and
2. document its intended behavior, so that people writing tests in the future will have an easier time choosing which of `testHeader` or `testHeaderAny` to use. It's not easy to figure out the difference in their behavior from just looking their source code.

willnorris: At quick glance, I still don't see why this method has multiple Accept headers.  We've never run into this problem before, so what's different here?

And if we do in fact need to support multiple values for a given header, I'd rather update `testHeader` to support that, which we can do in a way that doesn't require updating all the existing tests.  Something like:

```go
func testHeader(t *testing.T, r *http.Request, header string, want ...string) {
        got := r.Header[header]
        sort.Strings(got)
        sort.Strings(want)
        if !reflect.DeepEqual(got, want) {
                t.Errorf("Header.Get(%q) returned %s, want %s", header, got, want)
        }
}
```

gsquire: My understanding was to keep the original `Accept` header and add the new one per the [documentation on the endpoint](https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button). So that's why I added this function in even though it is temporary.

So I will defer the choice to keep/rename the function I added or adjust the function as you suggested.

github/pulls.go
```diff
@@ -248,19 +248,29 @@ type PullRequestMergeResult struct {
...
-		CommitMessage: &commitMessage,
-	})
+	pullRequestBody := &pullRequestMergeRequest{CommitMessage: &commitMessage}
+	if options != nil {
+		pullRequestBody.Squash = &options.Squash
+	}
+	req, err := s.client.NewRequest("PUT", u, pullRequestBody)
+
+	// TODO: This header will be unnecessary when the API is no longer in preview.
+	req.Header.Add("Accept", mediaTypeSquashPreview)
```
shurcooL: > At quick glance, I still don't see why this method has multiple Accept headers.

I see it now. /cc @willnorris It's because of this line, it does `req.Header.Add("Accept", ...Preview)` here, instead of what's done everywhere else which is `req.Header.Set("Accept", ...Preview)`, e.g., see [here](https://github.com/gsquire/go-github/blob/09a37d57ac99a8938444e44375db8660c2184696/github/reactions.go#L167).

`NewRequest` adds the "Accept" header value to `mediaTypeV3` [here](https://github.com/google/go-github/blob/c2beba44cffbb17740cc3ad8d54f2b303060027b/github/github.go#L209).

@gsquire, if you change this line to be `req.Header.Add` to be consistent with how preview accept headers are set everywhere else, this should get rid of the need for changing `testHeader`.

@willnorris, it'll be good to confirm how multiple custom media types in the `Accept` header are supposed to overlap, and whether the current behavior of always having at most one `Accept` header value is correct (I think it is, but I haven't confirmed).

gsquire: Ok, cool. It will be good to get confirmation on whether both are needed. If I can just call `Set` then I can use the original `testHeader` function as you said and get rid of the duplicate code I have.

shurcooL: Looked a little more into it, and I can confirm it should be `Set("Accept", "mediaTypeSquashPreview")` here, not `Add`.

`mediaTypeSquashPreview` already includes the entire media type value:

```
application/vnd.github.polaris-preview+json
```

Which overrides the `v3` (the default version) with `github.polaris-preview` preview API version.

```
application/vnd.github.v3+json
```

Having both `Accept` headers wouldn't make sense.

```bash
$ curl https://api.github.com/users/hubot -I \
  -H "Accept: application/vnd.github.v3.full+json" \
  -H "Accept: application/vnd.github.polaris-preview+json" 2>/dev/null | grep X-GitHub-Media-Type
X-GitHub-Media-Type: github.v3; param=full; format=json

$ curl https://api.github.com/users/hubot -I \
  -H "Accept: application/vnd.github.polaris-preview+json" 2>/dev/null | grep X-GitHub-Media-Type
X-GitHub-Media-Type: github.polaris-preview; format=json
```

I'm not sure what one would do if some endpoint wanted to opt into two API previews at the same time. Maybe that never happens?

gsquire: Thanks for looking into it @shurcooL and sorry for the trouble. I changed the code to use `Set` and removed the extra test function.

gsquire:
Rebased and fixed superfluous code.

github/pulls.go
```diff
@@ -248,19 +248,29 @@ type PullRequestMergeResult struct {
 	Message *string `json:"message,omitempty"`
 }
 
+type PullRequestOptions struct {
```
shurcooL: Go lint would say `PullRequestOptions` is exported and should be documented `// PullRequestOptions ...`.

shurcooL:
One remaining issue needs to be fixed, but I don't see any other issues. LGTM. But I haven't tested it out.

gsquire:
@shurcooL Okay, amended the commit message and added a simple comment.

PR #357, opened 2016-05-13
add support for Reactions API

### LINKED ISSUE #356: Add support for Reactions API
announcement: https://developer.github.com/changes/2016-05-12-reactions-api-preview/

gmlewis:
Fixes #356.

Change-Id: Iecc8d7d7c11a8ce9ed38c6d024683815495cb8b4

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -36,6 +36,7 @@ type Issue struct {
 	Milestone        *Milestone        `json:"milestone,omitempty"`
 	PullRequestLinks *PullRequestLinks `json:"pull_request,omitempty"`
 	Repository       *Repository       `json:"repository,omitempty"`
+	Reactions        *Reactions        `json:"reactions,omitempty"`
 
 	// TextMatches is only populated from search results that request text matches
 	// See: search.go and https://developer.github.com/v3/search/#text-match-metadata
@@ -131,6 +132,9 @@ func (s *IssuesService) listIssues(u string, opt *IssueListOptions) ([]Issue, *R
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
+
 	issues := new([]Issue)
 	resp, err := s.client.Do(req, issues)
 	if err != nil {
@@ -195,6 +199,9 @@ func (s *IssuesService) ListByRepo(owner string, repo string, opt *IssueListByRe
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
+
 	issues := new([]Issue)
 	resp, err := s.client.Do(req, issues)
 	if err != nil {
@@ -213,6 +220,9 @@ func (s *IssuesService) Get(owner string, repo string, number int) (*Issue, *Res
 	if err != nil {
 		return nil, nil, err
 	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
 
 	issue := new(Issue)
 	resp, err := s.client.Do(req, issue)
--- a/github/reactions.go
+++ b/github/reactions.go
@@ -1,0 +1,272 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import "fmt"
+
+// ReactionsService provides access to the reactions-related functions in the
+// GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/reactions/
+type ReactionsService struct {
+	client *Client
+}
+
+// Reaction represents a GitHub reaction.
+type Reaction struct {
+	// ID is the Reaction ID.
+	ID     *int `json:"id,omitempty"`
+	UserID *int `json:"user_id,omitempty"`
+	// Content is the type of reaction.
+	// Possible values are:
+	//     "+1", "-1", "laugh", "confused", "heart", "hooray".
+	Content *string `json:"content,omitempty"`
+}
+
+// Reactions represents a summary of GitHub reactions.
+type Reactions struct {
+	TotalCount *int    `json:"total_count,omitempty"`
+	PlusOne    *int    `json:"+1,omitempty"`
+	MinusOne   *int    `json:"-1,omitempty"`
+	Laugh      *int    `json:"laugh,omitempty"`
+	Confused   *int    `json:"confused,omitempty"`
+	Heart      *int    `json:"heart,omitempty"`
+	Hooray     *int    `json:"hooray,omitempty"`
+	URL        *string `json:"url,omitempty"`
+}
+
+func (r Reaction) String() string {
+	return Stringify(r)
+}
+
+// ListCommentReactions lists the reactions for a commit comment.
+//
+// GitHub API docs: https://developer.github.com/v3/reactions/#list-reactions-for-a-commit-comment
+func (s *ReactionsService) ListCommentReactions(owner, repo string, id int, opt *ListOptions) ([]*Reaction, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/comments/%v/reactions", owner, repo, id)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
+
+	var m []*Reaction
+	resp, err := s.client.Do(req, &m)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return m, resp, nil
+}
+
+// CreateCommentReaction creates a reaction for a commit comment.
+// Note that if you have already created a reaction of type content, the
+// previously created reaction will be returned with Status: 200 OK.
+//
+// GitHub API docs: https://developer.github.com/v3/reactions/#create-reaction-for-a-commit-comment
+func (s ReactionsService) CreateCommentReaction(owner, repo string, id int, content string) (*Reaction, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/comments/%v/reactions", owner, repo, id)
+
+	body := &Reaction{Content: String(content)}
+	req, err := s.client.NewRequest("POST", u, body)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
+
+	m := &Reaction{}
+	resp, err := s.client.Do(req, m)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return m, resp, nil
+}
+
+// ListIssueReactions lists the reactions for an issue.
+//
+// GitHub API docs: https://developer.github.com/v3/reactions/#list-reactions-for-an-issue
+func (s *ReactionsService) ListIssueReactions(owner, repo string, number int, opt *ListOptions) ([]*Reaction, *Response, error) {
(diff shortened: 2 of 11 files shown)


github/issues.go
```diff
@@ -36,6 +36,7 @@ type Issue struct {
 	Milestone        *Milestone        `json:"milestone,omitempty"`
 	PullRequestLinks *PullRequestLinks `json:"pull_request,omitempty"`
 	Repository       *Repository       `json:"repository,omitempty"`
+	Reactions        *ReactionsSummary `json:"reactions,omitempty"`
```
shurcooL: Is there a reason the type `ReactionsSummary` isn't named `Reactions`? That would be more consistent with the json field name (here and elsewhere). It'd also be more consistent with the other similar types (see lines above).

gmlewis: I didn't want any confused with `type Reaction struct {...}`.

In the GitHub API docs, whenever you see the "`reactions` object", the text describes them as "Reactions summary", so I felt this was appropriate.

@willnorris, what do you think?

github/reactions.go
```diff
@@ -0,0 +1,256 @@
...
+	// Possible values are:
+	//     "+1", "-1", "laugh", "confused", "heart", "hooray".
+	Content *string `json:"content,omitempty"`
+}
+
+// ReactionsSummary represents a summary of GitHub reactions.
+type ReactionsSummary struct {
+	TotalCount *int    `json:"total_count,omitempty"`
+	Plus1      *int    `json:"+1,omitempty"`
+	Minus1     *int    `json:"-1,omitempty"`
```
shurcooL: Have you considered `PlusOne` and `MinusOne`? Do you think that's more readable or less?

One (possible irrelevant) observation is that had the reaction been named `1+`, then you couldn't name it `1Plus`, but you could name it `OnePlus`. I think this demonstrates that this approach is slightly more scaleable.

Another possibility to consider is to write these as `ThumbsUp` and `ThumbsDown`. The rationale is that `:+1:` has an alias `:thumbsup:`, and `:-1:` has an alias `:thumbsdown:`. Proof: :+1: == :thumbsup:, :-1: == :thumbsdown:`. Also, if you check https://api.github.com/emojis, both variants show up and have equal values.

gmlewis: I'm totally open to suggestion...  I like both options... `PlusOne`/`MinusOne` or `ThumpsUp`/`ThumpsDown`... which do you prefer?

github/issues.go
shurcooL: > I didn't want any confusion with `type Reaction struct {...}`.

I was worried there was a stronger reason, like a naming conflict because `Reactions` was already taken by something else.

If that's not it, I think `Reactions` is much nicer and more consistent.

IMO `Reaction` and `Reactions` are fine to have, and they are great names/description for the types. We also have https://godoc.org/github.com/google/go-github/github#Page and https://godoc.org/github.com/google/go-github/github#Pages.

gmlewis: You convinced me.  :+1:

github/reactions.go
shurcooL: I like both of those options too. It doesn't seem that I have a preference for either one, because I can't choose what to say. :)

Maybe I have an ever so slight preference towards `PlusOne`/`MinusOne` over `ThumbsUp`/`ThumbsDown`, because I tend to think of those 2 emoji in my mind as "plus one/minus one", I type them as `:+1:`, and it's slightly shorter. Plus more different/easier to tell apart.

gmlewis: Done.

github/reactions.go
```diff
@@ -0,0 +1,256 @@
...
+
+	return m, resp, nil
+}
+
+// CreatePullRequestReviewCommentReaction creates a reaction for a pull request review comment.
+// Note that if you have already created a reaction of type content, the
+// previously created reaction will be returned with Status: 200 OK.
+//
+// GitHub API docs: https://developer.github.com/v3/reactions/#create-reaction-for-an-issue-comment
+func (s ReactionsService) CreatePullRequestReviewCommentReaction(owner, repo string, id int, content string) (*Reaction, *Response, error) {
```
shurcooL: I suggest leaving out `Review` from `CreatePullRequestReviewCommentReaction`. It's not neccessary, doesn't help, and this method name is already quite long.

Consider `CreateCommentReaction` method - you're not calling it `CreateCommitCommentReaction`.

If you make this change, the mapping between method names and API endpoints will be more consistent:

```
CreateCommentReaction
"repos/%v/%v/comments/%v/reactions"

CreateIssueReaction
"repos/%v/%v/issues/%v/reactions"

CreateIssueCommentReaction
"repos/%v/%v/issues/comments/%v/reactions"

CreatePullRequestCommentReaction
"repos/%v/%v/pulls/comments/%v/reactions"
```

gmlewis: SGTM.  Done.

shurcooL:
Thanks for considering and resolving the comments I've posted so far Glenn.

Apologies for a spread-over-long-amount of time review rather than all-at-once, but I'm discovering issues as I make progress on using this API to implement something (and no way to have drafts in GitHub). Consider the review a WIP until I send the last message saying so. :)

I've found an issue that should be fixed.

All the List* endpoints currently don't take any options, making it impossible to get anything but the first page of results. This is problematic for entries with huge number of reactions (e.g., https://github.com/golang/go/issues/15292#issue-148234131).

So, the List* methods should take an extra `opt *ListOptions` parameter and use it, to enable pagination to be controlled. E.g., similar to https://github.com/gmlewis/go-github/blob/a68a8707f3c18a3a47bb0975ca29a9239792dc01/github/activity_events.go#L88-L89.

github/repos_comments.go
```diff
@@ -109,6 +116,9 @@ func (s *RepositoriesService) GetComment(owner, repo string, id int) (*Repositor
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
+
 	c := new(RepositoryComment)
 	resp, err := s.client.Do(req, c)
 	if err != nil {
```
shurcooL: Do you know if `UpdateComment` needs to include the custom Accept header too, or not?

shurcooL:
Ok, finished reviewing. See new comments above, but aside from that, I don't see anything else, LGTM.

github/repos_comments.go
gmlewis: Thanks for the extensive review and testing, Dmitry!  I really appreciate it.

I've added the `ListOptions` and updated the tests, but I don't think any docs need updating based upon the docs found in your example from `activity_events.go`.

According to the docs for `UpdateComment`:
https://developer.github.com/v3/repos/comments/#update-a-commit-comment
it does not return the reactions object.

However, I just ran an experiment... If I add the custom `Accept` header, it *does* return the `Reactions` struct, however it does *not* allow me to modify the `Reactions` during the `UpdateComment`!!!

So I'm thinking it would cause less user confusion if we leave *off* the custom Accept header and let the user get an error if they provide a `Reactions` struct in the provided `RepositoryComment` instead of accepting it and giving them the false hope that they can update it.

Thoughts?

Meanwhile, I will update the code with the latest requested changes.
Thanks again for your reviews!

shurcooL: > I've added the `ListOptions` and updated the tests, but I don't think any docs need updating based upon the docs found in your example from `activity_events.go.`

Sounds good.

> According to the docs for `UpdateComment`:
https://developer.github.com/v3/repos/comments/#update-a-commit-comment
it does not return the reactions object.

Oh, I see now that the following 3 endpoints are actually documented with a new "Reactions summary" section to have reactions present with that accept header set:

- https://developer.github.com/v3/repos/comments/#list-commit-comments-for-a-repository
- https://developer.github.com/v3/repos/comments/#list-comments-for-a-single-commit
- https://developer.github.com/v3/repos/comments/#get-a-single-commit-comment

The `UpdateComment` endpoint indeed does not mention it.

I didn't see that before, which is why I asked that question.

> So I'm thinking it would cause less user confusion if we leave off the custom Accept header and let the user get an error if they provide a Reactions struct in the provided RepositoryComment instead of accepting it and giving them the false hope that they can update it.

> Thoughts?

Full agreement. You should follow the documentation that only mentions the 3 endpoints, and not add it for `UpdateComment`.

github/issues.go
```diff
@@ -131,6 +132,9 @@ func (s *IssuesService) listIssues(u string, opt *IssueListOptions) ([]Issue, *R
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeReactionsPreview)
```
shurcooL: I don't think this is related to this library, but rather the GitHub API... But I'll mention it here because I don't know of a better place.

It's just an observation. I see that supplying this custom Accept header has a small change on the behavior of `pull_request` field inside the `Issue` struct when it's an issue (not pull request).

Anyway, the change is that with the custom Accept header, that `pull_request` field is no longer omitted from the JSON. Notice:

```diff
~ $ diff -u <(curl 'https://api.github.com/repos/gorilla/mux/issues/2') \
            <(curl 'https://api.github.com/repos/gorilla/mux/issues/2' -H "Accept: application/vnd.github.squirrel-girl-preview")
--- /dev/fd/63	2016-05-14 13:36:15.000000000 -0700
+++ /dev/fd/62	2016-05-14 13:36:15.000000000 -0700
@@ -65,6 +65,22 @@
   "created_at": "2012-10-05T13:25:48Z",
   "updated_at": "2016-03-12T19:50:58Z",
   "closed_at": null,
+  "pull_request": {
+    "url": null,
+    "html_url": null,
+    "diff_url": null,
+    "patch_url": null
+  },
   "body": "Old issue: http://code.google.com/p/gorilla/issues/detail?id=51\r\n\r\nAccording to section 14.23 of RFC 2616 the Host header can include the port number if the default value of 80 is not used.\r\n\r\nSee https://github.com/gorilla/mux/blob/master/mux_test.go#L119 for a commented-out test case that fails.\r\n\r\nPotential fix is editing getHost() https://github.com/gorilla/mux/blob/master/regexp.go#L237 so it has a flag as to whether to strip off the port information. A new field would need to be probably need to be added somewhere such as routeRegexp.\r\n\r\nThe main reason I was using Host() match with a port in the first place is for trying to build reverse routes that include the absolute path as opposed to a relative one.",
-  "closed_by": null
+  "closed_by": null,
+  "reactions": {
+    "url": "https://api.github.com/repos/gorilla/mux/issues/2/reactions",
+    "total_count": 0,
+    "+1": 0,
+    "-1": 0,
+    "laugh": 0,
+    "hooray": 0,
+    "confused": 0,
+    "heart": 0
+  }
 }
```

But it means, with this PR applied, it's no longer to do this to check if an issue is a PR:

```Go
for _, issue := range ghIssuesAndPRs {
	// Filter out PRs.
	if issue.PullRequestLinks != nil {
		continue
	}
	... process issues only
```

Instead, one must now do something like this:

```Go
for _, issue := range ghIssuesAndPRs {
	// Filter out PRs.
	if issue.PullRequestLinks != nil && issue.PullRequestLinks.URL != nil {
		continue
	}
	... process issues only
```

Anyway, I wanted to point this out, because I don't think this change is intended by you @gmlewis. At the same time, it's not really our place to be modifying JSON that GitHub API returns. Maybe this is a common thing to all preview APIs? Or maybe it's an issue we should let them know about? I don't know.

shurcooL: If you read the GitHub API docs carefully, it sounds like a bug on their end. From https://developer.github.com/v3/issues/#list-issues:

> This endpoint may also return pull requests in the response. If an issue is a pull request, the object will include a pull_request key.

But with this Accept header set, `pull_request` key is included in all objects, both issues and PRs (but the contents of `pull_request` do vary, of course).

gmlewis: Very interesting!  Yes, it sounds like a bug on GitHub's side.  Would you mind reporting it, Dmitry?  I'm heading off to OSCON.  Thanks!

shurcooL: Yeah, I will. I think it's orthogonal to this PR so feel free to ignore it.

shurcooL: I've emailed them, they told me they've opened an internal issue to investigate. It appears to be fixed now, so nothing more needs to be done here. /cc @izuzak

izuzak: > /cc @izuzak It appears to be fixed now, so nothing more needs to be done here.

Yep, should be fixed now. Thanks again for the report! :bow:

PR #358, opened 2016-05-17
fix unit tests for go1.7 release

dsnet:
An upcoming change to Go 1.7 allows the json package to be able to encode maps
with the integer keys. Thus, this is no longer a good way to generate intentional
JSON failures.

Reference: https://github.com/golang/go/commit/f05c3aa24d815cd3869153750c9875e35fc48a6e


shurcooL:
LGTM.

PR #359, opened 2016-05-23
add Message field for "custom" errors

vetinari:
see https://developer.github.com/v3/#client-errors

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -497,6 +497,7 @@ type Error struct {
 	Resource string `json:"resource"` // resource on which the error occurred
 	Field    string `json:"field"`    // field on which the error occurred
 	Code     string `json:"code"`     // validation error code
+	Message  string `json:"message"`  // error message for "custom" error codes
 }
 
 func (e *Error) Error() string {
--- a/github/github.go
+++ b/github/github.go
@@ -438,7 +438,7 @@ type ErrorResponse struct {
 	// Most errors will also include a documentation_url field pointing
 	// to some content that might help you resolve the error, see
 	// https://developer.github.com/v3/#client-errors
-	DocumentationURL string `json:"documentation_url"`
+	DocumentationURL string `json:"documentation_url,omitempty"`
 }
 
 func (r *ErrorResponse) Error() string {
@@ -504,10 +504,7 @@ type Error struct {
 	Resource string `json:"resource"` // resource on which the error occurred
 	Field    string `json:"field"`    // field on which the error occurred
 	Code     string `json:"code"`     // validation error code
-	// For .Code == "custom" we always get a message.
-	// The github.User.CreateKey() returns such a code when trying to add a key
-	// which is already in use.
-	Message string `json:"message"`
+	Message  string `json:"message"`  // Message describing the error. Errors with Code == "custom" will always have this set.
 }
 
 func (e *Error) Error() string {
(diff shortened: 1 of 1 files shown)


gmlewis:
Do you have an example of a "custom" error response?
Do you also want to provide a `DocumentationURL` field as well?
You might want to separate out these two new fields with a comment, possibly including the https://developer.github.com/v3/#client-errors URL and when they might expect to get a "custom" error.

vetinari:
The `client.Users.CreateKey()` returns a 422 with
```json
{
  "documentation_url": "https://developer.github.com/enterprise/2.6/v3/users/keys/#create-a-public-key",
  "errors": [
    {
      "message": "key is already in use",
      "field": "key",
      "code": "custom",
      "resource": "PublicKey"
    }
  ],
  "message": "Validation Failed"
}
```
as body when a key is already in use.
The current code is used like:
```go
        if _, _, err := cl.Users.CreateKey(&github.Key{Key: &sshKey, Title: &mailAddr}); err != nil {
            if gherr, ok := err.(*github.ErrorResponse); ok && gherr.Errors[0].Code == "custom" {
                return fmt.Errorf("failed to create key for %s: %s, see %s\n", userName, gherr.Errors[0].Message, gherr.DocumentationURL)
            }
            return fmt.Errorf("failed to create key for %s: %s\n", userName, err)
        }
        return nil
```

github/github.go
```diff
@@ -435,6 +435,10 @@ type ErrorResponse struct {
 		Reason    string     `json:"reason,omitempty"`
 		CreatedAt *Timestamp `json:"created_at,omitempty"`
 	} `json:"block,omitempty"`
+	// Most errors willi also include a documentation_url field pointing
```
shurcooL: `willi` has a typo.

gmlewis:
LGTM :+1: 
@shurcooL - does this LGTY?

github/github.go
```diff
@@ -435,6 +435,10 @@ type ErrorResponse struct {
 		Reason    string     `json:"reason,omitempty"`
 		CreatedAt *Timestamp `json:"created_at,omitempty"`
 	} `json:"block,omitempty"`
+	// Most errors will also include a documentation_url field pointing
+	// to some content that might help you resolve the error, see
+	// https://developer.github.com/v3/#client-errors
+	DocumentationURL string `json:"documentation_url"`
```
shurcooL: Since this applies to "most errors", not all errors, should it include `,omitempty`?

github/github.go
```diff
@@ -490,13 +494,20 @@ These are the possible validation error codes:
...
 GitHub API docs: http://developer.github.com/v3/#client-errors
 */
 type Error struct {
 	Resource string `json:"resource"` // resource on which the error occurred
 	Field    string `json:"field"`    // field on which the error occurred
 	Code     string `json:"code"`     // validation error code
+	// For .Code == "custom" we always get a message.
+	// The github.User.CreateKey() returns such a code when trying to add a key
+	// which is already in use.
+	Message string `json:"message"`
```
shurcooL: I think you can simplify the docs here. Example of when it happens is already described above, in the doc for `Error` type.

This can be just:

```Go
// Message field describing the error. Errors with Code == "custom" will always have this set.
```

shurcooL:
Some minor additional feedback, aside from that LGTM.

gmlewis:
Merged in d2e7389.

PR #363, opened 2016-06-07
remove org permission content type

### LINKED ISSUE #262: remove org permission content type
The org permissions changes are now official, so the custom media type is no longer necessary.  This would also be a good time to make sure that we have coverage for all of those changes (though I'm pretty sure we do, since we use this part of the API pretty extensively inside Google)

announcement: https://developer.github.com/changes/2016-01-05-api-enhancements-for-working-with-organization-permissions-are-now-official/

gmlewis:
Fixes #262.

Change-Id: Ie9acf859346fce6b797b039d35b17e46f404e2a3

### DIFF
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -94,10 +94,6 @@ func (s *OrganizationsService) CreateTeam(org string, team *Team) (*Team, *Respo
 		return nil, nil, err
 	}
 
-	if team.Privacy != nil {
-		req.Header.Set("Accept", mediaTypeOrgPermissionPreview)
-	}
-
 	t := new(Team)
 	resp, err := s.client.Do(req, t)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


willnorris:
LGTM

gmlewis:
I made the changes mentioned in #262. PTAL.

github/orgs_teams.go
```diff
@@ -226,14 +214,15 @@ func (s *OrganizationsService) ListTeamRepos(team int, opt *ListOptions) ([]Repo
 // permissions team has for that repo.
 //
 // GitHub API docs: http://developer.github.com/v3/orgs/teams/#get-team-repo
```
shurcooL: This might be out of scope for this PR, but this link http://developer.github.com/v3/orgs/teams/#get-team-repo doesn't seem to be valid anymore. Should it be updated to something else, or replaced by the new one you added on the following line?

shurcooL:
Left one comment, otherwise LGTM.

github/orgs_teams.go
gmlewis: Thanks, @shurcooL!  TL;DR: I'll delete the first link.
One possible option is to replace it with https://developer.github.com/v3/orgs/teams/#list-team-repos since it describes this call without the custom media type header, but when comparing the docs content of the two links, it seems to me that the first link isn't very useful compared to the second.  I'll delete it.

PR #367, opened 2016-06-15
document 'since' pagination for Users.ListAll

gmlewis:
Fixes #278.

Change-Id: Ia1d0c1f6b395d493714d91e0f6e79dc2506f8fde

### DIFF
--- a/github/users.go
+++ b/github/users.go
@@ -144,6 +144,22 @@ type UserListOptions struct {
 
 // ListAll lists all GitHub users.
 //
+// To paginate through all users, populate 'Since' with the ID of the last user.
+// Example (warning: this will quickly eat through your GitHub API quota!):
+//
+// 	opts := &github.UserListOptions{}
+// 	for {
+// 		users, _, err := client.Users.ListAll(opts)
+// 		if err != nil {
+// 			log.Fatalf("error listing users: %v", err)
+// 		}
+// 		if len(users) == 0 {
+// 			break
+// 		}
+// 		opts.Since = *users[len(users)-1].ID
+// 		// Process users...
+// 	}
+//
 // GitHub API docs: http://developer.github.com/v3/users/#get-all-users
 func (s *UsersService) ListAll(opt *UserListOptions) ([]User, *Response, error) {
 	u, err := addOptions("users", opt)


shurcooL:
The actual content in this PR looks good to me, I don't see any issues. LGTM.

I do have 2 style comments that I'll leave inline for your discretion. The first one (writing `gofmt`ed Go code, even if it's inside a comment) is highly recommended. The second one is completely optional and up to you.

github/users.go
```diff
@@ -144,6 +144,22 @@ type UserListOptions struct {
...
+//        users, _, err := client.Users.ListAll(opts)
+//        if err != nil {
+//          log.Fatalf("error listing users: %v", err)
+//        }
+//        if len(users) == 0 {
+//          break
+//        }
+//        opts.Since = *users[len(users)-1].ID
+//        // Process users...
+//      }
```
shurcooL: I would suggest making this Go code `gofmt`ed. Right now it uses 1 space indent, followed by 2 spaces. How about this:

https://gist.githubusercontent.com/shurcooL/43a5922765e1918ae00f29f315e9b41c/raw/e1de820beeece8b3a6a881e8f0d1913901e7c71c/comment.go

(Note that the comment always begins with 2 slashes and a space. The rest is its content. The code section is indented by 1 tab. This is the most idiomatic and consistent way to write Go comments, at least if you're as picky about them as me.)

github/users.go
```diff
@@ -144,6 +144,22 @@ type UserListOptions struct {
 
 // ListAll lists all GitHub users.
 //
+// To paginate through all users, populate 'Since' with the ID of the last user.
+// Example (warning: this will quickly eat through your GitHub API quota!):
```
shurcooL: My second comment is completely optional.

Have you considered using real Go examples here, instead of embedding it into the comment? E.g., https://godoc.org/testing#hdr-Examples. Would that be better or not?

github/users.go
gmlewis: Very nice. Thanks, @shurcooL!

github/users.go
gmlewis: In general, I really like real Go examples and think that it improves the documentation.
However, I'm concerned about making this particular one a real example because then it will actually be run during "go test" which we definitely do not want to do.
So I'll leave this as-is. Thanks!

shurcooL: > However, I'm concerned about making this particular one a real example because then it will actually be run during "go test" which we definitely do not want to do.

Makes sense.

But to clarify, examples only run as part of tests if they include an `// Output:` comment at the end. Otherwise, `go test` simply checks that it compiles, but does not execute it.

> Example functions may include a concluding line comment that begins with "Output:" and is compared with the standard output of the function when the tests are run.

Did you already know that, or is this a new factor?

gmlewis: Ah!  I did not know that... but I see it is written right there... sorry I didn't read it through fully.
That is definitely a new factor.

OK, now comes the question: should the example go in `users_test.go` or do you have a preference for a different file naming convention for examples?

I can make a new PR that moves this example to where we think it should go.

willnorris: I did start playing around with inline examples in 32a84d1 and discussed it a little in #323.  One things I haven't liked about them (at least how the current ones are written), is that because they're in the github package, the code is not directly copy and pasteable.  Though now I wonder if we put them in `package github_test`, would that work since we would import the github package like normal.  But would godoc inline the examples with the code if they're in the github_test package?  That might be worth trying out to see, since I actually don't know.

PR #369, opened 2016-06-15
add support for Issue Timeline API

### LINKED ISSUE #360: Add support for Issue Timeline API
announcement: https://developer.github.com/changes/2016-05-23-timeline-preview-api/

docs: https://developer.github.com/v3/issues/timeline/

gmlewis:
Fixes #360.

Change-Id: I69846ddeb5fd289ec81712cccaefde3f3931fd09

### DIFF
--- a/github/issues_timeline.go
+++ b/github/issues_timeline.go
@@ -1,0 +1,148 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"fmt"
+	"time"
+)
+
+// Timeline represents an event that occurred around an Issue or Pull Request.
+//
+// It is similar to an IssueEvent but may contain more information.
+// GitHub API docs: https://developer.github.com/v3/issues/timeline/
+type Timeline struct {
+	ID        *int    `json:"id,omitempty"`
+	URL       *string `json:"url,omitempty"`
+	CommitURL *string `json:"commit_url,omitempty"`
+
+	// The User object that generated the event.
+	Actor *User `json:"actor,omitempty"`
+
+	// Event identifies the actual type of Event that occurred. Possible values
+	// are:
+	//
+	//     assigned
+	//       The issue was assigned to the assignee.
+	//
+	//     closed
+	//       The issue was closed by the actor. When the commit_id is present, it
+	//       identifies the commit that closed the issue using "closes / fixes #NN"
+	//       syntax.
+	//
+	//     commented
+	//       A comment was added to the issue.
+	//
+	//     committed
+	//       A commit was added to the pull request's 'HEAD' branch. Only provided
+	//       for pull requests.
+	//
+	//     cross-referenced
+	//       The issue was referenced from another issue. The 'source' attribute
+	//       contains the 'id', 'actor', and 'url' of the reference's source.
+	//
+	//     demilestoned
+	//       The issue was removed from a milestone.
+	//
+	//     head_ref_deleted
+	//       The pull request's branch was deleted.
+	//
+	//     head_ref_restored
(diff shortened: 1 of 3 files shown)


github/issues_timeline.go
```diff
@@ -0,0 +1,147 @@
...
+	//
+	//     locked
+	//       The issue was locked by the actor.
+	//
+	//     mentioned
+	//       The actor was @mentioned in an issue body.
+	//
+	//     merged
+	//     The issue was merged by the actor. The 'commit_id' attribute is the
+	//     SHA1 of the HEAD commit that was merged.
```
shurcooL: The 2-space indent is missing here, making this inconsistent.

github/issues_timeline.go
```diff
@@ -0,0 +1,147 @@
...
+	//       The assignee was unassigned from the issue.
+	//
+	//     unlabeled
+	//       A label was removed from the issue.
+	//
+	//     unlocked
+	//       The issue was unlocked by the actor.
+	//
+	//     unsubscribed
+	//     The actor unsubscribed to stop receiving notifications for an issue.
```
shurcooL: Also here.

github/issues_timeline.go
```diff
@@ -0,0 +1,147 @@
...
+	//
+	Event *string `json:"event,omitempty"`
+
+	// The string SHA of a commit that referenced this Issue.
```
shurcooL: > Timeline represents an event that occurred around an Issue or Pull Request.

Should this say "this Issue or Pull Request" then?

github/issues_timeline.go
```diff
@@ -0,0 +1,147 @@
...
+	CommitID *string `json:"commit_id,omitempty"`
+	// The timestamp indicating when the event occurred.
+	CreatedAt *time.Time `json:"created_at,omitempty"`
+	// The Label object including `name` and `color` attributes. Only provided for
+	// 'labeled' and 'unlabeled' events.
+	Label *Label `json:"label,omitempty"`
+	// The User object which was assigned to (or unassigned from) this Issue.
```
shurcooL: Also here.

github/issues_timeline.go
```diff
@@ -0,0 +1,147 @@
...
+	//       The issue was unlocked by the actor.
+	//
+	//     unsubscribed
+	//     The actor unsubscribed to stop receiving notifications for an issue.
+	//
+	Event *string `json:"event,omitempty"`
+
+	// The string SHA of a commit that referenced this Issue.
+	CommitID *string `json:"commit_id,omitempty"`
+	// The timestamp indicating when the event occurred.
```
shurcooL: There's a `commit_url` property in the example at https://developer.github.com/v3/issues/timeline/ but it's not available here. Do we want to do anything about `commit_url`?

gmlewis: SGTM. I'll add it to the top. Nice catch, @shurcooL!

github/issues_timeline_test.go
```diff
@@ -0,0 +1,40 @@
...
+			"page":     "1",
+			"per_page": "2",
+		})
+		fmt.Fprint(w, `[{"id":1}]`)
+	})
+
+	opt := &ListOptions{Page: 1, PerPage: 2}
+	events, _, err := client.Issues.ListIssueTimeline("o", "r", 1, opt)
+
+	if err != nil {
```
shurcooL: There's a blank line between assigning `err` and checking it. I suggest removing it. It's not needed, and makes the code less readable (the blank line suggests a two sections of code are separate concerns).

I can see there are a few other rare places where this is done in this code base, but the vast majority of code does not, so this would also increase consistency.

shurcooL:
There's one unnecessary blank line between a method call and its error checking, see comment above.

LGTM.

I have not tested this, I've only reviewed the code.

Issue #79, opened 2014-01-31
Using pointers all over the places?


unknwon:
I'm just curious, even for bool, int and string, is there any specific reason for doing this?

Thanks!

Issue #19, opened 2013-07-09
issue setting empty values


willnorris:
Depending on how our types are defined, we run into various interesting problems with setting empty values.
### Option 1

This is the current behavior throughout the library.

As currently setup, we can't set empty values at all on Create or Update methods.  For example, this won't work to remove the description from an existing repository:

``` go
// remove the repository description
r := &github.Repository{Description:""}
client.Repositories.Edit("user", "repo", r)
```

That is actually a no-op because the empty Description field gets dropped, since it is currently defined with `omitempty`.  Fortunately, there are only a handful of mutable values where the zero value is actually meaningful, most notably `bool` values.
### Option 2

We could then instead drop the `omitempty`, but that has potentially _really_ bad side-effects, particularly on edit methods.  Take the above code sample.  Without `omitempty` on any of our fields, this would work as intended.  However, it would also have the unintended side-effect of wiping out the repo homepage, [and once we add the additional fields...] making it public, and disabling issues and the wiki for the repo, since all of those fields would be passing their zero values.  The solution there is to first call `Get`, update the response, and then call `Edit` with the updated object:

``` go
// remove the repository description
r, _ := client.Repositories.Get("user", "repo")
r.Description = ""
client.Repositories.Edit("user", "repo", r)
```

If you forget to follow that flow, you're gonna have a bad time.
### Option 3

The third option is to do what [goprotobuf](https://code.google.com/p/goprotobuf) does and actually use pointers for all non-repeated fields, since that allows a clear distinction between "unset" (nil) and "empty" (non-nil, zero value).  That would result in types like:

``` go
type Repository struct {
    ID          *int        `json:"id,omitempty"`
    Owner       *User       `json:"owner,omitempty"`
    Name        *string     `json:"name,omitempty"`
    Description *string     `json:"description,omitempty"`
    CreatedAt   *time.Time  `json:"created_at,omitempty"`
    PushedAt    *time.Time  `json:"pushed_at,omitempty"`
    UpdatedAt   *time.Time  `json:"updated_at,omitempty"`
}
```

This is by far the safest approach, but does make working with the library a bit cumbersome, since creating pointers to primitive types takes a little extra work (multiplied by the number of fields you are setting).  For example, the above code sample now becomes:

``` go
// remove the repository description
d = ""
r := &github.Repository{Description:&d}
client.Repositories.Edit("user", "repo", r)
```

[...]

PR #373, opened 2016-06-16
add support for new repository invitations

### LINKED ISSUE #371: Add support for new repository invitations
announcement: https://developer.github.com/changes/2016-06-14-repository-invitations/

gmlewis:
Fixes #371.

Change-Id: I97f2fc28ce048442ee8af3bd11c48b5b2084e09b

### DIFF
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -58,19 +58,22 @@ type RepositoryAddCollaboratorOptions struct {
 	//     push - team members can pull and push, but not administer this repository
 	//     admin - team members can pull, push and administer this repository
 	//
-	// Default value is "pull".  This option is only valid for organization-owned repositories.
+	// Default value is "push".  This option is only valid for organization-owned repositories.
 	Permission string `json:"permission,omitempty"`
 }
 
 // AddCollaborator adds the specified Github user as collaborator to the given repo.
 //
-// GitHub API docs: http://developer.github.com/v3/repos/collaborators/#add-collaborator
+// GitHub API docs: https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator
 func (s *RepositoriesService) AddCollaborator(owner, repo, user string, opt *RepositoryAddCollaboratorOptions) (*Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/collaborators/%v", owner, repo, user)
 	req, err := s.client.NewRequest("PUT", u, opt)
 	if err != nil {
 		return nil, err
 	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeRepositoryInvitationsPreview)
 
 	return s.client.Do(req, nil)
 }
--- a/github/repos_invitations.go
+++ b/github/repos_invitations.go
@@ -1,0 +1,91 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import "fmt"
+
+// RepositoryInvitation represents an invitation to collaborate on a repo.
+type RepositoryInvitation struct {
+	ID      *int        `json:"id,omitempty"`
+	Repo    *Repository `json:"repository,omitempty"`
+	Invitee *User       `json:"invitee,omitempty"`
+	Inviter *User       `json:"inviter,omitempty"`
+
+	// Permissions represents the permissions that the associated user will have
+	// on the repository. Possible values are: "read", "write", "admin".
+	Permissions *string    `json:"permissions,omitempty"`
+	CreatedAt   *Timestamp `json:"created_at,omitempty"`
+	URL         *string    `json:"url,omitempty"`
+	HTMLURL     *string    `json:"html_url,omitempty"`
+}
+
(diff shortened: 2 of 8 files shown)


github/repos_collaborators.go
```diff
@@ -58,20 +58,23 @@ type RepositoryAddCollaboratorOptions struct {
...
+// GitHub API docs: https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator
 func (s *RepositoriesService) AddCollaborator(owner, repo, user string, opt *RepositoryAddCollaboratorOptions) (*Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/collaborators/%v", owner, repo, user)
 	req, err := s.client.NewRequest("PUT", u, opt)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeRepositoryInvitationsPreview)
```
shurcooL: It's interesting that this takes an existing, previously stable endpoint, and changes it slightly to use the (potentially less stable) preview API. (Usually preview APIs are for new endpoints.)

I think it's fine per our discussion in #376 because people who have really really specific needs will be vendoring this library anyway.

github/repos_invitations.go
```diff
@@ -0,0 +1,91 @@
...
+// RepositoryInvitation represents an invitation to collaborate on a repo.
+type RepositoryInvitation struct {
+	ID      *int        `json:"id,omitempty"`
+	Repo    *Repository `json:"repository,omitempty"`
+	Invitee *User       `json:"invitee,omitempty"`
+	Inviter *User       `json:"inviter,omitempty"`
+
+	// Permissions represents the permissions that the associated user will have
+	// on the repository. Possible values are: "read", "write", "admin".
+	Permissions string     `json:"permissions,omitempty"`
```
shurcooL: What's the reason `Permissions` is a `string` rather than `*string`, unlike all other fields here?

I'm not opposed, just want to clarify/understand.

github/repos_invitations.go
```diff
@@ -0,0 +1,91 @@
...
+// invitation.
+//
+// permissions represents the permissions that the associated user will have
+// on the repository. Possible values are: "read", "write", "admin".
+//
+// GitHub API docs: https://developer.github.com/v3/repos/invitations/#update-a-repository-invitation
+func (s *RepositoriesService) UpdateInvitation(repoID, invitationID int, permissions string) (*RepositoryInvitation, *Response, error) {
+	opts := &struct {
+		Permissions *string `json:"permissions,omitempty"`
+	}{Permissions: String(permissions)}
```
shurcooL: Minor optional suggestion here.

I feel like this could be simplified a little. The reason pointers are used in receiving structs (i.e., those whose values are populated by unmarshalling the JSON payload from GitHub API response) is to be able to differentiate zero values from missing values (i.e., see #79 and #19).

However, here, you are simply sending a JSON payload to GitHub API. It will always have a `Permission` value (it may or may not be zero, but it will never be missing because the method signature has `permissions string`, not `permissions *string`.

Therefore, you can simplify this ever so slightly to:

```Go
opts := &struct {
	Permissions string `json:"permissions,omitempty"`
}{Permissions: permissions}
```

What do you think? If you agree, you should apply this in another place at https://github.com/google/go-github/blob/3e167707c0e7fec4ff2e1afde8e923dc934d2a70/github/users_gpg_keys.go#L92-L96.

github/users.go
```diff
@@ -166,3 +166,59 @@ func (s *UsersService) ListAll(opt *UserListOptions) ([]User, *Response, error)
 
 	return *users, resp, err
 }
+
+// ListInvitations lists all currently-open repository invitations for the
+// authenticated user.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository
+func (s *UsersService) ListInvitations() ([]RepositoryInvitation, *Response, error) {
```
shurcooL: Given #375, wouldn't you want to return `[]*RepositoryInvitation` here rather than `[]RepositoryInvitation`?

shurcooL:
I've left 4 inline comments, otherwise LGTM.

I did not test the code, just reviewed it.

github/repos_invitations.go
gmlewis: This was a mistake on my part.  Thanks, @shurcooL - I'll fix it.

github/repos_invitations.go
gmlewis: Yes, I agree.  Thanks, @shurcooL.

github/users.go
gmlewis: I definitely do. Thanks for catching that!

github/users.go
```diff
@@ -166,3 +166,59 @@ func (s *UsersService) ListAll(opt *UserListOptions) ([]User, *Response, error)
 
 	return *users, resp, err
 }
+
+// ListInvitations lists all currently-open repository invitations for the
+// authenticated user.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository
```
shurcooL: This link should be https://developer.github.com/v3/repos/invitations/#list-a-users-repository-invitations.

gmlewis: Good catch!  Thanks, @shurcooL

PR #375, opened 2016-06-17
replace []Issue with []*Issue and for other large structs as well

### LINKED ISSUE #180: replace []Issue with []*Issue
In general having slices of large structs is awkward because it means that when you range over the slice, you either have to write:

    for i := range list {
        x := &list[i]
        ...
    }

or you write

    for _, x := range list {
        ...
    }

and put up with the fact that (1) the iteration is doing a (more expensive than pointer copy) struct copy  for each element and (2) you cannot modify the elements by changing x, since x is a copy.

If list were a []*github.Issue it wouldn't have these problems. 

The encoding/json package can handle slices of pointers just fine, but changing the types would have the drawback of breaking existing clients.

gmlewis:
Note that this is an API-breaking change but should have minimal
impact on users of this package due to the nice inference
properties of the Go programming language.

Bumped `libraryVersion` to `2` due to API-breaking change as
discussed in #376.

Fixes #180.

Change-Id: Ib386135e6b8f306d1f54278968c576f3ceccc4e7

### DIFF
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -85,7 +85,7 @@ func (e *Event) Payload() (payload interface{}) {
 // ListEvents drinks from the firehose of all public events across GitHub.
 //
 // GitHub API docs: http://developer.github.com/v3/activity/events/#list-public-events
-func (s *ActivityService) ListEvents(opt *ListOptions) ([]Event, *Response, error) {
+func (s *ActivityService) ListEvents(opt *ListOptions) ([]*Event, *Response, error) {
 	u, err := addOptions("events", opt)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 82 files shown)


shurcooL:
LGTM.

Most clients will not need to change their code at all. Some will need to update it, if they have wrappers or mock implementations of these interfaces. But it's easy to do so, and it makes the API better/resolves an issue.

It is a breaking API change (even if slightly), so I'd suggest waiting to hear @willnorris's thoughts too. But I think it's worth it (at least I'm not opposed).

P.S. You would not be able to do this in C++ as easily, because each `.` in user code would have to become `->`. :)

willnorris:
I think more people are going to have to change their code than you think.  For example, just looking at the Issues.List method, [both of the two people using it](https://sourcegraph.com/github.com/google/go-github/-/info/GoPackage/github.com/google/go-github/github/-/IssuesService/List) would have compilation errors.  Now granted, it's a very easy fix for them, but it's still a breaking change nonetheless.

That said, I'm still fine with doing it.

PR #377, opened 2016-06-19
Remove leading slash from ReactionsService.DeleteReaction URL.

shurcooL:
It's harmless, but inconsistent. All other such URLs in this package do not use a leading slash.

See https://github.com/google/go-github/commit/09a37d57ac99a8938444e44375db8660c2184696#commitcomment-17927607.


gmlewis:
LGTM. 👍

PR #378, opened 2016-06-20
make integration tests pass even if already watching this repo

### LINKED ISSUE #366: TestUsers_* integration tests are failing.
Needs investigation.

gmlewis:
Note that `TestUsers_Keys` still fails for general users
but will need to be addressed later.

Fixes #366.

Change-Id: I35d8a73c2daed08df7d60085b1585f48caee9a95

### DIFF
--- a/tests/integration/activity_test.go
+++ b/tests/integration/activity_test.go
@@ -13,14 +13,19 @@ import (
 	"github.com/google/go-github/github"
 )
 
+const (
+	owner = "google"
+	repo  = "go-github"
+)
+
 func TestActivity_Starring(t *testing.T) {
-	stargazers, _, err := client.Activity.ListStargazers("google", "go-github", nil)
+	stargazers, _, err := client.Activity.ListStargazers(owner, repo, nil)
 	if err != nil {
 		t.Fatalf("Activity.ListStargazers returned error: %v", err)
 	}
 
 	if len(stargazers) == 0 {
-		t.Errorf("Activity.ListStargazers('google', 'go-github') returned no stargazers")
+		t.Errorf("Activity.ListStargazers(%q, %q) returned no stargazers", owner, repo)
 	}
 
 	// the rest of the tests requires auth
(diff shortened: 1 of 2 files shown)


tests/integration/activity_test.go
```diff
@@ -29,7 +34,7 @@ func TestActivity_Starring(t *testing.T) {
 	}
 
 	// first, check if already starred google/go-github
```
shurcooL: Now that owner and repo are factored out into constants outside, it makes sense to update the comment to not mention google/go-github repo specifically, but rather use more broad terms. For example:

```
// first, check if already starred the target repository
```

tests/integration/activity_test.go
```diff
@@ -68,58 +73,70 @@ func TestActivity_Starring(t *testing.T) {
...
-	// first, check if already watching google/go-github
-	sub, _, err := client.Activity.GetRepositorySubscription("google", "go-github")
+	// check again and verify not watching
+	sub, _, err := client.Activity.GetRepositorySubscription(owner, repo)
 	if err != nil {
 		t.Fatalf("Activity.GetRepositorySubscription returned error: %v", err)
 	}
 	if sub != nil {
-		t.Fatalf("Already watching google/go-github.  Please manually stop watching it first.")
+		t.Fatalf("Still watching google/go-github after deleting subscription.")
```
shurcooL: Also here.

tests/integration/activity_test.go
```diff
@@ -68,58 +73,70 @@ func TestActivity_Starring(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Activity.GetRepositorySubscription returned error: %v", err)
 	}
+
+	// If already subscribing, delete then recreate subscription.
 	if sub != nil {
-		t.Fatalf("Still watching google/go-github after deleting subscription.")
+		deleteSubscription(t)
+		createSubscription(t)
+		return
```
shurcooL: Really minor style thought for your consideration.

When it comes to error handling, I really like to [indent error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow). That way, the "normal" path is at normal indentation, and the more unusual error handling is indented.

Here, I think using this "early return" style is not as nice, because the two cases are equal in my books, neither is an error, and neither is less likely than the other. They're just 2 branching paths, either you're already watching or not.

So I think it's slightly nicer to express that the two paths are essentially equal by not using early return, but instead... Personally, I'd use my favorite `switch`:

```Go
switch {
case sub != nil: // If already subscribing, delete then recreate subscription.
	deleteSubscription(t)
	createSubscription(t)
case sub == nil: // Otherwise, create subscription and then delete it.
	createSubscription(t)
	deleteSubscription(t)
}
```

But if you're not as big a fan of `switch` as me and [Andrew](https://twitter.com/enneff/status/575126465236332545), you could just `if ... else ...`.

Current code is absolutely fine, I'm just being nitpicky and looking for ways to make it even better.

shurcooL:
I think it'd be nice to update hardcoded references to google/go-github in comments and error messages, since you've moved those into package level constants. See inline comments.

Aside from that, LGTM.

PR #370, opened 2016-06-16
Update event_types.go

simpleelegant:
In my real project, I found this error.

PR #379, opened 2016-06-21
better support both Events API and Webhook events in PushEventCommit

gmlewis:
Fixes #370.

Change-Id: I0a8567ae352bd2e9e973ac73383894d2b4cdfb9d

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -316,12 +316,19 @@ func (p PushEvent) String() string {
 
 // PushEventCommit represents a git commit in a GitHub PushEvent.
 type PushEventCommit struct {
-	SHA       *string       `json:"sha,omitempty"`
-	Message   *string       `json:"message,omitempty"`
-	Author    *CommitAuthor `json:"author,omitempty"`
+	Message  *string       `json:"message,omitempty"`
+	Author   *CommitAuthor `json:"author,omitempty"`
+	URL      *string       `json:"url,omitempty"`
+	Distinct *bool         `json:"distinct,omitempty"`
+
+	// The following fields are only populated by Events API.
+	SHA *string `json:"sha,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	ID        *string       `json:"id,omitempty"`
+	TreeID    *string       `json:"tree_id,omitempty"`
+	Timestamp *Timestamp    `json:"timestamp,omitempty"`
 	Committer *CommitAuthor `json:"committer,omitempty"`
-	URL       *string       `json:"url,omitempty"`
-	Distinct  *bool         `json:"distinct,omitempty"`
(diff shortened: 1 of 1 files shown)


shurcooL:
This looks good as far as adding the missing fields and resolving #370.

But I want to ask, Glenn, have you vetted the rest of the common fields? There are some of them that I think are Webhook-only as well, because they're not mentioned in [Events API payload](https://developer.github.com/v3/activity/events/types/#events-api-payload-19), but are included in the [Webhook payload example](https://developer.github.com/v3/activity/events/types/#webhook-payload-example-15).

As far as I can see, these common `PushEventCommit ` fields are Webhook-only as well:

- `Committer`
- `Added`
- `Removed`
- `Modified`

If you prefer to consider that out of scope for this PR, I have no objections to that. This PR indeed resolves the task in #370. LGTM.

gmlewis:
@shurcooL - as far as I can tell, the [Events API payload](https://developer.github.com/v3/activity/events/types/#events-api-payload-19) appears exactly as described, so you are correct that those fields are Webhook-only as well.

We should go ahead and clean those up too in this PR while we are discussing it.

It turns out that I had also missed `timestamp`, so I'm adding it.
PTAL.

github/event_types.go
```diff
@@ -316,15 +316,22 @@ func (p PushEvent) String() string {
...
+	// The following fields are only populated by Webhook events.
+	ID        *string       `json:"id,omitempty"`
+	TreeID    *string       `json:"tree_id,omitempty"`
 	Committer *CommitAuthor `json:"committer,omitempty"`
-	URL       *string       `json:"url,omitempty"`
-	Distinct  *bool         `json:"distinct,omitempty"`
 	Added     []string      `json:"added,omitempty"`
 	Removed   []string      `json:"removed,omitempty"`
 	Modified  []string      `json:"modified,omitempty"`
+	Timestamp *Timestamp    `json:"timestamp,omitempty"`
```
shurcooL: Sometimes my comments get marginally trivial (but it's because everything else is great), but I think it might be nicer to arrange the fields in a slightly more logical order. I think `Timestamp` might be a better fit just before or after `Committer`, what do you think?

shurcooL:
> We should go ahead and clean those up too in this PR while we are discussing it.

Great, thanks!

Nice catch on `timestamp`, I missed that one.

LGTM. I left one trivial optional comment.

github/event_types.go
gmlewis: @shurcooL - It seems to me that your review comments have always led to improvements in the code and they are very welcome and greatly appreciated. Thank you!

Yes, I agree.  I'll move it before the `Committer` so that it is with the more mundane fields like `ID` and `TreeID`.

PR #383, opened 2016-06-24
add GitHub URL parsing utility function

### LINKED ISSUE #380: add GitHub URL parsing utility function
It seems that clients of this package frequently need to parse GitHub URLs in order to validate and/or extract useful information from them.  It would be nice to have all this code deduplicated into a single, central location, and this package might be a nice home for it.

Maybe something like:

```go
func ParseURL(url string) (owner, repo, ref string, err error) {...}
```

Common URLs to parse could be ones similar to:
```
https://github.com/bazelbuild/bazel/tree/master
http://github.com/bazelbuild/bazel/tree/master
https://www.github.com/bazelbuild/bazel/tree/master
https://github.com/bazelbuild/bazel/releases/tag/0.2.1
https://github.com/bazelbuild/bazel/commit/19b5675725caf69008a717082149237400260edc
https://github.com/bazelbuild/bazel/archive/0.2.1.zip
```

Thoughts?

gmlewis:
Fixes #380.

Change-Id: Ibe2f97028835d84c87e047baa8724edb4ae05ca0

### DIFF
--- a/github/misc.go
+++ b/github/misc.go
@@ -9,6 +9,8 @@ import (
 	"bytes"
 	"fmt"
 	"net/url"
+	"path"
+	"strings"
 )
 
 // MarkdownOptions specifies optional parameters to the Markdown method.
@@ -194,4 +196,34 @@ func (c *Client) ListServiceHooks() ([]*ServiceHook, *Response, error) {
 	}
 
 	return *hooks, resp, err
+}
+
+// ParseURL parses a GitHub URL and returns the owner and repo (if provided).
+// It returns an error if it is unable to parse the URL.
+//
+// Examples of valid URLs are:
+//     https://github.com/bazelbuild/bazel
+//     http://github.com/bazelbuild/bazel/tree/master
+//     https://www.github.com/bazelbuild/bazel/tree/master
+//     https://github.com/bazelbuild/bazel/releases/tag/0.2.1
+//     https://github.com/bazelbuild/bazel/commit/19b5675725caf69008a717082149237400260edc
+//     https://github.com/bazelbuild/bazel/archive/0.2.1.zip
+func ParseURL(rawurl string) (owner, repo string, err error) {
+	u, err := url.Parse(rawurl)
+	if err != nil {
+		return "", "", fmt.Errorf("unrecognized GitHub URL %q: %v", rawurl, err)
+	}
+	if u.Host != "github.com" && u.Host != "www.github.com" {
+		return "", "", fmt.Errorf("unrecognized GitHub URL: %q", rawurl)
+	}
+
+	parts := strings.SplitN(path.Clean(u.Path), "/", 4)
+	switch {
+	case len(parts) < 2 || parts[1] == "":
+		return "", "", fmt.Errorf("unrecognized GitHub URL: %q", rawurl)
+	case len(parts) == 2:
+		return parts[1], "", nil
+	default:
+		return parts[1], parts[2], nil
+	}
 }
--- a/github/misc_test.go
+++ b/github/misc_test.go
@@ -168,3 +168,49 @@ func TestRepositoriesService_ListServiceHooks(t *testing.T) {
 		t.Errorf("Repositories.ListServiceHooks returned %+v, want %+v", hooks, want)
 	}
 }
+
+func TestParseURL(t *testing.T) {
+	const (
+		owner, repo = "bazelbuild", "bazel"
+	)
+	tests := []struct {
+		url                 string
+		wantOwner, wantRepo string
+		wantErr             bool
+	}{
+		{url: "https://github.com/bazelbuild/bazel", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild", wantOwner: owner},
+		{url: "https://github.com/bazelbuild/", wantOwner: owner},
+		{url: "https://github.com////bazelbuild////bazel", wantOwner: owner, wantRepo: repo},
+		{url: "http://github.com/bazelbuild/bazel/tree/master", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/archive/0.2.1.zip", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/commit/19b...edc", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/releases/tag/0.2.1", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/tree/master", wantOwner: owner, wantRepo: repo},
+		{url: "https://www.github.com/bazelbuild/bazel/tree/master", wantOwner: owner, wantRepo: repo},
+		{url: "https://mygithub.com/bazelbuild", wantErr: true},
+		{url: "https://github.com///", wantErr: true},
+		{url: "https://github.com//", wantErr: true},
+		{url: "https://github.com/", wantErr: true},
+		{url: "https://github.com", wantErr: true},
+	}
+
+	for _, test := range tests {
+		gotOwner, gotRepo, err := ParseURL(test.url)
+		if err != nil {
+			if !test.wantErr {
+				t.Errorf("ParseURL(%q) = %v, want nil", test.url, err)
+			}
+			continue
+		}
+		if test.wantErr {
+			t.Errorf("ParseURL(%q) = nil, want error", test.url)
+			continue
+		}
+
+		if gotOwner != test.wantOwner || gotRepo != test.wantRepo {
+			t.Errorf("ParseURL(%q) = (%q, %q), want (%q, %q)", test.url, gotOwner, gotRepo, test.wantOwner, test.wantRepo)
+		}
+	}
+}


github/misc.go
```diff
@@ -195,3 +197,34 @@ func (c *Client) ListServiceHooks() ([]*ServiceHook, *Response, error) {
 
 	return *hooks, resp, err
 }
+
+var (
+	githubRE = regexp.MustCompile(`https?://(www\.)?github\.com/`)
```
shurcooL: Can we do this using `url.Parse` instead of a regexp?

gmlewis: Sure.

github/misc_test.go
```diff
@@ -168,3 +168,39 @@ func TestRepositoriesService_ListServiceHooks(t *testing.T) {
...
+		{url: "https://github.com/bazelbuild/", wantOwner: owner},
+		{url: "https://github.com/bazelbuild", wantOwner: owner},
+		{url: "https://mygithub.com/bazelbuild", wantErr: true},
+		{url: "https://github.com/", wantErr: true},
+		{url: "https://github.com", wantErr: true},
+	}
+
+	for _, test := range tests {
+		gotOwner, gotRepo, err := ParseURL(test.url)
+		if err != nil && !test.wantErr {
```
shurcooL: The `wantErr` check here is ineffective. It verifies that `err` must be `nil` when `wantErr` is false. However, when `wantErr` is true, either `nil` or not `nil` value of `err` will pass without test failing.

As a result, it didn't catch that there's an inconsistency in the `{url: "https://github.com/", wantErr: true},` line. `ParseURL("https://github.com/")`, as is currently written, actually returns `nil` error.

shurcooL: There are many ways to fix it, pick your favorite. Some examples:

- https://github.com/shurcooL/gostatus/blob/f8723cde71fe2c54eb101408c/status/url_test.go#L60-L66
- https://github.com/sourcegraph/sourcegraph/blob/47f606ac15d64/pkg/routevar/user_test.go#L135-L145
- https://github.com/sourcegraph/sourcegraph/blob/47f606ac15d64/app/internal/returnto/returnto_test.go#L32-L39

github/misc_test.go
```diff
@@ -168,3 +168,39 @@ func TestRepositoriesService_ListServiceHooks(t *testing.T) {
...
+		{url: "https://www.github.com/bazelbuild/bazel/tree/master", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/releases/tag/0.2.1", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/commit/19b...edc", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/archive/0.2.1.zip", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/", wantOwner: owner},
+		{url: "https://github.com/bazelbuild", wantOwner: owner},
+		{url: "https://mygithub.com/bazelbuild", wantErr: true},
+		{url: "https://github.com/", wantErr: true},
```
shurcooL: This case returns nil error. See my previous comment.

github/misc.go
shurcooL: Thank you so much! The new version that uses `url.Parse` instead of regexp is so so so much nicer IMO! It's easier to read, understand, and verify. And it's probably faster. (I am really not a fan of regexpes. Especially in places where more standard tools can be used to parse a well known spec.)

github/misc_test.go
gmlewis: Good catch, @shurcooL!  Thanks.
PTAL.

github/misc_test.go
```diff
@@ -168,3 +168,48 @@ func TestRepositoriesService_ListServiceHooks(t *testing.T) {
...
+		{url: "https://www.github.com/bazelbuild/bazel/tree/master", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/releases/tag/0.2.1", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/commit/19b...edc", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/archive/0.2.1.zip", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel/", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/bazel", wantOwner: owner, wantRepo: repo},
+		{url: "https://github.com/bazelbuild/", wantOwner: owner},
+		{url: "https://github.com/bazelbuild", wantOwner: owner},
+		{url: "https://mygithub.com/bazelbuild", wantErr: true},
+		{url: "https://github.com///", wantErr: true},
```
shurcooL: I don't know if you want to handle this case, but the following URL actually works:

https://github.com/////google/////go-github

If you do want to handle it, it can probably be done by doing `parts := strings.SplitN(path.Clean(u.Path), "/", 4)`.

github/misc.go
```diff
@@ -195,3 +196,33 @@ func (c *Client) ListServiceHooks() ([]*ServiceHook, *Response, error) {
...
+// ParseURL parses a GitHub URL and returns the owner and repo (if provided).
+// It returns an error if it is unable to parse the URL.
+//
+// Examples of valid URLs are:
+//     https://github.com/bazelbuild/bazel/tree/master
+//     http://github.com/bazelbuild/bazel/tree/master
+//     https://www.github.com/bazelbuild/bazel/tree/master
+//     https://github.com/bazelbuild/bazel/releases/tag/0.2.1
+//     https://github.com/bazelbuild/bazel/commit/19b5675725caf69008a717082149237400260edc
+//     https://github.com/bazelbuild/bazel/archive/0.2.1.zip
```
shurcooL: It sounded like you had some specific use cases in #380 that warranted adding this helper. I don't have have a current need for this, so it's hard for me to confirm it's usefulness. However, if I did want to use this helper, the number one type of URL I'd like to use it on would be of the form:

    https://github.com/bazelbuild/bazel

Just a plain reference to a repository, and I'd want `bazelbuild`, `bazel`, `nil` as the return values. From reviewing the source code, I know that URL type would work. However, if I were a new user of this library, I think I'd be very confused by these examples, since they do not list a plain `https://github.com/bazelbuild/bazel` URL, yet the return values only include `owner` and `repo`.

So, my suggestion would be, at the very least add `https://github.com/bazelbuild/bazel` as the first example of valid URL type.

shurcooL:
Another general observation, IMO, the usefulness of a utility is generally inversely proportional to how specific it is. Supporting more varied URL types might makes the function less helpful, rather than more helpful. But of course it'll depend on the exact project needs, which is why utilities are often best placed as unexported helpers right where they're needed, so that they serve a very concrete and specific need. Unless there's a very common general need shared across many projects.

As one data point, I know we couldn't use it to replace [`githubutil.SplitRepoURI`](https://sourcegraph.com/sourcegraph/sourcegraph/-/def/GoPackage/sourcegraph.com/sourcegraph/sourcegraph/pkg/githubutil/-/SplitRepoURI) because we need that helper to reject URLs that are not strictly `github.com/owner/repo` (plus it has custom logic for replacing the host, etc.).

Another data point, I couldn't use it to replace [`githubapi.ghRepoSpec`](https://github.com/shurcooL/notifications/blob/0d8dc47e48827a488c0a91ef829136d64fd5b60e/githubapi/githubapi.go#L251-L260) either, because its input already has host stripped, and it must reject all suffixes as well. For these [other helpers](https://github.com/shurcooL/notifications/blob/0d8dc47e48827a488c0a91ef829136d64fd5b60e/githubapi/githubapi.go#L197-L244), I need them to parse very specific GitHub URL types and return additional information (not just owner and repo).

Since I don't have have a current need for this utility, it's hard for me to confirm it's usefulness (I trust you and @willnorris on that). Since @willnorris said:

> Just to repeat what we talked about, we have at least four places (that I know of) inside Google where we are parsing URLs to extract one or more of these values.

I think it'd be a good idea to evaluate and confirm whether this utility would actually be helpful in replacing those. Making the API slightly larger has a cost (quite small in this case), so it should be done if it's worth it.

I don't see any other issues, so LGTM.

github/misc_test.go
gmlewis: Wow.  Learn something new every day.  Thanks, @shurcooL - I'll clean the path.

gmlewis:
OK, if I understand you correctly, @shurcooL, you feel that returning just `(owner, repo, err)` here has limited usefulness, and in fact you would not be able to use this helpful function at all in the two places where it could possibly have been useful.

Can you think of an API change that would make this useful to you and also to our use case where we don't want there to be four different instances of URL parsers that minimally need the `owner` and `repo` extracted from a verified GitHub URL?

We've already ruled out returning everything after the `repo`, which I thought might be useful, so I'm not sure what is left.

github/misc.go
gmlewis: It turns out that the examples you listed were already in the tests... but I'll move them to the top so that they are easier to see (and I'll embed the wacky `////` version in the middle).

shurcooL:
> OK, if I understand you correctly, @shurcooL, you feel that returning just `(owner, repo, err)` here has limited usefulness, and in fact you would not be able to use this helpful function at all in the two places where it could possibly have been useful.

That's right.

> Can you think of an API change that would make this useful to you and also to our use case where we don't want there to be four different instances of URL parsers that minimally need the owner and repo extracted from a verified GitHub URL?

I don't think I can.

To make it more useful to me, the function would have to be more specific. But as it becomes more specific, there are too many different variations of the same thing. For example:

- ParseURL that accepts URLs of only form "github.com/owner/repo", where there's no scheme, and host must match "github.com" exactly, and there must not be anything after "repo".
- ParseURL that accepts URLs of only form "github.com/owner/repo", where there's no scheme, and host can match "github.com" or "api.github.com" exactly, and there must not be anything after "repo".
- ParseURL that accepts URLs of only form "github.com/owner/repo", where there's no scheme, and host can match "github.com" or "api.github.com" exactly, and there may be anything after "repo" but it gets ignored.
- ParseURL that accepts URLs of only form "https://github.com/owner/repo", where the scheme must be "https", and host must match "github.com" exactly, and there may not be anything after "repo".
- ParseURL that accepts URLs of only form "owner/repo".

In all those cases, I would just write the helper myself in the project that needs the specific functionality. It's easier to do that, than to try to find a utility that does exactly what the project requires. I can't think of one general API that would be generally useful.

That said, if you think it can be useful to your projects and other people's projects, I don't think you shouldn't merge it because I can't find it useful, I'm only one data point.

gmlewis:
Will and I have decided to not go through with this. Thank you for your thoughtful comments and time invested, @shurcooL.

PR #384, opened 2016-06-25
Add support for affiliation and visibility in RepositoryListOptions.

### LINKED ISSUE #231: Add Support for "Affiliation" in RepositoryListOptions
In the [Github API Docs](https://developer.github.com/v3/repos/#list-your-repositories) there is a field for `affiliation` when doing a list of all your repositories. As far as I can tell, this doesn't exist in the `RepositoryListOptions` struct though. 

Am I missing something?

wlynch:
Also updates comments in RepositoryListOptions to match the Github API docs
(https://developer.github.com/v3/repos/#list-your-repositories).

Fixes #231.

### DIFF
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -42,15 +42,49 @@ func TestRepositoriesService_List_specifiedUser(t *testing.T) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", mediaTypeLicensesPreview)
 		testFormValues(t, r, values{
-			"type":      "owner",
-			"sort":      "created",
-			"direction": "asc",
-			"page":      "2",
+			"visibility":  "public",
+			"affiliation": "owner,collaborator",
+			"sort":        "created",
+			"direction":   "asc",
+			"page":        "2",
 		})
 		fmt.Fprint(w, `[{"id":1}]`)
 	})
 
-	opt := &RepositoryListOptions{"owner", "created", "asc", ListOptions{Page: 2}}
+	opt := &RepositoryListOptions{
+		Visibility:  "public",
+		Affiliation: "owner,collaborator",
+		Sort:        "created",
+		Direction:   "asc",
+		ListOptions: ListOptions{Page: 2},
+	}
+	repos, _, err := client.Repositories.List("u", opt)
+	if err != nil {
+		t.Errorf("Repositories.List returned error: %v", err)
+	}
+
+	want := []*Repository{{ID: Int(1)}}
+	if !reflect.DeepEqual(repos, want) {
+		t.Errorf("Repositories.List returned %+v, want %+v", repos, want)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


github/repos_test.go
```diff
@@ -42,15 +42,17 @@ func TestRepositoriesService_List_specifiedUser(t *testing.T) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", mediaTypeLicensesPreview)
 		testFormValues(t, r, values{
-			"type":      "owner",
-			"sort":      "created",
-			"direction": "asc",
-			"page":      "2",
+			"visibility":  "public",
+			"affiliation": "owner,collaborator",
+			"type":        "owner",
```
shurcooL: Type is documented as:

> Will cause a 422 error if used in the same request as visibility or affiliation.

So this is not a very realistic scenario. I know it's mock values and will not actually cause 422, but it's probably still better to simulate a real request.

github/repos_test.go
```diff
@@ -42,15 +42,17 @@ func TestRepositoriesService_List_specifiedUser(t *testing.T) {
...
+			"sort":        "created",
+			"direction":   "asc",
+			"page":        "2",
 		})
 		fmt.Fprint(w, `[{"id":1}]`)
 	})
 
-	opt := &RepositoryListOptions{"owner", "created", "asc", ListOptions{Page: 2}}
+	opt := &RepositoryListOptions{"public", "owner,collaborator", "owner", "created", "asc", ListOptions{Page: 2}}
```
shurcooL: With this many fields, I think it's time to start using keyed composite literals.

github/repos_test.go
wlynch: Done. Broke this out into 2 different tests so that both RepositoryListOptions are valid.

github/repos_test.go
wlynch: Done.

shurcooL:
LGTM. Thank you!

willnorris:
Thanks, Billy.

anshumanbh:
@wlynch @shurcooL @willnorris This PR does not seem to list the private repositories for a user when the `Visibility` is set to `private`. Can you please confirm?

shurcooL:
It might be because your oauth token doesn't have private repos scope. You should check whether or not the issue is in the `go-github` library, or on the GitHub API side by making the same request manually, without `go-github`, and see if the response is as expected. Then we can figure out what `go-github` is doing wrong, if anything.

anshumanbh:
@shurcooL my oauth token had all the scopes. I figured out the issue. I think its just how the Github API works. If you provide it a valid oauth token, hitting the /users/<user>/repos will only give you public repositories. If we want to list all repositories of a user, it is a different endpoint `/user/repos` all together. Thanks for looking into this!

mangatmodi:
@anshumanbh This was exactly my issue. Thanks

PR #385, opened 2016-06-26
Support list gist commits and list gist forks endpoints

beshrkayali:
Added support for [list gist commits](https://developer.github.com/v3/gists/#list-gist-commits) and [list gist forks](https://developer.github.com/v3/gists/#list-gist-forks) endpoints, including tests.

I was looking for a Go project to contribute to (as I'm still not very comfortable with Go) and I found that these two end points are missing (and no one has started working on them yet). Please let me know if I missed something in the guidelines :-)

### DIFF
--- a/github/gists.go
+++ b/github/gists.go
@@ -64,6 +64,7 @@ func (gc GistCommit) String() string {
 	return Stringify(gc)
 }
 
+// GistFork represents a fork of a gist
 type GistFork struct {
 	User      *User      `json:"user, omitempty"`
 	ID        *string    `json:"id,omitempty"`
@@ -234,27 +235,23 @@ func (s *GistsService) Edit(id string, gist *Gist) (*Gist, *Response, error) {
 	return g, resp, err
 }
 
-// List gist commits.
+// ListCommits lists commits of a gist.
 //
 // Github API docs: https://developer.github.com/v3/gists/#list-gist-commits
-
 func (s *GistsService) ListCommits(id string) ([]*GistCommit, *Response, error) {
 	u := fmt.Sprintf("gists/%v/commits", id)
 	req, err := s.client.NewRequest("GET", u, nil)
-
 	if err != nil {
 		return nil, nil, err
 	}
 
-	gist_commits := new([]*GistCommit)
-
-	resp, err := s.client.Do(req, gist_commits)
-
+	gistcommits := new([]*GistCommit)
+	resp, err := s.client.Do(req, gistcommits)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return *gist_commits, resp, err
+	return *gistcommits, resp, err
 }
 
 // Delete a gist.
@@ -326,25 +323,21 @@ func (s *GistsService) Fork(id string) (*Gist, *Response, error) {
 	return g, resp, err
 }
 
-// List gist forks.
+// ListForks lists forks ofa gist.
 //
 // Github API docs: https://developer.github.com/v3/gists/#list-gist-forks
-
 func (s *GistsService) ListForks(id string) ([]*GistFork, *Response, error) {
 	u := fmt.Sprintf("gists/%v/forks", id)
 	req, err := s.client.NewRequest("GET", u, nil)
-
 	if err != nil {
 		return nil, nil, err
 	}
 
-	gist_forks := new([]*GistFork)
-
-	resp, err := s.client.Do(req, gist_forks)
-
+	gistforks := new([]*GistFork)
+	resp, err := s.client.Do(req, gistforks)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return *gist_forks, resp, err
+	return *gistforks, resp, err
 }
--- a/github/gists.go
+++ b/github/gists.go
@@ -64,9 +64,9 @@ func (gc GistCommit) String() string {
 	return Stringify(gc)
 }
 
-// GistFork represents a fork of a gist
+// GistFork represents a fork of a gist.
 type GistFork struct {
-	User      *User      `json:"user, omitempty"`
+	User      *User      `json:"user,omitempty"`
 	ID        *string    `json:"id,omitempty"`
 	CreatedAt *Timestamp `json:"created_at,omitempty"`
 	UpdatedAt *Timestamp `json:"updated_at,omitempty"`
@@ -245,13 +245,13 @@ func (s *GistsService) ListCommits(id string) ([]*GistCommit, *Response, error) 
 		return nil, nil, err
 	}
 
-	gistcommits := new([]*GistCommit)
-	resp, err := s.client.Do(req, gistcommits)
+	gistCommits := new([]*GistCommit)
+	resp, err := s.client.Do(req, gistCommits)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return *gistcommits, resp, err
+	return *gistCommits, resp, err
 }
 
 // Delete a gist.
@@ -323,7 +323,7 @@ func (s *GistsService) Fork(id string) (*Gist, *Response, error) {
 	return g, resp, err
 }
 
-// ListForks lists forks ofa gist.
+// ListForks lists forks of a gist.
 //
 // Github API docs: https://developer.github.com/v3/gists/#list-gist-forks
 func (s *GistsService) ListForks(id string) ([]*GistFork, *Response, error) {
@@ -333,11 +333,11 @@ func (s *GistsService) ListForks(id string) ([]*GistFork, *Response, error) {
 		return nil, nil, err
 	}
 
-	gistforks := new([]*GistFork)
-	resp, err := s.client.Do(req, gistforks)
+	gistForks := new([]*GistFork)
+	resp, err := s.client.Do(req, gistForks)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return *gistforks, resp, err
+	return *gistForks, resp, err
 }
--- a/github/gists.go
+++ b/github/gists.go
@@ -54,10 +54,11 @@ func (g GistFile) String() string {
 
 // GistCommit represents a commit on a gist.
 type GistCommit struct {
-	Version      *string        `json:"version,omitempty"`
-	User         *User          `json:"user,omitempty"`
-	ChangeStatus map[string]int `json:"change_status,omitempty"`
-	CommitedAt   *Timestamp     `json:"commited_at,omitempty"`
+	URL          *string      `json:"url,omitempty"`
+	Version      *string      `json:"version,omitempty"`
+	User         *User        `json:"user,omitempty"`
+	ChangeStatus *CommitStats `json:"change_status,omitempty"`
+	CommitedAt   *Timestamp   `json:"commited_at,omitempty"`
 }
 
 func (gc GistCommit) String() string {
@@ -66,6 +67,7 @@ func (gc GistCommit) String() string {
 
 // GistFork represents a fork of a gist.
 type GistFork struct {
+	URL       *string    `json:"url,omitempty"`
 	User      *User      `json:"user,omitempty"`
 	ID        *string    `json:"id,omitempty"`
 	CreatedAt *Timestamp `json:"created_at,omitempty"`
--- a/github/gists.go
+++ b/github/gists.go
@@ -52,6 +52,29 @@ func (g GistFile) String() string {
 	return Stringify(g)
 }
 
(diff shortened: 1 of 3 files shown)


github/gists.go
```diff
@@ -211,6 +234,29 @@ func (s *GistsService) Edit(id string, gist *Gist) (*Gist, *Response, error) {
 	return g, resp, err
 }
 
+// List gist commits.
+//
+// Github API docs: https://developer.github.com/v3/gists/#list-gist-commits
+
```
shurcooL: In order for this comment to be considered as documentation by `godoc`, there can't be a blank line between it and the function. You should delete the blank line.

Source: https://golang.org/doc/effective_go.html#commentary

> Comments that appear before top-level declarations, with no intervening newlines, are extracted along with the declaration to serve as explanatory text for the item.

github/gists.go
```diff
@@ -211,6 +234,29 @@ func (s *GistsService) Edit(id string, gist *Gist) (*Gist, *Response, error) {
...
+func (s *GistsService) ListCommits(id string) ([]*GistCommit, *Response, error) {
+	u := fmt.Sprintf("gists/%v/commits", id)
+	req, err := s.client.NewRequest("GET", u, nil)
+
+	if err != nil {
```
shurcooL: It's better style to handle the error value immediately after `err` is assigned, and not have a blank line in between.

Blank lines in Go are used sparingly, and they are great for separating completely separate blocks of code. This is exactly the opposite of what you want when it comes to checking the value of `err`, since at first glance this may read as you didn't check value of `err` at all.

You should get rid of this blank line.

Look at some of the other methods in this file as an example for spacing. It's better to stay consistent.

github/gists.go
```diff
@@ -52,6 +52,29 @@ func (g GistFile) String() string {
...
+	User         *User          `json:"user,omitempty"`
+	ChangeStatus map[string]int `json:"change_status,omitempty"`
+	CommitedAt   *Timestamp     `json:"commited_at,omitempty"`
+}
+
+func (gc GistCommit) String() string {
+	return Stringify(gc)
+}
+
+type GistFork struct {
```
shurcooL: Since this is an exported symbol, it needs to be documented.

Source: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

> All top-level, exported names should have doc comments

It's also something go lint would catch:

```Go
github/gists.go:67: exported type GistFork should have comment or be unexported
```

github/gists_test.go
```diff
@@ -281,6 +281,48 @@ func TestGistsService_Edit_invalidID(t *testing.T) {
...
+			[
+				{"version": "1",
+				 "user": {"id": 1},
+				 "change_status": {
+				 	"deletions": 0,
+				 	"additions": 180,
+				 	"total": 180
+				 },
+				 "commited_at": "2010-01-01T00:00:00Z"
+				}
```
shurcooL: The formatting of this JSON looks kinda off. Can you make it look more consistent with what's in this file, or https://github.com/google/go-github/blob/45cbf2313cdd7e84ba9ef011885adbd72d6e6604/github/git_refs_test.go#L65-L84?

shurcooL:
Hey @beshrkayali, thanks a lot for your contribution, it is really appreciated!

This PR is off to a great start! I've left a few comments for style issues I've spotted, most of them are really minor and easy to fix. I've linked to sources so you can see where these style guidelines are coming from.

Thanks again, and I hope you enjoy learning more about Go and using it/making contributions! 💯

beshrkayali:
@shurcooL Thank for your feedback Dmitri! I've committed the necessary fixes.

github/gists.go
```diff
@@ -52,6 +52,30 @@ func (g GistFile) String() string {
...
+	CommitedAt   *Timestamp     `json:"commited_at,omitempty"`
+}
+
+func (gc GistCommit) String() string {
+	return Stringify(gc)
+}
+
+// GistFork represents a fork of a gist
+type GistFork struct {
+	User      *User      `json:"user, omitempty"`
```
gmlewis: nit: the extra space after the comma could be removed to match the other ones

github/gists.go
```diff
@@ -279,3 +322,22 @@ func (s *GistsService) Fork(id string) (*Gist, *Response, error) {
 
 	return g, resp, err
 }
+
+// ListForks lists forks ofa gist.
```
gmlewis: s/ofa/of a/

github/gists.go
```diff
@@ -211,6 +235,25 @@ func (s *GistsService) Edit(id string, gist *Gist) (*Gist, *Response, error) {
...
+//
+// Github API docs: https://developer.github.com/v3/gists/#list-gist-commits
+func (s *GistsService) ListCommits(id string) ([]*GistCommit, *Response, error) {
+	u := fmt.Sprintf("gists/%v/commits", id)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	gistcommits := new([]*GistCommit)
```
gmlewis: nit: I would capitalize Commits and Forks here and below: `gistCommits` and `gistForks`. Also in the unit tests.

gmlewis:
A few minor comments, otherwise LGTM 👍

beshrkayali:
Thanks for the feedback @gmlewis, issues resolved.

gmlewis:
Thanks, @beshrkayali!
Awaiting one more LGTM and then I will merge your changes.

shurcooL:
> Edit: One tiny issue to note is that the last few endpoints are actually implemented but not reflected in the google doc as `Done`.

What google doc? I'm not familiar with one.

github/gists.go
```diff
@@ -64,6 +64,7 @@ func (gc GistCommit) String() string {
 	return Stringify(gc)
 }
 
+// GistFork represents a fork of a gist
```
shurcooL: Nitpick. This sentence does not end with a period. Other comments have sentences that do end with periods, so it'd be nice and more consistent to include a period at the end of this sentence too.

github/gists.go
shurcooL: For reference, Go style suggests to use always `mixedCaps` style. You can read about it here:

Source: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

github/gists.go
```diff
@@ -52,6 +52,30 @@ func (g GistFile) String() string {
...
+	User         *User          `json:"user,omitempty"`
+	ChangeStatus map[string]int `json:"change_status,omitempty"`
+	CommitedAt   *Timestamp     `json:"commited_at,omitempty"`
+}
+
+func (gc GistCommit) String() string {
+	return Stringify(gc)
+}
+
+// GistFork represents a fork of a gist
```
shurcooL: Nitpick. This sentence does not end with a period. Other comments have sentences that do end with periods, so it'd be nice and more consistent to include a period at the end of this sentence too.

beshrkayali:
@shurcooL ah, it's a [spreadsheet](https://docs.google.com/spreadsheets/d/1wlHnoJSAN01nXUUF1JobsaKO4mJjAwZNWK8L_4ZSlls/edit) actually not a doc. It's highlighted in the [roadmap](https://github.com/google/go-github/blob/master/README.md#roadmap) section of the README.

willnorris:
> @shurcooL ah, it's a [spreadsheet](https://docs.google.com/spreadsheets/d/1wlHnoJSAN01nXUUF1JobsaKO4mJjAwZNWK8L_4ZSlls/edit) actually not a doc. It's highlighted in the [roadmap](https://github.com/google/go-github/blob/master/README.md#roadmap) section of the README.

just fyi, edit access to that doc should be open to anyone in the [go-github Google group](https://groups.google.com/group/go-github).

github/gists.go
```diff
@@ -52,6 +52,30 @@ func (g GistFile) String() string {
 	return Stringify(g)
 }
 
+// GistCommit represents a commit on a gist.
+type GistCommit struct {
+	Version      *string        `json:"version,omitempty"`
+	User         *User          `json:"user,omitempty"`
+	ChangeStatus map[string]int `json:"change_status,omitempty"`
```
shurcooL: `map[string]int` is great for dynamic content, where you don't know in advance what fields may or may not be present.

Based on my reading of the Gists API, it seems like change status is always three fields:

- "deletions"
- "additions"
- "total"

Is that not the case?

If so, it's much better to provide a static API for it. So, this line:

```Go
ChangeStatus map[string]int `json:"change_status,omitempty"`
```

Can become:

```Go
ChangeStatus *ChangeStatus `json:"change_status,omitempty"`
```

And you'd add that new struct:

```Go
// ChangeStatus represents the change status of a gist commit.
type ChangeStatus struct {
    Deletions *int `json:"deletions,omitempty"`
    Additions *int `json:"additions,omitempty"`
    Total     *int `json:"total,omitempty"`
}
```

That way, users of this library can look at at endpoint that returns `GistCommit` and immediately know what values will be populated inside `ChangeStatus`. When it's a map, you can't know that as easily. Also, editor auto-completion will help them discover those 3 values without leaving their editor. That's why it's an improvement that's worth doing whenever possible.

**Edit:** Actually, I just spotted a very similar struct [`CommitStats`](https://godoc.org/github.com/google/go-github/github#CommitStats). There is a chance we might want to reuse it instead of creating a separate one for Gist Commits. We should do that if we reasonably expect the two to stay in sync and that they represent the same thing. Even though they have slightly different names.

I'd probably lean towards trying to reuse it for now, and we can always create a dedicated `ChangeStatus` for gist later, should we notice the two diverge. So, this line can be simply replaced with:

```Go
ChangeStatus *ChangeStats `json:"change_status,omitempty"`
```

github/gists.go
```diff
@@ -52,6 +52,30 @@ func (g GistFile) String() string {
 	return Stringify(g)
 }
 
+// GistCommit represents a commit on a gist.
+type GistCommit struct {
```
shurcooL: Based on my reading of https://developer.github.com/v3/gists/#list-gist-commits, this struct seems to be missing the URL field:

```Go
URL *string `json:"url,omitempty"`
```

github/gists.go
```diff
@@ -52,6 +52,30 @@ func (g GistFile) String() string {
...
+	ChangeStatus map[string]int `json:"change_status,omitempty"`
+	CommitedAt   *Timestamp     `json:"commited_at,omitempty"`
+}
+
+func (gc GistCommit) String() string {
+	return Stringify(gc)
+}
+
+// GistFork represents a fork of a gist.
+type GistFork struct {
```
shurcooL: This is also missing URL field, based on https://developer.github.com/v3/gists/#list-gist-forks:

```Go
URL *string `json:"url,omitempty"`
```

github/gists.go
beshrkayali: @shurcooL I see. It didn't occur to me but you're right it seems like the change status has only these exact fields. Thanks.

Would it make sense to move `ChangeStats` type declaration out of `repos_commits.go` to somewhere more shared since it's not specific to repository commits anymore?

github/gists.go
beshrkayali: I noticed that other structs don't always map those one-to-one according to the doc (maybe I'm missing something though), for example: the `Gist` struct doesn't a URL field (nor `ForksURL` or `CommitsURL`) but they are mentioned in [the docs](https://developer.github.com/v3/gists/#get-a-single-gist). 

I have added the `URL` field to the `GistCommit` and `GistFork` structs.

github/gists.go
shurcooL: > Would it make sense to move `ChangeStats` type declaration out of `repos_commits.go` to somewhere more shared since it's not specific to repository commits anymore?

If there is such a place, moving it would be fine. But as long as you change its documentation to refer to both places it's used, I think there'll not be any problems.

github/gists.go
shurcooL: Fair enough. I think those are issues that should also be resolved.

It's possible the URL field is relatively new and didn't exist at the time those structs were added. Or maybe it was missed during the original code review.

I am not aware of any special rules or reasons to exclude the URL field.

shurcooL:
LGTM.

Huge thanks @beshrkayali! 🎉

gmlewis:
Incorporated in https://github.com/google/go-github/commit/e4b1c2870fd8416f2193ab449a4c889a9134b036.
Thank you, @beshrkayali!

beshrkayali:
Thanks @shurcooL + @gmlewis 👍

PR #390, opened 2016-06-28
Avoid unnecessary FooService allocations

### LINKED ISSUE #389: Optimization opportunity to reuse service struct allocations in github.Client
All of github.Client's service fields point to objects with identical underlying struct types.  Currently they're each separately heap allocated, but they could be made to reuse the same memory.  For example, add a field to github.Client:

           self       struct{ client *Client }

And then change github.NewClient to initialize the service fields as so:

           c.self.client = c
           c.Activity = (*ActivityService)(&c.self)
           c.Authorizations = (*AuthorizationsService)(&c.self)
           c.Gists = (*GistsService)(&c.self)
           ...

Just a random observation while looking at the API/implementation.  Feel free to close if you don't think this optimization is worthwhile.

mdempsky:
Fixes #389.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -107,7 +107,8 @@ type Client struct {
 	rateMu     sync.Mutex
 	rateLimits [categories]Rate // Rate limits for the client as determined by the most recent API calls.
 	mostRecent rateLimitCategory
-	common     service
+
+	common service
 
 	// Services used for talking to different parts of the GitHub API.
 	Activity       *ActivityService
--- a/github/github.go
+++ b/github/github.go
@@ -107,6 +107,7 @@ type Client struct {
 	rateMu     sync.Mutex
 	rateLimits [categories]Rate // Rate limits for the client as determined by the most recent API calls.
 	mostRecent rateLimitCategory
+	common     service
 
 	// Services used for talking to different parts of the GitHub API.
 	Activity       *ActivityService
@@ -123,6 +124,10 @@ type Client struct {
 	Licenses       *LicensesService
 	Migrations     *MigrationService
 	Reactions      *ReactionsService
+}
+
+type service struct {
+	client *Client
 }
 
 // ListOptions specifies the optional parameters to various List methods that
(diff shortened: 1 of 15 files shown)


github/github.go
```diff
@@ -107,6 +107,7 @@ type Client struct {
 	rateMu     sync.Mutex
 	rateLimits [categories]Rate // Rate limits for the client as determined by the most recent API calls.
 	mostRecent rateLimitCategory
+	common     service
```
shurcooL: I have a style suggestion.

```Go
rateMu     sync.Mutex
rateLimits [categories]Rate
mostRecent rateLimitCategory
```

Here, `rateMu` is a mutex hat. Like a hat, it sits on top of the variables that it protects.

So, **without needing to actually write the comment**, the above is implicitly understood to be equivalent to:

```Go
// rateMu protects rateLimits and mostRecent fields.
rateMu     sync.Mutex
rateLimits [categories]Rate
mostRecent rateLimitCategory
```

When adding a new, unrelated field that isn't protected by `rateMu`, don't do this:

```diff
 rateMu     sync.Mutex
 rateLimits [categories]Rate
 mostRecent rateLimitCategory
+common     service
```

Instead, do this:

```diff
 rateMu     sync.Mutex
 rateLimits [categories]Rate
 mostRecent rateLimitCategory
+
+common     service
```

How does that sound?

I can't remember where I first saw this pattern, but it might've been @davecheney. I really like it and recommend not breaking it. (I'll try to fix up the `clientMu` mutex on top to use this pattern too.)

mdempsky: > How does that sound?

SGTM, done.

github/github.go
```diff
@@ -174,20 +180,21 @@ func NewClient(httpClient *http.Client) *Client {
...
+	c.Gitignores = (*GitignoresService)(&c.common)
+	c.Issues = (*IssuesService)(&c.common)
+	c.Organizations = (*OrganizationsService)(&c.common)
+	c.PullRequests = (*PullRequestsService)(&c.common)
+	c.Repositories = (*RepositoriesService)(&c.common)
+	c.Search = (*SearchService)(&c.common)
+	c.Users = (*UsersService)(&c.common)
+	c.Licenses = (*LicensesService)(&c.common)
+	c.Migrations = (*MigrationService)(&c.common)
+	c.Reactions = (*ReactionsService)(&c.common)
```
willnorris: could you go ahead and sort these as well?  They were always supposed to be, but it looks like the last few got added to the bottom.

github/github.go
```diff
@@ -174,20 +180,21 @@ func NewClient(httpClient *http.Client) *Client {
...
-	c.Issues = &IssuesService{client: c}
-	c.Organizations = &OrganizationsService{client: c}
-	c.PullRequests = &PullRequestsService{client: c}
-	c.Repositories = &RepositoriesService{client: c}
-	c.Search = &SearchService{client: c}
-	c.Users = &UsersService{client: c}
-	c.Licenses = &LicensesService{client: c}
-	c.Migrations = &MigrationService{client: c}
-	c.Reactions = &ReactionsService{client: c}
+	c.common.client = c
```
shurcooL: Since this optimization is quite small and subtle, I think it'd be nice to leave a small, unobtrusive comment here.

That would reduce the chances of someone accidentally reverting this optimization a year from now, because they may not immediately understand why the code is arranged in this way instead of something simpler.

Something like:

```Go
c.common.client = c // Reuse a single struct instead of allocating one for each service on the heap.
```

(Or something along those lines, phrased in a better way.)

Placing it inline, on the right, is a cue to only read the comment it if you care about editing this section of code.

It's up to you, if you think it's worth it.

shurcooL: As a side benefit, someone who cares about finding more details/context can use find this commit via git blame.

So the comment can be very short, and a more detailed explanation can go into the commit message (which will reference the issue #390).

shurcooL:
See comments, but otherwise I think this looks great! LGTM.

My only ask is when merging this, squash into 1 commit so the relevant commit message pointing to more details/context is fewer git-blame hops away in the future.

willnorris:
made the minor changes suggested above, squashed and merged.  Thanks @mdempsky!

PR #391, opened 2016-06-29
Add ReadOnly boolean field to deploy keys

utay:
Hi there,

I just added the `ReadOnly` field (json: "read_only") to the Key structure (github/user_keys.go) so that we can specify with `true` or `false` if we want to make the deploy key read-only or not.

Here is the github API documentation which shows an example of read-only boolean:
https://developer.github.com/v3/repos/keys/#add-a-new-deploy-key

### DIFF
--- a/github/users_keys.go
+++ b/github/users_keys.go
@@ -9,10 +9,11 @@ import "fmt"
 
 // Key represents a public SSH key used to authenticate a user or deploy script.
 type Key struct {
-	ID    *int    `json:"id,omitempty"`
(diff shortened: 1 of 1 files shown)


utay:
I signed it!

gmlewis:
LGTM. :+1:
Thanks, @utay. I'll await another LGTM before merging.

willnorris:
LGTM

gmlewis:
Integrated in https://github.com/google/go-github/commit/8468db8b746699c51bb4b228ecfd498634c69088

PR #392, opened 2016-07-06
add (AuthorizationsService) CreateImpersonation/DeleteImpersonation

gmlewis:
Also add unit test for new methods.

Closes #342.

Change-Id: I657c9ac1af8334d6ff2aae56403a5531758fc2d1

### DIFF
--- a/github/authorizations.go
+++ b/github/authorizations.go
@@ -397,3 +397,40 @@ func (s *AuthorizationsService) DeleteGrant(id int) (*Response, error) {
 
 	return s.client.Do(req, nil)
 }
+
+// Create an impersonation OAuth token.
+//
+// This requires admin permissions. With the returned Authorization.Token
+// you can e.g. create or delete a user's public SSH key. NOTE: creating a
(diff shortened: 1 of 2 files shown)


gmlewis:
@willnorris - maybe I didn't go about this one correctly.
I pulled #342 into a pr/342 branch and then added the tests and amended the commit, so that the original author is @vetinari and the added tests were written by me.  Now googlebot is complaining that I'm not the original author, which is true.

Maybe I should just drop this one, create a new pr/342 (and fix all the merge problems), commit it, and then add a new PR with the unit tests for it.

I'll go ahead and close this.  Feel free to comment for future reference in case there is a better way to handle this situation.

willnorris:
After you pulled the changes in and amended the commit, you should have just submitted it directly.  No need to create a new pull request.

gmlewis:
Ah, I see.  I didn't do that because I wanted my unit tests reviewed... so it looks like I needed to split it up anyway.  Thanks, Will!

PR #396, opened 2016-07-09
Put sync.Mutex on top of variables it protects.

shurcooL:
This change makes the code more consistent with the `rateMu` below.

This is a followup to https://github.com/google/go-github/pull/390#discussion_r68863652.

I've also considered an alternative solution that kept the comments on top

```Go
// clientMu protects the client during calls that modify the CheckRedirect func.
clientMu sync.Mutex
// HTTP client used to communicate with the API.
client *http.Client
```

or

```Go
clientMu sync.Mutex // clientMu protects the client during calls that modify the CheckRedirect func.
// HTTP client used to communicate with the API.
client *http.Client
```

but decided to go with the inline version because I figured these are internal implementation details and unexported variables, so it's fine.

If there's any preference towards keeping the comments on top, I'm happy to change it back.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -88,10 +88,8 @@ const (
 
 // A Client manages communication with the GitHub API.
 type Client struct {
-	// HTTP client used to communicate with the API.
-	client *http.Client
-	// clientMu protects the client during calls that modify the CheckRedirect func.
-	clientMu sync.Mutex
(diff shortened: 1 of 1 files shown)


gmlewis:
LGTM 👍 

gmlewis:
Incorporated in https://github.com/google/go-github/commit/136ef5cdab23a105b08ae7ee273afdc557aa64b8

PR #398, opened 2016-07-10
Sync doc.go and README.md

### LINKED ISSUE #397: The text in package description in godoc and README.md has diverged.
This is a long-standing issue, and I'd really like for us to fix it.

The text in the package description in the godoc and in README.md has diverged. Each one describes and provides examples/documentations for certain things that the other doesn't, and vice versa.

- https://godoc.org/github.com/google/go-github/github
- https://github.com/google/go-github/blob/master/README.md#readme

This means it's harder to read that documentation, since you either need to consult both sources (that have a lot of overlap), or read only one and potentially miss out on something.

It also makes it harder to contribute improvements and changes to the documentation, since there are 2 places, it's unclear which one should be contributed to.

We should fix that by taking the best parts from each document and combine them into a single one. Then we can decide to move forward with one of these options:

- Keep both places for docs and manually keep them in sync.
    - **This is where my vote goes.**
- Remove most of README details and have it point to godoc.org documentation.
    - I would be okay with this option too.
- Remove most of godoc package documentation and keep the documentation only in README.md.
    - I would be strongly opposed to this option. It wouldn't be in spirit of Go to do that.
- Create/use some tool to generate README.md from the package documentation, make it so README.md is generated via `go generate`.
    - I don't think it's worth the effort here. The package docs don't change that often, so keeping it in sync manually is not very hard after we make the two texts identical.

      That said, I'll point out that the starting point of most of my Go repo README.md files is generated with [`gorepogen`](https://godoc.org/github.com/shurcooL/cmd/gorepogen) which does something like that, and it's very easy to extract the package documentation with [`go/doc`](https://godoc.org/go/doc#Package.Doc).

gmlewis:
Fixes #397.

Change-Id: I30ce336e3a36cf5758b9bf52b9dd77f7c1df5f27

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,33 +16,41 @@ import "github.com/google/go-github/github"
 ```
 
 Construct a new GitHub client, then use the various services on the client to
-access different parts of the GitHub API.  For example, to list all
-organizations for user "willnorris":
+access different parts of the GitHub API. For example:
 
 ```go
 client := github.NewClient(nil)
+
+// list all organizations for user "willnorris"
 orgs, _, err := client.Organizations.List("willnorris", nil)
 ```
 
-Some API methods have optional parameters that can be passed.  For example,
-to list public repositories for the "github" organization:
+Some API methods have optional parameters that can be passed. For example:
 
 ```go
 client := github.NewClient(nil)
-opt := &github.RepositoryListByOrgOptions{Type: "public"}
+
+// list recently updated repositories for org "github"
+opt := &github.RepositoryListByOrgOptions{Sort: "updated"}
 repos, _, err := client.Repositories.ListByOrg("github", opt)
 ```
 
+The services of a client divide the API into logical chunks and correspond to
+the structure of the GitHub API documentation at
+http://developer.github.com/v3/.
+
 ### Authentication ###
 
-The go-github library does not directly handle authentication.  Instead, when
+The go-github library does not directly handle authentication. Instead, when
 creating a new client, pass an `http.Client` that can handle authentication for
-you.  The easiest and recommended way to do this is using the [oauth2][]
+you. The easiest and recommended way to do this is using the [oauth2][]
 library, but you can always use any other library that provides an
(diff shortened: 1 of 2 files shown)


shurcooL:
This is so great! Thanks for doing this @gmlewis!

LGTM.

README.md
```diff
@@ -56,28 +64,100 @@ func main() {
...
+http://developer.github.com/v3/#rate-limiting.
+
+### Conditional Requests ###
+
+The GitHub API has good support for conditional requests which will help
+prevent you from burning through your rate limit, as well as help speed up your
+application. `go-github` does not handle conditional requests directly, but is
+instead designed to work with a caching `http.Transport`. We recommend using
+https://github.com/gregjones/httpcache, which can be used in conjunction with
+https://github.com/sourcegraph/apiproxy to provide additional flexibility and
```
shurcooL: I think we should remove the mention of https://github.com/sourcegraph/apiproxy here.

That repository does not compile successfully for over a year now (see https://github.com/sourcegraph/apiproxy/issues/1).

We've stopped supporting it or using it because it turned out that just using https://github.com/gregjones/httpcache for standards-compliant caching is good enough in most cases.

gmlewis: How does that last sentence sound now?

shurcooL: > `go-github` does not handle conditional requests directly, but is
> instead designed to work with a caching `http.Transport`. We recommend using
> https://github.com/gregjones/httpcache, which can be used to provide additional
> flexibility and control of caching rules.

I don't think that sounds right. It makes it sound like `gregjones/httpcache` is used to provide flexibility and control over caching rules, but in fact that package is what provides actual HTTP client-side caching. That sentence previously applied to `apicache`, which indeed did provide customization over http caching rules.

I think it should be more like this:

> `go-github` does not handle conditional requests directly, but is
> instead designed to work with a caching `http.Transport`. We recommend using
> https://github.com/gregjones/httpcache for that.

PR #400, opened 2016-07-13
remove CONTRIBUTORS file

willnorris:
Google no longer recommends CONTRIBUTORS files for projects, since this
data is better retrieved directly from version control.

Additionally, contributors are now instructed to add themselves (or
their employer, if appropriate) to the AUTHORS file, rather than the
maintainers being responsible for trying to keep this file up to date.
Ultimately, this should result in more accurate data, since in most
cases we don't know whether the code is copyright the contributor or
their employer.


parkr:
> Additionally, contributors are now instructed to add themselves (or
> their employer, if appropriate) to the AUTHORS file, rather than the
> maintainers being responsible for trying to keep this file up to date.

If I may, I would like to suggest that this be a part of some CI target. It would fail if the AUTHORS file excludes the information from the author's commit. Additionally, it should be indicated whether that change should be in a separate commit or all part of the same commit.

👍

willnorris:
> If I may, I would like to suggest that this be a part of some CI target. It would fail if the AUTHORS file excludes the information from the author's commit.

It is not a requirement that every author must be identified in this file.  The history of this file (in Google projects in general, not necessarily go-github in particular) is more for corporations that would like to have their contributions to open source projects be recognized, which are often not apparent just looking at the commit history.

> Additionally, it should be indicated whether that change should be in a separate commit or all part of the same commit.

Either is fine.  In practice, I don't expect this file to see a super high rate of change, so we can just see how it goes.

CONTRIBUTING.md
```diff
@@ -78,9 +78,13 @@ defined at <https://developer.github.com/v3/repos/hooks/> live in
...
-merging.
+**Verify CLAs.** CLAs must be on file for the pull request submitter and commit
+author(s). Google's CLA verification system should handle this automatically
+and will set commit statuses as appopriate. If there's ever any question about
+a pull request, ask [willnorris](https://github.com/willnorris).
+
+**Always try to maintain a clean, linear git history.** With very few
+exceptions, running `git log` should not show a bunch of branching and merging.
 
 Never use the GitHub "merge" button, since it always creates a merge commit.
```
shurcooL: Maybe this statement should be updated/revisited now that you can merge using "rebase and squash" strategy, which means the "since it always creates a merge commit" statement is no longer correct.

willnorris: I thought about that as I was making this change, but decided not to
include it here in order to keep this focused just on the contributors file.

shurcooL: Completely fine by me to consider it out of scope for this PR. 👍

CONTRIBUTING.md
```diff
@@ -78,9 +78,13 @@ defined at <https://developer.github.com/v3/repos/hooks/> live in
 
 (These notes are mostly only for people merging in pull requests.)
 
-**First and foremost, try to maintain a clean, linear git history.**  With very
-few exceptions, running `git log` should not show a bunch of branching and
-merging.
+**Verify CLAs.** CLAs must be on file for the pull request submitter and commit
+author(s). Google's CLA verification system should handle this automatically
+and will set commit statuses as appopriate. If there's ever any question about
```
shurcooL: s/appopriate/appropriate.

shurcooL:
I like this change a lot because it removes a manually maintained file which contains information that can be inferred automatically from git history. It's also one less file in the top level directory (without any loss of functionality). 🎉

LGTM.

CONTRIBUTING.md
willnorris: fixed

PR #404, opened 2016-07-18
Support qualifier for search query

huydx:
Reference: https://developer.github.com/v3/search
Currently without qualifier builder we have to build query manually.

```
searchService.Code("class+filename:foo.java")
```

which is quite inconvenient, especially for some time related qualifiers (created, updated...)

In my proposed implementation, qualifier for search query can be used by set appropriate values for `SearchQualifier` struct. Usage as seen in test code:

```
 qualifier := &SearchQualifiers{In: "in", Size: 123, Created: &now}
 opts := &SearchOptions{Qualifiers: qualifier}
```

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -6,7 +6,10 @@
 package github
 
 import (
+	"errors"
 	"fmt"
+	"reflect"
+	"time"
 
 	qs "github.com/google/go-querystring/query"
 )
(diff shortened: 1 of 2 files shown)


github/search.go
```diff
@@ -17,6 +20,42 @@ import (
...
+	Updated   *time.Time `qual:"pushed"`
+	Merged    *time.Time `qual:"merged"`
+	Closed    *time.Time `qual:"closed"`
+	User      string     `qual:"user"`
+	Repo      string     `qual:"repo"`
+	Repos     uint64     `qual:"repos"`
+	Language  string     `qual:"language"`
+	Stars     uint64     `qual:"stars"`
+	Path      string     `qual:"path"`
+	FileName  string     `qual:"filename"`
```
shurcooL: I'm not yet sure what to say to this PR at a high level (still thinking and waiting for others' comments), but at a low level, this should be `Filename`, not `FileName`. See similar fields:

- https://godoc.org/github.com/google/go-github/github#CommitFile.Filename
- https://godoc.org/github.com/google/go-github/github#GistFile.Filename
- https://godoc.org/github.com/google/go-github/github#GistFilename

willnorris:
How would you propose supporting the other comparison operators?  For example, see https://help.github.com/articles/search-syntax/ for all of the different options supported such as date ranges, less than and greater than, etc.  What you have only seems to work for exact value searches.

huydx:
@willnorris Oh now i know that search api support not only exact value searches. I will try enhance the builder to support another options also.

willnorris:
> @willnorris Oh now i know that search api support not only exact value searches. I will try enhance the builder to support another options also. Thanks for the suggestion.

Yeah, I think this was one of the main reasons we hadn't attempted to do this before... trying to come up with a Go data structure that can handle all of the possible permutations is non-trivial.  At the time, we just punted on it and pushed it onto the user to construct the query.  I'm not fundamentally opposed to the idea, I just worry about adding a lot of complexity to the library for marginal benefit.

huydx:
sorry i forgot to close this one. I've tried implemented but golang but can't found "just good" interface to support all the types. Some implements went too complex to solve just the simple problem, so i decide to close this one 😞 .

PR #368, opened 2016-06-15
add support for multiple issue assignees

gmlewis:
Fixes #362.

Change-Id: I39d142b4ef54bf514140a56b5fcbbcfa784f0c7f

PR #405, opened 2016-07-18
Add Request a Page Build API

### LINKED ISSUE #394: Add support for GitHub Pages preview API
https://developer.github.com/changes/2016-07-06-github-pages-preiew-api/

huytd:
Add new Request a Page Build API for GitHub Pages API.

Fixes #394 

The new API documented at this URL: https://developer.github.com/v3/repos/pages/#request-a-page-build

### DIFF
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -13,6 +13,7 @@ type Pages struct {
 	Status    *string `json:"status,omitempty"`
 	CNAME     *string `json:"cname,omitempty"`
 	Custom404 *bool   `json:"custom_404,omitempty"`
+	HTMLURL   *string `json:"html_url,omitempty"`
 }
 
 // PagesError represents a build error for a GitHub Pages site.
@@ -41,6 +42,9 @@ func (s *RepositoriesService) GetPagesInfo(owner string, repo string) (*Pages, *
 	if err != nil {
 		return nil, nil, err
 	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypePagesPreview)
 
 	site := new(Pages)
 	resp, err := s.client.Do(req, site)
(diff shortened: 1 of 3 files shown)


gmlewis:
Thank you for adding this, @huytd!

Could you please add "Fixes #394." to the commit message and edit the description above? (It would be cool if you squash the commits to keep it at 1 commit, but not strictly necessary.)

Also, you will need to add a custom media type to `github/github.go`  and both set the header and test for it. There are many examples of this in the repo... Check out this PR for one example: #368

huytd:
Thank you for the feedback @gmlewis 

I will make the changes and update this.

gmlewis:
Excellent!  For more details on the custom media type, please see the announcement: https://developer.github.com/changes/2016-07-06-github-pages-preiew-api/ or the documentation:
https://developer.github.com/v3/repos/pages/#request-a-page-build

huytd:
@gmlewis I made the changes and updated this PR. Could you help me check it please?

gmlewis:
@huytd - it is looking good... there are a couple more changes needed to fully address #394: 

`GetPagesInfo` needs the new custom media type and the `type Pages struct` needs a new HTMLURL field.

huytd:
@gmlewis Sorry for missing that one. I updated it.

There is one thing I'm still considering, is in `PagesError` struct, we're omitting the `Message` if it's empty, like this (am I correct about `omitempty`?):

```json
"error": {

},
```

But in the document, the `message` field should be there with a `null` value.

```json
"error": {
   "message": null
},
```

It seems not satisfying... Omitting this field if it's null will be a good idea? What do you think about this?

gmlewis:
@huytd - this should not be a problem because "omitempty" only affects the marshaling (encoding), not the unmarshaling (decoding).  For more information, please see: https://golang.org/pkg/encoding/json/#Marshal

I've tried to boil it down to a concrete example... hopefully this helps to explain the situation:
https://play.golang.org/p/9TbRm4yu97

Note that in Go, the `Error` field might be `nil` which is fine... meaning that there was no error. Presumably, if there was an error, the `Error` field will be populated and the unmarshaling will provide the Message field.

But maybe I am not understanding you correctly. What do you mean by:
> But in the document, ...

What document are you talking about?

huytd:
Ah yes, I was mentioned the API document here: https://developer.github.com/v3/repos/pages/#list-pages-builds

But it's fine, I understood it now.

Thank you so much.

gmlewis:
Thanks, @huytd!
LGTM :+1:

github/repos_pages.go
```diff
@@ -88,3 +92,25 @@ func (s *RepositoriesService) GetLatestPagesBuild(owner string, repo string) (*P
 
 	return build, resp, err
 }
+
+// RequestPageBuild request a build of a GitHub Pages site without needing to push new commit.
```
shurcooL: Minor, the second word here should be the verb "request**s**", not the noun "request". RequestPageBuild requests a build .

shurcooL:
Thanks for contributing @huytd!

See minor inline comment, otherwise LGTM, I didn't spot any other issues.

huytd:
Sorry my bad. I updated it.

Thank you so much for your feedback!

gmlewis:
Included in https://github.com/google/go-github/commit/769837dee39b31608be8a30304497e2f25240b6a

PR #403, opened 2016-07-18
Fix Mark{,Repository}NotificationsRead methods.

shurcooL:
Both of those endpoints use PUT method, which means they must send the parameters as a JSON encoded body rather than query parameters.

Source: https://developer.github.com/v3/#parameters

Fixes #402.

PR #406, opened 2016-07-18
Don't use Header.Add where Header.Set is sufficient.

shurcooL:
This is largely a style change. In the case of `NewRequest` and `NewUploadRequest`, it is completely a noop given current code (since a new request is created, and each header is added only one time). For `BasicAuthTransport`, it's a noop as long as incoming request doesn't already have "X-GitHub-OTP" header set (and if it did, then this is a better behavior to change its value rather than add a new one).

Even though these changes are largely noops, `Header.Set` is a simpler primitive to reason about, and it's more representative of the intended action. For "Content-Type", "Accept", and "User-Agent" headers, multiple such headers do not make sense, so it's better to use `Header.Set` than `Header.Add`.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -220,9 +220,9 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
 		return nil, err
 	}
 
-	req.Header.Add("Accept", mediaTypeV3)
+	req.Header.Set("Accept", mediaTypeV3)
 	if c.UserAgent != "" {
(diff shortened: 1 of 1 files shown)


willnorris:
LGTM

shurcooL:
I have another small change I'll be sending to be more compliant with the API specification. I sent this separately, since it was orthogonal, to make the reviews logically separate and easier. It depends on #403, so I'll send it after that makes it in.

PR #407, opened 2016-07-19
Set Content-Type to "application/json" for request bodies.

shurcooL:
According to https://developer.github.com/v3/#parameters:

> For POST, PATCH, PUT, and DELETE requests, parameters not included in the URL should be encoded as JSON with a Content-Type of 'application/json'

I'm not completely sure that we should merge this, but after I read what the GitHub API documentation said above, it seems like we should do it. Any thoughts?


gmlewis:
This seems totally reasonable to me.  I've tested this out on various non-GET requests (and also checked out some GET requests while I was at it), and it appears to work just fine.
LGTM. 👍

gmlewis:
Included in https://github.com/google/go-github/commit/b5e5babef39c18002f177a134fc49dc5013374ba

Issue #103, opened 2014-04-23
add missing struct fields


willnorris:
With the new [fields tool](https://github.com/google/go-github/blob/master/tests/fields/fields.go), it's now quite easy to identify which JSON fields are not being mapped into our resource structs.  Most of these are API URLs, but there are also a handful of other fields that got overlooked.

Originally, I had pushed back a bit on adding the API URLs since this library doesn't use them at all.  I've long since given up on that since a few use-cases have been mentioned, so we might as well go ahead and add them all.  I also have a few ideas for how the library could actually use these URLs or ones like them.

Related, the [API Versions](https://developer.github.com/v3/versions/) page documents some changes in API resources that we should make sure are addressed.

wesleimp:
@gmlewis this issue can be closed

gmlewis:
Closing.

PR #411, opened 2016-07-29
Adding some missing fields

jrbasso:
Refs #103

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -42,7 +42,8 @@ type PullRequest struct {
 	StatusesURL  *string    `json:"statuses_url,omitempty"`
 	DiffURL      *string    `json:"diff_url,omitempty"`
 	PatchURL     *string    `json:"patch_url,omitempty"`
-	Assignee     *User      `json:"assignee,omitempty"` // probably only in webhooks
+	Assignee     *User      `json:"assignee,omitempty"`
+	Assignees    []*User    `json:"assignees,omitempty"`
 
 	Head *PullRequestBranch `json:"head,omitempty"`
 	Base *PullRequestBranch `json:"base,omitempty"`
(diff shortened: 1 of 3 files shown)


shurcooL:
I haven't verified that the fields are valid, but I've looked at the code changes and not seeing any issues. LGTM if the fields are valid.

github/pulls.go
```diff
@@ -42,7 +42,8 @@ type PullRequest struct {
 	StatusesURL  *string    `json:"statuses_url,omitempty"`
 	DiffURL      *string    `json:"diff_url,omitempty"`
 	PatchURL     *string    `json:"patch_url,omitempty"`
-	Assignee     *User      `json:"assignee,omitempty"` // probably only in webhooks
+	Assignee     *User      `json:"assignee,omitempty"`
+	Assignees    []*User    `json:"assignees,omitempty"`
```
gmlewis: AFAICT, `Assignees` does not belong on a `PullRequest` struct, but does belong on an `Issue` struct and is already there on [line 39 of `issues.go`](https://github.com/google/go-github/blob/master/github/issues.go#L39).

If you still believe it belongs here, can you please point me to the GitHub documentation where it is mentioned?  Thanks!

jrbasso: @gmlewis According to [GH API](https://developer.github.com/v3/pulls/#labels-assignees-and-milestones), "Every pull request is an issue, but not every issue is a pull request."
That said, I believe all the fields from issues are applicable to pull requests.

Besides that, if you run the API request to this PR (`curl "https://api.github.com/repos/google/go-github/pulls/411"`) you going to see it there:

image

gmlewis: Ah!  Very good.  Thank you, @jrbasso!

gmlewis:
LGTM.  Merging.

github/pulls.go
jrbasso: @gmlewis You're welcome.

gmlewis:
Included in https://github.com/google/go-github/commit/56add9d4071bb1a06003a1239ebbc1500692b55d

PR #416, opened 2016-08-13
Adding missing field to search results

jrbasso:
Added the field incomplete_results. It is docummented on https://developer.github.com/v3/search/#timeouts-and-incomplete-results


shurcooL:
LGTM.

Thanks!

Issue #417, opened 2016-08-15
Add support for new repository traffic API


willnorris:
Announcement: https://developer.github.com/changes/2016-08-15-traffic-api-preview/

Docs: https://developer.github.com/v3/repos/traffic/

This is a tad bit more advanced, but would still be a good beginner task.  It doesn't require adding a new service (just put these methods on the existing RepositoriesService), but likely does involve adding a few new data types.

PR #393, opened 2016-07-06
add unit tests for #342 - CreateImpersonation and DeleteImpersonation

gmlewis:
Change-Id: Ib5fd4e99b88ff01e3363e049177326b7ee97be93

PR #418, opened 2016-08-20
Traffic api

Succo:
This is my take at https://github.com/google/go-github/issues/417

Critics welcome.


github/github.go
```diff
@@ -84,6 +84,9 @@ const (
 
 	// https://developer.github.com/changes/2016-07-06-github-pages-preiew-api/
 	mediaTypePagesPreview = "application/vnd.github.mister-fantastic-preview+json"
+
+	// https://developer.github.com/v3/repos/traffic/
+	mediaTypeTrafficPreview = "application/vnd.github.spiderman-preview"
```
shurcooL: There should be a `+json` suffix, see similar lines above.

For more context, see https://github.com/google/go-github/pull/337#discussion_r66712727.

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+	Count   *int    `json:"count,omitempty"`
+	Uniques *int    `json:"uniques,omitempty"`
+}
+
+// Time represents a time stamp as in a datapoint
+//
+// This is necessary becausegithub uses unix timestamp and not YYYY-MM-DDTHH:MM:SSZ
+type Time struct {
+	time.Time
+}
```
shurcooL: Can [`github.Timestamp`](https://godoc.org/github.com/google/go-github/github#Timestamp) be used, or is this different?

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+
+// Views represent information about views on the last 14 days
+type Views struct {
+	Views   *[]Datapoint `json:"views,omitempty"`
+	Count   *int         `json:"count,omitempty"`
+	Uniques *int         `json:"uniques,omitempty"`
+}
+
+// Clones represent information about clones on the last 14 days
```
shurcooL: This sentence is missing a period at the end. Also applies to most comments above in this file.

github/github.go
```diff
@@ -144,6 +147,11 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+// BreakdownOptions specifies the parameters to methods that support breakdown per day or week.
+type BreakdownOptions struct {
+	Per string `url:"per,omitempty"`
```
shurcooL: I think it'll be very helpful to document this `Per` parameter with something like this:

```
// Can be one of: day, week. Default: day.
```

For example, see https://github.com/google/go-github/blob/b9698164669e386680621e2006cab02271257611/github/repos.go#L113-L115.

shurcooL:
Thank you for contributing @Succo!

I've left a few inline comments as my initial review, please take a look and respond.

This is a great start.

github/github.go
```diff
@@ -144,6 +147,11 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+// BreakdownOptions specifies the parameters to methods that support breakdown per day or week.
+type BreakdownOptions struct {
```
parkr: Usually these options are scoped out to their method if only used once, e.g. TrafficBreakdownOptions. See OrgMembershipOptions, for example. But I defer to the real maintainers of this project for a second. :smile:

github/repos_traffic.go
Succo: Yeah I hadn't seen it.

Succo:
Updated the PR. I think the PR ran on a previous version of the commit, before I amended it with gofmt.

Not sure how to rerun it. (But I added gofmt on save to my IDE, so should not happen again).

shurcooL:
@Succo, would you like to work on adding tests for the new code?

You can see a good example of tests added in PR [#393](https://github.com/google/go-github/pull/393/files).

Also see an example PR [#373](https://github.com/google/go-github/pull/373/files) which added support forrepository invitations API, and also added tests. You can use that as a reference.

But really, any of the _test.go files in this package serve as good examples of tests.

Succo:
@shurcooL I added test based on the example you gave me.

willnorris:
(while @shurcooL is working with you on the code review, we do still need a [CLA](https://cla.developers.google.com/) from you before we can merge this.)

Succo:
@willnorris, I know, I'm working on getting access back to my google account.

Succo:
ok, should be good now for the CLA.

github/activity_events_test.go
```diff
@@ -229,7 +229,6 @@ func TestActivityService_ListEventsReceivedByUser_all(t *testing.T) {
 func TestActivityService_ListEventsReceivedByUser_publicOnly(t *testing.T) {
 	setup()
 	defer teardown()
-
```
shurcooL: The change on this line looks unintentional.

Succo: My bad, amended the commit.

github/github.go
```diff
@@ -144,6 +147,12 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+// TrafficBreakdownOptions specifies the parameters to methods that support breakdown per day or week.
```
willnorris: this should live in `repos_traffic.go` alongside the methods that use it.

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2013 The go-github AUTHORS. All rights reserved.
```
willnorris: use current year.  same in repos_traffic_test.go

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+	Path    *string `json:"path,omitempty"`
+	Title   *string `json:"title,omitempty"`
+	Count   *int    `json:"count,omitempty"`
+	Uniques *int    `json:"uniques,omitempty"`
+}
+
+// Time represents a time stamp as in a datapoint.
+//
+// This is necessary because github uses milliseconds timestamp.
+type Time struct {
```
willnorris: rather than creating a new type, perhaps we could modify https://godoc.org/github.com/google/go-github/github#Timestamp to handle millisecond precision times.  Right now it handles seconds, but using some pretty basic heuristics, we should be able to determine if a given value is intended to be seconds or milliseconds.  Something like:

```go
if i > 1e9 {
    i /= 1000
}
```

That'll at least work for dates up to the year 2286.  What do you think?

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+// Referrer represent a referrer information.
+type Referrer struct {
+	Referrer *string `json:"referrer,omitempty"`
+	Count    *int    `json:"count,omitempty"`
+	Uniques  *int    `json:"uniques,omitempty"`
+}
+
+// Path represent a referrer information.
```
willnorris: copy/paste error from Referrer type above

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+type Path struct {
```
willnorris: I worry a little about this name not being specific enough, given that file paths are used in a number of places throughout the library.  So maybe rename to `TrafficPath` or update the godoc to be very clear what this is referring to.

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return err
+	}
+	// We can drop the reaminder as returned values are days and it will always be 0
+	*t = Time{time.Unix(i/1000, 0)}
+	return nil
+}
+
+// Datapoint represent a view in views on clones.
+type Datapoint struct {
```
willnorris: maybe `TrafficView` here and `TrafficViews` below?  (I'm not really loving that myself, but these are awfully general names for such a specific part of the API)

github/repos_traffic.go
```diff
@@ -0,0 +1,167 @@
...
+type Clones struct {
+	Clones  *[]Datapoint `json:"clones,omitempty"`
+	Count   *int         `json:"count,omitempty"`
+	Uniques *int         `json:"uniques,omitempty"`
+}
+
+// ListReferrers list the top 10 referrers over the last 14 days.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/traffic/#list-referrers
+func (s *RepositoriesService) ListReferrers(owner, repo string) ([]*Referrer, *Response, error) {
```
willnorris: Same concern as above, but now on the methods.  I wonder if this should be `ListTrafficReferrers` (and similar naming style for methods below).  The fact that the URL for this is `/repos/o/r/traffic/popular/referrers` rather than simply `/repos/o/r/referrers` is a good indicator that maybe we should rename this.

github/repos_traffic_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"reflect"
+	"testing"
+	"time"
+)
+
+func TestRepositoriesService_ListReferrers(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/1/2/traffic/popular/referrers", func(w http.ResponseWriter, r *http.Request) {
```
willnorris: nit: as a matter of convention, we pretty much always use "o" and "r" for placeholder values to represent "owner" and "repo".  So, "/repos/o/r/traffic/popular/referrers"

github/repos_traffic.go
willnorris: (moving my comment from elsewhere, since I just saw this discussion...)

I do worry about this type causing confusion given the existing `Timestamp` type. Right now it handles seconds, but using some pretty basic heuristics, we should be able to determine if a given value is intended to be seconds or milliseconds. Something like:

```go
if i > 1e9 {
    i /= 1000
}
```

That'll at least work for dates up to the year 2286. What do you think?

willnorris: Although I guess this approach also causes problems with times before 1970-01-12 that are expressed in milliseconds (since they'd be interpreted as seconds and be about 30 years off).  Not sure how likely it is to have dates in that range in the traffic API though.  

Another option, which is *much* cleaner, but requires a bit more work is to support some basic tag options.  So you could have:

```go
struct {
    Timestamp *Timestamp `json:"timestamp,omitempty" precision:"millisecond"`
}
```

or something.  Though I'm pretty sure you don't have access to the struct tags inside of `Timestamp.UnmarshalJSON`, so a lot would have to be reworked.  Almost certainly not worth the trouble.

Succo:
@willnorris I went with a new naming scheme using traffic pretty much everywhere and clarifying the comments.

github/repos_traffic.go
Succo: I'm pretty sure some kind of workaround would be doable using "smart" Unmarshalling and estimation on the range of the timestamp (comparing it to `time.Now().Unix()` maybe) but that seems the kind of risky workaround that could surprise us later.

I tend to prefer clarity and safety to brevity, but as you want.

willnorris: Other alternative is just to rename the type for the time being (`TrafficTime` ? though that's still kinda terrible in other ways) so it doesn't get confused as being a general purpose time type.

willnorris: Or I guess `TimestampMS` or some variant like that might be better since there is nothing actually specific to the Traffic API about this.

Succo: That makes sense. I renamed it.

github/repos_traffic.go
willnorris: did this comment get overlooked?

Succo: Oh yeah sorry.

I renamed it, although it make it a bit less clearer for it's use in `TrafficClones`, not really inspired for a better name.

willnorris: ah, fair point.  I didn't realize it got used there also.  `TrafficData`, or even leave the "datapoint" like you had... `TrafficDatapoint`?

Succo: Wen't with TrafficData to avoid having a name too long.

willnorris:
This looks great.  Thanks for pushing through all the rounds of review! :) :+1:

PR #424, opened 2016-09-01
fix comment typo

yozamacs:


yozamacs:
I signed it!

github/search.go
```diff
@@ -70,7 +70,7 @@ func (s *SearchService) Issues(query string, opt *SearchOptions) (*IssuesSearchR
 	return result, resp, err
 }
 
-// UsersSearchResult represents the result of an issues search.
+// UsersSearchResult represents the result of an users search.
```
shurcooL: I think the article should also change to be "the result of **a** users search".

github/search.go
```diff
@@ -70,7 +70,7 @@ func (s *SearchService) Issues(query string, opt *SearchOptions) (*IssuesSearchR
...
 type UsersSearchResult struct {
 	Total             *int   `json:"total_count,omitempty"`
 	IncompleteResults *bool  `json:"incomplete_results,omitempty"`
```
shurcooL: Down below in this file it also says:

```
// CodeSearchResult represents the result of an code search.
type CodeSearchResult struct {
```

Mind correcting that as well to say "the result of **a** code search"?

shurcooL:
Thanks for contributing @yozamacs!

The change looks good so far. I've left some inline comments with suggestions for additional similar improvements you could do as part of this PR. Take a look and see if you want to take them.

yozamacs:
I did notice my commit message typo. That was autocorrect. I'll add the suggestions! Thanks!

willnorris:
squashed commits and merged.  Thanks!

PR #425, opened 2016-09-02
Add missing 'Type' field to 'GistFile' struct

bored-engineer:


shurcooL:
Confirmed by looking at https://developer.github.com/v3/gists/.

LGTM.

It looks like we're also missing `Language *string` and `Truncated *bool` fields in that struct. But that's not a blocker for this PR.

PR #427, opened 2016-09-07
Parse Webhook payload into actual event

apelisse:


apelisse:
Not sure this works, but would it make sense?

apelisse:
Actually it works

@willnorris is that something that could be merged?

github/messages.go
```diff
@@ -117,3 +120,41 @@ func validateSignature(signature string, payload, secretKey []byte) error {
...
+//
+// Example usage:
+//
+//     func (s *GitHubEventMonitor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+//       event, err := github.ParseWebHook(r, s.webhookSecretKey)
+//       if err != nil { ... }
+//       // Process event ...
+//     }
+//
+func ParseWebHook(r *http.Request, secretKey []byte) (interface{}, error) {
```
shurcooL: This returns an `interface{}`. Can you please help me understand what one would do with the return value of type `interface{}` from `ParseWebHook`?

The example just says "// Process event ...", but it's unclear to me how one would do that.

shurcooL: If I look at the implementation of what `ParseWebHook` does, then look at what `getWebhookEvent` does, I'll see that it returns `Event.Payload`. Then if I refer to documentation for [`Event.Payload`](https://godoc.org/github.com/google/go-github/github#Event.Payload), it begins to make sense:

> Payload returns the parsed event payload. For recognized event types, a value of the corresponding struct type will be returned.

The docs of `ParseWebHook` just say "returns an Event or a WebHookPayload" which isn't as clear.

apelisse:
Updated, does it make more sens @shurcooL

willnorris:
@gmlewis could you take a look and compare with our internal webhook processing code.  I do like how this reuses the existing logic in `Event.Payload`.

gmlewis:
@willnorris - It is similar to what we use internally, although the main difference is that we map the eventType to a specific handler, and then pass the `payload []byte` to the handler, which can then choose to use `Event.Payload` or not.

github/messages.go
```diff
@@ -117,3 +120,52 @@ func validateSignature(signature string, payload, secretKey []byte) error {
 	}
 	return nil
 }
+
+// getWebhookEvent parses the payload and request, and build the Event.
+func getWebhookEvent(r *http.Request, payload []byte) (interface{}, error) {
+	eventType := r.Header.Get(eventTypeHeader)
+	if eventType == "push" {
```
gmlewis: Why does the `push` `eventType` need to be handled specially?

github/messages.go
```diff
@@ -117,3 +120,52 @@ func validateSignature(signature string, payload, secretKey []byte) error {
...
+			return nil, err
+		}
+		return pushEvent, nil
+	}
+
+	event := Event{
+		Type:       &eventType,
+		RawPayload: (*json.RawMessage)(&payload),
+	}
+	return event.Payload(), nil
```
gmlewis: I don't think this is going to work for any of the other event types because the string returned in the header does not match the strings used in `Event.Payload`, correct?

apelisse: Yes, you're right. Unfortunately they are not the same type strings. Still, I wish go-github would include a piece of code that makes it easier to receive/parse webhooks.

Would it make sense to include a nice way to do it in go-github? Upcast made by the library, client code called with the actual object.

gmlewis: What about adding a simple `map[string]string` to convert these mismatched strings, and then "pull" would not be a special case at all and you could use `event.Payload()` for all processing?

apelisse: Sure, but where would you do `event.Payload`? `go-github` or client side?

gmlewis: I was actually suggesting adding it to this PR to make it work... but either way would be fine.

That's what we currently do, as I was saying earlier... we have a map from event type to event handler function, and it works great.

apelisse: Cool, I'll send a new PR. I think it would be a nice addition to this project.

Thanks for your help!

apelisse:
@gmlewis Updated

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
...
+		{
+			payload:     &WatchEvent{},
+			messageType: "watch",
+		},
+	}
+
+	for _, test := range tests {
+		p, err := json.Marshal(test.payload)
+		if err != nil {
+			t.Fatalf("Marshal(%#v): %v", test.payload, err)
```
gmlewis: For all of these test outputs, you might prefer "%+v" over "%#v" since we are dealing with pointers to structs... but your call... see which you output you think is clearer.

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
...
+		}
+		i, err := getWebhookEvent(test.messageType, p)
```
gmlewis: Can you please change `i` to `got` here and below?
I think it makes it easier to read the test and is a bit more idiomatic Go.

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("getWebHookEvent: %v", err)
+		}
+		if !reflect.DeepEqual(test.payload, i) {
+			t.Errorf("getWebHook(%#v, %#v) = %#v, wanted %#v", test.messageType, p, i, test.payload)
```
gmlewis: s/wanted/want/ is also a bit more idiomatic Go.

gmlewis:
A few minor nits - but I like how this turned out.
LGTM after the minor tweaks.

github/messages.go
```diff
@@ -117,3 +147,44 @@ func validateSignature(signature string, payload, secretKey []byte) error {
 	}
 	return nil
 }
+
+// getWebhookEvent parses the payload and request, and build the Event.
+func getWebhookEvent(messageType string, payload []byte) (interface{}, error) {
```
shurcooL: Rename to `parseWebhookEvent`, that seems like a much more fitting and consistent prefix than "get".

Also, should it be "Webhook" or "WebHook"? I see both used in this PR. But the rest of the codebase is also inconsistent. The exported code seems to use "WebHook", so we should probably go with that (since it's harder to change compared to documentation).

github/messages.go
```diff
@@ -117,3 +147,44 @@ func validateSignature(signature string, payload, secretKey []byte) error {
...
+//       switch event := event.(type) {
+//       case CommitCommentEvent:
+//           processCommitCommentEvent(event)
+//       case CreateEvent:
+//           processCreateEvent(event)
+//       ...
+//       }
+//     }
+//
+func ParseWebHook(r *http.Request, secretKey []byte) (interface{}, error) {
```
shurcooL: I still have a concern and a suggestion about the API design here.

The entire implementation of `ParseWebHook` currently consists of calling an exported func `ValidatePayload`, then an unexported `getWebhookEvent`. Calling a public API can be a warning sign.

I'm worried that with the way this function is written now, it leads to poorly readable usage code. Consider seeing:

```Go
func (s *GitHubEventMonitor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	event, err := github.ParseWebHook(r, s.webhookSecretKey)
	if err != nil { ... }
```

If I'm looking over code, I will likely think that there may be a security issue because the webhook is being parsed without the payload being first validated. I'd have to look up the docs for and see that "ParseWebhook parses the request, validates the signature", so it's safe. But I think it would be much more readable if `ParseWebhook` had smaller scope and only did one thing:

```Go
// ParseWebhook takes the request header and a validated payload, and returns
// an interface to an activity event type (as returned by Event.Payload()).
func ParseWebHook(header http.Header, validatedPayload []byte) (interface{}, error) {
```

Then usage becomes very explicit and clearly safe/correct:

```Go
func (s *GitHubEventMonitor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	validatedPayload, err := github.ValidatePayload(r, s.webhookSecretKey)
	if err != nil { ... }
	event, err := github.ParseWebHook(r.Header, validatedPayload)
	if err != nil { ... }
```

Since this deals with security, I'd much prefer being explicit and having readable usage code. Verbosity is okay.

Alternatively, you could consider keeping the method functionality as is but rename it to `ValidateAndParseWebHook` to better reflect what it does.

apelisse: Very valid comment, thanks for the feedback.

To go even further on your side, passing the secret key to `ParseWebHook` may be confusing. I guess I'll keep the two functions seperated from each other.

gmlewis:
Awaiting LGTM from @shurcooL

apelisse:
Yes, please wait, I'm about to push a new version :-)

apelisse:
@shurcooL updated

A couple of point:
- Not sure where you saw WebHook being used rather than Webhook. The existing documentation reads Webhook
-  I added a `ParseWebhookType` function. I don't want `ParseWebhook()` to depend on the request

Hope you like it!

shurcooL:
Thanks, I'll be able to review this tonight.

> Not sure where you saw WebHook being used rather than Webhook. The existing documentation reads Webhook

I see `WebHook` used in all the exported types.

See https://godoc.org/github.com/google/go-github/github.

image

apelisse:
OK Good catch, let me change that now so you can review it tonight.

Thanks

apelisse:
Updated

github/messages_test.go
shurcooL: Did you see this comment @apelisse?

It's not mandatory to change this, I just want to find out if you saw it and decided not to change it, or didn't see the comment.

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
...
+	for _, test := range tests {
+		p, err := json.Marshal(test.payload)
+		if err != nil {
+			t.Fatalf("Marshal(%#v): %v", test.payload, err)
+		}
+		got, err := ParseWebHook(test.messageType, p)
+		if err != nil {
+			t.Fatalf("getWebHookEvent: %v", err)
+		}
+		if !reflect.DeepEqual(test.payload, got) {
```
shurcooL: The common style is to compare got with want, in that order. Not the other way around.

```
github $ cat *.go | grep "got, want" | wc -l
     152
github $ cat *.go | grep "want, got" | wc -l
       0
```

It might be more readable as:

```Go
if want := test.payload; !reflect.DeepEqual(got, want) {
```

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
...
+			t.Errorf("getWebHook(%#v, %#v) = %#v, want %#v", test.messageType, p, got, test.payload)
```
shurcooL: `getWebHook` is out of date, it's called `ParseWebHook` now.

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
 		}
 	}
 }
+
+func TestGetWebHookEvent(t *testing.T) {
```
shurcooL: It should be `TestParseWebHook` now.

github/messages.go
```diff
@@ -117,3 +147,42 @@ func validateSignature(signature string, payload, secretKey []byte) error {
...
 	return nil
 }
+
+// ParseWebHookType returns the type of payload
+func ParseWebHookType(r *http.Request) string {
+	return r.Header.Get(eventTypeHeader)
+}
+
+// ParseWebHook parses the request into an interface to an activity
+// event type (as returned by Event.Payload())
```
shurcooL: This sentence is missing a period at the end, which is inconsistent.

The phrasing could be improved. There's no more request, its parameter is `payload []byte`. No need to mention interface, that's already in the return signature. What it returns is a struct type corresponding to the event type. Taking from `Event.Payload` docs, perhaps something like this:

```
// ParseWebHook parses the event payload. For recognized event types, a value of
// the corresponding struct type will be returned (as returned by Event.Payload()).
```

github/messages_test.go
```diff
@@ -79,3 +81,110 @@ func TestValidatePayload(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		p, err := json.Marshal(test.payload)
+		if err != nil {
+			t.Fatalf("Marshal(%#v): %v", test.payload, err)
+		}
+		got, err := ParseWebHook(test.messageType, p)
+		if err != nil {
+			t.Fatalf("getWebHookEvent: %v", err)
```
shurcooL: Same here. "ParseWebHook".

github/messages.go
```diff
@@ -117,3 +147,42 @@ func validateSignature(signature string, payload, secretKey []byte) error {
 	}
 	return nil
 }
+
+// ParseWebHookType returns the type of payload
+func ParseWebHookType(r *http.Request) string {
```
shurcooL: In this case, I wouldn't call this func parse.

Parse is a good name for something that can fail, and the signature usually has (Something, error) form.

Here, you're really just getting the type of payload from the request headers.

How about renaming it to just `WebHookType(r *http.Request)`? It feels like leaving out `Get` is better.

```
// WebHookType returns the event type of webhook request r.
func WebHookType(r *http.Request) string {
```

If not that, then call it `GetWebHookType`, but `ParseWebHookType` is definitely not the best name.

github/messages.go
```diff
@@ -30,6 +31,35 @@ const (
 	sha512Prefix = "sha512"
 	// signatureHeader is the GitHub header key used to pass the HMAC hexdigest.
 	signatureHeader = "X-Hub-Signature"
+	// eventTypeHeader is the Github header key used to pass the event type
```
shurcooL: Minor style nitpick. This sentence is missing a period at the end, which is inconsistent with the Go style and this codebase.

github/messages.go
```diff
@@ -30,6 +31,35 @@ const (
...
+	eventTypeHeader = "X-Github-Event"
+)
+
+var (
+	// eventTypeMapping is a conversion map between go-github types and webhooks types
```
shurcooL: Also missing a period.

The phrasing is odd. Isn't it a map from webhook types (e.g., "commit_comment") and go-github types (e.g., `CommitCommentEvent`)? The current phrasing makes it sound like the inverse of that.

github/messages.go
```diff
@@ -117,3 +147,42 @@ func validateSignature(signature string, payload, secretKey []byte) error {
...
+
+// ParseWebHookType returns the type of payload
+func ParseWebHookType(r *http.Request) string {
+	return r.Header.Get(eventTypeHeader)
+}
+
+// ParseWebHook parses the request into an interface to an activity
+// event type (as returned by Event.Payload())
+//
+// secretKey is the GitHub Webhook secret message.
```
shurcooL: There's no secret key anymore.

github/messages.go
```diff
@@ -117,3 +147,42 @@ func validateSignature(signature string, payload, secretKey []byte) error {
...
+//       switch event := event.(type) {
+//       case CommitCommentEvent:
+//           processCommitCommentEvent(event)
+//       case CreateEvent:
+//           processCreateEvent(event)
+//       ...
+//       }
+//     }
+//
+func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
```
shurcooL: This signature looks as nice as I expect it'll ever look, no further suggestions from me.

apelisse:
Updated, I think I've addressed each individual comment. There are no sentence not ending with dots.

Thanks again for your excellent feedback!

github/messages_test.go
shurcooL: @apelisse, did you see this?

github/messages.go
```diff
@@ -30,6 +31,35 @@ const (
 	sha512Prefix = "sha512"
 	// signatureHeader is the GitHub header key used to pass the HMAC hexdigest.
 	signatureHeader = "X-Hub-Signature"
+	// eventTypeHeader is the Github header key used to pass the event type.
+	eventTypeHeader = "X-Github-Event"
+)
+
+var (
+	// eventTypeMapping is a conversion map between go-github types and webhooks types.
```
shurcooL: What about improving the phrasing to say:

```
// eventTypeMapping maps webhooks types to their corresponding go-github struct types.
```

Or is that not a more accurate statement?

github/messages_test.go
apelisse: Yes, I did, I answered (but never published my comment). I like the `%#v` more because it prints the type, and we care about the type a lot here

github/messages.go
apelisse: OK Let's use that exact sentence, it's perfect (I think)

apelisse:
Updated :-)

shurcooL:
You've addressed all my feedback. Thanks!

I'll let @gmlewis or @willnorris take it from here.

gmlewis:
Very nice!  Thank you, @apelisse and @shurcooL !
LGTM. Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/94a3cd9f531888fe4a03487276c0adb887428e77

PR #428, opened 2016-09-10
add watch and unwatch

qianlnk:
https://developer.github.com/v3/activity/watching/#watch-a-repository-legacy

Watch a repository (LEGACY)
Stop watching a repository (LEGACY)

### DIFF
--- a/github/activity_watching.go
+++ b/github/activity_watching.go
@@ -134,3 +134,24 @@ func (s *ActivityService) DeleteRepositorySubscription(owner, repo string) (*Res
 
 	return s.client.Do(req, nil)
 }
+
+//Watch a repository as the authenticated user
+//https://developer.github.com/v3/activity/watching/#watch-a-repository-legacy
+func (s *ActivityService) Watch(owner, repo string) (*Response, error) {
+	u := fmt.Sprintf("user/subscriptions/%v/%v", owner, repo)
+	req, err := s.client.NewRequest("PUT", u, nil)
+	if err != nil {
+		return nil, err
+	}
+	return s.client.Do(req, nil)
+}
+
+// Unwatch a repository as the authenticated user.
+func (s *ActivityService) Unwatch(owner, repo string) (*Response, error) {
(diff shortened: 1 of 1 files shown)


qianlnk:
I signed it!

qianlnk:
the git commit email is the other one of mine, What should I do？

shurcooL:
Are there newer replacements available for those legacy endpoints?

qianlnk:
does func  SetRepositorySubscription can replace it? But I think it is not intuitive.

shurcooL:
I think so, yes.

From https://developer.github.com/v3/activity/watching/#set-a-repository-subscription:

> If you would like to watch a repository, set `subscribed` to `true`. If you would like to ignore notifications made within a repository, set `ignored` to `true`. If you would like to stop watching a repository, delete the repository's subscription completely.

That seems very reasonable to me. What about it do you find unintuitive?

Or are you referring to the [documentation of `SetRepositorySubscription`](https://godoc.org/github.com/google/go-github/github#ActivityService.SetRepositorySubscription) in go-github? If so, perhaps we should improve its documentation to be as clear as the GitHub API docs are.

I think we should avoid adding legacy/deprecated endpoints to go-github, if they're not neccessary (because a better replacement exists) and if they didn't already exist.

qianlnk:
yeah, you are right. Before, I don't know how to set the subscription.  So I wrote watch and unwatch according to API.

thanks!

shurcooL:
I think it'd be useful to include the text inside blue boxes in go-github documentation. Would you like to make a PR that adds these short paragraphs?

image
image

image

shurcooL:
I'll close this PR because we've come to conclusion that these legacy GitHub API endpoints are not needed, but instead we should improve go-github [documentation of `SetRepositorySubscription`](https://godoc.org/github.com/google/go-github/github#ActivityService.SetRepositorySubscription) to mention they can be used to watch/unwatch repos.

I've opened issue #477 to track that. Closing this. Thanks @qianlnk.

PR #434, opened 2016-09-25
Add installation webhook events

cyberdelia:
Add support for the new webhooks event for installations.

See for details:
- https://developer.github.com/early-access/integrations/webhooks/


cyberdelia:
I signed it!

github/event_types.go
```diff
@@ -146,6 +146,18 @@ type EditChange struct {
...
 
+// IntegrationEvent is triggered when an integration is created or modified.
+// The Webhook event name is "integration_installation" or
+// "integration_installation_repositories".
+//
+// GitHub docs: https://developer.github.com/early-access/integrations/webhooks/
+type IntegrationEvent struct {
+	Action            *string       `json:"action,omitempty"`
+	Installation      *Installation `json:"installation,omitempty"`
+	RepositoryAdded   []*Repository `json:"repositories_added,omitempty"`
```
gmlewis: s/RepositoryAdded/RepositoriesAdded/

github/event_types.go
```diff
@@ -146,6 +146,18 @@ type EditChange struct {
...
+	RepositoryRemoved []*Repository `json:"repositories_removed,omitempty"`
```
gmlewis: s/RepositoryRemoved/RepositoriesRemoved/

github/installation.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2013 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2013/2016/

github/event_types.go
```diff
@@ -146,6 +146,18 @@ type EditChange struct {
...
+}
```
gmlewis: Can you please add:
```go
Sender *User `json:"sender,omitempty"`
```
to this struct?

github/event_types.go
```diff
@@ -146,6 +146,18 @@ type EditChange struct {
 	} `json:"body,omitempty"`
 }
 
+// IntegrationEvent is triggered when an integration is created or modified.
+// The Webhook event name is "integration_installation" or
+// "integration_installation_repositories".
+//
+// GitHub docs: https://developer.github.com/early-access/integrations/webhooks/
+type IntegrationEvent struct {
+	Action            *string       `json:"action,omitempty"`
```
gmlewis: Can you please document this field?  For example:
```go
// The action that was performed. Possible values for an "integration_installation"
// event are: "created", "deleted". Possible values for an "integration_installation_repositories"
// event are: "added", "removed".
Action ...
```
(Following the style of the documentation found in other files in this repo.)

cyberdelia:
@gmlewis I addressed your feedback, thanks!

github/installation.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: On second thought, does this need its own separate file?
How about defining the `Installation` struct with the `IntegrationEvent` in `event_types.go`?

cyberdelia: I was merely trying to follow the structure of the other events, but that's a fairly special case, so maybe?

gmlewis: OK, let's see what @shurcooL says.

gmlewis:
@shurcooL - whenever you have time to check this out, that would be greatly appreciated.
Thanks!

github/installation.go
shurcooL: I agree this seems to fit really nicely right under the definition of `IntegrationEvent` that first mentions it. It doesn't seem like a win to break it out into a separate tiny file.

> I was merely trying to follow the structure of the other events

I don't quite see where that's the case, can you point it out please?

github/event_types.go
```diff
@@ -146,6 +146,22 @@ type EditChange struct {
...
+	// The action that was performed. Possible values for an "integration_installation"
+	// event are: "created", "deleted". Possible values for an "integration_installation_repositories"
+	// event are: "added", "removed".
+	Action              *string       `json:"action,omitempty"`
+	Installation        *Installation `json:"installation,omitempty"`
+	RepositoriesAdded   []*Repository `json:"repositories_added,omitempty"`
+	RepositoriesRemoved []*Repository `json:"repositories_removed,omitempty"`
+	Sender              *User         `json:"sender,omitempty"`
+}
+
```
shurcooL: I.e., `Installation ` could go right here.

github/messages.go
```diff
@@ -38,27 +38,29 @@ const (
...
-		"watch":                       "WatchEvent",
+		"commit_comment":                        "CommitCommentEvent",
+		"create":                                "CreateEvent",
+		"delete":                                "DeleteEvent",
+		"deployment":                            "DeploymentEvent",
+		"deployment_status":                     "DeploymentStatusEvent",
+		"fork":                                  "ForkEvent",
+		"gollum":                                "GollumEvent",
+		"integration_installation":              "IntegrationEvent",
+		"integration_installation_repositories": "IntegrationEvent",
```
shurcooL: Maybe I'm missing something, but why are these two events/webhook types combined into a single shared struct type? According to https://developer.github.com/early-access/integrations/webhooks/, they have different fields.

Why not have `IntegrationInstallationEvent` and `IntegrationInstallationRepositoriesEvent`?

Somewhat similar to `pull_request_review_comment` and `pull_request` mapping to their own separate types.

Are they different or special in a way? I'm not very familiar with the details here, so I'm asking.

cyberdelia:
I moved ``Installation`` in the same file as the events, and separated the two webhooks events.

shurcooL:
LGTM. I fixed a typo and made the URLs more specific, hope that's okay.

cyberdelia:
Looks good to me, thanks.

gmlewis:
LGTM.  Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/b0b867f2d5ec7c0139fad20ab0fe9267dcd8312b
Thank you, @cyberdelia and @shurcooL !

PR #436, opened 2016-09-28
Add support for integration service

cyberdelia:
Installations do have an one endpoint that list the repository they have access to. 

Sadly, it's fairly undocumented for now, the only mention of it can be found the webhooks documentation, see: https://developer.github.com/early-access/integrations/webhooks/#integrationinstallationevent

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -90,6 +90,9 @@ const (
 
 	// https://developer.github.com/changes/2016-09-14-projects-api/
 	mediaTypeProjectsPreview = "application/vnd.github.inertia-preview+json"
+
+	// https://developer.github.com/v3/integrations/
+	mediaTypeIntegrationPreview = "application/vnd.github.machine-man-preview+json"
 )
 
 // A Client manages communication with the GitHub API.
@@ -120,6 +123,7 @@ type Client struct {
 	Gists          *GistsService
 	Git            *GitService
 	Gitignores     *GitignoresService
+	Integrations   *IntegrationsService
 	Issues         *IssuesService
 	Organizations  *OrganizationsService
 	PullRequests   *PullRequestsService
@@ -190,6 +194,7 @@ func NewClient(httpClient *http.Client) *Client {
 	c.Gists = (*GistsService)(&c.common)
 	c.Git = (*GitService)(&c.common)
 	c.Gitignores = (*GitignoresService)(&c.common)
+	c.Integrations = (*IntegrationsService)(&c.common)
 	c.Issues = (*IssuesService)(&c.common)
 	c.Licenses = (*LicensesService)(&c.common)
 	c.Migrations = (*MigrationService)(&c.common)
--- a/github/integration.go
+++ b/github/integration.go
@@ -1,0 +1,37 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+// IntegrationsService provides access to the installation related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/integrations/
+type IntegrationsService service
+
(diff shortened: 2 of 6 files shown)


willnorris:
haven't looked closely, just a cursory comment on naming.  Services in go-github typically map to how the API documentation is structured.  That's a little hard right now, if this is an undocumented API, and given that the integrations stuff is in preview.  However, I think that installations are always done in the context of integrations.  So working with installations will almost certainly need to be a set of functions on an `IntegrationsService`, rather than having a standalone `InstallationsService`.

cyberdelia:
Some documentation appeared, I will makes the changes to match it: https://developer.github.com/v3/integrations/

cyberdelia:
I've rebased, matched and added links to the new documentation.

github/integration_installation.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
```
gmlewis: Before this line, could you please add:
`// TODO: remove custom Accept header when this API fully launches.`

github/integration.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
```
gmlewis: Before this line, could you please add:
`// TODO: remove custom Accept header when this API fully launches.`

github/integration.go
```diff
@@ -0,0 +1,37 @@
...
+// IntegrationsService provides access to the installation related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/integrations/
+type IntegrationsService service
+
+// ListInstallations lists the installations that the current integration has.
+//
+// GitHub API docs: https://developer.github.com/v3/integrations/#find-installations
+func (s *IntegrationsService) ListInstallations(opt *ListOptions) ([]*Installation, *Response, error) {
```
gmlewis: Can you please add a test function for this one?

github/integration_installation.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (i Installation) String() string {
+	return Stringify(i)
+}
+
+// ListRepos lists the repositories that the current installation has access to.
+//
+// GitHub API docs: https://developer.github.com/v3/integrations/installations/#list-repositories
+func (s *IntegrationsService) ListRepos(opt *ListOptions) ([]*Repository, *Response, error) {
```
gmlewis: Can you please add a test function for this one?

cyberdelia:
@gmlewis I added the TODO comments and tests.

gmlewis:
Thank you, @cyberdelia!
LGTM.  Awaiting secondary LGTM before merging.

github/github.go
```diff
@@ -90,6 +90,9 @@ const (
 
 	// https://developer.github.com/changes/2016-09-14-projects-api/
 	mediaTypeProjectsPreview = "application/vnd.github.inertia-preview+json"
+
+	// https://developer.github.com/v3/integrations/
```
shurcooL: Extremely minor, but most other comments above link to the changes page corresponding to the accept header value. So this could be https://developer.github.com/changes/2016-09-14-Integrations-Early-Access/.

gmlewis: Thanks, @shurcooL - I can address this one in my merge.

gmlewis:
Integrated in https://github.com/google/go-github/commit/93d3f730320429a6e1ecf2ac1778db832930948a

PR #438, opened 2016-09-28
Add support for new Projects Preview API

### LINKED ISSUE #432: Add support for new Projects Preview API
See blog post for more information:
https://developer.github.com/changes/2016-09-14-projects-api/

gmlewis:
Fixes #432.

This also fixes a small typo in repos_forks.go for a GitHub API URL.

Change-Id: I4792114482f412f0eff54a6899a759fc9393ff88

### DIFF
--- a/github/repos_forks.go
+++ b/github/repos_forks.go
@@ -50,7 +50,7 @@ type RepositoryCreateForkOptions struct {
 
 // CreateFork creates a fork of the specified repository.
 //
-// GitHub API docs: http://developer.github.com/v3/repos/forks/#list-forks
+// GitHub API docs: https://developer.github.com/v3/repos/forks/#create-a-fork
 func (s *RepositoriesService) CreateFork(owner, repo string, opt *RepositoryCreateForkOptions) (*Repository, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/forks", owner, repo)
 	u, err := addOptions(u, opt)
--- a/github/repos_projects.go
+++ b/github/repos_projects.go
@@ -1,0 +1,464 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"fmt"
+)
+
+// Project represents a GitHub Project.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/projects/
+type Project struct {
+	ID        *int       `json:"id,omitempty"`
+	URL       *string    `json:"url,omitempty"`
+	OwnerURL  *string    `json:"owner_url,omitempty"`
+	Name      *string    `json:"name,omitempty"`
+	Body      *string    `json:"body,omitempty"`
+	Number    *int       `json:"number,omitempty"`
+	CreatedAt *Timestamp `json:"created_at,omitempty"`
+	UpdatedAt *Timestamp `json:"updated_at,omitempty"`
+
+	// The User object that generated the project.
+	Creator *User `json:"creator,omitempty"`
+}
+
+func (p Project) String() string {
+	return Stringify(p)
+}
+
+// ListProjects lists the projects for a repo.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/projects/#list-projects
+func (s *RepositoriesService) ListProjects(owner, repo string, opt *ListOptions) ([]*Project, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/projects", owner, repo)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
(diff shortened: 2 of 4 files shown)


github/repos_projects.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	return project, resp, err
+}
+
+// ProjectInfo specifies the parameters to the
+// RepositoriesService.CreateProject and RepositoriesService.UpdateProject methods.
+type ProjectInfo struct {
+	// The name of the project. (Required for creation; optional for update.)
+	Name string `json:"name"`
```
shurcooL: If it's optional for update, shouldn't it need `,omitempty` in that case?

github/repos_projects.go
```diff
@@ -0,0 +1,461 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return project, resp, err
+}
+
+// ProjectInfo specifies the parameters to the
+// RepositoriesService.CreateProject and RepositoriesService.UpdateProject methods.
+type ProjectInfo struct {
```
shurcooL: All other similar structs (that are used as last argument to the method in order to provide additional fields) are usually named with "Options" suffix in this library:

- https://godoc.org/github.com/google/go-github/github#ActivityListStarredOptions
- https://godoc.org/github.com/google/go-github/github#CommitsListOptions
- https://godoc.org/github.com/google/go-github/github#IssueListByRepoOptions
- https://godoc.org/github.com/google/go-github/github#RepositoryContentFileOptions
- ...

There aren't any with an Info suffix like in this PR.

Did you choose to use Info instead of Options on purpose? If so, what was the reason?

github/repos_projects.go
```diff
@@ -0,0 +1,461 @@
...
+// RepositoriesService.CreateProjectColumn and RepositoriesService.UpdateProjectColumn methods.
+type ProjectColumnInfo struct {
+	// The name of the project column. (Required for creation and update.)
+	Name string `json:"name"`
+}
+
+// CreateProjectColumn creates a column for the specified (by number) project.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/projects/#create-a-column
+func (s *RepositoriesService) CreateProjectColumn(owner, repo string, number int, columnInfo *ProjectColumnInfo) (*ProjectColumn, *Response, error) {
```
shurcooL: In this case, `ProjectColumnInfo` has only one field and it's required (not optional). Yet `columnInfo` is a pointer instead of a value, so it can be `nil`.

I wanted to check if this was consistent with other parts of this library, and whether we could make it a value instead.

It appears to be very consistent, all similar structs are always passed as a pointer. I'm not sure why that's the case, since it seems like friendlier API to not give an option to pass nil when it's meaningless/unsupported. Perhaps it was done for performance.

In any case, this is fine, just wanted to comment on it.

github/repos_forks.go
```diff
@@ -50,7 +50,7 @@ type RepositoryCreateForkOptions struct {
 
 // CreateFork creates a fork of the specified repository.
 //
-// GitHub API docs: http://developer.github.com/v3/repos/forks/#list-forks
+// GitHub API docs: https://developer.github.com/v3/repos/forks/#create-a-fork
```
shurcooL: 👍

github/repos_projects.go
gmlewis: Good catch!  I pulled "Info" out of the thin air.  Thanks, @shurcooL !  I'll fix that.

gmlewis:
All comments addressed. Thank you, @shurcooL for the great review!
I'll let this sit another day in case there are any other comments since it is rather large.

github/repos_projects.go
```diff
@@ -0,0 +1,464 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return projects, resp, err
+}
+
+// GetProject gets a GitHub Project for a repo.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/projects/#list-a-project
```
shurcooL: Minor issue here. The fragment part of URL is invalid, it should be:

https://developer.github.com/v3/repos/projects/#get-a-project

github/repos_projects.go
```diff
@@ -0,0 +1,464 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return cards, resp, err
+}
+
+// GetProjectCard gets a card in a column of a GitHub Project.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/projects/#list-a-project-card
```
shurcooL: Similarly here.

https://developer.github.com/v3/repos/projects/#get-a-project-card

github/repos_projects.go
gmlewis: Ah, yes!  GitHub Support changed the URLs at my request.  😄 
Thanks, @shurcooL!

PR #439, opened 2016-10-02
Fix type of RepositoryRelease Author field.

dpetersen:
The author JSON is in [the format of a user](https://developer.github.com/v3/repos/releases/#get-a-single-release), not a commit author, and so the fields for the author returned in all the release fetching methods were nil.

There wasn't a test specifically for the decoding of the user, so I added a simple one in the fashion of [a similar test in the commit fetching code](https://github.com/google/go-github/blob/master/github/git_commits_test.go#L16-L35). The unit tests pass locally and this seems to work for me in some code that is using the library and was seeing a nil author for a release where the GitHub API is correctly returning one. I did not run the integration tests because they don't appear to cover releases anyway, and I need my token to not be exhausted right now.

Thanks for this library, it works great!

### DIFF
--- a/github/repos_releases_test.go
+++ b/github/repos_releases_test.go
@@ -44,7 +44,7 @@ func TestRepositoriesService_GetRelease(t *testing.T) {
 
 	mux.HandleFunc("/repos/o/r/releases/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		fmt.Fprint(w, `{"id":1}`)
+		fmt.Fprint(w, `{"id":1,"author":{"login":"l"}}`)
 	})
 
 	release, resp, err := client.Repositories.GetRelease("o", "r", 1)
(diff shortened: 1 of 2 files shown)


dpetersen:
I signed it!

github/repos_releases_test.go
```diff
@@ -44,15 +44,15 @@ func TestRepositoriesService_GetRelease(t *testing.T) {
 
 	mux.HandleFunc("/repos/o/r/releases/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		fmt.Fprint(w, `{"id":1}`)
+		fmt.Fprint(w, `{"id":1, "author": { "login": "authorname" }}`)
```
shurcooL: Minor style suggestion. The whitespace in the JSON is not consistent with most other places in this repo. When written as a single line, they omit all optional whitespace. See examples:

- https://github.com/google/go-github/blob/3e246d29992beb5a5d2fc253adebea86997c45b0/github/repos_releases_test.go#L264
- https://github.com/google/go-github/blob/3e246d29992beb5a5d2fc253adebea86997c45b0/github/pulls_test.go#L76

shurcooL: Also, most other test data uses `"login":"l"` as login value, not "authorname". It doesn't matter much, but it's nice to be more consistent.

dpetersen:
Thanks for looking at this. I didn't know about the webhook, but I [found it in the documentation](https://developer.github.com/v3/activity/events/types/#releaseevent) and it appears to use the exact same author format as the main releases endpoints.

I've also amended my commit to match your style suggestions.

gmlewis:
Thank you, @dpetersen and @shurcooL!
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/a1746f282059a5d71c4d8a270753698e164151e5

dpetersen:
Great, thank you!

PR #440, opened 2016-10-03
Local check for comments which are too long

eparis:
If a comment exceeds 64k github will reject it:

POST https://api.github.com/repos/kubernetes/kubernetes/issues/33381/comments: 422 Validation Failed [{Resource:IssueComment Field:body Code:custom Message:body is too long (maximum is 65536 characters)}]

Do a local check and error in the library instead.

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -236,8 +236,14 @@ func (s *IssuesService) Get(owner string, repo string, number int) (*Issue, *Res
 
 // Create a new issue on the specified repository.
 //
+// Issues with a body longer than 65536 will return an error.
+//
 // GitHub API docs: http://developer.github.com/v3/issues/#create-an-issue
 func (s *IssuesService) Create(owner string, repo string, issue *IssueRequest) (*Issue, *Response, error) {
+	if issue != nil && issue.Body != nil && len(*issue.Body) > maxCommentLen {
+		return nil, nil, fmt.Errorf("body is too long (maximum is %d characters)", maxCommentLen)
+	}
+
 	u := fmt.Sprintf("repos/%v/%v/issues", owner, repo)
 	req, err := s.client.NewRequest("POST", u, issue)
 	if err != nil {
--- a/github/issues_comments.go
+++ b/github/issues_comments.go
@@ -10,6 +10,9 @@ import (
 	"time"
 )
 
+// maxCommentLen is the maximum length of a comment that github will accept without a validation error.
+const maxCommentLen = 65536
+
 // IssueComment represents a comment left on an issue.
 type IssueComment struct {
 	ID        *int       `json:"id,omitempty"`
@@ -100,8 +103,13 @@ func (s *IssuesService) GetComment(owner string, repo string, id int) (*IssueCom
 
 // CreateComment creates a new comment on the specified issue.
 //
+// Comments longer than 65536 will return an error.
+//
 // GitHub API docs: http://developer.github.com/v3/issues/comments/#create-a-comment
 func (s *IssuesService) CreateComment(owner string, repo string, number int, comment *IssueComment) (*IssueComment, *Response, error) {
+	if comment != nil && comment.Body != nil && len(*comment.Body) > maxCommentLen {
+		return nil, nil, fmt.Errorf("body is too long (maximum is %d characters)", maxCommentLen)
+	}
 	u := fmt.Sprintf("repos/%v/%v/issues/%d/comments", owner, repo, number)
 	req, err := s.client.NewRequest("POST", u, comment)
 	if err != nil {
--- a/github/pulls_comments.go
+++ b/github/pulls_comments.go
@@ -107,8 +107,14 @@ func (s *PullRequestsService) GetComment(owner string, repo string, number int) 
 
 // CreateComment creates a new comment on the specified pull request.
 //
+// Comments longer than 65536 will return an error.
+//
 // GitHub API docs: https://developer.github.com/v3/pulls/comments/#create-a-comment
 func (s *PullRequestsService) CreateComment(owner string, repo string, number int, comment *PullRequestComment) (*PullRequestComment, *Response, error) {
+	if comment != nil && comment.Body != nil && len(*comment.Body) > maxCommentLen {
+		return nil, nil, fmt.Errorf("body is too long (maximum is %d characters)", maxCommentLen)
+	}
+
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/comments", owner, repo, number)
 	req, err := s.client.NewRequest("POST", u, comment)
 	if err != nil {


eparis:
If you do it too many times github will start giving you:
`You have triggered an abuse detection mechanism and have been temporarily blocked from content creation. Please retry your request again later.`
And then other calls will start failing. So better to never get into that situation.

github/issues_comments.go
```diff
@@ -10,6 +10,9 @@ import (
 	"time"
 )
 
+// MaxCommentLen is the maximum length of a comment that github will accept without a validationg error.
+var MaxCommentLen = 65535
```
shurcooL: In the x, it says:

x

Also, does this need to be exported? It looks to me that it's not helpful to be a part of API and can be made unexported. Any arguments for why it should be exported?

eparis: Whoops, will make it 65536. I swear I double checked that.

I made it external so developers could update the library if github ever changed their limit. No, it does not need to be external.

willnorris: If GitHub ever changed their limit, then we should simply update the library.  I think I'd rather leave this unexported for the time being.  We could always expose it later if there is a strong argument for it, but it's much harder to go the other way.

eparis:
Updated. If we want more or better docs let me know. I wasn't sure how much we wanted to say...

eparis:
1.5 months later. updated.

github/issues_comments.go
```diff
@@ -100,8 +103,13 @@ func (s *IssuesService) GetComment(owner string, repo string, id int) (*IssueCom
 
 // CreateComment creates a new comment on the specified issue.
 //
+// Comments longer than 65536 will return an error.
+//
 // GitHub API docs: http://developer.github.com/v3/issues/comments/#create-a-comment
 func (s *IssuesService) CreateComment(owner string, repo string, number int, comment *IssueComment) (*IssueComment, *Response, error) {
+	if comment != nil && comment.Body != nil && len(*comment.Body) > maxCommentLen {
+		return nil, nil, fmt.Errorf("Comments may not be longer than %d", maxCommentLen)
```
shurcooL: Error strings should not be capitalized, see https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

Actually, can we use the same phrasing as GitHub? According to the PR description, it would be something like:

> body is too long (maximum is 65536 characters)

shurcooL: Error strings should not be capitalized, see https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

Actually, can we use the same phrasing as what GitHub API returns? According to the PR description, it would be something like:

> body is too long (maximum is 65536 characters)

github/issues_comments.go
```diff
@@ -10,6 +10,9 @@ import (
 	"time"
 )
 
+// maxCommentLen is the maximum length of a comment that github will accept without a validationg error.
```
shurcooL: Typo in "validationg", should be "validation".

github/issues_comments.go
```diff
@@ -100,8 +103,13 @@ func (s *IssuesService) GetComment(owner string, repo string, id int) (*IssueCom
...
 // CreateComment creates a new comment on the specified issue.
 //
+// Comments longer than 65536 will return an error.
+//
 // GitHub API docs: http://developer.github.com/v3/issues/comments/#create-a-comment
 func (s *IssuesService) CreateComment(owner string, repo string, number int, comment *IssueComment) (*IssueComment, *Response, error) {
+	if comment != nil && comment.Body != nil && len(*comment.Body) > maxCommentLen {
+		return nil, nil, fmt.Errorf("Comments may not be longer than %d", maxCommentLen)
+	}
 	u := fmt.Sprintf("repos/%v/%v/issues/%d/comments", owner, repo, number)
```
shurcooL: Add a blank line before `u := ...` to separate it, like you've done in `IssuesService.Create`. The local validation is a separate section that has nothing to do with this code.

github/issues.go
```diff
@@ -238,6 +238,10 @@ func (s *IssuesService) Get(owner string, repo string, number int) (*Issue, *Res
 //
 // GitHub API docs: http://developer.github.com/v3/issues/#create-an-issue
```
shurcooL: Both `IssuesService.CreateComment` and `PullRequestsService.CreateComment` have a new "Comments longer than 65536 will return an error." sentence in the comment mentioning the limit. Why not here?

github/issues.go
```diff
@@ -236,8 +236,14 @@ func (s *IssuesService) Get(owner string, repo string, number int) (*Issue, *Res
 
 // Create a new issue on the specified repository.
 //
+// Issues with a body longer than 65536 will return an error.
```
shurcooL: "longer than 65536 **characters**" would sound better.

github/issues_comments.go
```diff
@@ -100,8 +103,13 @@ func (s *IssuesService) GetComment(owner string, repo string, id int) (*IssueCom
 
 // CreateComment creates a new comment on the specified issue.
 //
+// Comments longer than 65536 will return an error.
```
shurcooL: "longer than 65536 **characters**".

github/pulls_comments.go
```diff
@@ -107,8 +107,14 @@ func (s *PullRequestsService) GetComment(owner string, repo string, number int)
 
 // CreateComment creates a new comment on the specified pull request.
 //
+// Comments longer than 65536 will return an error.
```
shurcooL: "longer than 65536 **characters**" here as well.

github/issues_comments.go
```diff
@@ -100,8 +103,13 @@ func (s *IssuesService) GetComment(owner string, repo string, id int) (*IssueCom
...
+//
 // GitHub API docs: http://developer.github.com/v3/issues/comments/#create-a-comment
 func (s *IssuesService) CreateComment(owner string, repo string, number int, comment *IssueComment) (*IssueComment, *Response, error) {
+	if comment != nil && comment.Body != nil && len(*comment.Body) > maxCommentLen {
+		return nil, nil, fmt.Errorf("body is too long (maximum is %d characters)", maxCommentLen)
+	}
 	u := fmt.Sprintf("repos/%v/%v/issues/%d/comments", owner, repo, number)
```
shurcooL: Add a blank line separating the two section here as well, so it's consistent with the other two.

github/issues_comments.go
```diff
@@ -10,6 +10,9 @@ import (
 	"time"
 )
 
+// maxCommentLen is the maximum length of a comment that github will accept without a validation error.
+const maxCommentLen = 65536
```
shurcooL: Nitpick, but can we rename this to `maxCommentLength` or `maxCommentSize`? 

Having a trimmed word ("len") next to full words like "comment" looks weird.

See https://github.com/golang/go/blob/master/src/net/http/client.go#L549 for example from Go source code.

eparis:
I'm going to close this PR. I was just trying to test byte vs rune and was able to land a 30k 3 byte runes (90k bytes). So I decided to make sure it was working and a 160k byte 160k rune comment worked. Maybe github lifted the restriction I was hitting before? I'm using the same call that failed previously. In any case, no reason to code my broken logic into the library!

Issue #431, opened 2016-09-21
Where is the best layer to add throttling to avoid being marked for abuse?


dmitshur:
GitHub API has some guidelines to follow when accessing their API to avoid hitting abuse rate limit:

https://developer.github.com/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits

An application should follow those rules. Some of them can be enforced in a general/automated way via throttling, I think, and so they can be factored out into a library (rather than having to be done manually inside the application layer). Specifically:
1. > Make requests for a single user or client ID serially. Do not make requests for a single user or client ID concurrently.
   
   If we can assume a single `*github.Client` maps to a single user or client ID, then this can be done by using a mutex or semaphore or similar.
2. > If you're making a large number of `POST`, `PATCH`, `PUT`, or `DELETE` requests for a single user or client ID, wait at least one second between each request.
   
   Again, if we can assume a single `*github.Client` maps to a single user or client ID, then this can be done by keeping track of the last `POST`, `PATCH`, `PUT`, or `DELETE` request time, and sleeping if neccessary to make the next one happen at least one second later.
3. > When you have been limited, wait the number of seconds specified in the `Retry-After` response header.
   
   We can detect abuse rate limit response (it's documented [here](https://developer.github.com/v3/#abuse-rate-limits)), read the number of seconds specified in the `Retry-After` response header, and not allow making network requests until that time has passed (probably returning `*RateLimitError` error, or an abuse-specific version thereof, right away for all requests).

I see two places where this can be done:
1. Possibly inside go-github library. Inside `Client` itself. Probably as a configurable option.
2. Or outside of this library. As a higher-level wrapper layer around go-github library.

So far, go-github has been a relatively "low-level" API client, meaning it helps you out with type safety and provides structured output, but aside from that, it maps relatively closely to 1:1 to making HTTP requests to GitHub API.

It has taken some steps to be a smarter/higher-level client where it could be done so transparently, without blocking. For example, after #277, `RateLimitError` type was added, and after #347, go-github is smart enough to track rate limit reset time and not make network calls when the rate limit is known to still be exceeded, returning `RateLimitError` right away.

So far it has also been a "return-right-away" API, where none of the calls are artificially throttled on the client side, they just return an error if there's a rate limit problem. So it seems adding throttling to the go-github client itself may be out of place, and would belong in a higher-level wrapper around go-github.

[...]

PR #441, opened 2016-10-05
Add and detect AbuseRateLimitError.

shurcooL:
It's similar to `RateLimitError`, but it's a different type of rate limit error. It is documented at:

https://developer.github.com/v3/#abuse-rate-limits

Parse and include the `Retry-After` header value, if present. It is documented at:

https://developer.github.com/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits

According to GitHub support, its type is an integer value representing seconds.

Helps #431.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -494,6 +494,24 @@ func (r *RateLimitError) Error() string {
 		r.Response.StatusCode, r.Message, r.Rate.Reset.Time.Sub(time.Now()))
 }
 
+// AbuseRateLimitError occurs when GitHub returns 403 Forbidden response with the
+// "documentation_url" field value equal to "https://developer.github.com/v3#abuse-rate-limits".
+type AbuseRateLimitError struct {
+	Response *http.Response // HTTP response that caused this error
+	Message  string         `json:"message"` // error message
+
+	// RetryAfter is provided with some abuse rate limit errors. If present,
+	// it is the amount of time that the client should wait before retrying.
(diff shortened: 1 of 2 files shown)


shurcooL:
Some notes on this.

I'm implementing this based on GitHub documentation (and a clarifying/confirmation email I've sent to support@github.com). I've not had the opportunity to test it for real because I don't want to try to reproduce an abuse rate limit situation on purpose just for testing this PR.

github/github.go
```diff
@@ -564,6 +582,20 @@ func CheckResponse(r *http.Response) error {
 			Response: errorResponse.Response,
 			Message:  errorResponse.Message,
 		}
+	case r.StatusCode == http.StatusForbidden && errorResponse.DocumentationURL == "https://developer.github.com/v3#abuse-rate-limits":
```
shurcooL: The reason I matched based on the documentation_url value rather than a prefix of the message is because this seems more reliable.

I know there are multiple different types of abuse rate limit errors, and they have different message text. They may have the same prefix, but I don't know for sure what are all the possible messages. If the documentation_url value is "https://developer.github.com/v3#abuse-rate-limits" (note the "abuse" in the URL, it's not https://developer.github.com/v3/#rate-limiting), then it seems quite reasonable to expect this is an abuse rate limit error rather than something else.

If there are improvement suggestions, I'd be glad to hear them.

gmlewis:
I like it.  Thank you, @shurcooL!
Merging.

PR #442, opened 2016-10-05
Added license_template 

anshumanbh:
Added the license_template in the repository struct to be able to provide a license file when creating a repo


anshumanbh:
I signed it!

PR #443, opened 2016-10-05
added license_template to repo struct

anshumanbh:


shurcooL:
Is this new field referring to the `license_template` field mentioned in the Input section of repositories create endpoint here?

https://developer.github.com/v3/repos/#input

anshumanbh:
@shurcooL Yes, thats right!

While creating a new repo, we need to provide a license_template. If we don't do that, the repo still gets created but whenever you try to use the Github API to query anything, it says "the Git repository is empty". With the addition of this field in the repository struct, we don't get that message anymore and it is possible to use the Github API to query different methods.

github/repos.go
```diff
@@ -48,7 +48,8 @@ type Repository struct {
 	Permissions      *map[string]bool `json:"permissions,omitempty"`
 
 	// Only provided when using RepositoriesService.Get while in preview
-	License *License `json:"license,omitempty"`
+	License         *License `json:"license,omitempty"`
+	LicenseTemplate *string  `json:"license_template,omitempty"`
 
 	// Additional mutable fields when creating and editing a repository
```
shurcooL: > While creating a new repo, we need to provide a license_template. 

In that case, you should move the new `LicenseTemplate` field into _this_ section, it has all the other fields for creating/editing a repository. Or maybe a sub-section there at the end, since this field is relevant at create-time only, you can't use it when editing an existing repo.

It's not relevant to the `RepositoriesService.Get` section you've added it to currently.

Additionally, you can add `gitignore_template`, which is also missing, if you want. Source: https://developer.github.com/v3/repos/#input.

gmlewis:
LGTM.
Awaiting secondary LGTM before merging.

gmlewis:
@anshumanbh - if you take a look at this:
https://patch-diff.githubusercontent.com/raw/google/go-github/pull/443.patch
you can see different email addresses in the patch.
If you could please squash all your commits, that would be great.
Thanks.

anshumanbh:
@gmlewis Apologies for that. I switched laptops hence the error. I think it should be okay now

gmlewis:
No problem at all, @anshumanbh - it just makes googlebot happier.  😄 
Awaiting LGTM from @shurcooL.

gmlewis:
I can remove the blank line on merge.

gmlewis:
Thank you, @anshumanbh and @shurcooL!
Integrated in https://github.com/google/go-github/commit/1bc362c7737e51014af7299e016444b654095ad9

Issue #437, opened 2016-09-28
Add support for new preview rebase in the Pull Request Merge API and new Squash API enhancements


gmlewis:
See blog post for more information:
https://developer.github.com/changes/2016-09-26-pull-request-merge-api-update/

gsquire:
I would like to do this one, if nobody else has claimed it.

gmlewis:
SGTM, @gsquire - it is yours.  Thanks in advance!

Issue #376, opened 2016-06-17
versioning and tagging releases


willnorris:
We generally try not to have egregious breaking changes in this library, but we're also not afraid to do them when necessary (the latest example being #375).  We've gotten a couple of people asking for us to use semver on google/go-querystring#11, which we could also consider doing here.

That doesn't help us for #375, but would allow us to do future breaking changes at a version boundary.  I have no intention of changing the import path to use gopkg.in or anything like it, but it would still help set appropriate expectations for folks when they do upgrade to a new version.  And keeping a `CHANGELOG.md` file might be nice (although a bit of work to keep up to date) to point out new APIs that we support, etc.

@gmlewis @shurcooL what do you think?

parkr:
To me, vendoring makes this a moot issue - users can upgrade when they are ready and vendor the version they wish to use. This is ideal, however, and does not represent everyone's workflow. But it is something to suggest as people come pointing fingers about blame.

I had a package break due to the addition of a parameter to PullRequestsService.Merge. I fixed it and moved on. Go makes it very easy to deal with small breaking changes like that.

Moving the canonical source to gopkg.in is, in itself, a breaking change, right? Now that Go forces you to use the canonical import path?

gmlewis:
I suppose there are varying levels of supporting semantic versioning as well.
The bare minimum is to have a file somewhere in the repo with a version string in it.
A step up would have a compiled file with a const string in it that gets bumped.

This repo already has [`libraryVersion`](https://github.com/google/go-github/blob/master/github/github.go#L32) which appears to have remained at `0.1` since the beginning.

On other projects that I supported, maintaining a `CHANGELOG.md` was a pain and I always wondered if anyone ever really looked at it or found it useful.

All that said, I don't have really strong opinions either way.

My recommendation moving forward is to bump `libraryVersion` to 0.2 in #375 since it is a breaking change and let the commit message suffice for the change log. Let people vendor the library as they see fit.

willnorris:
> Moving the canonical source to gopkg.in is, in itself, a breaking change, right?

[...]

PR #446, opened 2016-10-10
Add support for merge method and repository squash settings.

gsquire:
This PR adds merge method support and squash API options for repository get and update.

I'm unsure of what a good unit test might be, so I can update this if need be.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -68,6 +68,7 @@ const (
 	mediaTypeReactionsPreview = "application/vnd.github.squirrel-girl-preview"
 
 	// https://developer.github.com/changes/2016-04-01-squash-api-preview/
+	// https://developer.github.com/changes/2016-09-26-pull-request-merge-api-update/
 	mediaTypeSquashPreview = "application/vnd.github.polaris-preview+json"
 
 	// https://developer.github.com/changes/2016-04-04-git-signing-api-preview/
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -253,12 +253,16 @@ type PullRequestMergeResult struct {
 
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
-	Squash bool
+	CommitTitle string
+
+	// The merge method to use. Possible values include: "merge", "squash", and "rebase" with the default being merge.
+	MergeMethod string
 }
 
 type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
-	Squash        *bool   `json:"squash,omitempty"`
+	CommitTitle   *string `json:"commit_title,omitempty"`
+	MergeMethod   *string `json:"merge_method,omitempty"`
 }
 
 // Merge a pull request (Merge Button™).
@@ -269,7 +273,8 @@ func (s *PullRequestsService) Merge(owner string, repo string, number int, commi
 
 	pullRequestBody := &pullRequestMergeRequest{CommitMessage: &commitMessage}
 	if options != nil {
-		pullRequestBody.Squash = &options.Squash
+		pullRequestBody.CommitTitle = &options.CommitTitle
+		pullRequestBody.MergeMethod = &options.MergeMethod
 	}
 	req, err := s.client.NewRequest("PUT", u, pullRequestBody)
 
--- a/github/repos.go
+++ b/github/repos.go
@@ -5,7 +5,10 @@
 
 package github
 
-import "fmt"
+import (
+	"fmt"
+	"strings"
+)
 
 // RepositoriesService handles communication with the repository related
 // methods of the GitHub API.
@@ -46,6 +49,9 @@ type Repository struct {
 	Source           *Repository      `json:"source,omitempty"`
 	Organization     *Organization    `json:"organization,omitempty"`
 	Permissions      *map[string]bool `json:"permissions,omitempty"`
+	AllowRebaseMerge *bool            `json:"allow_rebase_merge,omitempty"`
+	AllowSquashMerge *bool            `json:"allow_squash_merge,omitempty"`
+	AllowMergeCommit *bool            `json:"allow_merge_commit,omitempty"`
 
 	// Only provided when using RepositoriesService.Get while in preview
 	License *License `json:"license,omitempty"`
@@ -286,7 +292,8 @@ func (s *RepositoriesService) Get(owner, repo string) (*Repository, *Response, e
 
 	// TODO: remove custom Accept header when the license support fully launches
 	// https://developer.github.com/v3/licenses/#get-a-repositorys-license
-	req.Header.Set("Accept", mediaTypeLicensesPreview)
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeSquashPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	repository := new(Repository)
 	resp, err := s.client.Do(req, repository)
@@ -329,6 +336,9 @@ func (s *RepositoriesService) Edit(owner, repo string, repository *Repository) (
 	if err != nil {
 		return nil, nil, err
 	}
+
+	// TODO: Remove this preview header after API is fully vetted.
+	req.Header.Add("Accept", mediaTypeSquashPreview)
 
 	r := new(Repository)
 	resp, err := s.client.Do(req, r)
--- a/github/pulls_test.go
+++ b/github/pulls_test.go
@@ -363,7 +363,7 @@ func TestPullRequestsService_Merge(t *testing.T) {
 			}`)
 	})
 
-	options := &PullRequestOptions{Squash: true}
+	options := &PullRequestOptions{MergeMethod: "rebase"}
 	merge, _, err := client.PullRequests.Merge("o", "r", 1, "merging pull request", options)
 	if err != nil {
 		t.Errorf("PullRequests.Merge returned error: %v", err)
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 	"testing"
 )
 
@@ -225,9 +226,10 @@ func TestRepositoriesService_Get(t *testing.T) {
 	setup()
 	defer teardown()
 
+	acceptHeader := []string{mediaTypeLicensesPreview, mediaTypeSquashPreview}
 	mux.HandleFunc("/repos/o/r", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		testHeader(t, r, "Accept", mediaTypeLicensesPreview)
+		testHeader(t, r, "Accept", strings.Join(acceptHeader, ", "))
 		fmt.Fprint(w, `{"id":1,"name":"n","description":"d","owner":{"login":"l"},"license":{"key":"mit"}}`)
 	})
 


github/pulls.go
```diff
@@ -251,12 +251,14 @@ type PullRequestMergeResult struct {
 
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
-	Squash bool
+	Squash      bool
```
gmlewis: It looks to me like `Squash` is no longer supported and can be removed..
https://developer.github.com/changes/2016-04-01-squash-api-preview/ seems to confirm, although it is not explicit.

github/pulls.go
```diff
@@ -251,12 +251,14 @@ type PullRequestMergeResult struct {
...
+	MergeMethod string
```
gmlewis: Can you please document this field?
e.g.: `Merge method to use. Possible values are: "merge", "squash", "rebase". Default is "merge".`

github/pulls.go
```diff
@@ -251,12 +251,14 @@ type PullRequestMergeResult struct {
 
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
```
gmlewis: Could you please add `CommitTitle` to this struct?
https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

github/pulls.go
```diff
@@ -251,12 +251,14 @@ type PullRequestMergeResult struct {
...
-	Squash bool
+	Squash      bool
+	MergeMethod string
 }
 
 type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
 	Squash        *bool   `json:"squash,omitempty"`
```
gmlewis: You will need to remove `Squash` here, too.

github/pulls.go
```diff
@@ -251,12 +251,14 @@ type PullRequestMergeResult struct {
 
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
-	Squash bool
+	Squash      bool
+	MergeMethod string
 }
 
 type pullRequestMergeRequest struct {
```
gmlewis: You will need to add `CommitTitle` here too.

github/pulls.go
```diff
@@ -268,10 +270,12 @@ func (s *PullRequestsService) Merge(owner string, repo string, number int, commi
 	pullRequestBody := &pullRequestMergeRequest{CommitMessage: &commitMessage}
 	if options != nil {
```
gmlewis: And update this section too.

github/repos.go
```diff
@@ -283,7 +286,9 @@ func (s *RepositoriesService) Get(owner, repo string) (*Repository, *Response, e
 
 	// TODO: remove custom Accept header when the license support fully launches
 	// https://developer.github.com/v3/licenses/#get-a-repositorys-license
+	// https://developer.github.com/changes/2016-09-26-pull-request-merge-api-update
```
gmlewis: The blog posts are usually referenced in the github.go file with the preview tags - you could add this to the `polaris` one.

github/pulls.go
gsquire: Yes, I certainly missed that in the blog post.

github/pulls.go
gsquire: I agree, it looks to be deprecated.

github/repos.go
gsquire: Sure, will do.

github/pulls.go
```diff
@@ -251,12 +251,16 @@ type PullRequestMergeResult struct {
...
+	CommitTitle string
+	MergeMethod string
 }
 
 type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
-	Squash        *bool   `json:"squash,omitempty"`
+	CommitTitle   *string `json:"commit_title,omitempty"`
+
+	// The merge method to use. Possible values include: "merge", "squash", and "rebase" with the default being merge.
```
gmlewis: While this is good, I think it would be much better to move this to the exported struct above (`PullRequestOptions`) so that the auto-generated `godoc`s will read better.

gmlewis:
LGTM after a minor `godoc` tweak.
Awaiting secondary LGTM before merging.

github/pulls.go
gsquire: No problem, should have caught that.

gsquire:
Thank you, I will collapse my commits when it gets signed off on.

gsquire:
I rebased onto master and squashed.

shurcooL:
We should add a "Resolves #437." or equivalent message to the commit message, so that issue gets closed when this is merged. But @gmlewis can do that when merging, so it's not a big deal.

github/repos.go
```diff
@@ -284,6 +287,7 @@ func (s *RepositoriesService) Get(owner, repo string) (*Repository, *Response, e
 	// TODO: remove custom Accept header when the license support fully launches
 	// https://developer.github.com/v3/licenses/#get-a-repositorys-license
 	req.Header.Set("Accept", mediaTypeLicensesPreview)
+	req.Header.Add("Accept", mediaTypeSquashPreview)
```
shurcooL: Has anyone tested that multiple `Accept` headers actually work by combining both preview APIs, instead of a random one being used?

I'm asking because this is the very first time I see it done in this library, it has never been done before as far as I know.

I tried reading https://developer.github.com/v3/media/ and it doesn't seem to say that multiple Accept headers are the way to go (neither does it say it shouldn't be done).

github/github.go
```diff
@@ -68,6 +68,7 @@ const (
 	mediaTypeReactionsPreview = "application/vnd.github.squirrel-girl-preview"
 
 	// https://developer.github.com/changes/2016-04-01-squash-api-preview/
+	// https://developer.github.com/changes/2016-09-26-pull-request-merge-api-update
```
shurcooL: Nitpick, but the missing trailing slash is inconsistent.

github/repos.go
shurcooL: Yeah, pretty sure that 2 accept headers don't work:

```bash
$ curl https://api.github.com/users/technoweenie -I -s -H "Accept: application/vnd.github.drax-preview+json" -H "Accept: application/vnd.github.polaris-preview+json" | grep X-GitHub-Media-Type
X-GitHub-Media-Type: github.drax-preview; format=json
```

Only one seems to have an effect.

I'm not sure if it's even possible to opt into two preview APIs in the same request... Any thoughts on this?

gsquire: I didn't know if that would matter or not. I assumed that you would need both since this API request now has two preview features. What should happen here then?

gsquire:
@shurcooL I have a "Closes" line in my commit message already.

gsquire:
Rebased onto master again.

github/repos.go
gmlewis: Sorry for the delay.  I will ask GitHub support and see what they recommend.

gsquire: It's no rush, I didn't know what the course of action was for this. Thanks!

shurcooL: > I will ask GitHub support and see what they recommend.

Good idea, thanks for doing that!

If it turns out that it's not possible to combine multiple beta APIs in same request, we might have to make 2 requests (each with a different beta API), or just pick 1 beta that's more important and remove the other. Or wait until at least one of them is merged into main v3 API, then use the remaining beta header.

gmlewis: It turns out that my hunch for using comma-separated values was correct:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

From GitHub support:
> Example:
> Accept: application/vnd.github.mockingbird-preview, application/vnd.github.polaris-preview

Of course, this will need to be tested.

gmlewis:
Hmmm... it looks like I dropped the ball on this one... I'm sorry about that... Merging.

gmlewis:
Ah, yes... thanks, @shurcooL for the reminder.  This was held up on the header issue.
I haven't had a chance to test the comma-separated header yet.  Hopefully today.
I'll test before I merge.

eparis:
This breaks the existing 'squash' support. Our API is changing.

I don't even know if squash is still in the github API though. Or if we care (I don't).

I'll rebase my work to support SHA on top of this. Great!

gmlewis:
SGTM.  Yes, "squash" is still supported in the GitHub API via the `MergeMethod`.
Thank you, @eparis!

eparis:
@gmlewis but our API is breaking in this PR is what I was trying to say, not that I care, just pointing it out in case anyone else wanted to complain that we know this will cause code to be unable to compile.

gmlewis:
Ah!  I didn't understand that, @eparis. Can you please copy/paste the error message(s)?

github/pulls.go
```diff
@@ -253,12 +253,16 @@ type PullRequestMergeResult struct {
 
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
-	Squash bool
+	CommitTitle string
```
eparis: Shouldn't these be *string? Or should we check for `!= ""` before we assign to pullRequestMergeRequest? Otherwise setting any PullRequestOptions will result in all options be set (to explicit `""`)

github/pulls.go
```diff
@@ -253,12 +253,16 @@ type PullRequestMergeResult struct {
 
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
-	Squash bool
```
eparis: by removing this any user of our library who is doing
```
ops := &PullRequestOptions {
    squash: true,
}
```
will be unable to compile. We "could" keep supporting them by automatically turning `Squash` into a `MergeMethod = "squash"` if we so chose. We don't have to. If github breaks the API I don't know if we really need to be forward compatible with our users....

gmlewis: Yes, GitHub is changing their API.
Please read: https://developer.github.com/changes/2016-04-01-squash-api-preview/

eparis: But we don't have to. We may decide we want to, and that's fine. But it is easy to write:
```
type PullRequestOptions struct {
    Squash bool
    MergeMethod string
}
```
And then inside https://github.com/google/go-github/pull/446/files#diff-aafac035cce1f6a65b02566c5345b805R276
```
if options.Squash {
    mm := "squash"
    pullRequestBody.MergeMethod = &mm
}
```
Just pointing it out. I'm actually ok dropping OUR legacy API and break compile for our users. But its also pretty easy to be a friendly library if that's the path we wish to take.

gmlewis: Hmmm... @willnorris and @shurcooL - what are your thoughts about breaking the current (preview) API involving "squashes" when the GitHub Preview API itself is changing? Do we match the GitHub API or do we attempt to retain a "more friendly" API toward squashes?
See: https://developer.github.com/changes/2016-04-01-squash-api-preview/ for more info.

To remain consistent with the rest of this package, I'm leaning toward matching the GitHub API as close as possible to not create any surprises for users. But if you feel we should depart from this consistency and make the API friendlier for squashing, I'm not adverse to that either.

gsquire: I just caught up on this thread. I'll standby until we settle on what this API should have. Please feel free to ping me so I can make the appropriate changes.

Thanks for all of the comments!

shurcooL: > Hmmm... @willnorris and @shurcooL - what are your thoughts about breaking the current (preview) API involving "squashes" when the GitHub Preview API itself is changing? Do we match the GitHub API or do we attempt to retain a "more friendly" API toward squashes?

We've had this discussion in the past. Let me try to find it.

I think it's [this comment](https://github.com/google/go-github/issues/376#issuecomment-227021470) and its thread that I had in mind.

Basically, my POV is that we're welcome to make breaking API changes to endpoints that are in preview. We absolutely should, that's the whole point of API preview period, it lets GitHub iterate on the API and this library should follow. Using the very first iteration is going to be counter-productive for users after the API exits preview period.

@gmlewis I'm not sure if your question is about whether it's okay to break go-github APIs for preview endpoints (IMO it is), or if it's about whether we should deviate from GitHub API in what go-github provides (IMO we should try to stay close, but still be idiomatic Go, no need to compromise on usability/friendliness).

> To remain consistent with the rest of this package, I'm leaning toward matching the GitHub API as close as possible to not create any surprises for users. But if you feel we should depart from this consistency and make the API friendlier for squashing, I'm not adverse to that either.

I think your question is about the latter. Can you elaborate on what 2 alternatives you're considering here?

Also, we should open an issue that this PR resolves and continue the discussion there. It's hard to see this "outdated comment" in this PR.

github/pulls.go
gmlewis: As far as I can tell from https://developer.github.com/changes/2016-09-26-pull-request-merge-api-update/ and https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button
this field is mandatory, although below we make it omitempty.
Yes, I'm thinking that if we just test for != "" before assigning to `pullRequestMergeRequest` below, that this would be safe.

github/pulls.go
gmlewis: Creating new issue for discussion, now that GitHub no longer appears to be under DDOS attack.

gsquire:
@gmlewis I hope this PR hasn't gone stale. Based on the issue you wrote discussing how we should handle the merge method, the code is already set up to do that.

Please let me know if there are any more items I should address in the meantime.

gmlewis:
Thanks for the ping, @gsquire, and I apologize for the delay!
AFAICT, this PR is ready to be merged, and I'll go ahead and do that now.

gmlewis:
Integrated in https://github.com/google/go-github/commit/575b0b37560d9485f6d3bcb220a9111495b9c248.

gsquire:
@gmlewis It's no problem, thanks for your patience. 👍

PR #451, opened 2016-10-12
Add support for PullRequestReviewEvent webhook.

### LINKED ISSUE #450: Add support for new PullRequestReviewEvent webhook
https://developer.github.com/changes/2016-10-11-pull-request-review-webhooks/

Example: https://gist.github.com/parkr/25b9fcb88b771eb73e7bf246d647c141

parkr:
Fixes #450. This is documented here: https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent

/cc @willnorris @gmlewis for review.

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -327,6 +327,26 @@ type PullRequestEvent struct {
 	Sender  *User       `json:"sender,omitempty"`
 }
 
+// PullRequestReviewEvent is triggered when a review is submitted on a pull
+// request.
+// The Webhook event name is "pull_request_review".
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent
+type PullRequestReviewEvent struct {
+	// Action is always "submitted".
+	Action      *string            `json:"action,omitempty"`
+	Review      *PullRequestReview `json:"review,omitempty"`
+	PullRequest *PullRequest       `json:"pull_request,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+
+	// The following field is only present when the webhook is triggered on
+	// a repository belonging to an organization.
+	Organization *Organization `json:"organization,omitempty"`
+}
+
 // PullRequestReviewCommentEvent is triggered when a comment is created on a
 // portion of the unified diff of a pull request.
 // The Webhook event name is "pull_request_review_comment".
--- a/github/messages.go
+++ b/github/messages.go
@@ -55,6 +55,7 @@ var (
 		"milestone":                             "MilestoneEvent",
 		"page_build":                            "PageBuildEvent",
 		"public":                                "PublicEvent",
+		"pull_request_review":                   "PullRequestReviewEvent",
 		"pull_request_review_comment":           "PullRequestReviewCommentEvent",
 		"pull_request":                          "PullRequestEvent",
 		"push":                                  "PushEvent",
--- a/github/pulls_reviews.go
+++ b/github/pulls_reviews.go
@@ -1,0 +1,14 @@
+package github
+
+import "time"
+
+// PullRequestReview represents a review of a pull request.
+type PullRequestReview struct {
+	ID          *int       `json:"id,omitempty"`
+	User        *User      `json:"user,omitempty"`
+	Body        *string    `json:"body,omitempty"`
+	SubmittedAt *time.Time `json:"submitted_at,omitempty"`
+
+	// State can be "approved", "rejected", or "commented".
+	State *string `json:"state,omitempty"`
+}
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -69,6 +69,8 @@ func (e *Event) Payload() (payload interface{}) {
 		payload = &PublicEvent{}
 	case "PullRequestEvent":
 		payload = &PullRequestEvent{}
+	case "PullRequestReviewEvent":
+		payload = &PullRequestReviewEvent{}
 	case "PullRequestReviewCommentEvent":
 		payload = &PullRequestReviewCommentEvent{}
 	case "PushEvent":
--- a/github/messages_test.go
+++ b/github/messages_test.go
@@ -153,6 +153,10 @@ func TestParseWebHook(t *testing.T) {
 			messageType: "pull_request",
 		},
 		{
+			payload:     &PullRequestReviewEvent{},
+			messageType: "pull_request_review",
+		},
+		{
 			payload:     &PullRequestReviewCommentEvent{},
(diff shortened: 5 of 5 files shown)


github/pulls_reviews.go
```diff
@@ -0,0 +1,14 @@
+package github
+
+import "time"
+
+// No docs are out for this quite yet.
```
shurcooL: This comment is not helpful. How about:

```Go
// PullRequestReview represents a review of a pull request.
type PullRequestReview struct {
```

github/event_types.go
```diff
@@ -294,6 +294,26 @@ type PullRequestEvent struct {
...
 
+// PullRequestReviewEvent is triggered when a review is submitted on a pull
+// request.
+// The Webhook event name is "pull_request_review".
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent
+type PullRequestReviewEvent struct {
+	// Action is always "submitted".
+	Action            *string            `json:"action,omitempty"`
+	PullRequestReview *PullRequestReview `json:"review,omitempty"`
```
shurcooL: Suggestion. The JSON field is named `review`, it seems better to name the Go field the same to match:

```Go
Review *PullRequestReview `json:"review,omitempty"`
```

github/pulls_reviews.go
```diff
@@ -0,0 +1,14 @@
...
+type PullRequestReview struct {
+	ID          *int       `json:"id,omitempty"`
+	User        *User      `json:"user,omitempty"`
+	Body        *string    `json:"body,omitempty"`
+	SubmittedAt *time.Time `json:"submitted_at,omitempty"`
+
+	// State can be "approved", "rejected", or "commented".
+	State *string `json:"state,omitempty"`
+}
```
shurcooL: Suggestion. This single struct doesn't really deserve an entirely new file. How do you feel about moving it under `PullRequestReviewEvent`, which is the first (and currently only) place it's used?

If it ever becomes used in many other places and there's much more review related code, then moving it into a new separate file would make sense. It seems too early to do that now.

github/pulls_reviews.go
```diff
@@ -0,0 +1,14 @@
+package github
+
+import "time"
+
+// No docs are out for this quite yet.
+type PullRequestReview struct {
```
shurcooL: Nothing is currently using `Review` name. Have you considered using that, instead of `PullRequestReview`? Since that's what the JSON field is called. Plus it's documented as "The review that was affected", rather "The pull request review that was affected".

Or do you expect that there may be other types of reviews in the future, so it's better to proactively name this `PullRequestReview` to avoid future collisions?

shurcooL:
Friendly ping @parkr.

I've left some review comments above, did you see them? Are you interested in addressing them (either applying changes, or letting me know if you disagree or don't think something should be done)? If you prefer not to work on this PR anymore, just let us know and we'll try to take it from here.

Thanks for making getting it this far already!

shurcooL:
> Has the REST API for reviews been incorporated?

Sorry, I'm not sure which REST API for reviews you're referring to. Is it a new one? If so, can you link to its https://developer.github.com/changes/ blog post so I can cross-check?

Or maybe someone else knows.

parkr:
Ah, there is usually parity between the REST API and the webhooks for various features. It looks like the REST API for pull request reviews doesn't exist. I'll update per your comments.

github/pulls_reviews.go
parkr: @shurcooL I considered using just `Review`. I did not go with it because these reviews are strictly for pull requests and the namespacing conventions I grokked here seemed to indicate that it would be most appropriate to be reasonably specific about what the review affects. Reviews are scoped to pull requests, therefore `PullRequestReview`. Does that seem reasonable?

github/pulls_reviews.go
parkr: @shurcooL I put it in its own file in anticipation of the REST API being released (there's an API for nearly every feature on the site). My understanding is that this is where the eventual REST API code will go. I want a struct that is unified between the webhooks and REST API, and `github/event_types.go` is already nearly 550 lines long.

parkr:
Comments addressed. /cc @gmlewis @willnorris and @shurcooL for review.

github/pulls_reviews.go
shurcooL: Generally, I would advise against doing something suboptimal now in anticipation of something happening in the future. It's usually better to do the bare minimum for today's needs, and adjust in future as needed. It's very easy to move something into a new file later if there ends up being more stuff later.

But, it's not a big deal, and I see you're at GitHub, so I imagine you have a better ability to anticipate changes. So it's okay with me.

github/messages.go
```diff
@@ -56,6 +56,7 @@ var (
 		"page_build":                            "PageBuildEvent",
 		"public":                                "PublicEvent",
 		"pull_request_review_comment":           "PullRequestReviewCommentEvent",
+		"pull_request_review":                   "PullRequestReviewEvent",
```
shurcooL: For consistency with elsewhere, you can move `pull_request_review` on top of `pull_request_review_comment`.

github/event_types.go
```diff
@@ -327,6 +327,26 @@ type PullRequestEvent struct {
...
+	// Action is always "submitted".
+	Action      *string            `json:"action,omitempty"`
+	Review      *PullRequestReview `json:"review,omitempty"`
+	PullRequest *PullRequest       `json:"pull_request,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	Repo   *Repository `json:"repository,omitempty"`
+	Sender *User       `json:"sender,omitempty"`
+
+	// The follow field is only present when the webhook is triggered on a
```
shurcooL: Did you mean "The follow***ing*** field"?

github/messages.go
parkr: Should `pull_request` then go above `pull_request_review`?

github/event_types.go
parkr: Yep! Great catch.

github/messages.go
shurcooL: I suppose so. That seems to be the order in other files (activity_events.go, event_types.go, messages_test.go, etc.), and it's alphabetical ordering, right?

parkr:
Is there anything I'm missing here?

shurcooL:
Not as far as I can tell. This is just waiting for @gmlewis or @willnorris to review and merge.

gmlewis:
Sorry for the delay. LGTM. Merging.

gmlewis:
Included in https://github.com/google/go-github/commit/511bc5f4a8cb442d34f9eaecb7cb2372fa706965

parkr:
thank you, @gmlewis!

PR #456, opened 2016-10-20
Support SHA in merges

eparis:
Adds support for SHA

https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -281,6 +281,7 @@ type PullRequestMergeResult struct {
 // PullRequestOptions lets you define how a pull request will be merged.
 type PullRequestOptions struct {
 	CommitTitle string
+	SHA         string // SHA that pull request head must match to allow merge. (Optional.)
 
 	// The merge method to use. Possible values include: "merge", "squash", and "rebase" with the default being merge.
 	MergeMethod string
@@ -290,6 +291,7 @@ type pullRequestMergeRequest struct {
 	CommitMessage *string `json:"commit_message"`
 	CommitTitle   *string `json:"commit_title,omitempty"`
 	MergeMethod   *string `json:"merge_method,omitempty"`
+	SHA           *string `json:"sha,omitempty"`
 }
 
 // Merge a pull request (Merge Button™).
@@ -302,6 +304,7 @@ func (s *PullRequestsService) Merge(owner string, repo string, number int, commi
 	if options != nil {
 		pullRequestBody.CommitTitle = &options.CommitTitle
 		pullRequestBody.MergeMethod = &options.MergeMethod
+		pullRequestBody.SHA = &options.SHA
 	}
 	req, err := s.client.NewRequest("PUT", u, pullRequestBody)
 


eparis:
This is untested. I am going to work on testing now, but wanted to post for review.

gmlewis:
Please take a look at https://github.com/google/go-github/pull/446 before continuing.
I'm not sure why I haven't merged that one yet... I'm going to go take a closer look.

shurcooL:
> I'm not sure why I haven't merged that one yet... I'm going to go take a closer look.

It's because there's an overlap of 2 beta APIs and we're not sure if it's possible to do that in a single request. See https://github.com/google/go-github/pull/446#discussion_r84140678.

But it looks like there's a resolution there now.

gmlewis:
Ah, thank you, @shurcooL ! It is not obvious from a quick scan of the page... I wish the most recent comments were at the bottom of the page.  :-(

gsquire:
@eparis I just realized you are using my PR as a base for yours. We can close mine and make the changes on yours since there is overlap.

How does that sound?

eparis:
I'd be happy letting you finish your hard work and I'll rebase when you are done. I'm adding .1% on top of your excellent work. If anything you should roll this tiny addition into yours!

shurcooL:
> If anything you should roll this tiny addition into yours!

You can make a PR into his PR lol. Do a "compare across forks" when creating the PR, then set the base repo as https://github.com/gsquire/go-github, and the base branch as [`api-updates`](https://github.com/gsquire/go-github/commits/api-updates).

shurcooL:
Now that #446 has been merged, I think this PR needs to be rebased on top.

github/pulls.go
```diff
@@ -275,6 +277,9 @@ func (s *PullRequestsService) Merge(owner string, repo string, number int, commi
 	if options != nil {
 		pullRequestBody.CommitTitle = &options.CommitTitle
 		pullRequestBody.MergeMethod = &options.MergeMethod
+		if len(options.SHA) != 0 {
```
shurcooL: Style suggestion, I think writing this as `if options.SHA != "" {` would be more readable. See my rationale [here](https://dmitri.shuralyov.com/idiomatic-go#empty-string-check). Feel free to change it if you agree.

eparis:
Updated for style.
Added comment.
Also made CommitTitle and MergeMethod optional. (Since github support tells me they are optional)

gmlewis:
Thanks, @eparis!

Just FYI... the way that `CommitTitle` and `MergeMethod` were written before already made them optional.

Specifically, because each field in `pullRequestBody` has `omitempty`, if the string is `""` then it will not be sent by the JSON encoder in the request. For more details, see: https://godoc.org/encoding/json#Marshal

So, taking the address of the empty string is identical to not setting the field at all, since both values resolve to the "zero value" of a string.

Thus, I recommend that you revert your changes to `CommitTitle` and `MergeMethod` and follow the same pattern.  If you feel like it would be beneficial to other developers looking at the code, you could add a comment that those optional fields will not be sent if they are empty strings due to the `omitempty`.  Your call.

eparis:
done, fixed.

gmlewis:
Very nice!  Thanks again, @eparis, and thank you, @shurcooL for the reviews.
LGTM.
Merging.

shurcooL:
> Just FYI... the way that `CommitTitle` and `MergeMethod` were written before already made them optional.
>
> Specifically, because each field in `pullRequestBody` has `omitempty`, if the string is `""` then it will not be sent by the JSON encoder in the request. For more details, see: https://godoc.org/encoding/json#Marshal
>
> So, taking the address of the empty string is identical to not setting the field at all, since both values resolve to the "zero value" of a string.

That isn't true.

> Struct values encode as JSON objects. Each exported struct field becomes a member of the object unless
>
>     - the field's tag is "-", or
>     - the field is empty and its tag specifies the "omitempty" option.
>
> The empty values are false, 0, any nil pointer or interface value, and any array, slice, map, or string of length zero.

What you said would apply if `PullRequestOptions` type was being JSON encoded, but that's not the case. `pullRequestBody` has type `pullRequestMergeRequest`, and that contains:

```
SHA *string `json:"sha,omitempty"`
```

Notice that `SHA` is a _pointer_ to a string. Its zero value is `nil` and any other non-nil pointer value is a non-zero value.

So, if `SHA` is set to the address of an empty string, it still gets encoded.

See a [playground sample](https://play.golang.org/p/NDkgDGybWJ).

So this `if` statement is needed to be kept:

```Go
if options.SHA != "" {
	pullRequestBody.SHA = &options.SHA
}
```

Otherwise `SHA` will always be included, even if its the empty string.

As an aside, it's really sad how in GitHub PRs you lose access to previous versions of the PR when the branch is force pushed to. In Gerrit, it's always possible to see previous patch sets, which is very helpful in a situation like this.

gmlewis:
Ah, wow... Today I Learned!
Thank you, @shurcooL!

To recreate the history, he had changed the other two options to also check for `""`.

Now the question becomes... is what I merged incorrect?
I would think that a `SHA` value of `""` would behave identically to not passing it... but I could be wrong there too.

With this newly-gained knowledge on my part, should the other two options also include this `!= ""` check before setting them?

shurcooL:
> Now the question becomes... is what I merged incorrect?

I believe so. With code as is in master now, SHA is documented to be optional but in fact we send it always if `options *PullRequestOptions` provided is not nil, regardless of options.SHA is is empty string or not.

A simple fix is to change:

```Go
pullRequestBody.SHA = &options.SHA
```

To:

```Go
if options.SHA != "" {
	pullRequestBody.SHA = &options.SHA
}
```

> I would think that a `SHA` value of `""` would behave identically to not passing it... but I could be wrong there too.

Sorry, you'll have to be more specific. There's at least 3 "SHA" values you could be referring to. One in `PullRequestOptions` type, another in `pullRequestMergeRequest` type, or the JSON field accepted by GitHub API.

> With this newly-gained knowledge on my part, should the other two options also include this != "" check before setting them?

They're not optional (are they?), so I think they're fine as they are.

shurcooL:
@gmlewis Can you open a new issue and describe which fields are meant to be optional, then we can make a PR to resolve it. Instead of continuing the conversation here, we can link to that issue.

gmlewis:
Issue created: #500.

PR #469, opened 2016-11-05
changed updatedAt json to correct json value

lucasalcantara:


shurcooL:
LGTM. Thanks!

shurcooL:
For posterity, in case someone wants to write a linter rule that would catch this kind of bug, the latest `vet` should catch this type of issue, as of https://go-review.googlesource.com/#/c/16704/ (thanks to @dominikh for finding that CL).

However, it's not a part of Go 1.7, it will be a part of Go 1.8.

We run `vet` in CI, but it's not a critical failure for `tip`, and we didn't notice this. But it was caught there. See https://travis-ci.org/google/go-github/builds/173415020 and the [tip build](https://travis-ci.org/google/go-github/jobs/173415024):

```
$ go tool vet .
github/repos_pages.go:33: struct field UpdatedAt repeats json tag "created_at" also at github/repos_pages.go:32
```

lucasalcantara:
@gmlewis Can you take a look as well?

gmlewis:
Excellent!  Thank you, @lucasalcantara and thank you @shurcooL for reviewing!
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/f561108eef042b61588e174172c8c942ba0d57a4.

PR #470, opened 2016-11-08
repos_collaborators: add an option for adding collaborators directly 

mehmetalisavas:
```
To send an invitation to a collaborator rather than directly adding them, you must provide a custom media type in the Accept header:
```
https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator

Header is set for everytime. Its not possible to add a user to the repository as a collaborators without invitation. With this option we can add a user direclty.

### DIFF
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -60,6 +60,15 @@ type RepositoryAddCollaboratorOptions struct {
 	//
 	// Default value is "push".  This option is only valid for organization-owned repositories.
 	Permission string `json:"permission,omitempty"`
+
+	// AddDirectly specifies the invitations will be sent or not
+	// To send an invitation to a collaborator rather than directly adding them,
+	// you must provide a custom media type in the Accept header
+	// default value is "false"
+	//
+	// To add a user as a collaborator directly into a repository
+	// set AddDirectly value "true"
+	AddDirectly bool `json:"-"`
 }
 
 // AddCollaborator adds the specified Github user as collaborator to the given repo.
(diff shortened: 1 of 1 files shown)


mehmetalisavas:
Signed it

shurcooL:
Hmm, I suspect that the ability to add people directly is going away (to be replaced with invitations that need to be accepted by user, to avoid abuse).

However, the GitHub docs don't say anything explicit about that, so this might be okay. Curious what others think about this.

mehmetalisavas:
Currently, its not possible without sending invitation to user when i want to add the user as a collaborator to the repository.
There is 2 document part about issue;
https://developer.github.com/changes/2016-06-14-repository-invitations/
https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator

Besides of it i will update comments to be more clear

gmlewis:
So if I understand correctly, you are saying that when this API comes out of Preview period, there will be no way to add a collaborator without sending an invite?  That seems rather odd.  This will require more investigation and/or communication with GitHub Tech Support.

shurcooL:
> when this API comes out of Preview period, there will be no way to add a collaborator without sending an invite? That seems rather odd.

That seems expected to me. As I understand, the whole reason GitHub is moving towards sending invites that need to be accepted rather than outright adding people is to solve a spam problem. E.g., a spammer creates a repo they want many to see, and gives Linus Torvalds or someone with many followers push rights to it, and that repo shows up in the activity feed of everyone who follows Linus Torvalds.

Also, being made a collaborator creates responsibility, so sending an invite seems fair.

So it seems expected that once the API preview period is done, the old way of adding people as collaborators without their granted permission would no longer be possible.

Unless I'm missing something.

gmlewis:
Ah, OK, then.  Since that is the direction that GitHub is moving, and our goal with this repo is to stay current as much as possible with all Preview APIs, then I think we do not want to introduce a change like this. Agree to close?

shurcooL:
Agree to close.

Thanks @mehmetalisavas for contributing, but this is not a change we can merge.

PR #472, opened 2016-11-12
Update License struct and add new RepositoryLicense struct

### LINKED ISSUE #471: Licenses struct is out of date
The `Licenses` struct has diverged from the latest returned by the GitHub API (https://developer.github.com/v3/licenses/) and should be updated to reflect the latest state.

nmiyake:
Update structs to reflect the new reponses returned from GitHub.

Fixes #471

### DIFF
--- a/github/licenses.go
+++ b/github/licenses.go
@@ -10,8 +10,29 @@ import "fmt"
 // LicensesService handles communication with the license related
 // methods of the GitHub API.
 //
-// GitHub API docs: http://developer.github.com/v3/pulls/
+// GitHub API docs: https://developer.github.com/v3/licenses/
 type LicensesService service
+
+// RepositoryLicense represents the license for a repository.
+type RepositoryLicense struct {
+	Name *string `json:"name,omitempty"`
+	Path *string `json:"path,omitempty"`
+
+	SHA         *string  `json:"sha,omitempty"`
+	Size        *int     `json:"size,omitempty"`
+	URL         *string  `json:"url,omitempty"`
+	HTMLURL     *string  `json:"html_url,omitempty"`
+	GitURL      *string  `json:"git_url,omitempty"`
+	DownloadURL *string  `json:"download_url,omitempty"`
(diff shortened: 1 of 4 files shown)


github/licenses.go
```diff
@@ -10,23 +10,49 @@ import "fmt"
 // LicensesService handles communication with the license related
 // methods of the GitHub API.
 //
-// GitHub API docs: http://developer.github.com/v3/pulls/
+// GitHub API docs: http://developer.github.com/v3/licenses/
```
shurcooL: Mind changing the scheme to https here?

shurcooL:
LGTM. Thanks for reporting and fixing this issue @nmiyake!

nmiyake:
Great, thanks @shurcooL . Any ETA on when it might get merged?

shurcooL:
As soon as @gmlewis or @willnorris or someone else reviews it and gives it a second LGTM.

github/licenses.go
```diff
@@ -10,23 +10,49 @@ import "fmt"
...
+	SHA         *string `json:"sha,omitempty"`
+	Size        *int    `json:"size,omitempty"`
+	URL         *string `json:"url,omitempty"`
+	HTMLURL     *string `json:"html_url,omitempty"`
+	GitURL      *string `json:"git_url,omitempty"`
+	DownloadURL *string `json:"download_url,omitempty"`
+	Type        *string `json:"type,omitempty"`
+	Content     *string `json:"content,omitempty"`
+	Encoding    *string `json:"encoding,omitempty"`
+	Links       *struct {
```
willnorris: The Links structure simply duplicates the various *URL fields.  I don't think we've included it on any other structs in go-github, so probably doesn't make sense here either.

nmiyake: Yes, the only other place I see it is in the `Feeds` struct and it has more pertinent information there -- fine with removing this. Will do so and update PR.

nmiyake:
Updated PR to address feedback (removed the `_links` field)

Issue #176, opened 2015-03-31
json unmarshal error from ListIssueEventsForRepository


kylebrandt:
Maybe I'm doing something wrong?

```
package main

import (
    "fmt"

    "github.com/google/go-github/github"

    "code.google.com/p/goauth2/oauth"
)

func main() {
    t := &oauth.Transport{
        Token: &oauth.Token{AccessToken: "..."},
    }
    client := github.NewClient(t.Client())
    issues, r, err := client.Activity.ListIssueEventsForRepository("bosun-monitor", "bosun", &github.ListOptions{Page: 1, PerPage: 20})
    fmt.Println(r.Status)
    if err != nil {
        panic(err)
    }
    fmt.Println(issues)
}
```

Returns:

```
    200 OK
    panic: json: cannot unmarshal number into Go value of type string

    goroutine 1 [running]:
    main.main()
        /Users/kbrandt/src/github.go:19 +0x2f3

    goroutine 9 [runnable]:
    net/http.(*persistConn).readLoop(0xc208060000)
        /usr/local/Cellar/go/1.4/libexec/src/net/http/transport.go:928 +0x9ce
    created by net/http.(*Transport).dialConn
        /usr/local/Cellar/go/1.4/libexec/src/net/http/transport.go:660 +0xc9f

    goroutine 17 [syscall, locked to thread]:
    runtime.goexit()
        /usr/local/Cellar/go/1.4/libexec/src/runtime/asm_amd64.s:2232 +0x1

    goroutine 10 [select]:
    net/http.(*persistConn).writeLoop(0xc208060000)
        /usr/local/Cellar/go/1.4/libexec/src/net/http/transport.go:945 +0x41d
    created by net/http.(*Transport).dialConn
        /usr/local/Cellar/go/1.4/libexec/src/net/http/transport.go:661 +0xcbc
    exit status 2

    exit status 1
```

yisiper:
The API and library doesn't match.
https://github.com/google/go-github/blob/master/github/activity_events.go#L153

```
// GitHub API docs: http://developer.github.com/v3/activity/events/#list-issue-events-for-a-repository
func (s *ActivityService) ListIssueEventsForRepository(owner, repo string, opt *ListOptions) ([]Event, *Response, error) {
	u := fmt.Sprintf("repos/%v/%v/issues/events", owner, repo)
	u, err := addOptions(u, opt)
	if err != nil {
		return nil, nil, err
	}
```

if you want this API
[...]

PR #474, opened 2016-11-24
Fix type for Event.ID field

whilp:
ID is an int, not a string, per:

https://developer.github.com/v3/issues/events/#get-a-single-event

Fixes https://github.com/google/go-github/issues/176


whilp:
Ah -- this is the incorrect approach: the problem is that `ListIssueEventsForRepository` needs to return `IssueEvent`s, not that the field is incorrectly typed. Will reroll.

PR #476, opened 2016-11-27
Updates for breaking changes to branch protection API

alindeman:
Per <https://developer.github.com/changes/2016-06-27-protected-branches-api-update/>, breaking changes were made to the protected branches API during the preview period.

This PR gets started implementing the new routes. Specifically: 
* Removes `Branch.Protection`, no longer returned by the API
* Removes `EditBranch` function, protection is now modified using separate API routes
* Adds `GetBranchProtection` function
* Adds `UpdateBranchProtection` function
* Adds `RemoveBranchProtection` function

`omitempty` is intentionally left off fields in `ProtectionRequest` because passing `nil` (JSON `null`) has semantic meaning (described in <https://developer.github.com/v3/repos/branches/#update-branch-protection>)

Like the issues API, we separate some request structures from response structures because on the request, users and teams are specified as strings of logins or slugs, respectively, but in responses, full representations are returned.

Feedback and suggestions welcomed :heart:

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -501,29 +501,56 @@ func (s *RepositoriesService) ListTags(owner string, repo string, opt *ListOptio
 
 // Branch represents a repository branch
 type Branch struct {
-	Name       *string     `json:"name,omitempty"`
-	Commit     *Commit     `json:"commit,omitempty"`
-	Protection *Protection `json:"protection,omitempty"`
+	Name      *string `json:"name,omitempty"`
+	Commit    *Commit `json:"commit,omitempty"`
+	Protected *bool   `json:"protected,omitempty"`
 }
 
 // Protection represents a repository branch's protection
 type Protection struct {
-	Enabled              *bool                 `json:"enabled,omitempty"`
-	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks,omitempty"`
+	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks"`
+	Restrictions         *BranchRestrictions   `json:"restrictions"`
+}
+
+// ProtectionRequest represents a request to create/edit a branch's protection.
+// It is separate from Protection above because the request structure is
+// different from the response structure.
+type ProtectionRequest struct {
+	RequiredStatusChecks *RequiredStatusChecks      `json:"required_status_checks"`
+	Restrictions         *BranchRestrictionsRequest `json:"restrictions"`
 }
 
 // RequiredStatusChecks represents the protection status of a individual branch
 type RequiredStatusChecks struct {
-	// Who required status checks apply to.
-	// Possible values are:
-	//     off
-	//     non_admins
-	//     everyone
-	EnforcementLevel *string `json:"enforcement_level,omitempty"`
-	// The list of status checks which are required
+	// Enforce required status checks for repository administrators.
+	IncludeAdmins *bool `json:"include_admins,omitempty"`
+	// Require branches to be up to date before merging.
+	Strict *bool `json:"strict,omitempty"`
+	// The list of status checks to require in order to merge into this
+	// branch
 	Contexts *[]string `json:"contexts,omitempty"`
 }
 
+// BranchRestrictions represents the restriction that only certain users or
+// teams may push to a branch
+type BranchRestrictions struct {
+	// The list of user logins with push access
+	Users []*User `json:"users,omitempty"`
+	// The list of team slugs with push access
+	Teams []*Team `json:"teams,omitempty"`
+}
+
+// BranchRestrictionsRequest represents the request to create/edit the
+// restriction that only certain users or teams may push to a branch. It is
+// separate from BranchRestrictions above because the request structure is
+type BranchRestrictionsRequest struct {
+	// The list of user logins with push access
+	Users *[]string `json:"users,omitempty"`
+	// The list of team slugs with push access
+	Teams *[]string `json:"teams,omitempty"`
+}
+
+// different from the response structure.
 // ListBranches lists branches for the specified repository.
 //
 // GitHub API docs: http://developer.github.com/v3/repos/#list-branches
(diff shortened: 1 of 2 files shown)


github/repos.go
```diff
@@ -501,29 +501,56 @@ func (s *RepositoriesService) ListTags(owner string, repo string, opt *ListOptio
...
 	Contexts *[]string `json:"contexts,omitempty"`
 }
 
+// BranchRestrictions represents the restriction that only certain users or
+// teams may push to a branch
+type BranchRestrictions struct {
+	// The list of user logins with push access
+	Users []*User `json:"users,omitempty"`
+	// The list of team slugs with push access
+	Teams []*Team `json:"teams,omitempty"`
```
shurcooL: This seems inconsistent with `BranchRestrictionsRequest` below. Both structs have fields that are slices, and both use `omitempty` JSON option.

However, this struct uses fields that are slices of pointers (to `User` and `Team` structs), but `BranchRestrictionsRequest` below uses pointers to slices (of `string`s).

Is that intentional, or would it be better to make it consistently pointers to slices?

I've checked, and the rest of the codebase actually does something similar to what you have here, so maybe this is fine. But I just wanted to raise this to your attention and check what you think.

github/repos.go
```diff
@@ -501,29 +501,56 @@ func (s *RepositoriesService) ListTags(owner string, repo string, opt *ListOptio
...
-	//     non_admins
-	//     everyone
-	EnforcementLevel *string `json:"enforcement_level,omitempty"`
-	// The list of status checks which are required
+	// Enforce required status checks for repository administrators.
+	IncludeAdmins *bool `json:"include_admins,omitempty"`
+	// Require branches to be up to date before merging.
+	Strict *bool `json:"strict,omitempty"`
+	// The list of status checks to require in order to merge into this
+	// branch
```
shurcooL: Nitpick. This sentence, unlike the ones above, is missing a period at the end.

github/repos.go
```diff
@@ -501,29 +501,56 @@ func (s *RepositoriesService) ListTags(owner string, repo string, opt *ListOptio
...
+type BranchRestrictions struct {
+	// The list of user logins with push access
+	Users []*User `json:"users,omitempty"`
+	// The list of team slugs with push access
+	Teams []*Team `json:"teams,omitempty"`
+}
+
+// BranchRestrictionsRequest represents the request to create/edit the
+// restriction that only certain users or teams may push to a branch. It is
+// separate from BranchRestrictions above because the request structure is
```
shurcooL: Also missing a period at the end of that sentence.

The documentation for the fields in this struct and one above as well.

alindeman: Good catch. Fixed.

github/repos.go
alindeman: Good catch. Fixed.

alindeman:
@shurcooL This is ready for another look, I think. Cheers.

github/repos.go
shurcooL: Totally fine to leave it as is, since it's consistent. Changing that would be out of scope of this PR.

It's pretty harmless aside from being a source of minor inconsistency. A pointer to a slice can be nil, or a slice itself can be nil.

github/repos.go
```diff
@@ -520,37 +520,37 @@ type ProtectionRequest struct {
...
 	Users []*User `json:"users,omitempty"`
-	// The list of team slugs with push access
+	// The list of team slugs with push access.
 	Teams []*Team `json:"teams,omitempty"`
 }
 
 // BranchRestrictionsRequest represents the request to create/edit the
 // restriction that only certain users or teams may push to a branch. It is
 // separate from BranchRestrictions above because the request structure is
+// different from the response structure.
```
shurcooL: Hehe, I was gonna say that it feels like something was missing here. ;)

github/repos.go
```diff
@@ -501,29 +501,56 @@ func (s *RepositoriesService) ListTags(owner string, repo string, opt *ListOptio
...
-	Enabled              *bool                 `json:"enabled,omitempty"`
-	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks,omitempty"`
+	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks"`
+	Restrictions         *BranchRestrictions   `json:"restrictions"`
 }
 
-// RequiredStatusChecks represents the protection status of a individual branch
+// ProtectionRequest represents a request to create/edit a branch's protection.
+// It is separate from Protection above because the request structure is
+// different from the response structure.
```
shurcooL: I _think_ the entire "It is separate from Protection above because the request structure is different from the response structure." sentence can be left out. It should be pretty self-evident from the code itself.

But I might be biased because I've seen tons of Go code that does this, so it's up to you if you wanna keep it.

github/repos.go
```diff
@@ -571,25 +598,58 @@ func (s *RepositoriesService) GetBranch(owner, repo, branch string) (*Branch, *R
...
-	u := fmt.Sprintf("repos/%v/%v/branches/%v", owner, repo, branchName)
-	req, err := s.client.NewRequest("PATCH", u, branch)
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#get-branch-protection
+func (s *RepositoriesService) GetBranchProtection(owner, repo, branch string) (*Protection, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection", owner, repo, branch)
+	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
+	req.Header.Set("Accept", mediaTypeProtectedBranchesPreview)
```
shurcooL: Minor, but I think it'd be best to stick to the usual formula here for consistency. Add a blank line before `req.Header.Set` and this comment:

```
```

Just to keep this endpoint code looking more [like others](https://github.com/google/go-github/blob/7baedea38fd13aae4e4e9cb0854250797569a91c/github/git_commits.go#L67-L69).

Here, and in 2 other places.

github/repos.go
alindeman: Makes sense. I think the `issue` code has a comment similar to this, but I agree it's pretty self-evident. I've removed it.

github/repos.go
alindeman: :+1: Done

gmlewis:
Thank you, @alindeman for your contribution, and @shurcooL for your code reviews!
LGTM.
Merging.

gmlewis:
Included in https://github.com/google/go-github/commit/3b96b535044d9bdb869b91c7f7d6807f327c12f2

github/repos.go
shurcooL: Great! It can be done for `BranchRestrictionsRequest` type too, it's even less needed there.

PR #479, opened 2016-11-29
Adds AdminService and functions for updating LDAP mappings

alindeman:
/cc: @sr

### DIFF
--- a/github/admin.go
+++ b/github/admin.go
@@ -1,0 +1,102 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import "fmt"
+
+// AdminService handles communication with the admin related methods of the
+// GitHub API. These API routes are normally only accessible for GitHub
+// Enterprise installations.
+//
+// GitHub API docs: https://developer.github.com/v3/enterprise/
+type AdminService service
+
+// TeamLDAPMapping represents the mapping between a GitHub team and an LDAP
+// group.
+type TeamLDAPMapping struct {
+	ID          *int    `json:"id,omitempty"`
+	LDAPDN      *string `json:"ldap_dn,omitempty"`
+	URL         *string `json:"url,omitempty"`
+	Name        *string `json:"name,omitempty"`
+	Slug        *string `json:"slug,omitempty"`
+	Description *string `json:"description,omitempty"`
+	Privacy     *string `json:"privacy,omitempty"`
+	Permission  *string `json:"permission,omitempty"`
+
+	MembersURL      *string `json:"members_url,omitempty"`
+	RepositoriesURL *string `json:"repositories_url,omitempty"`
+}
+
+func (m TeamLDAPMapping) String() string {
+	return Stringify(m)
+}
+
+// UserLDAPMapping represents the mapping between a GitHub user and an LDAP
+// user.
+type UserLDAPMapping struct {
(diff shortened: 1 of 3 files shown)


github/github.go
```diff
@@ -188,6 +189,7 @@ func NewClient(httpClient *http.Client) *Client {
 
 	c := &Client{client: httpClient, BaseURL: baseURL, UserAgent: userAgent, UploadURL: uploadURL}
 	c.common.client = c
+	c.Admin = (*AdminService)(&c.common)
```
gmlewis: nit: please move this one line below c.Activity to keep these sorted.

github/github.go
```diff
@@ -116,6 +116,7 @@ type Client struct {
 	common service // Reuse a single struct instead of allocating one for each service on the heap.
 
 	// Services used for talking to different parts of the GitHub API.
+	Admin          *AdminService
```
gmlewis: nit: please move this one line below Activity to keep these sorted.

alindeman:
@gmlewis Cheers, good idea. I've pushed a commit to put it in the correct alphabetical order.

gmlewis:
Looks great, @alindeman - thank you!

github/admin.go
```diff
@@ -0,0 +1,102 @@
...
+
+// AdminService handles communication with the admin related methods of the
+// GitHub API. These API routes are normally only accessible for GitHub
+// Enterprise installations.
+//
+// GitHub API docs: https://developer.github.com/v3/enterprise/
+type AdminService service
+
+// TeamLDAPMapping represents the mapping between a GitHub team and an LDAP
+// group.
```
shurcooL: Minor thought. If it were me, I wouldn't create a second line just for one word. [Why not](https://twitter.com/rob_pike/status/563801489190043648) keep it together with the rest of the sentence?

Either way is completely fine.

github/admin.go
```diff
@@ -0,0 +1,102 @@
...
+
+package github
+
+import "fmt"
+
+// AdminService handles communication with the admin related methods of the
+// GitHub API. These API routes are normally only accessible for GitHub
+// Enterprise installations.
+//
+// GitHub API docs: https://developer.github.com/v3/enterprise/
```
shurcooL: Should this URL be https://developer.github.com/v3/enterprise/ldap/? That seems to be the specific page for the 2 endpoints being supported by `AdminService` here.

Unless the plan is to expand `AdminService` to cover all other parts of https://developer.github.com/v3/enterprise/, including https://developer.github.com/v3/enterprise/admin_stats/, https://developer.github.com/v3/enterprise/license/, https://developer.github.com/v3/enterprise/management_console/, etc.?

alindeman: My intention was to add more routes from the enterprise/admin section at a later point, yep. There are other routes in the `admin/` namespace that I haven't added yet.

github/admin.go
alindeman: This is my editor trying to keep everything under 80. I should probably make it be a bit more liberal. Pushed an update for this :)

gmlewis:
Thanks for the LGTM, @shurcooL, and @alindeman for the tweaks!
Merging.

gmlewis:
Included in https://github.com/google/go-github/commit/69894f9ef7352c33b5d874b2e3f799424c0f7387

github/admin.go
shurcooL: Thanks. I noticed you've changed 2 instances of this for the method docs, but not the 2 instances for types. Just FYI, fixing that is very optional.

PR #480, opened 2016-12-01
Adjusts comment style based on feedback

alindeman:
@shurcooL: Thanks for catching this. Easy enough to fix.


gmlewis:
Thank you, @alindeman and @shurcooL.
LGTM.
Merging.

Issue #6, opened 2013-06-16
handling custom media types


willnorris:
A handful of API methods (like [issues](http://developer.github.com/v3/issues/)) support the use of [custom mime](http://developer.github.com/v3/media/) types to return additional fields in the response, using different formats.

As currently designed, go-github doesn't have a simple way to expose this, since most HTTP-specific details are hidden from the caller.  We could probably just add it to the Options parameter for the given method, but that feels a little weird.  Additionally, I don't yet have a real use-case for needing to use these custom mime-types.  Therefore, I'm currently leaning toward punting on this until someone actually needs it.

So if you need it, please comment here, and we can look at how to work that into the library.

cburmeister:
@willnorris is there a way to specify a media type for `GetContents`?

```
client := github.NewClient(tc)
repo, _, _, err := client.Repositories.GetContents("username", "repository", "path", nil)
```

As is the `contents` are base64 encoded.

willnorris:
We don't currently expose that for `GetContents`, but it wouldn't be too hard to add it to the existing [RepositoryContentGetOptions](https://godoc.org/github.com/google/go-github/github#RepositoryContentGetOptions)... just add a `MediaType` field like in the example above, then switch on that inside the method to set the Accept header.

Do you wanna take a pass at that?

parkr:
What about a global toggle for each feature?

```go
github.Enable(github.BetaLicenses)
```

willnorris:
@parkr this is a bit different than the media types that are used to enable pre-released APIs.  In those cases, we intentionally try to just enable those APIs transparently so that users of the library don't even realize that it's sending a custom media type.  So far this has worked well, and we haven't had any conflicts between multiple pre-release APIs running on the same endpoints.

What @cburmeister is asking for though is not a pre-release API, it's just an endpoint that supports responses in JSON, HTML, or raw (and just happens to use media type to indicate which one is preferred).  In this case we definitely do want to expose this to users of the library, since the desired response type is generally a conscious decision on the developer's part.

Of course now that I say all this, I realize that we might not simply be able to add MediaType to the Options struct and use the existing method, since the go method expects to return a `*RepositoryContent` and `[]*RepositoryContent`, which we won't have.  Maybe if a custom media type is specified, then we return nil for both of those, and the caller is expected to parse the body of the `*Response` manually?  I don't think there is really any other way to do it, we'll just need to make that clear on the docs for the MediaType field.

parkr:
[...]

PR #481, opened 2016-12-02
Add function for PullRequestService to download raw content of pull r…

huydx:
…equest

We already has an issue here https://github.com/google/go-github/issues/6
And as @willnorris said it's better to use Contents API to download, but Contents API seems to support blob or file content which is a part of a repository git tree only. (not include pull request)

I want to support patch/diff media type in pull request also (which may not be supported even in future release).
So i added method `GetRaw` to `PullRequestService` separate from original `Get` method (to maintain backward compatibility)

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -42,6 +42,8 @@ const (
 	mediaTypeV3                = "application/vnd.github.v3+json"
 	defaultMediaType           = "application/octet-stream"
 	mediaTypeV3SHA             = "application/vnd.github.v3.sha"
+	mediaTypeV3Diff            = "application/vnd.github.v3.diff"
+	mediaTypeV3Patch           = "application/vnd.github.v3.patch"
 	mediaTypeOrgPermissionRepo = "application/vnd.github.v3.repository+json"
 
 	// Media Type values to access preview APIs
@@ -152,6 +154,20 @@ type ListOptions struct {
 // UploadOptions specifies the parameters to methods that support uploads.
 type UploadOptions struct {
 	Name string `url:"name,omitempty"`
+}
+
+// RawType represents type of raw format of a request instead of JSON.
+type RawType uint8
+
+const (
+	Diff RawType = 1 + iota
+	Patch
+)
+
+// RawOptions specifies parameters when user wants to get raw format of
+// a response instead of JSON.
+type RawOptions struct {
+	Type RawType
 }
 
 // addOptions adds the parameters in opt as URL query parameters to s.  opt
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -6,6 +6,7 @@
 package github
 
 import (
+	"bytes"
 	"fmt"
 	"time"
 )
@@ -132,6 +133,32 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
 	}
 
 	return pull, resp, err
+}
+
+// GetRaw gets raw (diff or patch) format of a pull request.
+func (s *PullRequestsService) GetRaw(owner string, repo string, number int, opt RawOptions) (string, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pulls/%d", owner, repo, number)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return "", nil, err
+	}
+
+	switch opt.Type {
+	case Diff:
+		req.Header.Set("Accept", mediaTypeV3Diff)
+	case Patch:
+		req.Header.Set("Accept", mediaTypeV3Patch)
+	default:
+		return "", nil, fmt.Errorf("unsupported raw type %d", opt.Type)
+	}
+
+	ret := new(bytes.Buffer)
+	resp, err := s.client.Do(req, ret)
+	if err != nil {
+		return "", resp, err
+	}
+
+	return ret.String(), resp, err
 }
 
 // NewPullRequest represents a new pull request to be created.
--- a/github/pulls_test.go
+++ b/github/pulls_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 	"testing"
 )
 
@@ -64,6 +65,60 @@ func TestPullRequestsService_Get(t *testing.T) {
 	want := &PullRequest{Number: Int(1)}
 	if !reflect.DeepEqual(pull, want) {
 		t.Errorf("PullRequests.Get returned %+v, want %+v", pull, want)
+	}
+}
+
+func TestPullRequestService_GetRawPatch(t *testing.T) {
+	setup()
+	defer teardown()
+	const rawStr = "@@patch content"
+
+	mux.HandleFunc("/repos/o/r/pulls/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeV3Patch)
+		fmt.Fprint(w, rawStr)
+	})
+
+	ret, _, err := client.PullRequests.GetRaw("o", "r", 1, RawOptions{Patch})
+	if err != nil {
+		t.Errorf("PullRequests.GetRaw return error: %v", err)
+	}
+
+	if ret != rawStr {
+		t.Errorf("PullRequests.GetRaw returned %s want %s", ret, rawStr)
+	}
+}
+
+func TestPullRequestService_GetRawDiff(t *testing.T) {
+	setup()
+	defer teardown()
+	const rawStr = "@@diff content"
+
+	mux.HandleFunc("/repos/o/r/pulls/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeV3Diff)
+		fmt.Fprint(w, rawStr)
+	})
+
+	ret, _, err := client.PullRequests.GetRaw("o", "r", 1, RawOptions{Diff})
+	if err != nil {
+		t.Errorf("PullRequests.GetRaw return error: %v", err)
+	}
+
+	if ret != rawStr {
+		t.Errorf("PullRequests.GetRaw returned %s want %s", ret, rawStr)
+	}
+}
+
+func TestPullRequestService_GetRawInvalid(t *testing.T) {
+	setup()
+	defer teardown()
+	_, _, err := client.PullRequests.GetRaw("o", "r", 1, RawOptions{100})
+	if err == nil {
+		t.Error("PullRequests.GetRaw should return error")
+	}
+	if !strings.Contains(err.Error(), "unsupported raw type") {
+		t.Error("PullRequests.GetRaw should return unsupported raw type error")
 	}
 }
 


huydx:
I see the travis fails but not sure what does that mean, do anybody know?

gmlewis:
Yes.  Please run gofmt on both `pulls.go` and `pulls_test.go`.

huydx:
thanks @gmlewis, ran and pushed :)

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
 	return pull, resp, err
 }
 
+// Get pull request raw format (diff or patch)
```
shurcooL: To follow Go style, the comment should begin with the symbol name. It's `GetRaw`, so it should be `// GetRaw gets ...`. Also, it'd be nice to end the sentence with a period.

> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

Source: https://golang.org/doc/effective_go.html#commentary

github/github.go
```diff
@@ -52,6 +52,11 @@ const (
 	// https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/
 	mediaTypeStarringPreview = "application/vnd.github.v3.star+json"
 
+	//https://developer.github.com/v3/media/
```
shurcooL: To follow Go style and for consistency, there should be a single space at the beginning of this comment.

See https://dmitri.shuralyov.com/idiomatic-go#comments-for-humans-always-have-a-single-space-after-the-slashes for full rationale.

github/github.go
```diff
@@ -154,6 +159,17 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+type MediaRawType int
```
shurcooL: This (and a few more below) is an exported package-level symbol. Go style suggests:

> All top-level, exported names should have doc comments, as should non-trivial unexported type or function declarations.

Source: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
...
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return "", nil, err
+	}
+
+	if opt.Type == Diff {
+		req.Header.Set("Accept", mediaTypeDiff)
+	} else if opt.Type == Patch {
+		req.Header.Set("Accept", mediaTypePatch)
+	}
```
shurcooL: Consider using a `switch` here, it would look nicer.

github/github.go
```diff
@@ -52,6 +52,11 @@ const (
 	// https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/
 	mediaTypeStarringPreview = "application/vnd.github.v3.star+json"
 
+	//https://developer.github.com/v3/media/
+	mediaTypeDiff  = "application/vnd.github.v3.diff"
+	mediaTypePatch = "application/vnd.github.v3.patch"
+	mediaTypeSha   = "application/vnd.github.v3.sha"
```
shurcooL: `mediaTypeSha` already exists, it's [`mediaTypeV3SHA`](https://github.com/google/go-github/blob/dccec7b5f17234b8219494c2b5b03af034e08c04/github/github.go#L44). It's probably better to add the other 2 in that same general area.

Also, the reason it's `SHA` and not `Sha` is because Go style uses consistent case for acronyms. See [here](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) for rationale.

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
...
 
+// Get pull request raw format (diff or patch)
+func (s *PullRequestsService) GetRaw(owner string, repo string, number int, opt *MediaRawTypeOption) (string, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pulls/%d", owner, repo, number)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return "", nil, err
+	}
+
+	if opt.Type == Diff {
```
shurcooL: `opt` is a pointer, but you never check if its value is nil.

Usually a pointer is used when `nil` means using default options. Either do that, or make it a value.

Making it a value makes more sense, since I can't think of a good "default" value for `MediaRawTypeOption`.

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
 	return pull, resp, err
 }
 
+// Get pull request raw format (diff or patch)
+func (s *PullRequestsService) GetRaw(owner string, repo string, number int, opt *MediaRawTypeOption) (string, *Response, error) {
```
shurcooL: I think `MediaRawTypeOption` should be named `GetRawOptions`. See other funcs that take an options struct, that's the pattern they follow.

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
...
+		req.Header.Set("Accept", mediaTypePatch)
+	}
+
+	ret := new(bytes.Buffer)
+	resp, err := s.client.Do(req, ret)
+	if err != nil {
+		return "", resp, err
+	}
+
+	return string(ret.Bytes()), resp, err
```
shurcooL: `string(ret.Bytes())` can be simplified to `ret.String()`. But are you sure that returning a `string` rather than `[]byte` makes the most sense?

huydx:
@shurcooL Thanks for your nice comments, i agree with all of your comments and will fix after we agree on should we create new method or not.
I think we have some options here
- Reuse original [Get](https://github.com/google/go-github/blob/master/github/pulls.go#L121) method by put an option there and new field into `PullRequest` struct called `raw` (which is string type). This option is not good in term of github API mapping because actually there is not any `raw` in the real response
- Create new `GetRaw` method as I did
- Write generic `DownloadRaw` as below

```
func DownloadRaw(type {}interface, rawOption *RawOption) {

}
```

type could be `PullRequest`, `Commit` or types may support `patch` or `diff` in future.

HDYT?

shurcooL:
> @shurcooL Thanks for your nice comments, i agree with all of your comments and will fix after we agree on should we create new method or not.

Sure thing. Sounds good!

Of the three ideas you've listed, I think first one is pretty much not good and not viable. It'd be a breaking API change, and it would not be worth to touch such an important/common endpoint with an option that's unlikely to be used very often.

Ideas 2 and 3 seem like reasonable solutions. I can't seem to find one I think is better.

Idea 2 is a smaller step and more concrete, and you need it. Idea 3 seems like a better choice in the long run, but only if all other types are implemented and useful in the future.

We can also look at the API for downloading contents and see how this can compare and fit in: https://godoc.org/github.com/google/go-github/github#RepositoriesService.DownloadContents.

I wonder if we're missing any other ideas for how to do this. So far I can't think of anything else, other than doing a manual request with [`NewRequest`](https://godoc.org/github.com/google/go-github/github#Client.NewRequest) and [`Do`](https://godoc.org/github.com/google/go-github/github#Client.Do).

That's all I got for now. Thoughts from others?

huydx:
@shurcooL Thanks for throughout comment, I've take a look at download contents api, but the implementation is mapping 1-1 to separate API at github endpoint, so this PR may be not fit to go in there. As an user, i think `GetRaw` method is quite easy to reasoning, so keep as it is seems  fine to me :).

shurcooL:
All right, that sounds good to me.

@willnorris or @gmlewis, what do you think of the currently proposed approach of adding `GetRaw` method to `PullRequestsService` in order to provide ability to fetch the pull request with different media types?

I think your feedback and approval to go ahead with that is needed to make progress here. If accepted, then @huydx can work on addressing the review comments.

gmlewis:
Yes, I agree that [option 3 above](https://github.com/google/go-github/pull/481#issuecomment-264628086) is probably best for the long-haul but also will involve a great deal of work or at the very least is a non-trivial change.

I also agree that since this is needed now, option 2 (this PR) seems like the best way to move forward.

@willnorris has the final say on a change like this, and he may also have a better idea.

In the meantime, though, @willnorris has said that if @shurcooL and I agree, then we can move forward with that decision. I therefore recommend letting this PR sit for one or two more days to allow @willnorris to respond, and then go ahead and move forward with the excellent review comments provided by @shurcooL. Thank you, @huydx!

huydx:
gentle ping @willnorris

shurcooL:
@huydx, see last paragraph in the comment above:

> In the meantime, though, @willnorris has said that if @shurcooL and I agree, then we can move forward with that decision. I therefore recommend letting this PR sit for one or two more days to allow @willnorris to respond, and then go ahead and move forward with the excellent review comments provided by @shurcooL. Thank you, @huydx!

It has been a few days, and there's no pushback. So feel free to address the feedback and we'll be able to merge this.

huydx:
fixed @shurcooL . Return of `GetRaw` function now is []byte to avoid unncessary memory copy (which is not transparent from user view).

.gitignore
```diff
@@ -1 +1,2 @@
 *.test
+.idea
```
shurcooL: Please get rid of this, I don't think it belongs here. If you need to ignore `.idea` files, you can add it to your personal `.git/info/exclude` file, which doesn't get committed. See https://help.github.com/articles/ignoring-files/#explicit-repository-excludes.

github/github.go
```diff
@@ -154,6 +156,19 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+// MediaRawType represents raw return of a request instead of json format
```
shurcooL: Add a missing period at end of sentence and spell it as JSON, not json, since it's an acronym.

github/github.go
```diff
@@ -154,6 +156,19 @@ type UploadOptions struct {
...
+type MediaRawType int
+
+const (
+	Diff MediaRawType = 1 + iota
+	Patch
+)
+
+// GetRawOptions specifies the optional parameter when user want to get raw presentation of a response instead of json format
```
shurcooL: I think you can simplify the docs to just:

```
// GetRawOptions specifies the parameters to the PullRequestsService.GetRaw method.
```

The parameter is not optional, so it's not accurate to say that it is.

github/github.go
```diff
@@ -154,6 +156,19 @@ type UploadOptions struct {
...
+type GetRawOptions struct {
```
shurcooL: What do you think about shortening the name to `RawOptions`?

If you look at other structs in this package that have `Options` suffix, they're usually named after the general area of interest, not the exact method name.

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
 	return pull, resp, err
 }
 
+// GetRaw gets raw (diff or patch) format of a pull request.
+func (s *PullRequestsService) GetRaw(owner string, repo string, number int, opt GetRawOptions) ([]byte, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pulls/%d", owner, repo, number)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return []byte{}, nil, err
```
shurcooL: No need to allocate an empty byte slice when returning an error here, just return a `nil` byte slice.

```Go
return  nil, nil, err
```

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
...
+	}
+
+	if opt.Type == Diff {
+		req.Header.Set("Accept", mediaTypeV3Diff)
+	} else if opt.Type == Patch {
+		req.Header.Set("Accept", mediaTypeV3Patch)
+	}
```
shurcooL: Two things here.

1. A `switch` statement would be a good fit here.
2. How about handling when `opt.Type` value is not any of the supported types? I think returning an error would be most appropriate. Currently, it would just perform unspecified actions.

github/pulls.go
```diff
@@ -134,6 +135,29 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
...
+
+	ret := new(bytes.Buffer)
+	resp, err := s.client.Do(req, ret)
+	if err != nil {
+		return []byte{}, resp, err
```
shurcooL: Return a `nil` byte slice instead of allocating an empty one here also.

github/pulls_test.go
```diff
@@ -67,6 +67,48 @@ func TestPullRequestsService_Get(t *testing.T) {
 	}
 }
 
+func TestPullRequestService_GetRawPatch(t *testing.T) {
+	setup()
+	defer teardown()
+	var rawStr = "@@patch content"
```
shurcooL: This can be `const`, no need for `var`.

github/pulls_test.go
```diff
@@ -67,6 +67,48 @@ func TestPullRequestsService_Get(t *testing.T) {
...
+
+	if string(ret) != rawStr {
+		t.Errorf("PullRequests.GetRaw returned %s want %s", ret, rawStr)
+	}
+}
+
+func TestPullRequestService_GetRawDiff(t *testing.T) {
+	setup()
+	defer teardown()
+	var rawStr = "@@diff content"
```
shurcooL: This can be `const` too.

huydx:
@shurcooL thanks for your patience, fix for 2nd round. Added test for invalid option as well.

github/pulls.go
```diff
@@ -134,6 +135,32 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
...
+		return nil, nil, err
+	}
+
+	switch opt.Type {
+	case Diff:
+		req.Header.Set("Accept", mediaTypeV3Diff)
+	case Patch:
+		req.Header.Set("Accept", mediaTypeV3Patch)
+	default:
+		return nil, nil, fmt.Errorf("Unsupported raw type %d", opt.Type)
```
shurcooL: Style issue. Error strings should not be capitalized.

Rationale: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

github/pulls_test.go
```diff
@@ -67,6 +68,60 @@ func TestPullRequestsService_Get(t *testing.T) {
...
+	defer teardown()
+	const rawStr = "@@patch content"
+
+	mux.HandleFunc("/repos/o/r/pulls/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeV3Patch)
+		fmt.Fprint(w, rawStr)
+	})
+
+	ret, _, err := client.PullRequests.GetRaw("o", "r", 1, RawOptions{Patch})
```
shurcooL: Client code will not be able to write `github.RawOptions{github.Patch}`, because that'd cause a vet issue of composite literal using unkeyed fields.

They'd have to write `github.RawOptions{Type: github.Patch}`.

On the one hand, keeping the `RawOptions` struct means we can add additional options in the future, but on the other hand, it's more verbose than just passing a `RawType` parameter without a struct.

Just making an observation here.

shurcooL:
Thanks for addressing the feedback @huydx.

I've left a few more minor comments. Overall, this looks good so far. I'll review it more carefully and leave my formal LGTM tomorrow or so.

@gmlewis, now that this PR has almost the final form, do you wanna take another look and tell us how it looks to you?

github/github.go
```diff
@@ -154,6 +156,19 @@ type UploadOptions struct {
...
 
+// RawType represents type of raw format of a request instead of JSON.
+type RawType int
+
+const (
+	Diff RawType = 1 + iota
+	Patch
+)
+
+// RawOptions specifies parameters when user want to get raw format of a response instead of JSON.
```
gmlewis: nit: s/want/wants/
Also, if you browse around this repo, you will see that the comments attempt to stay within 80 columns although obviously this is not strictly necessary for Go programming. But just to keep things consistent, it might be nice to break this line.

huydx:
PTAL~ @gmlewis @shurcooL

github/pulls.go
```diff
@@ -134,6 +135,32 @@ func (s *PullRequestsService) Get(owner string, repo string, number int) (*PullR
 	return pull, resp, err
 }
 
+// GetRaw gets raw (diff or patch) format of a pull request.
+func (s *PullRequestsService) GetRaw(owner string, repo string, number int, opt RawOptions) ([]byte, *Response, error) {
```
shurcooL: What was your reason to go with returning `[]byte` instead of `string`?

I've thought about it some more, and `string` seems to be a better fit to me. Both diff and patch are textual formats. go-github already has a few endpoints that return text, such as [`Markdown`](https://godoc.org/github.com/google/go-github/github#Client.Markdown), [`Octocat`](https://godoc.org/github.com/google/go-github/github#Client.Octocat) and [`Zen`](https://godoc.org/github.com/google/go-github/github#Client.Zen).

Do you have arguments for `[]byte`?

github/github.go
```diff
@@ -154,6 +156,20 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+// RawType represents type of raw format of a request instead of JSON.
+type RawType int
```
shurcooL: Maybe use `uint8` instead of `int` here for consistency with https://github.com/google/go-github/blob/466070b0580728e63bd1a415e0019639e55d7148/github/github.go#L684.

huydx:
PTAL @shurcooL

github/github.go
```diff
@@ -154,6 +156,20 @@ type UploadOptions struct {
 	Name string `url:"name,omitempty"`
 }
 
+// RawType represents type of raw format of a request instead of JSON.
+type RawType uint8
+
+const (
+	Diff RawType = 1 + iota
+	Patch
```
shurcooL: This causes 2 golint issues:

```
github.go:163:2: exported const Diff should have comment (or a comment on this block) or be unexported
github.go:164:2: exported const Patch should have comment (or a comment on this block) or be unexported
```

shurcooL:
I've applied the minor changes above (also swapped order of GetRawDiff, GetRawPatch to be consistent) and merged this PR as 7edde482875504b49769ad097eab7c572547ff21. Thank you @huydx.

huydx:
thanks~ @shurcooL

tamird:
I just noticed this change, and was about to apply the following diff to my code:
```diff
diff --git a/pkg/cmd/github-pull-request-make/main.go b/pkg/cmd/github-pull-request-make/main.go
index 7d917bee6..bd3723011 100644
--- a/pkg/cmd/github-pull-request-make/main.go
+++ b/pkg/cmd/github-pull-request-make/main.go
@@ -32,7 +32,6 @@ import (
 	"go/build"
 	"io"
 	"log"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -134,16 +133,17 @@ func main() {
 		return
 	}
 
-	resp, err := http.Get(*currentPull.DiffURL)
+	diff, _, err := client.PullRequests.GetRaw(
+		*currentPull.Head.Repo.Organization.Name,
+		*currentPull.Head.Repo.Name,
+		*currentPull.Number,
+		github.RawOptions{Type: github.Patch},
+	)
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
-		log.Fatalf("http.Get(%s).Status = %s", *currentPull.DiffURL, resp.Status)
-	}
 
-	pkgs, err := pkgsFromDiff(resp.Body)
+	pkgs, err := pkgsFromDiff(diff)
 	if err != nil {
 		log.Fatal(err)
 	}

```
However, the fact that `GetRaw` returns a string is kind of unfortunate - returning the bytes.Buffer directly would be strictly more flexible, and more efficient.

shurcooL:
Just for the sake of having a complete picture, can you show what `pkgsFromDiff` does with the `io.Reader`?

This API is still very new and not many people can be depending on it already. If we can agree on an improvement, changing it is a lot more viable than breaking an API that has been out for years.

tamird:
https://github.com/cockroachdb/cockroach/blob/master/pkg/cmd/github-pull-request-make/main.go#L62

shurcooL:
I see. So you're wrapping the reader in a `bufio.NewReader`, and doing that for its `ReadLine` method.

All that could be rewritten to loop over the `string` with `strings.IndexByte(s, '\n')`, since buffering a `*bytes.Buffer` or `string` is not needed.

But, still, returning `*bytes.Buffer` instead of `buf.String()` does seem more flexible and equally friendly... If @gmlewis is in agreement, we'll take a PR for it.

tamird:
heh, maximizing efficiency would certainly mean not allocating a string to begin with. In any case, this isn't terrible (nobody is writing high performance GH API tooling), so whatever is otherwise the convention in this repo should determine the shape of this API.

shurcooL:
> so whatever is otherwise the convention in this repo should determine the shape of this API.

There wasn't, we had to come up with something new. That's why this is still open for discussion.

The PR originally started by returning `[]byte`, but I made the suggestion to use `string` instead. Most other methods in this library return structs like `*PullRequest` or `*Issue`.

tamird:
Probably fine as-is, then.

gmlewis:
Yes, I'm fine with both options. It's interesting how often we end up writing code at work that jumps back and forth between `[]byte` and `string` and I remember hearing some famous Gophers from the Go team lament this fact.

PR #483, opened 2016-12-03
Add Repositories.Deployments get methods

### LINKED ISSUE #482: How to get a Deployment or DeploymentStatus by ID
https://developer.github.com/v3/repos/deployments/

The API supports looking up `Deployment` and `DeploymentStatus` by ID with a simple GET request... eg.

`https://api.github.com/repos/octocat/example/deployments/1`

This doesn't seem to be supported in the repos_deployments code.

ahume:
Add methods for getting a single Deployment and single
DeploymentStatus.

Fixes #482.

### DIFF
--- a/github/repos_deployments.go
+++ b/github/repos_deployments.go
@@ -82,6 +82,27 @@ func (s *RepositoriesService) ListDeployments(owner, repo string, opt *Deploymen
 	return *deployments, resp, err
 }
 
+// GetDeployment returns a single deployment of a repository.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/deployments/
+// Note: GetDeployment uses the undocumented GitHub API endpoint /repos/:owner/:repo/deployments/:id.
+func (s *RepositoriesService) GetDeployment(owner, repo string, deploymentID int) (*Deployment, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/deployments/%v", owner, repo, deploymentID)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	deployment := new(Deployment)
+	resp, err := s.client.Do(req, deployment)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return deployment, resp, err
+}
+
 // CreateDeployment creates a new deployment for a repository.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/deployments/#create-a-deployment
--- a/github/repos_deployments_test.go
+++ b/github/repos_deployments_test.go
@@ -35,6 +35,27 @@ func TestRepositoriesService_ListDeployments(t *testing.T) {
 	}
 }
 
+func TestRepositoriesService_GetDeployment(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/deployments/3", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, `{"id":3}`)
+	})
+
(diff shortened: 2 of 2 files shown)


ahume:
I signed it!

github/repos_deployments_test.go
```diff
@@ -87,6 +108,27 @@ func TestRepositoriesService_ListDeploymentStatuses(t *testing.T) {
...
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/deployments/3/statuses/4", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, `{"id":4}`)
+	})
+
+	deployment, _, err := client.Repositories.GetDeploymentStatus("o", "r", 3, 4)
+	if err != nil {
+		t.Errorf("Repositories.GetDeployment returned error: %v", err)
```
shurcooL: Copy-paste error here, it should be "Repositories.GetDeploymentStatus returned".

github/repos_deployments_test.go
```diff
@@ -87,6 +108,27 @@ func TestRepositoriesService_ListDeploymentStatuses(t *testing.T) {
...
+	}
+
+	want := &DeploymentStatus{ID: Int(4)}
+
+	if !reflect.DeepEqual(deployment, want) {
+		t.Errorf("Repositories.GetDeployment returned %+v, want %+v", deployment, want)
```
shurcooL: Here also.

github/repos_deployments.go
```diff
@@ -82,6 +82,26 @@ func (s *RepositoriesService) ListDeployments(owner, repo string, opt *Deploymen
 	return *deployments, resp, err
 }
 
+// GetDeployment returns a single deployment of a repository.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/deployments/
```
shurcooL: Can you please update it to point to the exact endpoint in the docs?

Actually, see https://github.com/google/go-github/issues/482#issuecomment-264770261, unless I'm missing it, it seems these endpoints are not documented.

If so, we should mention that (because if they're not documented, they may change or get removed without warning... unlikely, but possible).

ahume:
Removed `GetDeploymentStatus` as this is not supported by the API.

github/repos_deployments.go
```diff
@@ -82,6 +82,27 @@ func (s *RepositoriesService) ListDeployments(owner, repo string, opt *Deploymen
...
+// Note: GetDeployment uses the undocumented GitHub API endpoint /repos/:owner/:repo/deployments/:id.
+func (s *RepositoriesService) GetDeployment(owner, repo string, deploymentID int) (*Deployment, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/deployments/%v", owner, repo, deploymentID)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	deployment := new(*Deployment)
```
shurcooL: I believe this should be `new(Deployment)`.

github/repos_deployments.go
```diff
@@ -82,6 +82,27 @@ func (s *RepositoriesService) ListDeployments(owner, repo string, opt *Deploymen
...
+	resp, err := s.client.Do(req, deployment)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return *deployment, resp, err
```
shurcooL: And `*deployment` will become just `deployment`.

ahume:
I added the note on line 88 to cover the fact it's not currently documented. We can keep #482 open to track whether/when Github add a more explicit section on that page.

gmlewis:
Thank you, @ahume and @shurcooL!
LGTM.
I will merge this and keep #482 open to track the undocumented API.

PR #485, opened 2016-12-07
Make GetArchiveLink behave the same as other calls

svanharmelen:
All available calls in the `go-github` package close the `resp.Body` before returning.

Because the `GetArchiveLink` method doesn’t use the [`Do`](https://github.com/google/go-github/blob/master/github/github.go#L381) method, we have to make sure the body is closed here as well.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -267,8 +267,12 @@ func (s *RepositoriesService) GetArchiveLink(owner, repo string, archiveformat a
 	} else {
 		resp, err = s.client.client.Transport.RoundTrip(req)
 	}
-	if err != nil || resp.StatusCode != http.StatusFound {
-		return nil, newResponse(resp), err
+	if err != nil {
+		return nil, nil, err
+	}
+	resp.Body.Close()
+	if resp.StatusCode != http.StatusFound {
+		return nil, newResponse(resp), fmt.Errorf("unexpected status code: %s", resp.Status)
 	}
 	parsedURL, err := url.Parse(resp.Header.Get("Location"))
 	return parsedURL, newResponse(resp), err


github/repos_contents.go
```diff
@@ -267,6 +267,12 @@ func (s *RepositoriesService) GetArchiveLink(owner, repo string, archiveformat a
...
 	}
+	defer func() {
+		// Make sure to close the response body if we have one
+		if resp != nil {
+			resp.Body.Close()
+		}
+	}()
 	if err != nil || resp.StatusCode != http.StatusFound {
 		return nil, newResponse(resp), err
 	}
```
shurcooL: The above code will work, but I have a suggestion. The canonical example of doing an HTTP request looks like this:

```Go
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
body, err := ioutil.ReadAll(resp.Body)
// ...
```

Notice the error is checked first, and ensuring `resp.Body` will be closed is after. I think it'd be cleaner to rewrite it in a similar way.

First check if `err != nil` from `RoundTrip`, handle that, then handle success case. It would require breaking up the `if err != nil || resp.StatusCode != http.StatusFound {` check into two, but that seems very appropriate. So it could look something like this:

```Go
var resp *http.Response
// Use http.DefaultTransport if no custom Transport is configured
if s.client.client.Transport == nil {
	resp, err = http.DefaultTransport.RoundTrip(req)
} else {
	resp, err = s.client.client.Transport.RoundTrip(req)
}
if err != nil {
	return nil, nil, err
}
resp.Body.Close() // Close the body right away, it's unused.
if resp.StatusCode != http.StatusFound {
}
```

`defer` in front of `resp.Body.Close()` isn't even needed because we can just close it right away instead of when returning. It's never used.

What do you think?

svanharmelen: I thought about the same thing myself as well, but thought to change as little as possible... 

But I fully agree so will update the PR in a moment.

svanharmelen:
@shurcooL updated the PR according to your feedback.

github/repos_contents.go
```diff
@@ -267,7 +267,11 @@ func (s *RepositoriesService) GetArchiveLink(owner, repo string, archiveformat a
 	} else {
 		resp, err = s.client.client.Transport.RoundTrip(req)
 	}
-	if err != nil || resp.StatusCode != http.StatusFound {
+	if err != nil {
+		return nil, nil, err
+	}
+	resp.Body.Close()
+	if resp.StatusCode != http.StatusFound {
 		return nil, newResponse(resp), err
```
shurcooL: The value of `err` will always be `nil` here, so you're doing `return nil, newResponse(resp), nil` effectively.

We want to return a non-nil error here, since this is an error condition and there's no valid url being returned.

svanharmelen:
@shurcooL oeps! Seems I cannot do two things at the same time :wink: Updated!

github/repos_contents.go
```diff
@@ -267,9 +267,13 @@ func (s *RepositoriesService) GetArchiveLink(owner, repo string, archiveformat a
...
+	if err != nil {
+		return nil, nil, err
+	}
+	resp.Body.Close()
+	if resp.StatusCode != http.StatusFound {
+		return nil, newResponse(resp), fmt.Errorf("unexpected status code: %s", resp.Status)
 	}
 	parsedURL, err := url.Parse(resp.Header.Get("Location"))
-	return parsedURL, newResponse(resp), err
+	return parsedURL, newResponse(resp), nil
```
shurcooL: This line was actually correct before. You should undo the change here. It returns `err` from the line above:

```Go
```

And that `err` may be nil or not nil.

svanharmelen:
@shurcooL really sorry for all my sloppiness with this PR! Should be OK now...

svanharmelen:
@shurcooL any idea when this will be merged? Thx...

shurcooL:
This PR only needs a second review now.

@gmlewis Can You take a look at this as well?

gmlewis:
I apologize for the delay.
Thank you, @svanharmelen and @shurcooL!
LGTM. Merging.

gmlewis:
Included in https://github.com/google/go-github/commit/466070b0580728e63bd1a415e0019639e55d7148.

svanharmelen:
No problem! Thx!

PR #467, opened 2016-11-05
Update Projects API to support organizations projects

lucasalcantara:
Fixed #461 
Documented:
https://developer.github.com/changes/2016-10-27-changes-to-projects-api/

/cc @willnorris @gmlewis @shurcooL  for review

I created a const because I had to change in a lot of points the url and I think that can be easiest if we have the url in a const, because in the future this can happen again and we have to change all again.
But, of course, it is just a thought and I don't know if you agree with me.

And other point. I noticed that the code below repeats in some methods, Do you think is better to put the duplicates codes in a method? with this we will gain lines and avoid duplicates code

```
// TODO: remove custom Accept header when this API fully launches.
req.Header.Set("Accept", mediaTypeProjectsPreview)

projects := []*Project{}
resp, err := s.client.Do(req, &projects)
if err != nil {
	return nil, resp, err
}

return projects, resp, err
```

PR #487, opened 2016-12-08
Remove leading slash on Projects API

griffin-stewie:
I removed leading slash on Projects API
reasons
- unify with others
- constructing url does not work when in GitHub Enterprise environment


gmlewis:
LGTM.

shurcooL:
Thanks for catching and fixing this. It was an unintended oversight in #467.

LGTM.

gmlewis:
Thanks, @shurcooL. Merging.

Issue #486, opened 2016-12-07
Make empty responses due to 202 Accepted status code more visible.


caarlos0:
I've built a tool to gather all stats of all repos in a given org and rank the users, and, as I just found, some times github returns an empty hash on the endpoint that returns the statistics (`client.Repositories.ListContributorsStats` on go-github).

I changed `github.go` to debug the issue:

```diff
@@ -415,9 +408,13 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		if w, ok := v.(io.Writer); ok {
 			io.Copy(w, resp.Body)
 		} else {
-			err = json.NewDecoder(resp.Body).Decode(v)
+			content, _ := ioutil.ReadAll(response.Body)
+			body2 := ioutil.NopCloser(bytes.NewReader(content))
+			err = json.NewDecoder(body2).Decode(v)
 			if err == io.EOF {
 				err = nil // ignore EOF errors caused by empty response body
+			} else if err != nil {
+				println("Invalid JSON?", string(content))
 			}
 		}
 	}
```

And, it turns out, github returned an empty hash (`{}`) on a repo that wasn't changed the last year...

Is there anything doable on the go-github side? 

FYI: I'm reporting this issue to the GitHub support.

caarlos0:
@shurcooL I just did more tests, seem to be random.

https://github.com/caarlos0/org-stats 

You can exec it against the `docker` org for example.

Not sure it's a bug on github's side anymore.

shurcooL:
Can you reproduce it by hitting their RESP API directly via `curl` or something?

caarlos0:
Reply from github:

```
Hi Carlos,

Thanks for reaching out. An empty has should be valid JSON, and returning that empty hash is indeed expected in that case. The bit that you're missing is that the API is responding with a 202 status in that case, as documented here:

https://developer.github.com/v3/repos/statistics/#a-word-about-caching

When you get back a 202, that means that the statistics haven't been computed yet and you need to make another request to fetch the actual statistics.

I hope this helps!

Best,
Ivan
```

caarlos0:
So, basically, all I had to do is retry when response code is 202... Not sure if it's worth to abstract it in go-github or just leave it as is...

shurcooL:
Ah, yeah, I forgot about that.

I think the best/most appropriate thing for us to do is detect 202 in go-github and return a non-nil error that indicates Accepted status.

[...]

PR #488, opened 2016-12-08
deal with 202 Accepted status code

caarlos0:
When github returns a 202, it means that they don't have the data you asked, and it is considered a rater heavy operation to get it, so they schedule a background job to fetch and cache it, and subsequent calls to it might work.

go-github was failing to deal with that, because in some places it was trying to parse the body to a list of some object, and github actually returns a `{}`.

With this change we would be able to do the same as we do with rate limiting issues, e.g.:

```go
stats, _, err := client.Repositories.ListContributorsStats(org, repo)
if err != nil {
	if _, ok := err.(*github.AcceptedError); ok {
		// handle accepted error, likely by trying again later
	}
	// handle other errors
}
```

fixes https://github.com/google/go-github/issues/486

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,10 @@
 
 go-github is a Go client library for accessing the [GitHub API][].
 
-**Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)
-**Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)
-**Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)
-**Test Coverage:** [![Test Coverage](https://coveralls.io/repos/google/go-github/badge.svg?branch=master)](https://coveralls.io/r/google/go-github?branch=master) ([gocov report](https://drone.io/github.com/google/go-github/files/coverage.html))
+**Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)  
+**Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)  
+**Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)  
+**Test Coverage:** [![Test Coverage](https://coveralls.io/repos/google/go-github/badge.svg?branch=master)](https://coveralls.io/r/google/go-github?branch=master) ([gocov report](https://drone.io/github.com/google/go-github/files/coverage.html))  
 
 go-github requires Go version 1.4 or greater.
 
--- a/README.md
+++ b/README.md
@@ -99,20 +99,20 @@ if _, ok := err.(*github.RateLimitError); ok {
 Learn more about GitHub rate limiting at
 http://developer.github.com/v3/#rate-limiting.
 
-### Accepted status ###
+### Accepted Status ###
 
 Some endpoints may return a 202 Accepted status code, meaning that the
-information required is not yet ready and was scheduled to be gather in
-GitHub side. Methods know to behave like this are documented specifying
+information required is not yet ready and was scheduled to be gathered on
+GitHub side. Methods known to behave like this are documented specifying
 this behavior.
 
 To detect this condition of error, you can check if its type is
 `*github.AcceptedError`:
 
 ```go
-repos, _, err := client.Repositories.ListContributorsStats(org, repo)
+stats, _, err := client.Repositories.ListContributorsStats(org, repo)
 if _, ok := err.(*github.AcceptedError); ok {
-  log.Println("scheduled on github side")
+	log.Println("scheduled on github side")
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ go-github is a Go client library for accessing the [GitHub API][].
 **Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)  
 **Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)  
 **Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)  
-**Test Coverage:** [![Test Coverage](https://coveralls.io/repos/google/go-github/badge.svg?branch=master)](https://coveralls.io/r/google/go-github?branch=master) ([gocov report](https://drone.io/github.com/google/go-github/files/coverage.html))  
+**Test Coverage:** [![Test Coverage](https://coveralls.io/repos/google/go-github/badge.svg?branch=master)](https://coveralls.io/r/google/go-github?branch=master) ([gocov report](https://drone.io/github.com/google/go-github/files/coverage.html))
 
 go-github requires Go version 1.4 or greater.
 
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 go-github is a Go client library for accessing the [GitHub API][].
 
-**Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)  
-**Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)  
-**Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)  
+**Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)
+**Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)
+**Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)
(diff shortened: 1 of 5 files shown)


github/github.go
```diff
@@ -603,6 +612,8 @@ func CheckResponse(r *http.Response) error {
 			abuseRateLimitError.RetryAfter = &retryAfter
 		}
 		return abuseRateLimitError
+	case r.StatusCode == http.StatusAccepted:
+		return &AcceptedError{}
```
shurcooL: I'm tempted to suggest we turn this into a value rather than a pointer. It's an empty struct. Pointer seems unneeded and unhelpful.

```Go
return AcceptedError{}
```

github/github.go
```diff
@@ -501,6 +501,15 @@ func (r *RateLimitError) Error() string {
 		r.Response.StatusCode, r.Message, r.Rate.Reset.Time.Sub(time.Now()))
 }
 
+// AcceptedError occurs when GitHub returns 202 Accepted response with an
+// empty body, which means a job was scheduled on the GitHub side to process
+// the information needed and cache it.
+type AcceptedError struct{}
+
+func (r *AcceptedError) Error() string {
+	return "Job scheduled on GitHub side, try again later"
```
shurcooL: Error strings should not be capitalized. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

github/github.go
```diff
@@ -501,6 +501,15 @@ func (r *RateLimitError) Error() string {
 		r.Response.StatusCode, r.Message, r.Rate.Reset.Time.Sub(time.Now()))
 }
 
+// AcceptedError occurs when GitHub returns 202 Accepted response with an
+// empty body, which means a job was scheduled on the GitHub side to process
+// the information needed and cache it.
+type AcceptedError struct{}
+
+func (r *AcceptedError) Error() string {
```
shurcooL: Receiver is unused and can be removed. See rationale [here](https://dmitri.shuralyov.com/idiomatic-go#avoid-unused-method-receiver-names).

github/github.go
```diff
@@ -501,6 +501,15 @@ func (r *RateLimitError) Error() string {
 		r.Response.StatusCode, r.Message, r.Rate.Reset.Time.Sub(time.Now()))
 }
 
+// AcceptedError occurs when GitHub returns 202 Accepted response with an
+// empty body, which means a job was scheduled on the GitHub side to process
+// the information needed and cache it.
+type AcceptedError struct{}
```
shurcooL: I'm not sure if it's obvious or worth adding another sentence to the current documentation, something like:

> Technically, 202 Accepted is not a real error, it's just used to indicate that results are not ready yet, but should be available soon. The request should be repeated after some time.

The part that's missing to me is that the documentation should indicate that the this is a "retryable" error. Trying again after some time is completely normal and expected to succeed (but maybe not on the first try).

caarlos0:
@shurcooL thanks for the review! I think I fixed everything, I still not sure about the messages docs wording though.

Thanks!

github/github.go
```diff
@@ -564,14 +576,18 @@ func (e *Error) Error() string {
...
-// present.  A response is considered an error if it has a status code outside
-// the 200 range.  API error responses are expected to have either no response
+// present. A response is considered an error if it has a status code outside
+// the 200 range or equal to 202 Accepted.
+// API error responses are expected to have either no response
 // body, or a JSON response body that maps to ErrorResponse.  Any other
 // response body will be silently ignored.
 //
 // The error type will be *RateLimitError for rate limit exceeded errors,
 // and *TwoFactorAuthError for two-factor authentication errors.
```
shurcooL: Let's mention `AcceptedError` here.

```
// The error type will be *RateLimitError for rate limit exceeded errors,
// AcceptedError for 202 Accepted status codes,
// and *TwoFactorAuthError for two-factor authentication errors.
```

This makes me realize that there's actually a good reason to revert back to using a pointer to `*AcceptedError` rather than value: for consistency of type assertion code to other error types. Currently, it'd be:

```Go
switch err.(type) {
case *github.RateLimitError:
	// handle rate limit exceeded error
case *github.TwoFactorAuthError:
	// handle two-factor authentication error
case github.AcceptedError: // inconsistent
	// handle accepted error
}
```

If we revert to using a pointer to `*AcceptedError`, it becomes consistent.

```Go
switch err.(type) {
case *github.RateLimitError:
	// handle rate limit exceeded error
case *github.TwoFactorAuthError:
	// handle two-factor authentication error
case *github.AcceptedError: // consistent
	// handle accepted error
}
```

I think that's a good argument to use pointer here too. Do you agree?

shurcooL:
However, at this point I'd like to ping @willnorris or @gmlewis to take a look at this API change and see if they agree this is an improvement. It's also a slightly breaking API change (well, it changes the behavior more than API), so I wanna get their thoughts.

Also, it's kinda going against the HTTP spec which says that 2xx range status codes are "success" cases. But I think in Go it's expected/most idiomatic to return a non-nil error whenever the result you asked for cannot be returned to you.

My main argument for this change in https://github.com/google/go-github/issues/486#issuecomment-265816570 was:

> Returning a non-nil error when the result that was asked for is not returned seems very reasonable to me. Otherwise every user of this library and that function will need to check both `err` and `resp.StatusCode`, which isn't as friendly.

github/github.go
caarlos0: Yes, I agree... looks more consistent indeed

caarlos0:
> This is starting to look good.
> 
> See an inline comment. Additionally, I think this PR would benefit significantly if we can document the endpoints where AcceptedError can be returned. As far as I can tell, it'd be the 5 endpoints listed at https://developer.github.com/v3/repos/statistics/. Do you know/think any others can return 202?
> 
> An alternative is to document AcceptedError in doc.go package documentation. That'd be easier, but not as helpful IMO.

I did some search and found:

- https://developer.github.com/v3/enterprise/management_console/
- https://developer.github.com/v3/enterprise/search_indexing/
- https://developer.github.com/v3/enterprise/pre_receive_environments/
- https://developer.github.com/v3/repos/statistics/
- https://developer.github.com/v3/repos/hooks/
- https://developer.github.com/v3/repos/forks/

shurcooL:
Thanks for doing that.

Thinking more about it, I think we should document the 202 Accepted status being translated into `AcceptedError` at the package level rather than individual functions. It seems that this can apply to many endpoints, and it's a general behavior, not specific to a few endpoints. That way, as new endpoints suddenly decide to use 202 status, our documentation won't get outdated.

How about adding a new section to the package doc.go (also to README.md, the two should be in sync now) then. It could mention when `*github.AcceptedError` is returned, and a snippet how to detect it (similar to `*github.RateLimitError`).

Feel free to delay working on this until we hear from another maintainer of this library that this is a good change to go with.

github/repos_stats.go
```diff
@@ -78,7 +78,7 @@ func (w WeeklyCommitActivity) String() string {
 // starting on Sunday.
 //
 // If this is the first time these statistics are requested for the given
-// repository, this method will return a non-nil error and a status code of
+// repository, this method will return an *AcceptedError and a status code of
 // 202. This is because this is the status that github returns to signify that
 // it is now computing the requested statistics. A follow up request, after a
 // delay of a second or so, should result in a successful request.
```
shurcooL: Oh, I didn't see we already had that written up. In that case, it makes a lot of sense to update it as you've done, and copy it over to the other methods (as you've also already done).

caarlos0: Yeah, I haven't seen that before either =)

caarlos0:
@shurcooL agreed, done.

README.md
```diff
@@ -2,9 +2,9 @@
 
 go-github is a Go client library for accessing the [GitHub API][].
 
-**Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)  
-**Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)  
-**Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)  
+**Documentation:** [![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)
+**Mailing List:** [go-github@googlegroups.com](https://groups.google.com/group/go-github)
+**Build Status:** [![Build Status](https://travis-ci.org/google/go-github.svg?branch=master)](https://travis-ci.org/google/go-github)
```
shurcooL: You should undo these changes. I'm guessing your editor did them for you.

The two spaces at the end of those lines are significant, they mark a line break after each line.

Compare what the README looks like now:

image

And after this change, it's less readable:

image

shurcooL:
@caarlos0, I've applied a few typo fixes in 50886af7feb995e308e2841e67ac8d1f748664f5.

README.md
caarlos0: good catch, fixed

caarlos0:
> @caarlos0, I've applied a few typo fixes in 50886af.

thanks for that =)

I also fixed the header of the readme.

shurcooL:
To make the latest status of this PR more visible, I'm waiting to hear from @gmlewis or another maintainer on their thoughts. All my feedback so far already been addressed.

That said, I'm more confident this is a good direction after realizing that the following is already the current behavior for `ListCommitActivity` and similar, as documented:

```
// If this is the first time these statistics are requested for the given
// repository, this method will return a non-nil error and a status code of
// 202.
```

So this PR is about turning "non-nil error" into "*AcceptedError", not something more drastic as I originally thought.

gmlewis:
Yes, I like this and think it will be very helpful for developers.
LGTM.
Thank you, @caarlos0 and @shurcooL!
I'm going to tweak a couple minor nits like `s/github/GitHub/g` and then merge it.

gmlewis:
Included in https://github.com/google/go-github/commit/231da324fa98e91386bd1be76dfaf6034725621f.

caarlos0:
Thanks @gmlewis 👍

PR #490, opened 2016-12-10
Fix GetBranch response

isqua:
I want to get a time of the last commit in branch. I read [Github API: Get Branch](https://developer.github.com/v3/repos/branches/#get-branch) docs and found the [RepositoriesService.GetBranch](https://github.com/google/go-github/blob/5c45bca952b96dedace6c307caff915483013003/github/repos.go#L579) method. I called it like so:

```go
client.Repositories.GetBranch(orgName, repoName, branchName)
```

But get the following data (e.g. for this repo):

```go
{
	github.Commit {
		SHA:"5c45bca952b96dedace6c307caff915483013003",
		Author:github.CommitAuthor{},
		Committer:github.CommitAuthor{},
		Parents:[
			github.Commit{
				SHA:"5c5ff1dadae3bb0b306004d480a88e7ddc1714ca",
				URL:"https://api.github.com/repos/google/go-github/commits/5c5ff1dadae3bb0b306004d480a88e7ddc1714ca"
			}
		],
		URL:"https://api.github.com/repos/google/go-github/commits/5c45bca952b96dedace6c307caff915483013003"
	}
}
```
As you see, the Author and Committer fields has no info. I tried to call the endpoint with curl:
```
curl -H 'Accept: application/vnd.github.v3+json' \
     -H "Authorization: token ${MY_GITHUB_TOKEN}" \
     "https://api.github.com/repos/google/go-github/branches/master"
```
And get data I expected:
```js
{
  "name": "master",
  "commit": {
    "sha": "5c45bca952b96dedace6c307caff915483013003",
    "commit": {
      "author": {
        "name": "Ronak Jain",
        "email": "ronakjain@outlook.in",
        "date": "2016-12-09T18:12:45Z"
      },
      "committer": {
        "name": "Glenn Lewis",
        "email": "gmlewis@google.com",
        "date": "2016-12-09T18:58:51Z"
      },
      "message": "improve docs ...",
      "tree": {
        "sha": "de558bd3716b5304a7d40e90c4871a5173ac40a2",
        "url": "https://api.github.com/repos/google/go-github/git/trees/de558bd3716b5304a7d40e90c4871a5173ac40a2"
      },
      "url": "https://api.github.com/repos/google/go-github/git/commits/5c45bca952b96dedace6c307caff915483013003",
      "comment_count": 0
    },
    ...
  }
}
```
Looks like `commit` at the root of response is a [RepositoryCommit](https://github.com/google/go-github/blob/5c45bca952b96dedace6c307caff915483013003/github/repos_commits.go#L14).

### DIFF
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -434,7 +434,7 @@ func TestRepositoriesService_ListBranches(t *testing.T) {
 		t.Errorf("Repositories.ListBranches returned error: %v", err)
 	}
 
-	want := []*Branch{{Name: String("master"), Commit: &Commit{SHA: String("a57781"), URL: String("https://api.github.com/repos/o/r/commits/a57781")}}}
+	want := []*Branch{{Name: String("master"), Commit: &RepositoryCommit{SHA: String("a57781"), URL: String("https://api.github.com/repos/o/r/commits/a57781")}}}
 	if !reflect.DeepEqual(branches, want) {
 		t.Errorf("Repositories.ListBranches returned %+v, want %+v", branches, want)
 	}
@@ -447,7 +447,7 @@ func TestRepositoriesService_GetBranch(t *testing.T) {
 	mux.HandleFunc("/repos/o/r/branches/b", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
-		fmt.Fprint(w, `{"name":"n", "commit":{"sha":"s"}, "protected":true}`)
+		fmt.Fprint(w, `{"name":"n", "commit":{"sha":"s","commit":{"message":"m"}}, "protected":true}`)
 	})
 
 	branch, _, err := client.Repositories.GetBranch("o", "r", "b")
@@ -456,8 +456,13 @@ func TestRepositoriesService_GetBranch(t *testing.T) {
 	}
 
 	want := &Branch{
-		Name:      String("n"),
-		Commit:    &Commit{SHA: String("s")},
+		Name: String("n"),
+		Commit: &RepositoryCommit{
+			SHA: String("s"),
+			Commit: &Commit{
+				Message: String("m"),
+			},
+		},
 		Protected: Bool(true),
 	}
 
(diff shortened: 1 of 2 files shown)


isqua:
I signed it!

shurcooL:
You're absolutely right. It was surprising to me that we'd have such a glaring error, so I dug through git blame to see how it occurred (in order to make sure I'm not overlooking something).

It seems we've had the issue since its very [initial implementation](81fd27092a73ae36269b7ecad4fe32f7971c3722) in 2014. From that commit, I have a pretty good of how it happened. `ListBranches` was implemented first, and `Branch` type was created for it.

If you look at https://developer.github.com/v3/repos/branches/#list-branches, you'll see the example that the list endpoint provides has only "sha" and "url" fields inside the "commit" object. A coincidence is that both `Commit` and `RepositoryCommit` have those fields, so it's easy to use one instead of the other.

`GetBranch` was added later and seemingly no one else ran into this until now (or reported it).

Thanks for the patch!

github/repos_test.go
```diff
@@ -457,7 +457,7 @@ func TestRepositoriesService_GetBranch(t *testing.T) {
 
 	want := &Branch{
 		Name:      String("n"),
-		Commit:    &Commit{SHA: String("s")},
+		Commit:    &RepositoryCommit{SHA: String("s")},
```
shurcooL: This is good, but I think it's a good opportunity to expand the test data to include a `Commit` inside. That way it'd be a little more representative of the real data you get from GitHub, ala https://developer.github.com/v3/repos/branches/#get-branch.

Something like this:

```Go
want := &Branch{
	Name: String("n"),
	Commit: &RepositoryCommit{
		SHA: String("s"),
		Commit: &Commit{
			Message: String("m"),
		},
	},
	Protected: Bool(true),
}
```

You'd have to update the `mux.HandleFunc("/repos/o/r/branches/b"` above, of course.

Doing this is optional, if you agree it'd be an improvement for this Go package.

isqua: Well, I’ve done it: bb127587802101cdfcc97fa72ca1c799b64a5e99

shurcooL: Looks great, thanks!

gmlewis:
Thank you, @isqua and @shurcooL!
LGTM. Merging.

gmlewis:
Included in https://github.com/google/go-github/commit/a77ccc9a588d217ea8c8b069620a3a070ed201c3.

PR #491, opened 2016-12-12
Fix broken test TestRepositoriesService_CompareCommits.

shurcooL:
Previously, the test would pass if `got` and `want` were different, and fail if they were equal.

Update expected data in `TestRepositoriesService_CompareCommits`. This makes the test pass, by matching the mock JSON data that the test handler returns.

Re-arrange fields in want to match order in the JSON data.

Followup to #49.

### DIFF
--- a/github/repos_commits_test.go
+++ b/github/repos_commits_test.go
@@ -231,7 +231,7 @@ func TestRepositoriesService_CompareCommits(t *testing.T) {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


gmlewis:
LGTM. Merging.

gmlewis:
Included in https://github.com/google/go-github/commit/aacf907cae3fc95a86cabb933125c0d76f210f57.

PR #492, opened 2016-12-12
Remove RepositoryCommit.Message field.

shurcooL:
This field is unused. There is no "commit message" inside of `RepositoryCommit` (higher level GitHub object), it's only inside of `Commit` (lower level git object).

It was added accidentally in #49 and not caught via test due to #491.

Followup to #49. See https://github.com/google/go-github/pull/49/files#r92034462.


shurcooL:
> This field is unused. There is no "commit message" inside of `RepositoryCommit` (higher level GitHub object), it's only inside of `Commit` (lower level git object).

I hope I'm not wrong about that in any contexts that `RepositoryCommit` is used. But if I am, we can revert this and add a test case.

gmlewis:
LGTM.

PR #497, opened 2016-12-16
Add support for new pull request review API

### LINKED ISSUE #495: Add support for new Pull Request Reviews preview API.
Announcement: https://developer.github.com/changes/2016-12-14-reviews-api/.

Docs: https://developer.github.com/v3/pulls/reviews/.

This was announced today.

sahildua2305:
fixes #495 

@shurcooL @haya14busa please review. 🙂


github/pulls_reviews.go
```diff
@@ -17,3 +20,29 @@ type PullRequestReview struct {
...
+
+func (p PullRequestReview) String() string {
+	return Stringify(p)
+}
+
+// ListReviews lists all reviews on the specified pull request.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request
+func (s *PullRequestsService) ListReviews(owner string, repo string, number int) ([]*PullRequestReview, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pulls/%v/reviews", owner, repo, number)
```
shurcooL: You can use "%d" instead of "%v" for number (but keep "%v" for owner and repo), in order to be more consistent with the rest of the code in this repository. See:

https://github.com/google/go-github/blob/466070b0580728e63bd1a415e0019639e55d7148/github/pulls.go#L212

github/pulls_reviews.go
```diff
@@ -17,3 +20,29 @@ type PullRequestReview struct {
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req.Header.Set("Accept", mediaTypePullRequestReview)
```
shurcooL: Add a `// TODO: This header will be unnecessary when the API is no longer in preview.` comment above this line, for consistency.

github/pulls_reviews_test.go
```diff
@@ -0,0 +1,34 @@
...
+		testHeader(t, r, "Accept", mediaTypePullRequestReview)
+		fmt.Fprint(w, `[{"id":1},{"id":2}]`)
+	})
+
+	reviews, _, err := client.PullRequests.ListReviews("o", "r", 1)
+	if err != nil {
+		t.Errorf("PullRequests.ListComments returned error: %v", err)
+	}
+
+	want := []*PullRequestReview{{ID: Int(1)}, {ID: Int(2)}}
```
shurcooL: Minor thought, but I'd make it more visible there are 2 reviews here by using multiple lines, like so:

```Go
want := []*PullRequestReview{
	{ID: Int(1)},
	{ID: Int(2)},
}
```

For example, see [here](https://github.com/google/go-github/blob/466070b0580728e63bd1a415e0019639e55d7148/github/pulls_test.go#L305-L324).

github/pulls_reviews_test.go
```diff
@@ -0,0 +1,34 @@
...
+	if !reflect.DeepEqual(reviews, want) {
+		t.Errorf("PullRequests.ListComments returned %+v, want %+v", reviews, want)
```
shurcooL: You mean `PullRequests.ListReviews` here, not `ListComments`. :)

github/pulls_reviews_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	mux.HandleFunc("/repos/o/r/pulls/1/reviews", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypePullRequestReview)
+		fmt.Fprint(w, `[{"id":1},{"id":2}]`)
+	})
+
+	reviews, _, err := client.PullRequests.ListReviews("o", "r", 1)
+	if err != nil {
+		t.Errorf("PullRequests.ListComments returned error: %v", err)
```
shurcooL: Also typo here, `ListComments` should be `ListReviews`.

github/github.go
```diff
@@ -91,6 +91,9 @@ const (
 
 	// https://developer.github.com/changes/2016-09-14-Integrations-Early-Access/
 	mediaTypeIntegrationPreview = "application/vnd.github.machine-man-preview+json"
+
+	// https://developer.github.com/changes/2016-12-14-reviews-api/
+	mediaTypePullRequestReview = "application/vnd.github.black-cat-preview+json"
```
shurcooL: I know it's a mouthful, but, there should be a `Preview` stuffix at the end. These are all preview APIs, this one happens to be named "Pull Request Reviews". So:

```Go
mediaTypePullRequestReviewsPreview = ...
```

github/pulls_reviews.go
```diff
@@ -5,7 +5,10 @@
...
 package github
 
-import "time"
+import (
+	"fmt"
+	"time"
+)
 
 // PullRequestReview represents a review of a pull request.
 type PullRequestReview struct {
```
haya14busa: https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request

Some fields of list review API response are missing.

- commit_id
- html_url
- pull_request_url

You can see similar fields in PullRequestComment
https://github.com/google/go-github/blob/7b47f69ea1f1cb428354f0a6361242e9cb2cf63b/github/pulls_comments.go#L14

And "state" seems "UPPRECASE" and can be "APPROVED", "DISMISSED", "COMMENTED".

PullRequestReview  is also used in PullRequestReviewEvent https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent , so maybe we should check consistencies and give feed backs to GitHub support perhaps if the differences are not reasonable.

haya14busa: "state" also can be "CHANGES_REQUESTED"

sahildua2305:
@shurcooL @haya14busa please review again. Made changes according to the initial feedback 😄

github/pulls_reviews.go
haya14busa: Sorry, I mean, "state" can be "ACCEPTED", "DISMISSED", "CHANGES_REQUESTED" or "COMMENTED" for **Rest API** we are working with now, but "state" also can be "approved", "rejected", or "commented" for pull-request **event API**. https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent

PullRequestReview is used here as a field of event API.
https://github.com/google/go-github/blob/7b47f69ea1f1cb428354f0a6361242e9cb2cf63b/github/event_types.go#L338

So, it's not a good idea to remove existed comment, i guess.
Or, we can remove this comment completely and leave users to check GitHub API document, but i'm not sure.

And, there is a typo. `s/ACCEPTED/APPROVED/`

shurcooL:
> please review again. Made changes according to the initial feedback 😄 

The changes look great, thanks!

One suggestion. Since you're still iterating on this PR and making progress, it's better to push additional commits without rewriting history and squashing everything into a single commit. That way, when a reviewer is reviewing new changes, it's a lot easier to see what has changed since the last time they've looked. Otherwise, the reviewer has to review the entire thing from beginning, instead of just what's changed.

Squashing isn't during PR because we can easily do that at the end, when merging.

sahildua2305:
Got it! I'll push in separate commits from now.

github/pulls_reviews.go
```diff
@@ -50,3 +62,49 @@ func (s *PullRequestsService) ListReviews(owner string, repo string, number int)
...
+	}
+
+	return review, resp, err
+}
+
+// ListReviewComments lists all the comments for the specified review
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments
+func (s *PullRequestsService) ListReviewComments(owner string, number int, id int) ([]*PullRequestReviewComment, *Response, error) {
+	u := fmt.Sprintf("repos/%v/pulls/%d/reviews/%d/comments", owner, number, id)
```
shurcooL: This endpoint is missing the `repo` parameter too. We should add it, because it doesn't work otherwise.

github/pulls_reviews.go
```diff
@@ -50,3 +62,49 @@ func (s *PullRequestsService) ListReviews(owner string, repo string, number int)
 
 	return *reviews, resp, err
 }
+
+// GetReview fetches the specified pull request review.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-review
+func (s *PullRequestsService) GetReview(owner string, number int, id int) (*PullRequestReview, *Response, error) {
+	u := fmt.Sprintf("repos/%v/pulls/%d/reviews/%d", owner, number, id)
```
shurcooL: GitHub API documentation has a mistake for this endpoint and the one below. They're both missing the `repo` argument.

I've tested the API, and confirmed it's a documentation issue.

For example, this works okay:

```
curl -i -H "Accept: application/vnd.github.black-cat-preview+json" 'https://api.github.com/repos/google/go-github/pulls/497/reviews'
```

One if the IDs it returns is `13441488`. Then, if we use GitHub API docs to get that single URL, without the repo parameter, it's a 404:

```
$ curl -i -H "Accept: application/vnd.github.black-cat-preview+json" 'https://api.github.com/repos/google/pulls/497/reviews/13441488'
HTTP/1.1 404 Not Found

{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3"
}
```

But if we include repo (as one would expect... it makes no sense to specify a PR with just owner and PR number, without repo name), it works:

```
$ curl -i -H "Accept: application/vnd.github.black-cat-preview+json" 'https://api.github.com/repos/google/go-github/pulls/497/reviews/13441488'
HTTP/1.1 200 OK

{
  "id": 13441488,
  ...
}
```

sahildua2305: Thanks for noticing this @shurcooL! Even I was suspicious at the time of writing that method however, didn't delve deeper.

I have contacted GitHub about the error in their documentation. I will update the endpoint in our code.

sahildua2305:
@shurcooL I'm not sure what to do with the `event` in the [input](https://developer.github.com/v3/pulls/reviews/#input) for `CreateReview` method.

github/pulls_reviews.go
shurcooL: Excellent, thank you @sahildua2305!

shurcooL:
> I'm not sure what to do with the `event` in the [input](https://developer.github.com/v3/pulls/reviews/#input) for `CreateReview` method.

There are 15 structs in go-github that end with "Request". We want to use a similar pattern. There'll be a new struct for creating new PR reviews, named `PullRequestReviewRequest`.

For example, see the [`github.AuthorizationRequest`](https://gotools.org/github.com/google/go-github/github#AuthorizationRequest) struct, it's used as a parameter in [`AuthorizationsService.Create`](https://gotools.org/github.com/google/go-github/github#AuthorizationsService.Create) method to create authorizations. Or [`IssuesService.Create`](https://gotools.org/github.com/google/go-github/github#IssuesService.Create) and its `IssueRequest ` struct.

> How to handle comments parameter?

`comments` is documented as an "array of draft review comment objects", and:

> Draft review comments must include these values
>
> Name | Type | Description
> ------|------|------------
> path | string | The path to the file being commented on
> position | integer | The position in the file to be commented on
> body | string | Body of the comment

So I'd expect a `ReviewComment` struct (or so) with those 3 fields, and `PullRequestReviewRequest` would have a `Comments []ReviewComment` field. Does that make sense?

sahildua2305:
@shurcooL thanks for the detailed response! I think I got it. I will implement `CreateReview` accordingly and then you can review. 😄

sahildua2305:
@shurcooL @haya14busa  This is ready for review 🙂

github/pulls_reviews.go
```diff
@@ -5,15 +5,192 @@
...
+}
+
+// PullRequestReviewRequest represents a request to create a review
+type PullRequestReviewRequest struct {
+	Body     *string              `json:"body,omitempty"`
+	Event    *string              `json:"event,omitempty"`
+	Comments []DraftReviewComment `json:"comments,omitempty"`
+
+	// Message is used (optionally) while dismissing a review
+	Message *string `json:"message,omitempty"`
```
haya14busa: Hmm... I think it's better to create another struct like `PullRequestReviewDissmisalRequest` or `PullRequestReviewDissmisal` because the format is completely different.

github/pulls_reviews.go
```diff
@@ -5,15 +5,192 @@
...
+	Path     *string `json:"path,omitempty"`
+	Position *int    `json:"position,omitempty"`
+	Body     *string `json:"body,omitempty"`
+}
+
+// PullRequestReviewRequest represents a request to create a review
+type PullRequestReviewRequest struct {
+	Body     *string              `json:"body,omitempty"`
+	Event    *string              `json:"event,omitempty"`
+	Comments []DraftReviewComment `json:"comments,omitempty"`
```
haya14busa: I guess it should be `Comments []*DraftReviewComment` for consistency.

shurcooL: Are any of the fields optional? They should only be pointers with `,omitempty` if they're optional. Otherwise, plain values and no `,omitempty` would be more fitting.

haya14busa: I suggested this because I rememerbed go-github API changes from `[]T` to `[]*T` like (`[]github.PullRequestComment` -> `[]*github.PullRequestComment`) several month ago.
The pull-request is this. https://github.com/google/go-github/pull/375

But, on second thought, `PullRequestReviewRequest` is only used as a request, not as a response.
So, I agree with you. Pointer is not needed here.

github/pulls_reviews.go
```diff
@@ -5,15 +5,192 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, err
+}
+
+// SubmitReview submits a specified review on the specified pull request
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review
```
haya14busa: Just FYI: I reported a problem regarding this API as below. We cannot left "event" empty.

<details>

> If left blank, the review will be in the PENDING state.
> -- https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review

Document says we can left "event" field for "Submit a pull request review" API as "PENDING" stete, but API doesn't support it.

Response:

```
{
"message": "Invalid request. "event" wasn't supplied.",
"documentation_url": "https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review"
}
```
</details>

github/pulls_reviews.go
```diff
@@ -5,15 +5,192 @@
...
+
+	req, err := s.client.NewRequest("POST", u, review)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypePullRequestReviewsPreview)
+
+	r := new(PullRequestReview)
```
haya14busa: Let's keep variable name consistent. In `GetReview`,  The variable name is "review".
In Go, short name is preferred if it makes senses, so I think it's ok to use `r`, but in this case, i like "review".

haya14busa: Sorry, "review" is used as argument. Please ignore this.

haya14busa:
As I posted request and error response example (https://github.com/google/go-github/pull/497#discussion_r92933362), error format for review API is not same as http://developer.github.com/v3/#client-errors

"errors" is list of string for review API.

```json
{
  "message": "Validation Failed",
  "errors": [
    "Cannot return null for non-nullable field PullRequestReviewComment.position"
  ],
  "documentation_url": "https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments"
}
```

Corresponding go-github struct (`ErrorResponse`)  is here. https://github.com/google/go-github/blob/fbc15caccf54600900355e302a0fafabf7e2d733/github/github.go#L463

Because `ErrorResponse` doesn't support review API error format, error information is dropped with unmarshalling and it's really difficult to debug error using go-github.
(I tried to implement create review API in my personal project https://github.com/haya14busa/reviewdog/pull/63/ using go-github, but it was really hard to debug error... )


I'll report this problem to GitHub support, but this review API error fromat could be intentional.
If so, we have to handle this problem I guess.

github/pulls_reviews.go
```diff
@@ -5,15 +5,106 @@
...
+		return nil, resp, err
+	}
+
+	return review, resp, err
+}
+
+// ListReviewComments lists all the comments for the specified review
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments
+func (s *PullRequestsService) ListReviewComments(owner string, number int, id int) ([]*PullRequestReviewComment, *Response, error) {
```
haya14busa: Never mind. It's not go-github implementation issue. I reported to GitHub support.

sahildua2305: What's the error you're getting?

github/pulls_reviews.go
sahildua2305: Yeah, I was confused between the two approaches (making a new struct and adding an optional field in `PullRequestReviewRequest` struct). I will make a new struct 👍

github/pulls_reviews.go
```diff
@@ -5,15 +5,106 @@
...
+		return nil, resp, err
+	}
+
+	return *reviews, resp, err
+}
+
+// GetReview fetches the specified pull request review.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-review
+func (s *PullRequestsService) GetReview(owner string, number int, id int) (*PullRequestReview, *Response, error) {
```
sahildua2305: I noticed that even in `issues_comments.go`, all methods use `id`. I kept as `id` to make the code consistent. Do we want to change there as well?

github/pulls_reviews.go
```diff
@@ -5,15 +5,192 @@
...
+	ID             *int       `json:"id,omitempty"`
+	User           *User      `json:"user,omitempty"`
+	Body           *string    `json:"body,omitempty"`
+	SubmittedAt    *time.Time `json:"submitted_at,omitempty"`
+	CommitID       *string    `json:"commit_id,omitempty"`
+	HTMLURL        *string    `json:"html_url,omitempty"`
+	PullRequestURL *string    `json:"pull_request_url,omitempty"`
 
-	// State can be "approved", "rejected", or "commented".
+	// State can be "ACCEPTED", "DISMISSED", "CHANGES_REQUESTED" or "COMMENTED".
```
sahildua2305: Will add 👍

sahildua2305:
@haya14busa I have updated the PR accordingly. Please review again. (\cc: @shurcooL)

github/pulls_reviews.go
```diff
@@ -5,15 +5,194 @@
...
+	// State can be "ACCEPTED", "DISMISSED", "CHANGES_REQUESTED", "COMMENTED" or "PENDING".
 	State *string `json:"state,omitempty"`
 }
+
+func (p PullRequestReview) String() string {
+	return Stringify(p)
+}
+
+// PullRequestReviewComment represents a comment left on a pull request review.
+type PullRequestReviewComment struct {
```
haya14busa: Please add `String()` method for this type as well, including other new added structs.

```go
func (p PullRequestReviewComment) String() string {
	return Stringify(p)
}
```

In addition, we can test `String()` method in ` func TestString` https://github.com/google/go-github/blob/4dcd653e8d6b7aa32a6d50d938cecca80a8b4b61/github/strings_test.go#L87

github/pulls_reviews.go
```diff
@@ -5,15 +5,194 @@
...
+}
+
+// DraftReviewComment represents a comment part of the review.
+type DraftReviewComment struct {
+	Path     *string `json:"path,omitempty"`
+	Position *int    `json:"position,omitempty"`
+	Body     *string `json:"body,omitempty"`
+}
+
+// PullRequestReviewRequest represents a request to create a review.
```
haya14busa: This is not go-github implementation issue, but GitHub API one, but I wonder why we cannot specify `CommitID` to `DraftReviewComment` nor `PullRequestReviewRequest`.

Create a Pull Request comment API support commit_id   https://developer.github.com/v3/pulls/comments/#create-a-comment

> When passing the commit_id, use the SHA of the latest commit in the pull request or your comment may appear as "outdated" if the specified position has been modified in a subsequent commit.

As document noted, commit_id is important to specify a line because subsequent commit may overwrite it.

It might be better to contact to GitHub support about this issue.

github/pulls_reviews_test.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	want := []*PullRequestReview{
+		{ID: Int(1)},
+		{ID: Int(2)},
+	}
+	if !reflect.DeepEqual(reviews, want) {
+		t.Errorf("PullRequests.ListReviews returned %+v, want %+v", reviews, want)
+	}
+}
```
haya14busa: I found that there are test cases for invalid request for other API.

e.g. https://github.com/google/go-github/blob/4dcd653e8d6b7aa32a6d50d938cecca80a8b4b61/github/pulls_comments_test.go#L71

```go
func TestPullRequestsService_ListComments_invalidOwner(t *testing.T) {
	_, _, err := client.PullRequests.ListComments("%", "r", 1, nil)
	testURLParseError(t, err)
}
```

I think it's not so important test but it might be better to add similar tests for review API as well.

github/pulls_reviews.go
haya14busa: Error:

```
$ curl -v -H "Accept: application/vnd.github.black-cat-preview+json" 'https://api.github.com/repos/google/go-github/pulls/497/reviews/13441488/comments'
*   Trying 192.30.253.117...
* Connected to api.github.com (192.30.253.117) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: none
* TLSv1.2 (OUT), TLS header, Certificate Status (22):
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server accepted to use http/1.1
* Server certificate:
*  subject: C=US; ST=California; L=San Francisco; O=GitHub, Inc.; CN=*.github.com
*  start date: Apr  8 00:00:00 2014 GMT
*  expire date: Apr 12 12:00:00 2017 GMT
*  subjectAltName: host "api.github.com" matched cert's "*.github.com"
*  issuer: C=US; O=DigiCert Inc; OU=www.digicert.com; CN=DigiCert SHA2 High Assurance Server CA
*  SSL certificate verify ok.
> GET /repos/google/go-github/pulls/497/reviews/13441488/comments HTTP/1.1
> Host: api.github.com
> User-Agent: curl/7.50.0-DEV
> Accept: application/vnd.github.black-cat-preview+json
>
< HTTP/1.1 422 Unprocessable Entity
< Server: GitHub.com
< Date: Thu, 22 Dec 2016 08:54:11 GMT
< Content-Type: application/json; charset=utf-8
< Content-Length: 573
< Status: 422 Unprocessable Entity
< X-RateLimit-Limit: 60
< X-RateLimit-Remaining: 20
< X-RateLimit-Reset: 1482399467
< X-GitHub-Media-Type: github.black-cat-preview; format=json
< Access-Control-Expose-Headers: ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval
< Access-Control-Allow-Origin: *
< Content-Security-Policy: default-src 'none'
< Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
< X-Content-Type-Options: nosniff
< X-Frame-Options: deny
< X-XSS-Protection: 1; mode=block
< X-GitHub-Request-Id: 31680E9E:C366:4D04E48:585B94B2
<
{
  "message": "Validation Failed",
  "errors": [
    "Cannot return null for non-nullable field PullRequestReviewComment.position",
    "Cannot return null for non-nullable field PullRequestReviewComment.position",
    "Cannot return null for non-nullable field PullRequestReviewComment.position",
    "Cannot return null for non-nullable field PullRequestReviewComment.position",
    "Cannot return null for non-nullable field PullRequestReviewComment.position"
  ],
  "documentation_url": "https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments"
}
```

I already reported this problem to GitHub support but it seems it's not fixed yet.

github/pulls_reviews.go
```diff
@@ -5,15 +5,194 @@
...
+		return nil, resp, err
+	}
+
+	return review, resp, err
+}
+
+// ListReviewComments lists all the comments for the specified review.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments
+func (s *PullRequestsService) ListReviewComments(owner string, repo string, number int, reviewID int) ([]*PullRequestReviewComment, *Response, error) {
```
haya14busa: returned type should be `([]*PullRequestComment, *Response, error)`

doc: https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments

github/pulls_reviews.go
sahildua2305: Did you contact GitHub about this issue? So, we will have to wait till they fix it before we can merge it?

github/pulls_reviews.go
sahildua2305: Done

sahildua2305: Done

github/pulls_reviews_test.go
sahildua2305: Added some of the tests, not for all new methods however.

github/pulls_reviews.go
haya14busa: Not, yet. Will do.

>  So, we will have to wait till they fix it before we can merge it?

No, I think we don't have to wait fix(?) about this problem. We can add a field later after merging.

However, I think it's better to wait for the fix or response about the error format problem. https://github.com/google/go-github/pull/497#issuecomment-267810912
If it's intentional, go-github should support the new error format as well.

github/pulls_reviews_test.go
haya14busa: Looks nice! Thanks :+1: https://github.com/google/go-github/pull/497/commits/8c7f0fb864466393e53815c0b7f9e127b824ce83

Could you please adding similar tests for the rest of methods for consistency?

github/pulls_reviews.go
sahildua2305: Nice catch! Fixed

github/pulls_reviews_test.go
sahildua2305: Done

sahildua2305:
@shurcooL @haya14busa please review again!

haya14busa:
Just FYI: I reported a GitHub Review API problem regarding cannceling or adding line comments to a pending review as follow, it's not go-github one but I just let you know I reported a problem regarding GitHub Review API.

<details>

Cannot cancel a pending review nor add line comments (draft review comment) to a pending review via Review API.

It seems that we cannot canncel a pending review via API nor add line comments, although we can do it via Web UI.
In addition, we cannot create more than one pending reviews.

So, once we create a pending review via [Create a pull request review API](https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review),
we have no choice but to submit the review through [Submit a pull request review API](https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review).

I think Reivew API should support to cancel a pending review or/and add a line comments to a pending review.

Thanks in advance.

</details>

github/pulls_reviews.go
sahildua2305: I will prefer removing this comment totally. @haya14busa

haya14busa: Agree. I think it's ok to remove it.

sahildua2305:
@shurcooL please check 😃

github/pulls_reviews.go
```diff
@@ -5,15 +5,194 @@
...
+	r := new(PullRequestReview)
+	resp, err := s.client.Do(req, r)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, err
+}
+
+// SubmitReview submits a specified review on the specified pull request.
```
shurcooL: Thanks for reporting it to GitHub support @haya14busa and looking into it.

If I'm reading this right, it sounds like the problem is a purely permissions one and is on their backend. If they resolve it by changing the permission requirements, then there wouldn't be any API changes on our side, would there?

If so, I would suggest we lean towards not letting this issue block the PR from being merged.

sahildua2305:
Thanks for the review @shurcooL :)

Where do we add the corresponding TODO as you mentioned in the review?
Also, do we add preview API endpoints to [this spreadsheet](https://docs.google.com/spreadsheets/d/1wlHnoJSAN01nXUUF1JobsaKO4mJjAwZNWK8L_4ZSlls/edit#gid=0)? If so, please add. :)

github/pulls_reviews.go
haya14busa: > If so, I would suggest we lean towards not letting this issue block the PR from being merged.

Agree. I just let you know to share the problem not to make you get into the same problem.

However, as for the different problem I described in this comment https://github.com/google/go-github/pull/497#issuecomment-267810912 , it's might be better to wait the response because it's related go-github implementation.
It will be really hard to get error messages via go-github at current implementation.

But i'm not sure. We can also merge p-r and  fix it later. I leave it to you.
GitHub support kindly contacted me several times, but they didn't fix it yet.
They may overlook this problem because I reported several other problems.

May I ping them about the problem?

haya14busa:
> The only thing we should do to make the PR merge-able now is to document the known issue with the SubmitReview endpoint

No. As I said in this comment, the only thing we should care about is problem about error response format.
https://github.com/google/go-github/pull/497#discussion_r94887799

I think we can ignore other problems because they are not related with go-github implementation.

github/pulls_reviews.go
shurcooL: Pinging after some time should be fine. If they're already on it, they'll let you know.

sahildua2305: So, while we wait for them to respond/fix the issue, should we go ahead with this PR and land it? It's been hanging around for a while now 😛

shurcooL:
Ah, okay, I missed the issue with the unexpected error response format. That's worth waiting a bit more for. Do you know if it affects all 6 endpoints being added here?

> Where do we add the corresponding TODO as you mentioned in the review?

It would be as part of the method documentation. I meant something like this:

```Go
// SubmitReview submits a specified review on the specified pull request.
//
// Note: There is a known issue with this API preview endpoint, ...
// TODO: Remove this after we arrive to a resolution with GitHub support.
//
// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#submit-a-pull-request-review
func (s *PullRequestsService) SubmitReview ...
```

github/pulls_reviews.go
shurcooL: I'm okay with landing this sooner as long as the affected endpoints have the problem documented and a TODO comment saying what needs to be followed up with.

haya14busa:
> Do you know if it affects all 6 endpoints being added here?

I didn't get errors from all 6 endpoints, but list reviews and create review comments API returns the same error format as far as I remember. I assume all review API returns the same format.

Documentation looks nice :+1:

github/pulls_reviews.go
sahildua2305: Thanks, @haya14busa. Let us know once you hear back from them 😃

github/pulls_reviews.go
haya14busa: GitHub fixed this problem! :heart:

github/pulls_reviews.go
sahildua2305: @haya14busa did you get any update?

haya14busa: GitHub support tells me that he contacted to developer teams to share the problem, but no document or implementation fix yet.

haya14busa:
GitHub adds new API to delete apending review. https://developer.github.com/v3/pulls/reviews/#delete-a-pending-review

As for the error format problem, I pinged them right now again.
But, I think it's ok to merge it and fix it later if you want... It works for valid request after all ...
Maybe we can wait a week or so because I pinged them again right now and we needs to implement addigional delete pending reviews in a meantime.

sahildua2305:
@haya14busa I have added the new API endpoint too. Please review. :)

Also, since it's been one week, did you hear back from GitHub support?

sahildua2305:
@gmlewis can you please restart tests?

gmlewis:
Restarted build.

github/pulls_reviews.go
```diff
@@ -5,15 +5,220 @@
...
+//
+// Note: There is a known issue with this preview endpoint.
+// Read more about it here - https://github.com/google/go-github/pull/497#discussion_r94593877
+//
+// TODO: Remove this once we arrive at resolution with GitHub Support.
```
haya14busa: This should be fixed now. Can you confirm and remove this?

github/pulls_reviews.go
```diff
@@ -103,6 +103,29 @@ func (s *PullRequestsService) GetReview(owner string, repo string, number int, r
 	return review, resp, err
 }
 
+// DeletePendingReview deletes the specified pull request pending review.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#delete-a-pending-review
+func (s *PullRequestsService) DeletePendingReview(owner string, repo string, number int, reviewID int) (*PullRequestReview, *Response, error) {
```
haya14busa: LGTM :+1:

github/pulls_reviews.go
```diff
@@ -5,15 +5,220 @@
...
+		return nil, resp, err
+	}
+
+	return r, resp, err
+}
+
+// DismissReview dismisses a specified review on the specified pull request.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#dismiss-a-pull-request-review
+func (s *PullRequestsService) DismissReview(owner string, repo string, number int, reviewID int, review *PullRequestReviewDismissalRequest) (*PullRequestReview, *Response, error) {
```
gmlewis: `(owner, repo string, number, reviewID int, review *PullRequestReviewDismissalRequest)`

github/pulls_reviews.go
```diff
@@ -5,15 +5,220 @@
...
+	Event    *string              `json:"event,omitempty"`
+	Comments []DraftReviewComment `json:"comments,omitempty"`
+}
+
+func (r PullRequestReviewRequest) String() string {
+	return Stringify(r)
+}
+
+// PullRequestReviewDismissalRequest represents a request to dismiss a review.
+type PullRequestReviewDismissalRequest struct {
```
gmlewis: I wish we could reduce the stutter on this type... it reminds me too much of Java.  :smiley:

github/pulls_reviews.go
```diff
@@ -5,15 +5,220 @@
...
+
+	r := new(PullRequestReview)
+	resp, err := s.client.Do(req, r)
+	if err != nil {
+		return nil, resp, err
+	}
 
-	// State can be "approved", "rejected", or "commented".
-	State *string `json:"state,omitempty"`
+	return r, resp, err
```
gmlewis: `return r, resp, nil`

github/pulls_reviews.go
sahildua2305: One thing I can think of is to use `PRRequestReviewDismissalRequest` but that will disturb the consistency. 😞

gmlewis: Agreed.  I couldn't think of a better alternative.  I was just lamenting.  :smile:

sahildua2305:
@gmlewis thanks for the review. I went ahead and fixed all the minor issues. Please review again 😄

haya14busa:
@sahildua2305 

> I think it's ok to merge this p-r, but could you please adding comments to each method about this problem?
> https://github.com/google/go-github/pull/497#pullrequestreview-19529473

Could you please adding comments for know issues as well?

shurcooL:
Still LGTM, I don't have change requests from myself. However, I think the point @haya14busa is making should be addressed (it's just a documentation change).

sahildua2305:
@haya14busa according to our previous discussion, all endpoints in this PR have the aforementioned issue except list reviews and create review. Is that right?

haya14busa:
> except list reviews and create review. 

No. Create review endpoint also has the same problem.
I didn't confirm all endpoints by hand, but all review API returns the same error format.
You can check error of create review by passing wrong line number or something.

You might be able to skip `GET` endpoints because we cannot reproduce the problem again(GitHub fixed some problems for those endpoints), but just to be on the safe side, I think it's better to note the problem for each endpoints.

sahildua2305:
@haya14busa done 😄

haya14busa:
Thank you for the fix @sahildua2305 :+1: :+1: :+1:
LGTM

github/pulls_reviews.go
```diff
@@ -59,6 +59,11 @@ func (r PullRequestReviewDismissalRequest) String() string {
 
 // ListReviews lists all reviews on the specified pull request.
 //
+// TODO: Follow up with GitHub support about an issue with this method's
+// returned error format and remove this comment once it's fixed.
+// Read mroe about it here -
+// https://github.com/google/go-github/pull/497#issuecomment-267810912
```
shurcooL: There's a typo here, "mroe" should be "more".

Instead of linking to that PR with 100+ of comments, let's use a new issue to track this and point people at it. This will be much friendlier.

I've made https://github.com/google/go-github/issues/540 - please use it.

sahildua2305:
@shurcooL Done! Fixed 😄

gmlewis:
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/27c7c32b6d369610435bd2ad7b4d8554f235eb01

PR #498, opened 2016-12-19
Fix lint issues with projectID parameter name.

shurcooL:
Go style suggests that acronyms, including "ID", should have a consistent case.

> This rule also applies to "ID" when it is short for "identifier," so write
> "appID" instead of "appId".

Source: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

This change fixes the following lint issues:

	github/projects.go:119:46: method parameter projectId should be projectID
	github/projects.go:176:47: method parameter projectId should be projectID

Updates #467.

### DIFF
--- a/github/projects.go
+++ b/github/projects.go
@@ -116,8 +116,8 @@ type ProjectColumn struct {
 // ListProjectColumns lists the columns of a GitHub Project for a repo.
 //
 // GitHub API docs: https://developer.github.com/v3/projects/columns/#list-project-columns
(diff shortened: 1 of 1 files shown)


gmlewis:
Thank you, @shurcooL!
LGTM.
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/8dc18ba2c1f0efa36679c7d29dbafa09e714d880.

PR #499, opened 2016-12-21
Add the "PingEvent" structure.

octo:
GitHub docs: https://developer.github.com/webhooks/#ping-event

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -295,6 +295,18 @@ type PageBuildEvent struct {
 	Sender *User       `json:"sender,omitempty"`
 }
 
+// PingEvent is triggered when a WebHook is added to GitHub.
+//
+// GitHub docs: https://developer.github.com/webhooks/#ping-event
+type PingEvent struct {
+	// Random string of GitHub zen.
+	Zen *string `json:"zen,omitempty"`
+	// The ID of the webhook that triggered the ping.
+	HookID *string `json:"hook_id,omitempty"`
+	// The webhook configuration.
+	Hood *Hook `json:"hook,omitempty"`
+}
+
 // PublicEvent is triggered when a private repository is open sourced.
 // According to GitHub: "Without a doubt: the best GitHub event."
 // The Webhook event name is "public".
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -295,7 +295,7 @@ type PageBuildEvent struct {
 	Sender *User       `json:"sender,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


github/event_types.go
```diff
@@ -295,6 +295,18 @@ type PageBuildEvent struct {
 	Sender *User       `json:"sender,omitempty"`
 }
 
+// PingEvent is triggered when a WebHook is added to GitHub.
```
shurcooL: Minor, but for consistency with what I see above and below, call it "Webhook", not "WebHook".

github/event_types.go
```diff
@@ -295,6 +295,18 @@ type PageBuildEvent struct {
...
+//
+// GitHub docs: https://developer.github.com/webhooks/#ping-event
+type PingEvent struct {
+	// Random string of GitHub zen.
+	Zen *string `json:"zen,omitempty"`
+	// The ID of the webhook that triggered the ping.
+	HookID *string `json:"hook_id,omitempty"`
+	// The webhook configuration.
+	Hood *Hook `json:"hook,omitempty"`
```
shurcooL: s/Hood/Hook.

github/event_types.go
octo: Done.

github/event_types.go
octo: Done.

parkr:
Yay! I've been doing this [manually for a while](https://github.com/parkr/auto-reply/blob/e5c55feb1811aad9f2426fbe49891c50a3376e8f/hooks/event_types.go#L36-L38) and I appreciate that this may be added ❤️

octo:
Thanks for the review @shurcooL!

Sorry for the repeated typos, I should probably go to bed ;) I fixed the commit message, not that it matters much if you typically squash commits when merging ;)

shurcooL:
Friendly ping @gmlewis (no pun intended). Do you get notified when I [request a review](https://github.com/google/go-github/pull/499#event-902113237)?

gmlewis:
Sorry for the delay, @shurcooL. Yes, I do get notified, but unfortunately if there are comments in the thread after that, the notification gets buried, so I missed it. Poor excuse, but the only one I have in this case. :smiley:
LGTM.
Feel free to merge, or I will later tonight.

shurcooL:
> Sorry for the delay,

No problem! I was just checking. :) Thanks, I'll merge now.

Thank you @octo!

PR #501, opened 2016-12-22
Fix PullRequests.Merge incorrectly sending empty optional values.

### LINKED ISSUE #500: Optional fields of pullRequestMergeRequest should not be populated
In #456, I had given incorrect advice to the author of the PR and then merged the PR without allowing @shurcooL to review the requested changes, all of which was my fault.

As a result, the optional fields in `pullRequestMergeRequest` are being sent even if empty.

The GitHub API docs will need to be read to determine which fields are optional, and then only pass them in the JSON request if non-zero.

shurcooL:
Resolves #500.

This PR consists of 4 _logical_ commits, please review by looking at each commit individually and read its commit message for description.

First commit is unrelated but got in my way, sorry.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -301,7 +301,7 @@ type pullRequestMergeRequest struct {
 func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string, options *PullRequestOptions) (*PullRequestMergeResult, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/merge", owner, repo, number)
 
-	pullRequestBody := pullRequestMergeRequest{CommitMessage: commitMessage}
+	pullRequestBody := &pullRequestMergeRequest{CommitMessage: commitMessage}
 	if options != nil {
 		pullRequestBody.CommitTitle = options.CommitTitle
 		pullRequestBody.MergeMethod = options.MergeMethod
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -288,10 +288,10 @@ type PullRequestOptions struct {
 }
 
 type pullRequestMergeRequest struct {
-	CommitMessage *string `json:"commit_message"`
-	CommitTitle   *string `json:"commit_title,omitempty"`
-	MergeMethod   *string `json:"merge_method,omitempty"`
-	SHA           *string `json:"sha,omitempty"`
+	CommitMessage string `json:"commit_message"`
+	CommitTitle   string `json:"commit_title,omitempty"`
+	MergeMethod   string `json:"merge_method,omitempty"`
+	SHA           string `json:"sha,omitempty"`
 }
 
 // Merge a pull request (Merge Button™).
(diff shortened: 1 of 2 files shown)


shurcooL:
CI [passes](https://travis-ci.org/google/go-github/builds/185993334) for 1.7 and tip, but fails for 1.6 and older because I used [subtests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks), forgetting they did not exist [before 1.7](https://golang.org/doc/go1.7#testing). :(

I guess I have no choice but to rid of subtests then, do I?

eparis:
It LGTM.

github/pulls.go
```diff
@@ -301,11 +301,11 @@ type pullRequestMergeRequest struct {
 func (s *PullRequestsService) Merge(owner string, repo string, number int, commitMessage string, options *PullRequestOptions) (*PullRequestMergeResult, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/merge", owner, repo, number)
 
-	pullRequestBody := &pullRequestMergeRequest{CommitMessage: &commitMessage}
+	pullRequestBody := pullRequestMergeRequest{CommitMessage: commitMessage}
```
gmlewis: I don't agree with making `pullRequestBody` a value rather than a pointer.

Although this code is typically not performance critical, it is not necessary to make copies of structs (and therefore make copies of its members) on function calls and I think we want to demonstrate the more efficient version.

Additionally, this change would be inconsistent with the rest of this file using pointers, and more generally with the `go-github` package.

shurcooL: Hearing this kind of rationale for why we're using pointer is exactly why I did it, thanks @gmlewis. :)

At least I have a reason for why it's there now.

However, I'd be curious to put that claim to the test and benchmark. The struct contains 4 strings, and strings are quite inexpensive to copy (they're a [16-byte struct](https://godoc.org/reflect#StringHeader) on 64-bit systems). So if I'm not mistaken, it's a difference of copying a 4*16 = 64 byte struct on the stack, vs allocating a pointer on the heap. I don't know off the top of my head how the two would compare in performance and allocations, and I'd be curious to write a bench to learn more about that.

But, definitely not as part of this PR. I'll revert the change for this PR.

shurcooL:
I agree making 1.7 the min version supported by Travis is too early and not a viable option. I'll remove subtests to make it work everywhere.

I'll push 2 fixup commits for you to review, and once you've done that, please give me a chance to rebase them before merging, because I'd like to merge this PR as the 4 logical commits instead of squashing into 1.

shurcooL:
> I'll push 2 fixup commits for you to review, and once you've done that, please give me a chance to rebase them before merging, because I'd like to merge this PR as the 4 logical commits instead of squashing into 1.

Done. CI is all green. PTAL, @gmlewis.

gmlewis:
Very nice, @shurcooL, thank you!
LGTM.
Merge at will.

shurcooL:
Merged as 4 commits https://github.com/google/go-github/compare/64431e18df1ef20e7f08b42de4ffa3d1e90754fe...54b4d4b95f2c2abb33b9094f6a57056832bb0e3a.

PR #502, opened 2016-12-23
change PingEvent.HookID to be int, not string

jzhoucliqr:
See Hook type struct: 

type Hook struct {
	CreatedAt *time.Time             `json:"created_at,omitempty"`
	UpdatedAt *time.Time             `json:"updated_at,omitempty"`
	Name      *string                `json:"name,omitempty"`
	URL       *string                `json:"url,omitempty"`
	Events    []string               `json:"events,omitempty"`
	Active    *bool                  `json:"active,omitempty"`
	Config    map[string]interface{} `json:"config,omitempty"`
	_**ID        *int                   `json:"id,omitempty"`**_
}

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -302,7 +302,7 @@ type PingEvent struct {
 	// Random string of GitHub zen.
 	Zen *string `json:"zen,omitempty"`
 	// The ID of the webhook that triggered the ping.
-	HookID *string `json:"hook_id,omitempty"`
+	HookID *int `json:"hook_id,omitempty"`
 	// The webhook configuration.
 	Hook *Hook `json:"hook,omitempty"`
 }


shurcooL:
Hmm, seems reasonable. https://developer.github.com/webhooks/#ping-event itself doesn't specify, but https://developer.github.com/v3/repos/hooks/#get-single-hook suggests its ID is indeed a number, not string.

@octo, since you've worked on this very recently in #499, would you like to review this change?

> You can trigger a ping again by calling the [ping endpoint](https://developer.github.com/v3/repos/hooks/#ping-a-hook).

It should be possible to verify by hitting that endpoint. I don't have a webhook to test with handy, so hopefully someone else can.

jzhoucliqr:
yeah, I've tested it. It's an int. 

{"zen":"Responsive is better than fast.","hook_id":11261869,"hook":{"type":"Repository" ...}

gmlewis:
Thank you, @jzhoucliqr and @shurcooL!
LGTM.
Merging.

PR #503, opened 2016-12-27
Document that Issue struct can represent PRs too.

shurcooL:
GitHub API documentation says:

> If an issue is a pull request, the object will include a pull_request key.

Source: https://developer.github.com/v3/issues/#list-issues


gmlewis:
Very nice, @shurcooL... thank you!
LGTM.
Merge at will or I will merge this evening.

PR #505, opened 2016-12-28
Add optional Installation object to Event

bradleyfalzon:
https://developer.github.com/webhooks/#payloads has been recently updated
to include installation object to support GitHub's new Integrations.

This is a WIP, I wanted to discuss this as well as still testing to ensure this works myself (but I should be more confident in a day or so).

Looking for feedback to see if this is the correct place to add the installation object which should be common across all events (when sent to an installation). Also looking for guidance on correct testing.

### DIFF
--- a/github/activity_events_test.go
+++ b/github/activity_events_test.go
@@ -275,7 +275,7 @@ func TestActivityService_ListUserEventsForOrganization(t *testing.T) {
 	}
 }
 
-func TestActivity_EventPayload_typed(t *testing.T) {
+func TestActivityService_EventPayload_typed(t *testing.T) {
 	raw := []byte(`{"type": "PushEvent","payload":{"push_id": 1}}`)
 	var event *Event
 	if err := json.Unmarshal(raw, &event); err != nil {
@@ -291,7 +291,7 @@ func TestActivity_EventPayload_typed(t *testing.T) {
 // TestEvent_Payload_untyped checks that unrecognized events are parsed to an
 // interface{} value (instead of being discarded or throwing an error), for
 // forward compatibility with new event types.
-func TestActivity_EventPayload_untyped(t *testing.T) {
+func TestActivityService_EventPayload_untyped(t *testing.T) {
 	raw := []byte(`{"type": "UnrecognizedEvent","payload":{"field": "val"}}`)
 	var event *Event
 	if err := json.Unmarshal(raw, &event); err != nil {
@@ -299,6 +299,19 @@ func TestActivity_EventPayload_untyped(t *testing.T) {
 	}
 
 	want := map[string]interface{}{"field": "val"}
+	if !reflect.DeepEqual(event.Payload(), want) {
+		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	}
+}
+
+func TestActivityService_EventPayload_installation(t *testing.T) {
+	raw := []byte(`{"type": "PullRequestEvent","payload":{"installation":{"id":1}}}`)
+	var event *Event
+	if err := json.Unmarshal(raw, &event); err != nil {
+		t.Fatalf("Unmarshal Event returned error: %v", err)
+	}
+
+	want := &PullRequestEvent{Installation: &Installation{ID: Int(1)}}
 	if !reflect.DeepEqual(event.Payload(), want) {
 		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
 	}
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -15,9 +15,10 @@ type CommitCommentEvent struct {
 	Comment *RepositoryComment `json:"comment,omitempty"`
 
 	// The following fields are only populated by Webhook events.
-	Action *string     `json:"action,omitempty"`
-	Repo   *Repository `json:"repository,omitempty"`
-	Sender *User       `json:"sender,omitempty"`
+	Action       *string       `json:"action,omitempty"`
+	Repo         *Repository   `json:"repository,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
 }
 
 // CreateEvent represents a created repository, branch, or tag.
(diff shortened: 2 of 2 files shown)


shurcooL:
Hopefully someone more familiar with webhooks and events can correct or confirm this, but this is what I see after looking more.

There's the top level JSON object for events, as described at https://developer.github.com/v3/activity/events/#events.

image

As far as I can tell, the paragraph you've linked to at https://developer.github.com/webhooks/#payloads refers to contents of the `payload` JSON object inside the top-level event JSON object.

> In addition to the fields documented for each event, webhook **payloads** include the user who performed the event (`sender`) as well as the organization (`organization`) and/or repository (`repository`) which the event occurred on, and for an Integration's webhook may include the installation (`installation`) which an event relates to. An example is given in the `PullRequestEvent` **payload**.

Emphasis mine, but it really says that the extra fields, including the new `installation` one, are additional fields inside the **payload**, not the top-level event.

I think it should go inside the individual payload structs, such as here:

- https://github.com/google/go-github/blob/4b026b323c199177ad865434c0266e8bcd5ec83d/github/event_types.go#L17-L20
- https://github.com/google/go-github/blob/4b026b323c199177ad865434c0266e8bcd5ec83d/github/event_types.go#L38-L41

It may make sense to factor all those common extra fields into a struct which is then embedded into all payload structs. See https://play.golang.org/p/sJPU5yBX48 for example of what I mean.

bradleyfalzon:
Thanks @shurcool, and yes, you're right, unfortunately, I incorrectly tested this by testing my fork not the actual vendored copy (which I had hardcoded when testing). My apologies for the confusion.

So the way I see it:

1. Add `Installation` to each and every webhook.
1. Create `WebhookCommon` add `Installation` as a member, as they should be sent with all events when used for an installation.
1. As above, but also add `Sender`, `Organization`, and `Repository` to `WebhookCommon` and remove from each Event. This I'm not 100% certain, it's not clear whether they all are global and will this conceivably introduce backwards incompatible changes?

I'm happy to PR up anything, and I'm confident in 1 or 2 being acceptable solutions, looking to others for thoughts on 3.

I've pushed d5faeb04d58c99114c9e1a43bbd343471b245ea5 as an example of 2.

gmlewis:
I'm not convinced that adding `WebhookCommon` to each struct is the way to go, especially since it contains just a single `Installation` struct. If it were me, I would simply add the `Installation` struct to any structs where it is missing.

bradleyfalzon:
I'm fine with anything, I thought the `WebhookCommon` (or similar) was a good idea to help future proof other common members in the future.

Should I send through another commit with `Installation` added to each webhook? If so, should I just make one new test with one hook or edit every hook's test I modify to check Installation is set?

gmlewis:
Let's wait to see what @shurcooL thinks of my comment before I have you change anything, to minimize the amount of rework this might cause.

shurcooL:
> I'm not convinced that adding `WebhookCommon` to each struct is the way to go, especially since it contains just a single `Installation` struct. If it were me, I would simply add the `Installation` struct to any structs where it is missing.

I agree it's temping to just add `Installation` field everywhere, it's simpler and less disruptive.

Let me elaborate on what situation I think we should consider a common embedded struct in. It would make more sense to do that if GitHub API documentation clearly said that "all webhooks have the following additional fields." That way, having such a common struct that's embedded into all webhook structs would allow us to stay closer to the GitHub API and just update that struct whenever a new common field is added. It would be a more direct mapping to the GitHub API.

The GitHub API does say something like that:

> In addition to the fields documented for each event, webhook payloads include the user who performed the event (`sender`) as well as the organization (`organization`) and/or repository (`repository`) which the event occurred on, and for an Integration's webhook may include the installation (`installation`) which an event relates to. An example is given in the `PullRequestEvent` payload.

But it's not quite enough for us to change our approach now, for the following reasons:

1. there aren't that many fields, and they don't get added often
2. it's not very clear that absolutely _all_ payloads have those extra fields, e.g., it says only "Integration's webooks" might have `installation`.
3. it'd be a more disruptive change given that we'd want to move `sender`, `organization`, `repository` in there.

By using an embedded struct, it's possible to make it a lesser breaking API change, but it wouldn't be perfectly so.

Because of all that, I agree we should just add `installation` field to all payloads and not try to do more now. But we can re-consider the common struct later if it begins to make more sense.

bradleyfalzon:
I've force pushed these changes, don't merge yet, I've yet to test them fully locally, and I'll add a test for just one webhook unless there's further tests you'd like added.

bradleyfalzon:
Testing locally now, but how would you like this unit tested?

shurcooL:
I think a simple test similar to `TestActivity_EventPayload_typed`, except one that includes an `"installation"` object would work.

https://developer.github.com/webhooks/#payloads mentions that:

> An example is given in the [PullRequestEvent payload](https://developer.github.com/v3/activity/events/types/#webhook-payload-example-16).

So let's use a short version of that event payload that includes an installation. Something like:

```Go
func TestActivityService_EventPayload_pullRequestEvent(t *testing.T) {
	raw := []byte(`{"type": "PullRequestEvent","payload": {
		...,
		"installation": {
		  "id": 234
		}}`)
	var event *Event
	if err := json.Unmarshal(raw, &event); err != nil {
		t.Fatalf("Unmarshal Event returned error: %v", err)
	}

	want := &PushEvent{PushID: Int(1)}
	if !reflect.DeepEqual(event.Payload(), want) {
	...
```

Also, if you don't mind, can you rename the two `TestActivity_` tests to `TestActivityService_` for consistency? It's okay if not.

shurcooL:
Friendly ping. This is quite close. Let me know if the above response was not clear, or if you're still blocked on anything.

bradleyfalzon:
Terribly sorry, I was continuing to test my fork and this has completely slipped my mind, I'll update in the next couple days if not sooner.

bradleyfalzon:
Hey, I think this is done, happy to get a review when convenient.

gmlewis:
Thank you, @bradleyfalzon and @shurcooL!
LGTM.
Merging.

PR #506, opened 2017-01-05
tests/integration: Point to a better place for more information.

shurcooL:
Point to the README, section on integration tests, instead of an empty directory.


willnorris:
LGTM

PR #509, opened 2017-01-08
Update integration tests for new branch protection API.

varadarajana:
Integration tests update branches were broken since old structures were used. This has been modified. Integration folder requires ProtectionRequest structure to be available by RepositoryService. Protection also has been changed.

Resolves #507.
Updates #310.

PR #512, opened 2017-01-09
Adds required pull request reviews

alindeman:
`required_pull_request_reviews` is a recently introduced object that will soon be required:
* https://developer.github.com/v3/repos/branches/#get-branch-protection
* https://developer.github.com/v3/repos/branches/#update-branch-protection

/cc: @sr

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -523,12 +523,12 @@ type ProtectionRequest struct {
 // RequiredStatusChecks represents the protection status of a individual branch.
 type RequiredStatusChecks struct {
 	// Enforce required status checks for repository administrators.
-	IncludeAdmins *bool `json:"include_admins,omitempty"`
+	IncludeAdmins bool `json:"include_admins"`
 	// Require branches to be up to date before merging.
-	Strict *bool `json:"strict,omitempty"`
+	Strict bool `json:"strict"`
 	// The list of status checks to require in order to merge into this
 	// branch.
-	Contexts *[]string `json:"contexts,omitempty"`
+	Contexts []string `json:"contexts"`
 }
 
 // RequiredPullRequestReviews represents the protection configuration for pull requests.
--- a/github/repos.go
+++ b/github/repos.go
@@ -534,7 +534,7 @@ type RequiredStatusChecks struct {
 // RequiredPullRequestReviews represents the protection configuration for pull requests.
 type RequiredPullRequestReviews struct {
 	// Enforce pull request reviews for repository administrators.
-	IncludeAdmins *bool `json:"include_admins,omitempty"`
+	IncludeAdmins bool `json:"include_admins"`
 }
 
 // BranchRestrictions represents the restriction that only certain users or
--- a/github/repos.go
+++ b/github/repos.go
@@ -541,9 +541,9 @@ type RequiredPullRequestReviews struct {
 // teams may push to a branch.
 type BranchRestrictions struct {
 	// The list of user logins with push access.
-	Users []*User `json:"users,omitempty"`
+	Users []*User `json:"users"`
 	// The list of team slugs with push access.
-	Teams []*Team `json:"teams,omitempty"`
+	Teams []*Team `json:"teams"`
 }
 
 // BranchRestrictionsRequest represents the request to create/edit the
@@ -552,9 +552,9 @@ type BranchRestrictions struct {
 // different from the response structure.
 type BranchRestrictionsRequest struct {
 	// The list of user logins with push access.
-	Users *[]string `json:"users,omitempty"`
+	Users []string `json:"users"`
 	// The list of team slugs with push access.
-	Teams *[]string `json:"teams,omitempty"`
+	Teams []string `json:"teams"`
 }
 
 // ListBranches lists branches for the specified repository.
--- a/github/repos.go
+++ b/github/repos.go
@@ -508,14 +508,16 @@ type Branch struct {
 
 // Protection represents a repository branch's protection.
 type Protection struct {
-	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks"`
-	Restrictions         *BranchRestrictions   `json:"restrictions"`
+	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
+	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
+	Restrictions               *BranchRestrictions         `json:"restrictions"`
 }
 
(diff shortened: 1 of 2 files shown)


github/repos.go
```diff
@@ -529,6 +531,12 @@ type RequiredStatusChecks struct {
 	Contexts *[]string `json:"contexts,omitempty"`
 }
 
+// RequiredPullRequestReviews represents the protection configuration for pull requests.
+type RequiredPullRequestReviews struct {
+	// Enforce pull request reviews for repository administrators.
+	IncludeAdmins *bool `json:"include_admins,omitempty"`
```
shurcooL: `include_admins` key is [documented](https://developer.github.com/v3/repos/branches/#update-branch-protection) as:

> **Required.** Enforce required status checks for repository administrators.

Shouldn't we not use `omitempty` option for required parameters?

alindeman: Thanks for bringing this up.

I was mostly following what I thought was a convention. As another example, something like [creating a repository](https://developer.github.com/v3/repos/#create) requires a `name` in the same manner, but [`name` is `omitempty` on the `Repository` struct](https://github.com/google/go-github/blob/dcda0b96bd591fbeb40dcaddec3fb40a7fcaca34/github/repos.go#L23). It seems this pattern is pretty common in this codebase.

I also initially thought that in the absence of `omitempty`, failing to provide a value would send `null` along which might evaluate to `false` server-side, and therefore be surprising. However, on actually testing this assumption, it appears that GitHub validates that `include_admins` is strictly a boolean and will raise an error if `null` is passed. The assumption was wrong.

If you think that going forward it's a good idea for required values to, well, omit `omitempty`, I'm on board and will make it so for this property. Let me know what you think.

shurcooL: Yeah, I saw that similar fields above that are required also have omitempty, and I figured you probably just followed the pattern. I wanted to bring it up so we could discuss the best thing to do from now on.

I do think leaving out `omitempty` would be a more appropriate thing to do, and it's better to start doing it late than never. There are at least a few places where it's already done, e.g., see here:

- https://github.com/google/go-github/blob/dcda0b96bd591fbeb40dcaddec3fb40a7fcaca34/github/projects.go#L169-L170
- https://github.com/google/go-github/blob/dcda0b96bd591fbeb40dcaddec3fb40a7fcaca34/github/pulls.go#L291

What if we go a step further to increase the mapping between a required field and it always being included by making it a value rather than pointer? E.g.:

```Go
IncludeAdmins bool `json:"include_admins"` // Required.
```

alindeman: @shurcooL :+1: I like that it exposes the requirements of the API in the type itself. I've made it so.

I also pushed 7634edb to backport this convention to `RequiredStatusChecks` because it also includes similarly required parameters like `include_admins`. I realize this is a breaking change, but I feel like it's a good idea to remain consistent within the structures that the `{Create,Update}BranchProtection` functions use. Furthermore, the GitHub HTTP API is itself still beta and open to breaking changes. If you think it's a bad idea, though, I can lop that commit off.

shurcooL: > I like that it exposes the requirements of the API in the type itself. I've made it so.

Great!

> If you think it's a bad idea, though, I can lop that commit off.

Not at all, on the contrary, I think that's absolutely the way to go. Thank you for doing that.

Yes, the GitHub API is still in beta so I think it's fine to break it slightly, since it's an improvement.

/cc @varadarajana FYI, see above. You'll need to make a small change to your PR #509 as a result.

shurcooL:
Hmm, what about `BranchRestrictionsRequest`?

I can see that in https://developer.github.com/v3/repos/branches/#update-branch-protection it has 2 keys, and unlike the `required_status_checks` and `required_pull_request_reviews`, they don't have **Required.** text. However, it does say the following for all 3:

> The `required_status_checks` object must have the following keys

> The `required_pull_request_reviews` object must have the following keys

> The `restrictions` object **must** have the following keys

Emphasis mine. It kinda sounds like it's implied that they're also required... Not sure what to do here. Do you think it's worth clarifying with GitHub support, @alindeman?

shurcooL:
Another observation I want to point out. Both `Protection` and `ProtectionRequest` reuse the same structs for `RequiredStatusChecks` and `RequiredPullRequestReviews`.

It's true that the fields in those structs are documented as required for the https://developer.github.com/v3/repos/branches/#update-branch-protection endpoint, but the change from pointer to value will affect the `Protection` object used during unmarshaling. We are assuming that since the input fields are required, they should be always present as part of the output from https://developer.github.com/v3/repos/branches/#get-branch-protection endpoint too.

If that isn't the case, it would no longer be possible to detect if a zero value came from the response, or if the response didn't contain a value.

This is quite subtle, and I think it's not going to be a problem, but I just wanted to point it out for review purposes.

gmlewis:
Thank you, @alindeman and @shurcooL!
It seems to me like your interpretation of the GitHub API docs is totally reasonable and that we can run with it. We can always revisit if we discover issues.
LGTM.
I'll let this sit for a day if there are any other comments, then merge.

alindeman:
@shurcooL I didn't get to look into your questions yesterday like I hoped, but I'll give these a look later today my time (US East)

alindeman:
> I can see that in https://developer.github.com/v3/repos/branches/#update-branch-protection it has 2 keys, and unlike the required_status_checks and required_pull_request_reviews, they don't have Required. text. However, it does say the following for all 3:
> 
> The required_status_checks object must have the following keys
> The required_pull_request_reviews object must have the following keys
> The restrictions object must have the following keys
> Emphasis mine. It kinda sounds like it's implied that they're also required... Not sure what to do here. Do you think it's worth clarifying with GitHub support, @alindeman?

I explored the API by sending it some different permutations of requests and found that, you're right, `users` and `teams` are required if `restrictions` is passed a non-`null` value. And if it's passed, the response always contains the `users` and `teams` as well. I'll make them non-pointers as well in an upcoming commit.

alindeman:
> If that isn't the case, it would no longer be possible to detect if a zero value came from the response, or if the response didn't contain a value.
> 
> This is quite subtle, and I think it's not going to be a problem, but I just wanted to point it out for review purposes.

Agreed. While I think it's true that it would be possible for the GitHub API to return only a subset of the required keys in responses, I think in practice it would be a bug given the constraints they've documented and appear to be enforcing server-side with JSON Schema.

gmlewis:
LGTM.
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/5e7fada1f17c60816ececcbc72ba183dbe7a44fb

PR #513, opened 2017-01-12
Changed "event" by "action" in issues events

yograterol:
When the IssueEvent is used the event field is empty because the json payload has "action" and not "event" as key.

https://developer.github.com/v3/activity/events/types/#webhook-payload-example-8

### DIFF
--- a/github/issues_events.go
+++ b/github/issues_events.go
@@ -57,7 +57,7 @@ type IssueEvent struct {
 	//     head_ref_deleted, head_ref_restored
 	//       The pull request’s branch was deleted or restored.
 	//
-	Event *string `json:"event,omitempty"`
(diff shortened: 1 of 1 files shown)


yograterol:
I signed it!

shurcooL:
This change doesn't seem to be valid. It would break the https://developer.github.com/v3/issues/events/#list-events-for-an-issue endpoint.

shurcooL:
I think you've mixed up `IssueEvent` and `IssuesEvent`.

See https://github.com/google/go-github/blob/dcda0b96bd591fbeb40dcaddec3fb40a7fcaca34/github/event_types.go#L201.

yograterol:
@shurcooL you're right, thanks!

Issue #515, opened 2017-01-12
Remove preview header for Traffic API


gmlewis:
Announcement:
https://developer.github.com/changes/2016-12-28-end-traffic-api-preview/

varadarajana:
Please assign this to me, I am almost done with integration test issues.

PR #517, opened 2017-01-14
Remove preview header for traffic API

davidducros:
The traffic API is now formally released and a preview header is no longer required.

https://developer.github.com/changes/2016-12-28-end-traffic-api-preview/


davidducros:
This fixes #515

PR #518, opened 2017-01-14
(#515) Remove Preview header for Traffic API

rtscowen:
Remove preview header for Traffic API.


rtscowen:
Fixes 515

shurcooL:
Thanks, but it looks like #517 was created just before this PR.

The good news is both PRs are identical, so that helps confirm this is a valid change.

PR #519, opened 2017-01-17
Document required branch protection fields.

shurcooL:
All of those fields are documented as required at https://developer.github.com/v3/repos/branches/#update-branch-protection.

In #512, we changed the API to better reflect that. However, it should still be documented that the fields are required. This is consistent with the other required fields that are also documented as required.

Additionally, note that empty slices must be non-`nil`. From running integration tests, I've seen that passing `nil` slices (which get encoded as `null` via `encoding/json`) resulted in invalid request errors such as:

	For 'anyOf/1', {"users"=>nil, "teams"=>nil} is not a null. []

Follows #512.

/cc @alindeman since you worked on #512 where we discussed this.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -522,18 +522,18 @@ type ProtectionRequest struct {
 
 // RequiredStatusChecks represents the protection status of a individual branch.
 type RequiredStatusChecks struct {
-	// Enforce required status checks for repository administrators.
+	// Enforce required status checks for repository administrators. (Required.)
 	IncludeAdmins bool `json:"include_admins"`
-	// Require branches to be up to date before merging.
+	// Require branches to be up to date before merging. (Required.)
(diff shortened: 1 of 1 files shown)


shurcooL:
An alternative solution to consider is for us to automatically convert nil slices into empty (non-nil) slices when sending requests. That way we don't have to document that nil slices shouldn't be used.

I suspect doing that could be slightly better, but I'm also okay with the current solution if others think it's fine.

gmlewis:
I think that documenting this is excellent.  If someone would like to automatically convert the nil slices to non-nil empty slices and remove the comments, that might make for a nice starter project and I would approve that too.

shurcooL:
Sounds good. I'll leave this for a few days in case @alindeman would like to leave any comments, then merge.

PR #520, opened 2017-01-17
Add support for Commit Search endpoint

### LINKED ISSUE #508: Add support for Commit Search preview API.
Announcement: https://developer.github.com/changes/2017-01-05-commit-search-api/.

Docs: https://developer.github.com/v3/search/#search-commits.

This was announced today.

sahildua2305:
Added support for this new preview endpoint which lets users
search for commits based on various conditions.

GitHub announcement -
https://developer.github.com/changes/2017-01-05-commit-search-api/
Docs - https://developer.github.com/v3/search/#search-commits

Fixes #508.

\cc @shurcooL @gmlewis

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -183,13 +183,14 @@ func (s *SearchService) search(searchType string, query string, opt *SearchOptio
 	}
 
 	switch {
+	case searchType == "commits":
+		// Accept header for search commits preview endpoint
+		// TODO: remove custom Accept header when this API fully launches.
+		req.Header.Set("Accept", mediaTypeCommitSearchPreview)
 	case opt != nil && opt.TextMatch:
 		// Accept header defaults to "application/vnd.github.v3+json"
 		// We change it here to fetch back text-match metadata
 		req.Header.Set("Accept", "application/vnd.github.v3.text-match+json")
-	case searchType == "commits":
-		// Accept header for search commits preview endpoint
-		req.Header.Set("Accept", mediaTypeCommitSearchPreview)
 	}
 
 	return s.client.Do(req, result)
--- a/github/search.go
+++ b/github/search.go
@@ -55,6 +55,7 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 	return result, resp, err
 }
 
+// SingleCommitResult represents a commit object as returned in commit search endpoint response.
 type SingleCommitResult struct {
 	Hash           *string     `json:"hash,omitempty"`
 	Message        *string     `json:"message,omitempty"`
--- a/github/search.go
+++ b/github/search.go
@@ -58,11 +58,11 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 type SingleCommitResult struct {
 	Hash           *string     `json:"hash,omitempty"`
 	Message        *string     `json:"message,omitempty"`
-	AuthorID       int         `json:"author_id,omitempty"`
+	AuthorID       *int        `json:"author_id,omitempty"`
 	AuthorName     *string     `json:"author_name,omitempty"`
 	AuthorEmail    *string     `json:"author_email,omitempty"`
 	AuthorDate     *string     `json:"author_date,omitempty"`
-	CommitterID    int         `json:"committer_id,omitempty"`
+	CommitterID    *int        `json:"committer_id,omitempty"`
 	CommitterName  *string     `json:"committer_name,omitempty"`
 	CommitterEmail *string     `json:"committer_email,omitempty"`
 	CommitterDate  *string     `json:"committer_date,omitempty"`
--- a/github/search.go
+++ b/github/search.go
@@ -55,27 +55,28 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 	return result, resp, err
 }
 
-// SingleCommitResult represents a commit object as returned in commit search endpoint response.
-type SingleCommitResult struct {
+// CommitsSearchResult represents the result of a commits search.
+type CommitsSearchResult struct {
+	Total             *int           `json:"total_count,omitempty"`
+	IncompleteResults *bool          `json:"incomplete_results,omitempty"`
+	Commits           []CommitResult `json:"items,omitempty"`
+}
+
+// CommitResult represents a commit object as returned in commit search endpoint response.
+type CommitResult struct {
 	Hash           *string     `json:"hash,omitempty"`
 	Message        *string     `json:"message,omitempty"`
 	AuthorID       *int        `json:"author_id,omitempty"`
 	AuthorName     *string     `json:"author_name,omitempty"`
 	AuthorEmail    *string     `json:"author_email,omitempty"`
-	AuthorDate     *string     `json:"author_date,omitempty"`
+	AuthorDate     *Timestamp  `json:"author_date,omitempty"`
 	CommitterID    *int        `json:"committer_id,omitempty"`
 	CommitterName  *string     `json:"committer_name,omitempty"`
 	CommitterEmail *string     `json:"committer_email,omitempty"`
-	CommitterDate  *string     `json:"committer_date,omitempty"`
+	CommitterDate  *Timestamp  `json:"committer_date,omitempty"`
 	Repository     *Repository `json:"repository,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


github/search.go
```diff
@@ -54,6 +55,36 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
...
 }
 
+type SingleCommitResult struct {
+	Hash           *string     `json:"hash,omitempty"`
+	Message        *string     `json:"message,omitempty"`
+	AuthorID       int         `json:"author_id,omitempty"`
+	AuthorName     *string     `json:"author_name,omitempty"`
+	AuthorEmail    *string     `json:"author_email,omitempty"`
+	AuthorDate     *string     `json:"author_date,omitempty"`
+	CommitterID    int         `json:"committer_id,omitempty"`
```
shurcooL: Any reason these two ints (`AuthorID` and `CommitterID`) aren't pointers, like everywhere else?

sahildua2305: Oops! No, that's a bug, not a feature.

github/search.go
```diff
@@ -54,6 +55,36 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 	return result, resp, err
 }
 
+type SingleCommitResult struct {
```
shurcooL: golint issue here, exported struct `SingleCommitResult` should have documentation.

sahildua2305: Fixed

sahildua2305:
@shurcooL can you please review again? 🙂

github/search.go
```diff
@@ -54,6 +55,37 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
...
+	Hash           *string     `json:"hash,omitempty"`
+	Message        *string     `json:"message,omitempty"`
+	AuthorID       *int        `json:"author_id,omitempty"`
+	AuthorName     *string     `json:"author_name,omitempty"`
+	AuthorEmail    *string     `json:"author_email,omitempty"`
+	AuthorDate     *string     `json:"author_date,omitempty"`
+	CommitterID    *int        `json:"committer_id,omitempty"`
+	CommitterName  *string     `json:"committer_name,omitempty"`
+	CommitterEmail *string     `json:"committer_email,omitempty"`
+	CommitterDate  *string     `json:"committer_date,omitempty"`
```
shurcooL: Are you sure `AuthorDate` and `CommitterDate` should be `*string` instead of `*Timestamp`?

github/search.go
```diff
@@ -54,6 +55,37 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 	return result, resp, err
 }
 
+// SingleCommitResult represents a commit object as returned in commit search endpoint response.
+type SingleCommitResult struct {
```
shurcooL: What do you think of following the style set forward by `CodeResult` and name this `CommitResult`?

```Go
// CodeResult represents a single search result.
type CodeResult struct {
```

Source: https://github.com/google/go-github/blob/c9c37fd6ce0247943d1aa34fc3813c62c5cb0225/github/search.go#L115-L116.

The prefix `Single` doesn't seem to be needed, is it? Unless it's needed, it's better to remove it to have a simpler struct name.

sahildua2305:
Thanks @shurcooL, I will make the changes as suggested.

github/search.go
sahildua2305: No, it should be `*Timestamp` only. Changed it.

sahildua2305:
@shurcooL I have made changes as you suggested.

sahildua2305:
@shurcooL can you please this again?

github/search.go
```diff
@@ -182,15 +183,14 @@ func (s *SearchService) search(searchType string, query string, opt *SearchOptio
...
-		// We change it here to fetch back text-match metadata
-		req.Header.Set("Accept", "application/vnd.github.v3.text-match+json")
+	switch {
+		case opt != nil && opt.TextMatch:
+			// Accept header defaults to "application/vnd.github.v3+json"
+			// We change it here to fetch back text-match metadata
+			req.Header.Set("Accept", "application/vnd.github.v3.text-match+json")
+		case searchType == "commits":
+			// Accept header for search commits preview endpoint
+			req.Header.Set("Accept", mediaTypeCommitSearchPreview)
```
shurcooL: Two things here.

1. Add a `// TODO: remove custom Accept header when this API fully launches.` comment.

2. Now that these cases are in the same switch, it's a little easier to reason about them. I think we should give the `searchType == "commits"` higher priority than `opt != nil && opt.TextMatch`. It may not even be valid to set `opt.TextMatch: true` for a commit search, but if someone does, it would fail to return any results at all because the custom accept header wouldn't be set.

    So, let's be safe and put the `searchType == "commits"` case on top so that the API preview header is always set when doing a commit search.

sahildua2305: @shurcooL made the changes. Please have a look. :)

github/search.go
```diff
@@ -54,6 +55,37 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 	return result, resp, err
 }
 
+// CommitsSearchResult represents the result of a commits search.
+type CommitsSearchResult struct {
+	Total             *int           `json:"total_count,omitempty"`
+	IncompleteResults *bool          `json:"incomplete_results,omitempty"`
+	Commits           []CommitResult `json:"items,omitempty"`
```
gmlewis: Please change this to:
```go
Commits           []*CommitResult `json:"items,omitempty"`
```

Also, wouldn't we want `Items` here?

github/search.go
```diff
@@ -54,6 +55,37 @@ func (s *SearchService) Repositories(query string, opt *SearchOptions) (*Reposit
 	return result, resp, err
 }
 
+// CommitsSearchResult represents the result of a commits search.
+type CommitsSearchResult struct {
+	Total             *int           `json:"total_count,omitempty"`
```
gmlewis: Wouldn't it reduce confusion if the Go field name matched the JSON field name?
In other words, use `TotalCount` here?

github/search_test.go
```diff
@@ -46,6 +46,37 @@ func TestSearchService_Repositories(t *testing.T) {
...
+	opts := &SearchOptions{Sort: "author-date", Order: "desc"}
+	result, _, err := client.Search.Commits("blah", opts)
+	if err != nil {
+		t.Errorf("Search.Commits returned error: %v", err)
+	}
+
+	want := &CommitsSearchResult{
+		Total:             Int(4),
+		IncompleteResults: Bool(false),
+		Commits:           []CommitResult{{Hash: String("random_hash1")}, {Hash: String("random_hash2")}},
```
gmlewis: `[]*CommitResult...`

github/search.go
shurcooL: In a normal situation, I would full you agree.

However, please look at the other similar structs in this file @gmlewis. See `RepositoriesSearchResult`, `CodeSearchResult`, etc.

They consistently call total_count as Total.

github/search.go
shurcooL: Again, the current version of not using Items as the name, and not having pointers is consistent with other structs.

However, we can probably deviate on the pointer. What do you think?

github/search.go
gmlewis: OK, SGTM.

github/search.go
gmlewis: I definitely think the pointer version is better despite any inconsistencies, especially for new structs. See #180 for discussion.

As for `Commits`, that's fine with me.

github/search.go
shurcooL: If you think it's worth changing (outside this PR), we can make a separate issue/PR to discuss that. But I suspect the cost (breaking API) wouldn't be worth the small benefit (more consistency).

github/search.go
shurcooL: SGTM. @sahildua2305, please go ahead with this `[]*CommitResult` change.

sahildua2305: @gmlewis @shurcooL done 👍

gmlewis:
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/3b59f32154c06d3e31093f10e58700aeaf039405

sahildua2305:
Thanks @gmlewis

PR #523, opened 2017-01-19
Redact secret in error from HTTP client Do.

### LINKED ISSUE #522: Errors caused by internal HTTP client's Do() don't have client_secret redacted.
`go-github` tries to redact `client_secret` URL query parameter value from errors that it returns, to prevent leaking sensitive information in case they errors are exposed to users. See:

- https://github.com/google/go-github/blob/173908142398084e14d7f14c39924f0004d43679/github/github.go#L552-L564
- https://github.com/google/go-github/blob/173908142398084e14d7f14c39924f0004d43679/github/github.go#L497
- https://github.com/google/go-github/blob/173908142398084e14d7f14c39924f0004d43679/github/github.go#L613
- https://github.com/google/go-github/blob/173908142398084e14d7f14c39924f0004d43679/github/github.go#L425

However, there exists a possible path where an error is returned without sanitization. See:

https://github.com/google/go-github/blob/173908142398084e14d7f14c39924f0004d43679/github/github.go#L409

The underlying HTTP client's Do() can return an error, for example, if GitHub servers are down or having connectivity issues. The returned error can be a `*url.Error` which contains the original query URL with the secret.

E.g.:

image

(Actual status.github.com graph spike from last night.)

```
Get https://api.github.com/...?client_id=xxx&client_secret=yyy: dial tcp 192.30.253.117:443 i/o timeout
```

Since we're already sanitizing those errors in most cases, for consistency, we should also do it here, to prevent `client_secret` being included in the returned error and potentially displayed to users.

shurcooL:
Fixes #522. See that issue and commit messages for explanation.

I see that CI succeeds on Go 1.7 and newer, but fails on 1.6 and older. I'm going to think about how to best address that, and update this PR.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -550,7 +550,7 @@ func (r *AbuseRateLimitError) Error() string {
 }
 
 // sanitizeURL redacts the client_secret parameter from the URL which may be
-// exposed to the user, specifically in the ErrorResponse error message.
+// exposed to the user.
 func sanitizeURL(uri *url.URL) *url.URL {
 	if uri == nil {
 		return nil
--- a/github/github.go
+++ b/github/github.go
@@ -406,6 +406,12 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 
 	resp, err := c.client.Do(req)
 	if err != nil {
+		if e, ok := err.(*url.Error); ok {
+			if url, err := url.Parse(e.URL); err == nil {
+				e.URL = sanitizeURL(url).String()
+				return nil, e
+			}
+		}
 		return nil, err
 	}
 
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -392,8 +392,7 @@ func TestDo_sanitizeURL(t *testing.T) {
 	if err == nil {
 		t.Fatal("Expected error to be returned.")
 	}
-	if !strings.Contains(err.Error(), "client_secret=REDACTED") ||
-		strings.Contains(err.Error(), "client_secret=secret") {
+	if strings.Contains(err.Error(), "client_secret=secret") {
 		t.Errorf("Do error contains secret, should be redacted:\n%q", err)
 	}
 }
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -375,6 +375,29 @@ func TestDo_redirectLoop(t *testing.T) {
 	}
 }
 
+// Test that an error caused by the internal http client's Do() function
+// does not leak the client secret.
+func TestDo_sanitizeURL(t *testing.T) {
+	tp := &UnauthenticatedRateLimitedTransport{
+		ClientID:     "id",
+		ClientSecret: "secret",
+	}
+	unauthedClient := NewClient(tp.Client())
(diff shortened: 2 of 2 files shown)


shurcooL:
When working on this, I ended up drafting up a few alternative solutions. I'll post them here, because I think they're worth considering. I wasn't sure which one to use, and went with the simplest, most specific fix to the exact issue I've seen (instead of trying to be more general).

But I might want to rethink that.

### Alternative A

```Go
	...
	resp, err := c.client.Do(req)
	if err != nil {
		return nil, sanitizeError(err)
	}
	...
}

// sanitizeError redacts the client_secret parameter from the URL in error
// which may be exposed to the user.
func sanitizeError(err error) error {
	// Check if it's *url.Error with a URL that parses successfully.
	if e, ok := err.(*url.Error); ok {
		if url, err := url.Parse(e.URL); err == nil {
			e.URL = sanitizeURL(url).String()
			return e
		}
	}

	// Check if the error string contains "client_secret=".
	e := err.Error()
	if i := strings.Index(e, "client_secret="); i != -1 {
		return errors.New(e[:i] + "client_secret=REDACTED...")
	}

	return err
}
```

### Alternative B

I also considered factoring out the sanitization of `*url.Error` into a helper func:

```Go
	...
	resp, err := c.client.Do(req)
	if err != nil {
		if err, ok := err.(*url.Error); ok {
			return nil, sanitizeURLError(err)
		}
		return nil, err
	}
	...
}

// sanitizeURLError redacts the client_secret parameter from the URL in url.Error
// which may be exposed to the user.
func sanitizeURLError(e *url.Error) error {
	url, err := url.Parse(e.URL)
	if err != nil {
		return e
	}
	e.URL = sanitizeURL(url).String()
	return e
}
```

shurcooL:
Ok, I've made CI pass on older version of Go with bf43674e694881dccec129bdfcf8db58cf504bf6. I still want to think more about this and see what's the best overall solution.

Meanwhile, @gmlewis could you review this and let me know your thoughts for how exactly to best resolve #522?

gmlewis:
Thank you for the thoughtful analysis and consideration here, @shurcooL!
I'm thinking that I like your current solution the very best. It is straightforward, easy to read, and easy to understand... with "Alternative B" being a very close second choice.

I'm not sure we want to go with Alternative A for fear of chopping off extra information, although leaking the client secret is equally bad (or worse possibly?).  I think we don't have a known situation at this point where we need something like Alternative A... but I could easily be wrong here.

LGTM.

shurcooL:
> I'm not sure we want to go with Alternative A for fear of chopping off extra information ... I think we don't have a known situation at this point where we need something like Alternative A...

Agreed with this. If someone can find a reproducible way to leak the secret that the current approach doesn't handle, they should report that and we can deal with it then. I'd rather not start doing guesswork and jumping ahead. This PR is already an improvement over current state, so it's safe to merge as is.

> Thank you for the thoughtful analysis and consideration here, @shurcooL!
I'm thinking that I like your current solution the very best. It is straightforward, easy to read, and easy to understand... with "Alternative B" being a very close second choice.

Sounds good. I agree, and I'll merge it as is now. If we need to follow this up in the future, it's very likely we'll want to pull it out into a func then, but it doesn't need to be done now.

I'll merge this and #522 will be resolved.

PR #527, opened 2017-01-21
Add ListPendingOrgInvitations method and Invitation.Inviter field.

### LINKED ISSUE #514: Support update to organization and team invitations APIs
Announcement:
https://developer.github.com/changes/2017-01-12-organization-team-invitation-update/

varadarajana:
This change is made to incorporate changes to GitHub for Issue #514 
Here I have tried to change the test case to include the entire response.
I have not been able to compare time.Time field in Invitation struct. I need to work on the Date time conversions.

Resolves #514.


varadarajana:
@shurcooL @gmlewis Can you please review this?

varadarajana:
The golang documentation of time, requires structs to access as time.Time and not *time.Time. So changed it as it.
The URL https://developer.github.com/v3/orgs/teams/#list-pending-team-invitations does not provide correct date format I think, it is not consistent with other API in the documentation. I have changed the return JSON to the type consistent and mentioned in the documentation of GitHub.

varadarajana:
I have now added support for Orgs List Members Invitation also. Now this contains both List Members as well as Team Invitations.

github/orgs_members.go
```diff
@@ -270,3 +270,33 @@ func (s *OrganizationsService) RemoveOrgMembership(user, org string) (*Response,
 
 	return s.client.Do(req, nil)
 }
+
+//ListPendingOrgInvitations returns a list of contains a role field which refers to the
+//Organization Invitation role and will be one of the following values:
+//direct_member, admin, billing_manager, hiring_manager, or reinstate.
+//If the invitee is not a GitHub member, the login field in the return hash will be null.
```
shurcooL: Don't forget to include a space between the text and `//`.

Rationale: https://dmitri.shuralyov.com/idiomatic-go#comments-for-humans-always-have-a-single-space-after-the-slashes

github/orgs_members.go
```diff
@@ -270,3 +270,33 @@ func (s *OrganizationsService) RemoveOrgMembership(user, org string) (*Response,
...
+//
+// GitHub API docs:
+// https://developer.github.com/v3/orgs/members/#list-pending-organization-invitations
```
shurcooL: There's no benefit in putting the long URL on a new line. We're not hard-limited to the width of a [punch card](https://twitter.com/rob_pike/status/563801489190043648), it's fine to keep it on the same line.

github/orgs_members.go
shurcooL: This documentation is hard to read and strangely detailed (it refers to specifics of the returned values inside `Invitation` struct).

We should either omit it altogether, just say `// ListPendingOrgInvitations returns a list of pending organization invitations.`, or update it to be specific to go-github. The returned thing is `[]Invitation`, a slice of invitations, not a hash. There's no null, it's nil. Etc.

It would make sense to me to move the documentation of `role` to the `Invitation` struct itself, not here. Unless it's specific to this method, is it?

github/orgs_teams.go
```diff
@@ -46,17 +46,46 @@ func (t Team) String() string {
...
-	Email     *string    `json:"email,omitempty"`
-	Role      *string    `json:"role,omitempty"`
-	CreatedAt *time.Time `json:"created_at,omitempty"`
+	ID    *int    `json:"id,omitempty"`
+	Login *string `json:"login,omitempty"`
+	Email *string `json:"email,omitempty"`
+	Role  *string `json:"role,omitempty"`
+	//golang documentation states
+	//Programs using times should typically store and pass them as values, not pointers.
+	//That is, time variables and struct fields should be of type time.Time, not *time.Time
```
shurcooL: This comment shouldn't be here as part of the final PR.

I will respond to it.

While it's true that it's fine to use `time.Time` as a value, the reason we're using it as a point here is because of JSON encoding. We want to be able to tell (sometimes) that a field is absent by checking if its value is nil.

So, we want to keep `CreatedAt` as a pointer to `*time.Time`. If you want to discuss this further, I suggest opening a new issue, because it's outside of the scope of this PR.

github/orgs_teams.go
```diff
@@ -46,17 +46,46 @@ func (t Team) String() string {
...
+	CreatedAt time.Time `json:"created_at,omitempty"`
+	Inviter   *Inviter  `json:"inviter,omitempty"`
 }
 
 func (i Invitation) String() string {
 	return Stringify(i)
 }
 
+// Inviter represents the entity introduced in team and org invitations
```
shurcooL: A period is missing at the end of this sentence.

Also, you can write the full word "organization", there's no need for a short form here.

github/orgs_teams.go
shurcooL: Okay, question, have you considered using `*Timestamp` here instead of `*time.Time`?

Can you be more specific about the problem you're seeing with `*time.Time` as is?

github/orgs_members.go
varadarajana: @shurcooL  Yes, this has been changed.

github/orgs_members.go
varadarajana: @shurcooL this is done.

github/orgs_teams.go
varadarajana: @shurcooL in fact this time i tried pointer to time.Time, and it worked. i assigned it to a variable and then passed its address.

github/orgs_teams.go
varadarajana: @shurcooL this is done.

github/orgs_teams.go
```diff
@@ -44,19 +44,46 @@ func (t Team) String() string {
...
+	Following_url       *string `json:"following_url,omitempty"`
+	Gists_url           *string `json:"gists_url,omitempty"`
+	Starred_url         *string `json:"starred_url,omitempty"`
+	Subscriptions_url   *string `json:"subscriptions_url,omitempty"`
+	Organizations_url   *string `json:"organizations_url,omitempty"`
+	Repos_url           *string `json:"repos_url,omitempty"`
+	Events_url          *string `json:"events_url,omitempty"`
+	Received_events_url *string `json:"received_events_url,omitempty"`
+	Type                *string `json:"type,omitempty"`
+	Site_admin          bool    `json:"site_admin,omitempty"`
```
shurcooL: Most of these variables need to be renamed to follow Go style.

See:

- https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps
- https://github.com/golang/go/wiki/CodeReviewComments#initialisms

So, it should be `SiteAdmin`, not `Site_admin`. `URL`, not `Url`. `HTMLURL`, not `Html_url`, etc.

github/orgs_members_test.go
```diff
@@ -353,3 +354,81 @@ func TestOrganizationsService_RemoveOrgMembership(t *testing.T) {
...
+			]`)
+	})
+
+	opt := &ListOptions{Page: 1}
+	invitations, _, err := client.Organizations.ListPendingOrgInvitations(1, opt)
+	if err != nil {
+		t.Errorf("Organizations.ListPendingOrgInvitations returned error: %v", err)
+	}
+
+	tm := time.Date(2017, 01, 21, 0, 0, 0, 0, time.UTC)
```
varadarajana: @shurcooL yes, tm was not proper, i have set it as createdAt

github/orgs_teams.go
```diff
@@ -44,19 +44,46 @@ func (t Team) String() string {
 	return Stringify(t)
 }
 
-// Invitation represents a team member's inviation status
+// Invitation represents a team member's invitation status.
+// Role can be one of the values - direct_member, admin, billing_manager, hiring_manager, or reinstate.
```
varadarajana: @shurcooL yes this is done.

github/orgs_teams.go
varadarajana: @shurcooL this is also done now.

github/orgs_teams.go
```diff
@@ -44,19 +44,46 @@ func (t Team) String() string {
...
+	GistsURL          *string `json:"gists_url,omitempty"`
+	StarredURL        *string `json:"starred_url,omitempty"`
+	SubscriptionsURL  *string `json:"subscriptions_url,omitempty"`
+	OrganizationsURL  *string `json:"organizations_url,omitempty"`
+	ReposURL          *string `json:"repos_url,omitempty"`
+	EventsURL         *string `json:"events_url,omitempty"`
+	ReceivedEventsURL *string `json:"received_events_url,omitempty"`
+	Type              *string `json:"type,omitempty"`
+	SiteAdmin         *bool   `json:"site_admin,omitempty"`
+}
```
shurcooL: I think this new struct is just a subset of [`User`](https://godoc.org/github.com/google/go-github/github#User). I think we can and should use `User` instead of creating a new struct. So you can delete this.

github/orgs_teams.go
```diff
@@ -44,19 +44,46 @@ func (t Team) String() string {
...
-	ID        *int       `json:"id,omitempty"`
-	Login     *string    `json:"login,omitempty"`
-	Email     *string    `json:"email,omitempty"`
+	ID    *int    `json:"id,omitempty"`
+	Login *string `json:"login,omitempty"`
+	Email *string `json:"email,omitempty"`
+	// Role can be one of the values - direct_member, admin, billing_manager, hiring_manager, or reinstate.
 	Role      *string    `json:"role,omitempty"`
 	CreatedAt *time.Time `json:"created_at,omitempty"`
+	Inviter   *Inviter   `json:"inviter,omitempty"`
```
shurcooL: Use `User` instead of `Inviter` here:

```Go
Inviter   *User   `json:"inviter,omitempty"`
```

See here for precedent:

https://github.com/google/go-github/blob/c9c37fd6/github/repos_invitations.go#L15

varadarajana: @shurcooL yes, thank you, i missed it. I have corrected it and checked in.

github/orgs_teams.go
varadarajana: @shurcooL yes, thank you, i missed it. I have corrected it and checked in.

github/orgs_members.go
```diff
@@ -270,3 +270,29 @@ func (s *OrganizationsService) RemoveOrgMembership(user, org string) (*Response,
 
 	return s.client.Do(req, nil)
 }
+
+// ListPendingOrgInvitations returns a list of pending invitations.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/members/#list-pending-organization-invitations
+func (s *OrganizationsService) ListPendingOrgInvitations(org int, opt *ListOptions) ([]*Invitation, *Response, error) {
```
shurcooL: One more thought. Should we spell out the entire word "Organization"? I don't think we shorten "Organization" to "Org" anywhere else. E.g.:

https://godoc.org/github.com/google/go-github/github#ActivityService.ListEventsForOrganization

github/orgs_members.go
```diff
@@ -270,3 +270,29 @@ func (s *OrganizationsService) RemoveOrgMembership(user, org string) (*Response,
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeOrgMembershipPreview)
+
+	pendingInvitations := new([]*Invitation)
```
gmlewis: Just for the record, I don't like calling new to make a pointer to a slice (since it is already a reference type)... but I see this happens all over in this repo, so let's leave it... and maybe some day I can throw together a PR to replace it and see what the other authors think about the change.

github/orgs_teams.go
```diff
@@ -44,13 +44,15 @@ func (t Team) String() string {
...
-// Invitation represents a team member's inviation status
+// Invitation represents a team member's invitation status.
 type Invitation struct {
-	ID        *int       `json:"id,omitempty"`
-	Login     *string    `json:"login,omitempty"`
-	Email     *string    `json:"email,omitempty"`
+	ID    *int    `json:"id,omitempty"`
+	Login *string `json:"login,omitempty"`
+	Email *string `json:"email,omitempty"`
+	// Role can be one of the values - direct_member, admin, billing_manager, hiring_manager, or reinstate.
```
gmlewis: nit - would you mind putting the acceptable values in quotes to follow the style of the other documentation in the repo?

github/orgs_teams_test.go
```diff
@@ -508,7 +509,34 @@ func TestOrganizationsService_ListPendingTeamInvitations(t *testing.T) {
 		testMethod(t, r, "GET")
 		testFormValues(t, r, values{"page": "1"})
 		testHeader(t, r, "Accept", mediaTypeOrgMembershipPreview)
-		fmt.Fprint(w, `[{"id":1}]`)
+		fmt.Fprint(w, `[
```
gmlewis: Can we please reduce this test to just fields of interest?

github/orgs_members_test.go
```diff
@@ -353,3 +354,81 @@ func TestOrganizationsService_RemoveOrgMembership(t *testing.T) {
...
+
+func TestOrganizationsService_ListPendingOrgInvitations(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/orgs/1/invitations", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testFormValues(t, r, values{"page": "1"})
+		testHeader(t, r, "Accept", mediaTypeOrgMembershipPreview)
+		fmt.Fprint(w, `[
```
gmlewis: Can we please reduce this test to just fields of interest?

github/orgs_members.go
shurcooL: Agree that this would be best to tackle in a separate PR since it happens everywhere.

I feel similarly about the whole pattern of returning `err` instead of `nil` in situations where `err` is known to be `nil`; I will make a PR for that sometime.

github/orgs_members.go
shurcooL: Good catch, that's even better!

shurcooL: > I don't think we shorten "Organization" to "Org" anywhere else.

Well, I just found a counterexample of that:

https://github.com/varadarajana/go-github/blob/master/github/orgs_members.go#L178-L181

gmlewis: Yeah... I don't like that as much.  ☹️ 
Your call, @shurcooL as a tie-breaker.  :smiley:

shurcooL: Hmm. Looking at the methods of `OrganizationsService` more closely, it seems there's many, many uses of `Org`. So I'm afraid it might make sense to stay consistent with that and keep this as is.

image

It seems valuable to keep Org because it makes it more clear that the pending invitation applies to org, rather than team (which also exists).

github/orgs_teams_test.go
varadarajana: @gmlewis, @shurcooL  I agree I made the test case big with the full test data. I have used all values that are returned by this API call, i felt test coverage will be better that way. I was not sure which field to omit, as JSON will return all of them. 
Is there a guideline to choose some of these fields from from the response?

github/orgs_members_test.go
varadarajana: @gmlewis, @shurcooL, As I mentioned above, I felt all fields need to be tested for assignment to the struct and compare values. Will it be ok only to test id field as before?

github/orgs_teams_test.go
shurcooL: I personally don't mind having all the fields here, since they come directly from GitHub API examples.

If we reduce it to just IDs, keep the new inviter field with an ID too.

github/orgs_members.go
gmlewis: OK, I can address this change in a separate PR.
For the record, it will become:

```go
	var pendingInvitations []*Invitation
	resp, err := s.client.Do(req, &pendingInvitations)
	if err != nil {
		return nil, resp, err
	}
	return pendingInvitations, resp, nil
```

gmlewis:
LGTM.
Merging.

github/orgs_members_test.go
varadarajana: @shurcooL @gmlewis thank you

gmlewis:
Integrated in https://github.com/google/go-github/commit/0e7db468186d70bb8568a7e7322ba7eba6f7dd5d

github/orgs_members_test.go
gmlewis: Thank you, @varadarajana!

Issue #421, opened 2016-08-23
Add support for changing base branch on pull requests


willnorris:
announcement: https://developer.github.com/changes/2016-08-23-change-base/

This is going to be problematic since `PullRequestsService.Edit` takes a `PullRequest` whose current `base` field is of type `PullRequestBranch`.

parkr:
> This is going to be problematic since PullRequestsService.Edit takes a PullRequest whose current base field is of type PullRequestBranch.

I might be way off base, but is `PullRequestsService.EditBase` a no-go?

willnorris:
> I might be way off base

I see what you did there :smirk:

The other option is to add `state` to [NewPullRequest](https://godoc.org/github.com/google/go-github/github#NewPullRequest), and change `PullRequestsService.Edit` to take that as a parameter.  Which is a breaking change and also really ugly since the name `NewPullRequest` doesn't make sense in that context, and it means if you already have a `PullRequest` object you can't simply pass it to the `Edit` method, you'd have to convert it first.

I gotta say, design decisions like this in the GitHub API drive me absolutely insane.  it would have been a simple matter to have the payload for [updating a pull request](https://developer.github.com/v3/pulls/#update-a-pull-request) be:

```json
{
  "title": "Amazing new feature",
  "body": "Please pull this in!",
  "head": "octocat:new-feature",
  "base": {
    "ref: "master"
  }
}
```

And then it would have been compatible with the existing Pull Request JSON object.  Come to think of it, why doesn't GitHub simply accept both styles? (Maybe it does? I haven't tried it)

shurcooL:
> I gotta say, design decisions like this in the GitHub API drive me absolutely insane.  it would have been a simple matter to have the payload for [updating a pull request](https://developer.github.com/v3/pulls/#update-a-pull-request) be:
> ...
> And then it would have been compatible with the existing Pull Request JSON object.  Come to think of it, why doesn't GitHub simply accept both styles? (Maybe it does? I haven't tried it)

If you think so, it might be worth to send support@github.com an email with that question/suggestion. In my experience, they do a good job of either elaborating on their rationale for an API decision, or taking feedback into account and improving it. At least while the API is in preview period and they're purposefully asking for feedback on it:

> If you have any questions or feedback, please [let us know](https://github.com/contact?form%5Bsubject%5D=Change+base+on+Pull+Requests)!

lucasalcantara:
How this issue going? Is it still have to be done?

[...]

PR #528, opened 2017-01-22
pulls: Support updating base branch

abhinav:
This adds support for changing the base branch of a PR.

As per the discussion in #421, the `PullRequest` is converted into an
unexported type `pullRequestUpdate` which matches the shape expected by the
pulls PATCH endpoint.

Resolves issue #421.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -198,6 +198,9 @@ type pullRequestUpdate struct {
 
 // Edit a pull request.
 //
+// The following fields are editable: Title, Body, State, and Base.Ref.
+// Base.Ref updates the base branch of the pull request.
+//
 // GitHub API docs: https://developer.github.com/v3/pulls/#update-a-pull-request
 func (s *PullRequestsService) Edit(owner string, repo string, number int, pull *PullRequest) (*PullRequest, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d", owner, repo, number)
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -189,12 +189,30 @@ func (s *PullRequestsService) Create(owner string, repo string, pull *NewPullReq
 	return p, resp, err
 }
 
+type pullRequestUpdate struct {
+	Title *string `json:"title,omitempty"`
+	Body  *string `json:"body,omitempty"`
+	State *string `json:"state,omitempty"`
+	Base  *string `json:"base,omitempty"`
+}
+
 // Edit a pull request.
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/#update-a-pull-request
 func (s *PullRequestsService) Edit(owner string, repo string, number int, pull *PullRequest) (*PullRequest, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d", owner, repo, number)
-	req, err := s.client.NewRequest("PATCH", u, pull)
+
+	update := new(pullRequestUpdate)
+	if pull != nil {
+		update.Title = pull.Title
+		update.Body = pull.Body
+		update.State = pull.State
+		if pull.Base != nil {
+			update.Base = pull.Base.Ref
+		}
+	}
+
+	req, err := s.client.NewRequest("PATCH", u, update)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/pulls_test.go
+++ b/github/pulls_test.go
@@ -9,7 +9,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"reflect"
 	"strings"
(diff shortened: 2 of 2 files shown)


github/pulls_test.go
```diff
@@ -232,31 +233,63 @@ func TestPullRequestsService_Create_invalidOwner(t *testing.T) {
 }
 
 func TestPullRequestsService_Edit(t *testing.T) {
-	setup()
-	defer teardown()
+	tests := []struct {
+		input        *PullRequest
+		sendResponse string
+
+		wantUpdate   *pullRequestUpdate
```
shurcooL: This is optional, but I think it'd be better if you made the type of `wantUpdate` a `string` and have it contain the expected JSON.

There's no need to decode the update JSON into our own struct that we used for encoding, and compare those.

What do you think?

github/pulls_test.go
```diff
@@ -232,31 +233,63 @@ func TestPullRequestsService_Create_invalidOwner(t *testing.T) {
...
+		wantResponse *PullRequest
+	}{
+		{
+			input:        &PullRequest{Title: String("t")},
+			sendResponse: `{"number":1}`,
+			wantUpdate:   &pullRequestUpdate{Title: String("t")},
```
shurcooL: Then, the value for `wantUpdate` here would be something like `{"title":"t"}`.

It's easier to reason about it being correct, since GitHub API deals with JSON input and we can compare its documentation with this.

github/pulls.go
```diff
@@ -189,12 +189,30 @@ func (s *PullRequestsService) Create(owner string, repo string, pull *NewPullReq
...
+type pullRequestUpdate struct {
+	Title *string `json:"title,omitempty"`
+	Body  *string `json:"body,omitempty"`
+	State *string `json:"state,omitempty"`
+	Base  *string `json:"base,omitempty"`
+}
+
 // Edit a pull request.
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/#update-a-pull-request
```
abhinav: Agreed. Added docs for this.

github/pulls_test.go
abhinav: I was concerned that the order in which the struct fields would
be serialized is not deterministic but that doesn't appear to be
the case upon trial. Updated.

shurcooL: I see.

Yeah, `encoding/json` encoder uses the order of fields in a struct when encoding, so it is deterministic. It even sorts map keys.

github/pulls_test.go
```diff
@@ -270,12 +271,12 @@ func TestPullRequestsService_Edit(t *testing.T) {
...
-
 				testMethod(t, r, "PATCH")
-				if !reflect.DeepEqual(v, tt.wantUpdate) {
-					t.Errorf("Request body = %+v, want %+v", v, tt.wantUpdate)
+
+				gotUpdate, _ := ioutil.ReadAll(r.Body)
+				wantUpdate := tt.wantUpdate + "\n" // json encoder adds a newline
+				if string(gotUpdate) != wantUpdate {
+					t.Errorf("Request body = %q, want %q", gotUpdate, wantUpdate)
 				}
```
shurcooL: You can use [`testBody`](https://github.com/google/go-github/blob/c9c37fd6ce0247943d1aa34fc3813c62c5cb0225/github/github_test.go#L117-L125) helper to simplify this.

See https://github.com/google/go-github/blob/c9c37fd6ce0247943d1aa34fc3813c62c5cb0225/github/pulls_test.go#L478 for example.

github/pulls_test.go
```diff
@@ -232,31 +234,63 @@ func TestPullRequestsService_Create_invalidOwner(t *testing.T) {
...
+				Base:   &PullRequestBranch{Ref: String("master")},
+			},
+		},
+	}
 
-	input := &PullRequest{Title: String("t")}
+	for _, tt := range tests {
+		func() {
+			setup()
+			defer teardown()
```
shurcooL: In fact, you can use the idea from `TestPullRequestsService_Merge_options` to remove the need for the nested closure with per-testcase `setup()/defer teardown()`. The idea there is to use `i` as the pull request ID, so that registered routes don't overlap.

See https://github.com/google/go-github/blob/c9c37fd6ce0247943d1aa34fc3813c62c5cb0225/github/pulls_test.go#L438-L486.

github/pulls_test.go
```diff
@@ -235,28 +236,58 @@ func TestPullRequestsService_Edit(t *testing.T) {
...
-	input := &PullRequest{Title: String("t")}
+	tests := []struct {
+		input        *PullRequest
+		sendResponse string
 
-	mux.HandleFunc("/repos/o/r/pulls/1", func(w http.ResponseWriter, r *http.Request) {
-		v := new(PullRequest)
-		json.NewDecoder(r.Body).Decode(v)
+		wantUpdate   string
+		wantResponse *PullRequest
```
gmlewis: nit: I find `wantResponse` a bit confusing here... what if we just called it `want` ?
Either way is fine, though... your call.

PR #529, opened 2017-01-22
Add support for context package.

### LINKED ISSUE #526: support the context package
It would be nice if go-github supported the context package, so you could coordinate cancelation across many different goroutines.

It's now available in the standard library as https://golang.org/pkg/context/.

shurcooL:
~~**DO NOT MERGE.** I plan to propose changes before we merge.~~ **Edit:** Changes proposed in https://github.com/google/go-github/issues/526#issuecomment-280985393 and applied to this PR.

Resolves #526.

### DIFF
--- a/examples/basicauth/main.go
+++ b/examples/basicauth/main.go
@@ -10,6 +10,7 @@ package main
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"os"
 	"strings"
@@ -34,14 +35,14 @@ func main() {
 	}
 
 	client := github.NewClient(tp.Client())
-	user, _, err := client.Users.Get("")
+	user, _, err := client.Users.Get(context.Background(), "")
 
 	// Is this a two-factor auth error? If so, prompt for OTP and try again.
 	if _, ok := err.(*github.TwoFactorAuthError); err != nil && ok {
 		fmt.Print("\nGitHub OTP: ")
 		otp, _ := r.ReadString('\n')
 		tp.OTP = strings.TrimSpace(otp)
-		user, _, err = client.Users.Get("")
+		user, _, err = client.Users.Get(context.Background(), "")
 	}
 
 	if err != nil {
--- a/examples/basicauth/main.go
+++ b/examples/basicauth/main.go
@@ -35,14 +35,15 @@ func main() {
 	}
 
 	client := github.NewClient(tp.Client())
-	user, _, err := client.Users.Get(context.Background(), "")
+	ctx := context.Background()
+	user, _, err := client.Users.Get(ctx, "")
 
 	// Is this a two-factor auth error? If so, prompt for OTP and try again.
 	if _, ok := err.(*github.TwoFactorAuthError); err != nil && ok {
 		fmt.Print("\nGitHub OTP: ")
 		otp, _ := r.ReadString('\n')
 		tp.OTP = strings.TrimSpace(otp)
-		user, _, err = client.Users.Get(context.Background(), "")
+		user, _, err = client.Users.Get(ctx, "")
 	}
 
 	if err != nil {
--- a/github/github.go
+++ b/github/github.go
@@ -385,7 +385,12 @@ func parseRate(r *http.Response) Rate {
 // interface, the raw response body will be written to v, without attempting to
 // first decode it. If rate limit is exceeded and reset time is in the future,
 // Do returns *RateLimitError immediately without making a network API call.
+//
+// The provided ctx must be non-nil. If it is canceled or times out,
+// ctx.Err() will be returned.
 func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
+	req = req.WithContext(ctx)
+
 	rateLimitCategory := category(req.URL.Path)
 
 	// If we've hit rate limit, don't make further requests before Reset time.
@@ -393,14 +398,24 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	resp, err := c.client.Do(req.WithContext(ctx))
+	resp, err := c.client.Do(req)
 	if err != nil {
+		// If we got an error, and the context has been canceled,
+		// the context's error is probably more useful.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
+		// If the error type is *url.Error, sanitize its URL before returning.
 		if e, ok := err.(*url.Error); ok {
 			if url, err := url.Parse(e.URL); err == nil {
 				e.URL = sanitizeURL(url).String()
 				return nil, e
 			}
 		}
+
 		return nil, err
 	}
 
--- a/github/github.go
+++ b/github/github.go
@@ -7,6 +7,7 @@ package github
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -385,6 +386,14 @@ func parseRate(r *http.Response) Rate {
 // first decode it. If rate limit is exceeded and reset time is in the future,
 // Do returns *RateLimitError immediately without making a network API call.
 func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
+	return c.do(context.TODO(), req, v)
+}
+
+// TODO: The signature of Do is kept unmodified temporarily, in order for tests
+//       to pass during the transition period. Once all endpoints are updated
+//       to accept ctx parameter, the old Do method can be removed, and this
+//       new do method with ctx parameter should be made to take its place.
+func (c *Client) do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
 	rateLimitCategory := category(req.URL.Path)
 
 	// If we've hit rate limit, don't make further requests before Reset time.
@@ -392,7 +401,7 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		return nil, err
 	}
 
-	resp, err := c.client.Do(req)
+	resp, err := c.client.Do(req.WithContext(ctx))
 	if err != nil {
 		if e, ok := err.(*url.Error); ok {
 			if url, err := url.Parse(e.URL); err == nil {
--- a/github/github.go
+++ b/github/github.go
@@ -385,7 +385,7 @@ func parseRate(r *http.Response) Rate {
 // interface, the raw response body will be written to v, without attempting to
 // first decode it. If rate limit is exceeded and reset time is in the future,
 // Do returns *RateLimitError immediately without making a network API call.
-func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
+func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	rateLimitCategory := category(req.URL.Path)
 
 	// If we've hit rate limit, don't make further requests before Reset time.
@@ -393,7 +393,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	resp, err := c.client.Do(req.WithContext(ctx))
+	resp, err := c.client.Do(req)
 	if err != nil {
 		if e, ok := err.(*url.Error); ok {
 			if url, err := url.Parse(e.URL); err == nil {
@@ -723,7 +723,7 @@ func (c *Client) RateLimits(ctx context.Context) (*RateLimits, *Response, error)
 	response := new(struct {
 		Resources *RateLimits `json:"resources"`
 	})
-	resp, err := c.Do(ctx, req, response)
+	resp, err := c.Do(req.WithContext(ctx), response)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/github.go
+++ b/github/github.go
@@ -385,7 +385,7 @@ func parseRate(r *http.Response) Rate {
 // interface, the raw response body will be written to v, without attempting to
 // first decode it. If rate limit is exceeded and reset time is in the future,
 // Do returns *RateLimitError immediately without making a network API call.
-func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
+func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
 	rateLimitCategory := category(req.URL.Path)
 
 	// If we've hit rate limit, don't make further requests before Reset time.
@@ -393,7 +393,7 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		return nil, err
 	}
 
-	resp, err := c.client.Do(req)
+	resp, err := c.client.Do(req.WithContext(ctx))
 	if err != nil {
 		if e, ok := err.(*url.Error); ok {
 			if url, err := url.Parse(e.URL); err == nil {
@@ -723,7 +723,7 @@ func (c *Client) RateLimits(ctx context.Context) (*RateLimits, *Response, error)
 	response := new(struct {
 		Resources *RateLimits `json:"resources"`
 	})
-	resp, err := c.Do(req.WithContext(ctx), response)
+	resp, err := c.Do(ctx, req, response)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/github.go
+++ b/github/github.go
@@ -714,7 +714,7 @@ func category(path string) rateLimitCategory {
 }
 
 // RateLimits returns the rate limits for the current client.
-func (c *Client) RateLimits() (*RateLimits, *Response, error) {
+func (c *Client) RateLimits(ctx context.Context) (*RateLimits, *Response, error) {
 	req, err := c.NewRequest("GET", "rate_limit", nil)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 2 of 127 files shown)


github/github.go
```diff
@@ -397,14 +398,22 @@ func (c *Client) Rate() Rate {
...
 func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
+	return c.do(context.TODO(), req, v)
+}
+
+// TODO: The signature of Do is kept unmodified temporarily, in order for tests
+//       to pass during the transition period. Once all endpoints are updated
+//       to accept ctx parameter, the old Do method can be removed, and this
+//       new do method with ctx parameter should be made to take its place.
+func (c *Client) do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
 	rateLimitCategory := category(req.URL.Path)
```
kevinburke: would a nil check make sense here?

.travis.yml
```diff
@@ -1,18 +1,11 @@
 sudo: false
 language: go
 go:
-  - 1.5.4
-  - 1.6.3
-  - 1.7
-  - tip
+  - 1.7.x
```
kevinburke: should we add 1.8.x here?

github/github.go
```diff
@@ -397,14 +398,22 @@ func (c *Client) Rate() Rate {
...
 
 	// If we've hit rate limit, don't make further requests before Reset time.
 	if err := c.checkRateLimitBeforeDo(req, rateLimitCategory); err != nil {
 		return nil, err
 	}
 
-	resp, err := c.client.Do(req)
+	resp, err := c.client.Do(req.WithContext(ctx))
```
kevinburke: the WithContext docs state that "the provided ctx must be non-nil." worth erroring about here, or is the panic in net/http okay?

github/github.go
shurcooL: I think it's unnecessary.

The `context` package is already [documented](https://godoc.org/context) to say that `nil` value of `context.Context` shouldn't be passed:

> Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use.

So all valid code will always have non-nil.

Invalid code can be easily caught by static analysis tools such as go vet, as the `context` package docs allude:

> Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation

> TODO is recognized by static analysis tools that determine whether Contexts are propagated correctly in a program.

.travis.yml
shurcooL: We should, once it comes out.

gmlewis:
I like this approach, @shurcooL.
Thank you for doing this!

kevinburke:
Ok this looks good to me

github/issues_test.go
```diff
@@ -149,7 +150,7 @@ func TestIssuesService_Get(t *testing.T) {
 		fmt.Fprint(w, `{"number":1, "labels": [{"url": "u", "name": "n", "color": "c"}]}`)
 	})
 
-	issue, _, err := client.Issues.Get("o", "r", 1)
+	issue, _, err := client.Issues.Get(context.Background(), "o", "r", 1)
```
willnorris: just for brevity, we might want to have a package-level:

```go
var ctx = context.Background()
```

(or is that discouraged somewhere?)

shurcooL: I personally liked `context.Background()` everywhere (even in tests), but it was mostly because I wasn't sure if reusing the same context value in multiple places was okay. It [turns out it is](https://twitter.com/shurcooL/status/799378333590949893) (/cc @rakyll), because of how it's [implemented](https://gotools.org/context#Background).

I still kinda like/don't mind `context.Background()` everywhere because it's readable and I don't think saving a bit of typing is worth it, but I'm also okay with with `var bg = context.Background()` or similar (in tests only).

Previous discussions:

- https://github.com/tambet/go-asana/pull/4#discussion_r80197313
- https://twitter.com/rogpeppe/status/798970167338422272 (/cc @rakyll)

shurcooL:
Okay, now that 1.8 is out, this is unblocked.

I wanted to make some progress here. I decided to prototype the rest of the change. Basically, change all the endpoints to accept `context.Context` and finish this PR. The idea is to see if doing that provides any additional insights. With some assistance from tools, doing it wasn't too hard.

Also, doing a global replace of `context.Background()` in tests with `bg`, if we decide it's worth doing, is trivial. For now, I just used `context.Background()` because it was easier for me.

I'll push the final version and post my observations afterwards.

shurcooL:
Okay, here are my 3 observations from prototyping the entire PR:

1. When I started updating the entire API, I noticed some of the endpoints I had to update were deprecated. For example, `RateLimit` is deprecated in favor of `RateLimits`, `RepositoriesService.ListServiceHooks` is deprecated in favor of `Client.ListServiceHooks`, etc.

    Sure, I could just update them too, but I figured, since their API will change anyway... and this is going to be a huge breaking API change, why not take the opportunity to also delete the deprecated methods and such. If you're going to change your APIs, you might as well stop using deprecated methods; it shouldn't be much extra effort.

   Thoughts on that? Is it a good idea to remove the deprecated stuff as part of this PR, or should I just update their APIs and keep them?

    Additionally, I kinda went ham and removed all deprecated things, even ones that wouldn't be affected by API changes. Please see the entire commit e11b46114b187f07b954ee1c25432b5b494de97b and review it carefully, and let me know your thoughts. Do we delete only the affected deprecated APIs? All of them (as currently is)? Or don't remove any?

2. In https://github.com/google/go-github/issues/526#issuecomment-274365094, I outlined some of the options for new API. One of the open questions was whether or not `Do` method should get `context.Context` as first argument or not. It's absolutely possible to go either way, because `req *http.Request` is the next parameter, and it's always possible to pass the context via the request.

  I currently went with `Do(ctx context.Context, req *http.Request, v interface{})` but after doing the  entire PR, I'm starting to get second thoughts.

   Basically, it feels weird that there are 2 ways to pass the context. It feels especially weird when you don't want to pass it, as in the case of some tests. Leaving signature of `Do` as `Do(req *http.Request, v interface{})` would make it a lot easier not to set a context.

  Consider this code from some of the tests:

	```Go
	req, _ := client.NewRequest("GET", "/", nil)
	resp, err := client.Do(context.Background(), req, nil)
	```

    If we simply passed the context to `Do` via the request, it becomes much more natural not to set the context:

	```Go
	req, _ := client.NewRequest("GET", "/", nil)
	resp, err := client.Do(req, nil)
	```

Or set it explicitly:

	```Go
	req, _ := client.NewRequest("GET", "/", nil)
	resp, err := client.Do(req.WithContext(ctx), nil)
	```

3. This is a huge PR with many files, lines touched. Is there anything I can do to make it easier to review? It's already split into separate logical commits, but I'm talking about the large commit that adds context to all endpoints.

     Any ideas for making it more review friendly? @gmlewis?

I'm going to let this sit as is and think about point 2 myself. If I still think it's a good idea, I might try to change signature of `Do` and see how well it feels.

.travis.yml
shurcooL: Now that 1.8 is out, this is done in 1ebf55f82387495e2e461928e4d8e248c4cd19cd (see line below).

shurcooL:
I've thought about point 2 some more, and decided to make it 2 arguments after all.

See commit 1563515de0c1a97e4c1fcfb74385f50f79e8cb17, the rationale is inside its commit message:

> Pass context to Do via request, instead of as a separate parameter.
> 
> ```Go
> Do(req *http.Request, v interface{})
> ```
> 
> It seems cleaner because it eliminates the question of where to put
> context. With the previous version that took 3 parameters, it was
> possible (in theory) to pass the context via request or via first
> parameter. Now, that ambiguity is gone, there's only one way.
> 
> It should be most helpful in situations where you have a request
> that was made elsewhere, and you want to preserve its original context,
> so you don't have to do something as awkward as:
>
> ```Go
> ghClient.Do(req.Context(), req, ...)
> ```
> 
> Instead, it's simply:
> 
> ```Go
> ghClient.Do(req, ...)
> ```

examples/basicauth/main.go
```diff
@@ -34,14 +35,14 @@ func main() {
...
-	user, _, err := client.Users.Get("")
+	user, _, err := client.Users.Get(context.Background(), "")
 
 	// Is this a two-factor auth error? If so, prompt for OTP and try again.
 	if _, ok := err.(*github.TwoFactorAuthError); err != nil && ok {
 		fmt.Print("\nGitHub OTP: ")
 		otp, _ := r.ReadString('\n')
 		tp.OTP = strings.TrimSpace(otp)
-		user, _, err = client.Users.Get("")
+		user, _, err = client.Users.Get(context.Background(), "")
```
gmlewis: I don't think we want to show calling `context.Background` twice in a row within a single function as an example to be followed. Instead, please assign `ctx := context.Background()` once before or after line 37 and then use `ctx` in both locations in place of `context.Background()`.

github/event_types.go
```diff
@@ -130,19 +130,6 @@ type GollumEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
-// IssueActivityEvent represents the payload delivered by Issue webhook.
```
gmlewis: I read your discussion about deprecation and agree, but now that I look at this, I think we want 3 separate PRs to make for a clean and easy-to-understand history:

* PR#1 - Update repo to Go 1.7 and Go 1.8 and bump version number
* PR#2 - Remove everything that is deprecated
* PR#3 - Support context.Context (only)

If that order is not correct, then put them in the correct order.
My point here is that PR#2 and PR#3 are completely orthogonal, or should be made so as much as possible.

Thoughts?

.travis.yml
kevinburke: thanks!

examples/basicauth/main.go
kevinburke: In examples I think it would be better to use `context.TODO()` tbh to emphasize to the user that they should set as appropriate.

kevinburke:
> Pass context to Do via request, instead of as a separate parameter.

Agree on this

github/event_types.go
shurcooL: Sounds great to me. I'll break it up into 3 PRs, it's just a matter of splitting up those commits, I won't have to do anything extra. :)

However, the tests of 3rd PR will fail until 1st is merged.

shurcooL: > However, the tests of 3rd PR will fail until 1st is merged.

Actually, I just came up with a really sneaky way of getting around that. :stuck_out_tongue_winking_eye:

examples/basicauth/main.go
gmlewis: Hmmm... I'm not sure I agree with that.

Most clients that are not being used within a server will simply want `ctx := context.Background()` and as such is a fine example, especially for newbie Go devs. More experienced Go programmers will realize that they can replace the context with any one that they already have in operation and will have no problem swapping it out for their own.

But having said that, I don't feel really strongly about `TODO` vs `Background`.

I feel more strongly, though, about not duplicating the call and instead assigning to a `ctx` variable, as that is more idiomatic Go.

shurcooL:
**DO NOT MERGE:** Please note the base of this PR is has been changed to `remove-deprecated-apis` branch rather than `master`. This PR is for review purposes only. Once #555 is merged, I will change base of this PR to `master` and it can be mergeable then.

This is part 3 of 3 PRs to resolve #526 (as was suggested to split PR #529 in https://github.com/google/go-github/pull/529#discussion_r101895336).

examples/basicauth/main.go
shurcooL: Yeah, using `context.TODO()` can be seen in two ways:

1. It's a TODO for the reader to replace it with a proper context.
2. It's a TODO that we, the library authors, haven't resolved.

I think you want it to be seen as 1, but it might be seen as 2... So I'd rather use `context.Background()` over `context.TODO()` for that reason.

shurcooL:
> @kevinburke: If there are any sorts of tools that can detect unused function parameters, those might be useful for detecting (in an automated way) a missing `req.WithContext(ctx)`.

Indeed. I believe the relevant issue is https://github.com/golang/go/issues/16742.

That said, for this PR, I can guarantee that context is propagated in all instances. This is why.

I first used `gofmt -r` to rewrite all instances of `Do(req, x) to `Do(ctx, req, x)` in all .go files in `github` directory. That initially caused a lot of build errors, because the `ctx` parameter did not exist. I was guaranteed to get them all, because the signature of `Do` was changed to accept 3 parameters, so all callers must be updated before build succeeds.

Then, I added `ctx` parameter to all methods, which fixed all "undefined: ctx" errors in all calls to `Do`.

After that, I changed signature of `Do` to take 2 parameters, and then I once again used `gofmt -r` to rewrite all instances of `Do(ctx, req, x)` with `Do(req.WithContext(ctx), x)`. The lack of build errors meant that I got all instances.

Of course, future PRs that add new mehods may forget to propagate `ctx` parameter, so the tool will still be useful in the future. Until it exists, we'll need to pay more attention to that during code review.

examples/basicauth/main.go
shurcooL: > @gmlewis: I feel more strongly, though, about not duplicating the call and instead assigning to a ctx variable, as that is more idiomatic Go.

Done in ed4cdcede68dee8b12b2e4c1d7cdf0cb13ac851f, PTAL.

shurcooL:
Thanks for reviewing @gmlewis!

Given how large surface breaking API change this is, I will let this sit until tomorrow morning (Sunday morning), and merge then if nothing comes up. It will be the middle of a weekend, so I think that's a a good time to merge this, since it's less likely to disrupt anyone's work in progress.

On the plus side, updating API usage should be very easy: it's just a matter of sending available context, or `context.Background()` otherwise, as first parameter.

shurcooL:
An update here, the status of this PR continues to be **DO NOT MERGE.** We're going to delay merging because I have some updates. /cc @gmlewis

I've done some more "prototyping" by using this branch locally and updating all of my code that uses go-github for the new API. I did that because I wanted to see if it would lead to any additional insights, and confirm that everything looks good and works as expected (plus, I would have to do it eventually anyway).

As a result of of using the new API (which worked well), I've come to realize there may be a change we should consider making that would lead to an even better outcome.

I'll write up the proposal and my rationale for why I think doing it differently is better in issue #526, so we can discuss it there and decide on the next steps.

shurcooL:
@gmlewis, I have pushed 2 additional commits to this PR, PTAL.

As usual, see the commit messages for rationale. It's also outlined in full detail in https://github.com/google/go-github/issues/526#issuecomment-280985393.

shurcooL:
Everything still looks good, I'm proceeding with the plan to merge this PR now.

ztec:
Hello,

I'm a relatively new go dev but this PR just broke completely the API of this library (as said in the description of https://github.com/google/go-github/commit/23d6cb9cacb5aa314e93d600fe20a48496a718d4) .

I'm not able (too noob in go) to discus choices made with this context added in all Method of the API, and not willing to discus about that.
In many language we have vendor versioning with vendor locking mechanism to avoid breaking change breaking things.

- I did not see any branch v1/v2 in the repository to decide which version (with or without context) I would like to use. 
- I'm not aware of any integrated tool in go ecosystem to fix the exact version/SHA1 `go get` will get. 
 - I'm new to go and willing to know how to solve this "vendor" issue 

What is your solution for people that do not want to fix all already existing codebase to add a mysterious "context" (for me) in all Method call (even if the compiler yell clearly about whats wrong) ?


Regards,
Zed

willnorris:
@ztec: take a look at https://github.com/golang/go/wiki/PackageManagementTools

ztec:
@willnorris So you are saying that [not-so-go](https://github.com/blindpirate/gogradle)/[unofficial](https://github.com/Masterminds/glide)/[unfinished/bogous](https://github.com/golang/dep) tools should be used to do vendor locking ? I find it weird because almost everything in go is "standard" and integrated in the go environment.

I'v already found libraries doing basic versioning for major breaking change like the one of this PR. They use different branches (v1/v2/...) and often state that older versions are not maintained anymore. This allow updates for older versions as well as usage by existing code.
This is fully compatible with standard tools in go.  

What about your (this repo managers) responsibility about defining versions with clear information about breaking changes of the library ? Like Semver specify it for example. Or simply by creating a branch for older version with warning about maintenance ? 


ps: I find tools of your link good ides and solutions.

willnorris:
I'm not trying to tell anyone what they *should* do, I was merely answering your question about how to handle vendoring in Go.  As you probably saw on that page, there is an [official tool](https://github.com/golang/dep) in the works, but it's not ready yet.

The other option is to manually snapshot specific versions of your dependencies and check them in to your version control system, which is basically what many people did before the availability of the above tools.  It's a lot more work, but addresses your concern about relying on unofficial or unfinished tools.

There has been discussion of versioning in general as well as tagging our releases in git (see #376).  At the time, we opted not to do git tags and to increment the libraryVersion const on breaking changes.  Admittedly, that doesn't really help with build breakages, since you can't read the const at that point anyway.

Managing dependency versions has been a problem within the Go community for a long time.  In the time since #376, there has been much more effort to organize a "real" solution, and I'm happy to make whatever changes we need to work in that world.  I'm happy to continue this discussion over on #376, so we can keep it all together.

ztec:
I did not made my "user" job completely as I did not search for discussion about versioning and did not found #376. Sorry. 

Thanks for your great responses.
I will follow https://github.com/google/go-github/issues/376 from now, hoping actions can be done to help users of this library.

PR #532, opened 2017-01-28
Add 'language' property to GistFile struct

sHesl:
Just added the language property to the GistFile struct.


sHesl:
I signed it!

PR #533, opened 2017-01-29
Add StatusesURL to PushEventRepository

bradleyfalzon:
https://developer.github.com/v3/activity/events/types/#pushevent defines the statuses_url sent with each PushEvent webhook.

There wasn't any tests around this event currently (that I could see), but I can add them if you can provide some directions/examples if you'd like.

The position in the struct is also a little strange to me, as all the fields should be populated as this struct represents a webhook event, so I don't understand the comment `The following fields are only populated by Webhook events.`, but I've added the member there as it did seem to be the most appropriate place. Happy to move around.

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -504,6 +504,7 @@ type PushEventRepository struct {
 	// The following fields are only populated by Webhook events.
 	URL          *string       `json:"url,omitempty"`
 	HTMLURL      *string       `json:"html_url,omitempty"`
+	StatusesURL  *string       `json:"statuses_url,omitempty"`
 	Installation *Installation `json:"installation,omitempty"`
 }
 
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -423,26 +423,26 @@ type PullRequestReviewCommentEvent struct {
 //
 // GitHub API docs: http://developer.github.com/v3/activity/events/types/#pushevent
 type PushEvent struct {
-	PushID       *int                 `json:"push_id,omitempty"`
-	Head         *string              `json:"head,omitempty"`
-	Ref          *string              `json:"ref,omitempty"`
-	Size         *int                 `json:"size,omitempty"`
-	Commits      []PushEventCommit    `json:"commits,omitempty"`
-	Repo         *PushEventRepository `json:"repository,omitempty"`
-	Before       *string              `json:"before,omitempty"`
-	DistinctSize *int                 `json:"distinct_size,omitempty"`
+	PushID       *int              `json:"push_id,omitempty"`
+	Head         *string           `json:"head,omitempty"`
+	Ref          *string           `json:"ref,omitempty"`
+	Size         *int              `json:"size,omitempty"`
+	Commits      []PushEventCommit `json:"commits,omitempty"`
+	Before       *string           `json:"before,omitempty"`
+	DistinctSize *int              `json:"distinct_size,omitempty"`
 
 	// The following fields are only populated by Webhook events.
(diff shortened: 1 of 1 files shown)


shurcooL:
> https://developer.github.com/v3/activity/events/types/#pushevent defines the statuses_url sent with each PushEvent webhook.

Can you point out what exactly you're referring to? Is it something on that page, or is it not there?

shurcooL:
> The position in the struct is also a little strange to me, as all the fields should be populated as this struct represents a webhook event, so I don't understand the comment `The following fields are only populated by Webhook events.`,

There are events that you can get via https://developer.github.com/v3/activity/events/ methods. And then there are webhook events that send similar events, but often with some extra fields. Those extra fields are only present in the webhook event, not "normal" event coming from https://developer.github.com/v3/activity/events/ endpoints.

The idea is to put these webhook-only fields under those `// The following fields are only populated by Webhook events.` comments.

It's pretty hard to keep track of accuracy of that, because the github documentation doesn't really specify what fields are a part of normal events and webhook events for all events, so the only way to try to learn is to hit the real API and see what you get. But finding/receving specific events isn't easy.

shurcooL:
I think I see what you meant. The `PushEventRepository` struct is currently only used inside `PushEvent`, for `"repository"` field, and that field is only present in the webhook push event (not normal push event) according to https://developer.github.com/v3/activity/events/types/#events-api-payload-26. So all fields in `PushEventRepository` are webhook-only, as you said.

Do you want to take on a little more in this PR and try to improve that? Here's what I think we need to do:

1. Remove `// The following fields are only populated by Webhook events.` from `PushEventRepository`, or move it to the very top. All fields are for webhook push event only (at this time at least).

2. Move ``Repo *PushEventRepository `json:"repository,omitempty"` `` field in `PushEvent` down, below `// The following fields are only populated by Webhook events.` line. It's a webhook-only field.

3. Follow up on https://github.com/google/go-github/pull/505#commitcomment-20655387 and https://github.com/google/go-github/pull/505#commitcomment-20655389.

The code change is correct to me, so let me know how much of the above you want to take on. It's okay to leave that out for future PRs.

bradleyfalzon:
Thanks @shurcooL, I'm keen to address the 3 points.

bradleyfalzon:
I've made the following changes, as a separate commit, I can combine them once everyone is happy.

1. Removed `// The following fields are only populated by Webhook events.` from `PushEventRepository` as you originally suggested (instead of moving it to the top, else one might think all child type might need to the same doc and we may reuse the type in another struct which is populated by non-webhook events).
1. Moved location of `PushEventRepository` in `PushEvent`  to below `// The following fields are only populated by Webhook events.` - but should I do this for `PushEvent.Commits` type `[]PushEventCommit` too? It's not clear to me where GitHub documents which fields should appear when.
1. Removed `Installation` field from `PushEventCommit` and `PushEventRepository`

shurcooL:
> but should I do this for `PushEvent.Commits` type `[]PushEventCommit` too? It's not clear to me where GitHub documents which fields should appear when.

I'm using https://developer.github.com/v3/activity/events/types/#events-api-payload-26 as a reference. It states that `commits` array and certain fields are present as part of the normal (not webhook) event. So I think it's fine.

gmlewis:
Thank you, @bradleyfalzon and @shurcooL!
LGTM.
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/4baf08e3bbebb0f2105099a14e850cdeb31b4cf2

PR #534, opened 2017-01-30
Adds support for GET /repos/:owner/:repo/collaborators/:username/permission

alindeman:
GitHub API docs: https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level

### DIFF
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -75,7 +75,7 @@ func (s *RepositoriesService) GetPermissionLevel(owner, repo, user string) (*Rep
 	if err != nil {
 		return nil, resp, err
 	}
-	return rpl, resp, err
+	return rpl, resp, nil
 }
 
 // RepositoryAddCollaboratorOptions specifies the optional parameters to the
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -49,6 +49,35 @@ func (s *RepositoriesService) IsCollaborator(owner, repo, user string) (bool, *R
 	return isCollab, resp, err
 }
 
+// RepositoryPermissionLevel represents the permission level an organization
+// member has for a given repository.
+type RepositoryPermissionLevel struct {
+	// Possible values: "admin", "write", "read", "none"
+	Permission *string `json:"permission,omitempty"`
+
+	User *User `json:"user,omitempty"`
+}
+
+// GetPermissionLevel retrieves the specific permission level a collaborator has for a given repository.
+// GitHub API docs: https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level
+func (s *RepositoriesService) GetPermissionLevel(owner, repo, user string) (*RepositoryPermissionLevel, *Response, error) {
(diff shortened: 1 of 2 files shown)


github/repos_collaborators.go
```diff
@@ -49,6 +49,35 @@ func (s *RepositoriesService) IsCollaborator(owner, repo, user string) (bool, *R
 	return isCollab, resp, err
 }
 
+// RepositoryPermissionLevel represents the permission level an organization
+// member has for a given repository.
+type RepositoryPermissionLevel struct {
+	// Possible values: "admin", "write", "read", "none"
+	Permission string `json:"permission"`
```
shurcooL: What's the reason you decided to use a `string` here instead of `*string`, as is commonly done in other structs?

github/repos_collaborators.go
```diff
@@ -49,6 +49,35 @@ func (s *RepositoriesService) IsCollaborator(owner, repo, user string) (bool, *R
...
+
+	User *User `json:"user"`
```
shurcooL: Was there a reason to avoid `,omitempty` option for both `User` and `Permission`, unlike all other structs we have?

alindeman:
> This is looking good. A few questions.

@shurcooL Thinking back to the [conversation we had over in #512](https://github.com/google/go-github/pull/512#issuecomment-272350152), I thought that since--as far as I can tell--these attributes will always be returned from the API, we could specify them as non-pointer types.

Of course, that's assuming new API routes will not be added in the future that only return subsets of attributes. If you think it'd be better if they matched up with the typical convention of pointers and `omitempty`, I'm not opposed.

shurcooL:
Thanks for answering. That makes sense. I wanted to discuss this a bit, and share some thoughts/findings.

My comment ended up being lengthy, so I decided to take it out from this PR and make it a separate issue, so that others can weigh in their opinions thee, without derailing this PR.

Please see https://github.com/google/go-github/issues/537#issue-204414025.

alindeman:
@shurcooL I've amended my commit based on the conclusion in #537

github/repos_collaborators.go
```diff
@@ -49,6 +49,35 @@ func (s *RepositoriesService) IsCollaborator(owner, repo, user string) (bool, *R
...
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeOrgMembershipPreview)
+
+	rpl := new(RepositoryPermissionLevel)
+	resp, err := s.client.Do(req, rpl)
+	if err != nil {
+		return nil, resp, err
+	}
+	return rpl, resp, err
```
gmlewis: We are about to explicitly return `nil` when there is no error, so maybe you could start this practice:
`return rpl, resp, nil`
See #536.

alindeman:
@gmlewis Sounds good. I thought the convention was a bit odd, but I admit to blindly following it for consistently. Fixed in the latest push.

gmlewis:
Thank you, @alindeman!
LGTM.
Merging.

PR #535, opened 2017-01-31
Change new([]*...) to slice zero value

gmlewis:
This change is a proposal and is open for discussion.

Change-Id: I62da2980c76d6220ce2ef12e6a1e64b60769ab01

### DIFF
--- a/github/integration.go
+++ b/github/integration.go
@@ -28,11 +28,11 @@ func (s *IntegrationsService) ListInstallations(opt *ListOptions) ([]*Installati
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeIntegrationPreview)
 
-	i := new([]*Installation)
+	var i []*Installation
 	resp, err := s.client.Do(req, &i)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return *i, resp, err
+	return i, resp, nil
 }
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -76,12 +76,12 @@ func (s *RepositoriesService) ListReleases(owner, repo string, opt *ListOptions)
 		return nil, nil, err
 	}
 
-	releases := new([]*RepositoryRelease)
-	resp, err := s.client.Do(req, releases)
+	var releases []*RepositoryRelease
+	resp, err := s.client.Do(req, &releases)
 	if err != nil {
 		return nil, resp, err
 	}
-	return *releases, resp, err
+	return releases, resp, nil
 }
 
 // GetRelease fetches a single release.
@@ -188,12 +188,12 @@ func (s *RepositoriesService) ListReleaseAssets(owner, repo string, id int, opt 
 		return nil, nil, err
 	}
 
-	assets := new([]*ReleaseAsset)
-	resp, err := s.client.Do(req, assets)
+	var assets []*ReleaseAsset
+	resp, err := s.client.Do(req, &assets)
 	if err != nil {
-		return nil, resp, nil
+		return nil, resp, err
 	}
-	return *assets, resp, err
+	return assets, resp, nil
 }
 
 // GetReleaseAsset fetches a single release asset.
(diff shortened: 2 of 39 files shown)


github/repos_releases.go
```diff
@@ -188,12 +188,12 @@ func (s *RepositoriesService) ListReleaseAssets(owner, repo string, id int, opt
 		return nil, nil, err
 	}
 
-	assets := new([]*ReleaseAsset)
-	resp, err := s.client.Do(req, assets)
+	var assets []*ReleaseAsset
+	resp, err := s.client.Do(req, &assets)
 	if err != nil {
 		return nil, resp, nil
```
shurcooL: I spotted this. It also happens in one more place in this file. See https://github.com/google/go-github/pull/63#pullrequestreview-19332985.

Do you want to include a fix for that in this PR @gmlewis?

github/integration.go
```diff
@@ -28,11 +28,11 @@ func (s *IntegrationsService) ListInstallations(opt *ListOptions) ([]*Installati
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeIntegrationPreview)
 
-	i := new([]*Installation)
+	var i []*Installation
 	resp, err := s.client.Do(req, &i)
```
shurcooL: I wonder if this even worked before, given the double pointer given to `client.Do`, i.e., `**[]*Installation`.

A quick test reveals that it should be fine, see https://play.golang.org/p/O5lq1RfseF. But it's still nice to fix to have simpler code.

github/repos_releases.go
```diff
@@ -188,12 +188,12 @@ func (s *RepositoriesService) ListReleaseAssets(owner, repo string, id int, opt
 		return nil, nil, err
 	}
 
-	assets := new([]*ReleaseAsset)
-	resp, err := s.client.Do(req, assets)
+	var assets []*ReleaseAsset
+	resp, err := s.client.Do(req, &assets)
 	if err != nil {
-		return nil, resp, nil
+		return nil, resp, err
```
gmlewis: Fixed err return here. See https://github.com/google/go-github/pull/63#pullrequestreview-19332985

github/repos_releases.go
```diff
@@ -210,7 +210,7 @@ func (s *RepositoriesService) GetReleaseAsset(owner, repo string, id int) (*Rele
 	asset := new(ReleaseAsset)
 	resp, err := s.client.Do(req, asset)
 	if err != nil {
-		return nil, resp, nil
+		return nil, resp, err
```
gmlewis: Fixed err return here. See https://github.com/google/go-github/pull/63#pullrequestreview-19332985

github/repos_releases.go
```diff
@@ -248,7 +248,7 @@ func (s *RepositoriesService) DownloadReleaseAsset(owner, repo string, id int) (
 		if !strings.Contains(err.Error(), "disable redirect") {
 			return nil, "", err
 		}
-		return nil, loc, nil
+		return nil, loc, nil // Intentionally return no error with valid redirect URL.
```
gmlewis: Added comment here. See https://github.com/google/go-github/pull/312#commitcomment-20680837

gmlewis:
Thank you, @shurcooL! I believe I addressed your comments.

gmlewis:
Other readers - please note that the general handling of `return ..., nil` instead of `return ..., err` when `err == nil` will be handled in a separate PR.  I'll file an issue.

gmlewis:
Yes, by "general handling", I mean all other instances.
Sorry for the auto-squash... the tools that I use perform this on my behalf. I'll try to remember to stop using them.  They were a requirement for working with the Go repos I worked on.

gmlewis:
Merging.

shurcooL:
> Yes, by "general handling", I mean all other instances.

Gotcha. 👍

> Sorry for the auto-squash... the tools that I use perform this on my behalf. I'll try to remember to stop using them. They were a requirement for working with the Go repos I worked on.

No problem, and thanks. I hope I'm not inconveniencing you too much. If this were Gerrit, it would keep track of your individual pushes so I could always see a diff between patch sets, but GitHub doesn't do that.

PR #538, opened 2017-02-01
Added listing outside collaborators for an organization

glapul:
2fa_disabled filter is supported too.

I created a new file [orgs_outside_collaborators.go], but I'm not sure whether it wouldn't be better to rename [orgs_members.go] to something like [orgs_members_and_outside_collaborators.go] as it's not clear to me where would the function that converts a member to an outside collaborator fit best.

[orgs_outside_collaborators_test.go] contains tests that are basically recycled from [orgs_members_test.go]. 

I thought about adding an integration test, but I think that setting up a test scenario is either very hard or straight impossible right now (as we would have to be able to create an organisation and users, which I think is impossible without an Github Enterprise account). 
So I fell back to testing it a bit by hand, and those tests passed.

By the way, it's not clear to me why only the organisation's owner can list the organisation's outside collaborators (that's the response API gave me when I tried listing google's). It seems that this is inferrable by just going through all the organisation's repositories, so being a member should suffice. Or am I missing something?

### DIFF
--- a/github/orgs_outside_collaborators.go
+++ b/github/orgs_outside_collaborators.go
@@ -1,4 +1,4 @@
-// Copyright 2016 The go-github AUTHORS. All rights reserved.
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
@@ -24,7 +24,7 @@ type ListOutsideCollaboratorsOptions struct {
 // Warning: The API may change without advance notice during the preview period.
 // Preview features are not supported for production use.
 //
-// GitHub API docs: https://developer.github.com/v3/orgs/outside_collaborators/
+// GitHub API docs: https://developer.github.com/v3/orgs/outside_collaborators/#list-outside-collaborators
 func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOutsideCollaboratorsOptions) ([]*User, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/outside_collaborators", org)
 	u, err := addOptions(u, opt)
@@ -33,7 +33,6 @@ func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOut
 	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
-
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/orgs_outside_collaborators.go
+++ b/github/orgs_outside_collaborators.go
@@ -46,5 +46,5 @@ func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOut
 		return nil, resp, err
 	}
 
-	return members, resp, err
+	return members, resp, nil
 }
--- a/github/orgs_outside_collaborators.go
+++ b/github/orgs_outside_collaborators.go
@@ -1,0 +1,51 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import "fmt"
+
+// ListOutsideCollaboratorsOptions specifies optional parameters to the
+// OrganizationsService.ListOutsideCollaborators method.
+type ListOutsideCollaboratorsOptions struct {
+	// Filter outside collaborators returned in the list. Possible values are:
+	// 2fa_disabled, all.  Default is "all".
+	Filter string `url:"filter,omitempty"`
+
+	ListOptions
+}
+
+// ListOutsideCollaborators lists outside collaborators of organization's repositories.
+// This will only work if the authenticated
+// user is an owner of the organization.
+//
+// Warning: The API may change without advance notice during the preview period.
+// Preview features are not supported for production use.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/outside_collaborators/
+func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOutsideCollaboratorsOptions) ([]*User, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/outside_collaborators", org)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeOrgMembershipPreview)
+
+	members := new([]*User)
+	resp, err := s.client.Do(req, members)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return *members, resp, err
+}
--- a/github/orgs_outside_collaborators.go
+++ b/github/orgs_outside_collaborators.go
@@ -40,11 +40,11 @@ func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOut
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeOrgMembershipPreview)
 
-	members := new([]*User)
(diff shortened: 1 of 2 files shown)


gmlewis:
Why not extend the already-existing `Organizations.ListMembers` method (in `orgs_members.go`) and make `OutsideCollaborators bool` a new option of `ListMembersOptions` ?

It currently already handles the `PublicOnly` option and seems like a good fit to me.

As an added bonus, it already has all the `Filter`  and `ListOptions` ready-to-go.

shurcooL:
@gmlewis, that's a good idea to consider.

However, I see that GitHub API really lists this as a separate endpoint. Compare:

- Members list - https://developer.github.com/v3/orgs/members/#members-list
- List outside collaborators - https://developer.github.com/v3/orgs/outside_collaborators/#list-outside-collaborators

Additionally, removing an outside collaborator is also different than removing a normal member:

- Remove a member - https://developer.github.com/v3/orgs/members/#remove-a-member
- Remove outside collaborator - https://developer.github.com/v3/orgs/outside_collaborators/#remove-outside-collaborator

You can also "convert member to outside collaborator". It suggests to me these things are different enough that listing them via different endpoints makes sense.

While the idea of merging the listing endpoints makes sense now, I worry that future changes/additions to the API that further diverge outside collaborators and members may cause trouble for our API.

What do you think?

gmlewis:
OK, I'm fine with keeping it separate then, @shurcooL.

github/orgs_outside_collaborators.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: nit: s/2016/2017/

github/orgs_outside_collaborators.go
```diff
@@ -0,0 +1,51 @@
...
+// GitHub API docs: https://developer.github.com/v3/orgs/outside_collaborators/
+func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOutsideCollaboratorsOptions) ([]*User, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/outside_collaborators", org)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+
```
gmlewis: nix blank line between API call and err check

github/orgs_outside_collaborators.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeOrgMembershipPreview)
+
+	members := new([]*User)
```
gmlewis:
```go
	var members []*User
	resp, err := s.client.Do(req, &members)
	if err != nil {
		return nil, resp, err
	}

	return members, resp, nil
```

github/orgs_outside_collaborators_test.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: nit: s/2016/2017/

github/orgs_outside_collaborators_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestOrganizationsService_ListOutsideCollaborators(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/orgs/o/outside_collaborators", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
```
gmlewis: please add test for preview header... see other examples in repo

github/orgs_outside_collaborators.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// ListOutsideCollaborators lists outside collaborators of organization's repositories.
+// This will only work if the authenticated
+// user is an owner of the organization.
+//
+// Warning: The API may change without advance notice during the preview period.
+// Preview features are not supported for production use.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/outside_collaborators/
```
gmlewis: Change URL to:
https://developer.github.com/v3/orgs/outside_collaborators/#list-outside-collaborators

glapul:
Done.

github/orgs_outside_collaborators.go
gmlewis: I don't see these changes... do you need to push them?

github/orgs_outside_collaborators.go
```diff
@@ -40,11 +40,11 @@ func (s *OrganizationsService) ListOutsideCollaborators(org string, opt *ListOut
...
-	members := new([]*User)
-	resp, err := s.client.Do(req, members)
+	var members []*User
+	resp, err := s.client.Do(req, &members)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return *members, resp, err
+	return members, resp, err
```
gmlewis: return members, resp, nil

gmlewis:
Awaiting second LGTM.

github/orgs_outside_collaborators.go
```diff
@@ -0,0 +1,50 @@
...
+
+	ListOptions
+}
+
+// ListOutsideCollaborators lists outside collaborators of organization's repositories.
+// This will only work if the authenticated
+// user is an owner of the organization.
+//
+// Warning: The API may change without advance notice during the preview period.
+// Preview features are not supported for production use.
```
shurcooL: We don't usually include this warning in all other similar situations. Should we start doing it here?

It'll create more work and might get out of date. We have the `// TODO: remove custom Accept header when this API fully launches.` comments in code, and GitHub API itself is a good source for knowing which endpoints are stable or not.

I'm not against it, just a thought, because this is the first comment of its kind.

github/orgs_outside_collaborators.go
```diff
@@ -0,0 +1,50 @@
...
+	// Filter outside collaborators returned in the list. Possible values are:
+	// 2fa_disabled, all.  Default is "all".
+	Filter string `url:"filter,omitempty"`
+
+	ListOptions
+}
+
+// ListOutsideCollaborators lists outside collaborators of organization's repositories.
+// This will only work if the authenticated
+// user is an owner of the organization.
```
shurcooL: Nitpick, the sentence "This will only work if the authenticated user is an owner of the organization." can probably fit on a single line, instead of being split into two lines of lengths 40 and 37 characaters.

Also, question, what's the source of this statement? I don't see it mentioned in https://developer.github.com/v3/orgs/outside_collaborators/#list-outside-collaborators.

glapul: While trying to use the API by hand I was informed (with an error) that only the owner can do this.

github/orgs_outside_collaborators.go
glapul: It's not the first one, I copied it from another file using the same Accept header. (and thought that's the convention). I'll find the file when I'll get back home.

shurcooL: Oh okay, keeping it is perfectly fine then. Thanks!

gmlewis:
It looks like this was ready to merge and I think I dropped the ball by not getting it in before the context changes.

I'm thinking that while I'm rebasing, that I could add a commit before squashing that adds context.

Does that sound reasonable to you, @shurcooL, or do you have an alternative proposal?

shurcooL:
Yes, this PR is just 1 logical commit + 3 adjustments based on review comments. I think you should rebase them, add context support and squash everything into one commit before merging into master.

> Add more commits by pushing to the `outside_collaborators` branch on `glapul/go-github`.

You can also do that.

gmlewis:
SGTM. Adjusting and merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/93f95bf99fdd49bad208a6a555090ce882d2653e

PR #539, opened 2017-02-01
Document Git.CreateCommit and PullRequests.Edit arguments

### LINKED ISSUE #531: What happens if you Git.CreateCommit(..., nil) or PullRequests.Edit(..., nil)?
`GitService.CreateCommit` is [documented](https://godoc.org/github.com/google/go-github/github#GitService.CreateCommit) as:

> CreateCommit creates a new commit in a repository.
>
> The commit.Committer is optional and will be filled with the commit.Author data if omitted. If the commit.Author is omitted, it will be filled in with the authenticated user’s information and the current date.
>
> GitHub API docs: http://developer.github.com/v3/git/commits/#create-a-commit

If it's legal, what happens if a user calls `CreateCommit("owner", "repo", nil)`? Should we document it, because I don't think it's intuitive or clear.

If it's not legal, that should be documented.

---

Same question for `PullRequestsService.Edit`, [documented](https://godoc.org/github.com/google/go-github/github#PullRequestsService.Edit) as:

> Edit a pull request.
>
> GitHub API docs: https://developer.github.com/v3/pulls/#update-a-pull-request

---

In case nil values are not legal, why are pointers being accepted? Is it for performance reasons, or something else? If a breaking API change is viable, would it be a good idea to accept a value instead?

---

These questions came up while reviewing #528 which resolves #421 (but they're outside of scope of that PR, may be in scope of the issue).

gmlewis:
This also changes the logic of `Git.CreateCommit` to ensure that the `commit` argument is non-nil.
I am against changing this argument to a value because it is a non-trivial struct and makes sense (to me) to be passed by reference.

Fixes #531.

Change-Id: Ifc6640db138e4ea936f3c3077c1563fe3eedd450

### DIFF
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -88,6 +88,7 @@ type createCommit struct {
 }
 
 // CreateCommit creates a new commit in a repository.
+// commit must not be nil.
 //
 // The commit.Committer is optional and will be filled with the commit.Author
 // data if omitted. If the commit.Author is omitted, it will be filled in with
@@ -95,22 +96,25 @@ type createCommit struct {
 //
 // GitHub API docs: https://developer.github.com/v3/git/commits/#create-a-commit
 func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string, commit *Commit) (*Commit, *Response, error) {
+	if commit == nil {
+		return nil, nil, fmt.Errorf("commit must be provided")
+	}
+
 	u := fmt.Sprintf("repos/%v/%v/git/commits", owner, repo)
 
-	body := &createCommit{}
-	if commit != nil {
-		parents := make([]string, len(commit.Parents))
-		for i, parent := range commit.Parents {
-			parents[i] = *parent.SHA
-		}
+	parents := make([]string, len(commit.Parents))
+	for i, parent := range commit.Parents {
+		parents[i] = *parent.SHA
+	}
 
-		body = &createCommit{
-			Author:    commit.Author,
-			Committer: commit.Committer,
-			Message:   commit.Message,
-			Tree:      commit.Tree.SHA,
-			Parents:   parents,
-		}
+	body := &createCommit{
+		Author:    commit.Author,
+		Committer: commit.Committer,
+		Message:   commit.Message,
+		Parents:   parents,
+	}
+	if commit.Tree != nil {
+		body.Tree = commit.Tree.SHA
 	}
 
 	req, err := s.client.NewRequest("POST", u, body)
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -198,6 +198,7 @@ type pullRequestUpdate struct {
 }
 
 // Edit a pull request.
+// pull may be nil, in which case the requested PR is returned unmodified.
 //
 // The following fields are editable: Title, Body, State, and Base.Ref.
 // Base.Ref updates the base branch of the pull request.
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -198,23 +198,26 @@ type pullRequestUpdate struct {
 }
 
 // Edit a pull request.
-// pull may be nil, in which case the requested PR is returned unmodified.
+// pull must not be nil.
 //
 // The following fields are editable: Title, Body, State, and Base.Ref.
 // Base.Ref updates the base branch of the pull request.
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/#update-a-pull-request
 func (s *PullRequestsService) Edit(ctx context.Context, owner string, repo string, number int, pull *PullRequest) (*PullRequest, *Response, error) {
+	if pull == nil {
+		return nil, nil, fmt.Errorf("pull must be provided")
+	}
+
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d", owner, repo, number)
 
-	update := new(pullRequestUpdate)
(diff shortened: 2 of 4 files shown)


github/pulls.go
```diff
@@ -197,6 +197,7 @@ type pullRequestUpdate struct {
 }
 
 // Edit a pull request.
+// pull may be nil, in which case the requested PR is returned unmodified.
```
shurcooL: I'm trying to think about the best thing to do in a situation where `pull` is `nil`.

Have you also considered returning an error, similar to when `commit` is `nil` in `CreateCommit`?

If so, how did you pick which one to go with?

I can see that for the `CreateCommit` case, there's really no other viable choice but to return error if `commit` is `nil`. You can't create a commit if it weren't supplied, and the only reason `Commit` is being given as a pointer is to copy less memory, not because `nil` value has any meaning.

For `Edit`ing a pull request, it's a little less clear and both options are viable.

My main concern with accepting `pull` value of `nil` as a valid operation is that it would encourage doing that in some situations, but it seems silly to make a network request to GitHub API servers with an noop operation, doesn't it? So maybe we should be more aggressive and return error here too, if `pull` is `nil`?

Or do you think it makes sense to allow the pull request parameter value of `nil`? In that case, should we consider not making the network request, or is that "too smart" for what's meant to be more or less a direct wrapper to the GH API?

gmlewis: Sorry I let this one slip by... Yes, I think you are right.
There really is no need to call `Edit` with `pull == nil` since `Get` could be called instead.
I'll make this an error too.

github/git_commits.go
```diff
@@ -88,29 +88,33 @@ type createCommit struct {
...
+
+	body := &createCommit{
+		Author:    commit.Author,
+		Committer: commit.Committer,
+		Message:   commit.Message,
+		Parents:   parents,
+	}
+	if commit.Tree != nil {
+		body.Tree = commit.Tree.SHA
 	}
```
shurcooL: One question about this new `if commit.Tree != nil {` statement. Do you know why it wasn't here before, and why did you decide to add it now?

I don't see this change being mentioned in the commit message, and I think it's worth including.

github/git_commits.go
```diff
@@ -88,29 +88,33 @@ type createCommit struct {
...
+// commit must not be nil.
 //
 // The commit.Committer is optional and will be filled with the commit.Author
 // data if omitted. If the commit.Author is omitted, it will be filled in with
 // the authenticated user’s information and the current date.
 //
 // GitHub API docs: https://developer.github.com/v3/git/commits/#create-a-commit
 func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string, commit *Commit) (*Commit, *Response, error) {
+	if commit == nil {
+		return nil, nil, fmt.Errorf("commit must be provided")
```
shurcooL: Given that this func (and `PullRequestsService.Edit`) is now documented that "commit must not be nil", and it returns an error, we have two choices of what to do if the user uses the API incorrectly and `commit` is actually `nil`:

1. Panic.
2. Return an error (currently being done).

Which of those two options do you think is better @gmlewis any why?

gmlewis: [Effective Go](https://golang.org/doc/effective_go.html#panic) recommends:

> ... real library functions should avoid panic.

and I strongly agree. I would much rather that my server/program spit out an error message and continue on (or choose to die, depending on its particular circumstances) than to have some library choose for me to kill my program.

github/git_commits.go
gmlewis: It looks to me like I saw a potential runtime panic and chose to plug it.

I'll add a comment to the commit message.

gmlewis:
I believe I have addressed all feedback, @shurcooL.
PTAL.

github/git_commits.go
shurcooL: I completely agree that library functions should avoid panic during normal processing.

But that same Effective Go section also says "One possible counterexample is during initialization", and I think another counterexample could be for invalid API usage. For example, consider [`(crypto/rand).Int`](https://godoc.org/crypto/rand#Int) function:

```Go
// Int returns a uniform random value in [0, max). It panics if max <= 0.
func Int(rand io.Reader, max *big.Int) (n *big.Int, err error) {
```

Panicking (rather than returning error) on incorrect API usage seems totally appropriate, since the goal is to let the user know that they are using the API incorrectly as soon as possible, so the most drastic option (such as panic) is desired.

Most other places in the standard library that panic on invalid API usage don't have an error return value (e.g., [`(math/big).Int.Sqrt`](https://godoc.org/math/big#Int.Sqrt), so returning a `errors.New("x must be non-negative")` is not even an option there.

However, that is not very applicable in this situation.

The biggest reason why I think we have to go with returning error here in go-github (rather than panicking) is the previous behavior of this method. Previous behavior was to accept `nil` value for `*Commit` and not panic (but return an error that comes from GitHub response), so it would be completely unacceptable for us to suddenly change that.

If we were designing a new API from scratch, then panicking on incorrect API usage, or choosing to use a value to make it impossible to misuse the API, is something we could've considered. But we're not.

So 👍 to returning error. :)

shurcooL:
Friendly ping. I'll let you merge it @gmlewis since it's your PR.

gmlewis:
Thanks for the ping, @shurcooL!
Merging.

PR #541, opened 2017-02-02
Additional Clone related URLs to PushEventRepository

bradleyfalzon:
As defined in https://developer.github.com/v3/activity/events/types/#pushevent

Sorry, these should've been merged in with #533 at the time, and have the same conditions applied, that they're only defined in the webhook payload which contains `PushEventRepository` which is commented to only exist during a webhook. So I think it should be a simple addition.

Let me know if you'd like changes.


shurcooL:
Confirmed with https://developer.github.com/v3/activity/events/types/#webhook-payload-example-22.

LGTM.

Thanks!

PR #543, opened 2017-02-03
Generate accessors for structs with pointer fields

### LINKED ISSUE #45: add Getter methods for data structures
as part of resolving #19, all data structures now use pointer field values.  To ease the burden on developers to constantly perform nil checks, we should generate `Get*()` funcs on all our data structures, similar to what goprotobuf does.  If a field is non-nil, return it.  Otherwise, return the zero value for that type.  So for example,

``` go
func (m *User) GetLogin() string {
    if m != nil && m.Login != nil {
        return *m.Login
    }
    return ""
}
```

Go provides good capabilities to read and manipulate go source code, so this can be completely generated.  We might need to move all these types into a single `data.go` file, I'm not sure.

gmlewis:
Fixes #45.

Change-Id: Ib2b6cc5d713e2eb833ee3c7fcfbd804bfe8fa313


shurcooL:
Thanks for working on this @gmlewis. Before I look at the details of the implementation, I'd like to discuss this change at a higher level. I have some concerns about the general direction and I'd like to see where others stand. Let's maybe do that in #45 itself, and leave this PR to discuss implementation details. I'll ping you there.

shurcooL:
As I posted in https://github.com/google/go-github/issues/45#issuecomment-279566195, my potential concerns are resolved and I have no objections to doing this. I'll leave some review comments on the implementation now.

shurcooL:
First, some high level questions/thoughts for @gmlewis, looking only at the final API after this PR.

1. The "Get" prefix is slightly unfortunate, given Go style suggests to avoid "Get" prefix from getters (source: https://golang.org/doc/effective_go.html#Getters).

    However, I understand this is unavoidable because otherwise field name and method name would have same name, right? Also, this is the same solution that protobuf uses and is therefore consistent, right?

    Just mentioning this, I don't think we need to change it. Also, one way to justify the "Get" prefix is that it's not a simple getter, it has logic for returning zero value if the pointer is nil.

2. The documentation text for methods seems to prioritize mentioning the nil case:

    ```
    // GetMessage returns the zero value if nil, otherwise the actual value.
    ```

    Is that the way to go? I'd expect the nil case to be the less common path, and the actual value being the primary path. So would it better to reflect that in the docs by swapping the order? _Something_ that mentions the actual value first, nil case second, like:

    ```
    // GetMessage returns the Message field if its value is non-nil, and zero value otherwise.
    ```

    (That phrasing doesn't sound better, but perhaps it can be tweaked. Just want to get your thoughts.)

3. This is actually somewhat implementation-related. Currently, this PR generates 56 "-accessors.go" files. Given they're all generated and most users/developers won't need to look at/modify the detailed contents of those files, maybe it'd be better to consolidate them all in a single  generated .go file?

    That way, when working on this package, there are less files get in your way that you don't want to modify. Also smaller risk of accidentally modifying a generated file, because there are fewer of them.

    Compare the file list [before](https://cloud.githubusercontent.com/assets/1924134/22909491/6174f118-f222-11e6-85cf-b405a3b3eedb.png) and [after this PR](https://cloud.githubusercontent.com/assets/1924134/22909514/82c2e1a4-f222-11e6-817e-6789566be59f.png).

    Please don't change this yet, I just want to get your thoughts on this first. We should consider the trade-offs of both approaches.

gmlewis:
@shurcooL - addressing your questions/thoughts:

1. Yes, "Get" is consistent with the proto2 implementation.
2. I went through a few iterations on the comment, and am fine with changing it. Both versions seem clear to me, but am happy to emphasize as suggested.
3. Sure, I'm happy to consolidate to a single file... I was originally thinking it would be nice to easily jump to just the accessors for a particular file, but with all the editor tooling we have these days, maybe a single file is better after all.

willnorris:
+1 to single file

github/github.go
```diff
@@ -3,6 +3,16 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// go-github authors should run the following commands before
+// sending PRs to GitHub:
+//
+//     cd github.com/go-github
```
shurcooL: `cd github.com/go-github`? That doesn't sound right.

github/github.go
```diff
@@ -3,6 +3,16 @@
...
+//     go generate ./...
+//     go test ./...
+//     go vet ./...
```
shurcooL: How about deleting the `cd` step and making these commands use absolute import paths rather than relative? E.g.:

```
go generate github.com/google/go-github/...
```

Then the same command can be run from any directory, and it'll do the right thing.

github/github-accessors.go
```diff
@@ -0,0 +1,7229 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// github-accessors.go generated by gen-accessors; DO NOT EDIT
```
shurcooL: The [standardized generated file disclaimer header format](https://github.com/golang/go/issues/13560#issuecomment-277804473) not official yet, but it seems to be on its way to become so. Let's start following that standard.

```
// Code generated by gen-accessors; DO NOT EDIT.
```

cmd/tools/gen-accessors.go
```diff
@@ -0,0 +1,300 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
```
shurcooL: I think you should add:

```
// +build ignore
```

It won't affect go generate functionality since the .go file is being `go run` directly.

But it'll prevent a poorly named and unneccessary `tools` binary from being placed in users `$GOPATH/bin` directory if they install all packages in go-github repository.

In fact, is there an advantage in placing this file in a separate directory, instead of just inside `github` folder, the same folder as the `github` package where it's being used?

cmd/tools/gen-accessors.go
```diff
@@ -0,0 +1,300 @@
...
+const source = `// Copyright {{.Year}} The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// {{.Filename}} generated by gen-accessors; DO NOT EDIT
+
+package {{.Package}}
+{{if .Imports}}
```
shurcooL: It's possible to simplify the template in this instance by using `{{with .Imports}}` instead of `{{if .Imports}}`.

cmd/tools/gen-accessors.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	// Sort getters by ReceiverType.FieldName
+	sort.Sort(byName(t.Getters))
+
+	var buf bytes.Buffer
+	if err := sourceTmpl.Execute(&buf, t); err != nil {
+		return err
+	}
+	clean, err := format.Source(buf.Bytes())
```
shurcooL: Does `gofmt`ing the output from template have an effect, i.e., does it change something, or is it being done "just in case"?

If it changes something, what is it? Asking because I'm curious.

gmlewis: Yes, it correctly converts spaces to tabs. I worked on this project:
https://godoc.org/google.golang.org/api
and calling gofmt is standard operating procedure when generating Go code.

cmd/tools/gen-accessors.go
gmlewis: We don't want `gen-accessors.go` in the same `github` directory because it is not in the `github` package and may cause confusion. It seems to be common practice to move command-line tools to a subdirectory called `tools` or `cmd` or `tools/cmd` or `cmd/tools`... I'm flexible with the directory name, but I think we want to keep it in a separate directory.

gmlewis:
I believe I've addressed the review comments from @shurcooL.
PTAL.

github/github.go
```diff
@@ -3,6 +3,15 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// go-github authors should run the following commands before
```
willnorris: these instructions should be added to CONTRIBUTING.md.  Either in addition to, or probably instead of, here.

gmlewis: Done.

cmd/gen-accessors/gen-accessors.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// blacklist lists which "struct.method" combos to not generate.
+	blacklist = map[string]bool{
+		"RepositoryContent.GetContent":    true,
+		"Client.GetBaseURL":               true,
+		"Client.GetUploadURL":             true,
+		"ErrorResponse.GetResponse":       true,
+		"RateLimitError.GetResponse":      true,
+		"AbuseRateLimitError.GetResponse": true,
+	}
```
shurcooL: This is a question. What's the pattern of what's included in the blacklist?

Are all entries here simply methods that already exist, that happen to begin with "Get"? And the blacklist is needed to prevent the generated code from creating a duplicate method name, is that so?

cmd/gen-accessors/gen-accessors.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func (t *templateData) dump() error {
+	if len(t.Getters) == 0 {
+		logf("No getters for %v; skipping.", t.filename)
+		return nil
+	}
+
+	// Sort getters by ReceiverType.FieldName
+	sort.Sort(byName(t.Getters))
```
shurcooL: FWIW, you could use `sort.Slice` now to make this a little shorter (i.e., `byName` type and its 3 methods can go away), if you wanted to, since #554 is merged.

(Completely optional, just mentioning it.)

cmd/gen-accessors/gen-accessors.go
gmlewis: The pattern is `<receiver type> + "." + <method name>`.

Yes, the blacklist prevents these methods from being created... either because they already exist or because they would otherwise create problems if they were to be created.

cmd/gen-accessors/gen-accessors.go
gmlewis: I just tried this with Go 1.7 and it failed.
It looks like `sort.Slice` is Go 1.8 only:
https://beta.golang.org/doc/go1.8#sort_slice
and since we still support Go 1.7, I will not yet switch this to `sort.Slice`.

gmlewis:
Thank you for the detailed discussion regarding generators, @shurcooL!
I was not aware of this common usage, and it makes sense, so I've moved `gen-accessors.go` into the `github` directory.

cmd/gen-accessors/gen-accessors.go
shurcooL: Ah, yes, my mistake. I was off by one. :)

shurcooL: Although, this is code that runs during `go generate`, only by developers of this package, not users. So in theory we could require/expect developers to be running 1.8. But, this is absolutely not worth it to save a few lines. Just a fun observation.

gmlewis:
Merging.

gmlewis:
Integrated in https://github.com/google/go-github/commit/cd756c0dfc435cf14d9c696e3b44012f6a4fb5aa

Issue #530, opened 2017-01-26
Support new Webhooks for Projects


gmlewis:
Announcement:
https://developer.github.com/changes/2017-01-24-project-webhooks/

varadarajana:
@gmlewis I would like to work on this issue. Can you please assign this to me?

gmlewis:
Done. Thank you, @varadarajana!

sahildua2305:
@varadarajana how's it going with this one?

shurcooL:
There's an open PR #530 that mentions it.

sahildua2305:
@shurcooL oh yes, alright!

@varadarajana it's always good to mention the issue you're trying to resolve in the PR so that others also know that there is/was a PR for this issue and may get to know the history of an issue. :smile:

varadarajana:
Sure @sahildua2305, I have changed it now.

shurcooL:
This issue was resolved via PR #544, which was recently merged, so closing.

Note, it wasn't closed automatically because neither the PR description, nor the commit message of 2694dcaaa60686a8c69d923b3d365fb9375d7e9a said "Closes #530." or equivalent. /cc @gmlewis FYI.

gmlewis:
Ah, so that is the case where I would need to add to the commit message during "Squash and merge" then. Thanks, @shurcooL.

shurcooL:
Yep, at least one of the PR description or commit message needs to mention that it resolves an issue, otherwise GitHub has no way of knowing what issue, if any, to automatically close.

References:

- https://help.github.com/articles/closing-issues-via-commit-messages/
- https://github.com/blog/1506-closing-issues-via-pull-requests

Note that Pull Request title does not count, it has to be in the body, and it has to match the expected syntax exactly:

> All you have to do is include the special keyword syntax (eg. `fixes #5`) in the **body** of your Pull Request.

PR #544, opened 2017-02-04
Issue530: Resolves fully Issue 530: new web hooks  project, project card and project column

varadarajana:
@shurcooL @gmlewis please review the code, I have written for project, project card and project column


varadarajana:
I created a new branch, and find my earlier changes also  there. Is that an issue?

shurcooL:
> I created a new branch, and find my earlier changes also there. Is that an issue?

Well, they're definitely not needed, and it'd be nice to get rid of them if you can.

The final diff looks okay though, so that's what matters more. All the commits will be squashed into 1 when merging this PR anyway.

varadarajana:
@shurcooL Can you please review these changes.

shurcooL:
Reviewing now.

@varadarajana, is this PR meant to fully resolve issue #530? I see it mentioned in the title, but you should add a phrase like "Resolves #530." to the PR body if that's the case. Otherwise, it's unclear if that issue can be closed after this merged.

github/event_types.go
```diff
@@ -345,6 +345,42 @@ type PingEvent struct {
...
+// ProjectEvent is triggered when project is created, modified or deleted.
+// the webhook event name is "project".
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectevent
+type ProjectEvent struct {
+	Action        *string       `json:"action,omitempty"`
+	Project       *Project      `json:"project,omitempty"`
+	Repository    *Repository   `json:"repository,omitempty"`
+	Organizaition *Organization `json:"organization,omitempty"`
+	Sender        *User         `json:"sender,omitempty"`
```
shurcooL: You should add an extra `Installation` field here.

```
```

Based on what https://developer.github.com/webhooks/#payloads says, all events may have this extra optional field:

> In addition to the fields documented for each event, webhook payloads include the user who performed the event (`sender`) as well as the organization (`organization`) and/or repository (`repository`) which the event occurred on, and for an Integration's webhook may include the installation (`installation`) which an event relates to.

This is related to #505. /cc @bradleyfalzon

github/event_types.go
```diff
@@ -345,6 +345,42 @@ type PingEvent struct {
...
 }
 
+// ProjectEvent is triggered when project is created, modified or deleted.
+// the webhook event name is "project".
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectevent
+type ProjectEvent struct {
+	Action        *string       `json:"action,omitempty"`
+	Project       *Project      `json:"project,omitempty"`
+	Repository    *Repository   `json:"repository,omitempty"`
```
shurcooL: All other events seem to rename this field to `Repo` rather than `Repository`. I think we should do the same here, for consistency.

shurcooL: Also, please add the following comment separator for webhook-only fields:

```
```

github/event_types.go
```diff
@@ -345,6 +345,42 @@ type PingEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// ProjectEvent is triggered when project is created, modified or deleted.
+// the webhook event name is "project".
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectevent
+type ProjectEvent struct {
+	Action        *string       `json:"action,omitempty"`
+	Project       *Project      `json:"project,omitempty"`
```
shurcooL: This event is missing `Changes` field specified in https://developer.github.com/v3/activity/events/types/#events-api-payload-13.

github/event_types.go
```diff
@@ -345,6 +345,42 @@ type PingEvent struct {
...
+// the webhook event name is "project_column".
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectcolumnevent
+type ProjectColumnEvent struct {
+	Action        *string        `json:"action,omitempty"`
+	ProjectColumn *ProjectColumn `json:"project_column,omitempty"`
+	Repository    *Repository    `json:"repository,omitempty"`
+	Organization  *Organization  `json:"organization,omitempty"`
+	Sender        *User          `json:"sender,omitempty"`
+}
```
shurcooL: Same feedback as for above. `Repository` -> `Repo`, add `Installation` field, missing `// The following fields are only populated by Webhook events.` comment, missing `Changes` field, missing `AfterID` field from https://developer.github.com/v3/activity/events/types/#events-api-payload-20.

github/event_types.go
```diff
@@ -345,6 +345,42 @@ type PingEvent struct {
...
+// the webhook event name is "project_card".
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#projectcardevent
+type ProjectCardEvent struct {
+	Action       *string       `json:"action,omitempty"`
+	ProjectCard  *ProjectCard  `json:"project_card,omitempty"`
+	Repository   *Repository   `json:"repository,omitempty"`
+	Organization *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+}
```
shurcooL: Same feedback as for above. `Repository` -> `Repo`, `Organization` -> `Org`, add `Installation` field, missing `// The following fields are only populated by Webhook events.` comment, missing `Changes` field, missing `AfterID` field from https://developer.github.com/v3/activity/events/types/#events-api-payload-19.

github/event_types.go
```diff
@@ -345,6 +345,42 @@ type PingEvent struct {
...
 
+// ProjectEvent is triggered when project is created, modified or deleted.
+// the webhook event name is "project".
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectevent
+type ProjectEvent struct {
+	Action        *string       `json:"action,omitempty"`
+	Project       *Project      `json:"project,omitempty"`
+	Repository    *Repository   `json:"repository,omitempty"`
+	Organizaition *Organization `json:"organization,omitempty"`
```
shurcooL: All other events seem to rename this field to `Org` rather than `Organization`. I think we should do the same here, for consistency.

shurcooL:
@varadarajana, some higher level suggestions. It looks to me that you've taken the "Webhook payload example" sections from the GitHub documentation and mapped that directly to a Go struct.

image

That's a good start, but you seem to have missed the "Events API payload" sections. Those are important because they contain well specified fields in the event. The webhook example may not include all of them, but the structs need to.

However, we need to keep the new structs consistent with what we've already started doing in `event_types.go`. I suggest looking up and down in that file for other event struct definitions and making sure your new events are well aligned. That's where my feedback above comes from.

varadarajana:
@shurcooL Thanks, I got your point, there are few fields missing in example. Let me add them today and submit.

github/event_types.go
varadarajana: @shurcooL This is complete, I have added installation field all the structs.

github/event_types.go
varadarajana: @shurcooL I have added Changes field. I had to change EditChange to include the two nodes Note, From.

github/event_types.go
varadarajana: @shurcooL I have added them. These fields I assumed to be the ones available in example but not in event API Payload.

github/event_types.go
varadarajana: @shurcooL I have added them

github/event_types.go
varadarajana: @shurcooL i have added them, can you please review?

github/event_types.go
varadarajana: @shurcooL I have modified and added this comment.

varadarajana:
@shurcooL can you please review this when you find time?

github/event_types.go
```diff
@@ -149,6 +149,12 @@ type EditChange struct {
 	Title *struct {
 		From *string `json:"from,omitempty"`
 	} `json:"title,omitempty"`
+	Name *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"title,omitempty"`
+	Note *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"title,omitempty"`
```
shurcooL: This should be `json:"note,omitempty"`, not `json:"title,omitempty"`, shouldn't it?

github/event_types.go
```diff
@@ -149,6 +149,12 @@ type EditChange struct {
 	Title *struct {
 		From *string `json:"from,omitempty"`
 	} `json:"title,omitempty"`
+	Name *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"title,omitempty"`
```
shurcooL: This should be `json:"name,omitempty"`, not `json:"title,omitempty"`, shouldn't it?

github/event_types.go
```diff
@@ -149,6 +149,12 @@ type EditChange struct {
...
+	Note *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"title,omitempty"`
 	Body *struct {
 		From *string `json:"from,omitempty"`
 	} `json:"body,omitempty"`
```
shurcooL: Look at the description of `EditChange`:

```Go
// EditChange represents the changes when an issue, pull request, or comment has
// been edited.
type EditChange struct {
```

I'm not sure if it makes the most sense to keep reusing it for project changes, project column changes and project card changes. That becomes a lot of different things all for one common struct, isn't it? (Maybe it's fine, not sure...)

I was considering creating a new struct instead...

But now I'm starting to think, maybe it'd be simpler/better to just inline the struct into the event?

That way, each `changes` field is going to contain exactly the fields that are supposed to be there, nothing more. I think it'd be better API.

It would be harder/more verbose to create new instances of these structs, but I'm not sure if that's something people do often, is it?

This is what it'd look like:

```Go
// ProjectEvent is triggered when project is created, modified or deleted.
// the webhook event name is "project".
//
// GitHub docs https://developer.github.com/v3/activity/events/types/#projectevent
type ProjectEvent struct {
	...
	Changes *struct{
		Name *struct {
		} `json:"name,omitempty"`
		Body *struct {
		} `json:"body,omitempty"`
	} `json:"changes,omitempty"`
	...
}

// ProjectCardEvent is triggered when a project card is created, updated, moved, converted to an issue, or deleted.
// the webhook event name is "project_card".
//
// GitHub docs: https://developer.github.com/v3/activity/events/types/#projectcardevent
type ProjectCardEvent struct {
	...
	Changes *struct{
		Note *struct {
		} `json:"note,omitempty"`
	} `json:"changes,omitempty"`
	...
}

// ProjectColumnEvent is triggered when a project column is created, updated, moved, or deleted.
// the webhook event name is "project_column".
//
// GitHub docs https://developer.github.com/v3/activity/events/types/#projectcolumnevent
type ProjectColumnEvent struct {
	...
	Changes {
		Name *struct {
		} `json:"name,omitempty"`
	} `json:"changes,omitempty"`
	...
}
```

What do you think?

github/event_types.go
```diff
@@ -345,6 +351,56 @@ type PingEvent struct {
...
+	Action  *string     `json:"action,omitempty"`
+	Changes *EditChange `json:"changes,omitempty"`
+	Project *Project    `json:"project,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	Repo         *Repository   `json:"repository,omitempty"`
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
+}
```
shurcooL: Do you mind re-arranging these in the same consistent order as in `activity_events.go` and `messages.go`?

`ProjectEvent`, then `ProjectCardEvent`, then `ProjectColumnEvent`.

github/event_types.go
varadarajana: @shurcooL inline Changes is very good. I agree for since Changes are not used anywhere else, they keep varying across APIs. I will make these and submit.

github/event_types.go
varadarajana: @shurcooL I have re-ordered them.

github/event_types.go
varadarajana: @shurcooL I have made changes to all Changes referenced in event_types.go

github/event_types.go
varadarajana: @shurcool these changes are present in inline struct

github/event_types.go
varadarajana: @shurcool these changes are present in inline struct

github/event_types.go
```diff
@@ -130,17 +130,6 @@ type GollumEvent struct {
...
-// EditChange represents the changes when an issue, pull request, or comment has
-// been edited.
-type EditChange struct {
-	Title *struct {
-		From *string `json:"from,omitempty"`
-	} `json:"title,omitempty"`
-	Body *struct {
-		From *string `json:"from,omitempty"`
-	} `json:"body,omitempty"`
-}
```
shurcooL: When I made https://github.com/google/go-github/pull/544#discussion_r102259209, I didn't mean for you to actually remove `EditChange` and modify the other existing structs, I was referring only to the 3 new event types being added for #530.

However, given you've already done it, and this is probably better in the long term, we should consider keeping it. @gmlewis, what do you think about this?

It shouldn't be a breaking API change for most uses that just access the fields ala `something.Title.From`, but it might still be a breaking change for code that tries to copy or manipulate these structs.

varadarajana: @gmlewis As @shurcooL says, is this fine? Do i need to rollback my changes?

github/event_types.go
gmlewis: I'm not sure that I fully understand the benefit of inlining the structs.

> That way, each changes field is going to contain exactly the fields that are supposed to be there, nothing more. I think it'd be better API.

Inlining the structs (by itself) does not solve this problem. There is no reason why a typed struct could not be created that contains exactly the fields that are supposed to be there with nothing more. It may mean that we have more typed structs, of course (which I think is OK).

As a user of the API, I would rather have an exported typed struct that I can fill in as opposed to having to copy/paste the full source code for the inlined struct if I need to use it.

I think we should err on the side of making the API easier to use for the end user, and inlining the structs seems to me that it makes it more difficult for the end user.

github/event_types.go
gmlewis: I left comments in #544 that I'm not sure I fully agree with inlining the structs, as I think that makes the API more challenging to use for the end user.

shurcooL: In that case, @varadarajana please revert the changes to other structs for now, so we can further discuss just the structs needed for the new functionality being added as part of this PR only. That'll make it easier.

github/event_types.go
shurcooL: @gmlewis Having longer discussions inside inlined comments that may get collapsed is hard, so I'll quote your comment and reply at the bottom.

shurcooL:
@gmlewis, your inline https://github.com/google/go-github/pull/544#discussion_r103988781:

> I'm not sure that I fully understand the benefit of inlining the structs.
>
> > That way, each changes field is going to contain exactly the fields that are supposed to be there, nothing more. I think it'd be better API.
>
> Inlining the structs (by itself) does not solve this problem. There is no reason why a typed struct could not be created that contains exactly the fields that are supposed to be there with nothing more. It may mean that we have more typed structs, of course (which I think is OK).
>
> As a user of the API, I would rather have an exported typed struct that I can fill in as opposed to having to copy/paste the full source code for the inlined struct if I need to use it.
>
> I think we should err on the side of making the API easier to use for the end user, and inlining the structs seems to me that it makes it more difficult for the end user.

Replying here because it's hard to have a long conversation in an inline comment.

> Inlining the structs (by itself) does not solve this problem.

That is true.

In practice, the current situation is that we've reused the same `EditChange` struct for edits to things as varied as issues, pull requests, and comments. There was definitely some overlap between them, which is why it made sense, but it wasn't deemed worth it to increase number of `EditChange` like structs from 1 to many in order to offer more precise fields.

We can reconsider that now.

Let's limit the scope to the new functionality being added in this PR only for this discussion (we can come back to the other existing things later).

There are 3 new event structs being added:

- `ProjectEvent`
- `ProjectCardEvent`
- `ProjectColumnEvent`

Each one has `Changes` field, but all 3 have different contents that are possible inside them.

The reason I suggested considering inlining them is because it makes it effortless to customize the `Changes` contents to match exactly what's documented to be available in the event.

If we were to give them names, it would become 3 new `EditChange`-like structs:

```Go
// ProjectChange represents the changes when a project has been edited.
type ProjectChange struct {
	Name *struct {
	} `json:"name,omitempty"`
	Body *struct {
	} `json:"body,omitempty"`
}

// ProjectCardChange represents the changes when a project card has been edited.
type ProjectCardChange struct {
	Note *struct {
	} `json:"note,omitempty"`
}

// ProjectColumnChange represents the changes when a project column has been edited.
type ProjectColumnChange struct {
	Name *struct {
	} `json:"name,omitempty"`
}
```

How does that look/sound to you?

If we go this way, we'll need to make sure that we don't use up the public API namespace for something more important in the future. Depending on that, we can consider longer (or shorter) names for those Change structs.

gmlewis:
> How does that look/sound to you?

I like that much better.  I think longer names are OK for these structs... editors can help with this situation, and having these named structs I think is a definite win for the end user.

shurcooL:
Ok, SGTM, let's go with that.

> editors can help with this situation, and having these named structs I think is a definite win for the end user.

Do you expect people to be setting them often? I figured that in a vast majority of use cases, their values would be read from a webhook event that was received. The code for that would look the same regardless if the struct is inlined or not:

```Go
func processProjectEvent(e *github.ProjectEvent) error {
	// ...
	oldName := e.Changes.Name.From
	oldBody := e.Changes.Body.From
	// ...
}

**Comment by gmlewis** [comment]
No, I doubt it will be used often, but still prefer the typed structure name.

File: github/event_types.go (hunk 11 above)
Comment (varadarajana) [inline github/event_types.go]: @shurcooL @gmlewis I have made changes for defined structs of ProjectCard, ProjectColumn and project events. Old ones that refer to EditChanges have been kept as old. Please do review and comment.

Comment (varadarajana) [inline github/event_types.go]: @shurcooL @gmlewis I have made changes for defined structs of ProjectCard, ProjectColumn and project events. Old ones that refer to EditChanges have been kept as old. Please do review and comment.

Comment (varadarajana) [inline github/event_types.go]: @shurcooL @gmlewis I have made changes for defined structs of ProjectCard, ProjectColumn and project events. Old ones that refer to EditChanges have been kept as old. Please do review and comment.

Comment (varadarajana) [inline github/event_types.go]: @shurcooL , @gmlewis I have made changes for defined structs of ProjectCard, ProjectColumn and project events. Old ones that refer to EditChanges have been kept as old. Please do review and comment.

Comment (varadarajana) [inline github/event_types.go]: @shurcooL , @gmlewis I have made changes for defined structs of ProjectCard, ProjectColumn and project events. Old ones that refer to EditChanges have been kept as old. Please do review and comment.

File: github/event_types.go [hunk 12]
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// ProjectEvent is triggered when project is created, modified or deleted.
+// the webhook event name is "project".
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectevent
```
gmlewis: s/GitHub docs /GitHub API docs: /

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
...
+	Repo         *Repository   `json:"repository,omitempty"`
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
+}
+
+// ProjectCardEvent is triggered when a project card is created, updated, moved, converted to an issue, or deleted.
+// the webhook event name is "project_card".
+//
+// GitHub docs: https://developer.github.com/v3/activity/events/types/#projectcardevent
```
gmlewis: s/GitHub docs /GitHub API docs: /

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// ProjectEvent is triggered when project is created, modified or deleted.
+// the webhook event name is "project".
```
gmlewis: s/the/The/

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
...
+
+	// The following fields are only populated by Webhook events.
+	Repo         *Repository   `json:"repository,omitempty"`
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
+}
+
+// ProjectCardEvent is triggered when a project card is created, updated, moved, converted to an issue, or deleted.
+// the webhook event name is "project_card".
```
gmlewis: s/the/The/

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
...
+
+	// The following fields are only populated by Webhook events.
+	Repo         *Repository   `json:"repository,omitempty"`
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
+}
+
+// ProjectColumnEvent is triggered when a project column is created, updated, moved, or deleted.
+// the webhook event name is "project_column".
```
gmlewis: s/the/The/

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
...
+//
+// GitHub docs https://developer.github.com/v3/activity/events/types/#projectcolumnevent
```
gmlewis: s/GitHub docs /GitHub API docs: /

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
...
+	Project *Project       `json:"project,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	Repo         *Repository   `json:"repository,omitempty"`
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
+}
+
+// ProjectCardEvent is triggered when a project card is created, updated, moved, converted to an issue, or deleted.
```
gmlewis: nit: that line seems slightly long and I think we attempt to keep the godoc comments shorter in this package, but I know Go style allows any length, and I see other long lines in `github.go` in godoc comment strings, so it's your call.

varadarajana: @gmlewis , The line was not changed keeping in mind earlier comment from @shurcooL, guidelines from go style. So I retained it.

github/event_types.go
varadarajana: @gmlewis this is done

github/event_types.go
varadarajana: @gmlewis this is done

github/event_types.go
varadarajana: @gmlewis this is done

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// ProjectEvent is triggered when project is created, modified or deleted.
+// The webhook event name is "project".
+//
+// GitHub API docs https://developer.github.com/v3/activity/events/types/#projectevent
```
shurcooL: Still missing a colon here, please include it for consistency.

See relevant PR #562.

github/event_types.go
```diff
@@ -332,6 +356,56 @@ type PingEvent struct {
...
+	Repo         *Repository   `json:"repository,omitempty"`
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
+}
+
+// ProjectColumnEvent is triggered when a project column is created, updated, moved, or deleted.
+// The webhook event name is "project_column".
+//
+// GitHub API docs https://developer.github.com/v3/activity/events/types/#projectcolumnevent
```
shurcooL: Also missing colon here, please insert.

github/event_types.go
varadarajana: @shurcooL this is done

github/event_types.go
varadarajana: @shurcooL this is done

varadarajana:
@gmlewis Can you approve these changes too?

gmlewis:
Did you follow the instructions in `CONTRIBUTING.md` like `go generate ...`, etc.?

varadarajana:
@gmlewis 
I normally do a get, diff, vet and test. Now I did a generate, test, vet now. Here is the output.
```
$ go generate github.com/google/go-github/...
$  go test github.com/google/go-github/...
?       github.com/google/go-github/examples/basicauth  [no test files]
ok      github.com/google/go-github/github      7.070s
?       github.com/google/go-github/tests/fields        [no test files]
?       github.com/google/go-github/tests/integration   [no test files]

$  go vet github.com/google/go-github/...
```

gmlewis:
OK, thanks @varadarajana - what I meant to ask was if `go generate` modified `github-accessors.go`?
I was expecting some newly-created accessors for your new named structs that you created in this PR. If it did, can you please add a commit to this PR that includes the changes to this file?

varadarajana:
@gmlewis 
I did a go generate . and found the following added:

```
func (p *ProjectCard) GetNote() string {
        if p == nil || p.Note == nil {
                return ""
        }
        return *p.Note
}

// GetUpdatedAt returns the UpdatedAt field if it's non-nil, zero value otherwise.
func (p *ProjectCard) GetUpdatedAt() Timestamp {
        if p == nil || p.UpdatedAt == nil {
                return Timestamp{}
        }
        return *p.UpdatedAt
}

// GetAction returns the Action field if it's non-nil, zero value otherwise.
func (p *ProjectCardEvent) GetAction() string {
        if p == nil || p.Action == nil {
                return ""
        }
        return *p.Action
}

// GetAfterID returns the AfterID field if it's non-nil, zero value otherwise.
func (p *ProjectCardEvent) GetAfterID() int {
        if p == nil || p.AfterID == nil {
                return 0
        }
        return *p.AfterID
}

// GetCreatedAt returns the CreatedAt field if it's non-nil, zero value otherwise.
func (p *ProjectColumn) GetCreatedAt() Timestamp {
        if p == nil || p.CreatedAt == nil {
                return Timestamp{}
        }
        return *p.CreatedAt
}

// GetID returns the ID field if it's non-nil, zero value otherwise.
func (p *ProjectColumn) GetID() int {
        if p == nil || p.ID == nil {
                return 0
        }
        return *p.ID
}

// GetName returns the Name field if it's non-nil, zero value otherwise.
func (p *ProjectColumn) GetName() string {
        if p == nil || p.Name == nil {
                return ""
        }
        return *p.Name
}
```
I will commit them now.

varadarajana:
@gmlewis @shurcooL , 
Do we need to change contributing.md should include workspace folder structure of the individual instead of github.com/ and so on? 

I wanted to mention
go generate <your working folder> 

instead of what we have now.

shurcooL:
@varadarajana The steps in CONTRIBUTING.md are accurate, but they may not work depending on the exact setup you're using. The steps assume you're working on the package without modifying its import path, which is the best and recommended approach in a general case. But I also think their goal is to describe the intention of what _should_ be done.

If your working directory is at the project root, doing `go generate ./...` does the same thing, and it'll work even if you've modified the import path (for example, to your fork's import path).

Now we check if there's a mismatch between a PR and what should be generated in Travis (see #579), I hope it'll be easier for people to spot what needs to be done.

What you've added in 25cadb0 looks good.

shurcooL:
@gmlewis This PR has both our LGTMs now, I'll let you merge it since I recently merged another one. (Unless you want me to, then I can do it.)

gmlewis:
SGTM, @shurcooL.
I'll merge it now.

gmlewis:
@shurcooL - my `git` foo is severely stressed at this point.  Maybe you can guide me here.

I've gone through this procedure three times now... and got different results, none of which match the ending state shown on the GitHub web page.

I download all the commits to a separate branch, called `pr/544` and switch to that branch.
I run:

`git rebase -i master`
and then go through a series of:
`git mergetool` (resolving conflicts as a I go... the `local` changes appear to be correct, not the `remote` ones, but it is honestly quite confusing with all the intermediary debugging commits... and on the 3rd try I chose all the `remote` changes and still no joy.)
`git rebase --continue`
until it finally says it is happy.
The end results, however, don't match the diffs we want.

Is this your process as well, or do you simply click the green arrow on this page and choose `Rebase and merge` ?

shurcooL:
> Maybe you can guide me here.

Sure.

> Is this your process as well, or do you simply click the green arrow on this page and choose `Rebase and merge` ?

In general, I usually merge PRs via the green "Squash and merge" button when the outcome I'd like is to have the entire PR squashed into one commit, while preserving the individual commits inside the PR, and have GitHub be able to detect that the PR was "merged" rather than "closed".

For that reason, I'd recommend using the green "Squash and merge" button.

However, I also have no problem with doing the merge manually on the command line, but when done that way, I have to pick one of the two:

- Mark the PR as "merged" rather than "closed"
- Preserve the commit history in the PR

It's not possible to merge via command line and achieve both (as far as I know).

Let's assume you'd want to merge via command line and preserve the commit history in the PR, let me describe the steps for that.

Essentially, you want to think of it as these operations in this specific order:

1. Checkout the PR branch.
2. **Squash** all the commits in the PR branch (in this case, 26 commits) into 1 commit.
3. **Rebase** that one final commit onto latest `master` of go-github.
4. Edit the commit message and insert "Closes #544." line at the end.

Does that make sense?

When doing step 2, I usually perform that by doing an interactive rebase on top of the commit this PR was based on (not on top of latest master), and marking all commits except the first one as "squash" instead of "pick". That's a simple way of doing it, but I'm sure there are other ways.

Let me know if you have any further questions or if something's unclear, I'm very happy to help further.

shurcooL:
> `git rebase -i master`
> and then go through a series of:
> `git mergetool` (resolving conflicts as a I go... the `local` changes appear to be correct, not the `remote` ones, but it is honestly quite confusing with all the intermediary debugging commits... and on the 3rd try I chose all the `remote` changes and still no joy.)
> `git rebase --continue`
> until it finally says it is happy.
> The end results, however, don't match the diffs we want.

You definitely should not have to resolve any conflicts while merging this. That is unnecessary work, and it's error prone. This PR has no conflicts and should be merge-able without any manual conflict resolution. The key is to squash all the commits first (all sequential commits always merge effortlessly), before rebasing it.

shurcooL:
@gmlewis I think there's a higher level git command you can use to make this operation much simpler:

```
git merge --squash <PR-branch>
git commit
```

See [this SO question and answer](http://stackoverflow.com/questions/3697178/git-merge-all-changes-from-another-branch-as-a-single-commit).

Basically, that makes performing the squash operation very easy and infallible. It's possible because this PR has no merge conflicts.

So, the overall process can be something like this:

```
git checkout master
git fetch origin pull/544/head:pr-544 # Fetch the PR into local branch named "pr-544".
git merge --squash
git commit # And write a nice commit message, and add "Closes 544." at the end, or else this PR won't be closed.
# test anything locally, and push to remote master once happy
```

I've [tried it locally](https://cloud.githubusercontent.com/assets/1924134/23690617/0a99dbee-0391-11e7-8dd1-df7c3ffae596.png) (minus the push step, of course) and the diff of the final commit looks exactly like in this PR's "Files changed" tab.

But I'd still recommend using the green "Squash and merge" button because it would mark this PR as "merged" instead of "closed", and no need to add "Closes #544." to the commit message.

gmlewis:
That was extremely helpful, @shurcooL!
I've learned a lot and appreciate your time and patience in writing up that explanation!
OK, I'm going to use the "Squash and merge" button now.

gmlewis:
I didn't edit the commit message at all for fear of changing the hashes. I hope that is OK.

shurcooL:
> That was extremely helpful, @shurcooL!
> I've learned a lot and appreciate your time and patience in writing up that explanation!

No problem, happy to be useful!

> OK, I'm going to use the "Squash and merge" button now.

Looks great!

> I didn't edit the commit message at all for fear of changing the hashes. I hope that is OK.

It's not a problem, but I think it would've been better to write a clean commit message. The current commit message includes many details that were only relevant as part of the work in progress. That history is still available in this PR, and it's not helpful in master. The commit message, ideally, should describe what the final commit does and why.

I'm not sure what you mean by "fear of changing the hashes". You should feel free to edit the commit message when using the "Squash and merge" button, it works as one would expect without any issues.

You can look at the commit message of previous commit cfbfe91dfc6f331a809a2c942f64b6bad65d9c91 as an example - I wrote that with via the "Squash and merge" interface on GitHub as well.

gmlewis:
Oh, I see.  Thanks, @shurcooL!

Now I understand that my final message is actually the final commit itself and GitHub will still take this as the single commit.  I was hesitant because I thought that if I change that message, it was going to add on yet another commit for my message change which, in retrospect, would not make sense. Good to know.

PR #546, opened 2017-02-12
Return nil when there is no error 

### LINKED ISSUE #536: Prefer to explicitly return nil when there is no error for code readability
A common pattern in this repo is:

```go
if err != nil {
  return ..., err
}
return ..., err
```

and we would like to change the last return to explicitly `return ..., nil` for Go readability.

nspragg:
Return nil explicitly when there is no error code to aid readability (and consistency). 

Fixes #536

### DIFF
--- a/github/activity_notifications.go
+++ b/github/activity_notifications.go
@@ -67,7 +67,7 @@ func (s *ActivityService) ListNotifications(opt *NotificationListOptions) ([]*No
 		return nil, resp, err
 	}
 
-	return notifications, resp, err
+	return notifications, resp, nil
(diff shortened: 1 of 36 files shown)


shurcooL:
That said, this does a lot to help resolve #536, but it won't be fully resolved. Before this PR, I found around 148 instances of ", resp, err\n}\n\n", and after this PR there's 35 more. Some of them are valid code, but some can still be simplified.

nspragg:
@shurcooL - I'm happy to fix the reminding (35) instances.

PR #547, opened 2017-02-12
fix doc link for IssuesService.AddLabelsToIssue

stapelberg:


github/issues_labels.go
```diff
@@ -137,7 +137,7 @@ func (s *IssuesService) ListLabelsByIssue(owner string, repo string, number int,
```
shurcooL: If you'd like, you can fix the method `ListLabelsByIssue ` above too. It also has an incorrect documentation link. It should be https://developer.github.com/v3/issues/labels/#list-labels-on-an-issue instead of http://developer.github.com/v3/issues/labels/#list-all-labels-for-this-repository.

stapelberg: Good catch, done.

PR #550, opened 2017-02-15
Return nil explicitly when there is no error.

### LINKED ISSUE #536: Prefer to explicitly return nil when there is no error for code readability

nspragg:
Resolves #536. Explicitly return nil when there is no error. Hopefully, this should address outstanding instances of the following pattern: 

```go
if err != nil {
  return ..., err
}
return ..., err
```

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -739,7 +739,7 @@ func (c *Client) RateLimit() (*Rate, *Response, error) {
 		return nil, resp, err
 	}
 	if limits == nil {
-		return nil, resp, fmt.Errorf("RateLimits returned nil limits and error; unable to extract Core rate limit")
+		return nil, resp, errors.New("RateLimits returned nil limits and error; unable to extract Core rate limit")
 	}
 	return limits.Core, resp, nil
 }
--- a/github/github.go
+++ b/github/github.go
@@ -739,7 +739,7 @@ func (c *Client) RateLimit() (*Rate, *Response, error) {
 		return nil, nil, err
 	}
 
-	return limits.Core, resp, err
+	return limits.Core, resp, nil
 }
 
 // RateLimits returns the rate limits for the current client.
--- a/github/github.go
+++ b/github/github.go
@@ -741,7 +741,6 @@ func (c *Client) RateLimit() (*Rate, *Response, error) {
 	if limits == nil {
 		return nil, resp, fmt.Errorf("RateLimits returned nil limits and error; unable to extract Core rate limit")
 	}
-
 	return limits.Core, resp, nil
 }
 
--- a/github/github.go
+++ b/github/github.go
@@ -736,10 +736,10 @@ func category(path string) rateLimitCategory {
 func (c *Client) RateLimit() (*Rate, *Response, error) {
 	limits, resp, err := c.RateLimits()
 	if err != nil {
-		return nil, nil, err
+		return nil, resp, err
 	}
 	if limits == nil {
-		return nil, nil, fmt.Errorf("RateLimits returned nil limits and error. Unable to extract Core rate limit")
+		return nil, resp, fmt.Errorf("RateLimits returned nil limits and error; unable to extract Core rate limit")
 	}
 
 	return limits.Core, resp, nil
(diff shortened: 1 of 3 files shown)


github/github.go
```diff
@@ -739,7 +739,7 @@ func (c *Client) RateLimit() (*Rate, *Response, error) {
 		return nil, nil, err
 	}
 
-	return limits.Core, resp, err
+	return limits.Core, resp, nil
```
shurcooL: Hmm, this one won't work, it changes behavior. But I think the problem lies above.

Currently, it checks if limits is nil, and always returns err. In the (possibly impossible) path where limits is not nil, but err is also not nil, this change would incorrectly return nil error.

However, I think you should change the checks above to be like this:

```Go
limits, resp, err := c.RateLimits()
if err != nil {
	return nil, nil, err
}
if limits == nil {
	return nil, nil, fmt.Errorf("RateLimits() returned nil error and nil limits, cannot extract Core rate limit")
}
```

Then it's fine to do `return limits.Core, resp, nil` here.

We could also leave this as is, but I think it's unfriendly API to return `nil, nil, nil`, so changing it to always return non-nil error on failure would be better.

nspragg: Well spotted. I agree with checking both limit and error in this case (better than returning nil, nil, nil)

Incidentally, no tests failed. Is there any motivation to increase the test coverage in the project in the future?

shurcooL: Having high but practical test coverage is always a goal, and I think we're doing pretty well there (look at all the _test.go files). This finding doesn't change that. This is a particular rare edge case that I wouldn't expect to be fully caught by unit tests.

Also note that this is a deprecated method, so it's not worth spending a lot of time/effort on.

github/github.go
```diff
@@ -735,9 +735,12 @@ func category(path string) rateLimitCategory {
 // Deprecated: RateLimit is deprecated, use RateLimits instead.
 func (c *Client) RateLimit() (*Rate, *Response, error) {
 	limits, resp, err := c.RateLimits()
-	if limits == nil {
+	if err != nil {
 		return nil, nil, err
 	}
+	if limits == nil {
+		return nil, nil, fmt.Errorf("RateLimits returned nil limits and error. Unable to extract Core rate limit")
```
shurcooL: I think we should return original `resp` instead of `nil` here. It's either going to be nil or some valid response. If there's an error, and a valid response, it's better to return that to the caller so they can look into why the error happened, rather than dropping the response information.

Here, and in the if check above.

shurcooL: About the error message, see https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

I don't think you should separate it into multiple sentences and capitalize the letters. First letter happens to be capital because it refers to `RateLimits` the proper noun.

github/github.go
```diff
@@ -735,11 +735,14 @@ func category(path string) rateLimitCategory {
 // Deprecated: RateLimit is deprecated, use RateLimits instead.
 func (c *Client) RateLimit() (*Rate, *Response, error) {
 	limits, resp, err := c.RateLimits()
+	if err != nil {
+		return nil, resp, err
+	}
 	if limits == nil {
-		return nil, nil, err
+		return nil, resp, fmt.Errorf("RateLimits returned nil limits and error; unable to extract Core rate limit")
 	}
```
shurcooL: Nitpick, but I would remove this blank line now. It does not provide value at this point IMO.

Without the blank line separating them, it would be a little more visible that `limits` is checked for being non-`nil` before dereferencing `limits.Core`.

github/github.go
```diff
@@ -735,11 +735,13 @@ func category(path string) rateLimitCategory {
 // Deprecated: RateLimit is deprecated, use RateLimits instead.
 func (c *Client) RateLimit() (*Rate, *Response, error) {
 	limits, resp, err := c.RateLimits()
+	if err != nil {
+		return nil, resp, err
+	}
 	if limits == nil {
-		return nil, nil, err
+		return nil, resp, fmt.Errorf("RateLimits returned nil limits and error; unable to extract Core rate limit")
```
gmlewis: s/fmt.Errorf/errors.New/

PR #555, opened 2017-02-18
Remove deprecated identifiers.

### LINKED ISSUE #526: support the context package

shurcooL:
**DO NOT MERGE:** Please note the base of this PR is currently the `require-go17-plus` branch rather than `master`. This PR is for review purposes only. Once #554 is merged, I will change base of this PR to `master` and it can be mergeable then.

This is part 2 of 3 PRs to resolve #526 (as was suggested to split PR #529 in https://github.com/google/go-github/pull/529#discussion_r101895336).

Please see commit message for change details:

> This is a breaking API change. However, it removes things that were deprecated and shouldn't be used anymore (including something that's available in standard library).
>
> Next commit will be a large breaking API change anyway, so it makes more sense to get rid of the deprecated things instead of updating their API.
>
> `http.StatusUnprocessableEntity` was added in Go 1.7, and so we can use it instead of own equivalent constant. Anyone who was previously using it should switch to using `http.StatusUnprocessableEntity` as well. Consider this commit to deprecate `StatusUnprocessableEntity` and remove it in one.

As I wrote in https://github.com/google/go-github/pull/529#issuecomment-280819989:

>  I kinda went to an extreme and removed all deprecated things, even ones that wouldn't be affected by API changes. Please see the entire change and review it carefully, and let me know your thoughts. Do we delete only the affected deprecated APIs? All of them (as currently is)? Or don't remove any?

### DIFF
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -49,8 +49,6 @@ func (e *Event) Payload() (payload interface{}) {
 		payload = &IntegrationInstallationEvent{}
 	case "IntegrationInstallationRepositoriesEvent":
 		payload = &IntegrationInstallationRepositoriesEvent{}
-	case "IssueActivityEvent":
-		payload = &IssueActivityEvent{}
 	case "IssueCommentEvent":
 		payload = &IssueCommentEvent{}
 	case "IssuesEvent":
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -130,19 +130,6 @@ type GollumEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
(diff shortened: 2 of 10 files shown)


shurcooL:
Ok, pulling the trigger on this and #554. Then we'll just have #529 to clean up, review, and eventually merge.

shurcooL:
Side note, I had to rebase this on latest master to resolve conflicts, because the commit message of the previous commit changed (added PR number to the title). The actual contents of the change haven't changed.

PR #557, opened 2017-02-18
Add requestreviewers

### LINKED ISSUE #516: Add support for Preview Review Requests API
Announcement:
https://developer.github.com/changes/2016-12-16-review-requests-api/

sevki:
this commit adds the RequestReviewers function as described here

https://developer.github.com/v3/pulls/review_requests/#create-a-review-request

btw: sorry I'm new to the codebase and followed the CONTRIBUTING to the best
of my ability, if you have any fixes I'll do my best to address them 

Resolves #516.


shurcooL:
@sevki, is this meant to resolve (partially or in full) #516?

sevki:
@shurcooL complete but I wanted to get something clear before I did the other two (list, delete).
I am not sure about the naming, it's not analogous to the GitHub API because [PullRequestReviewRequest](https://github.com/google/go-github/blob/master/github/pulls_reviews.go#L42) was already a thing. I think the request there, correct me if I'm wrong, refers to the HTTP request. and `PullRequestReviewRequestRequest` sounded very Java-like. So I wanted to avoid that. Are you ok with the patch as is?

shurcooL:
> [PullRequestReviewRequest](https://github.com/google/go-github/blob/master/github/pulls_reviews.go#L42) was already a thing. I think the request there, correct me if I'm wrong, refers to the HTTP request. and `PullRequestReviewRequestRequest` sounded very Java-like.

We have a pattern where some structs have `Request` or `Options` suffix. Usually, `Request` suffix is used for methods that create new things. It's a request to create that thing. And options are usually used for other methods, like options for listing. E.g.:

```Go
// IssueRequest represents a request to create/edit an issue.
// It is separate from Issue above because otherwise Labels
// and Assignee fail to serialize to the correct JSON.
type IssueRequest struct {
```

```Go
// MilestoneListOptions specifies the optional parameters to the
// IssuesService.ListMilestones method.
type MilestoneListOptions struct 
```

So, in the case of `PullRequestReviewRequest`:

```Go
// PullRequestReviewRequest represents a request to create a review.
type PullRequestReviewRequest struct {
```

It's a request to create a PR review. It's a parameter for `CreateReview` and `SubmitReview` methods.

From what I see in https://developer.github.com/v3/pulls/review_requests/, there are 3 operations that can be done with "review requests":

- List
- Create
- Delete

But review requests are basically github usernames. Both Create and Delete operations contain nothing more than an array of user logins. List operation returns an array of users.

I think we can get away without creating any structs for these 3 methods. That would be simpler.

How about something like this?

```Go
// ListReviewers lists users whose reviews have been requested on the specified pull request.
//
// GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#list-review-requests
func (s *PullRequestsService) ListReviewers(owner, repo string, number int) (*[]User, *Response, error) {
	...
}

// RequestReviewers creates a review request for the provided GitHub users for the specified pull request.
//
// GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#create-a-review-request
func (s *PullRequestsService) RequestReviewers(owner, repo string, number int, reviewers []string) (*PullRequest, *Response, error) {
	...
}

// RemoveReviewers removes the review request for the provided GitHub users for the specified pull request.
//
// GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#delete-a-review-request
func (s *PullRequestsService) RemoveReviewers(owner, repo string, number int, reviewers []string) (*Response, error) {
	...
}
```

It is inspired by similar operations that can be done on issue assignees. See https://github.com/google/go-github/blob/master/github/issues_assignees.go.

varadarajana:
@sevki in your PR createRequest is done, please tell me if I can take up list and Delete methods or you have already finished them.

sevki:
@varadarajana ah yeah they are, sorry forgot to push them.

varadarajana:
@sevki The build is failing here.

```
github/pulls_reviews_request.go:79: not enough arguments in call to s.client.Do
The command "go test -v -race ./..." exited with 2.
```
I think it might be due to context parameter added recently, can you please check if you have the latest version from the mainline.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		t.Errorf("PullRequests.RequestReviewers returned error: %v", err)
+	}
+	_login := "octocat"
+	_id := 1
+
+	want := []User{
+		{
+			Login: &_login,
+			ID:    &_id,
```
shurcooL: Go style doesn't use underscores in identifier names, see https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps.

Also, these variables are not needed. This can be replaced with:

```Go
{
	Login: String("octocat"),
	ID:    Int(1),
}
```

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+]`
+
+	mux.HandleFunc("/repos/o/r/pulls/1/requested_reviewers", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypePullRequestReviewsPreview)
+		fmt.Fprintf(w, sampleResponse)
+	})
+
+	// This returns a PR, unmarshalling of which is tested elsewhere
+	have, _, err := client.PullRequests.ListReviewers(context.Background(), "o", "r", 1)
```
shurcooL: Rename `have` to `reviewers` for consistency with other tests.

Alternatively, use the word `got` rather than `have`. Some other tests use `got`, but no one uses `have`.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+  }
+]`
+
+	mux.HandleFunc("/repos/o/r/pulls/1/requested_reviewers", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypePullRequestReviewsPreview)
+		fmt.Fprintf(w, sampleResponse)
+	})
+
+	// This returns a PR, unmarshalling of which is tested elsewhere
```
shurcooL: I don't understand this comment. `ListReviewers` returns a `[]*User`, not `*PullRequest`. Is there a mistake here?

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+	_login := "octocat"
+	_id := 1
+
+	want := []User{
+		{
+			Login: &_login,
+			ID:    &_id,
+		},
+	}
+	if !reflect.DeepEqual(have, &want) {
```
shurcooL: Would it work without taking address of `want`, i.e.:

```Go
if !reflect.DeepEqual(have, want) {
```

Or would that fail? Other tests don't seem to take address of `want` usually.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+	_, err := client.PullRequests.RemoveReviewers(context.Background(), "o", "r", 1, logins)
+	if err != nil {
+		t.Errorf("PullRequests.RequestReviewers returned error: %v", err)
+	}
+
+	want := reviewers{
+		Reviewers: logins,
+	}
+	if !reflect.DeepEqual(have, want) {
+		t.Errorf("PullRequests.ListReviews returned %+v, want %+v", have, want)
```
shurcooL: This is misleading an not accurate. It says "PullRequests.ListReviews", but the method that was called was PullRequests.RemoveReviewers".

I think you should move this check directly inside `mux.HandleFunc`.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func TestRequestReviewers(t *testing.T) {
+	setup()
+	defer teardown()
+
+	type reviewers struct {
+		Reviewers []string `json:"reviewers,omitempty"`
+	}
+	have := reviewers{}
```
shurcooL: Don't use `have` here, see detailed comments below.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func TestRemoveReviewers(t *testing.T) {
+	setup()
+	defer teardown()
+
+	type reviewers struct {
+		Reviewers []string `json:"reviewers,omitempty"`
+	}
+	have := reviewers{}
```
shurcooL: Don't use `have` here either, see comments below.

github/pulls_reviewers.go
```diff
@@ -0,0 +1,83 @@
...
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// RequestReviewers submits a set of logins to be potential reviewers on a PR.
```
shurcooL: I think it'd be nicer to stay closer to the language the GitHub API docs use and say that this method creates a "review request". What do you think of what I wrote earlier?

```
// RequestReviewers creates a review request for the provided GitHub users for the specified pull request.
```

That seems more consistent with the other endpoints (like the remove one).

github/pulls_reviewers_test.go
sevki: :+1: Looks like I mixed Remove and Request :smiley:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: it breaks the tests. 
image

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
shurcooL: What does the error message say?

shurcooL: Never mind, I see that it's because `ListReviewers` returns `*[]User` and not `[]User`. Your new version looks good.

sevki: @shurcooL I looked at a couple of things and saw they return pointers to arrays, is that not the consistent thing to do? if not I can change it

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+		t.Errorf("PullRequests.RequestReviewers returned error: %v", err)
+	}
+
+}
+
+func TestListReviewers(t *testing.T) {
+	setup()
+	defer teardown()
+
+	sampleResponse := `[
```
shurcooL: You can inline this, it's used in one place. Also, make it a single line, it's short enough to be readable:

```JSON
[{"login": "octocat", "id": 1}]
```

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+		if err := json.Unmarshal(b, &reviewers); err != nil {
+			return
+		}
+
+		want := reviewersRequest{
+			Reviewers: logins,
+		}
+		if !reflect.DeepEqual(reviewers, want) {
+			t.Errorf("PullRequests.RemoveReviewers returned %+v, want %+v", reviewers, want)
+		}
```
shurcooL: I think you can replace all of the above, starting with `b, err := ioutil.ReadAll(r.Body)`, with:

```Go
testBody(t, r, `{"reviewers":["octocat","googlebot"]}`)
```

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+	_, _, err := client.PullRequests.RequestReviewers(context.Background(), "o", "r", 1, logins)
+	if err != nil {
+		t.Errorf("PullRequests.RequestReviewers returned error: %v", err)
+	}
+}
+
+func TestRemoveReviewers(t *testing.T) {
+	setup()
+	defer teardown()
+	logins := []string{"octocat", "googlebot"}
```
shurcooL: Inline `logins`, it's used in one place.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+		reviewers := reviewersRequest{}
+		if err := json.Unmarshal(b, &reviewers); err != nil {
+			return
+		}
+		want := reviewersRequest{
+			Reviewers: logins,
+		}
+		if !reflect.DeepEqual(reviewers, want) {
+			t.Errorf("PullRequests.RequestReviewers returned %+v, want %+v", reviewers, want)
+		}
```
shurcooL: I think you can replace all of the above, starting with `b, err := ioutil.ReadAll(r.Body)`, with:

```Go
testBody(t, r, `{"reviewers":["octocat","googlebot"]}`)
```

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+type reviewersRequest struct {
+	Reviewers []string `json:"reviewers,omitempty"`
+}
+
+func TestRequestReviewers(t *testing.T) {
+	setup()
+	defer teardown()
+
+	logins := []string{"octocat", "googlebot"}
```
shurcooL: Inline `logins`, it's used in one place.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+type reviewersRequest struct {
+	Reviewers []string `json:"reviewers,omitempty"`
+}
```
shurcooL: If you apply my suggestions below, this won't be needed, so remove it.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		want := reviewersRequest{
+			Reviewers: logins,
+		}
+		if !reflect.DeepEqual(reviewers, want) {
+			t.Errorf("PullRequests.RequestReviewers returned %+v, want %+v", reviewers, want)
+		}
+	})
+
+	// This returns a PR, unmarshalling of which is tested elsewhere
```
shurcooL: It's not hard to test it. You don't need to test all fields. Just include the PR number.

In the response handler above, add this at the end:

```Go
fmt.Fprint(w, `{"number":1}`)
```

And then do:

```Go
pull, _, err := client.PullRequests.RequestReviewers(...)
...

want := &PullRequest{Number: Int(1)}
if !reflect.DeepEqual(pull, want) {
	t.Errorf("PullRequests.RequestReviewers returned %+v, want %+v", pull, want)
}
```

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+		if !reflect.DeepEqual(reviewers, want) {
+			t.Errorf("PullRequests.RemoveReviewers returned %+v, want %+v", reviewers, want)
+		}
+	})
+
+	_, err := client.PullRequests.RemoveReviewers(context.Background(), "o", "r", 1, logins)
+	if err != nil {
+		t.Errorf("PullRequests.RequestReviewers returned error: %v", err)
+	}
+
```
shurcooL: Nitpick, remove the unnecessary and inconsistent blank line.

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,119 @@
...
+  {
+    "login": "octocat",
+    "id": 1
+  }
+]`
+
+	mux.HandleFunc("/repos/o/r/pulls/1/requested_reviewers", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypePullRequestReviewsPreview)
+		fmt.Fprintf(w, sampleResponse)
```
shurcooL: Don't use `fmt.Fprintf` where `fmt.Fprint` will suffice.

github/pulls_reviewers.go
```diff
@@ -0,0 +1,83 @@
...
+	"context"
+	"fmt"
+)
+
+// RequestReviewers creates a review request for the provided GitHub users for the specified pull request.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#create-a-review-request
+func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo string, number int, logins []string) (*PullRequest, *Response, error) {
+	u := fmt.Sprintf("repos/%s/%s/pulls/%d/requested_reviewers", owner, repo, number)
+	reviewers := struct {
```
shurcooL: Nitpick, `RemoveReviewers` has a blank line between `u` and `reviewers`, but this doesn't. Please make it consistent.

sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
sevki: :+1:

github/pulls_reviewers_test.go
shurcooL: You said 👍 here, but you didn't apply this change. Did you miss it?

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	want := &[]User{
+		{
+			Login: String("octocat"),
+			ID:    Int(1),
+		},
+	}
+	if !reflect.DeepEqual(reviewers, want) {
+		t.Errorf("PullRequests.ListReviews returned %+v, want %+v", reviewers, want)
```
shurcooL: s/PullRequests.ListReviews/PullRequests.ListReviewers

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	mux.HandleFunc("/repos/o/r/pulls/1/requested_reviewers", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		testHeader(t, r, "Accept", mediaTypePullRequestReviewsPreview)
+		testBody(t, r, `{"reviewers":["octocat","googlebot"]}`)
+	})
+
+	_, err := client.PullRequests.RemoveReviewers(context.Background(), "o", "r", 1, []string{"octocat", "googlebot"})
+	if err != nil {
+		t.Errorf("PullRequests.RequestReviewers returned error: %v", err)
```
shurcooL: s/PullRequests.RequestReviewers/PullRequests.RemoveReviewers

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,83 @@
...
+  {
+    "login": "octocat",
+    "id": 1
+  }
+]`)
+	})
+
+	reviewers, _, err := client.PullRequests.ListReviewers(context.Background(), "o", "r", 1)
+	if err != nil {
+		t.Errorf("PullRequests.ListReviewers error: %v", err)
```
shurcooL: For consitency with lines 34 and 50 (and elsewhere in the project), insert the word "returned":

```Go
t.Errorf("PullRequests.ListReviewers returned error: %v", err)
```

github/pulls_reviewers_test.go
sevki: ooops.

sevki:
@shurcooL so the latest changes are going to wail because `testBody()` don't seem to trim whitespace. the request body seems to end with a new line but the inlined JSON doesn't. Is this an issue you are aware of or am I doing something wrong?

shurcooL:
> @shurcooL so the latest changes are going to wail because testBody() don't seem to trim whitespace. the request body seems to end with a new line but the inlined JSON doesn't. Is this an issue you are aware of or am I doing something wrong?

That's fine, just add a newline to expected JSON output. See here for example:

https://github.com/google/go-github/blob/9be6ce168fa69a3cc2db27a776950a42d18c2ae0/github/activity_notifications_test.go#L78

sevki:
can I send a PR to trim whitespace when reading the testBody?

shurcooL:
Also, @sevki, please prefer to push new commits rather than amending previous ones and force pushing. New commits make it easier to review just the changes you've done.

shurcooL:
> can I send a PR to trim whitespace when reading the testBody?

I'd rather we don't do that, I think it's better to be explicit. If we trim whitespace in `testBody`, it wouldn't catch an issue where we accidentally send too much unwanted whitespace.

sevki:
cool

github/pulls_reviewers_test.go
```diff
@@ -58,12 +58,12 @@ func TestListReviewers(t *testing.T) {
 	mux.HandleFunc("/repos/o/r/pulls/1/requested_reviewers", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", mediaTypePullRequestReviewsPreview)
-		fmt.Fprint(w, `[{"login": "octocat","id": 1 }]`)
+		fmt.Fprint(w, `[{"login": "octocat","id":1}]`)
```
shurcooL: Is there a reason you decided to keep the space between `"login":` and `"octocat"`, but not between `"id":` and `1`?

sevki: not using monospace fonts so I missed it.

sevki:
@shurcooL thanks for the review.

github/pulls_reviewers.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, resp, err
+	}
+
+	return r, resp, nil
+}
+
+// ListReviewers lists users whose reviews have been requested on the specified pull request.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#list-review-requests
+func (s *PullRequestsService) ListReviewers(ctx context.Context, owner, repo string, number int) (*[]User, *Response, error) {
```
gmlewis: There is no need to return a pointer to a slice.  Just return the slice which is already a reference type.

`...([]User, *Response, error) {`

github/pulls_reviewers.go
```diff
@@ -0,0 +1,84 @@
...
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypePullRequestReviewsPreview)
+
+	var users []User
+	resp, err := s.client.Do(ctx, req, &users)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return &users, resp, nil
```
gmlewis: `return users, resp, nil`

github/pulls_reviewers_test.go
```diff
@@ -0,0 +1,78 @@
...
+		testHeader(t, r, "Accept", mediaTypePullRequestReviewsPreview)
+		fmt.Fprint(w, `[{"login":"octocat","id":1}]`)
+	})
+
+	reviewers, _, err := client.PullRequests.ListReviewers(context.Background(), "o", "r", 1)
+	if err != nil {
+		t.Errorf("PullRequests.ListReviewers returned error: %v", err)
+	}
+
+	want := &[]User{
```
gmlewis: `want := []User{`

github/pulls_reviewers.go
shurcooL: @gmlewis, per https://godoc.org/github.com/google/go-github/github#UsersService.ListAll and https://github.com/google/go-github/issues/180#issuecomment-226816236, shouldn't we go with `[]*User`?

gmlewis: Wow! Yes! Where was my brain?

github/pulls_reviewers_test.go
shurcooL: Based on https://github.com/google/go-github/pull/557#discussion_r103596225, this should become `want := []*User{`.

github/pulls_reviewers_test.go
```diff
@@ -66,8 +66,8 @@ func TestListReviewers(t *testing.T) {
 		t.Errorf("PullRequests.ListReviewers returned error: %v", err)
 	}
 
-	want := &[]User{
-		{
+	want := []*User{
+		&User{
```
shurcooL: It's not neccessary to include `&User{` here, just `{` as before will do.

In fact, it causes Travis to fail because we check that `gofmt -s` doesn't find any missed opportunities to simplify:

image

The rest of this change is perfect.

sevki: :+1:

sevki:
@shurcooL done.

shurcooL:
@gmlewis PTAL.

sevki:
@gmlewis my pleasure. @shurcooL did all the heavy lifting :smile:

shurcooL:
Thanks @sevki!

PR #561, opened 2017-02-22
Update API inside documentation.

shurcooL:
- Update documentation with new context parameter.
- Update remaining uses of oauth2.NoContext.

This is a followup to #529.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ func main() {
   ts := oauth2.StaticTokenSource(
     &oauth2.Token{AccessToken: "... your access token ..."},
   )
-  tc := oauth2.NewClient(oauth2.NoContext, ts)
+  tc := oauth2.NewClient(context.Background(), ts)
 
   client := github.NewClient(tc)
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ access different parts of the GitHub API. For example:
 client := github.NewClient(nil)
 
 // list all organizations for user "willnorris"
-orgs, _, err := client.Organizations.List("willnorris", nil)
+orgs, _, err := client.Organizations.List(ctx, "willnorris", nil)
 ```
 
 Some API methods have optional parameters that can be passed. For example:
@@ -32,7 +32,7 @@ client := github.NewClient(nil)
 
 // list public repositories for org "github"
 opt := &github.RepositoryListByOrgOptions{Type: "public"}
-repos, _, err := client.Repositories.ListByOrg("github", opt)
+repos, _, err := client.Repositories.ListByOrg(ctx, "github", opt)
 ```
 
 The services of a client divide the API into logical chunks and correspond to
@@ -60,7 +60,7 @@ func main() {
   client := github.NewClient(tc)
 
   // list all repositories for the authenticated user
-  repos, _, err := client.Repositories.List("", nil)
+  repos, _, err := client.Repositories.List(ctx, "", nil)
 }
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -55,12 +55,12 @@ func main() {
   ts := oauth2.StaticTokenSource(
     &oauth2.Token{AccessToken: "... your access token ..."},
   )
-  tc := oauth2.NewClient(oauth2.NoContext, ts)
+  tc := oauth2.NewClient(context.Background(), ts)
```
gmlewis: What about adding a line before this:
`ctx := context.Background()`
and then changing this line to use `ctx` ?

github/doc.go
```diff
@@ -45,12 +45,12 @@ use it with the oauth2 library using:
 		ts := oauth2.StaticTokenSource(
 			&oauth2.Token{AccessToken: "... your access token ..."},
 		)
-		tc := oauth2.NewClient(oauth2.NoContext, ts)
+		tc := oauth2.NewClient(context.Background(), ts)
```
gmlewis: Same comment here:
What about adding a line before this:
`ctx := context.Background()`
and then changing this line to use `ctx` ?

README.md
shurcooL: That seems reasonable.

gmlewis:
Merging.

gmlewis:
@shurcooL - did I mess something up here?  I see `Closed with unmerged commits` which I don't remember seeing before.

shurcooL:
> I see `Closed with unmerged commits` which I don't remember seeing before.

Where did you see that?

It definitely looks like your local repository is in a weird state. You should probably make a backup of your current local master branch, then reset your local master to match the upstream master (so that it's not both behind and ahead).

I can see that the remote master has 2 new commits since last night. They are 93f95bf99fdd49bad208a6a555090ce882d2653e and 28065814ab889552aebe99cd41b35bc5e31063fe.

1.	93f95bf99fdd49bad208a6a555090ce882d2653e is you merging #538.

	I've compared the difference between https://patch-diff.githubusercontent.com/raw/google/go-github/pull/538.diff (the PR) and https://github.com/google/go-github/commit/93f95bf99fdd49bad208a6a555090ce882d2653e.diff, and it looks fine, you just added context.

2.	28065814ab889552aebe99cd41b35bc5e31063fe is you merging this PR #561.

	As far as I can see, the commit 28065814ab889552aebe99cd41b35bc5e31063fe contains all of the changes from the 3 commits in this PR, squashed into one commit. However, its commit message only contains text from the first commit and drops everything from 2nd (which was meaningful, not just code review edits). It would've been nice to include that, but it's not very critical. The main thing is that all changes are in.

	You can confirm that by checking that https://patch-diff.githubusercontent.com/raw/google/go-github/pull/561.diff and https://github.com/google/go-github/commit/28065814ab889552aebe99cd41b35bc5e31063fe.diff are identical. They are.

	```
	$ $ diff <(curl -sS 'https://patch-diff.githubusercontent.com/raw/google/go-github/pull/561.diff') <(curl -sS 'https://github.com/google/go-github/commit/28065814ab889552aebe99cd41b35bc5e31063fe.diff')
	$ 
	```

So this PR is fully merged in 28065814ab889552aebe99cd41b35bc5e31063fe, and we can close it. You should resolve your local master branch issues, however.

Also, thanks for showing me that `-v` option for `git branch` command displays behind/ahead counts. I made [`git-branches`](https://github.com/shurcooL/git-branches) partially because I didn't know that.

shurcooL:
Oh, I see what you were talking about regarding "Closed with unmerged commits". It only shows up when the PR is closed:

image

I think it's fine. I'm not completely sure why/how GitHub displays it, but I think it's because you merged the commit by squashing these 3 commits into one and closed it via commit message (the "Closes #561." text).

gmlewis:
Thank you, @shurcooL! I appreciate the analysis and confirmation.
I was able to get my master branch back to normal (by a `reset --hard` to an older commit and then syncing to the latest), fortunately.

PR #565, opened 2017-02-28
UsersService: Method to get named users GPG keys.

thorduri:


github/users_gpg_keys.go
```diff
@@ -40,6 +40,28 @@ type GPGEmail struct {
 	Verified *bool   `json:"verified,omitempty"`
 }
 
+// GetGPGKeys fetches GPG keys for a user by name.
+//
+//GitHub API docs: https://developer.github.com/v3/users/gpg_keys/#list-gpg-keys-for-a-user
```
shurcooL: Minor, don't forget a space between the slashes and the comment text.

Rationale: https://dmitri.shuralyov.com/idiomatic-go#comments-for-humans-always-have-a-single-space-after-the-slashes.

github/users_gpg_keys.go
```diff
@@ -40,6 +40,28 @@ type GPGEmail struct {
...
+func (s *UsersService) GetGPGKeys(ctx context.Context, user string) ([]*GPGKey, *Response, error) {
```
shurcooL: A bigger issue I see is with the method name. It's called "Get", yet it's really list. It returns a slice of GPG keys, not a single GPG key.

I see we already have `ListGPGKeys(ctx context.Context) ([]*GPGKey, *Response, error)`, which is probably why you opted to use a different name.

I think we should actually merge both these endpoints into one, by adding a "username string" parameter and documenting it that using an empty value lists GPG keys for current user, otherwise for the specified user.

We already do this in some other places, for example:

https://godoc.org/github.com/google/go-github/github#UsersService.ListKeys

> ListKeys lists the verified public keys for a user. Passing the empty string will fetch keys for the authenticated user.

This is a preview API endpoint, so it's more okay to update it.

Another option is to rename this method to something like `ListGPGKeysForUser` but it doesn't seem to me that it would be better.

/cc @gmlewis for your opinion.

gmlewis: I agree, @shurcooL - let's double-up on `ListGPGKeys` and add a `username string` parameter to it with documentation for what happens when it's empty.

thorduri: @shurcooL That was the rationale for the naming choice.

I'll post an updated PR doubling-up on ListGPGKeys and adding a username string.

Thanks for the review.

thorduri:
@shurcooL  I've updated the PR with the proposed changes.

PR #567, opened 2017-03-01
Prefer setting over adding, where setting is sufficient and better fits intention.

shurcooL:
The common theme of this cleanup PR is to prefer setting things, instead of adding, in places where setting is sufficient (i.e., the current adding behavior is provably never done more than once for the same key), and it's a better fit for the intended behavior.

The motivation to do so is because set communicates the underlying intended action better, and it's simpler to reason about, more readable. Readers can see that there will not be multiple entries of the same key without having to verify that by reading additional code. It also fixes inconsistencies in the code in some places.

The PR consists of 4 commits, see commit message of each commit for details and rationale of each distinct change.

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -403,9 +403,9 @@ func TestDo_rateLimit(t *testing.T) {
 	defer teardown()
 
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Add(headerRateLimit, "60")
-		w.Header().Add(headerRateRemaining, "59")
-		w.Header().Add(headerRateReset, "1372700873")
+		w.Header().Set(headerRateLimit, "60")
+		w.Header().Set(headerRateRemaining, "59")
+		w.Header().Set(headerRateReset, "1372700873")
 	})
 
 	req, _ := client.NewRequest("GET", "/", nil)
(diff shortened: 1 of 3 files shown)


shurcooL:
> it's simpler to reason about, more readable. Readers can see that there will not be multiple entries of the same key without having to verify that by reading additional code. It also fixes inconsistencies in the code in some places.

To expand on what I mean by that, here's an example.

When reading Go code and seeing the following line only:

```Go
// ...
query.Set("foo", "bar")
// Without having to read more code above, I can be confident that at this
// point, query will contain exactly one entry with the key "foo", and its value
// will be equal to "bar".
```

On the other hand:

```Go
// ...
query.Add("foo", "bar")
// Without reading more code above, I cannot know exactly how many entries
// with the key "foo" there will be in query. I know there will be at least one with
// value "bar", but there may also be additional entries with key "foo" that
// were set in/added to query previously.
```

Of course, if my intention is to add an extra instance of "foo", regardless of previous potential such entries, then using `Add` is the right thing to do. But if the goal is to set the one and only "foo" entry, then using `Set` should be preferable.

shurcooL:
Friendly ping.

shurcooL:
@gmlewis, I've merged this PR via "Rebase and merge" strategy, because the 4 commits are logical commits with descriptive commit messages, and I wanted to preserve that. They are independent changes, so squashing them into 1 commit is counter-productive. This is not a case of a single logical change with "amend" commits on top.

PR #572, opened 2017-03-03
Return rate limit information immediately after `checkRateLimitBeforeDo`

### LINKED ISSUE #571: With removal of deprecated client.Rate in #555, rate limit info needs more visibility
It used to be that clients might make API calls like:

```go
	repos, resp, err := client.Repositories.ListByOrg(org, opt)
	if err != nil {
		rate := client.Rate()
		if rate.Remaining == 0 {
			//...
		}
		//...
	}
```

but after #555, it turns out that `checkRateLimitBeforeDo` will return an error without rate information.
I have a PR on the way that addresses this issue.

gmlewis:
Fixes #571.

Change-Id: Iefcc687d9fc56c40964d5e6f1cdbc54d020dae3c

### DIFF
--- a/github/doc.go
+++ b/github/doc.go
@@ -71,8 +71,10 @@ limited to 60 requests per hour, while authenticated clients can make up to
 that are not issued on behalf of a user, use the
 UnauthenticatedRateLimitedTransport.
 
-You can always call RateLimits() directly to get the most up-to-date
-rate limit data for the client.
+The returned Response.Rate value contains the rate limit information
+from the most recent API call. If a recent enough response isn't
+available, you can use RateLimits to fetch the most up-to-date rate
+limit data for the client.
 
 To detect an API rate limit error, you can check if its type is *github.RateLimitError:
 
--- a/github/doc.go
+++ b/github/doc.go
@@ -71,11 +71,8 @@ limited to 60 requests per hour, while authenticated clients can make up to
 that are not issued on behalf of a user, use the
 UnauthenticatedRateLimitedTransport.
 
-The Rate method on a client returns the rate limit information based on the most
-recent API call. This is updated on every call, but may be out of date if it's
-been some time since the last API call and other clients have made subsequent
-requests since then. You can always call RateLimits() directly to get the most
-up-to-date rate limit data for the client.
+You can always call RateLimits() directly to get the most up-to-date
+rate limit data for the client.
 
 To detect an API rate limit error, you can check if its type is *github.RateLimitError:
 
--- a/github/github.go
+++ b/github/github.go
@@ -396,7 +396,10 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 	// If we've hit rate limit, don't make further requests before Reset time.
 	if err := c.checkRateLimitBeforeDo(req, rateLimitCategory); err != nil {
-		return nil, err
+		return &Response{
+			Response: err.Response,
+			Rate:     err.Rate,
+		}, err
 	}
 
 	resp, err := c.client.Do(req)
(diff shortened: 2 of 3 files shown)


github/doc.go
```diff
@@ -71,11 +71,8 @@ limited to 60 requests per hour, while authenticated clients can make up to
 that are not issued on behalf of a user, use the
 UnauthenticatedRateLimitedTransport.
 
-The Rate method on a client returns the rate limit information based on the most
-recent API call. This is updated on every call, but may be out of date if it's
-been some time since the last API call and other clients have made subsequent
-requests since then. You can always call RateLimits() directly to get the most
-up-to-date rate limit data for the client.
+You can always call RateLimits() directly to get the most up-to-date
+rate limit data for the client.
```
shurcooL: In addition to removing the outdated part about `Rate` method on github client, I think we should put in its place a mention of the best way to access the remaining rate limit, and that is by accessing the `resp.Rate` of the previous call.

That [used to be](https://github.com/google/go-github/commit/2a4b92075e5c5e42ee0e55819b70f49b02c1ac64#diff-9bd6c2575adb220521a096df92e2d502L390) already written as part of the removed `Rate` method:

> Use the Response.Rate returned from most recent API call.

So, how about something like this:

> The returned Response.Rate value contains the rate limit information from the most recent API call. You can use RateLimits to get the most up-to-date rate limit data for the client.

github/github.go
```diff
@@ -396,7 +396,10 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 	// If we've hit rate limit, don't make further requests before Reset time.
 	if err := c.checkRateLimitBeforeDo(req, rateLimitCategory); err != nil {
-		return nil, err
+		return &Response{
+			Response: err.Response,
```
shurcooL: I have one potential concern about this change.

I think it's somewhat unfortunate that we ended up being forced to create a fake/mock `*http.Response` even in situations where no remote HTTP request was actually made. I am referring to [this](https://github.com/google/go-github/blob/9be6ce168fa69a3cc2db27a776950a42d18c2ae0/github/github.go#L465-L472).

It was originally done in #347 because we (or I) didn't want to break clients that assumed `*http.Response` was never `nil`, since it used to never be `nil` in prior to that change.

So I think it's somewhat disadvantageous that we would now be propagating the fake `*http.Response` in more situations.

However, even if we had control over this, I'm not sure what's better. If a remote API request is not made (because we've predicted it'll fail due to exceeded rate limit), should the client be able to tell it apart from when a remote API request was made? It wasn't documented in the past, but one could've done that by checking if `resp` was nil. This changes that.

I'm pointing out the above because I wanted to think it through. But I don't think it's a reason to avoid going with this change. This change actually increases consistency (in that we return `resp` and `resp.Rate` in more cases).

So, I think I'm okay with doing this, but I want to take another day to think it over. Meanwhile, I'd love to hear your perspective on the above @gmlewis.

github/doc.go
gmlewis: I like that.  I'll hopefully make these changes early next week.

github/github.go
gmlewis: You make excellent points and I have mixed feelings too. Please know that I am completely comfortable with the decision to simply revert the code changes and keep the updates to the documentation.

There is no urgency here so we can let this sit for some time in case others have opinions.

shurcooL:
I've thought about this and I'm in favor of making this change.

Given we already do have a fake response, I think this change simplifies things/adds consistency by returning the fake response in more places, just like a real response.

The subtle difference I outlined in https://github.com/google/go-github/pull/572#discussion_r104279097, where a really determined individual could tell apart a fake response (no network call made) vs real response (from a network call that was made), is not documented, and cannot really be relied on.

So, I'm okay with doing this because it simplifies/streamlines things. I don't think it's going to make a practical difference for most people, given what we've discussed in https://github.com/google/go-github/issues/571#issuecomment-284106437. But I wanted to make a decision on this PR to move forward, and I think merging it makes more sense than not.

However, to merge this, @gmlewis you still need to address https://github.com/google/go-github/pull/572/files#r104281339.

gmlewis:
Thanks, @shurcooL.  I believe I have addressed the review feedback, but please let me know if I missed something.

shurcooL:
The change to `github/doc.go` looks good, please apply it to `README.md` too, to keep them in sync.

gmlewis:
Ah, thank you, @shurcooL!
Done.  PTAL.

shurcooL:
Friendly ping @gmlewis.

Now that this has my approval, I'm leave this for you to merge, since you authored this PR.

PR #573, opened 2017-03-04
Add issue.Locked field

### LINKED ISSUE #570: support issue.locked
The v3 API returns a `locked` boolean indicating whether an issue has been locked, at least according to https://developer.github.com/v3/issues/#get-a-single-issue, but I don't see it in the Issue struct.

kevinburke:
Fixes #570.

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -27,6 +27,7 @@ type Issue struct {
 	ID               *int              `json:"id,omitempty"`
 	Number           *int              `json:"number,omitempty"`
 	State            *string           `json:"state,omitempty"`
+	Locked           *bool             `json:"locked,omitempty"`
 	Title            *string           `json:"title,omitempty"`
 	Body             *string           `json:"body,omitempty"`
 	User             *User             `json:"user,omitempty"`


github/issues.go
```diff
@@ -47,6 +47,7 @@ type Issue struct {
 	// TextMatches is only populated from search results that request text matches
 	// See: search.go and https://developer.github.com/v3/search/#text-match-metadata
 	TextMatches []TextMatch `json:"text_matches,omitempty"`
+	Locked      *bool       `json:"locked,omitempty"`
```
shurcooL: Please move it up, together with all the other core fields.

This makes it look like its grouped with `TextMatches`, but it isn't.

You can place it after `Assignees`, or somewhere else if you think it's more relevant/related to another existing field.

Perhaps under `State` is a good fit? Or above `Comments`, since that's where it shows up in examples on GitHub API docs.

kevinburke: I thought there was some compatibility related reason to keep adding on to the end instead of in the middle, but I'll switch it up

shurcooL: Nope, this is Go. :)

shurcooL:
@gmlewis I've just realized we forgot to go generate and add the new method to access this.

I merged this because I thought it was a trivial change, similar to ones in the past. But I forgot about this change since then.

It's not a big deal because it'll be included in the next PR that adds a field and regenerates, but it's something I should be mindful of next time.

@gmlewis Given how simple our go generate task is, it should be possible to run it inside Travis and complain if it finds a diff. That way we'll have an easier time catching this in the future. What do you think about that?

gmlewis:
Yes, I totally agree... I often find myself forgetting to run that too.  Now I have a `test.sh` script that I run instead of `go test ./...` that does everything for me so I don't forget. 😄 

Hey, I think putting that in Travis is a great idea!
Feel free to do that or file an issue and I can address it soon hopefully.

PR #575, opened 2017-03-05
Add GetLocked() accessor

kevinburke:


shurcooL:
Thanks for following up @kevinburke!

PR #577, opened 2017-03-05
add issue.ClosedBy field and add issue test into tests/field

### LINKED ISSUE #574: How do I access issue/closed_by attribute?
API descripition here: https://developer.github.com/v3/issues/#get-a-single-issue

I want to use the `closed_by` attribute in `Get a single issue` which looks like this

```
  "closed_by": {
    "login": "octocat",
    "id": 1,
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    "gravatar_id": "",
    "url": "https://api.github.com/users/octocat",
    "html_url": "https://github.com/octocat",
    .....
  }
```

How can I access it?

Pezhin:
Fixes #574 and also add issue into tests/field

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -37,6 +37,7 @@ type Issue struct {
 	ClosedAt         *time.Time        `json:"closed_at,omitempty"`
 	CreatedAt        *time.Time        `json:"created_at,omitempty"`
 	UpdatedAt        *time.Time        `json:"updated_at,omitempty"`
+	ClosedBy         *User             `json:"closed_by,omitempty"`
 	URL              *string           `json:"url,omitempty"`
 	HTMLURL          *string           `json:"html_url,omitempty"`
 	Milestone        *Milestone        `json:"milestone,omitempty"`
--- a/tests/fields/fields.go
+++ b/tests/fields/fields.go
@@ -64,6 +64,7 @@ func main() {
 		{"users/willnorris/keys", &[]github.Key{}},
 		{"orgs/google-test", &github.Organization{}},
 		{"repos/google/go-github", &github.Repository{}},
+		{"repos/google/go-github/issues/1", &github.Issue{}},
 		{"/gists/9257657", &github.Gist{}},
 	} {
 		err := testType(tt.url, tt.typ)


Pezhin:
I signed it!

tests/fields/fields.go
```diff
@@ -65,6 +65,7 @@ func main() {
 		{"orgs/google-test", &github.Organization{}},
 		{"repos/google/go-github", &github.Repository{}},
 		{"/gists/9257657", &github.Gist{}},
+		{"repos/google/go-github/issues/1", &github.Issue{}},
```
shurcooL: You can move it up by one line. That way it'll be grouped with the similar `{"repos/google/go-github", &github.Repository{}},` entry.

github/issues.go
```diff
@@ -44,6 +44,7 @@ type Issue struct {
 	Repository       *Repository       `json:"repository,omitempty"`
 	Reactions        *Reactions        `json:"reactions,omitempty"`
 	Assignees        []*User           `json:"assignees,omitempty"`
+	ClosedBy         *User             `json:"closed_by,omitempty"`
```
shurcooL: What do you think about moving this up, and placing it under `ClosedAt` field?

shurcooL: Alternatively, you can place it under `UpdatedAt`, so that it's consistent with the example GitHub API docs use:

image

Pezhin: I will the reordering soon.

BTW, I am relatively new to the `generate` stuff but it seems to be ignoring the field with `User` type? After calling generate for adding a boolean field(like locked), I saw a new getter was generated but there was no change after adding a `User` field. Is it supposed to be so?

shurcooL: There should be a single new method called `GetClosedBy()`, is that what you saw? It should be similar to #575, that one was `GetLocked()`.

shurcooL: You might need to either merge the latest master branch into your feature branch, or rebase it on top of latest master.

Pezhin: @shurcooL I mean..on my side, the generator is ignoring my change with adding `closed_by`...

And inside the `github-accessors` file, there's no getter that returns type of either `User` or `Repository`... It seems to me that the genearator is ignoring all user-defined types...?

shurcooL: You are right. In that case, nothing more needs to be done for this PR.

@gmlewis Just checking, but is it intentional that `*User` fields don't get accessors? What's the reason?

Pezhin: @shurcooL Might be a bug xD... potentially an issue is on the way, xD

gmlewis: Yes, that is intentional - I didn't give access to named structs... I figured you could just use the accessors provided for that named struct, like:

`name := resp.User.GetName()`

(totally hypothetical - not looking at the code at the moment).  If you know that `resp` is not `nil`, then even if `User` is `nil` this should still work because `GetName` (or whatever field of the `User` struct you are interested in getting) can take `nil` as the receiver... if I'm making any sense.

I'm happy to change that if you think it would be useful, but I couldn't think of a scenario that would be useful.  (Note that I have not read this entire thread yet... and am just dropping by.)
Does that help any?  If not, I can investigate further.

gmlewis: (deleted comment... looking at `type User struct`...)

Pezhin: @gmlewis Yes, thanks! Since it is intended, I think it is ok to merge this PR

shurcooL: > (Note that named structs themselves don't get accessors [intentionally]... there is no `GetUser` method, for example, because you can just do something like `resp.User.GetName()` for example and it will just work... even if `resp.User == nil`... does that answer the question?)

I see. That answers my question, thanks.

shurcooL:
@Pezhin All that's left is for you to regenerate this package after your changes, as I mentioned above:

> You will also need to regenerate the github package after making this change. See step 5 of https://github.com/google/go-github/blob/master/CONTRIBUTING.md#submitting-a-patch.

Then we can merge this.

Pezhin:
@shurcooL: as we discussed in the changes, there's nothing I can generate because I just add a User field where User field has nothing to generate.

shurcooL:
Oh, you're right, I forgot about that. Thanks.

shurcooL:
@gmlewis, does this have your LGTM too?

gmlewis:
LGTM.

shurcooL:
Ok, I'll merge.

PR #579, opened 2017-03-05
Travis: Check that go generate produces zero diff in Travis CI.

shurcooL:
~~Testing this out and iterating on it.~~ Ready to review and merge now.

This should help us catch situations where we forget to run `go generate` after a change.

This idea comes from https://github.com/google/go-github/pull/573#issuecomment-284206455. /cc @gmlewis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,6 @@ install:
 script:
   - go get -t -v ./...
   - diff -u <(echo -n) <(gofmt -d -s .)
+  - go generate -x ./... && git diff --exit-code
   - go tool vet .
   - go test -v -race ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ install:
 script:
   - go get -t -v ./...
   - diff -u <(echo -n) <(gofmt -d -s .)
-  - go generate -x ./... && git diff --exit-code
+  - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; exit $code
+  - git diff --exit-code
   - go tool vet .
   - go test -v -race ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,6 @@ install:
 script:
   - go get -t -v ./...
   - diff -u <(echo -n) <(gofmt -d -s .)
-  - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; (exit $code)
+  - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; (exit $code) # Check that go generate ./... produces a zero diff; clean up any changes afterwards.
   - go tool vet .
   - go test -v -race ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
 script:
   - go get -t -v ./...
   - diff -u <(echo -n) <(gofmt -d -s .)
-  - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; exit $code
+  - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; (exit $code)
   - git diff --exit-code
   - go tool vet .
   - go test -v -race ./...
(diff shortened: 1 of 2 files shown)


shurcooL:
@gmlewis Check it out. How's this look? https://travis-ci.org/google/go-github/jobs/207836057

gmlewis:
This is fantastic, @shurcooL!
I'm not sure why I can't see the results for Go 1.8: https://travis-ci.org/google/go-github/jobs/207836057
but otherwise it looks great!

shurcooL:
> I'm not sure why I can't see the results for Go 1.8

I'm not sure either.

I've restarted that particular job, and a build log seems to be showing up okay now. I guess it was some Travis CI issue.

https://travis-ci.org/google/go-github/jobs/207836057

shurcooL:
A couple of observations and questions.

1. Doing `git diff --exit-code` will not pick up new files. If the `go generate` step were to create new files, they would be untracked, and `git diff` ignores those until you `git add .` them.

    Currently, we know that `go generate` only updates one existing file (github-accessors.go), so this isn't a big deal. But in a general case, this is a limitation, so if there was a nice solution that doesn't add much verbosity, that'd be great.

    The only workaround I can currently think of is to do `go generate -x ./... && git add . && git diff --cached --exit-code`. But that's more verbose, and probably harder to understand. I don't think it's worth it because of that.

    If there are any better ideas, I'd love to hear them.

2. What about the order? I've currently placed it after `diff -u <(echo -n) <(gofmt -d -s .)` and before `go vet` and `go test`.

    In a situation where go generate happens to create a non-zero difference, do we want `go vet` and `go test` to run on that diff included?

    Should `gofmt -d -s` also run after the `go generate` check?

    It doesn't make a huge difference, but still, I'd rather think about the optimal placement before we merge this. Any thoughts or preferences?

gmlewis:
For part 1, extra verbosity doesn't really bother me in the travis config file... a comment could explain that it is being done to catch any differences made by `go generate`... I can't think of a better way at the moment... and that seems like a good solution to me despite the verbosity.

For part 2, my initial gut reaction is that it will be more likely that someone forgets to run `go generate` than to forget to run `gofmt` or `goimports` (especially with the editor integration of these latter tools these days)... so I'm thinking that maybe the `go generate` test could easily be before the existing `diff`... although when you get right down to it, I don't think the order of these two really matters. In fact, I'm feeling more confident that the order doesn't matter because these operations really could be done in parallel and if either of them failed, the build should fail. (But there's really no need to run them in parallel, I think.)

shurcooL:
All right. I agree, it's better to think of each step as an independent check, as if they could run in parallel. It's better to have the other checks display the actual results of running on the code as-submitted, not as-submitted-with-go-generate-run afterwards. That could be confusing, if it causes the local test results to not match remote.

So, I think I'll add the `git checkout -- .` cleanup step to make it a truly orthogonal check.

As for ordering... Agreed, doesn't really matter. Another minor factor is that `go test` creates lots of output, so stuff at the bottom of a build log is harder to see. I guess I'll just keep it where it is.

shurcooL:
@gmlewis I've implemented what I described above. PTAL.

gmlewis:
Very nice!  I like it a lot!  I also learned about exiting a subshell.  :smile:
Now, maybe just a comment for that tricky line would be great.
Alternatively, you could make an alias for that last expression and comment the alias instead...

```bash
# git-diff-then-revert checks if any changes were made to the code then reverts those changes.
# The exit code of git-diff-then-revert is the result of the `git diff`.
alias git-diff-then-revert='git diff --exit-code; code=$?; git checkout -- .; (exit $code)'
go generate -x ./... && git-diff-then-revert
```

Totally your call... feel free to ignore this.

shurcooL:
Sure, I'll add a comment that describes what happens in the line to the right. It's not visually disruptive, and someone looking for more details of that line will find it.

I'm also going to remove the temporary testing change, because I think this is done and ready to merge.

gmlewis:
LGTM.

PR #580, opened 2017-03-06
Retains ability to add users directly as collaborators

alindeman:
GitHub is previewing a media type that [sends invitations instead of directly adding collaborators](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator).

Currently `go-github` always sends the preview header, but I think we should retain the ability to add a collaborator directly because this header is a fundamental behavior change--rather than a typical preview which enables new functionality or sends back additional information.

Concretely, the change to always use the preview header is affecting my ability to upgrade `go-github` in the [terraform](https://github.com/hashicorp/terraform) project without changing the behavior of the `github_repository_collaborator` resource. In the version that's currently vendored, the preview header is not sent, meaning the user is added directly. If I upgrade go-github to get access to new (unrelated) functionality, I fundamentally change how an unrelated resource works.

I think eventually programs like terraform will need to switch to using invitations, but I propose the old functionality continue to be exposed in `go-github` since, in fact, the GitHub API itself still supports it.

### DIFF
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -94,7 +94,7 @@ type RepositoryAddCollaboratorOptions struct {
 	Permission string `json:"permission,omitempty"`
 }
 
-// AddCollaborator adds the specified GitHub user as collaborator to the given repo.
+// AddCollaborator invites the specified GitHub user as a collaborator to the given repo.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator
 func (s *RepositoriesService) AddCollaborator(ctx context.Context, owner, repo, user string, opt *RepositoryAddCollaboratorOptions) (*Response, error) {
@@ -106,6 +106,19 @@ func (s *RepositoriesService) AddCollaborator(ctx context.Context, owner, repo, 
 
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeRepositoryInvitationsPreview)
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// AddCollaboratorDirectly adds the specified GitHub user as a collaborator to the given repo.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator
+func (s *RepositoriesService) AddCollaboratorDirectly(ctx context.Context, owner, repo, user string, opt *RepositoryAddCollaboratorOptions) (*Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/collaborators/%v", owner, repo, user)
+	req, err := s.client.NewRequest("PUT", u, opt)
+	if err != nil {
+		return nil, err
+	}
 
 	return s.client.Do(ctx, req, nil)
 }
--- a/github/repos_collaborators_test.go
+++ b/github/repos_collaborators_test.go
@@ -141,6 +141,31 @@ func TestRepositoriesService_AddCollaborator_invalidUser(t *testing.T) {
 	testURLParseError(t, err)
 }
 
+func TestRepositoriesService_AddCollaboratorDirectly(t *testing.T) {
+	setup()
+	defer teardown()
+
+	opt := &RepositoryAddCollaboratorOptions{Permission: "admin"}
+
+	mux.HandleFunc("/repos/o/r/collaborators/u", func(w http.ResponseWriter, r *http.Request) {
+		v := new(RepositoryAddCollaboratorOptions)
+		json.NewDecoder(r.Body).Decode(v)
+
+		testMethod(t, r, "PUT")
+		testHeader(t, r, "Accept", mediaTypeV3)
+		if !reflect.DeepEqual(v, opt) {
+			t.Errorf("Request body = %+v, want %+v", v, opt)
+		}
+
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	_, err := client.Repositories.AddCollaboratorDirectly(context.Background(), "o", "r", "u", opt)
+	if err != nil {
+		t.Errorf("Repositories.AddCollaborator returned error: %v", err)
+	}
+}
+
 func TestRepositoriesService_RemoveCollaborator(t *testing.T) {
 	setup()
 	defer teardown()


github/repos_collaborators.go
```diff
@@ -94,7 +94,7 @@ type RepositoryAddCollaboratorOptions struct {
 	Permission string `json:"permission,omitempty"`
 }
 
-// AddCollaborator adds the specified GitHub user as collaborator to the given repo.
+// AddCollaborator invites the specified GitHub user as a collaborator to the given repo.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator
 func (s *RepositoriesService) AddCollaborator(ctx context.Context, owner, repo, user string, opt *RepositoryAddCollaboratorOptions) (*Response, error) {
```
alindeman: Another thought I had: rename this to `InviteCollaborator` and revert `AddCollaborator` to the 'add directly' behavior. But this is yet another behavior change in `go-github` which might be confusing to clients. I'm open to feedback though.

shurcooL:
I believe this is an exact duplicate of PR #470 which we came to the conclusion it's better not to merge.

@alindeman, can you take a look and see what you think of our discussion there? How do you feel about it?

alindeman:
@shurcooL I still think this library removed the functionality a bit too early: the functionality to add users directly still exists and hasn't been officially called deprecated yet as far as I know. At the end of the day, though, this isn't super important to me: I'll defer to the library maintainers and work around it in my own code if needed.

shurcooL:
> I still think this library removed the functionality a bit too early: the functionality to add users directly still exists and hasn't been officially called deprecated yet as far as I know.

Fair enough.

How about we take the following approach then. Given that all evidence (and please correct me if I'm wrong about this) suggests that the ability to use API to invite forcibly (i.e., skipping asking for consent via invitation) will soon/eventually go away, but hasn't yet, let's add that ability but optimize towards being able to remove it in the future, if/once GitHub removes such functionality from the API.

What I mean by that is, let's add separate methods that re-add this "invite forcibly" functionality, instead of changing existing ones that send invitations. Then these methods can be deprecated or removed more easily, if/when that needs to happen.

Based on a quick look over the PR, it looks to me that it's already the approach you're currently taken. I'll look at the implementation details sometime tomorrow in more detail, I just wanted to share my higher level thoughts sooner.

github/repos_collaborators.go
shurcooL: I definitely don't think this is a good idea, for 2 reasons. We generally try to prioritize having the method names match what the GitHub API documentation calls the endpoints. In this case:

1. It would be yet a breaking API change, one which is not neccessary, and without strong benefits.
2. https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator calls it "Add user as a collaborator".

alindeman: Sounds good :+1:

alindeman:
> What I mean by that is, let's add separate methods that re-add this "invite forcibly" functionality, instead of changing existing ones that send invitations. Then these methods can be deprecated or removed more easily, if/when that needs to happen.

I agree, and that is the approach I took. I'm looking forward to your more in-depth feedback.

github/repos_collaborators.go
```diff
@@ -110,6 +110,19 @@ func (s *RepositoriesService) AddCollaborator(ctx context.Context, owner, repo,
 	return s.client.Do(ctx, req, nil)
 }
 
+// AddCollaboratorDirectly adds the specified GitHub user as a collaborator to the given repo.
+//
```
gmlewis: How about adding "without sending an invite"?
Also, maybe note that this will most likely be going away soon... or maybe even mark it "Deprecated" here on creation.

github/repos_collaborators.go
```diff
@@ -94,7 +94,7 @@ type RepositoryAddCollaboratorOptions struct {
 	Permission string `json:"permission,omitempty"`
 }
 
-// AddCollaborator adds the specified GitHub user as collaborator to the given repo.
+// AddCollaborator invites the specified GitHub user as a collaborator to the given repo.
```
shurcooL: This is a good improvement, thanks. Let's make it even more explicit about what this does. This is based on wording in [GitHub API docs](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator):

> To send an invitation to a collaborator rather than directly adding them,

So, how about:

```
// AddCollaborator sends an invitation to the specified GitHub user to become a collaborator to the given repo.
```

github/repos_collaborators.go
shurcooL: I think it's okay to add a note saying that all signs indicate that this might be going away, but we shouldn't actually deprecate it. As of this moment, this is actually a part of the stable GitHub API, and sending invitations is a part of a preview API.

But it'd be good to set correct expectations and help users understand why both methods are available So, I agree that adding a note that AddCollaboratorDirectly does so rather than sending an invitation would be nice. Perhaps put it in parenthesis?

github/repos_collaborators_test.go
```diff
@@ -141,6 +141,31 @@ func TestRepositoriesService_AddCollaborator_invalidUser(t *testing.T) {
...
+		if !reflect.DeepEqual(v, opt) {
+			t.Errorf("Request body = %+v, want %+v", v, opt)
+		}
+
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	_, err := client.Repositories.AddCollaboratorDirectly(context.Background(), "o", "r", "u", opt)
+	if err != nil {
+		t.Errorf("Repositories.AddCollaborator returned error: %v", err)
```
shurcooL: AddCollaboratorDirectly, not AddCollaborator.

shurcooL:
An important update on this subject.

https://developer.github.com/changes/2017-05-05-end-repository-invitation-preview/

> We're making the Repository Invitation API part of the official GitHub API on June 26, 2017. This will be a breaking change to the API and the API endpoint for directly adding a collaborator to a repository. **You will no longer be able to directly add a user to a repository.** Instead, the user will receive an invitation, which they can accept or decline via email, notification, or API endpoint.

If I understand the situation correctly, it means there's no point in merging this PR, since it is now confirmed to become non-functional in the near future.

Those who still want to take advantage of the ability to invite directly, instead of sending an invite, can vendor an older version of go-github, a fork, or custom code (but it'll stop working on June 26 anyway). Does that sound reasonable?

alindeman:
> Those who still want to take advantage of the ability to invite directly, instead of sending an invite, can vendor an older version of go-github, a fork, or custom code (but it'll stop working on June 26 anyway). Does that sound reasonable?

Agreed.

PR #581, opened 2017-03-06
Fail the build if "go generate" fails

kevinburke:
I wrote a little binary that will run a command and error if any files
tracked by Git are modified: github.com/kevinburke/differ.

I used it here to check whether "go generate" generates any changes,
and also to simplify the gofmt command invocation. I tested that they
worked by forking from the commit that didn't run "go generate."

I also updated the .travis.yml file with a go_import_path so
forks work correctly - Travis can run into problems building e.g.
github.com/kevinburke/go-github since the import paths for subprojects
all mention github.com/google/go-github.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: go
+go_import_path: github.com/google/go-github
 go:
   - 1.7.x
   - 1.8.x
@@ -8,10 +9,15 @@ matrix:
   allow_failures:
     - go: master
   fast_finish: true
+before_script:
+    - mkdir -p $HOME/bin
+    - curl --silent --location https://github.com/kevinburke/differ/releases/download/0.4/differ-linux-amd64 > $HOME/bin/differ
+    - chmod 755 $HOME/bin/differ
 install:
   - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
 script:
   - go get -t -v ./...
-  - diff -u <(echo -n) <(gofmt -d -s .)
+  - differ gofmt -s .
+  - differ go generate ./...
   - go tool vet .
   - go test -v -race ./...


gmlewis:
@kevinburke - please take a look at #579... I think that may be the direction we want to take.
Please let @shurcooL and I know if that is missing any functionality, but I don't think it is.

kevinburke:
OK; I think you should at least consider this, since my approach gives better error messages and it's clearer to a) read the command, b) figure out what the problem is, and c) apply the same technique to more problems in the future, without complex shell hackery, but whatever you like is fine with me.

gmlewis:
I see one significant problem with this approach.
Using curl to retrieve a binary linux executable is a security risk and is not transparent.
I much prefer using the commented shell command that anyone can easily understand than a third-party binary executable that I have never seen before and have not looked at its source code.

kevinburke:
Sure, though I'd also point out that I've contributed to this library and about 30 changesets to the Go standard library and associated tools, the build script calls `go get`, the build script installs `libc-bin libc-dev-bin libc6 libc6-dev` C libraries via apt-get, runs a binary called `gimme` to install Go, and point you to Reflections on Trusting Trust, etc.

The source code is here, it's about 40 lines, you could just as easily install it with `go get`. https://github.com/kevinburke/differ/blob/master/main.go

gmlewis:
OK, but just for the record, I was not saying that *you* are a security risk, I was saying that curling a binary executable from the internet is, in general, a security risk. No offense was intended and I apologize for not being clear.

And while I'm here, I would like to thank you for your contributions to Go and all the other great things you have worked on. Your work is greatly appreciated.

Yes, I would prefer a `go get` solution, but before you change anything, let's hear what @willnorris and/or @shurcooL think about this approach.

willnorris:
Travis is not a trusted secure environment, so running arbitrary binaries there is not a huge deal.  That said, a go gettable approach would be much preferable.  differ certainly looks neat, but I can't say that I have strong opinions on this vs #579

shurcooL:
Thanks @kevinburke for wanting to contribute.

I'd like to get #579 in first, it's very simple (1 line, no additional dependencies), functional, and I think it makes a good starting point. We can consider a version of this PR rebased on top as a potential enhancement and discuss that.

shurcooL:
> approach gives better error messages

Can you provide an example? I don't know what the error message is, so I can't compare.

For reference, this is the error message that the current version of #579 results in when `go generate ./...` produces a non-zero diff:

image

(Source: https://travis-ci.org/google/go-github/jobs/208035644.)

PR #582, opened 2017-03-06
Support Google App Engine with recent context changes

### LINKED ISSUE #578: Using go-github on Google App Engine with recent context changes - any interest?
After importing the recent context changes in #529 into Google, I became aware that it is a bit trickier now to use this same code base either within an App Engine project or outside of one.

I had to pull a few tricks with `// +build appengine` and `// +build !appengine` in a couple new files and patched a couple other files to make it all work.

I'm not sure if anyone is really interested in these changes, but if you are, please let me know and I'll see if I can replicate the needed changes in a new PR and incorporate them into the library.

gmlewis:
Fixes #578.

Change-Id: Icc6b26877bcbc5e34845238c998d50a873fb7d65

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -390,7 +390,7 @@ func parseRate(r *http.Response) Rate {
 // The provided ctx must be non-nil. If it is canceled or times out,
 // ctx.Err() will be returned.
 func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
-	req = req.WithContext(ctx)
+	ctx, req = addContext(ctx, req)
 
 	rateLimitCategory := category(req.URL.Path)
 
--- a/github/with_appengine.go
+++ b/github/with_appengine.go
@@ -1,0 +1,21 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This file provides glue for making github work on App Engine.
+
+// +build appengine
+
+package github
+
+import (
+	"context"
+	"net/http"
+
+	"google.golang.org/appengine"
+)
+
+func addContext(ctx context.Context, req *http.Request) (context.Context, *http.Request) {
+	return appengine.WithContext(ctx, req), req
+}
--- a/github/without_appengine.go
+++ b/github/without_appengine.go
@@ -1,0 +1,19 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This file provides glue for making github work without App Engine.
+
+// +build !appengine
+
+package github
+
+import (
+	"context"
+	"net/http"
+)
+
+func addContext(ctx context.Context, req *http.Request) (context.Context, *http.Request) {
+	return ctx, req.WithContext(ctx)
+}
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -248,10 +248,11 @@ func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo st
 	}
 	var resp *http.Response
 	// Use http.DefaultTransport if no custom Transport is configured
+	ctx, req = addContext(ctx, req)
 	if s.client.client.Transport == nil {
-		resp, err = http.DefaultTransport.RoundTrip(req.WithContext(ctx))
+		resp, err = http.DefaultTransport.RoundTrip(req)
 	} else {
-		resp, err = s.client.client.Transport.RoundTrip(req.WithContext(ctx))
+		resp, err = s.client.client.Transport.RoundTrip(req)
 	}
 	if err != nil {
 		return nil, nil, err
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -244,7 +244,8 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
 	}
 	defer func() { s.client.client.CheckRedirect = saveRedirect }()
 
-	resp, err := s.client.client.Do(req.WithContext(ctx))
+	ctx, req = addContext(ctx, req)
+	resp, err := s.client.client.Do(req)
 	if err != nil {
 		if !strings.Contains(err.Error(), "disable redirect") {
 			return nil, "", err


github/without_appengine.go
```diff
@@ -0,0 +1,19 @@
...
+// +build !appengine
+
+package github
+
+import (
+	"context"
+	"net/http"
+)
+
+func addContext(ctx context.Context, req *http.Request) (context.Context, *http.Request) {
```
shurcooL: Minor, but `withContext` seems like a better fitting name, given the two implementations call:

```Go
req.WithContext(ctx)
appengine.WithContext(ctx, req)
```

github/github.go
```diff
@@ -390,7 +390,7 @@ func parseRate(r *http.Response) Rate {
 // The provided ctx must be non-nil. If it is canceled or times out,
 // ctx.Err() will be returned.
 func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
-	req = req.WithContext(ctx)
+	ctx, req = addContext(ctx, req)
```
shurcooL: Aren't we glad that we decided to add `ctx` parameter to `Do`, instead of doing `Do(req.WithContext(ctx), ...)` in all the call sites? 😆

github/with_appengine.go
```diff
@@ -0,0 +1,21 @@
...
+// license that can be found in the LICENSE file.
+
+// This file provides glue for making github work on App Engine.
+
+// +build appengine
+
+package github
+
+import (
+	"context"
```
shurcooL: I don't use App Engine, so I'm not familiar. If https://github.com/google/go-github/issues/526#issuecomment-284177413 is accurate, and App Engine uses/forces you to use Go 1.6, how does this work? Package `context` was added to Go standard library in Go 1.7, it wasn't available in Go 1.6.

github/with_appengine.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// This file provides glue for making github work on App Engine.
+
+// +build appengine
```
shurcooL: Minor and has no effect, but I'd suggest swapping the order of these two lines, like so:

```
// +build appengine

// This file provides glue for making github work on App Engine.
```

This is because I think the build constraints have higher priority and should be closer to top of the file. It's better for people to see the constraints as soon as possible, before the rest of the file. The only exception is that we place the copyright comment on top of that.

Somewhat of a precedent is [here](https://github.com/google/go-github/blob/04b64a7f8ccd7701ed660231f3805bd585fb486c/github/gen-accessors.go#L6-L8).

github/github.go
shurcooL: Question about the path on App Engine. Again, I'm not familiar with how it works, but I'd like to understand for this PR.

Given that ctx is not really used, and `addContext` modifies only `ctx`, not `req`, does it actually have an effect on App Engine? Do you know how?

gmlewis: Yes, it has an effect on App Engine, and what App Engine does depends on what environment the code is run under.

Here is an example:
https://github.com/golang/appengine/blob/master/internal/api_classic.go#L56

Basically, it associates each App Engine request with a context so that logging (and other things) can be associated with a particular request.
When you are dealing with hundreds or thousands of queries per second, it is a really nice feature in the developers console to associate the log lines with the request that caused the them.
You can then mentally follow the flow of the single request.

github/with_appengine.go
gmlewis: That's an excellent point, and I believe this was an oversight on my part.
Internally, we have our own `context` so I didn't remember this was an issue.
I'll change it.

github/github.go
shurcooL: I see. To make my question more specific, how does it have an effect given that the returned modified value of `ctx` is never used (aside from `ctx.Done()` and `ctx.Err()`) and never returned from `Do`?

Is it because the implementation of `appengine.WithContext` has side effects, other than just returning a modified copy of `ctx`?

github/with_appengine.go
gmlewis: Although now that I'm thinking about this a little more, I'm wondering how the other `import "context"` in the rest of this package are going to work.

I think I'm going to have to do some more experimentation with this PR before we call it ready for review.  Sorry, @shurcooL.

shurcooL: If that's the case, that's going to be an issue. Just about every single file in package `github` imports `context`.

- https://gotools.org/github.com/google/go-github/github?rev=33bd2319b7816a4781538fb06f712d054d4c5271#activity.go-L8
- https://gotools.org/github.com/google/go-github/github?rev=33bd2319b7816a4781538fb06f712d054d4c5271#activity-events.go-L9
- https://gotools.org/github.com/google/go-github/github?rev=33bd2319b7816a4781538fb06f712d054d4c5271#activity-notifications.go-L9
- (60+ more files)

- https://github.com/google/go-github/blob/33bd2319b7816a4781538fb06f712d054d4c5271/github/activity.go#L8
- https://github.com/google/go-github/blob/33bd2319b7816a4781538fb06f712d054d4c5271/github/activity_events.go#L9
- https://github.com/google/go-github/blob/33bd2319b7816a4781538fb06f712d054d4c5271/github/activity_notifications.go#L9
- (60+ more files)

If we need appengine to import `golang.org/x/net/context` instead, we will have to make a copy of each and every .go file.

That would be a pretty significant disturbance, and it was one of the main reasons we decided to wait for Go 1.8 to come out, so that we could drop support for 1.6 and older before adding context support to this package.

gmlewis: Yes.  I seem to remember hearing Dave Symonds say that the two implementations are 100% compatible... so maybe I don't need to change things after all... I'm just not sure if the Go 1.6 compiler will actually find the bare `"context"` import or not.
I definitely think that we don't need to make those kinds of accommodations here in this repo, though.

vvakame:
How's it going? I need this change...

gmlewis:
Sorry for the delay, @vvakame ... feel free to patch this into your code and try it out to see if it works while you are waiting for me to try it out myself. I've had some other pressing issues I'm attending to.

gmlewis:
After discussing this with @willnorris a bit, we are thinking that you will need to run the following command to make this run with the current version of Go on App Engine Classic (which as of this writing uses Go 1.6):

`gofmt -w -r '"context" -> "golang.org/x/net/context"' *.go`

This will rewrite all the `context` imports so that it should compile with Go 1.6.

We do not plan on supporting multiple versions of the Go compiler in this package, so we may simply have to add instructions to the `with_appengine.go` file on how to perform this transformation as long as App Engine Classic has not moved to Go 1.8.

github/with_appengine.go
```diff
@@ -0,0 +1,26 @@
...
+// Fortunately, this is easy with "gofmt":
+//
+//     gofmt -w -r '"context" -> "golang.org/x/net/context"' *.go
+
+package github
+
+import (
+	"net/http"
+
+	"golang.org/x/net/context"
```
shurcooL: Since all other files import `"context"`, and the `gofmt -w -r '"context" -> "golang.org/x/net/context"' *.go` step is necessary, is there any advantage to importing `"golang.org/x/net/context"` here instead of `"context"`?

In the end, it doesn't matter, but I'm leaning slightly towards preferring `"context"` because it's more consistent. Either `"context"` package doesn't exist, and the rewrite is neccessary everywhere, or it does exist, and nothing needs to be done anywhere. It would avoid importing both types of context package if `"context"` happens to be (or becomes) available.

I don't feel too strongly about this, so feel free to go with whatever you think is best.

github/with_appengine.go
```diff
@@ -0,0 +1,26 @@
...
+// license that can be found in the LICENSE file.
+
+// +build appengine
+
+// This file provides glue for making github work on App Engine.
+// In order to get the entire github package to compile with
+// Go 1.6, you will need to rewrite all the import "context" lines.
+// Fortunately, this is easy with "gofmt":
+//
+//     gofmt -w -r '"context" -> "golang.org/x/net/context"' *.go
```
shurcooL: Asking users to modify the source code of a package in order to get it to work is pretty unusual, and I feel this comment would be hard to spot.

I think a nice thing we can do for users is mention this in a new "App Engine" section of README (and package comment, to keep them in sync). It can simply state the current situation, that the package is expected not to build by default and needs to be modified, and point to this file for further instructions. Similar to what you wrote in https://github.com/google/go-github/pull/582#issuecomment-286822789.

gmlewis: SGTM.
Done.

github/with_appengine.go
gmlewis: Excellent point.
Done.

samuelkaufman:
Awesome! Thanks guys, I'll try and test this out today. I'm using godep already so formatting the code in the vendor dir and adding the gofmt command to my build process should not be hard.

PR #583, opened 2017-03-07
Don't panic when processing event payload

begizi:
The `Event.Payload` function can fail in instances when it receives invalid
json. (example: when parsing a webhook that is set to application/x-www-form-urlencoded)
If this happens, we shouldn't panic but instead return an error to let the
caller know about the failure.

Added in a new function, `Event.ParsePayload` to take its place a depreciated
the old function.

Also, it looks like this is the only place where we are panicing in this library. https://github.com/google/go-github/search?q=panic&type=Code&utf8=%E2%9C%93

### DIFF
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -28,9 +28,9 @@ func (e Event) String() string {
 	return Stringify(e)
 }
 
-// Payload returns the parsed event payload. For recognized event types,
+// ParsePayload parses the event payload. For recognized event types,
 // a value of the corresponding struct type will be returned.
-func (e *Event) Payload() (payload interface{}) {
+func (e *Event) ParsePayload() (payload interface{}, err error) {
 	switch *e.Type {
 	case "CommitCommentEvent":
 		payload = &CommitCommentEvent{}
@@ -89,8 +89,20 @@ func (e *Event) Payload() (payload interface{}) {
 	case "WatchEvent":
 		payload = &WatchEvent{}
 	}
-	if err := json.Unmarshal(*e.RawPayload, &payload); err != nil {
-		panic(err.Error())
+	err = json.Unmarshal(*e.RawPayload, &payload)
+	return payload, err
+}
+
+// Payload returns the parsed event payload. For recognized event types,
+// a value of the corresponding struct type will be returned.
+//
+// Deprecated: Use ParsePayload instead, which returns an error
+// rather than panics if JSON unmarshaling raw payload fails.
+func (e *Event) Payload() (payload interface{}) {
+	var err error
+	payload, err = e.ParsePayload()
+	if err != nil {
+		panic(err)
 	}
 	return payload
 }
--- a/github/activity_events_test.go
+++ b/github/activity_events_test.go
@@ -276,7 +276,7 @@ func TestActivityService_ListUserEventsForOrganization(t *testing.T) {
 	}
 }
 
-func TestActivityService_EventPayload_typed(t *testing.T) {
+func TestActivityService_EventParsePayload_typed(t *testing.T) {
 	raw := []byte(`{"type": "PushEvent","payload":{"push_id": 1}}`)
 	var event *Event
 	if err := json.Unmarshal(raw, &event); err != nil {
@@ -284,15 +284,19 @@ func TestActivityService_EventPayload_typed(t *testing.T) {
 	}
 
 	want := &PushEvent{PushID: Int(1)}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	got, err := event.ParsePayload()
+	if err != nil {
+		t.Fatalf("ParsePayload returned unexpected error: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Event.ParsePayload returned %+v, want %+v", got, want)
 	}
 }
 
 // TestEvent_Payload_untyped checks that unrecognized events are parsed to an
 // interface{} value (instead of being discarded or throwing an error), for
 // forward compatibility with new event types.
-func TestActivityService_EventPayload_untyped(t *testing.T) {
+func TestActivityService_EventParsePayload_untyped(t *testing.T) {
 	raw := []byte(`{"type": "UnrecognizedEvent","payload":{"field": "val"}}`)
 	var event *Event
 	if err := json.Unmarshal(raw, &event); err != nil {
@@ -300,12 +304,16 @@ func TestActivityService_EventPayload_untyped(t *testing.T) {
 	}
 
 	want := map[string]interface{}{"field": "val"}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	got, err := event.ParsePayload()
+	if err != nil {
+		t.Fatalf("ParsePayload returned unexpected error: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Event.ParsePayload returned %+v, want %+v", got, want)
 	}
 }
 
-func TestActivityService_EventPayload_installation(t *testing.T) {
+func TestActivityService_EventParsePayload_installation(t *testing.T) {
 	raw := []byte(`{"type": "PullRequestEvent","payload":{"installation":{"id":1}}}`)
 	var event *Event
 	if err := json.Unmarshal(raw, &event); err != nil {
@@ -313,7 +321,11 @@ func TestActivityService_EventPayload_installation(t *testing.T) {
 	}
 
 	want := &PullRequestEvent{Installation: &Installation{ID: Int(1)}}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	got, err := event.ParsePayload()
+	if err != nil {
+		t.Fatalf("ParsePayload returned unexpected error: %v", err)
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Event.ParsePayload returned %+v, want %+v", got, want)
 	}
 }
--- a/github/messages.go
+++ b/github/messages.go
@@ -162,7 +162,7 @@ func WebHookType(r *http.Request) string {
 
 // ParseWebHook parses the event payload. For recognized event types, a
 // value of the corresponding struct type will be returned (as returned
-// by Event.Payload()). An error will be returned for unrecognized event
+// by Event.ParsePayload()). An error will be returned for unrecognized event
 // types.
 //
 // Example usage:
@@ -191,5 +191,5 @@ func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
 		Type:       &eventType,
 		RawPayload: (*json.RawMessage)(&payload),
 	}
-	return event.Payload(), nil
+	return event.ParsePayload()
 }


gmlewis:
Yes, I strongly believe that the library should not panic, ever.
We recently had this discussion in another issue.
Thank you, @begizi!
Reviewing changes...

github/activity_events.go
```diff
@@ -89,10 +89,10 @@ func (e *Event) Payload() (payload interface{}) {
 	case "WatchEvent":
 		payload = &WatchEvent{}
 	}
-	if err := json.Unmarshal(*e.RawPayload, &payload); err != nil {
-		panic(err.Error())
+	if err = json.Unmarshal(*e.RawPayload, &payload); err != nil {
```
gmlewis: nit: there is no harm in leaving this as `:=` here and it is a good pattern in Go.

github/activity_events_test.go
```diff
@@ -284,8 +284,9 @@ func TestActivityService_EventPayload_typed(t *testing.T) {
 	}
 
 	want := &PushEvent{PushID: Int(1)}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	p, _ := event.Payload()
```
gmlewis:
```go
p, err := event.Payload()
if err != nil {
	t.Errorf("Payload = %v, want nil", err)
}
```

github/activity_events_test.go
```diff
@@ -300,8 +301,9 @@ func TestActivityService_EventPayload_untyped(t *testing.T) {
 	}
 
 	want := map[string]interface{}{"field": "val"}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	p, _ := event.Payload()
```
gmlewis:
```go
p, err := event.Payload()
if err != nil {
	t.Errorf("Payload = %v, want nil", err)
}
```

github/activity_events_test.go
```diff
@@ -313,7 +315,8 @@ func TestActivityService_EventPayload_installation(t *testing.T) {
 	}
 
 	want := &PullRequestEvent{Installation: &Installation{ID: Int(1)}}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	p, _ := event.Payload()
```
gmlewis:
```go
p, err := event.Payload()
if err != nil {
	t.Errorf("Payload = %v, want nil", err)
}
```

begizi:
@gmlewis Addressed comments. Thanks!

begizi:
@gmlewis Updated to use `got`

github/activity_events_test.go
```diff
@@ -284,8 +284,12 @@ func TestActivityService_EventPayload_typed(t *testing.T) {
 	}
 
 	want := &PushEvent{PushID: Int(1)}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	got, err := event.Payload()
+	if err != nil {
+		t.Errorf("Payload = %v, want nil", err)
```
shurcooL: This should be `t.Fatalf`, not `t.Errorf`. If `event.Payload()` returned a non-nil error, there's no point in continuing testing the `got` value, it's only noise. /cc @gmlewis

Also in 2 similar places below.

github/activity_events.go
```diff
@@ -90,9 +90,9 @@ func (e *Event) Payload() (payload interface{}) {
 		payload = &WatchEvent{}
 	}
 	if err := json.Unmarshal(*e.RawPayload, &payload); err != nil {
-		panic(err.Error())
+		return nil, err
 	}
-	return payload
+	return payload, nil
```
shurcooL: The 4 lines above can be simplified to:

```Go
err := json.Unmarshal(*e.RawPayload, &payload)
return payload, err
```

github/activity_events.go
```diff
@@ -30,7 +30,7 @@ func (e Event) String() string {
 
 // Payload returns the parsed event payload. For recognized event types,
 // a value of the corresponding struct type will be returned.
-func (e *Event) Payload() (payload interface{}) {
+func (e *Event) Payload() (payload interface{}, err error) {
```
shurcooL: Given this is a breaking API change anyway, I think we should also rename the method to be a better fit for what it does.

`Payload` was okay when it returned a single value and never failed.

Now, I think a better name would be:

```Go
// ParsePayload parses the event payload. For recognized event types,
func (e *Event) ParsePayload() (payload interface{}, err error) {
```

It'll be similar to `ParseWebHook`.

shurcooL:
I'll also ask the following question.

Given we have a higher level [`ParseWebHook`](https://godoc.org/github.com/google/go-github/github#ParseWebHook) function, does this payload method even need to still be exported? Does it provide value, or only pollute the API needlessly?

github/activity_events_test.go
gmlewis: Agreed.  Thanks, @shurcooL.

begizi:
@shurcooL tests updated, return simplified, and name changed. I agree that we probably don't need to be exporting the `ParsePayload` function, but couldn't say with confidence someone might not find it useful. Happy to change it if we decide to go in that direction though.

gmlewis:
I'm thinking that if someone already has an `*Event`, possibly from one of these calls:

```
github/activity_events.go :101:func (s *ActivityService) ListEvents(ctx context.Context, opt *ListOptions) ([]*Event, *Response, error) {
github/activity_events.go :124:func (s *ActivityService) ListRepositoryEvents(ctx context.Context, owner, repo string, opt *ListOptions) ([]*Event, *Response, error) {
github/activity_events.go :172:func (s *ActivityService) ListEventsForRepoNetwork(ctx context.Context, owner, repo string, opt *ListOptions) ([]*Event, *Response, error) {
github/activity_events.go :196:func (s *ActivityService) ListEventsForOrganization(ctx context.Context, org string, opt *ListOptions) ([]*Event, *Response, error) {
github/activity_events.go :221:func (s *ActivityService) ListEventsPerformedByUser(ctx context.Context, user string, publicOnly bool, opt *ListOptions) ([]*Event, *Response, error) {
github/activity_events.go :251:func (s *ActivityService) ListEventsReceivedByUser(ctx context.Context, user string, publicOnly bool, opt *ListOptions) ([]*Event, *Response, error) {
github/activity_events.go :281:func (s *ActivityService) ListUserEventsForOrganization(ctx context.Context, org, user string, opt *ListOptions) ([]*Event, *Response, error) {
```

Then they will want to call `Payload` directly, and `ParseWebhook` will not be that useful to them.

So I think we want to keep the function.

shurcooL:
Thanks @gmlewis for confirming that.

Agreed, we need to keep the function exported.

github/activity_events_test.go
```diff
@@ -284,8 +284,12 @@ func TestActivityService_EventPayload_typed(t *testing.T) {
 	}
 
 	want := &PushEvent{PushID: Int(1)}
-	if !reflect.DeepEqual(event.Payload(), want) {
-		t.Errorf("Event Payload returned %+v, want %+v", event.Payload(), want)
+	got, err := event.ParsePayload()
+	if err != nil {
+		t.Fatalf("Payload = %v, want nil", err)
```
shurcooL: This should now say "ParsePayload", and perhaps be more clear that it was the error returned. How about:

```Go
t.Fatalf("ParsePayload returned unexpected error: %v", err)
```

Consistent with https://github.com/google/go-github/blob/cfbfe91dfc6f331a809a2c942f64b6bad65d9c91/github/github_test.go#L631.

github/activity_events_test.go
```diff
@@ -284,8 +284,12 @@ func TestActivityService_EventPayload_typed(t *testing.T) {
...
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Event Payload returned %+v, want %+v", got, want)
```
shurcooL: /s/Event Payload/Event.ParsePayload

begizi:
@shurcooL Addressed comments

begizi:
@shurcooL Addressed comments

github/messages.go
```diff
@@ -191,5 +191,5 @@ func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
 		Type:       &eventType,
```
shurcooL: You also missed a mention of old function name in the docs of this func.

It should be:

> as returned by Event.ParsePayload()

Rather than the current:

> as returned by Event.Payload()

begizi: Updated. Good catch 👍

github/activity_events.go
```diff
@@ -28,9 +28,9 @@ func (e Event) String() string {
 	return Stringify(e)
 }
 
-// Payload returns the parsed event payload. For recognized event types,
+// ParsePayload parses the event payload. For recognized event types,
 // a value of the corresponding struct type will be returned.
-func (e *Event) Payload() (payload interface{}) {
+func (e *Event) ParsePayload() (payload interface{}, err error) {
```
shurcooL: So, we actually have a choice here, and I just wanna make sure we've considered both options before going through with it.

1.	We make this a breaking API change, as done above, and bump libraryVersion by one when merging.

2.	We make this a not breaking API change by adding the old method, and marking it as deprecated. It would look like this:

	```Go
	// Payload returns the parsed event payload. For recognized event types,
	//
	// Deprecated: Use ParsePayload instead, which returns an error rather than panics if unmarshaling fails.
	func (e *Event) Payload() (payload interface{}) {
		var err error
		payload, err = e.ParsePayload()
		if err != nil {
			panic(err)
		}
	}
	```

gmlewis: I think option 2 is a great idea, @shurcooL. The fewer projects we break, the better, and this gives people the opportunity to update their code at their convenience.  I like it. Thanks.

begizi: Added in the old method with depreciation warning. Is it necessary to add in some tests for the old method since it's just proxying another method?

github/activity_events.go
```diff
@@ -30,7 +30,21 @@ func (e Event) String() string {
...
+// Deprecated: Use ParsePayload instead, which returns an error
+// rather than panics if JSON unmarshaling raw payload fails.
 func (e *Event) Payload() (payload interface{}) {
+	var err error
+	payload, err = e.ParsePayload()
+	if err != nil {
+		panic(err)
+	}
+	return payload
+}
```
shurcooL: Really minor, but would you mind swapping the order of `Payload` and `ParsePayload` in the source? I think the deprecated func should be below the preferred one.

begizi: Sure. Done!

shurcooL:
@begizi For future reference, when addressing feedback in a PR, feel free to push individual commits rather than amending and force pushing the single commit. It's really easy for us to squash the commits when merging the PR, after it has passed code review. Seeing the incremental changes in separate commits make it easier to review only what has changed, rather than having to review the entire commit each time.

github/activity_events.go
shurcooL: No need for extra tests for the deprecated method, it's trivial and deprecated.

shurcooL:
LGTM. Thanks @begizi!

I don't know what's going on with the Travis build, it's wedged.

gmlewis:
LGTM. We can merge after Travis comes back online.

shurcooL:
Looks like pushing a new commit (to edit the commit message) has unwedged it. @gmlewis, wanna take merging this one?

gmlewis:
SGTM. Merging.

PR #586, opened 2017-03-08
Invitation methods take :owner/:repo

alindeman:
All repository routes on the GitHub API can accessed via /repositories/:id _and_ /repos/:owner/:repo. For whatever reason, the documentation for the invitation routes only references the former, but both work correctly.

Repository invitations are the only functions in google/go-github to use the :id style routes, so I'd propose we change them to use :owner/:repo as that will likely be much more convenient for users of this library.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	libraryVersion = "3"
+	libraryVersion = "4"
 	defaultBaseURL = "https://api.github.com/"
 	uploadBaseURL  = "https://uploads.github.com/"
 	userAgent      = "go-github/" + libraryVersion
--- a/github/repos_invitations.go
+++ b/github/repos_invitations.go
@@ -28,8 +28,8 @@ type RepositoryInvitation struct {
 // ListInvitations lists all currently-open repository invitations.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository
-func (s *RepositoriesService) ListInvitations(ctx context.Context, repoID int, opt *ListOptions) ([]*RepositoryInvitation, *Response, error) {
-	u := fmt.Sprintf("repositories/%v/invitations", repoID)
+func (s *RepositoriesService) ListInvitations(ctx context.Context, owner, repo string, opt *ListOptions) ([]*RepositoryInvitation, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/invitations", owner, repo)
 	u, err := addOptions(u, opt)
 	if err != nil {
 		return nil, nil, err
@@ -55,8 +55,8 @@ func (s *RepositoriesService) ListInvitations(ctx context.Context, repoID int, o
 // DeleteInvitation deletes a repository invitation.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/invitations/#delete-a-repository-invitation
-func (s *RepositoriesService) DeleteInvitation(ctx context.Context, repoID, invitationID int) (*Response, error) {
-	u := fmt.Sprintf("repositories/%v/invitations/%v", repoID, invitationID)
+func (s *RepositoriesService) DeleteInvitation(ctx context.Context, owner, repo string, invitationID int) (*Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/invitations/%v", owner, repo, invitationID)
 	req, err := s.client.NewRequest("DELETE", u, nil)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


shurcooL:
Hmm, this is very interesting. On the surface, what you're saying is very true. I want to dig deeper to better understand how this happened and what's going on.

These methods were implemented 9 months ago, in #373, and reviewed by @gmlewis and me.

At first, I thought this was about one of the new features we've merged in the last few days, but 9 months? I'm surprised no one has spotted this by now.

Looking at the GitHub announcement the underlying issue was about, and then the documentation, I can see how this happened. The GitHub documentation actually uses repo IDs rather than owner/repo scheme.

image

(Source: https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository.)

As you said, both are indeed valid, but this library and most endpoints in GitHub API docs use the latter form.

I think we should contact GitHub support and ask them about this. Perhaps there's a good reason why these endpoints are documented using repo ID form rather than owner/repo. If so, we'll need to think harder about this.

Otherwise, and I suspect this is the case, it could be just an unintended outcome. If so, they might decide to fix their documentation to use the consistent owner/repo form.

If that happens, it should be much easier for us to make the decision to follow suit.

Given that this has been around for 9 months, it's a little worrisome about breaking the API... but unless I'm missing something, this alternate form must be more awkward to use anyway. Also, these endpoints are still considered as "preview API", it's not yet stable version, which gives us more room to fix things.

alindeman:
@shurcooL Here's what I got from GitHub support

> As for your question -- that looks like a problem in the documentation. You should definitely be able to use the standard routes like /repos/:owner/:repo/invitations. Not sure if you remember this, but most (if not all) repository-related API endpoints can be accessed through two routes:
>
> /repos/:owner/:repo/...
> 
> /repositories/:repo_id/...
>
> They're equivalent and route to the same code. Not sure why the team decided to document the one with the repository ID instead of the name for this specific endpoint, but we should be able to make that more consistent.
>
> The one with the name was even used in the related blog posts:
>
> https://developer.github.com/changes/2016-06-14-repository-invitations/
>
> I hope this answers your question and thanks for reaching out about this!

In my specific case, I'm trying to plug this API into code that only has `owner/repo` available to it. I have to fetch the repo with `Repositories.Get` just to get its ID to plug into these functions. I suspect many other implementations have to do the same.

gmlewis:
Hmmm... I'm now wondering if we should have another `go generate` function that makes these `Repo` APIs accessible either by "owner/repo" or by "repo_id" (keeping the same name as the original but adding `ByID` as a suffix to the method name)?

Maybe that is overkill... but I'm not sure I like breaking this API here either when both methods of access should be possible via the API.

Of course, a GraphQL version of the API should solve this, but we aren't there yet.

shurcooL:
Thanks for emailing them.

> As for your question -- that looks like a problem in the documentation. You should definitely be able to use the standard routes like /repos/:owner/:repo/invitations.
> ...
> Not sure why the team decided to document the one with the repository ID instead of the name for this specific endpoint, but we should be able to make that more consistent.
>
> The one with the name was even used in the related blog posts

That's great to hear, and I think that gives us a good chance of being able to fix up this unintended inconsistency in the API.

> Hmmm... I'm now wondering if we should have another `go generate` function that makes these `Repo` APIs accessible either by "owner/repo" or by "repo_id" (keeping the same name as the original but adding `ByID` as a suffix to the method name)?
>
> Maybe that is overkill... but I'm not sure I like breaking this API here either when both methods of access should be possible via the API.
> 
> Of course, a GraphQL version of the API should solve this, but we aren't there yet.

@gmlewis I think it'd be better to have a discussion about such a large and sweeping potential change in a separate issue. I'm really worried about anything that increases API sizes, and doubling is very significant. That said, I want to refer you to some relevant previous issues on the topic. Consider #68, which I think proposed a very good approach for being able to use both owner/repo and ID-based ways of referring to repositories, users. Unfortunately, it didn't make it in because it was too late to make such a breaking API change.

gmlewis:
Thanks for the pointer, @shurcooL, I had not seen that. Very clever.
Yes, sounds like fixing the inconsistency is the way to go, then, thus breaking the API.
We could bump the version again.

shurcooL:
> Yes, sounds like fixing the inconsistency is the way to go, then, thus breaking the API.
> We could bump the version again.

Let's do that.

alindeman:
Great, thanks! You all will take the integration from here?

gmlewis:
It would be helpful if you, @alindeman, bump the library version number in `github.go` and add a commit to this PR.  Thank you!

alindeman:
> It would be helpful if you, @alindeman, bump the library version number in github.go and add a commit to this PR. Thank you!

@gmlewis Sure, done.

shurcooL:
Ok, merging.

PR #593, opened 2017-03-16
Add Label.ID field and accessor method.

bradfitz:


github/issues_labels.go
```diff
@@ -21,6 +22,13 @@ func (l Label) String() string {
 	return fmt.Sprint(*l.Name)
```
bradfitz: what was this code trying to do?

willnorris: Aside from the obvious of having Label implement Stringer?  Not sure why it's using fmt.Sprint; looks like it got added in b416fdaac841e7c2404b365cf6aa7b9f1d450d93.  Should just return `*l.Name` I guess.

github/issues_labels.go
```diff
@@ -21,6 +22,13 @@ func (l Label) String() string {
...
 }
 
+func (l Label) GetID() int {
```
willnorris: accessors should be generated using `go generate`.  /cc @gmlewis for details as this was just recently added.

gmlewis:
Right... you don't need to manually add accessors...  `go generate` now does it for you.
Please see CONTRIBUTING.md for details.

bradfitz:
Updated.

github/issues_labels.go
```diff
@@ -12,13 +12,14 @@ import (
...
 type Label struct {
+	ID    *int    `json:"id,omitempty"`
 	URL   *string `json:"url,omitempty"`
 	Name  *string `json:"name,omitempty"`
 	Color *string `json:"color,omitempty"`
 }
 
 func (l Label) String() string {
-	return fmt.Sprint(*l.Name)
+	return *l.Name
```
shurcooL: This change is an improvement, but really, this was meant to be `return Stringify(l)`. /cc @willnorris

```Go
func (l Label) String() string {
	return Stringify(l)
}
```

If you [look at the implementations](https://cloud.githubusercontent.com/assets/1924134/24018043/b70540d4-0a68-11e7-88bd-c901e4f5757e.png) of all other `String() string` methods on the go-github types, that's what they do. This is the only unintended exception (other than `Timestamp.String`, but that's a special case on purpose).

> what was this code trying to do?

The intent is to implement `fmt.Stringer` interface and print a more readable value for debugging purposes (since all the fields are pointers). There's no other contract it needs to satisfy.

github/issues_labels_test.go
```diff
@@ -160,7 +160,7 @@ func TestIssuesService_ListLabelsByIssue(t *testing.T) {
 	mux.HandleFunc("/repos/o/r/issues/1/labels", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testFormValues(t, r, values{"page": "2"})
-		fmt.Fprint(w, `[{"name": "a"},{"name": "b"}]`)
+		fmt.Fprint(w, `[{"name": "a","id": 1},{"name": "b", "id": 2}]`)
```
shurcooL: Nitpick. It's inconsistent how first added comma has no space after it, but the other comma has a space after it.

github/issues_labels.go
bradfitz: Done.

bradfitz: Actually, there's a test locking in this behavior:

                {Label{Name: String("l")}, "l"},

So I'm going to just remove this from this CL.

bradfitz:
PTAL, with String method change removed and a space added in the test's JSON.

bradfitz:
Actually, just removed all spaces to make it even more consistent. PTAL.

github/issues_labels_test.go
bradfitz: Done.

github/issues_labels.go
shurcooL: Sounds good. I actually dug up more history behind that decision, and filed #595 to deal with resolving it (definitely a good idea to leave it out of this PR).

shurcooL:
@gmlewis Does your original LGTM still apply, or do you want to look it over? The PR is now a subset of the one you approved, and I've created #595 to deal with the `Label.String` change separately.

gmlewis:
LGTM

shurcooL:
All right, I can merge this.

PR #598, opened 2017-03-19
add tokentype=basic to example code

queirozfcom:
If users obtain a github API Token (as suggested in the example), they need to add `TokenType: "Basic"` to the client definition, or they get `401 Unauthorized`.

Apparently, the default value is something else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ import "golang.org/x/oauth2"
 func main() {
   ctx := context.Background()
   ts := oauth2.StaticTokenSource(
-    &oauth2.Token{AccessToken: "... your access token ..."},
+    &oauth2.Token{TokenType: "Basic", AccessToken: "... your access token ..."},
   )
   tc := oauth2.NewClient(ctx, ts)
 


queirozfcom:
I signed the Google CLA.

shurcooL:
> Apparently, the default value is something else.

The default is "Bearer", it's documented here:

https://godoc.org/golang.org/x/oauth2#Token.TokenType

We try to keep the README and package comment in sync, so if you're making changes here, you should update https://github.com/google/go-github/blob/67f700645e97e82b8cf282ece3508805438be24f/github/doc.go#L47 as well.

> or they get `401 Unauthorized`.

I have a question. I've been using this very code, setting only `AccessToken`, and it works fine for me, no 401.

Do you have a source for that, or can you provide a way to reproduce this?

queirozfcom:
@shurcooL Hi.

I've setup a minimal example here: [https://github.com/queirozfcom/tracker-api/blob/token_issue/src/github.com/queirozfcom/trackerapi/main/main_token_issue.go](https://github.com/queirozfcom/tracker-api/blob/token_issue/src/github.com/queirozfcom/trackerapi/main/main_token_issue.go)

(you need a GITHUB_TOKEN environment variable before running this, though. If on linux, just to `export GITHUB_TOKEN=xxxxxxx` before runing the application)

shurcooL:
Thanks, I'll try to reproduce with that.

How did you create the GitHub token? Is it a personal access token? With what scopes? Something else?

shurcooL:
I cannot reproduce.

Here's what I did. I went to https://github.com/settings/tokens/new and created a new personal access token without any scopes:

image

Then I ran your program (slightly modified output), and it worked without errors:

```
$ go run main.go
err: <nil>
repos: [github.Repository{ID:23498380, Owner:github.User{Login:"ajhager", ID:42474, AvatarURL:"https://avatars3.githubusercontent.com/u/42474?v=3", ...
```

What did you do differently?

shurcooL:
Friendly ping @queirozfcom. We won't be able to proceed with this PR until there's more information to help reproduce and understand this issue.

kaakaa:
I encountered the same error and had resolved it.

My case is a special case, which may be different from the case of the Issue reporter.
But I will write it down if it is helpful for something.

Since the organization uses a self-signed certificate, it is necessary to set `InsecureSkipVerify: true` when performing https communication.
When setting an incorrect `InsecureSkipVerify`, it returns a 401 error like this Issue.

The incorrect setting is as follows.

```
        &oauth2.Token{AccessToken: accesstoken},
    )
    tc.Transport = &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
    }
    client := github.NewClient(oauthClient)
```

I changed the setting with reference to [this comments](https://github.com/golang/oauth2/issues/187#issuecomment-227811477), and now it works normally.

```
    tr := &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
    }
    sslcli := &http.Client{Transport: tr}
    ctx := context.TODO()
    ctx = context.WithValue(ctx, oauth2.HTTPClient, sslcli)

        &oauth2.Token{AccessToken: accesstoken},
    )
    tc := oauth2.NewClient(ctx, ts)
    client := github.NewClient(tc)
```

gmlewis:
Closing as resolved.

PR #599, opened 2017-03-19
OrganizationsService: Add projects methods

shawncatz:
ListProjects returns list of organization projects
CreateProject creates organization project

### DIFF
--- a/github/orgs_projects.go
+++ b/github/orgs_projects.go
@@ -1,4 +1,4 @@
-// Copyright 2016 The go-github AUTHORS. All rights reserved.
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
@@ -28,7 +28,7 @@ func (s *OrganizationsService) ListProjects(ctx context.Context, org string, opt
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeProjectsPreview)
 
-	projects := []*Project{}
+	var projects []*Project
 	resp, err := s.client.Do(ctx, req, &projects)
 	if err != nil {
 		return nil, resp, err
--- a/github/orgs_projects.go
+++ b/github/orgs_projects.go
@@ -12,7 +12,7 @@ import (
 
 // ListProjectsOptions specifies the optional parameters to the
 // OrganizationsService.ListProjects method.
-type ListProjectsOptions struct {
+type ProjectsListOptions struct {
 	// Indicates the state of the projects to return. Can be either open, closed, or all. Default: open
 	State string `url:"state,omitempty"`
 
@@ -22,7 +22,7 @@ type ListProjectsOptions struct {
 // ListProjects lists the projects for an organization.
 //
 // GitHub API docs: https://developer.github.com/v3/projects/#list-organization-projects
-func (s *OrganizationsService) ListProjects(ctx context.Context, org string, opt *ListProjectsOptions) ([]*Project, *Response, error) {
+func (s *OrganizationsService) ListProjects(ctx context.Context, org string, opt *ProjectsListOptions) ([]*Project, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/projects", org)
 	u, err := addOptions(u, opt)
 	if err != nil {
--- a/github/orgs_projects.go
+++ b/github/orgs_projects.go
@@ -10,9 +10,9 @@ import (
 	"fmt"
 )
 
-// ListProjectsOptions specifies the optional parameters to the
+// ProjectListOptions specifies the optional parameters to the
 // OrganizationsService.ListProjects method.
-type ProjectsListOptions struct {
+type ProjectListOptions struct {
 	// Indicates the state of the projects to return. Can be either open, closed, or all. Default: open
 	State string `url:"state,omitempty"`
 
@@ -22,7 +22,7 @@ type ProjectsListOptions struct {
 // ListProjects lists the projects for an organization.
 //
 // GitHub API docs: https://developer.github.com/v3/projects/#list-organization-projects
-func (s *OrganizationsService) ListProjects(ctx context.Context, org string, opt *ProjectsListOptions) ([]*Project, *Response, error) {
+func (s *OrganizationsService) ListProjects(ctx context.Context, org string, opt *ProjectListOptions) ([]*Project, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/projects", org)
 	u, err := addOptions(u, opt)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


github/orgs_projects.go
```diff
@@ -0,0 +1,69 @@
...
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// ListProjectsOptions specifies the optional parameters to the
+// OrganizationsService.ListProjects method.
+type ListProjectsOptions struct {
```
shurcooL: What do you think of `ProjectListOptions` as an alternative name for this struct?

Also, we should probably update `RepositoriesService.ListProjects` method to also use it, given that https://developer.github.com/v3/projects/#list-repository-projects has the same list parameters as https://developer.github.com/v3/projects/#list-organization-projects. Are you okay with doing that as part of this PR?

shurcooL:
Thanks for working on this @shawncatz, it's looking great overall.

github/orgs_projects.go
shawncatz: Yeah, no problem. Happy to change it.

I wasn't sure which name to use... it seems like there are some that use both naming conventions...

Some examples:
repos_commits.go:90:type CommitsListOptions struct {
orgs_members.go:49:type ListMembersOptions struct {

I'll change the name to `ProjectsListOptions` to match the convention used by `CommitsListOptions`

Do you have a preference for where it should be defined? Since both `orgs_projects.go` and `repos_projects.go` will be using it?

github/orgs_projects.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // ListProjectsOptions specifies the optional parameters to the
 // OrganizationsService.ListProjects method.
-type ListProjectsOptions struct {
+type ProjectsListOptions struct {
```
shurcooL: `ProjectListOptions` sounds better than `ProjectsListOptions`.

Also, you'll need to update the comment to use the new name.

shawncatz: yeah, missed the comment, my fail :)
Are we ok leaving the definition in `orgs_projects.go`?

shurcooL: Yes, the filename looks good. It's consistent with `repos_projects.go` and all the other `orgs_*.go` files.

shawncatz: > Are we ok leaving the definition in orgs_projects.go?

What I mean by this is that `ProjectListOptions` is used by `orgs_projects.go` and `repo_projects.go`, but is defined in `orgs_projects.go`... just wanted to be sure we didn't want to move it to somewhere else, since it's shared between the files.

shurcooL: Ah, I see.

I think it should be in one of the two files, either `repos_projects.go` or `orgs_projects.go`. It doesn't really matter which of the two.

You could move it to `repos_projects.go` since that one feels a little more "primary", but either way is completely fine. Thanks. :)

shawncatz:
@shurcooL I've moved the definition and updated the comment to include references to both places it's used.
This should be complete, let me know if there's anything else I can do.
Thanks :)

shurcooL:
Because the change of signature of `RepositoriesService.ListProjects` method is a breaking change, can you bump up the version by one at https://github.com/shawncatz/go-github/blob/1fe303ba3b6789190206f9652c8316f8d9d96bb5/github/github.go#L30?

shawncatz:
of course, didn't even think of that

github/orgs_projects.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2016/2017/

github/repos_projects.go
```diff
@@ -10,10 +10,19 @@ import (
 	"fmt"
 )
 
+// ProjectListOptions specifies the optional parameters to the
+// OrganizationsService.ListProjects and RepositoriesService.ListProjects methods.
+type ProjectListOptions struct {
+	// Indicates the state of the projects to return. Can be either open, closed, or all. Default: open
+	State string `url:"state,omitempty"`
```
gmlewis: @shurcooL - do you know why there are many instances of
```
`url:"..."`
```
throughout this repo instead of
```
`json:"..."`
```
?

```bash
$ pt -l url:
github/activity_notifications.go
github/activity_star.go
github/gists.go
github/git_refs.go
github/issues_comments.go
github/issues.go
github/github.go
github/issues_milestones.go
github/orgs.go
github/orgs_outside_collaborators.go
github/orgs_members.go
github/orgs_teams.go
github/pulls_comments.go
github/pulls.go
github/repos_commits.go
github/repos_deployments.go
github/repos_forks.go
github/repos.go
github/repos_traffic.go
github/search.go
github/repos_contents.go
github/users.go
```

I just did a Google search and couldn't find any mention of `url:` for use with `encoding/json` and the [docs](https://golang.org/pkg/encoding/json/) don't seem to mention anything either.

In fact, reading through https://golang.org/pkg/encoding/json/#Marshal, it appears to me that all of these format strings are being completely ignored, which is probably not our intention.  :smile:

I'm thinking that we should probably change this one to
```
`json:"..."`
```
and then open a new issue to change the other ones... unless you point out what I'm missing.

github/orgs_projects_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2016/2017/

github/orgs_projects.go
```diff
@@ -0,0 +1,60 @@
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeProjectsPreview)
+
+	projects := []*Project{}
```
gmlewis: var projects []*Project

github/repos_projects.go
shurcooL: This is normal. `url` struct field tag controls how the field gets marshaled into a query parameter.

Take a look at `addOptions` helper which operates on this struct:

https://github.com/google/go-github/blob/master/github/github.go#L191

Also look at the [`github.com/google/go-querystring/query`](github.com/google/go-querystring/query) package being used.

You can confirm for yourself by removing the `url` struct field tag, and see how the test fails. The query parameter will be encoded as "State" instead of "state" without it.

gmlewis: Ah, great.  Thank you, @shurcooL!
Then please totally disregard this comment, @shawncatz.

github/orgs_projects.go
shurcooL: That change should apply to `RepositoriesService.ListProjects` method too. It was taken from there.

shawncatz:
could I get an idea of when this might be merged? don't mean to push, just curious

shawncatz:
Thanks!

-- 
Shawn Catanzarite
408-605-7744
http://shawncatz.com/
http://linkedin.com/in/shawncatanzarite

Issue #601, opened 2017-03-28
Search Commits API Results Problem


amirilovic:
Issue: When calling search commits api (go-github/github/search.go:84) returned results contain correct number of items but each item properties are empty.

As per github search docs https://developer.github.com/v3/search/#search-commits

Search Commits is returning results as:
```
{
  "total_count": 1,
  "incomplete_results": false,
  "items": [
    {
      "url": "https://api.github.com/repos/octocat/Spoon-Knife/commits/bb4cc8d3b2e14b3af5df699876dd4ff3acd00b7f",
      "sha": "bb4cc8d3b2e14b3af5df699876dd4ff3acd00b7f",
      "html_url": "https://github.com/octocat/Spoon-Knife/commit/bb4cc8d3b2e14b3af5df699876dd4ff3acd00b7f",
      "comments_url": "https://api.github.com/repos/octocat/Spoon-Knife/commits/bb4cc8d3b2e14b3af5df699876dd4ff3acd00b7f/comments",
      "commit": {
        "url": "https://api.github.com/repos/octocat/Spoon-Knife/git/commits/bb4cc8d3b2e14b3af5df699876dd4ff3acd00b7f",
        "author": {
          "date": "2014-02-04T14:38:36-08:00",
          "name": "The Octocat",
          "email": "octocat@nowhere.com"
        },
        "committer": {
          "date": "2014-02-12T15:18:55-08:00",
          "name": "The Octocat",
          "email": "octocat@nowhere.com"
        },
        "message": "Create styles.css and updated README",
        "tree": {
          "url": "https://api.github.com/repos/octocat/Spoon-Knife/git/trees/a639e96f9038797fba6e0469f94a4b0cc459fa68",
          "sha": "a639e96f9038797fba6e0469f94a4b0cc459fa68"
        },
        "comment_count": 8
      },
      "author": {
        // author fields ...
      },
      "committer": {
       // committer fields ...
      },
      "parents": [
        {
          "url": "https://api.github.com/repos/octocat/Spoon-Knife/commits/a30c19e3f13765a3b48829788bc1cb8b4e95cee4",
          "html_url": "https://github.com/octocat/Spoon-Knife/commit/a30c19e3f13765a3b48829788bc1cb8b4e95cee4",
          "sha": "a30c19e3f13765a3b48829788bc1cb8b4e95cee4"
        }
      ],
      "repository": {
       // repo fields ...
      },
      "score": 4.9971514
    }
  ]
}
```

[...]

PR #602, opened 2017-03-28
Fixing issue with commit search results deserialization #601

amirilovic:


github/search.go
```diff
@@ -65,17 +65,16 @@ type CommitsSearchResult struct {
...
-	CommitterID    *int        `json:"committer_id,omitempty"`
-	CommitterName  *string     `json:"committer_name,omitempty"`
-	CommitterEmail *string     `json:"committer_email,omitempty"`
-	CommitterDate  *Timestamp  `json:"committer_date,omitempty"`
-	Repository     *Repository `json:"repository,omitempty"`
+	SHA         *string     `json:"sha,omitempty"`
+	Commit      *Commit     `json:"commit,omitempty"`
+	Author      *User       `json:"author,omitempty"`
+	Committer   *User       `json:"committer,omitempty"`
+	Parents     []Commit    `json:"parents,omitempty"`
```
shurcooL: Based on #180 and https://github.com/google/go-github/pull/520#discussion_r98745029 (/cc @gmlewis), I think we'd want to go with `[]*Commit` here. Even though it's a little inconsistent with other similar structs.

At least that's what I think @gmlewis would request.

amirilovic: Got it, will fix.

gmlewis:
As part of this cleanup, can you please add `HTMLURL` to `type Commit struct` in `git_commits.go` (I cam across this while looking at the `parents` response)?

Also, in `search.go` can you please rename `Total` in `type CommitsSearchResult struct` to `TotalCount` for consistency?

Since these are breaking API changes, we also need to bump the version number.

Thank you!

amirilovic:
@gmlewis All fine for TotalCount.

For HTMLURL in Commit, I think it's fine right now. 
If you take a look at https://developer.github.com/v3/search/#search-commits results are represented as CommitResult which does have HTMLURL

```
type CommitResult struct {
       ...
        Commit      *Commit  `json:"commit,omitempty"`
	HTMLURL     *string     `json:"html_url,omitempty"`
	....
}
```

Also for RepositoryCommit as a result of https://developer.github.com/v3/repos/commits/#get-a-single-commit

```
type RepositoryCommit struct {
	...
        Commit      *Commit  `json:"commit,omitempty"`
        HTMLURL     *string  `json:"html_url,omitempty"`
	...
}
```

So, as I understand both result structures are wrappers around a commit with a specific HTMLURL, this is inline with github docs.

Also parents in get single commit don't have html_url
```
...
"parents": [
    {
      "url": "https://api.github.com/repos/octocat/Hello-World/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e",
      "sha": "6dcb09b5b57875f334f61aebed695e2e4193db5e"
    }
  ],
...
```

As far as I see in docs for other endpoints parents array is always just a reference to parent commits it never contains actual commit data, so maybe the best thing would be to create new struct to that represents that like:

```
type CommitReference {
        SHA *string  `json:"sha,omitempty"`
        URL *string  `json:"url,omitempty"`
        HTMLURL *string  `json:"html_url,omitempty"`
}
```
and change RepositoryResult and CommitResult to be:
```
type RepositoryCommit struct {
        ...
	Parents     []CommitReference `json:"parents,omitempty"`
        ...
}

type CommitResult struct {
        ...
	Parents     []CommitReference `json:"parents,omitempty"`
        ...
}
```

What do you think?

shurcooL:
> Also, in `search.go` can you please rename `Total` in `type CommitsSearchResult struct` to `TotalCount` for consistency?

@gmlewis, I don't think we should do this. At least not in this PR. It makes things inconsistent. Look at the rest of `search.go`:

```Go
// RepositoriesSearchResult represents the result of a repositories search.
type RepositoriesSearchResult struct {
	Total             *int         `json:"total_count,omitempty"`
	IncompleteResults *bool        `json:"incomplete_results,omitempty"`
	Repositories      []Repository `json:"items,omitempty"`
}

// IssuesSearchResult represents the result of an issues search.
type IssuesSearchResult struct {
	Total             *int    `json:"total_count,omitempty"`
	IncompleteResults *bool   `json:"incomplete_results,omitempty"`
	Issues            []Issue `json:"items,omitempty"`
}

// UsersSearchResult represents the result of a users search.
type UsersSearchResult struct {
	Total             *int   `json:"total_count,omitempty"`
	IncompleteResults *bool  `json:"incomplete_results,omitempty"`
	Users             []User `json:"items,omitempty"`
}
```

Do you see how it'd be inconsistent to change `Total` to `TotalCount` only in `CommitsSearchResult`?

```Go
// CommitsSearchResult represents the result of a commits search.
type CommitsSearchResult struct {
	TotalCount        *int            `json:"total_count,omitempty"`
	IncompleteResults *bool           `json:"incomplete_results,omitempty"`
	Commits           []*CommitResult `json:"items,omitempty"`
}

// CodeSearchResult represents the result of a code search.
type CodeSearchResult struct {
	Total             *int         `json:"total_count,omitempty"`
	IncompleteResults *bool        `json:"incomplete_results,omitempty"`
	CodeResults       []CodeResult `json:"items,omitempty"`
}
```

We've discussed this at https://github.com/google/go-github/pull/520#discussion_r98727959 previously.

@gmlewis Can you open an issue about this instead, so we can come to a decision and address all similar structs at once? But let's keep it out of scope of this PR.

@amirilovic Please revert `TotalCount` to `Total` for now.

gmlewis:
OK, SGTM.
Opened: https://github.com/google/go-github/issues/605
Sorry for the trouble, @amirilovic.

shurcooL:
About `HTMLURL` in `Commit`, I agree with @gmlewis that we should add it there.

@amirilovic, it's pretty subtle as to why.

If you look at our https://godoc.org/github.com/google/go-github/github#GitService.GetCommit method, it returns `Commit`. It maps to https://developer.github.com/v3/git/commits/#get-a-commit GitHub API endpoint, and its example output does not contain a "html_url" field:

```JSON
{
  "sha": "7638417db6d59f3c431d3e1f261cc637155684cd",
  "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/7638417db6d59f3c431d3e1f261cc637155684cd",
  "author": {
    "date": "2014-11-07T22:01:45Z",
    "name": "Scott Chacon",
    "email": "schacon@gmail.com"
  },
  "committer": {
    "date": "2014-11-07T22:01:45Z",
    "name": "Scott Chacon",
    "email": "schacon@gmail.com"
  },
  "message": "added readme, because im a good github citizen",
  "tree": {
    "url": "https://api.github.com/repos/octocat/Hello-World/git/trees/691272480426f78a0138979dd3ce63b77f706feb",
    "sha": "691272480426f78a0138979dd3ce63b77f706feb"
  },
  "parents": [
    {
      "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/1acc419d4d6a9ce985db7be48c6349a0475975b5",
      "sha": "1acc419d4d6a9ce985db7be48c6349a0475975b5"
    }
  ]
}
```

However, GitHub API documentation is not always 100% up to date and does not always show all fields that the real API returns. This is an example of that.

If you hit that endpoint for real, it includes "html_url" field. For example, try:

https://api.github.com/repos/octocat/Hello-World/git/commits/7fd1a60b01f91b314f59955a4e4d4e80d8edf11d

```JSON
{
  "sha": "7fd1a60b01f91b314f59955a4e4d4e80d8edf11d",
  "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/7fd1a60b01f91b314f59955a4e4d4e80d8edf11d",
  "html_url": "https://github.com/octocat/Hello-World/commit/7fd1a60b01f91b314f59955a4e4d4e80d8edf11d",
  "author": {
    "name": "The Octocat",
    "email": "octocat@nowhere.com",
    "date": "2012-03-06T23:06:50Z"
  },
  "committer": {
    "name": "The Octocat",
    "email": "octocat@nowhere.com",
    "date": "2012-03-06T23:06:50Z"
  },
  "tree": {
    "sha": "b4eecafa9be2f2006ce1b709d6857b07069b4608",
    "url": "https://api.github.com/repos/octocat/Hello-World/git/trees/b4eecafa9be2f2006ce1b709d6857b07069b4608"
  },
  "message": "Merge pull request #6 from Spaceghost/patch-1\n\nNew line at end of file.",
  "parents": [
    {
      "sha": "553c2077f0edc3d5dc5d17262f6aa498e69d6f8e",
      "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/553c2077f0edc3d5dc5d17262f6aa498e69d6f8e",
      "html_url": "https://github.com/octocat/Hello-World/commit/553c2077f0edc3d5dc5d17262f6aa498e69d6f8e"
    },
    {
      "sha": "762941318ee16e59dabbacb1b4049eec22f0d303",
      "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/762941318ee16e59dabbacb1b4049eec22f0d303",
      "html_url": "https://github.com/octocat/Hello-World/commit/762941318ee16e59dabbacb1b4049eec22f0d303"
    }
  ]
}
```

Also note that parents have "html_url" field too.

So, `Commit` should have `HTMLURL`.

They've probably added it more recently, and didn't update their old documentation examples.

amirilovic:
@shurcooL Good point.

github/git_commits.go
```diff
@@ -22,6 +22,7 @@ type SignatureVerification struct {
 // Commit represents a GitHub commit.
 type Commit struct {
 	SHA          *string                `json:"sha,omitempty"`
+	HTMLURL      *string                `json:"html_url,omitempty"`
```
shurcooL: Move it near `URL`, that should be a more fitting place for it.

See https://github.com/google/go-github/blob/https://github.com/google/go-github/blob/master/github/repos_commits.go#L24/github/repos_commits.go#L24.

amirilovic:
@shurcooL Anything else that needs to be done here?

github/search.go
```diff
@@ -65,17 +65,16 @@ type CommitsSearchResult struct {
...
+	SHA         *string     `json:"sha,omitempty"`
+	Commit      *Commit     `json:"commit,omitempty"`
+	Author      *User       `json:"author,omitempty"`
+	Committer   *User       `json:"committer,omitempty"`
+	Parents     []*Commit   `json:"parents,omitempty"`
+	HTMLURL     *string     `json:"html_url,omitempty"`
+	URL         *string     `json:"url,omitempty"`
+	CommentsURL *string     `json:"comments_url,omitempty"`
+	Repository  *Repository `json:"repository,omitempty"`
+	Score       *float64    `json:"score,omitempty"`
```
shurcooL: As an observation, `CommitResult` now looks very similar to [RepositoryCommit](https://godoc.org/github.com/google/go-github/github#RepositoryCommit). But it differs in the last 2 fields. `RepositoryCommit` has `Stats` and `Files`, which `CommitResult` does not, and `CommitResult` has `Repository` and `Score`, that `RepositoryCommit` doesn't.

That's completely fine, just pointing out.

However, I think it'd be nice to add a blank line between `CommentsURL` field and `Repository` field, to separate the common vs differing fields, similar to the blank line separator that `RepositoryCommit` already has.

gmlewis:
Merging.

Issue #588, opened 2017-03-10
Github API for GetRef can return an array, but go-github doesn't handle it.


samprakos:
https://developer.github.com/v3/git/refs/#get-a-reference

If my ref string matches as a prefix to 1 or more refs, an array is returned.

In this case, I get this error at line 63 of git_refs.go:  json: cannot unmarshal array into Go value of type github.Reference

shurcooL:
Thanks for reporting this. This looks like a valid issue that we need to fix.

Unfortunately, it seems we'd need to change the signature of that endpoint to be able to return multiple refs, since that's what the GitHub API endpoint can sometimes return.

aleksclark:
I'm taking a look at this issue.

aleksclark:
@shurcooL / @gmlewis  just noticing that people generally ask to have issues assigned instead of just submitting PRs, could I be assigned to this?

gmlewis:
Sure, I'll send you an invite which you will need to accept before I can assign you.

aleksclark:
@gmlewis done!

aleksclark:
@shurcooL What if instead of breaking the api, we mark GetRef as deprecated, and add the new code as GetRefs, which would more accurately reflect the nature of the returned value?

shurcooL:
@aleksclark That's actually a pretty great idea. We're lucky that the current naming allows for it.

In fact, I'm thinking perhaps we don't need to deprecate the current method. It depends on how people use them, but I suspect most people only care about a single result, and `GetRef` is easier to work with since it returns a single result rather than a slice containing a single entry. If that's the case, we can keep signature of `GetRef` as is, and people can keep using it if they're happy with at most 1 result.

Then `GetRefs` can be added with ability to return multiple refs.

One thing, I think the current `GetRef` gives an error if GitHub API returns multiple entries, so we should fix that.

From https://developer.github.com/v3/git/refs/#get-a-reference:

> If the `ref` doesn't exist in the repository, but existing refs start with `ref` they will be returned as an array. For example, a call to get the data for a branch named `feature`, which doesn't exist, would return head refs including `featureA` and `featureB` which do.

So, if we're going to keep `GetRef` and not deprecate it, I think the best behavior is to make it return `nil` if there's no exact match (and GitHub API returns multiple similar results).

[...]

PR #603, opened 2017-03-28
#588 Handle multiple References returned by the GH refs/ endpoint.

aleksclark:
This (breaking) change alters the api of GetRef() to return a slice of
References instead of a single Reference, in order to account for the
fact that GH will return an array of references in the event that
there is not a precise match to the queried reference.

This is accomplished by passing a buffer to client.Do(), and making a
second attempt at unmarshalling in the event that the single-object
case fails.

Addresses issue #588

### DIFF
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -50,6 +50,8 @@ type updateRefRequest struct {
 // GetRef fetches a single Reference object for a given Git ref.
 // If there is no exact match, GetRef will return an error.
 //
+// Note: the GitHub API can return multiple matches, if you wish to use this functionality please use the GetRefs() method.
+//
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
 func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref string) (*Reference, *Response, error) {
 	ref = strings.TrimPrefix(ref, "refs/")
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -61,7 +61,6 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref 
 
 	r := new(Reference)
 	resp, err := s.client.Do(ctx, req, r)
-
 	if _, ok := err.(*json.UnmarshalTypeError); ok {
 		// Multiple refs, means there wasn't an exact match.
 		return nil, resp, nil
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 )
@@ -61,7 +62,8 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref 
 	r := new(Reference)
 	resp, err := s.client.Do(ctx, req, r)
 	if _, ok := err.(*json.UnmarshalTypeError); ok {
-		return nil, resp, fmt.Errorf("No exact match found for this ref")
+		// Multiple refs, means there wasn't an exact match.
+		return nil, resp, errors.New("no exact match found for this ref")
 	} else if err != nil {
 		return nil, resp, err
 	}
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -76,9 +76,9 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref 
 // If returned error is nil, there will be at least 1 ref returned.
 // For example:
 //
-// 	"heads/featureA"  -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
-// 	"heads/feature"   -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
-// 	"heads/notexist"  -> []                                              // Returns an error.
+// 	"heads/featureA" -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
+// 	"heads/feature"  -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
+// 	"heads/notexist" -> []                                              // Returns an error.
 //
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
 func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref string) ([]*Reference, *Response, error) {
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -78,7 +78,7 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref 
 //
 // 	"heads/featureA"  -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
 // 	"heads/feature"   -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
-// 	"heads/notexist"  -> []                                             // Returns an error.
+// 	"heads/notexist"  -> []                                              // Returns an error.
 //
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
 func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref string) ([]*Reference, *Response, error) {
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -73,10 +73,12 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref 
 // GetRefs fetches a slice of Reference objects for a given Git ref.
 // If there is an exact match, only that ref is returned.
 // If there is no exact match, GitHub returns all refs that start with ref.
+// If returned error is nil, there will be at least 1 ref returned.
 // For example:
 //
 // 	"heads/featureA" -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
 // 	"heads/feature"  -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
+// 	"heads/feature"  -> []                                              // Returns an error.
 //
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
 func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref string) ([]*Reference, *Response, error) {
@@ -104,6 +106,9 @@ func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref
 	var rs []*Reference
 	multipleUnmarshalError := json.Unmarshal(rawJSON, &rs)
 	if multipleUnmarshalError == nil {
+		if len(rs) == 0 {
+			return nil, resp, fmt.Errorf("unexpected response from GitHub API: an array of refs with length 0")
+		}
 		return rs, resp, nil
 	}
 
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -47,12 +47,49 @@ type updateRefRequest struct {
 	Force *bool   `json:"force"`
 }
 
-// GetRef fetches a slice of Reference objects for a given Git ref.
-// If a Git ref does not match exactly, GitHub will return all refs that match it as a prefix
-// E.g. feature -> [featureA, featureB]
+// GetRef fetches a single Reference object for a given Git ref.
+// If a Git ref does not match exactly, this method will return a nil result.
 //
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
-func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref string) ([]*Reference, *Response, error) {
+func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref string) (*Reference, *Response, error) {
+	ref = strings.TrimPrefix(ref, "refs/")
+	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, ref)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	buf := new(bytes.Buffer)
+	resp, err := s.client.Do(ctx, req, buf)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	rbuf := bytes.NewBuffer(buf.Bytes())
+	r := new(Reference)
+
+	err = json.NewDecoder(rbuf).Decode(r)
+	if err != nil {
+		_, ok := err.(*json.UnmarshalTypeError)
+		if ok {
+			// multiple responses
+			return nil, resp, nil
+		} else {
+			return nil, resp, err
+		}
+	}
+
+	return r, resp, nil
+}
+
+// GetRefs fetches a slice of Reference objects for a given Git ref.
+// If a Git ref does not match exactly, GitHub will return all refs that match it as a prefix.
+// Example:
+// feature -> [featureA, featureB]
+// featureA -> [featureA]
+//
+// GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
+func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref string) ([]*Reference, *Response, error) {
 	ref = strings.TrimPrefix(ref, "refs/")
 	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, ref)
 	req, err := s.client.NewRequest("GET", u, nil)
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -59,15 +59,9 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref 
 		return nil, nil, err
 	}
 
-	buf := new(bytes.Buffer)
-	resp, err := s.client.Do(ctx, req, buf)
-	if err != nil {
-		return nil, resp, err
-	}
-
 	r := new(Reference)
+	resp, err := s.client.Do(ctx, req, r)
 
-	err = json.NewDecoder(buf).Decode(r)
 	if _, ok := err.(*json.UnmarshalTypeError); ok {
 		// Multiple refs, means there wasn't an exact match.
 		return nil, resp, nil
@@ -99,7 +93,7 @@ func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref
 		return nil, resp, err
 	}
 
-	rbuf := bytes.NewBuffer(buf.Bytes())
+	rbuf := bytes.NewReader(buf.Bytes())
 
 	var rs []*Reference
 	r := new(Reference)
@@ -109,10 +103,10 @@ func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref
 		rs = append(rs, r)
 	} else if _, ok := err.(*json.UnmarshalTypeError); ok {
 		// check for multiple refs returned
-		rbuf = bytes.NewBuffer(buf.Bytes())
-		serr := json.NewDecoder(rbuf).Decode(&rs)
-		if serr != nil {
-			return nil, resp, serr
+		rbuf = bytes.NewReader(buf.Bytes())
+		err := json.NewDecoder(rbuf).Decode(&rs)
+		if err != nil {
+			return nil, resp, err
 		}
 	} else {
 		return nil, resp, err
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -6,7 +6,6 @@
 package github
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -86,32 +85,26 @@ func (s *GitService) GetRefs(ctx context.Context, owner string, repo string, ref
 		return nil, nil, err
 	}
 
-	buf := new(bytes.Buffer)
-	resp, err := s.client.Do(ctx, req, buf)
+	var rawJSON json.RawMessage
+	resp, err := s.client.Do(ctx, req, &rawJSON)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	rbuf := bytes.NewReader(buf.Bytes())
-
 	var rs []*Reference
 	r := new(Reference)
 	// prioritize the most common case: a single returned ref
-	err = json.NewDecoder(rbuf).Decode(r)
-	if err == nil {
-		rs = append(rs, r)
-	} else if _, ok := err.(*json.UnmarshalTypeError); ok {
-		// check for multiple refs returned
-		rbuf = bytes.NewReader(buf.Bytes())
-		err := json.NewDecoder(rbuf).Decode(&rs)
-		if err != nil {
-			return nil, resp, err
-		}
-	} else {
-		return nil, resp, err
+	singleUnmarshalErr := json.Unmarshal(rawJSON, r)
+	if singleUnmarshalErr == nil {
+		return append(rs, r), resp, nil
 	}
 
-	return rs, resp, nil
+	multipleUnmarshalErr := json.Unmarshal(rawJSON, &rs)
+	if multipleUnmarshalErr == nil {
+		return rs, resp, nil
+	}
+
+	return nil, resp, fmt.Errorf("unmarshalling failed for both single and multiple responses: %s and %s ", singleUnmarshalErr, multipleUnmarshalErr)
 }
 
 // ReferenceListOptions specifies optional parameters to the
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -6,7 +6,9 @@
 package github
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 )
@@ -45,10 +47,12 @@ type updateRefRequest struct {
 	Force *bool   `json:"force"`
 }
 
-// GetRef fetches the Reference object for a given Git ref.
+// GetRef fetches a slice of Reference objects for a given Git ref.
+// If a Git ref does not match exactly, GitHub will return all refs that match it as a prefix
+// E.g. feature -> [featureA, featureB]
 //
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
-func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref string) (*Reference, *Response, error) {
+func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref string) ([]*Reference, *Response, error) {
 	ref = strings.TrimPrefix(ref, "refs/")
 	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, ref)
 	req, err := s.client.NewRequest("GET", u, nil)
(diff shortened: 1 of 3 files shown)


aleksclark:
@shurcooL this is ready for review I think, tests and everything :)

github/git_refs.go
```diff
@@ -56,15 +59,75 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	err = json.NewDecoder(rbuf).Decode(r)
+	if err != nil {
+		_, ok := err.(*json.UnmarshalTypeError)
+		if ok {
+			// multiple responses
+			return nil, resp, nil
+		} else {
+			return nil, resp, err
+		}
+	}
```
shurcooL: I think this can be cleaner if we reduce the indentation by one level:

```Go
err = json.NewDecoder(rbuf).Decode(r)
if _, ok := err.(*json.UnmarshalTypeError); ok {
	// Multiple refs, means there wasn't an exact match.
	return nil, resp, nil
} else if err != nil {
	return nil, resp, err
}
```

Similar can be applied below.

aleksclark:
@shurcooL got it, thanks. I somehow have the feeling that it's not idiomatic in go to check for `err == nil` because 99% of the time it's:
```
if err != nil {
    return err
}
```

but in this case it feels optimal to handle the happy path(s) first, in case of `GetRefs`. Thoughts?

github/git_refs.go
```diff
@@ -56,15 +59,69 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
 		return nil, nil, err
 	}
 
-	r := new(Reference)
-	resp, err := s.client.Do(ctx, req, r)
+	buf := new(bytes.Buffer)
```
shurcooL: Given you're not trying to parse the input twice in `GetRef`, this change to use `bytes.Buffer` seems unnecessary.

shurcooL: To clarify, I meant to ask if you can keep the code as before, have it `r := new(Reference); resp, err := s.client.Do(ctx, req, r)`? Or does that obscure the `*json.UnmarshalTypeError`?

github/git_refs.go
```diff
@@ -56,15 +59,68 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+
+	var rs []*Reference
+	r := new(Reference)
+	// prioritize the most common case: a single returned ref
+	err = json.NewDecoder(rbuf).Decode(r)
+	if err == nil {
+		rs = append(rs, r)
+	} else if _, ok := err.(*json.UnmarshalTypeError); ok {
+		// check for multiple refs returned
+		rbuf = bytes.NewBuffer(buf.Bytes())
```
shurcooL: Using `bytes.NewReader` is less expensive than buffer and will suffice for the needs of `json.NewDecoder`.

aleksclark:
@shurcooL ok, performed requested changes, although I'm getting the feeling that you're spending more time reviewing than it would have taken you to fix it in the first place :)

github/git_refs.go
```diff
@@ -58,13 +61,60 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
 
 	r := new(Reference)
 	resp, err := s.client.Do(ctx, req, r)
-	if err != nil {
+
```
shurcooL: No need for a blank line here. It makes it harder to see that the `err` value is being handled on next line.

github/git_refs.go
```diff
@@ -58,13 +61,60 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+		rbuf = bytes.NewReader(buf.Bytes())
+		err := json.NewDecoder(rbuf).Decode(&rs)
+		if err != nil {
+			return nil, resp, err
+		}
+	} else {
+		return nil, resp, err
+	}
+
+	return rs, resp, nil
```
shurcooL: @aleksclark, I think we can do all of the above in a cleaner and more readable way by making use of `json.RawMessage`.

Take a look at https://developer.github.com/v3/repos/contents/#get-contents, it's another GitHub API endpoint that can return either an array of objects, or a single object. This is how it's implemented in our code:

https://github.com/google/go-github/blob/d705ac853513f240ee84a8e4772b7126498f9591/github/repos_contents.go#L154-L167

That seems like a better way to do this, do you agree? If so, you can update your code to follow that pattern.

aleksclark: yes, that is much cleaner, and I like the preservation of the unmarshalling errors. code has been updated.

github/git_refs.go
```diff
@@ -45,7 +46,8 @@ type updateRefRequest struct {
 	Force *bool   `json:"force"`
 }
 
-// GetRef fetches the Reference object for a given Git ref.
+// GetRef fetches a single Reference object for a given Git ref.
+// If a Git ref does not match exactly, this method will return a nil result.
```
shurcooL: I would rephrase this slightly:

"If a Git ref does not match exactly" -> "If there is no exact match"
"this method" -> "GetRef"
"nil result" -> "nil Reference"

So:

```
// If there is no exact match, GetRef will return a nil Reference.
```

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
 	return r, resp, nil
 }
 
+// GetRefs fetches a slice of Reference objects for a given Git ref.
+// If a Git ref does not match exactly, GitHub will return all refs that match it as a prefix.
+// Example:
+// feature -> [featureA, featureB]
+// featureA -> [featureA]
+//
+// GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
```
shurcooL: We should tweak this comment's formatting. Currently, it will not look good in godoc:

image

I would suggest formatting it somewhat like this:

```
// GetRefs fetches a slice of Reference objects for a given Git ref.
// If there is an exact match, only that ref is returned.
// If there is no exact match, GitHub returns all refs that start with ref.
// For example:
//
// 	"heads/featureA" -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
// 	"heads/feature"  -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
//
// GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
```

Which looks like this:

image

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+		return nil, nil, err
+	}
+
+	var rawJSON json.RawMessage
+	resp, err := s.client.Do(ctx, req, &rawJSON)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	var rs []*Reference
```
shurcooL: A nice thing to do is to declare variables just before they're needed, not sooner. We can get away with declaring `rs` much later, or not at all. Remove it here.

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	r := new(Reference)
+	// prioritize the most common case: a single returned ref
```
shurcooL: Go style suggests using proper english sentences, unless the comment is a short phrase. So capitalize the first letter and add a period to this comment. I'd also put it on top of `r := new(Reference)`, so:

```Go
// Prioritize the most common case: a single returned ref.
r := new(Reference)
singleUnmarshalErr := json.Unmarshal(rawJSON, r)
```

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	singleUnmarshalErr := json.Unmarshal(rawJSON, r)
+	if singleUnmarshalErr == nil {
+		return append(rs, r), resp, nil
```
shurcooL: `return []*Reference{r}, resp, nil` is simpler.

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	}
+
+	multipleUnmarshalErr := json.Unmarshal(rawJSON, &rs)
```
shurcooL: Move `var rs []*Reference` on top of this line.

Also rename `multipleUnmarshalErr` to `multipleUnmarshalError` for same reason as above.

So:

```Go
var rs []*Reference
multipleUnmarshalError := json.Unmarshal(rawJSON, &rs)
```

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	var rawJSON json.RawMessage
+	resp, err := s.client.Do(ctx, req, &rawJSON)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	var rs []*Reference
+	r := new(Reference)
+	// prioritize the most common case: a single returned ref
+	singleUnmarshalErr := json.Unmarshal(rawJSON, r)
```
shurcooL: For consistency, call this variable `singleUnmarshalError`. It's only shortened to `err` when there is no prefix. That's what it's called in `RepositoriesService.GetContents` too.

See https://dmitri.shuralyov.com/idiomatic-go#error-variable-naming for motivation.

github/git_refs.go
```diff
@@ -58,13 +60,53 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	if singleUnmarshalErr == nil {
+		return append(rs, r), resp, nil
+	}
+
+	multipleUnmarshalErr := json.Unmarshal(rawJSON, &rs)
+	if multipleUnmarshalErr == nil {
+		return rs, resp, nil
+	}
+
+	return nil, resp, fmt.Errorf("unmarshalling failed for both single and multiple responses: %s and %s ", singleUnmarshalErr, multipleUnmarshalErr)
```
shurcooL: "single and multiple responses" -> "single and multiple refs". 'Responses' doesn't make sense, there's only one response.

Get rid of the trailing space here. If you don't mind, also get rid of it in `RepositoriesService.GetContents` (it's there too).

github/git_refs_test.go
```diff
@@ -14,7 +14,7 @@ import (
 	"testing"
 )
 
-func TestGitService_GetRef(t *testing.T) {
+func TestGitService_GetRef_SingleResult(t *testing.T) {
```
shurcooL: The naming convention we apply for tests (it's based on the convention examples) is to begin the test label – "SingleResult" in this case – with lower case.

For example, see https://github.com/google/go-github/blob/e96a1efec3efbfec63cf32984b21bf98933042a2/github/pulls_comments_test.go#L18, https://github.com/google/go-github/blob/e96a1efec3efbfec63cf32984b21bf98933042a2/github/pulls_comments_test.go#L51.

Source: https://godoc.org/testing#hdr-Examples.

So, this should be renamed to `TestGitService_GetRef_singleResult`.

However, let's be more specific and say it's single ref, rather than single result.

So, `TestGitService_GetRef_singleRef`.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
 	}
 }
 
+func TestGitService_GetRef_MultipleResult(t *testing.T) {
```
shurcooL: `TestGitService_GetRef_multipleRefs` for same reason as above.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+		t.Errorf("Git.GetRef returned error: %v", err)
+	}
+
+	if ref != nil {
+		t.Errorf("Git.GetRef returned %+v, want nil", ref)
+	}
+
+}
+
+func TestGitService_GetRefs_SingleResult(t *testing.T) {
```
shurcooL: `TestGitService_GetRefs_singleRef` for same reason as above.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+		t.Errorf("Git.GetRef returned %+v, want %+v", ref, want)
+	}
+
+	// without 'refs/' prefix
+	if _, _, err := client.Git.GetRefs(context.Background(), "o", "r", "heads/b"); err != nil {
+		t.Errorf("Git.GetRef returned error: %v", err)
+	}
+}
+
+func TestGitService_GetRefs_MultipleResult(t *testing.T) {
```
shurcooL: `TestGitService_GetRefs_multipleRefs` for same reason as above.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+		      "type": "commit",
+		      "sha": "aa218f56b14c9653891f9e74264a383fa43fefbd",
+		      "url": "https://api.github.com/repos/o/r/git/commits/aa218f56b14c9653891f9e74264a383fa43fefbd"
+		    }
+		  }`)
+	})
+
+	refs, _, err := client.Git.GetRefs(context.Background(), "o", "r", "refs/heads/b")
+	if err != nil {
+		t.Fatalf("Git.GetRef returned error: %v", err)
```
shurcooL: `Git.GetRefs`.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+		Ref: String("refs/heads/b"),
+		URL: String("https://api.github.com/repos/o/r/git/refs/heads/b"),
+		Object: &GitObject{
+			Type: String("commit"),
+			SHA:  String("aa218f56b14c9653891f9e74264a383fa43fefbd"),
+			URL:  String("https://api.github.com/repos/o/r/git/commits/aa218f56b14c9653891f9e74264a383fa43fefbd"),
+		},
+	}
+	if !reflect.DeepEqual(ref, want) {
+		t.Errorf("Git.GetRef returned %+v, want %+v", ref, want)
```
shurcooL: `Git.GetRefs`.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+	}
+
+	// without 'refs/' prefix
+	if _, _, err := client.Git.GetRefs(context.Background(), "o", "r", "heads/b"); err != nil {
+		t.Errorf("Git.GetRef returned error: %v", err)
```
shurcooL: `Git.GetRefs`.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+		        "url": "https://api.github.com/repos/o/r/git/commits/612077ae6dffb4d2fbd8ce0cccaa58893b07b5ac"
+		      }
+		    }
+		  ]
+		`)
+	})
+
+	refs, _, err := client.Git.GetRefs(context.Background(), "o", "r", "refs/heads/b")
+	if err != nil {
+		t.Errorf("Git.GetRef returned error: %v", err)
```
shurcooL: `Git.GetRefs`.

github/git_refs_test.go
```diff
@@ -56,6 +56,139 @@ func TestGitService_GetRef(t *testing.T) {
...
+		Ref: String("refs/heads/booger"),
+		URL: String("https://api.github.com/repos/o/r/git/refs/heads/booger"),
+		Object: &GitObject{
+			Type: String("commit"),
+			SHA:  String("aa218f56b14c9653891f9e74264a383fa43fefbd"),
+			URL:  String("https://api.github.com/repos/o/r/git/commits/aa218f56b14c9653891f9e74264a383fa43fefbd"),
+		},
+	}
+	if !reflect.DeepEqual(refs[0], want) {
+		t.Errorf("Git.GetRef returned %+v, want %+v", refs[0], want)
```
shurcooL: `Git.GetRefs`.

shurcooL:
@aleksclark Do you want to apply those changes yourself? If you want, I can apply them when merging this PR. It's up to you.

aleksclark:
@shurcooL I'll do it - it's good penance :)

shurcooL:
All right.

To be clear, you did nothing wrong, these are just opportunities to improve this already good PR even further that I was able to spot. :) Thanks!

github/git_refs.go
```diff
@@ -58,13 +60,56 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	} else if err != nil {
 		return nil, resp, err
 	}
 
 	return r, resp, nil
 }
 
+// GetRefs fetches a slice of Reference objects for a given Git ref.
+// If there is an exact match, only that ref is returned.
+// If there is no exact match, GitHub returns all refs that start with ref.
```
shurcooL: From my testing, when there are multiple matches returned, there's always at least 1 entry, otherwise it's a 404 not found error.

I think we can take advantage of that to make the API friendlier. We can make a guarantee that if returned `error` is `nil`, then there's at least one match, making it safe to access `refs[0]`.

I think doing that would make it easier for people to use `GetRefs`. What are your thoughts on this?

I'm imagining something like:

```
// If there is an exact match, only that ref is returned.
// If there is no exact match, GitHub returns all refs that start with ref.
// If returned error is nil, there will be at least 1 ref returned.
```

Or is this not worth doing?

github/git_refs.go
```diff
@@ -58,13 +60,56 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	r := new(Reference)
+	singleUnmarshalError := json.Unmarshal(rawJSON, r)
+	if singleUnmarshalError == nil {
+		return []*Reference{r}, resp, nil
+	}
+
+	// Attempt to unmarshal multiple refs.
+	var rs []*Reference
+	multipleUnmarshalError := json.Unmarshal(rawJSON, &rs)
+	if multipleUnmarshalError == nil {
```
shurcooL: If we make the documentation change above, I think we'll need to be defensive about it. While currently the GitHub API seems to always return 404 not found if there are 0 matching refs, should that ever change or not be true, I think we'd want to catch that and prevent a Go program from panicking. So, we add a run-time check here:

```Go
if len(rs) == 0 {
    return nil, resp, fmt.Errorf("unexpected response from GitHub API: an array of refs with length 0")
}
```

That way we can guarantee we uphold the contract stated in the godoc.

github/git_refs.go
```diff
@@ -58,13 +60,61 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
 
+// GetRefs fetches a slice of Reference objects for a given Git ref.
+// If there is an exact match, only that ref is returned.
+// If there is no exact match, GitHub returns all refs that start with ref.
+// If returned error is nil, there will be at least 1 ref returned.
+// For example:
+//
+// 	"heads/featureA" -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
+// 	"heads/feature"  -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
+// 	"heads/feature"  -> []                                              // Returns an error.
```
shurcooL: Maybe rename the 3rd entry to "heads/notexist" or something. Otherwise, it's confusing how "heads/feature" returns 2 branches on second line, but 0 on third.

github/git_refs_test.go
```diff
@@ -56,6 +56,154 @@ func TestGitService_GetRef(t *testing.T) {
...
+	}
+}
+
+func TestGitService_GetRefs_noRefs(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/git/refs/heads/b", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, "[]")
```
shurcooL: This tests the handling of an unexpected response from GitHub API, is that what you intended?

aleksclark: Yes, I think that if we're going to explicitly handle this response, we should explicitly test it.

shurcooL: Sounds good. My only request is that you document that this test is for an _unexpected_ response from GitHub. It should never actually happen.

My concern is that if we don't document that, it will be confusing later.

You could also check that the error string is an exact match with "unexpected response from GitHub API: an array of refs with length 0". That way it ties the test closer with the exact thing it's testing.

github/git_refs.go
```diff
@@ -58,13 +60,61 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
 
+// GetRefs fetches a slice of Reference objects for a given Git ref.
+// If there is an exact match, only that ref is returned.
+// If there is no exact match, GitHub returns all refs that start with ref.
+// If returned error is nil, there will be at least 1 ref returned.
+// For example:
+//
+// 	"heads/featureA"  -> ["refs/heads/featureA"]                         // Exact match, single ref is returned.
+// 	"heads/feature"   -> ["refs/heads/featureA", "refs/heads/featureB"]  // All refs that start with ref.
+// 	"heads/notexist"  -> []                                             // Returns an error.
```
shurcooL: Nitpick, the 3rd comment `// Returns an error.` isn't aligned with the other two comments above it.

github/git_refs.go
```diff
@@ -58,13 +60,61 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	r := new(Reference)
+	singleUnmarshalError := json.Unmarshal(rawJSON, r)
+	if singleUnmarshalError == nil {
+		return []*Reference{r}, resp, nil
+	}
+
+	// Attempt to unmarshal multiple refs.
+	var rs []*Reference
+	multipleUnmarshalError := json.Unmarshal(rawJSON, &rs)
+	if multipleUnmarshalError == nil {
```
gmlewis: nit: It is more common Go style to end up the function with the no-error case. I think it is also easier to read as it reduces the indentation levels.

So this would be:

```go
if multipleUnmarshalError != nil {
	return nil, resp, fmt.Errorf("unmarshalling failed for both single and multiple refs: %s and %s", singleUnmarshalError, multipleUnmarshalError)
}
if len(rs) == 0 {
	return nil, resp, fmt.Errorf("unexpected response from GitHub API: an array of refs with length 0")
}

return rs, resp, nil
```

shurcooL: @gmlewis I don't think this applies here, it's a special case. We don't know what the success path is going to be until we try two paths and see if one of them succeeds.

This is code we iterated to from something that was IMO less readable, because of much higher levels of indentation. See the discussion in the PR above, and https://github.com/google/go-github/pull/603#discussion_r109051657.

Also, this is consistent with our code at https://github.com/google/go-github/blob/d705ac853513f240ee84a8e4772b7126498f9591/github/repos_contents.go#L154-L167.

shurcooL:
So, this turned out not to be a breaking API change, so I won't have to bump library version when merging.

I will fix a minor spacing nit in a comment.

Thanks for awesome work on this PR @aleksclark! Nice job coming up with the idea of non-breaking API solution in https://github.com/google/go-github/issues/588#issuecomment-290406367, I think it worked out really well.

github/git_refs.go
```diff
@@ -45,7 +46,8 @@ type updateRefRequest struct {
 	Force *bool   `json:"force"`
 }
 
-// GetRef fetches the Reference object for a given Git ref.
+// GetRef fetches a single Reference object for a given Git ref.
+// If there is no exact match, GetRef will return a nil Reference.
```
shurcooL: To be more specific, I'm proposing we change this to:

```
// GetRef fetches a single Reference object for a given Git ref.
// If there is no exact match, an error is returned.
```

github/git_refs.go
```diff
@@ -61,8 +61,7 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
 	r := new(Reference)
 	resp, err := s.client.Do(ctx, req, r)
 	if _, ok := err.(*json.UnmarshalTypeError); ok {
-		// Multiple refs, means there wasn't an exact match.
-		return nil, resp, nil
+		return nil, resp, fmt.Errorf("No exact match found for this ref")
```
shurcooL: 1. Keep the "// Multiple refs, means there wasn't an exact match." comment. It explains that getting `*json.UnmarshalTypeError` error means we're getting multiple refs. Without the comment, it's not clear.

2. Go style says error strings should begin with a lower case letter. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings. So, it should be "no exact match found for this ref".

3. Since there's no dynamic formatting taking place, use `errors.New("no exact match found for this ref")` rater than `fmt.Errorf`.

github/git_refs_test.go
```diff
@@ -86,13 +86,10 @@ func TestGitService_GetRef_multipleRefs(t *testing.T) {
...
-	if err != nil {
-		t.Errorf("Git.GetRef returned error: %v", err)
-	}
-
-	if ref != nil {
-		t.Errorf("Git.GetRef returned %+v, want nil", ref)
+	_, _, err := client.Git.GetRef(context.Background(), "o", "r", "refs/heads/b")
+	want := "No exact match found for this ref"
+	if err.Error() != want {
+		t.Errorf("Git.GetRefs returned %+v, want %+v", err, want)
```
shurcooL: Git.GetRef, not Git.GetRefs.

github/git_refs.go
```diff
@@ -45,7 +47,8 @@ type updateRefRequest struct {
 	Force *bool   `json:"force"`
 }
 
-// GetRef fetches the Reference object for a given Git ref.
+// GetRef fetches a single Reference object for a given Git ref.
+// If there is no exact match, GetRef will return an error.
 //
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
```
shurcooL: > > Do you think maybe we should make the error say something about using GetRefs if you want to handle 0...N refs?
>
> No need. The documentation says that.

Well, I guess I lied about that. The docs doesn't really say that. So I agree it'd be an improvement to add a paragraph here, something that mentions that users should consider `GetRefs` if they want a method that is a more direct mapping to the GitHub API endpoint, if they want to be able to get multiple refs.

shurcooL:
@gmlewis, there have been some changes here since your last approval, do you mind looking them over?

github/git_refs.go
```diff
@@ -50,6 +50,8 @@ type updateRefRequest struct {
 // GetRef fetches a single Reference object for a given Git ref.
 // If there is no exact match, GetRef will return an error.
 //
+// Note: the GitHub API can return multiple matches, if you wish to use this functionality please use the GetRefs() method.
```
gmlewis: nit: I would suggest wrapping this to match the length of the other comments.

shurcooL: Also, "The" should be capitalized, since "Note:" is usually treated as a prefix to a complete sentence, rather than being considered a part of the sentence.

E.g., see https://github.com/golang/go/blob/f75b5e99ed19e97ea78e62a406d03c1632efa7d4/src/cmd/asm/internal/asm/parse.go#L795.

aleksclark:
@gmlewis @shurcooL does that look good to you guys?

shurcooL:
Thanks a lot @aleksclark for working on this!

PR #609, opened 2017-03-31
Support PushEventRepository.ArchiveURL

ns-cweber:
Added support for PushEventRepository.ArchiveURL per https://developer.github.com/v3/activity/events/types/#pushevent


ns-cweber:
I signed it!

shurcooL:
Sure. But I'll make it a separate commit, since this change doesn't break the API.

Issue #606, opened 2017-03-30
Update Protected Branches API preview with Admin Enforcement


gmlewis:
GitHub Announcement:
https://developer.github.com/changes/2017-03-22-protected-branches-admin-improvements/

This will be a breaking API change and the package version will need to be bumped.

amirilovic:
@gmlewis I can take this.

gmlewis:
Let's close this issue and continue any discussion related to Protected Branches API in #617.

Issue #537, opened 2017-01-31
Policy on using pointers/values in structs for receiving data from GitHub servers.


shurcooL:
I'm making this issue to discuss a point that came up while reviewing #534. /cc @alindeman

### Background

In #534, @alindeman answered my questions:

> > @shurcooL: What's the reason to use `string` value rather than `*string` pointer, as is commonly done in other structs?
> > @shurcooL: What's the reason to omit `,omitempty` option, contrary to most similar structs we have?
>
> @alindeman: Thinking back to the [conversation we had over in #512](https://github.com/google/go-github/pull/512#issuecomment-272350152), I thought that since--as far as I can tell--these attributes will always be returned from the API, we could specify them as non-pointer types.

One key difference is that the conversation in #512 was primarily about the request structs - structs that are used to send values to github servers. The affect on structs for receiving and unmarshaling data from github servers was inadvertent and discovered at the end, and we decided to go ahead with it because it was very minor. This PR is all about the structs for receiving from the start, that's why I want to think about it more carefully.

As I understood it up until now (before doing the following research), the main reason pointers are used for most receiving structs is to be able to tell when a field is absent vs when it's present but set to zero.

That's what I saw in #79. That links to #19 where the initial discussion took place, and 3072d069f278a43f79da41c2178194fdcc371070 is the implementation commit. However, as I read over #19 in more detail now, I was surprised to see that the main motivation to have pointers there seemed to be for the ability to _set_ fields, because at the time many structs were reused for setting and getting. /cc @willnorris

It's interesting that we've seemingly evolved towards dedicated Request structs for editing/updating operations (e.g., https://godoc.org/github.com/google/go-github/github#IssueRequest and https://godoc.org/github.com/google/go-github/github#Issue; https://godoc.org/github.com/google/go-github/github#BranchRestrictionsRequest and https://godoc.org/github.com/google/go-github/github#BranchRestrictions, also see #528, etc.). This was typically forced on us by certain new GitHub endpoints that had differences in the JSON they expect when editing/updating a resource. Especially whey they have required fields in the request.

In #512 we discussed and came to the conclusion that it makes sense to use values rather than pointers for required fields in the request structs.

### Conclusion

So in the end, I think that it's valid to also use values rather than pointers for mandatory fields (i.e., those that are guaranteed to be present by GitHub API).

**However,** one limitation of that is that GitHub API documentation doesn't typically specify fields in the response as mandatory/required, so it's hard to be sure they'll always be present.

[...]

PR #610, opened 2017-04-01
Add admin enforcement to protected branches API (#606)

amirilovic:
Helps #606.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -511,6 +511,7 @@ type Branch struct {
 type Protection struct {
 	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
 	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
+	EnforceAdmins              *AdminEnforcement           `json:"enforce_admins"`
 	Restrictions               *BranchRestrictions         `json:"restrictions"`
 }
 
@@ -518,6 +519,7 @@ type Protection struct {
 type ProtectionRequest struct {
 	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
 	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
+	EnforceAdmins              bool                        `json:"enforce_admins"`
 	Restrictions               *BranchRestrictionsRequest  `json:"restrictions"`
 }
 
@@ -536,6 +538,12 @@ type RequiredStatusChecks struct {
 type RequiredPullRequestReviews struct {
 	// Enforce pull request reviews for repository administrators. (Required.)
 	IncludeAdmins bool `json:"include_admins"`
+}
+
+// AdminEnforcement represents the configuration to enforce required status checks for repository administrators.
+type AdminEnforcement struct {
+	URL     *string `json:"url,omitempty"`
+	Enabled bool    `json:"enabled"`
 }
 
 // BranchRestrictions represents the restriction that only certain users or
--- a/github/repos.go
+++ b/github/repos.go
@@ -509,22 +509,21 @@ type Branch struct {
 
 // Protection represents a repository branch's protection.
 type Protection struct {
-	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
-	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
-	EnforceAdmins              *AdminEnforcement           `json:"enforce_admins"`
-	Restrictions               *BranchRestrictions         `json:"restrictions"`
+	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks"`
+	EnforceAdmins        *AdminEnforcement     `json:"enforce_admins"`
+	Restrictions         *BranchRestrictions   `json:"restrictions"`
 }
 
 // ProtectionRequest represents a request to create/edit a branch's protection.
 type ProtectionRequest struct {
-	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
-	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
-	EnforceAdmins              bool                        `json:"enforce_admins"`
-	Restrictions               *BranchRestrictionsRequest  `json:"restrictions"`
+	RequiredStatusChecks *RequiredStatusChecks      `json:"required_status_checks"`
+	EnforceAdmins        bool                       `json:"enforce_admins"`
+	Restrictions         *BranchRestrictionsRequest `json:"restrictions"`
 }
 
 // RequiredStatusChecks represents the protection status of a individual branch.
 type RequiredStatusChecks struct {
+	// Deprecated: Use EnforceAdmins instead.
 	// Enforce required status checks for repository administrators. (Required.)
 	IncludeAdmins bool `json:"include_admins"`
 	// Require branches to be up to date before merging. (Required.)
@@ -532,12 +531,6 @@ type RequiredStatusChecks struct {
 	// The list of status checks to require in order to merge into this
 	// branch. (Required; use []string{} instead of nil for empty list.)
 	Contexts []string `json:"contexts"`
-}
-
-// RequiredPullRequestReviews represents the protection configuration for pull requests.
-type RequiredPullRequestReviews struct {
-	// Enforce pull request reviews for repository administrators. (Required.)
-	IncludeAdmins bool `json:"include_admins"`
 }
 
 // AdminEnforcement represents the configuration to enforce required status checks for repository administrators.
--- a/github/repos.go
+++ b/github/repos.go
@@ -509,16 +509,18 @@ type Branch struct {
 
 // Protection represents a repository branch's protection.
 type Protection struct {
-	RequiredStatusChecks *RequiredStatusChecks `json:"required_status_checks"`
-	EnforceAdmins        *AdminEnforcement     `json:"enforce_admins"`
-	Restrictions         *BranchRestrictions   `json:"restrictions"`
+	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
+	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
+	EnforceAdmins              *AdminEnforcement           `json:"enforce_admins"`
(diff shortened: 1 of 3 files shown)


github/repos.go
```diff
@@ -511,13 +511,15 @@ type Branch struct {
 type Protection struct {
 	RequiredStatusChecks       *RequiredStatusChecks       `json:"required_status_checks"`
 	RequiredPullRequestReviews *RequiredPullRequestReviews `json:"required_pull_request_reviews"`
+	EnforceAdmins              *AdminEnforcement           `json:"enforce_admins"`
 	Restrictions               *BranchRestrictions         `json:"restrictions"`
```
shurcooL: The names of other structs here take pretty closely after the field name itself.

Have you considered naming this struct something like `EnforceAdmins`? If so, what made you pick `AdminEnforcement` instead?

It seems like the latter sounds better, but I'm a little unsure about the deviation from the pattern. Do you have any comments on this?

I'm on the fence about this, that's why I want to ask for your thoughts.

github/repos.go
```diff
@@ -538,6 +540,12 @@ type RequiredPullRequestReviews struct {
 	IncludeAdmins bool `json:"include_admins"`
 }
 
+// AdminEnforcement represents the configuration to enforce required status checks for repository administrators.
+type AdminEnforcement struct {
+	URL     *string `json:"url,omitempty"`
+	Enabled bool    `json:"enabled"`
```
shurcooL: What's the reason `URL` is a pointer `*string` and has `,omitempty`, unlike `Enabled`?

I think I get the reason, but it seems a bit arbitrary. I don't have any suggestions for improvement, so this is okay with me.

github/repos.go
```diff
@@ -538,6 +540,12 @@ type RequiredPullRequestReviews struct {
 	IncludeAdmins bool `json:"include_admins"`
```
shurcooL: In order to fully resolve #606, I think you should mark `IncludeAdmins` as deprecated, or remove it.

Same for `IncludeAdmins` inside `RequiredStatusChecks` struct.

If you deprecate it, then change it to "Helps #606", so it won't be closed. We want to use that issue to track the entire change for https://developer.github.com/changes/2017-03-22-protected-branches-admin-improvements/, which includes removing the old way of setting admin.

shurcooL:
Whoops, I self-requested a review as a way of not forgetting to deal with this PR, but I forgot anyway. Thanks for reminding.

github/repos.go
amirilovic: I choose AdminEnforcement because of: https://developer.github.com/v3/repos/branches/#get-admin-enforcement-of-protected-branch
The GET request states: Get **admin enforcement** of protected branch, which makes sense because it is actually a nested resource, not just a flag.

github/repos.go
amirilovic: @shurcooL I'm not sure about deprecation, seems like an overhead, maybe it's better just to remove old fields.

shurcooL: Feel free to go for it and make that a part of this PR.

github/repos.go
shurcooL: Sounds good, that's convincing rationale, thanks for providing it.

github/repos.go
navarrano: @amirilovic, if you don't mind, I'd love to know the response for shurcooL question, because I'd like to understand your approach here.

shurcooL: @navarrano See https://github.com/google/go-github/issues/537 and the issues it mentions for background information on this.

The TLDR is that this is a style choice, both ways work. Using pointers allows users to know whether the value was actually present in the GitHub response or not, but they have to deal with a pointer. Using values means you can't know if it's a zero value or it was present in GH response, but no pointers to deal with.

My understanding is that `Enabled` is supposed to be a field that's always included in the response, so it's kinda okay not making it a pointer. But URL is expected to be more optional, and it might be worth knowing if it's included or not.

It also has to do with consistency with other similar structs for this API.

But it could've gone in another way.

amirilovic: @shurcooL @navarrano sorry for late response, @shurcooL that was exactly my reasoning, don't know if you had similar cases, but it seemed logical to me.

github/repos.go
amirilovic: @shurcooL struct `RequiredPullRequestReviews` should be removed completely?

shurcooL: As I understand, we want to remove `IncludeAdmins` fields. I see that `RequiredPullRequestReviews` contains _only_ that field, and no others, so I understand where your question is coming from.

I would suspect that yes, we want to remove `RequiredPullRequestReviews`, since it has no other fields left, but the GitHub API docs say [some strange/conflicting things](https://developer.github.com/v3/repos/branches/#parameters-1) right now:

> GitHub recommends passing `required_pull_request_reviews` and `enforce_admins`. Passing these fields is not currently required, but may be required soon.

I'm not sure if that's out of date or a typo...

amirilovic: I'll contact support and try out how API behaves currently if no value is set for `RequiredPullRequestReviews`

amirilovic: @shurcooL No reply from support yet but here are test results:
1. When required_pull_request_reviews is not in the request, response status is 200 OK - 
existing required_pull_request_reviews is not affected
1. When required_pull_request_reviews is null, response is 200 OK - existing required_pull_request_reviews is removed from the resource
1. When required_pull_request_reviews is {}, response is 422 Unprocessable Entity, with following error in body:
```
{
  "message": "Invalid request.\n\nNo subschema in \"anyOf\" matched.\n\"include_admins\" wasn't supplied.\nNot all subschemas of \"allOf\" matched.\nFor 'anyOf/1', {} is not a null.",
  "documentation_url": "https://developer.github.com/v3/repos/branches/#update-branch-protection"
}
```
So, it's safe to remove required_pull_request_reviews from the request completely. 

For required_status_checks:
1. When required_status_checks is sent with include_admins, , response is 200 OK - existing required_status_checks is updated with new values
1. When required_status_checks is sent without include_admins, response is 422 Unprocessable Entity, with following error in body:
```
{
  "message": "Invalid request.\n\nNo subschema in \"anyOf\" matched.\n\"include_admins\" wasn't supplied.\nNot all subschemas of \"allOf\" matched.\nFor 'anyOf/1', {\"url\"=>\"https://api.github.com/repos/amirilovic/test-repo/branches/master/protection/required_status_checks\", \"strict\"=>true, \"contexts\"=>[], \"contexts_url\"=>\"https://api.github.com/repos/amirilovic/test-repo/branches/master/protection/required_status_checks/contexts\"} is not a null.",
  "documentation_url": "https://developer.github.com/v3/repos/branches/#update-branch-protection"
}
```

So, required_status_checks has to include include_admins field.

github/repos.go
```diff
@@ -509,22 +509,21 @@ type Branch struct {
...
-	Restrictions               *BranchRestrictionsRequest  `json:"restrictions"`
+	RequiredStatusChecks *RequiredStatusChecks      `json:"required_status_checks"`
+	EnforceAdmins        bool                       `json:"enforce_admins"`
+	Restrictions         *BranchRestrictionsRequest `json:"restrictions"`
 }
 
 // RequiredStatusChecks represents the protection status of a individual branch.
 type RequiredStatusChecks struct {
+	// Deprecated: Use EnforceAdmins instead.
 	// Enforce required status checks for repository administrators. (Required.)
```
shurcooL: You can't start the comment with "Deprecated:" first, because that breaks the Go style requirement of having documentation begin with identifier name. Put it second:

```
// Enforce required status checks for repository administrators. (Required.)
// Deprecated: Use EnforceAdmins instead.
```

Reference: https://github.com/golang/go/issues/10909#issuecomment-136492606.

amirilovic:
@shurcooL I've fixed the comment issue. I think there is no need to wait for response from support, I've tested current behavior of the API. If something comes up from GH support, I'll report the issue and fix it.

github/repos.go
shurcooL: The `RequiredStatusChecks.IncludeAdmins` field is now documented both as required and deprecated.

It's pretty puzzling to me how to use the new API. There are two `bool`s, do they need to be set to the same value if `true`, or does `EnforceAdmins` take higher precedence over `RequiredStatusChecks.IncludeAdmins`? Or can either one be set to `true`?

shurcooL:
Okay. I tried to review the PR and came up with a question at https://github.com/google/go-github/pull/610#discussion_r110760785.

shurcooL:
Also, see #617, which might be relevant here.

It seems to add `dismissal_restrictions` to `required_pull_request_reviews`, so perhaps we don't want to remove `required_pull_request_reviews` anymore.

github/repos.go
amirilovic: @shurcooL I've checked how it works. When setting either `RequiredStatusChecks.IncludeAdmins`,`RequiredPullRequestReviews.IncludeAdmins` or `EnforceAdmins` will set all other properties to the same value - changing one changes all the rest. `RequiredPullRequestReviews.IncludeAdmins` and `RequiredPullRequestReviews.IncludeAdmins` are required fields in terms of request structure correctness (request fails with `422 Unprocessable Entity`, if they are not included - explained in previous comment), but logically they shouldn't be used since `EnforceAdmins` overrides them both. So, they are required, but deprecated by `EnforceAdmins`.

amirilovic:
@shurcooL I've rolled back changes removing `RequiredPullRequestReviews`, also added deprecation comment for `RequiredPullRequestReviews.IncludeAdmins`. I've responded to your question about relationship between `RequiredPullRequestReviews.IncludeAdmins`, `RequiredStatusChecks.IncludeAdmins` and `EnforceAdmins`.

amirilovic:
@shurcooL Anything else that needs to be done here?

shurcooL:
Thanks for making progress on this. I'll try to review this soon.

github/repos.go
shurcooL: I still think this is confusing Go API for go-github users. But, this falls on GitHub, and it's temporary. According to https://developer.github.com/changes/2017-03-22-protected-branches-admin-improvements/, they'll remove the deprecated `include_admins` field on May 2, 2017, so this Go API will become less confusing then.

So, this is okay by me, I won't request further improvements (I can think of some, but I don't think it's worth the overhead).

shurcooL:
Thanks a lot @amirilovic!

Based on what it says at https://developer.github.com/changes/2017-03-22-protected-branches-admin-improvements/, we'll need to wait until May 2, 2017 to be able to completely resolve #606.

PR #612, opened 2017-04-05
RepositoriesService: Add required status checks methods

jimmidyson:
This PR adds required status checks methods as detailed in https://developer.github.com/v3/repos/branches/#get-required-status-checks-of-protected-branch
and https://developer.github.com/v3/repos/branches/#list-required-status-checks-contexts-of-protected-branch

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -629,6 +629,49 @@ func (s *RepositoriesService) GetBranchProtection(ctx context.Context, owner, re
 	return p, resp, nil
 }
 
+// GetRequiredStatusChecks gets the required status checks for a given protected branch.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#get-required-status-checks-of-protected-branch
+func (s *RepositoriesService) GetRequiredStatusChecks(ctx context.Context, owner, repo, branch string) (*RequiredStatusChecks, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_status_checks", owner, repo, branch)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypeProtectedBranchesPreview)
+
+	p := new(RequiredStatusChecks)
+	resp, err := s.client.Do(ctx, req, p)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return p, resp, nil
+}
+
+// ListRequiredStatusChecksContexts lists the required status checks contexts for a given protected branch.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#list-required-status-checks-contexts-of-protected-branch
+func (s *RepositoriesService) ListRequiredStatusChecksContexts(ctx context.Context, owner, repo, branch string) (contexts []string, resp *Response, err error) {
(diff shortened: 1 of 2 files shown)


github/repos.go
```diff
@@ -629,6 +629,50 @@ func (s *RepositoriesService) GetBranchProtection(ctx context.Context, owner, re
...
+		return nil, resp, err
+	}
+
+	return p, resp, nil
+}
+
+// ListRequiredStatusChecksContexts lists the required status checks contexts for a given protected branch.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#list-required-status-checks-contexts-of-protected-branch
+func (s *RepositoriesService) ListRequiredStatusChecksContexts(ctx context.Context, owner, repo, branch string) ([]string, *Response, error) {
```
shurcooL: It might make the API more readable to use named return parameters here, since `[]string` is not as readable as a named type. What do you think of this?

```Go
func (s *RepositoriesService) ListRequiredStatusChecksContexts(ctx context.Context, owner, repo, branch string) (contexts []string, resp *Response, err error) {
```

You can look at [`MigrationService.MigrationArchiveURL`](https://godoc.org/github.com/google/go-github/github#MigrationService.MigrationArchiveURL) as some precedent, it returns `(url string, err error)` rather than `(string, error)`.

Reference: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters.

However, it might be fine as is, since the method name `ListRequiredStatusChecksContexts` sorta implies that what's being returned are "contexts". But, the name is quite complex, so it might not be as obvious.

What do you think is better?

jimmidyson: Agree it is not obvious so switched to named return parameters to make it clearer. Shame about function name, but seems consistent with other function names and API docs.

github/repos_test.go
```diff
@@ -623,3 +623,55 @@ func TestRepositoriesService_License(t *testing.T) {
...
+		json.NewDecoder(r.Body).Decode(v)
+
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
+		fmt.Fprint(w, `{"include_admins": true,"strict": true,"contexts": ["x","y","z"]}`)
+	})
+
+	contexts, _, err := client.Repositories.GetRequiredStatusChecks(context.Background(), "o", "r", "b")
+	if err != nil {
+		t.Errorf("Repositories.ListRequiredStatusChecksContexts returned error: %v", err)
```
shurcooL: It looks like you've made some copy paste errors here, this should be `GetRequiredStatusChecks`, not `ListRequiredStatusChecksContexts`. Here, and below. Also, the returned value is better named `checks`, not `contexts`.

This is just a minor naming/error message issue in the tests, I can fix this while merging.

jimmidyson:
Happy with those changes - thanks for making them!

shurcooL:
Sadly, @googlebot and [collaborating on GitHub pull requests](https://github.com/blog/2247-improving-collaboration-with-forks) don't get along too well.

I checked that CLA was good before I pushed my commit, and I obviously consent to contribute my change (and I've signed CLA), so we can ignore that message and safely merge the PR.

However, it would be nice if @googlebot can be updated to handle this situation better. @willnorris, how realistic is it for that to happen? Is it a matter of doing some implementation work (that I can contribute to?), or will it be harder due to needing to be vetted by legal teams, etc.?

shurcooL:
Thanks for the PR @jimmidyson!

PR #613, opened 2017-04-06
Adds LDAPDN field to Team

alindeman:
In GitHub Enterprise, the teams API routes include `ldap_dn` if the team is being synchronized with LDAP.

### DIFF
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -39,6 +39,10 @@ type Team struct {
 	Organization    *Organization `json:"organization,omitempty"`
 	MembersURL      *string       `json:"members_url,omitempty"`
 	RepositoriesURL *string       `json:"repositories_url,omitempty"`
+
+	// LDAPDN is only available in GitHub Enterprise and when the team
+	// membership is synchronized with LDAP.
+	LDAPDN *string `json:"ldap_dn,omitempty"`
 }
 
 func (t Team) String() string {
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -6212,6 +6212,14 @@ func (t *Team) GetID() int {
 	return *t.ID
 }
 
+// GetLDAPDN returns the LDAPDN field if it's non-nil, zero value otherwise.
+func (t *Team) GetLDAPDN() string {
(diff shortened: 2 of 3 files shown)


github/orgs_teams.go
```diff
@@ -39,6 +39,10 @@ type Team struct {
 	Organization    *Organization `json:"organization,omitempty"`
 	MembersURL      *string       `json:"members_url,omitempty"`
 	RepositoriesURL *string       `json:"repositories_url,omitempty"`
+
+	// LDAPDN is only available in GitHub Enterprise and when the team
+	// membership is synchronized with LDAP
```
shurcooL: Add a missing period to the sentence please.

shurcooL:
Is this documented anywhere on GitHub API docs? I didn't see it at all at https://developer.github.com/v3/orgs/teams/, but I do see a similar field mentioned at https://developer.github.com/v3/enterprise/ldap/. Is that the best reference, or do you know of a better one?

This looks good to me after the minor missing period is fixed. Thanks!

alindeman:
> Is this documented anywhere on GitHub API docs? I didn't see it at all at https://developer.github.com/v3/orgs/teams/, but I do see a similar field mentioned at https://developer.github.com/v3/enterprise/ldap/. Is that the best reference, or do you know of a better one?

Unfortunately I can't find it documented either, but I can verify the field is returned when I run `GET /teams/:id` or `GET /orgs/:id/teams` against my GitHub Enterprise instance. The JSON looks something like:

```json
{
  "name": "Read-Only Users",
  "id": 67,
  "slug": "read-only-users",
  "description": "Read-Only Access",
  "privacy": "secret",
  "permission": "pull",
  "ldap_dn": "cn=ghereadonly,ou=Group,dc=example,dc=com",
  "members_count": 5,
  "repos_count": 10,
  // other fields elided
}
```

github/orgs_teams.go
alindeman: @shurcooL Thanks, fixed.

shurcooL:
Thanks @alindeman!

PR #614, opened 2017-04-09
Add Assigner field to issues_events

apelisse:
I've not updated the comment:

>        //     assigned, unassigned                                                                         
>        //       The Actor assigned the issue to or removed the assignment from the Assignee.               

to keep it consistent with [Github documentation](https://developer.github.com/v3/issues/events/) even though it is clearly incorrect.

### DIFF
--- a/github/issues_events.go
+++ b/github/issues_events.go
@@ -65,6 +65,7 @@ type IssueEvent struct {
 
 	// Only present on certain events; see above.
 	Assignee  *User      `json:"assignee,omitempty"`
+	Assigner  *User      `json:"assigner,omitempty"`
 	CommitID  *string    `json:"commit_id,omitempty"`
 	Milestone *Milestone `json:"milestone,omitempty"`
 	Label     *Label     `json:"label,omitempty"`


shurcooL:
Can you provide a reference for why this field should be added. Either a link to GitHub API docs that shows it, or if it's not documented, then a sample API query and response that contains it?

Also, I don't quite understand what you mean by this:

> I've not updated the comment:
>
> >        //     assigned, unassigned                                                                         
> >        //       The Actor assigned the issue to or removed the assignment from the Assignee.               
>
> to keep it consistent with [Github documentation](https://developer.github.com/v3/issues/events/) even though it is clearly incorrect.

Can you please elaborate? https://developer.github.com/v3/issues/events/ is a large page, what exactly are you referring to there?

apelisse:
Sure, it's described there: https://developer.github.com/v3/issues/events/#attributes

Inconsistency in the documentation:

- [Attributes](https://developer.github.com/v3/issues/events/#attributes):
> assigner
The User object that performed the assignment (or unassignment) for this Issue. Only provided for 'assigned' and 'unassigned' events.

- [Events](https://developer.github.com/v3/issues/events/#events-1)
> assigned
The issue was assigned to the actor.
unassigned
The actor was unassigned from the issue.

It sounds like "the actor" refers to the actor attribute, but that's not correct.

I'll send an email. Thanks.

shurcooL:
> It sounds like "the actor" refers to the actor attribute, but that's not correct.

I tried reading it multiple times, but I still don't quite see what you're referring to. Maybe it'd help if you showed what a potential patch would look like?

I can merge this PR which adds a missing field now. If there are additional doc enhancements possible, please feel free to follow up with another PR or issue.

PR #615, opened 2017-04-09
Update `IssueEvent.Event` assign/unassign doc

apelisse:
The current phrase reads as if the "Actor" field contains the
information about the user who changed the assignment, but this
information is actually in the "Assigner" field.

Fix the description of `assigned, unassigned` event to reflect
that.

Is that clearer?

### DIFF
--- a/github/issues_events.go
+++ b/github/issues_events.go
@@ -44,7 +44,7 @@ type IssueEvent struct {
 	//       Someone unspecified @mentioned the Actor [sic] in an issue comment body.
 	//
 	//     assigned, unassigned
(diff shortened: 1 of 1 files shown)


apelisse:
Actually, I re-read the github documentation, and it is correct:

- [Events](https://developer.github.com/v3/issues/events/#events-1)
> assigned
The issue was assigned to the actor.
unassigned
The actor was unassigned from the issue.

The `actor` field is the same as the `assignee`. That's OK.

This PR should still get merged though :-)

shurcooL:
Thanks a lot for this improvement!

Issue #618, opened 2017-04-18
Function to get the Delivery ID for a webhook request


drewfradette:
I would like to add a function to get the Delivery ID header (`X-GitHub-Delivery`) for a webhook request. This would essentially be the same as [`WebHookType`](https://github.com/google/go-github/blob/master/github/messages.go#L162) but a different header.

If there are no problems with this, I would be glad to open a PR.

gmlewis:
I'm not sure I understand the need for a new function. All that is needed is a single line:

```go
deliveryID := r.Header.Get("X-Github-Delivery")
```

Oh, I see... #427 added `WebHookType`.  Hmmm... I guess I missed that before. I'm not sure that was necessary, but I guess it shields clients from the details at the expense of a larger API surface.

I've got mixed feelings but could go either way.

@shurcooL - what do you think about adding a new function (maybe `DeliveryID`) to `messages.go` that is essentially a one-liner?

shurcooL:
SGTM.

PR #619, opened 2017-04-19
Add DeliveryID function for webhook events

drewfradette:
Addresses https://github.com/google/go-github/issues/618

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -161,11 +161,15 @@ func validateSignature(signature string, payload, secretKey []byte) error {
 }
 
 // WebHookType returns the event type of webhook request r.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/hooks/#webhook-headers
 func WebHookType(r *http.Request) string {
 	return r.Header.Get(eventTypeHeader)
 }
 
 // DeliveryID returns the unique delivery ID of webhook request r.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/hooks/#webhook-headers
 func DeliveryID(r *http.Request) string {
 	return r.Header.Get(deliveryIDHeader)
 }
--- a/github/messages.go
+++ b/github/messages.go
@@ -33,6 +33,8 @@ const (
 	signatureHeader = "X-Hub-Signature"
 	// eventTypeHeader is the GitHub header key used to pass the event type.
 	eventTypeHeader = "X-Github-Event"
+	// deliveryIDHeader is the GitHub header key used to pass the unique ID for the webhook event.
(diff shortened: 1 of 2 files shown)


github/messages.go
```diff
@@ -163,6 +165,11 @@ func WebHookType(r *http.Request) string {
 	return r.Header.Get(eventTypeHeader)
 }
 
+// DeliveryID returns the unique delivery ID of webhook request r.
```
shurcooL: Perhaps it's a good idea to link to the relevant GitHub API docs here?

```
// DeliveryID returns the unique delivery ID of webhook request r.
//
// GitHub API docs: https://developer.github.com/v3/repos/hooks/#webhook-headers
```

If you want to, you can add the same to docs of `WebHookType`.

drewfradette:
I was using the wrong git config, I've addressed the googlebot's issue above.

github/messages.go
drewfradette: Done!

github/messages_test.go
```diff
@@ -220,3 +220,17 @@ func TestParseWebHook(t *testing.T) {
...
+	id := "8970a780-244e-11e7-91ca-da3aabcb9793"
+	req, err := http.NewRequest("POST", "http://localhost", nil)
+	if err != nil {
+		t.Fatalf("DeliveryID: %v", err)
+	}
+	req.Header.Set("X-Github-Delivery", id)
+
+	got := DeliveryID(req)
+	if got != id {
+		t.Errorf("DeliveryID(%#v) = %#v, want %#v", req, got, id)
```
gmlewis: nit: since `got` and `id` are both strings, you may want to use `%q` for these values instead of `%#v`.

`t.Errorf("DeliveryID(%#v) = %q, want %q", req, got, id)`

drewfradette: Done!

shurcooL: Drew, the commit message of ffae808934ce7f830c244465e103a28ddda502c1 is "**Fix** formatting of test output". However, both `%#v` and `%q` produce same results for `string` type. It wasn't wrong, this is a mere style change. Just wanted to point that out FYI.

drewfradette: Yeah, sorry, bad wording. I meant "fixed" in the context of "for the review". Would you like me to change the commit message to be more precise?

shurcooL: No problem. No need to change the commit message, this PR will be squashed into 1 commit when merging anyway, and the review-specific parts don't need to be included.

shurcooL:
Thanks @drewfradette!

Issue #569, opened 2017-03-02
Support new user blocking APIs and webhook


gmlewis:
Announcement:
https://developer.github.com/changes/2017-02-28-user-blocking-apis-and-webhook/

GitHub API docs:
- [x] https://developer.github.com/v3/users/blocking/
- [x] https://developer.github.com/v3/orgs/blocking/
- [ ] https://developer.github.com/v3/activity/events/types/#orgblockevent

varadarajana:
@gmlewis please assign this to me, as this is similar to #530. I need one help with your review of Issue #530. Can you provide your comments on the usage of Changes struct?

gmlewis:
Done.  Thanks, @varadarajana! I'll take a look at 530.

varadarajana:
Today I came back to work after sometime and found difficulty in merging, here is what I get after I do  
'''
$ git fetch upstream
remote: Counting objects: 96, done.
remote: Compressing objects: 100% (15/15), done.
remote: Total 96 (delta 65), reused 60 (delta 60), pack-reused 21
Unpacking objects: 100% (96/96), done.
From https://github.com/google/go-github
 * [new branch]      master     -> upstream/master
'''
'''
$ git checkout master
Already on 'master'
Your branch is up-to-date with 'origin/master'.
'''

'''
$ git merge upstream/master
Auto-merging github/orgs_teams_test.go
Auto-merging github/orgs_teams.go
Auto-merging github/orgs_members_test.go
CONFLICT (content): Merge conflict in github/orgs_members_test.go
Auto-merging github/orgs_members.go
Automatic merge failed; fix conflicts and then commit the result.
'''

I tried stash, checkout --theirs, reset and manually tried to change the files, somehow I get this error. Is there something I am doing wrong?

shurcooL:
What are you trying to achieve?

If it were me, I would start with latest `upstream/master` (assuming your upstream remote is google/go-github), and work from there. If you had some previous work on a branch, I would rebase that on top of latest master.

varadarajana:
@shurcooL 
I want to sync up my fork with the upstream and catch up to the latest version on upstream. So I fetch upstream and then checkout my fork, and merge.

varadarajana:
@shurcooL 
This time I did a

'''
git add github/orgs_members_test.go
git commit
git merge upstream/master
'''
This worked. I hope I did the right way.

shurcooL:
You should make sure there aren't unintended merge conflicts as part of that file.

[...]

PR #600, opened 2017-03-25
#569:Partial Fix: Fix for Users Blocking

varadarajana:
Please find the 2 files containing the partial fixes for blocking users. I will create one more for organization.

PR #620, opened 2017-04-20
#569 part 2 : Add Organization Blocking Users API

varadarajana:
This is continuation of #569. I have made a first cut for review here.

### DIFF
--- a/github/orgs_users_blocking.go
+++ b/github/orgs_users_blocking.go
@@ -14,7 +14,6 @@ import (
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/blocking/#list-blocked-users
 func (s *OrganizationsService) ListBlockedUsers(ctx context.Context, org string, opt *ListOptions) ([]*User, *Response, error) {
-	// TODO: add a check for scope to be admin:org
 	u := fmt.Sprintf("orgs/%v/blocks", org)
 	u, err := addOptions(u, opt)
 	if err != nil {
@@ -42,7 +41,6 @@ func (s *OrganizationsService) ListBlockedUsers(ctx context.Context, org string,
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/blocking/#check-whether-a-user-is-blocked-from-an-organization
 func (s *OrganizationsService) IsBlocked(ctx context.Context, org string, user string) (bool, *Response, error) {
-	// TODO: add a check for scope to be admin:org
 	u := fmt.Sprintf("orgs/%v/blocks/%v", org, user)
 
 	req, err := s.client.NewRequest("GET", u, nil)
@@ -62,7 +60,6 @@ func (s *OrganizationsService) IsBlocked(ctx context.Context, org string, user s
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/blocking/#block-a-user
 func (s *OrganizationsService) BlockUser(ctx context.Context, org string, user string) (*Response, error) {
-	// TODO: add a check for scope to be admin:org
 	u := fmt.Sprintf("orgs/%v/blocks/%v", org, user)
 
 	req, err := s.client.NewRequest("PUT", u, nil)
@@ -80,7 +77,6 @@ func (s *OrganizationsService) BlockUser(ctx context.Context, org string, user s
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/blocking/#unblock-a-user
 func (s *OrganizationsService) UnblockUser(ctx context.Context, org string, user string) (*Response, error) {
-	// TODO: add a check for scope to be admin:org
 	u := fmt.Sprintf("user/%v/blocks/%v", org, user)
 
 	req, err := s.client.NewRequest("DELETE", u, nil)
--- a/github/orgs_users_blocking.go
+++ b/github/orgs_users_blocking.go
@@ -1,0 +1,92 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// ListBlockedUsers lists all the users blocked by an organization.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/blocking/#list-blocked-users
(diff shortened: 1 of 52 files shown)


shurcooL:
@varadarajana The [commits](https://github.com/google/go-github/pull/620/commits) in this PR are identical to the [commits in PR #600](https://github.com/google/go-github/pull/600/commits), which we already merged (as d23570d44313ca73dbcaadec71fc43eca4d29f8b).

varadarajana:
@shurcooL sorry, I missed pushing my code. Now it is done. you can check commit 7377254.

github/orgs_users_blocking.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// UnblockUser unblocks specified user from an organization.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/blocking/#unblock-a-user
+func (s *OrganizationsService) UnblockUser(ctx context.Context, org string, user string) (*Response, error) {
+	// TODO: add a check for scope to be admin:org
+	u := fmt.Sprintf("user/%v/blocks/%v", org, user)
```
shurcooL: Hmm. It's really surprising to see `user/%v/blocks/%v` here, instead of `orgs/%v/blocks/%v`.

However, I can see it matches with what the GitHub API says at https://developer.github.com/v3/orgs/blocking/#unblock-a-user.

But I suspect it's a mistake on their part. Look at how it's inconsistent compared to the user-specific blocks (on the right hand side):

image

This is worthy of sending them an email to clarify. I've done that just now, and I'll report back here when I hear back from them.

github/orgs_users_blocking.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"context"
+	"fmt"
+)
+
+// ListBlockedUsers lists all the users blocked by an organization.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/blocking/#list-blocked-users
+func (s *OrganizationsService) ListBlockedUsers(ctx context.Context, org string, opt *ListOptions) ([]*User, *Response, error) {
+	// TODO: add a check for scope to be admin:org
```
shurcooL: You have this `// TODO: add a check for scope to be admin:org` in 8 places. You also asked:

> I have a question. While i was reading the Blocking Users (Organizations) API, I find the following lines:
> ```
> The token used to authenticate the call must have the admin:org scope in order to make any
> blocking calls for an organization. Otherwise, the response returns HTTP 404.
> ```
> In authorizations.go we have all the scopes defines. I do not see any references to scope anywhere? Is there a central place where this scope validation is done?

Let me answer that here. That note is meant for users of GitHub API. When authenticating, they need to make sure the token they're using has that scope.

There's nothing we need to do for these endpoints inside go-github, that information is for users of go-github. Authentication is orthogonal matter, see https://github.com/google/go-github#authentication.

So, nothing needs to be done here, and you can safely remove these TODOs.

However, if you want, you can add that note to the documentation for these 4 endpoints. E.g.:

```
// ListBlockedUsers lists all the users blocked by an organization.
//
// The token used to authenticate the call must have the admin:org scope
// in order to make any blocking calls for an organization. Otherwise,
// the response returns HTTP 404.
//
// GitHub API docs: https://developer.github.com/v3/orgs/blocking/#list-blocked-users
```

We generally don't try to document this for all endpoints, people are expected to read GitHub API documentation to figure out what scopes are required. So I'm not sure if it's worth doing or not just here.

varadarajana: Thank you @shurcooL . I do not think there is a need to document this at end point since one needs to follow the API documentation. I have removed the TODO.

github/orgs_users_blocking.go
varadarajana: @shurcooL Thank you, I never realized the URL seems incorrect.

github/orgs_users_blocking_test.go
```diff
@@ -0,0 +1,90 @@
...
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/orgs/o1/blocks/u1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		testHeader(t, r, "Accept", mediaTypeBlockUsersPreview)
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	_, err := client.Organizations.BlockUser(context.Background(), "o1", "u1")
```
shurcooL: Is there a reason this test uses "o1" and "u1", while the rest use "o" and "u"? It seems a little inconsistent.

github/orgs_users_blocking.go
```diff
@@ -0,0 +1,91 @@
...
+	req.Header.Set("Accept", mediaTypeBlockUsersPreview)
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// UnblockUser unblocks specified user from an organization.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/blocking/#unblock-a-user
+func (s *OrganizationsService) UnblockUser(ctx context.Context, org string, user string) (*Response, error) {
+	u := fmt.Sprintf("user/%v/blocks/%v", org, user)
```
shurcooL: This should be changed to `"orgs/%v/blocks/%v"`.

github/orgs_users_blocking_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		t.Errorf("Organizations.BlockUser returned error: %v", err)
+	}
+}
+
+func TestOrganizationsService_UnblockUser(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/user/o/blocks/u", func(w http.ResponseWriter, r *http.Request) {
```
gmlewis: Obviously, this line will be changed too once @shurcooL's comments are addressed, to make the tests pass.

github/orgs_users_blocking_test.go
varadarajana: @shurcooL I was in a debugging effort and trying out combinations of URL. that's where I out o1 and u1. For consistency as you have mentioned, I have replaced o1 and u1 with o and u.

github/orgs_users_blocking.go
varadarajana: @shurcooL I have made these changes in the latest checkin.

PR #626, opened 2017-04-27
RepositoriesService: Return nil slice for required status checks on unprotected branches

### LINKED ISSUE #625: Branch protection status checks returns error for unprotected branches
Response is 404 so throws an error, instead of returning nil slice of required contexts. Sending in a quick PR.

jimmidyson:
Fixes #625

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -655,6 +655,8 @@ func (s *RepositoriesService) GetRequiredStatusChecks(ctx context.Context, owner
 	p := new(RequiredStatusChecks)
 	resp, err := s.client.Do(ctx, req, p)
 	if err != nil {
+		// if it's just a 404, don't return that as an error
+		_, err = parseBoolResponse(err)
 		return nil, resp, err
 	}
 
@@ -676,6 +678,8 @@ func (s *RepositoriesService) ListRequiredStatusChecksContexts(ctx context.Conte
 
 	resp, err = s.client.Do(ctx, req, &contexts)
 	if err != nil {
+		// if it's just a 404, don't return that as an error, but rather just the empty slice
+		_, err = parseBoolResponse(err)
 		return nil, resp, err
 	}
 
--- a/github/repos.go
+++ b/github/repos.go
@@ -639,7 +639,8 @@ func (s *RepositoriesService) GetBranchProtection(ctx context.Context, owner, re
 	return p, resp, nil
 }
 
-// GetRequiredStatusChecks gets the required status checks for a given protected branch.
+// GetRequiredStatusChecks gets the required status checks for a given protected branch,
+// or nil if the branch is not protected.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/branches/#get-required-status-checks-of-protected-branch
 func (s *RepositoriesService) GetRequiredStatusChecks(ctx context.Context, owner, repo, branch string) (*RequiredStatusChecks, *Response, error) {
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -717,7 +717,7 @@ func TestRepositoriesService_ListRequiredStatusChecksContexts_unprotectedBranch(
 		t.Errorf("Repositories.ListRequiredStatusChecksContexts returned error: %v", err)
 	}
 
-	if contexts != nil {
-		t.Errorf("Repositories.ListRequiredStatusChecksContexts returned %+v, want nil", contexts)
+	if len(contexts) != 0 {
+		t.Errorf("Repositories.ListRequiredStatusChecksContexts returned %+v, want zero length slice", contexts)
 	}
 }
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -633,9 +633,6 @@ func TestRepositoriesService_GetRequiredStatusChecks(t *testing.T) {
 	defer teardown()
 
 	mux.HandleFunc("/repos/o/r/branches/b/protection/required_status_checks", func(w http.ResponseWriter, r *http.Request) {
-		v := new(ProtectionRequest)
-		json.NewDecoder(r.Body).Decode(v)
-
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
 		fmt.Fprint(w, `{"include_admins": true,"strict": true,"contexts": ["x","y","z"]}`)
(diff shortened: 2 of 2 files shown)


github/repos.go
```diff
@@ -676,10 +677,11 @@ func (s *RepositoriesService) ListRequiredStatusChecksContexts(ctx context.Conte
 
 	resp, err = s.client.Do(ctx, req, &contexts)
 	if err != nil {
-		return nil, resp, err
+		// if it's just a 404, don't return that as an error, but rather just the empty slice
+		_, err = parseBoolResponse(err)
 	}
 
-	return contexts, resp, nil
+	return contexts, resp, err
```
shurcooL: Same here.

github/repos.go
```diff
@@ -655,10 +655,11 @@ func (s *RepositoriesService) GetRequiredStatusChecks(ctx context.Context, owner
 	p := new(RequiredStatusChecks)
 	resp, err := s.client.Do(ctx, req, p)
 	if err != nil {
-		return nil, resp, err
+		// if it's just a 404, don't return that as an error
+		_, err = parseBoolResponse(err)
 	}
 
-	return p, resp, nil
+	return p, resp, err
```
shurcooL: This last line can remain the success case:

```Go
return p, resp, nil
```

It'd also be consistent with https://github.com/google/go-github/blob/e8d46665e050742f457a58088b1e6b794b2ae966/github/activity_watching.go#L97-L103.

github/repos.go
```diff
@@ -655,10 +655,11 @@ func (s *RepositoriesService) GetRequiredStatusChecks(ctx context.Context, owner
 	p := new(RequiredStatusChecks)
 	resp, err := s.client.Do(ctx, req, p)
 	if err != nil {
-		return nil, resp, err
+		// if it's just a 404, don't return that as an error
+		_, err = parseBoolResponse(err)
 	}
```
shurcooL: I think it'd be more readable to handle the error completely here, rather than continuing:

```Go
if err != nil {
	// if it's just a 404, don't return that as an error
	_, err = parseBoolResponse(err)
	return nil, resp, err
}
```

jimmidyson:
@shurcooL Thanks for the review. Updated as you've recommended and updated tests to reflect the behaviour of returning nils.

github/repos_test.go
```diff
@@ -656,14 +653,35 @@ func TestRepositoriesService_GetRequiredStatusChecks(t *testing.T) {
...
+  "documentation_url": "https://developer.github.com/v3"
+}`)
+	})
+
+	checks, _, err := client.Repositories.GetRequiredStatusChecks(context.Background(), "o", "r", "b")
+	if err != nil {
+		t.Errorf("Repositories.GetRequiredStatusChecks returned error: %v", err)
+	}
+
+	if checks != nil {
```
shurcooL: How do you feel about verifying that `len(checks) == 0` instead?

github/repos_test.go
```diff
@@ -679,3 +697,27 @@ func TestRepositoriesService_ListRequiredStatusChecksContexts(t *testing.T) {
...
+  "documentation_url": "https://developer.github.com/v3"
+}`)
+	})
+
+	contexts, _, err := client.Repositories.ListRequiredStatusChecksContexts(context.Background(), "o", "r", "b")
+	if err != nil {
+		t.Errorf("Repositories.ListRequiredStatusChecksContexts returned error: %v", err)
+	}
+
+	if contexts != nil {
```
shurcooL: How do you feel about checking if `len(contexts) != 0` instead?

I'm not sure if we should verify whether the method returns `nil` slice or `[]string{}` empty slice. Do you think we should?

github/repos.go
```diff
@@ -655,6 +655,8 @@ func (s *RepositoriesService) GetRequiredStatusChecks(ctx context.Context, owner
 	p := new(RequiredStatusChecks)
```
shurcooL: Perhaps we should also update documentation to reflect this.

```
// GetRequiredStatusChecks gets the required status checks for a given protected branch,
// or nil if the branch is not protected.
```

github/repos_test.go
jimmidyson: Agree it's more idiomatic to check slice length. I'll change to that.

github/repos.go
jimmidyson: Updated.

shurcooL:
Code-wise, everything looks great to me now, thanks.

However, it concerns me that I haven't been able to get a non-404 response from GitHub API for any repository. Did you see my comment in #625 @jimmidyson? Can you help me out reproduce this?

github/repos_test.go
```diff
@@ -656,14 +653,35 @@ func TestRepositoriesService_GetRequiredStatusChecks(t *testing.T) {
...
+func TestRepositoriesService_GetRequiredStatusChecks_unprotectedBranch(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/branches/b/protection/required_status_checks", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprint(w, `{
+  "message": "Not Found",
```
shurcooL: In [my testing](https://github.com/google/go-github/issues/625#issuecomment-298063138), the message one gets when a branch is unprotected is "Branch not protected". It's "Not Found" only when the query is rejected because of insufficient permissions to the repo.

I think it's be good idea to display an accurate error message here, so that people in the future don't get confused.

github/repos_test.go
```diff
@@ -679,3 +697,27 @@ func TestRepositoriesService_ListRequiredStatusChecksContexts(t *testing.T) {
...
+func TestRepositoriesService_ListRequiredStatusChecksContexts_unprotectedBranch(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/branches/b/protection/required_status_checks/contexts", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprint(w, `{
+  "message": "Not Found",
```
shurcooL: Same here.

shurcooL:
I don't think this PR should be merged as is until we conclude the API discussion at https://github.com/google/go-github/issues/625#issuecomment-298063834.

I don't want this to be accidentally merged, and since this already has two LGTMs, the only way I can think of is to close this PR for now. When we finish the discussion in #625, we can re-open this PR and resume, if needed. I hope that's okay @jimmidyson.

Issue #553, opened 2017-02-16
Support new Community Health preview API


gmlewis:
Announcement:
https://developer.github.com/changes/2017-02-09-community-health/
Docs:
https://developer.github.com/v3/repos/community/

sahildua2305:
@gmlewis I will work on this. Please assign it to me. 😄

gmlewis:
Thanks, @sahildua2305! Done.

gmlewis:
Breaking updates to the preview API:
https://developer.github.com/changes/2017-03-07-community-health-api-enhancements/

sahildua2305:
@gmlewis I am looking into it now and thinking about what `type` can I have for the response. In `files`, there are four keys which are the file names and if that file is present, there will be corresponding fields.

shurcooL:
It can be a struct, something like this:

```Go
type CommunityHealthMetrics struct {
	HealthPercentage int `json:"health_percentage"`
	Files            struct {
		CodeOfConduct struct {
			Name    string `json:"name"`
			Key     string `json:"key"`
			URL     string `json:"url"`
			HTMLURL string `json:"html_url"`
		} `json:"code_of_conduct"`
		Contributing struct {
			URL     string `json:"url"`
			HTMLURL string `json:"html_url"`
		} `json:"contributing"`
		License struct {
			Name    string `json:"name"`
			Key     string `json:"key"`
			URL     string `json:"url"`
			HTMLURL string `json:"html_url"`
		} `json:"license"`
		Readme struct {
			URL     string `json:"url"`
			HTMLURL string `json:"html_url"`
		} `json:"readme"`
	} `json:"files"`
	UpdatedAt time.Time `json:"updated_at"`
}
```

And it looks like you'll need a slice of them, because the response is a JSON array, not a single object.

_(Thanks to @mholt for https://mholt.github.io/json-to-go/.)_

sahildua2305:
Thanks @shurcooL! I didn't know about this json-to-go tool. It's very helpful.

PR #628, opened 2017-05-04
Add Community Health metrics endpoint

sahildua2305:
This is a new API released by GitHub and is currently available as a
preview only.
Link - https://developer.github.com/v3/repos/community/#retrieve-community-health-metrics

Fixes: #553

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -97,6 +97,9 @@ const (
 
 	// https://developer.github.com/changes/2017-02-28-user-blocking-apis-and-webhook/
 	mediaTypeBlockUsersPreview = "application/vnd.github.giant-sentry-fist-preview+json"
+
+	// https://developer.github.com/changes/2017-02-09-community-health/
+	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview"
 )
 
 // A Client manages communication with the GitHub API.
--- a/github/github.go
+++ b/github/github.go
@@ -99,7 +99,7 @@ const (
 	mediaTypeBlockUsersPreview = "application/vnd.github.giant-sentry-fist-preview+json"
 
 	// https://developer.github.com/changes/2017-02-09-community-health/
-	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview"
+	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview+json"
 )
 
 // A Client manages communication with the GitHub API.
--- a/github/repos_community_health.go
+++ b/github/repos_community_health.go
@@ -43,7 +43,7 @@ type CommunityHealthMetrics struct {
 //
 // GitHub API docs: https://developer.github.com/v3/repos/community/#retrieve-community-health-metrics
 func (s *RepositoriesService) GetCommunityHealthMetrics(ctx context.Context, owner, repo string) (*CommunityHealthMetrics, *Response, error) {
-	u := fmt.Sprintf("repositories/%v/%v/community/profile", owner, repo)
+	u := fmt.Sprintf("repos/%v/%v/community/profile", owner, repo)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
@@ -54,5 +54,9 @@ func (s *RepositoriesService) GetCommunityHealthMetrics(ctx context.Context, own
 
 	metrics := &CommunityHealthMetrics{}
 	resp, err := s.client.Do(ctx, req, metrics)
-	return metrics, resp, err
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return metrics, resp, nil
 }
--- a/github/repos_community_health.go
+++ b/github/repos_community_health.go
@@ -21,7 +21,7 @@ type Metric struct {
 
 // CommunityHealthFiles represents the different files in the community health metrics response.
 type CommunityHealthFiles struct {
-	CodeOfConduct *Metric `json:"code_conduct"`
+	CodeOfConduct *Metric `json:"code_of_conduct"`
 	Contributing  *Metric `json:"contributing"`
 	License       *Metric `json:"license"`
 	Readme        *Metric `json:"readme"`
--- a/github/repos_community_health.go
+++ b/github/repos_community_health.go
@@ -1,0 +1,58 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+// CommunityHealthMetrics represents a response containing the community metrics of a repository.
+type CommunityHealthMetrics struct {
+	HealthPercentage *int `json:"health_percentage"`
+	Files            struct {
+		CodeOfConduct struct {
+			Name    *string `json:"name"`
+			Key     *string `json:"key"`
+			URL     *string `json:"url"`
+			HTMLURL *string `json:"html_url"`
+		} `json:"code_of_conduct"`
+		Contributing struct {
+			URL     *string `json:"url"`
+			HTMLURL *string `json:"html_url"`
+		} `json:"contributing"`
+		License struct {
+			Name    *string `json:"name"`
+			Key     *string `json:"key"`
+			URL     *string `json:"url"`
+			HTMLURL *string `json:"html_url"`
+		} `json:"license"`
+		Readme struct {
+			URL     *string `json:"url"`
+			HTMLURL *string `json:"html_url"`
+		} `json:"readme"`
+	} `json:"files"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+// GetCommunityHealthMetrics retrieves all the community health  metrics for a  repository.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/community/#retrieve-community-health-metrics
+func (s *RepositoriesService) GetCommunityHealthMetrics(ctx context.Context, owner, repo string) (*CommunityHealthMetrics, *Response, error) {
+	u := fmt.Sprintf("repositories/%v/%v/community/profile", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


github/repos_community_health.go
```diff
@@ -0,0 +1,58 @@
...
+			Key     *string `json:"key"`
+			URL     *string `json:"url"`
+			HTMLURL *string `json:"html_url"`
+		} `json:"license"`
+		Readme struct {
+			URL     *string `json:"url"`
+			HTMLURL *string `json:"html_url"`
+		} `json:"readme"`
+	} `json:"files"`
+	UpdatedAt time.Time `json:"updated_at"`
```
shurcooL: Should this be a pointer too, for consistency? Other structs have it as pointer.

sahildua2305: Yes indeed. I'll update it. Is it good otherwise?

github/repos_community_health.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// GetCommunityHealthMetrics retrieves all the community health  metrics for a  repository.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/community/#retrieve-community-health-metrics
+func (s *RepositoriesService) GetCommunityHealthMetrics(ctx context.Context, owner, repo string) (*CommunityHealthMetrics, *Response, error) {
+	u := fmt.Sprintf("repositories/%v/%v/community/profile", owner, repo)
```
shurcooL: This should be `repos/%v/%v/community/profile`.

`repositories/:repo_id` is used with numeric repo IDs, rather than `owner/repo`.

GitHub API docs are being inconsistent in using `repositories/:repo_id` URL style in their docs at https://developer.github.com/v3/repos/community/, and it'd be good to send them an email at support@github.com and let them know. Both are valid, it's just not consistent.

github/repos_community_health.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeRepositoryCommunityHealthMetricsPreview)
+
+	metrics := &CommunityHealthMetrics{}
+	resp, err := s.client.Do(ctx, req, metrics)
+	return metrics, resp, err
```
shurcooL: I think it'd be better to use same style as other endpoints here:

```Go
if err != nil {
	return nil, resp, err
}

return metrics, resp, nil
```

For consistency.

github/github.go
```diff
@@ -97,6 +97,9 @@ const (
 
 	// https://developer.github.com/changes/2017-02-28-user-blocking-apis-and-webhook/
 	mediaTypeBlockUsersPreview = "application/vnd.github.giant-sentry-fist-preview+json"
+
+	// https://developer.github.com/changes/2017-02-09-community-health/
+	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview"
```
shurcooL: Should have `+json` at the end, like above.

sahildua2305:
@shurcooL made changes. Please review again.

gmlewis:
@shurcooL - I'm concerned that by defining `CommunityHealthMetrics` with inline structs, we are missing out on not getting auto accessors for all the metrics fields.

I know there isn't a single struct that fits all the inline fields, and I don't think we want to make custom structs for each field since I get the impression from the GitHub API docs that they could add more metrics fields at any time. Custom structs would be a maintenance pain.

I suggest we make a single `Metric` struct that supports all the current metrics and a separate `MetricsFiles` struct... something like this:

```go
// Metric defines common fields used in CommunityHealthMetrics.
type Metric struct {
	Name    *string `json:"name"`
	Key     *string `json:"key"`
	URL     *string `json:"url"`
	HTMLURL *string `json:"html_url"`
}

// MetricsFiles represents files contained in CommunityHealthMetrics.
type MetricsFiles struct {
	CodeOfConduct *Metric `json:"code_of_conduct"`
	Contributing  *Metric `json:"contributing"`
	License       *Metric `json:"license"`
	Readme        *Metric `json:"readme"`
}

// CommunityHealthMetrics represents a response containing the community metrics of a repository.
type CommunityHealthMetrics struct {
	HealthPercentage *int          `json:"health_percentage"`
	Files            *MetricsFiles `json:"files"`
	UpdatedAt        *time.Time    `json:"updated_at"`
}
```

Thoughts?

shurcooL:
It's hard to predict how exactly the response will evolve, if at all, but that sounds reasonable to me. I'm not against it.

My only change suggestion is to call `MetricsFiles` as `CommunityHealthFiles` instead, what do you think?

sahildua2305:
@gmlewis thanks for the suggestions. I have made the corresponding changes.

github/repos_community_health.go
```diff
@@ -0,0 +1,55 @@
...
+
+package github
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+type Metric struct {
```
gmlewis: Please add comment for exported struct.

github/repos_community_health.go
```diff
@@ -0,0 +1,55 @@
...
+	Name    *string `json:"name"`
+	Key     *string `json:"key"`
+	URL     *string `json:"url"`
+	HTMLURL *string `json:"html_url"`
+}
+
+type CommunityHealthFiles struct {
```
gmlewis: Please add comment for exported struct.

github/repos_community_health.go
```diff
@@ -11,32 +11,25 @@ import (
...
 
+type Metric struct {
+	Name    *string `json:"name"`
+	Key     *string `json:"key"`
+	URL     *string `json:"url"`
+	HTMLURL *string `json:"html_url"`
+}
+
+type CommunityHealthFiles struct {
+	CodeOfConduct *Metric `json:"code_conduct"`
```
shurcooL: This should be be `json:"code_of_conduct"`.

github/repos_community_health_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func TestRepositoriesService_GetCommunityHealthMetrics(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/community/profile", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeRepositoryCommunityHealthMetricsPreview)
+		fmt.Fprintf(w, `{"health_percentage":75}`)
```
shurcooL: Perhaps it's a good idea to make this test have higher coverage. You can use the entire sample response from https://developer.github.com/v3/repos/community/#retrieve-community-health-metrics:

```JSON
{
  "health_percentage": 100,
  "files": {
    "code_of_conduct": {
      "name": "Contributor Covenant",
      "key": "contributor_covenant",
      "url": null,
      "html_url": "https://github.com/octocat/Hello-World`/blob/master/CODE_OF_CONDUCT.md"
    },
    "contributing": {
      "url": "https://api.github.com/repos/octocat/Hello-World/contents/CONTRIBUTING",
      "html_url": "https://github.com/octocat/Hello-World/blob/master/CONTRIBUTING"
    },
    "license": {
      "name": "MIT License",
      "key": "mit",
      "url": "https://api.github.com/licenses/mit",
      "html_url": "https://github.com/octocat/Hello-World/blob/master/LICENSE"
    },
    "readme": {
      "url": "https://api.github.com/repos/octocat/Hello-World/contents/README.md",
      "html_url": "https://github.com/octocat/Hello-World/blob/master/README.md"
    }
  },
  "updated_at": "2017-02-28T19:09:29Z"
}
```

sahildua2305:
@shurcooL I have fixed the bug in JSON tag and have increased the test coverage as well. However, for some reason, test is passing. Can you please help me debug it?

shurcooL:
> I have fixed the bug in JSON tag and have increased the test coverage as well.

That part (78d7b04ed1f33731d6ed503bc220372a43ab5050) looks good.

> However, for some reason, the test isn't passing. Can you please help me debug it?

What is the output from the test failure?

sahildua2305:
This is the failure message from the test
```
--- FAIL: TestRepositoriesService_GetCommunityHealthMetrics (0.00s)
	repos_community_health_test.go:83: Repositories.GetCommunityHealthMetrics = &{HealthPercentage:0xc420476098 Files:0xc4203056e0 UpdatedAt:2017-02-28 00:00:00 +0000 UTC}, want &{HealthPercentage:0xc420476320 Files:0xc420305900 UpdatedAt:2017-02-28 00:00:00 +0000 UTC}
FAIL
FAIL	github.com/google/go-github/github	0.660s
```

shurcooL:
It's pretty hard to see how exactly they differ because they are printed plainly:

```Go
t.Errorf("Repositories.GetCommunityHealthMetrics = %+v, want %+v", got, want)
```

Have you tried using [`Stringify`](https://godoc.org/github.com/google/go-github/github#Stringify) to print them instead?

```Go
t.Errorf("Repositories.GetCommunityHealthMetrics:\ngot:\n%v\nwant:\n%v", Stringify(got), Stringify(want))
```

By doing that, you should be able to spot where the issue is.

sahildua2305:
@shurcooL Thanks ❤️

It helped me spot the issue. Should I keep the `Stringify` in the error message for better debugging in future as well?

shurcooL:
Please do. Otherwise, the output of that test, should it happen to fail, is very hard to read.

github/repos_community_health_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	mux.HandleFunc("/repos/o/r/community/profile", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeRepositoryCommunityHealthMetricsPreview)
+		fmt.Fprintf(w, `{
+				"health_percentage": 100,
+				"files": {
+					"code_of_conduct": {
+						"name": "Contributor Covenant",
+						"key": "contributor_covenant",
```
shurcooL: Why not include `"url": null,` here, just like in the [example](https://developer.github.com/v3/repos/community/#retrieve-community-health-metrics)? Test would still pass, right?

sahildua2305: Yes, right! 😄

PR #633, opened 2017-05-07
Add OnRequest + OnResponse hooks

radeksimko:
I was unable to find a way to debug the communication between Github and the client and server, which is something we'd find extremely useful in [Terraform](https://github.com/hashicorp/terraform), especially when our nightly acceptance tests fail because of what appears to be eventual consistency.

The only way to debug such issues without this patch from my point of view is either:

1. tcpdump + some kind of MITM proxy (because all traffic is encrypted), or
2. capture _every single_ request and response from each call, e.g. from `activitySvc.ListFeeds(ctx)` which would involve a lot of duplicated logic in many places (esp. in the case of Terraform) assuming we want to capture all requests and responses.

We've been able to debug various issues with a very similar pattern that is already part of AWS Go SDK, but I'm open to alternative solutions/suggestions to solve that problem.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -99,6 +99,9 @@ const (
 	mediaTypeBlockUsersPreview = "application/vnd.github.giant-sentry-fist-preview+json"
 )
 
+type onRequestFunc func(r *http.Request)
+type onResponseFunc func(r *http.Response)
+
 // A Client manages communication with the GitHub API.
 type Client struct {
 	clientMu sync.Mutex   // clientMu protects the client during calls that modify the CheckRedirect func.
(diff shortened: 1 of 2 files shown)


shurcooL:
> We've been able to debug various issues with a very similar pattern that is already part of AWS Go SDK, but I'm open to alternative solutions/suggestions to solve that problem.

Have you considered a logging `http.RoundTripper` middleware? As far as I know, that's the most general way these things are done. As long as there's a way to specify the `*http.Client` being used, which there is, you can pass a custom `*http.Client`, or modify the current one being passed to include an `http.RoundTripper` that performs the `OnRequest` and `OnResponse` you want.

How does that sound to you?

radeksimko:
I did consider this option briefly and I originally thought I'd have to reimplement the default transport or the upstream oauth2 one, but it seems they're both mutable and passthrough behaviour is actually quite easy to implement, so it is a sufficient & elegant solution.

Thanks.

PR #635, opened 2017-05-09
Add pagination option to PullRequests.ListReviews.

sminnee:
This adds a ListOptions argument to ListReviews, so that pagination is
able to be supported. Without this it’s not possible to fetch the
2nd page of reviews, from what I can see.

Note that this does break the API, but I’ve added support for a nil
value in the argument, to minimise upgrade pain.

### DIFF
--- a/github/authorizations.go
+++ b/github/authorizations.go
@@ -343,8 +343,13 @@ func (s *AuthorizationsService) Revoke(ctx context.Context, clientID string, tok
 // tokens an application has generated for the user.
 //
 // GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#list-your-grants
-func (s *AuthorizationsService) ListGrants(ctx context.Context) ([]*Grant, *Response, error) {
-	req, err := s.client.NewRequest("GET", "applications/grants", nil)
+func (s *AuthorizationsService) ListGrants(ctx context.Context, opt *ListOptions) ([]*Grant, *Response, error) {
+	u, err := addOptions("applications/grants", opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/pulls_reviews.go
+++ b/github/pulls_reviews.go
@@ -149,9 +149,13 @@ func (s *PullRequestsService) DeletePendingReview(ctx context.Context, owner, re
 // returned error format and remove this comment once it's fixed.
 // Read more about it here - https://github.com/google/go-github/issues/540
 //
-// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments
-func (s *PullRequestsService) ListReviewComments(ctx context.Context, owner, repo string, number, reviewID int) ([]*PullRequestComment, *Response, error) {
+// GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-comments-for-a-single-review
+func (s *PullRequestsService) ListReviewComments(ctx context.Context, owner, repo string, number, reviewID int, opt *ListOptions) ([]*PullRequestComment, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/reviews/%d/comments", owner, repo, number, reviewID)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
--- a/github/pulls_reviews.go
+++ b/github/pulls_reviews.go
@@ -65,8 +65,12 @@ func (r PullRequestReviewDismissalRequest) String() string {
 // Read more about it here - https://github.com/google/go-github/issues/540
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request
-func (s *PullRequestsService) ListReviews(ctx context.Context, owner, repo string, number int) ([]*PullRequestReview, *Response, error) {
+func (s *PullRequestsService) ListReviews(ctx context.Context, owner, repo string, number int, opt *ListOptions) ([]*PullRequestReview, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/reviews", owner, repo, number)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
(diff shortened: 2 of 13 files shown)


sminnee:
I signed it!

github/pulls_reviews.go
```diff
@@ -65,9 +65,17 @@ func (r PullRequestReviewDismissalRequest) String() string {
 // Read more about it here - https://github.com/google/go-github/issues/540
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request
-func (s *PullRequestsService) ListReviews(ctx context.Context, owner, repo string, number int) ([]*PullRequestReview, *Response, error) {
+func (s *PullRequestsService) ListReviews(ctx context.Context, owner, repo string, number int, opt *ListOptions) ([]*PullRequestReview, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/pulls/%d/reviews", owner, repo, number)
 
+	if opt != nil {
```
shurcooL: No need to check that opt != nil, because `addOptions` handles that case.

See https://github.com/google/go-github/blob/c03cf196876fc0b7cf8db2516848536778ea4e29/github/github.go#L182-L184.

Please use the same style as all other places that call `addOptions`. You can use [`RepositoriesService.ListCommits`](https://github.com/google/go-github/blob/c03cf196876fc0b7cf8db2516848536778ea4e29/github/repos_commits.go#L114-L117) as reference:

```Go
u := fmt.Sprintf("repos/%v/%v/pulls/%d/reviews", owner, repo, number)
u, err := addOptions(u, opt)
if err != nil {
	return nil, nil, err
}
```

sminnee:
Thanks will tweak it

sminnee:
@shurcooL I've pushed an initial commit for #636 — I haven't updated the tests yet, however. Apart from the tests does the approach look reasonable?

github/authorizations.go
```diff
@@ -343,8 +343,14 @@ func (s *AuthorizationsService) Revoke(ctx context.Context, clientID string, tok
 // tokens an application has generated for the user.
 //
 // GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#list-your-grants
-func (s *AuthorizationsService) ListGrants(ctx context.Context) ([]*Grant, *Response, error) {
-	req, err := s.client.NewRequest("GET", "applications/grants", nil)
+func (s *AuthorizationsService) ListGrants(ctx context.Context, opt *ListOptions) ([]*Grant, *Response, error) {
+	u := "applications/grants"
```
shurcooL: Since the value of `u` here is just a constant, inline it into the next line:

```Go
u, err := addOptions("applications/grants", opt)
if err != nil {
	return nil, nil, err
}
```

It makes sense to keep it separate only when it's using `fmt.Sprintf`.

See existing code for precedent.

github/licenses.go
```diff
@@ -61,8 +61,14 @@ func (l License) String() string {
 // List popular open source licenses.
 //
 // GitHub API docs: https://developer.github.com/v3/licenses/#list-all-licenses
-func (s *LicensesService) List(ctx context.Context) ([]*License, *Response, error) {
-	req, err := s.client.NewRequest("GET", "licenses", nil)
+func (s *LicensesService) List(ctx context.Context, opt *ListOptions) ([]*License, *Response, error) {
+	u := "licenses"
```
shurcooL: Also here.

github/users.go
```diff
@@ -172,8 +172,14 @@ func (s *UsersService) ListAll(ctx context.Context, opt *UserListOptions) ([]*Us
 // authenticated user.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/invitations/#list-a-users-repository-invitations
-func (s *UsersService) ListInvitations(ctx context.Context) ([]*RepositoryInvitation, *Response, error) {
-	req, err := s.client.NewRequest("GET", "user/repository_invitations", nil)
+func (s *UsersService) ListInvitations(ctx context.Context, opt *ListOptions) ([]*RepositoryInvitation, *Response, error) {
+	u := "user/repository_invitations"
```
shurcooL: And here.

sminnee:
OK I have:

 * Added / fixed tests
 * Corrected the code style points you raised
 * Fixed `go fmt`

And got the green tick ;-)

github/licenses.go
```diff
@@ -61,8 +61,13 @@ func (l License) String() string {
...
 // GitHub API docs: https://developer.github.com/v3/licenses/#list-all-licenses
-func (s *LicensesService) List(ctx context.Context) ([]*License, *Response, error) {
-	req, err := s.client.NewRequest("GET", "licenses", nil)
+func (s *LicensesService) List(ctx context.Context, opt *ListOptions) ([]*License, *Response, error) {
+	u, err := addOptions("licenses", opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
```
shurcooL: You should leave `LicensesService.List` unmodified. See https://github.com/google/go-github/issues/636#issuecomment-300674337 for rationale.

github/pulls_reviews.go
```diff
@@ -146,8 +150,12 @@ func (s *PullRequestsService) DeletePendingReview(ctx context.Context, owner, re
 // Read more about it here - https://github.com/google/go-github/issues/540
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments
```
shurcooL: If you don't mind (because it's outside the scope), can you fix this GitHub API docs link? It should be:

https://developer.github.com/v3/pulls/reviews/#get-comments-for-a-single-review

github/users.go
```diff
@@ -172,8 +172,13 @@ func (s *UsersService) ListAll(ctx context.Context, opt *UserListOptions) ([]*Us
 // authenticated user.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/invitations/#list-a-users-repository-invitations
-func (s *UsersService) ListInvitations(ctx context.Context) ([]*RepositoryInvitation, *Response, error) {
-	req, err := s.client.NewRequest("GET", "user/repository_invitations", nil)
+func (s *UsersService) ListInvitations(ctx context.Context, opt *ListOptions) ([]*RepositoryInvitation, *Response, error) {
```
shurcooL: See https://github.com/google/go-github/issues/636#issuecomment-300675601 about this endpoint.

sminnee: I'm going to leave this as is because I'm pretty sure it's correct.

github/pulls_reviews.go
sminnee: Will do.

github/licenses.go
sminnee: Will do.

sminnee:
OK — updated the API doc link and removed licenses.go and licenses_test.go

shurcooL:
Did you amend the last commit and force push? For future, it'd be better if you hadn't, it's easier to review when I can see only what has changed since my last review.

sminnee:
Got a reply from GitHub:

Thanks for the report, Sam -- looks like a bug in the documentation, that endpoint should indeed support pagination. I'll ask the team to update the documentation.

shurcooL:
Great! I'll merge this soon.

shurcooL:
Thanks @sminnee for this original PR, fully investigating #636, and increasing the scope of this PR to resolve all of #636!

Issue #236, opened 2015-09-22
IssuesService.Edit cannot clear an issue's milestone


tchap:
Much like #181 I fail to see how to use go-github to clear an issue's milestone. The API requires the milestone be to set to `null` in the request so that it is cleared, but that is not possible since the field uses `omitempty`.

Not sure what the right solution would be, perhaps a standalone method for clearing milestones?

tbruyelle:
Is there a workaround for this ? 
If no I think a dedicated method is the way to go, something like `RemoveMilestonelForIssue`, I can work on this.

tchap:
The workaround is to send the request manually. You can basically copy `EditMilestone` and put your own struct there.

tbruyelle:
Indeed, we can workaround it like that (tested) : 

```go
       u := fmt.Sprintf("repos/%v/%v/issues/%d", owner, repo, issue)
       req, err := githubCli.NewRequest("PATCH", u, &struct {
               Milestone interface{} `json:"milestone"`
       }{})
       if err != nil {
               return err
       }
       _, err = githubCli.Do(req, nil)
       if err != nil {
               return err
       }
```

e-beach:
I'm surprised this hasn't been addressed in 2 years, though it's clear it's a problem as @marun has mentioned. Should we add a `removeMilestone` method to the `IssuesService`?

shurcooL:
I think that's the case because the GitHub API is very large, and removing a milestone from an issue is something only a few people ever need. It has a well described workaround here, so those who need it use that and move on.

To resolve the issue, we need to come up with a good API/good way of doing so.

alexellis:
Nice to see @elliott-beach again - this issue manifests as a bug with much head scratching and wasted time. Perhaps now that there are multiple consumers with this specific issue, we could fix this issue using this library. What are your thoughts 3 years on?

Alex

gmlewis:
It seems to me to be a relatively simple matter to resolve this issue by creating a new `RemoveMilestone` method with the [workaround above](https://github.com/google/go-github/issues/236#issuecomment-220560052).

Leaving this issue open.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

[...]

PR #637, opened 2017-05-10
[Issue#617] Update Protected Branches API preview

navarrano:
Adds pull request reviews enforcement structures and API methods, adds admin enforcement API methods and removes old IncludeAdmins field from the request.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -575,7 +575,7 @@ func (req PullRequestReviewsEnforcementRequest) MarshalJSON() ([]byte, error) {
 }
 
 // PullRequestReviewsEnforcementUpdate represents request to patch the pull request review
-// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcementUpdate above
+// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcementRequest above
 // because the patch request does not require all fields to be initialized.
 type PullRequestReviewsEnforcementUpdate struct {
 	// Specifies which users and teams can dismiss pull requets reviews. Can be ommitted.
@@ -804,7 +804,7 @@ func (s *RepositoriesService) License(ctx context.Context, owner, repo string) (
 //
 // GitHub API docs: https://developer.github.com/v3/repos/branches/#get-pull-request-review-enforcement-of-protected-branch
 func (s *RepositoriesService) GetPullRequestReviewEnforcement(ctx context.Context, owner, repo, branch string) (*PullRequestReviewsEnforcement, *Response, error) {
-	u := fmt.Sprintf("/repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
@@ -827,7 +827,7 @@ func (s *RepositoriesService) GetPullRequestReviewEnforcement(ctx context.Contex
 //
 // GitHub API docs: https://developer.github.com/v3/repos/branches/#update-pull-request-review-enforcement-of-protected-branch
 func (s *RepositoriesService) UpdatePullRequestReviewEnforcement(ctx context.Context, owner, repo, branch string, patch *PullRequestReviewsEnforcementUpdate) (*PullRequestReviewsEnforcement, *Response, error) {
-	u := fmt.Sprintf("/repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
 	req, err := s.client.NewRequest("PATCH", u, patch)
 	if err != nil {
 		return nil, nil, err
@@ -850,7 +850,7 @@ func (s *RepositoriesService) UpdatePullRequestReviewEnforcement(ctx context.Con
 //
 // GitHub API docs: https://developer.github.com/v3/repos/branches/#update-pull-request-review-enforcement-of-protected-branch
 func (s *RepositoriesService) DisableDismissalRestrictions(ctx context.Context, owner, repo, branch string) (*PullRequestReviewsEnforcement, *Response, error) {
-	u := fmt.Sprintf("/repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
 
 	data := struct {
 		R []interface{} `json:"dismissal_restrictions"`
@@ -877,7 +877,7 @@ func (s *RepositoriesService) DisableDismissalRestrictions(ctx context.Context, 
 //
 // GitHub API docs: https://developer.github.com/v3/repos/branches/#remove-pull-request-review-enforcement-of-protected-branch
 func (s *RepositoriesService) RemovePullRequestReviewEnforcement(ctx context.Context, owner, repo, branch string) (*Response, error) {
-	u := fmt.Sprintf("/repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
 	req, err := s.client.NewRequest("DELETE", u, nil)
 	if err != nil {
 		return nil, err
@@ -893,7 +893,7 @@ func (s *RepositoriesService) RemovePullRequestReviewEnforcement(ctx context.Con
 //
 // GitHub API docs: https://developer.github.com/v3/repos/branches/#get-admin-enforcement-of-protected-branch
 func (s *RepositoriesService) GetAdminEnforcement(ctx context.Context, owner, repo, branch string) (*AdminEnforcement, *Response, error) {
-	u := fmt.Sprintf("/repos/%v/%v/branches/%v/protection/enforce_admins", owner, repo, branch)
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/enforce_admins", owner, repo, branch)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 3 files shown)


github/repos.go
```diff
@@ -509,37 +509,49 @@ type Branch struct {
...
+	DismissalRestrictions *DismissalRestrictions `json:"dismissal_restrictions"`
+	DismissStaleReviews   bool                   `json:"dismiss_stale_reviews"`
+}
+
+// PullRequestReviewsEnforcementRequest represents request to set the pull request review
+// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcement above
+// because the request structure is different from the response structure.
+type PullRequestReviewsEnforcementRequest struct {
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+	DismissStaleReviews          bool                          `json:"dismiss_stale_reviews"`
```
shurcooL: Can you document the two fields?

```Go
// Specify which users and teams can dismiss pull request reviews. (Required.)
DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`

// Dismiss approved reviews automatically when a new commit is pushed. (Required.)
```

github/repos.go
```diff
@@ -509,37 +509,49 @@ type Branch struct {
...
+}
+
+// PullRequestReviewsEnforcementPatchRequest represents request to patch the pull request review
+// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcementPatchRequest above
+// because the patch request does not require all fields to be initialized.
+type PullRequestReviewsEnforcementPatchRequest struct {
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
+	DismissStaleReviews          *bool                         `json:"dismiss_stale_reviews,omitempty"`
```
shurcooL: Here as well. Except, the fields are not required here.

github/repos.go
```diff
@@ -509,37 +509,49 @@ type Branch struct {
...
+type PullRequestReviewsEnforcement struct {
+	DismissalRestrictions *DismissalRestrictions `json:"dismissal_restrictions"`
+	DismissStaleReviews   bool                   `json:"dismiss_stale_reviews"`
+}
+
+// PullRequestReviewsEnforcementRequest represents request to set the pull request review
+// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcement above
+// because the request structure is different from the response structure.
+type PullRequestReviewsEnforcementRequest struct {
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
```
shurcooL: Given this is a required field here, what do you think about using a value `DismissalRestrictionsRequest` rather than pointer `*DismissalRestrictionsRequest` (similar to `bool` below being a value rather than pointer).

Or is it neccessary to have a pointer to be able to send a `null` value?

github/repos.go
```diff
@@ -509,37 +509,49 @@ type Branch struct {
...
+	DismissStaleReviews          bool                          `json:"dismiss_stale_reviews"`
+}
+
+// PullRequestReviewsEnforcementPatchRequest represents request to patch the pull request review
+// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcementPatchRequest above
+// because the patch request does not require all fields to be initialized.
+type PullRequestReviewsEnforcementPatchRequest struct {
```
shurcooL: If it turns out we really do need to keep this struct after all, I'd suggest renaming it to `PullRequestReviewsEnforcementUpdate`. How's that sound?

github/repos.go
navarrano: True, I believe I overthought it. 

> You can disable dismissal restrictions by passing the dismissal_restrictions object as an empty array.

I will clarify it yet with GitHub support, since other structures allow `null` for disabling

> You must pass the following objects: required_status_checks, enforce_admins, and restrictions. They can have the value null for disabled.

In this case, the pointer would make sense and it would support disabling dismissal restrictions. If not, I'm not sure yet, how can I pass an empty array instead of empty object to disable it.

github/repos.go
navarrano: Will do and had so at the beginning, but I've noticed that Visual Studio Code `go fmt` plugin, does not align structure's fields to one line, if the fields have comments. So I thought that maybe commenting fields is not welcome.

github/repos.go
navarrano: I'm fine with that. However, taking into account, that the HTTP request is `PATCH`, I thought it will be better to use such a name, so it will explicitly say, that not all fields are required. But I believe `PullRequestReviewsEnforcementUpdate` is also fine

navarrano:
I believe the difference between `must` and `can` comes from the difference between HTTP `PUT` request of the [Update branch protection](https://developer.github.com/v3/repos/branches/#update-branch-protection) and HTTP `PATCH` request of the [Update pull request reviews enforcment](https://developer.github.com/v3/repos/branches/#update-branch-protection).

I wrote an e-mail to GitHub support to clarify those two endpoints

github/repos.go
shurcooL: They're certainly welcome. See some other structs in this file that have documented fields. The gofmt formatting is fine. Thanks.

github/repos.go
shurcooL: The HTTP method is indeed `PATCH`, but that's a low level implementation detail. The endpoint is called "Update pull request review enforcement of protected branch", which is where my suggestion of using "Update" comes from.

It's still not ideal, because "Request" is used for an endpoint called "Update branch protection"... But still, I think one word is better than "PatchRequest".

navarrano:
@shurcooL: 
After exchanging some e-mails with a GitHub support, I know, that we require two different structures for `Update branch protection` and `Update pull request review`. The difference between required and not required is intended. 

However, I'm still stuck at one point. For update branch protection, it is easy to say "if one pass nil value for `dismissal_restrictions`, he wants to disable them so we convert it to an empty array.
However, the same option should be available with the `Update pull request review` and here comes my design problem, because on the one hand we should be able to ignore `dismissal_restrictions` (that's why `omitempty`), but on the other hand, we should be able to notify GitHub that one wants to disable them and convert it to an empty array.

shurcooL:
The problem you're describing sounds very similar to https://github.com/google/go-github/issues/236.

navarrano:
@shurcooL: Thank you. I took your advise and implemented additional API method. I believe, my change set is ready.

github/repos.go
```diff
@@ -534,24 +535,53 @@ type RequiredStatusChecks struct {
...
+	}
+	newReq := struct {
+		R *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+		D bool                          `json:"dismiss_stale_reviews"`
+	}{
+		R: req.DismissalRestrictionsRequest,
+		D: req.DismissStaleReviews,
+	}
+	return json.Marshal(newReq)
+}
```
shurcooL: This is a really cool solution to work around the GitHub API weirdness. Nice idea! 👍

github/repos_test.go
```diff
@@ -772,6 +773,43 @@ func TestRepositoriesService_UpdatePullRequestReviewEnforcement(t *testing.T) {
...
+	mux.HandleFunc("/repos/o/r/branches/b/protection/required_pull_request_reviews", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PATCH")
+		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
+
+		buf := new(bytes.Buffer)
+		buf.ReadFrom(r.Body)
+		body := strings.TrimSpace(buf.String())
+		if wantBody != body {
+			t.Errorf("Request body = %+v, want %+v", body, wantBody)
+		}
```
shurcooL: You can use `testBody` helper. See how other tests use it.

github/repos_test.go
```diff
@@ -852,3 +852,19 @@ func TestRepositoriesService_RemoveAdminEnforcement(t *testing.T) {
...
+	req := PullRequestReviewsEnforcementRequest{}
+
+	json, err := json.Marshal(req)
+
+	if err != nil {
+		t.Errorf("PullRequestReviewsEnforcementRequest.MarshalJSON returned error: %v", err)
+	}
+
+	want := `{"dismissal_restrictions":[],"dismiss_stale_reviews":false}`
+
```
shurcooL: This blank line can be removed too.

github/repos_test.go
```diff
@@ -852,3 +852,19 @@ func TestRepositoriesService_RemoveAdminEnforcement(t *testing.T) {
 		t.Errorf("Repositories.RemoveAdminEnforcement returned error: %v", err)
 	}
 }
+
+func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctions(t *testing.T) {
+	req := PullRequestReviewsEnforcementRequest{}
+
+	json, err := json.Marshal(req)
+
+	if err != nil {
```
shurcooL: Remove the blank lines between `foo, err := something()` and `if err != nil {`. They hurt readability.

github/repos.go
```diff
@@ -509,37 +510,78 @@ type Branch struct {
...
+
+// PullRequestReviewsEnforcementUpdate represents request to patch the pull request review
+// enforcement of a protected branch. It is separate from PullRequestReviewsEnforcementUpdate above
```
gmlewis: s/PullRequestReviewsEnforcementUpdate/PullRequestReviewsEnforcementRequest/

shurcooL: @gmlewis The reason it's not already `PullRequestReviewsEnforcementRequest` is because that name is already taken. See 3rd paragraph of https://github.com/google/go-github/pull/637#pullrequestreview-37740813 and https://github.com/google/go-github/pull/637#pullrequestreview-37742570 for previous discussion about this.

gmlewis: Let X = "PullRequestReviewsEnforcementUpdate"

This section says:

// X represents... It is separate from X above ...
type X struct {

I think the second X needs to be replaced.  The first and third X are fine.

shurcooL: Ah, I see what you mean now. 👍 Indeed.

github/repos.go
```diff
@@ -738,3 +799,153 @@ func (s *RepositoriesService) License(ctx context.Context, owner, repo string) (
 
 	return r, resp, nil
 }
+
+// GetPullRequestReviewEnforcement gets pull request review enforcement of a protected branch.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#get-pull-request-review-enforcement-of-protected-branch
+func (s *RepositoriesService) GetPullRequestReviewEnforcement(ctx context.Context, owner, repo, branch string) (*PullRequestReviewsEnforcement, *Response, error) {
+	u := fmt.Sprintf("/repos/%v/%v/branches/%v/protection/required_pull_request_reviews", owner, repo, branch)
```
shurcooL: An important issue I've spotted is that all 7 of these URLs begin with a `/`, they should not (it breaks support for GitHub Enterprise).

```Go
u := fmt.Sprintf("repos/...`
```

github/repos_test.go
navarrano: Thank you for the tip, it made the test method definitely nicer.

github/repos.go
navarrano: Thank you for spotting it, no idea how I came with an idea of starting the URL with a slash... 
Should be alright now

github/repos.go
```diff
@@ -738,3 +799,153 @@ func (s *RepositoriesService) License(ctx context.Context, owner, repo string) (
...
+	r := new(AdminEnforcement)
+	resp, err := s.client.Do(ctx, req, r)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, err
+}
+
+// RemoveAdminEnforcement removess admin enforcement from a protected branch.
```
shurcooL: Minor, typo in "removess", should be "removes".

gmlewis:
Sorry for the delay!  LGTM. Merging.

PR #639, opened 2017-05-11
Simplify pagination example code.

shurcooL:
In [`RepositoryListByOrgOptions`](https://godoc.org/github.com/google/go-github/github#RepositoryListByOrgOptions) struct, [`ListOptions`](https://godoc.org/github.com/google/go-github/github#ListOptions) struct is an [embedded field](https://golang.org/ref/spec#Struct_types), so it's possible to dereference `Page` without explicitly going through `ListOptions`.

This is shorter, simpler and readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ for {
 	if resp.NextPage == 0 {
 		break
 	}
-	opt.ListOptions.Page = resp.NextPage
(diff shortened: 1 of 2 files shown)


gmlewis:
LGTM.
Merging.

Issue #638, opened 2017-05-11
Remove custom media type for Pull Request Reviews API


gmlewis:
The custom media type `application/vnd.github.black-cat-preview+json` can be removed from `go-github` as the "Pull Request Reviews API" is no longer in preview but has become official.

See GitHub Developer announcement:
https://developer.github.com/changes/2017-05-09-end-black-cat-preview/

varadarajana:
@gmlewis please assign this to me

varadarajana:
@gmlewis @shurcooL Pull Requests have three categories - Reviews, ReviewComments and Review Requests. Is this change required for all of them. The documentation points are reviews only.

shurcooL:
I think it's all endpoints that currently set the custom `application/vnd.github.black-cat-preview+json` Accept header. They no longer need to.

swsnider:
This change breaks Enterprise users -- we don't have reviews out of preview yet :(.

shurcooL:
Thanks for reporting @swsnider.

Re-opening for further consideration.

How should we go about dealing with this, @gmlewis?

A short term solution is to revert this removal of the preview API.

However, a long term solution is deciding on our policy. Should latest `master` target GitHub.com 
API only, or GitHub.com and GitHub Enterprise APIs? The challenge with doing the latter from what I can see is that their API chances/announcements are more difficult to track down for Enterprise API. But maybe I haven't looked hard enough.

swsnider:
The problem is that even after the Enterprise version with this fixed is released, you're typically still at the mercy of your IT org to upgrade your instance in a timely manner, so this may not be soluble by this project :(.

shurcooL:
Right. So it might be the case that a better/more viable solution is to, unfortunately, use an older version of go-github via vendoring.

swsnider:
Yeah, that's what I'm doing for now

e-beach:
I think the "good-for-beginners" label could be removed from this,

gmlewis:
I had a talk with @willnorris about this, and I believe our resolution was that semantic versioning (#376) would be the best way to support GitHub Enterprise API users (and that they would need to vendor the version that best supported their use-case), and that the latest `master` target the latest GitHub API.

As such, I think the "good-for-beginners" label still holds for this issue.

willnorris:
We can't help IT departments that are slow to upgrade, and vendoring (together with semver) is the right solution there.

[...]

PR #641, opened 2017-05-15
#638: Remove custom media type for Pull Request Reviews API

varadarajana:
@shurcooL I have made changes for the media type for the entire pulls reviews. Can you please review this?


shurcooL:
Look at #611 as an example of a PR that removes a custom media type that's no longer needed.

There's no need to set `mediaTypeV3` in individual methods, because it's already set when a request is created. See https://github.com/google/go-github/blob/4a42a1def4af554769703e8ac0ce5290c16371a2/github/github.go#L263.

varadarajana:
@shurcooL Sorry I missed the GitHub.go file where the normal headers are already set. Here is the corrected file.

Issue #627, opened 2017-04-27
Support new GitHub Apps (formerly Integrations) API.


gmlewis:
GitHub API Announcement:
https://developer.github.com/changes/2017-04-25-Integrations-Pre-Release/

vogxn:
This would be invaluable! The change for user-to-server requests appears to be adding the headers for the media type `application/vnd.github.machine-man-preview+json`.  I have not looked deep into the library but how difficult would this change be for a new contributor to delve into?

gmlewis:
Skimming the docs, this would be a fair amount of work to get this up and running, but would be a fantastic learning experience. :smiley:

gmlewis:
The Integrations API has now been deprecated in favor of the new GitHub Apps API.
This issue should now focus on the latest GitHub Developer API announcement:
https://developer.github.com/changes/2017-05-22-github-apps-production-ship/

shurcooL:
I think they were just renamed, not deprecated. But yes, it's called GitHub Apps now.

bradleyfalzon:
Just on this, the [Breaking Changes](https://developer.github.com/changes/2017-05-22-github-apps-production-ship/#breaking-changes) will need to be addressed, in particular, the integration specific events introduced in b0b867f2d5ec7c0139fad20ab0fe9267dcd8312b by @cyberdelia will need to be renamed. From my understanding, they are sending both events for the moment (and I can see they're doing this on my own installations, receiving `installation_repositories` and `integration_installation_repositories` as well as `installation` and `integration_installation`).

I'm happy to take this on in time if other's aren't. The first question would be exactly how go-github wants to handle this transition, considering both deprecated and new events appear to be supported at the moment, is the first stage simply to implement the new name and remove the old later? Or just a breaking change to rename the events? 

> Skimming the docs, this would be a fair amount of work to get this up and running, but would be a fantastic learning experience.

In case you hadn't seen it, I've been maintaining and using https://github.com/bradleyfalzon/ghinstallation with go-github for some time quite successfully.

shurcooL:
> In case you hadn't seen it, I've been maintaining and using https://github.com/bradleyfalzon/ghinstallation with go-github for some time quite successfully.

Do you think it's ready to be considered for addition to the README, in its own section? If so, want to send a PR that does that (and update both doc.go and README.md)?

bradleyfalzon:
[...]

Issue #455, opened 2016-10-18
GitHub Integrations Authentication Workflow


bradleyfalzon:
Hi,

Apart from the addition in #436, the new GitHub Integrations (https://developer.github.com/early-access/integrations/) requires a new authentication workflow, different from OAuth (https://developer.github.com/early-access/integrations/authentication/ as well as https://developer.github.com/early-access/integrations/integrations-vs-oauth-applications/).

As many APIs require this installation token (unique per installation of the integration, expires after short time period), I can easily (and do) provide a separate `http.Client` with a custom `http.RoundTripper` adding the necessary authentication headers, but, this requires a separate `http.Client` for every Integration installation, which means the underlying tcp connection is not being reused between installations (although I believe I can workaround that).

Further there is likely some room here to provide helpers for retrieving the required JWT tokens from a file as this is currently non-trivial and error prone (although just an example would be sufficient for most people).

I haven't developed a GitHub OAuth application before, so I'm not certain if the Integrations flow is any different, but ideally, I would like to be able to use a single client for all installations, swapping out the token on a per request basis - but I don't see how this could be possible without breaking api or not being thread safe (although I'm not certain on the safety guarantees already in place - I did look and couldn't see).

I haven't thought this through completely, and a separate client for each Integration installation maybe optimal (create another client, cache it for the life of the token), but I wanted to ask first if there has been some thought about how `go-github` would support with the new GitHub Integrations.

I have working prototype of a GitHub Integration and can provide samples if any maintainers want a quick start if no-one has started yet.

My current thinking (half tested):
- Single `http.Transport` - ensures all derived clients reuse http connection.
- Wrap that transport for each installation, the `installationTransport` adds necessary authorisation headers and is responsible for renewing tokens if they expire.
- Pass the `installationTransport` into a `http.Client` into a `github.Client`.

Thanks

bradleyfalzon:
So to provide this via a http.RoundTripper, I've done:

```go
// installationTransport provides a http.RoundTripper by wrapping an existing
// http.RoundTripper (that's shared between multiple installation transports to
// reuse underlying http connections), but provides GitHub Integration
// authentication as an installation.
//
// See https://developer.github.com/early-access/integrations/authentication/#as-an-installation
type installationTransport struct {
        tr             http.RoundTripper // tr is the underlying roundtripper being wrapped
        keyFile        string            // keyFile is the path to GitHub Intregration's PEM encoded private key
[...]

PR #649, opened 2017-05-26
Add link to GitHub Apps authentication package in docs.

bradleyfalzon:
As discussed in #627 I have a package which provides the GitHub Apps Authentication as an Installation `http.RoundTripper` that I use with this GitHub library successfully. I'm happy to keep maintaining it in an external repository, but a link from this library may assist others in finding the package.

Relates to #455.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,26 @@ See the [oauth2 docs][] for complete instructions on using that library.
 For API methods that require HTTP Basic Authentication, use the
 [`BasicAuthTransport`](https://godoc.org/github.com/google/go-github/github#BasicAuthTransport).
 
+GitHub Apps authentication can be provided by the [ghinstallation](https://github.com/bradleyfalzon/ghinstallation)
+package.
+
+```go
+import "github.com/bradleyfalzon/ghinstallation"
+
+func main() {
+  // Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
+  itr, err := ghinstallation.NewKeyFromFile(http.DefaultTransport, 1, 99, "2016-10-19.private-key.pem")
+  if err != nil {
+    // Handle error.
+  }
+
+  // Use installation transport with client.
+  client := github.NewClient(&http.Client{Transport: itr})
+
+  // Use client...
+}
+```
+
 ### Rate Limiting ###
 
 GitHub imposes a rate limit on all API clients. Unauthenticated clients are
--- a/github/doc.go
+++ b/github/doc.go
@@ -63,6 +63,24 @@ See the oauth2 docs for complete instructions on using that library.
 For API methods that require HTTP Basic Authentication, use the
 BasicAuthTransport.
 
+GitHub Apps authentication can be provided by the
+https://github.com/bradleyfalzon/ghinstallation package.
+
+	import "github.com/bradleyfalzon/ghinstallation"
+
+	func main() {
+		// Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
+		itr, err := ghinstallation.NewKeyFromFile(http.DefaultTransport, 1, 99, "2016-10-19.private-key.pem")
+		if err != nil {
+			// Handle error.
+		}
+
+		// Use installation transport with client
+		client := github.NewClient(&http.Client{Transport: itr})
+
+		// Use client...
+	}
+
 Rate Limiting
 
 GitHub imposes a rate limit on all API clients. Unauthenticated clients are


shurcooL:
I still feel it'd be nice to replace the period with a colon and add a simple basic usage example here, in this README. Just like there's a short example for `oauth2` above.

You said:

> I've sent through a minor one-liner, as I did begin to write a larger section which mostly just duplicated the docs already in https://github.com/bradleyfalzon/ghinstallation, so just pared it back to a minor note. Let me know in the PR if you'd like me to expand upon the docs though, happy to.

The way I see it, in this README, just include a basic usage example for people to get started. But leave advanced usage, details, explanations in the ghinstallation repo README.

Just like for oauth2, this repo README constrains a basic snippet for getting started, but oauth2 package docs contain more thorough detailed docs.

bradleyfalzon:
Sure, I can add a simple example.

Is this OK:

```
// Shared transport to reuse TCP connections.
tr := &http.Transport{}

// Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
itr, err := ghinstallation.NewKeyFromFile(tr, 1, 99, "2016-10-19.private-key.pem")

// Use installation transport with new client.
client := github.NewClient(http.Client{Transport: itr})
```

I can make it inline the `*http.Transport` if you like? The oauth2 example is a complete working program, would you prefer that too?

shurcooL:
I think it'd be helpful to include the import statement. It makes it easier to see the whole thing at a glance. Otherwise it's extra work and uncertainty for people to figure out the import path (it could be a vanity import path, so it requires extra steps to find out).

You can stop after creating a GitHub client, no need to call methods on it. Maybe just a `// Use client...` comment.

Also, why do you suggest creating a separate `http.RoundTripper` for it, instead of just using [`http.DefaultTransport`](https://godoc.org/net/http#DefaultTransport)? The zero value of `http.Transport{}` doesn't seem to have good defaults.

bradleyfalzon:
Sure, updated.

> Also, why do you suggest creating a separate http.RoundTripper for it, instead of just using http.DefaultTransport? The zero value of http.Transport{} doesn't seem to have good defaults.

Ah must just be an oversight, GopherCI uses DefaultTransport (https://github.com/bradleyfalzon/gopherci/blob/03badd2364a4a3e85c0e4389c88ca86b7e2f7a70/internal/github/github.go#L38), so I'll modify that example.

README.md
```diff
@@ -74,6 +74,26 @@ See the [oauth2 docs][] for complete instructions on using that library.
...
 [`BasicAuthTransport`](https://godoc.org/github.com/google/go-github/github#BasicAuthTransport).
 
+GitHub Apps authentication can be provided by [ghinstallation](https://github.com/bradleyfalzon/ghinstallation).
+
+```go
+import "github.com/bradleyfalzon/ghinstallation"
+
+func main() {
+  // Shared transport to reuse TCP connections.
+  tr := http.DefaultTransport
```
shurcooL: I think you can delete this line and its comment, and pass `http.DefaultTransport` directly to the `ghinstallation.NewKeyFromFile` call below.

It's pretty common knowledge that transports should be reused, unless there are specific reasons not to. I don't think you need to elaborate on that here. This is meant to be a minimal simple example to get started quickly.

README.md
```diff
@@ -74,6 +74,26 @@ See the [oauth2 docs][] for complete instructions on using that library.
...
+
+  // Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
+  itr, err := ghinstallation.NewKeyFromFile(tr, 1, 99, "2016-10-19.private-key.pem")
+  if err != nil {
+    log.Fatal(err)
```
shurcooL: Replace with `// handle error` to leave error handling up to user. We don't want to suggest bad practices, and doing better error handling can be verbose. It's common for usage examples to not elaborate on error handling, just that it should be done.

README.md
```diff
@@ -74,6 +74,26 @@ See the [oauth2 docs][] for complete instructions on using that library.
...
+  }
+
+  // Use installation transport with client
+  client := github.NewClient(http.Client{Transport: itr})
+}
```
shurcooL: Replace `}` with `// use client...`. This is an incomplete program anyway (it's missing a `package` clause), so leaving out the closing `}` is okay. I don't see why you'd want to include `}` and make it seem like `client` isn't used at all.

bradleyfalzon: Yeah, but I was half copying the style of oauth2, I'll make this change, but let me know if you want to make further edits there for consistency.

README.md
bradleyfalzon: Sure, done.

README.md
bradleyfalzon: Agreed, done.

README.md
```diff
@@ -74,6 +74,22 @@ See the [oauth2 docs][] for complete instructions on using that library.
...
 
+GitHub Apps authentication can be provided by [ghinstallation](https://github.com/bradleyfalzon/ghinstallation).
+
+```go
+import "github.com/bradleyfalzon/ghinstallation"
+
+func main() {
+  // Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
+  itr, err := ghinstallation.NewKeyFromFile(http.DefaultTransport, 1, 99, "2016-10-19.private-key.pem")
+  // Handle error.
```
shurcooL: I meant like this:

```Go
itr, err := ghinstallation.NewKeyFromFile(http.DefaultTransport, 1, 99, "2016-10-19.private-key.pem")
if err != nil {
	// Handle error.
}
```

I think it's a lot more readable at a glance this way, compared to just a plain text comment without the `if err != nil {` line.

github/doc.go
```diff
@@ -63,6 +63,21 @@ See the oauth2 docs for complete instructions on using that library.
...
+	func main() {
+		// Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
+		itr, err := ghinstallation.NewKeyFromFile(http.DefaultTransport, 1, 99, "2016-10-19.private-key.pem")
+		// Handle error.
+
+		// Use installation transport with client
+		client := github.NewClient(http.Client{Transport: itr})
+
+		// Use client...
+
```
shurcooL: Okay, I see how this looks kinda bad/unbalanced without the closing `}` for `func main() {` now. So let's add that back in. Keep the `// Use client...` comment of course.

README.md
```diff
@@ -74,6 +74,22 @@ See the [oauth2 docs][] for complete instructions on using that library.
 For API methods that require HTTP Basic Authentication, use the
 [`BasicAuthTransport`](https://godoc.org/github.com/google/go-github/github#BasicAuthTransport).
 
+GitHub Apps authentication can be provided by [ghinstallation](https://github.com/bradleyfalzon/ghinstallation).
```
shurcooL: Add the word "package" after ghinstallation maybe?

> GitHub Apps authentication can be provided by the [ghinstallation](https://github.com/bradleyfalzon/ghinstallation) package.

Notice how above it says:

> The easiest and recommended way to do this is using the [oauth2](https://github.com/golang/oauth2) **library**, ...

bradleyfalzon: Agreed, done.

README.md
bradleyfalzon: Of course, I wasn't thinking. Done.

github/doc.go
bradleyfalzon: Yeah sounds good. Done.

bradleyfalzon:
> are we okay with adding endorsement for the ghinstallation package in a separate repository?

The other option is to roll ghinstallation into go-github, I'd have no problem with that. Potentially with the idea to split the single Transport package into IntegrationTransport and InstallationTransport.

README.md
```diff
@@ -74,6 +74,26 @@ See the [oauth2 docs][] for complete instructions on using that library.
...
+
+func main() {
+  // Wrap the shared transport for use with the integration ID 1 authenticating with installation ID 99.
+  itr, err := ghinstallation.NewKeyFromFile(http.DefaultTransport, 1, 99, "2016-10-19.private-key.pem")
+  if err != nil {
+    // Handle error.
+  }
+
+  // Use installation transport with client.
+  client := github.NewClient(http.Client{Transport: itr})
```
shurcooL: Minor. Shouldn't this be `&http.Client`?

bradleyfalzon: Yes it should, fixed, and thanks.

shurcooL:
Friendly ping @gmlewis and @willnorris. I want to make sure you saw [my comment from 7 days ago](https://github.com/google/go-github/pull/649#pullrequestreview-40645150), in case it got buried by the minor typo discussion that followed.

gmlewis:
I'm fine with the endorsement, but before proceeding, I would like confirmation from @willnorris and I defer to his judgement. I'll ask him to review this.

shurcooL:
Great! Merging.

If there are any important updates or changes to the linked package, please let us know @bradleyfalzon. We're here to help. Thanks for creating it!

PR #650, opened 2017-05-28
Add Preview Suport To Conduct Api

### LINKED ISSUE #648: Support preview Codes of Conduct API
GitHub API announcement:
https://developer.github.com/changes/2017-05-23-coc-api/
GitHub Developers docs:
https://developer.github.com/v3/codes_of_conduct/

liubinyi:
Fixes #648 support review Codes of Conduct API

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -98,7 +98,7 @@ const (
 	// https://developer.github.com/changes/2017-02-09-community-health/
 	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview+json"
 
-	// https://developer.github.com/v3/codes_of_conduct/
+	// https://developer.github.com/changes/2017-05-23-coc-api/
 	mediaTypeCodesOfConductPreview = "application/vnd.github.scarlet-witch-preview+json"
 )
 
--- a/github/github.go
+++ b/github/github.go
@@ -97,6 +97,9 @@ const (
 
 	// https://developer.github.com/changes/2017-02-09-community-health/
 	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview+json"
+
+	// https://developer.github.com/v3/codes_of_conduct/
+	mediaTypeCodesOfConductPreview = "application/vnd.github.scarlet-witch-preview+json"
 )
 
 // A Client manages communication with the GitHub API.
--- a/github/misc.go
+++ b/github/misc.go
@@ -130,7 +130,7 @@ func (c *Client) GetCodeOfConduct(ctx context.Context, key string) (*CodeOfCondu
 	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
 
 	coc := new(CodeOfConduct)
-	resp, err := c.Do(ctx, req, &coc)
+	resp, err := c.Do(ctx, req, coc)
 	if err != nil {
 		return nil, resp, err
 	}
--- a/github/misc.go
+++ b/github/misc.go
@@ -83,22 +83,22 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response, 
 	return emoji, resp, nil
 }
 
-// Conduct represents metric that contains name, key, url, body.
-type Conduct struct {
+// CodeOfConduct represents a code of conduct
+type CodeOfConduct struct {
 	Name *string `json:"name,omitempty"`
 	Key  *string `json:"key,omitempty"`
 	URL  *string `json:"url,omitempty"`
-	BODY *string `json:"body,omitempty"`
+	Body *string `json:"body,omitempty"`
 }
 
-func (c *Conduct) String() string {
+func (c *CodeOfConduct) String() string {
 	return Stringify(c)
 }
 
 // ListCodesOfConduct returns all codes of conduct.
 //
-// https://developer.github.com/v3/codes_of_conduct/
-func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*Conduct, *Response, error) {
+// GitHub API docs: https://developer.github.com/v3/codes_of_conduct/
+func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*CodeOfConduct, *Response, error) {
 	req, err := c.NewRequest("GET", "codes_of_conduct", nil)
 	if err != nil {
 		return nil, nil, err
@@ -107,19 +107,19 @@ func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*Conduct, *Response,
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
 
-	var conducts []*Conduct
-	resp, err := c.Do(ctx, req, &conducts)
+	var codeOfConducts []*CodeOfConduct
+	resp, err := c.Do(ctx, req, &codeOfConducts)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return conducts, resp, nil
+	return codeOfConducts, resp, nil
 }
 
-// ListCodeOfConductByKey returns an individual code of conduct.
+// GetCodeOfConductByKey returns an individual code of conduct.
 //
 // https://developer.github.com/v3/codes_of_conduct/
-func (c *Client) ListCodeOfConductByKey(ctx context.Context, key string) (*Conduct, *Response, error) {
+func (c *Client) GetCodeOfConductByKey(ctx context.Context, key string) (*CodeOfConduct, *Response, error) {
 	u := fmt.Sprintf("codes_of_conduct/%s", key)
 	req, err := c.NewRequest("GET", u, nil)
 	if err != nil {
@@ -129,13 +129,13 @@ func (c *Client) ListCodeOfConductByKey(ctx context.Context, key string) (*Condu
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
 
-	conduct := new(Conduct)
-	resp, err := c.Do(ctx, req, &conduct)
+	codeOfConduct := new(CodeOfConduct)
+	resp, err := c.Do(ctx, req, &codeOfConduct)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return conduct, resp, nil
+	return codeOfConduct, resp, nil
 }
 
 // APIMeta represents metadata about the GitHub API.
--- a/github/misc.go
+++ b/github/misc.go
@@ -83,7 +83,7 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response, 
 	return emoji, resp, nil
 }
 
-// CodeOfConduct represents a code of conduct
+// CodeOfConduct represents a code of conduct.
 type CodeOfConduct struct {
 	Name *string `json:"name,omitempty"`
 	Key  *string `json:"key,omitempty"`
@@ -97,7 +97,7 @@ func (c *CodeOfConduct) String() string {
 
 // ListCodesOfConduct returns all codes of conduct.
 //
-// GitHub API docs: https://developer.github.com/v3/codes_of_conduct/
+// GitHub API docs: https://developer.github.com/v3/codes_of_conduct/#list-all-codes-of-conduct
 func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*CodeOfConduct, *Response, error) {
 	req, err := c.NewRequest("GET", "codes_of_conduct", nil)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


github/github.go
```diff
@@ -97,6 +97,9 @@ const (
 
 	// https://developer.github.com/changes/2017-02-09-community-health/
 	mediaTypeRepositoryCommunityHealthMetricsPreview = "application/vnd.github.black-panther-preview+json"
+
+	// https://developer.github.com/v3/codes_of_conduct/
```
shurcooL: This link isn't great, because the mention of preview API will eventually be removed.

Let's link to https://developer.github.com/changes/2017-05-23-coc-api/ here instead.

github/misc.go
```diff
@@ -83,6 +83,61 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
 	return emoji, resp, nil
 }
 
+// Conduct represents metric that contains name, key, url, body.
+type Conduct struct {
+	Name *string `json:"name,omitempty"`
+	Key  *string `json:"key,omitempty"`
+	URL  *string `json:"url,omitempty"`
+	BODY *string `json:"body,omitempty"`
```
shurcooL: `BODY` should be `Body`, it's a word, not initialism.

github/misc.go
```diff
@@ -83,6 +83,61 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
 	return emoji, resp, nil
 }
 
+// Conduct represents metric that contains name, key, url, body.
+type Conduct struct {
```
shurcooL: The name and description here seem off.

I think we should give it a more explicit and long name. `Conduct` isn't clear in context of all the types present in the package.

What do you think of `CodeOfConduct` instead?

The comment should probably say that `// CodeOfConduct represents a code of conduct.` or so.

github/misc.go
```diff
@@ -83,6 +83,61 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
...
+		return nil, resp, err
+	}
+
+	return conducts, resp, nil
+}
+
+// ListCodeOfConductByKey returns an individual code of conduct.
+//
+// https://developer.github.com/v3/codes_of_conduct/
+func (c *Client) ListCodeOfConductByKey(ctx context.Context, key string) (*Conduct, *Response, error) {
```
shurcooL: Oh, I forgot to leave a comment here, but you can call this method `GetCodeOfConduct`. "List" is for methods that return multiple entries, "Get" is for a single entry.

github/repos.go
```diff
@@ -305,6 +305,28 @@ func (s *RepositoriesService) Get(ctx context.Context, owner, repo string) (*Rep
 	return repository, resp, nil
 }
 
+// Get the contents of a repository's code of conduct.
```
shurcooL: `golint` will complain here, saying the doc string should begin with `// GetCodeOfConduct ...`, so you should write:

```Go
// GetCodeOfConduct gets the contents of a repository's code of conduct.
```

github/misc.go
```diff
@@ -83,6 +83,61 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
...
+	BODY *string `json:"body,omitempty"`
+}
+
+func (c *Conduct) String() string {
+	return Stringify(c)
+}
+
+// ListCodesOfConduct returns all codes of conduct.
+//
+// https://developer.github.com/v3/codes_of_conduct/
```
shurcooL: For consistency with our other endpoints, can you add a "GitHub API docs:" prefix to all the documentation URLs? Like so:

```Go
// GitHub API docs: https://developer.github.com/v3/codes_of_conduct/
```

Here, and in the method below where it's missing too.

github/github.go
liubinyi: yeah. updated

github/misc.go
liubinyi: yeah that sounds better. updated.

github/misc.go
liubinyi: updated

github/misc.go
liubinyi: updated

github/misc.go
liubinyi: i see. updated

github/repos.go
liubinyi: updated

liubinyi:
@shurcooL i updated per your comments. Let me know if you spotted other issues. Thanks for the suggestions!

github/misc.go
```diff
@@ -83,22 +83,22 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
 	return emoji, resp, nil
 }
 
-// Conduct represents metric that contains name, key, url, body.
-type Conduct struct {
+// CodeOfConduct represents a code of conduct
```
shurcooL: Minor, but the sentence is missing a period at the end.

github/misc.go
```diff
@@ -83,22 +83,22 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
...
-func (c *Conduct) String() string {
+func (c *CodeOfConduct) String() string {
 	return Stringify(c)
 }
 
 // ListCodesOfConduct returns all codes of conduct.
 //
-// https://developer.github.com/v3/codes_of_conduct/
-func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*Conduct, *Response, error) {
+// GitHub API docs: https://developer.github.com/v3/codes_of_conduct/
```
shurcooL: You can update the URL here to be more specific:

https://developer.github.com/v3/codes_of_conduct/#list-all-codes-of-conduct

github/misc.go
```diff
@@ -107,19 +107,19 @@ func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*Conduct, *Response,
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
 
-	var conducts []*Conduct
-	resp, err := c.Do(ctx, req, &conducts)
+	var codeOfConducts []*CodeOfConduct
+	resp, err := c.Do(ctx, req, &codeOfConducts)
```
shurcooL: Minor, but the plural is on the wrong word here. The method and API is called code**s** of conduct, yet this variable is code of conduct**s**.

You can make it much shorter, since it's being used in a very local context. Either `cs`, `codes`, or something like that.

Reference: https://github.com/golang/go/wiki/CodeReviewComments#variable-names.

github/misc.go
```diff
@@ -107,19 +107,19 @@ func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*Conduct, *Response,
...
-	return conducts, resp, nil
+	return codeOfConducts, resp, nil
 }
 
-// ListCodeOfConductByKey returns an individual code of conduct.
+// GetCodeOfConductByKey returns an individual code of conduct.
 //
 // https://developer.github.com/v3/codes_of_conduct/
-func (c *Client) ListCodeOfConductByKey(ctx context.Context, key string) (*Conduct, *Response, error) {
+func (c *Client) GetCodeOfConductByKey(ctx context.Context, key string) (*CodeOfConduct, *Response, error) {
```
shurcooL: Simplify it further to just `GetCodeOfConduct`. "ByKey" isn't needed, there's only one way to get a code of conduct, and that's by key.

github/misc.go
```diff
@@ -129,13 +129,13 @@ func (c *Client) ListCodeOfConductByKey(ctx context.Context, key string) (*Condu
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
 
-	conduct := new(Conduct)
-	resp, err := c.Do(ctx, req, &conduct)
+	codeOfConduct := new(CodeOfConduct)
```
shurcooL: I'd suggest using a shorter variable name here too. `coc := new(CodeOfConduct)` or `code := new(CodeOfConduct)` would be completely fine, and likely more readable.

github/misc_test.go
```diff
@@ -81,23 +81,23 @@ func TestListCodesOfConduct(t *testing.T) {
 						]`)
 	})
 
-	metrics, _, err := client.ListCodesOfConduct(context.Background())
+	codeOfConducts, _, err := client.ListCodesOfConduct(context.Background())
```
shurcooL: Longer name is okay here, but call it `codesOfConduct` for consistency.

github/misc.go
```diff
@@ -107,19 +107,19 @@ func (c *Client) ListCodesOfConduct(ctx context.Context) ([]*Conduct, *Response,
...
 	}
 
-	return conducts, resp, nil
+	return codeOfConducts, resp, nil
 }
 
-// ListCodeOfConductByKey returns an individual code of conduct.
+// GetCodeOfConductByKey returns an individual code of conduct.
 //
 // https://developer.github.com/v3/codes_of_conduct/
```
shurcooL: You can make the URL more specific:

https://developer.github.com/v3/codes_of_conduct/#get-an-individual-code-of-conduct

github/repos.go
```diff
@@ -305,10 +305,10 @@ func (s *RepositoriesService) Get(ctx context.Context, owner, repo string) (*Rep
 	return repository, resp, nil
 }
 
-// Get the contents of a repository's code of conduct.
+// GetCodeOfConduct gets the contents of a repository's code of conduct.
 //
 // GitHub API docs: https://developer.github.com/v3/codes_of_conduct/
```
shurcooL: URL can point to the very endpoint:

https://developer.github.com/v3/codes_of_conduct/#get-the-contents-of-a-repositorys-code-of-conduct

github/repos.go
```diff
@@ -318,13 +318,13 @@ func (s *RepositoriesService) GetCodeOfConduct(ctx context.Context, owner, repo
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
 
-	conduct := new(Conduct)
-	resp, err := s.client.Do(ctx, req, conduct)
+	codeOfconduct := new(CodeOfConduct)
```
shurcooL: `codeOfconduct` has lowercase second "c", instead of `codeOfConduct`. But you can use a shorter variable name here too.

github/misc.go
liubinyi: updated.

github/misc.go
liubinyi: like that. updated. I will be more careful next time.

github/misc.go
liubinyi: updated to cs

github/misc.go
liubinyi: updated

github/misc.go
liubinyi: updated. also i am curious what would be the better naming convention if there are more than one way to get a code of conduct.

GetCodeOfConductByKey(..., key string) or GetCodeOfConductBy(..., key string) (since key is required input)

github/misc.go
liubinyi: prefer using coc. updated.

github/misc_test.go
liubinyi: since cs was used in misc, i updated it to cs just to be more consistent.

github/repos.go
liubinyi: updated

github/repos.go
liubinyi: updated to coc for consistency

liubinyi:
@shurcooL all updated. Thanks for comments. That was good! Please let me know if you find other places that can be improved.

github/misc.go
shurcooL: > also i am curious what would be the better naming convention if there are more than one way to get a code of conduct

It happens rarely. But what we ended up doing in such a situation is to have:

- `GetFoo`.
- `GetFooByID`.

For such an example, see https://godoc.org/github.com/google/go-github/github#UsersService.Get and https://godoc.org/github.com/google/go-github/github#UsersService.GetByID.

github/repos.go
```diff
@@ -293,7 +293,7 @@ func (s *RepositoriesService) Get(ctx context.Context, owner, repo string) (*Rep
 
 	// TODO: remove custom Accept header when the license support fully launches
 	// https://developer.github.com/v3/licenses/#get-a-repositorys-license
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeSquashPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeSquashPreview, mediaTypeCodesOfConductPreview}
```
shurcooL: This is great. It means GitHub API will send a "code_of_conduct" key in the repository object response.

But you still need to add a ``CodeOfConduct `json:"code_of_conduct,omitempty"` `` to the [`Repository`](https://github.com/google/go-github/blob/9f4675c6fd826772a2fffe74b2a7b6c9822802a0/github/repos.go#L21) struct above.

Otherwise people won't be able to access it.

github/repos.go
```diff
@@ -75,6 +75,7 @@ type Repository struct {
 	AssigneesURL     *string `json:"assignees_url,omitempty"`
 	BlobsURL         *string `json:"blobs_url,omitempty"`
 	BranchesURL      *string `json:"branches_url,omitempty"`
+	CodeOfConduct    *string `json:"code_of_conduct,omitempty"`
```
shurcooL: Pretty sure you want the type to be `*CodeOfConduct`, not `*string`. ;)

This is the sample value in GitHub API docs:

image

shurcooL: Also, don't put it in the `// API URLs` section since it's not an API URL. Put it at the top with all the main fields... Maybe right under `Homepage` field?

liubinyi: yeah. pretty sure my hand is faster than my brain at that moment. Thanks for catching that typo!!! updated.

github/misc.go
```diff
@@ -83,6 +83,61 @@ func (c *Client) ListEmojis(ctx context.Context) (map[string]string, *Response,
...
+	req, err := c.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
+
+	coc := new(CodeOfConduct)
+	resp, err := c.Do(ctx, req, &coc)
```
shurcooL: Pretty minor, but inconsistent. `coc` is a pointer to `CodeOfConduct`. No need to take address of it, just do `c.Do(ctx, req, coc)`.

You already do that in the `RepositoriesService.GetCodeOfConduct` method, might as well be consistent. Vast majority of the code in this package passes a single pointer instead of a pointer to a pointer.

liubinyi: yep! updated Thank you!

PR #630, opened 2017-05-05
#569:3 Webhooks for OrgBlockEvent

varadarajana:
@shurcooL I have now been able to isolate only the changes for this file based on your inputs. Thank you for your inputs. I was struggling for long on how to remove this. Please review these changes.

Resolves #569.

PR #653, opened 2017-06-02
Add support for 'team' events

### LINKED ISSUE #652: Support TeamEvent
This event was announced in https://developer.github.com/changes/2016-11-29-user-management-webhooks/

The other events announced in that post (OrganizationEvent, MemberEvent) are already supported.

The API documentation is here:
https://developer.github.com/v3/activity/events/types/#teamevent

I'd like to tackle this one.

kytrinyx:
The team event was announced in this post:
https://developer.github.com/changes/2016-11-29-user-management-webhooks/

The API documentation is at:
https://developer.github.com/v3/activity/events/types/#teamevent

The "Changes" object in the response is not illustrated in
the example payload, but it is documented in the table of fields.

Closes #652.

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -165,6 +165,28 @@ type ProjectColumnChange struct {
 	} `json:"name,omitempty"`
 }
 
+// TeamChange represents the changes when a team has been edited.
+type TeamChange struct {
+	Description *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"description,omitempty"`
+	Name *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"name,omitempty"`
+	Privacy *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"privacy,omitempty"`
+	Repository *struct {
+		Permissions *struct {
+			From *struct {
+				Admin *string `json:"admin,omitempty"`
+				Pull  *string `json:"pull,omitempty"`
+				Push  *string `json:"push,omitempty"`
+			} `json:"from,omitempty"`
+		} `json:"permissions,omitempty"`
+	} `json:"repository,omitempty"`
+}
+
 // IntegrationInstallationEvent is triggered when an integration is created or deleted.
 // The Webhook event name is "integration_installation".
 //
@@ -651,6 +673,25 @@ type StatusEvent struct {
 	Repo         *Repository       `json:"repository,omitempty"`
 	Sender       *User             `json:"sender,omitempty"`
 	Installation *Installation     `json:"installation,omitempty"`
+}
+
+// TeamEvent is triggered when an organization's team is created, modified or deleted.
+// The Webhook event name is "team".
+//
+// Events of this type are not visible in timelines. These events are only used
+// to trigger hooks.
+//
+// GitHub API docs: https://developer.github.com/v3/activity/events/types/#teamevent
+type TeamEvent struct {
+	Action  *string     `json:"action,omitempty"`
+	Team    *Team       `json:"team,omitempty"`
+	Changes *TeamChange `json:"changes,omitempty"`
+	Repo    *Repository `json:"repository,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	Org          *Organization `json:"organization,omitempty"`
+	Sender       *User         `json:"sender,omitempty"`
+	Installation *Installation `json:"installation,omitempty"`
 }
 
 // TeamAddEvent is triggered when a repository is added to a team.
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -92,6 +92,8 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 		payload = &RepositoryEvent{}
 	case "StatusEvent":
 		payload = &StatusEvent{}
+	case "TeamEvent":
+		payload = &TeamEvent{}
 	case "TeamAddEvent":
 		payload = &TeamAddEvent{}
 	case "WatchEvent":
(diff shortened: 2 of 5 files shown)


github/event_types.go
```diff
@@ -653,6 +675,21 @@ type StatusEvent struct {
...
+// Events of this type are not visible in timelines. These events are only used
+// to trigger hooks.
+//
+// GitHub API docs: https://developer.github.com/v3/activity/events/types/#teamevent
+type TeamEvent struct {
+	Action  *string       `json:"action,omitempty"`
+	Team    *Team         `json:"team,omitempty"`
+	Changes *TeamChange   `json:"changes,omitempty"`
+	Org     *Organization `json:"organization,omitempty"`
+	Sender  *User         `json:"sender,omitempty"`
```
shurcooL: It looks like you're missing ``Repo *Repository `json:"repository,omitempty"` `` struct field. I see it's a part of Events API payload at https://developer.github.com/v3/activity/events/types/#events-api-payload-34:

image

Note this is different from the `repository` field inside `changes` field.

shurcooL: You should also include ``Installation *Installation     `json:"installation,omitempty"` `` here after `Sender`. It's implicitly needed for webhook payloads, according to:

> **In addition to the fields [documented for each event](https://developer.github.com/v3/activity/events/types/),** webhook payloads include the user who performed the event (`sender`) as well as the organization (`organization`) and/or repository (`repository`) which the event occurred on, and for a [GitHub App](https://developer.github.com/apps/)'s webhook **may include the installation (`installation`) which an event relates to.**

_(Source: https://developer.github.com/webhooks/#payloads.)_

Related to #505. /cc @bradleyfalzon FYI.

See events above and below for example. You can also see https://github.com/google/go-github/pull/630#discussion_r115077161 for precedent.

shurcooL: Finally, I think it'd be nice to separate the webhook-only fields from the events API ones, like this:

```Go
type TeamEvent struct {
	...

	Org     *Organization `json:"organization,omitempty"`
	Sender  *User         `json:"sender,omitempty"`
}
```

You can see the `TeamAddEvent` below for example.

github/event_types.go
```diff
@@ -165,6 +165,28 @@ type ProjectColumnChange struct {
...
+	} `json:"name,omitempty"`
+	Privacy *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"privacy,omitempty"`
+	Repository *struct {
+		Permissions *struct {
+			From *struct {
+				Admin *string `json:"admin,omitempty"`
+				Push  *string `json:"push,omitempty"`
+				Pull  *string `json:"pull,omitempty"`
```
shurcooL: This is _really_ minor, but the order in the documentation is `admin`, then `pull`, then `push`. Is there a specific reason you chose to swap the order of Push and Pull? If not, maybe it'd be better to stay consistent with the docs by swapping the two.

kytrinyx: Ah, I didn't notice that. I was going from most permissive to least, but it's better to follow the order of the documentation. I'll change that.

github/event_types.go
kytrinyx: Ah, I didn't know about the implicit fields. That's a great catch.

I didn't separate out the webhook events, because this is an event that is only used for webhooks—it doesn't appear in the timeline.

kytrinyx:
@shurcooL thanks for catching the missing fields. I've added them, and also separated out the implicit webhook fields.

github/event_types.go
shurcooL: > I was going from most permissive to least

I see, that makes sense. Probably even more sense than the order the docs have them in.

We could also re-arrange them to be `Pull`, `Push`, and `Admin` last. That way it's from least permissive to most, and consistent with:

- https://github.com/google/go-github/blob/ebfec748347a9af6793c723f8859afcd906860fb/github/repos_collaborators.go#L86-L88
- https://github.com/google/go-github/blob/ebfec748347a9af6793c723f8859afcd906860fb/github/orgs_teams.go#L263-L265

But I noticed that some sample responses in https://developer.github.com/v3/repos/ also have them in the alphabetical order:

```JSON
"permissions": {
  "admin": false,
  "push": false,
  "pull": true
},
```

So, there's really no single best way, they're all slightly inconsistent. I'll leave it to you to decide what to go with. I just wanted to bring your attention here and make sure the ordering choice was deliberate rather than accidental.

github/event_types.go
shurcooL: > I didn't separate out the webhook events, because this is an event that is only used for webhooks—it doesn't appear in the timeline.

As I understand, the event is still accessible via Events API, but it doesn't show up in the visual timelines. The way you have it now looks good to me.

kytrinyx:
> Let us know when this is ready from your side, and we can merge then.

@shurcooL This is ready from my side.

shurcooL:
@gmlewis Wanna look over the recent changes, and merge this if everything still looks good to you?

gmlewis:
LGTM.  Merging.

Thank you, @kytrinyx and @shurcooL!

github/event_types.go
```diff
@@ -165,6 +165,28 @@ type ProjectColumnChange struct {
...
+	} `json:"name,omitempty"`
+	Privacy *struct {
+		From *string `json:"from,omitempty"`
+	} `json:"privacy,omitempty"`
+	Repository *struct {
+		Permissions *struct {
+			From *struct {
+				Admin *string `json:"admin,omitempty"`
+				Pull  *string `json:"pull,omitempty"`
+				Push  *string `json:"push,omitempty"`
```
kytrinyx: Oh, shoot. These are not strings, they're bools. I'll submit a new PR fixing that.

gmlewis: Thanks for catching that, @kytrinyx!  Sorry I didn't catch it during review.

PR #656, opened 2017-06-13
Allow COC to be returned in lists

jamiehannaford:
This PR adds the experimental COC accept header to list calls

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -177,7 +177,7 @@ func (s *RepositoriesService) List(ctx context.Context, user string, opt *Reposi
 	}
 
 	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	var repos []*Repository
@@ -215,7 +215,7 @@ func (s *RepositoriesService) ListByOrg(ctx context.Context, org string, opt *Re
 	}
 
 	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	var repos []*Repository
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -19,7 +19,7 @@ func TestRepositoriesService_List_authenticatedUser(t *testing.T) {
 	setup()
 	defer teardown()
 
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	mux.HandleFunc("/user/repos", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", strings.Join(acceptHeaders, ", "))
@@ -41,7 +41,7 @@ func TestRepositoriesService_List_specifiedUser(t *testing.T) {
 	setup()
 	defer teardown()
 
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	mux.HandleFunc("/users/u/repos", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", strings.Join(acceptHeaders, ", "))
@@ -77,7 +77,7 @@ func TestRepositoriesService_List_specifiedUser_type(t *testing.T) {
 	setup()
 	defer teardown()
 
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	mux.HandleFunc("/users/u/repos", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", strings.Join(acceptHeaders, ", "))
@@ -110,7 +110,7 @@ func TestRepositoriesService_ListByOrg(t *testing.T) {
 	setup()
 	defer teardown()
 
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	mux.HandleFunc("/orgs/o/repos", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", strings.Join(acceptHeaders, ", "))


github/repos.go
```diff
@@ -245,6 +246,7 @@ func (s *RepositoriesService) ListAll(ctx context.Context, opt *RepositoryListAl
 	if err != nil {
 		return nil, nil, err
 	}
+	req.Header.Set("Accept", mediaTypeCodesOfConductPreview)
```
shurcooL: This line should have a blank line above it, and a `// TODO: remove custom Accept header when license support fully launches` comment for consistency.

jamiehannaford:
@shurcooL Feedback incorporated. I created a new test fn, let me know if you'd like that done another way.

github/repos_test.go
```diff
@@ -109,7 +109,8 @@ func TestRepositoriesService_ListByOrg(t *testing.T) {
 
 	mux.HandleFunc("/orgs/o/repos", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		testHeader(t, r, "Accept", mediaTypeLicensesPreview)
+		testInHeader(t, r, "Accept", mediaTypeLicensesPreview)
+		testInHeader(t, r, "Accept", mediaTypeCodesOfConductPreview)
```
shurcooL: This tests for presence of wanted media types in accept header, but not for absence of unwanted ones.

I'd suggest just using the existing `testHeader` helper. Similar to how it's done here:

https://github.com/google/go-github/blob/d20deacea658223895efa3600c28053e733a512f/github/repos_test.go#L233

jamiehannaford:
@shurcooL ping

shurcooL:
The code looks good.

However, the functionality doesn't seem to work. @jamiehannaford, have you tested this?

There are 2 different issues.

1. /repositories endpoint doesn't seem to support `mediaTypeCodesOfConductPreview` accept header at all.

    ```
    $ curl -s -H "Accept: application/vnd.github.scarlet-witch-preview+json" https://api.github.com/repositories | grep "code_of_conduct"
    $ 
    ```

    The announcement at https://developer.github.com/changes/2017-05-23-coc-api/, nor its docs at https://developer.github.com/v3/codes_of_conduct/ never suggest that /repositories endpoint is supported.

    The /orgs/:org/repos endpoint does seem to be supported. Also, /users/:username/repos and /users/repos. Did you mean to target those endpoints instead?

2. Even though /orgs/:org/repos endpoint supports the `mediaTypeCodesOfConductPreview` accept header individually, it doesn't seem to work alongside with its existing `mediaTypeLicensesPreview ` preview header for licenses.

    /cc @gmlewis I had my concerns about whether multiple preview headers can be passed at the same time, and this is a very vivid example of how it doesn't work. But I don't know if it's just these 2 headers that intersect poorly, or all others too.

    In my testing, you can have one, or the other, but not both:

    ```
    $ curl -s -H "Accept: application/vnd.github.scarlet-witch-preview+json" https://api.github.com/orgs/golang/repos?per_page=1 | grep "code_of_conduct"
        "code_of_conduct": {
    $ 
    ```

    ```
    $ curl -s -H "Accept: application/vnd.github.drax-preview+json" https://api.github.com/orgs/golang/repos?per_page=1 | grep "license"
        "license": {
          "url": "https://api.github.com/licenses/bsd-3-clause",
    $ 
    ```

    ```
    $ curl -s -H "Accept: application/vnd.github.drax-preview+json, application/vnd.github.scarlet-witch-preview+json" https://api.github.com/orgs/golang/repos?per_page=1 | grep "code_of_conduct"
    $ 
    ```

    Thoughts? Am I doing something wrong in my testing?

gmlewis:
@shurcooL, as far as I can tell, you are testing correctly. It might be time to reach out to GitHub support. Would you or @jamiehannaford mind doing that and reporting back their findings?

shurcooL:
I'll let @jamiehannaford spearhead the effort.

There's no point in asking them about /repositories endpoint, that's clearly not supported (and was never promised as such).

But it is worth asking them if it's possible to combine multiple preview headers, `application/vnd.github.scarlet-witch-preview+json` and `application/vnd.github.drax-preview+json` specifically, in order to opt-in for both the licenses and code of conduct previews at the same time (1 request).

I suspect the answer is that it's not possible and 2 separate requests must be done until at least one of the preview APIs becomes standard. But, it's worth a shot to ask and confirm.

jamiehannaford:
I've sent the GitHub support folks an email, will keep you updated.

shurcooL:
@jamiehannaford We've got a [reply](https://github.com/google/go-github/issues/670#issuecomment-315837388) from @kytrinyx at GitHub, letting us know that multiple preview headers are now supported (previously, they weren't).

I've just confirmed that's the case:

```
$ curl -s -H "Accept: application/vnd.github.drax-preview+json, application/vnd.github.scarlet-witch-preview+json" https://api.github.com/orgs/golang/repos?per_page=1 | tail -n 20
    "default_branch": "master",
    "permissions": {
      "admin": false,
      "push": false,
      "pull": true
    },
    "license": {
      "key": "bsd-3-clause",
      "name": "BSD 3-clause \"New\" or \"Revised\" License",
      "spdx_id": "BSD-3-Clause",
      "url": "https://api.github.com/licenses/bsd-3-clause",
      "featured": false
    },
    "code_of_conduct": {
      "key": "none",
      "name": "None",
      "url": null
    }
  }
]
```

That means we can proceed with this PR.

jamiehannaford:
That's awesome, thanks @kytrinyx! 🎉

shurcooL:
@jamiehannaford Can you address the first item I raised in https://github.com/google/go-github/pull/656#issuecomment-313720527? Second item is now resolved.

jamiehannaford:
@shurcooL Yep, will do when I get some free time this week

shurcooL:
Also see https://github.com/google/go-github/issues/675#issuecomment-316771743, which ran into a very similar issue. I found there's a few more endpoints that return `[]*github.Repository` and they may also need this header set:

- ❓ https://godoc.org/github.com/google/go-github/github#RepositoriesService.ListForks
- ❓ https://godoc.org/github.com/google/go-github/github#OrganizationsService.ListTeamRepos

I can't confirm whether COC is supported by those extra endpoints from looking at the github API docs, so you'll need to confirm by testing.

jamiehannaford:
@shurcooL Okay, comments addressed. Sorry this took so long to get to!

shurcooL:
@jamiehannaford Thanks! In the future, please avoid squashing and force pushing to the PR branch. That eliminates the change history and makes it harder to review only what has changed. We can squash all the commits at the end, when merging, so doing it preemptively is unnecessary.

It looks like you've removed the header from /repositories endpoint (`ListAll` method), which indeed addresses the first item in https://github.com/google/go-github/pull/656#issuecomment-313720527. The second is already addressed on GitHub's side.

That just leaves my new question in https://github.com/google/go-github/pull/656#issuecomment-316773955, have you seen it? Do `RepositoriesService.ListForks` and `OrganizationsService.ListTeamRepos` methods need this header too?

shurcooL:
Looks like `ListForks` does not support it:

```
$ curl -s -H "Accept: application/vnd.github.scarlet-witch-preview+json" https://api.github.com/repos/golang/go/forks?per_page=1 | grep "code_of_conduct"
$
```

That just leaves us to test `ListTeamRepos`.

shurcooL:
Finding it pretty hard to test `ListTeamRepos` endpoint. It needs admin rights to a repo, and that repo needs to have teams, and those teams need to have access to some repos...

Since the sample output in GitHub API docs (https://developer.github.com/v3/orgs/teams/#list-team-repos) doesn't include code of conduct, I think we can assume that's the case until proven otherwise.

Unless you have a way to confirm, I think we can move on.

github/repos.go
```diff
@@ -177,7 +177,7 @@ func (s *RepositoriesService) List(ctx context.Context, user string, opt *Reposi
 	}
 
 	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeTopicsPreview, mediaTypeCodesOfConductPreview}
```
shurcooL: `RepositoriesService.Get` method lists these in the following order:

```Go
acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
```

Let's update these to use the same, consistent order. Here, and all other lines in this PR.

jamiehannaford: Done :)

shurcooL:
Thanks @jamiehannaford!

PR #657, opened 2017-06-16
avoid stale responses

BOTBrad:
currently if you push/delete tags from a repo and then query the list of refs, github returns stale data. setting the cache-control header to "must-revalidate" alleviates this. [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) is a list of other values we could set if we'd prefer no caching or something.


BOTBrad:
lol wrong fork

PR #660, opened 2017-06-27
Guard against nil *http.Response

### LINKED ISSUE #659: net/http error can result in a nil *http.Response which causes panic
I was investigating this crash:

```
furan_1      | panic: runtime error: invalid memory address or nil pointer dereference
furan_1      | [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x7bb77a]
furan_1      |
furan_1      | goroutine 911 [running]:
furan_1      | github.com/dollarshaveclub/furan/vendor/github.com/google/go-github/github.(*Response).populatePageValues(0xc42040d180)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/vendor/github.com/google/go-github/github/github.go:294 +0x3a
furan_1      | github.com/dollarshaveclub/furan/vendor/github.com/google/go-github/github.newResponse(0x0, 0xc420435c00)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/vendor/github.com/google/go-github/github/github.go:286 +0x5a
furan_1      | github.com/dollarshaveclub/furan/vendor/github.com/google/go-github/github.(*RepositoriesService).GetArchiveLink(0xc420335b90, 0xc420363c00, 0xf, 0xc420363c10, 0x4, 0xc77aef, 0x7, 0xc42004bb68, 0x1d40e0e9f774ab6f, 0xa03bb7934bbb0dbe, ...)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/vendor/github.com/google/go-github/github/repos_contents.go:271 +0x273
furan_1      | github.com/dollarshaveclub/furan/lib.(*GitHubFetcher).Get(0xc42000ec10, 0xc420363c00, 0xf, 0xc420363c10, 0x4, 0xc42020768a, 0x6, 0x1, 0x106bb30, 0xc4204b4000, ...)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/lib/github_fetch.go:58 +0xc9
furan_1      | github.com/dollarshaveclub/furan/lib.(*ImageBuilder).Build(0xc4204a71e0, 0x7f03ad986eb0, 0xc42032ae00, 0xc420207550, 0x1d40e0e9f774ab6f, 0xa03bb7934bbb0dbe, 0x0, 0x1d40e0e9f774ab6f, 0xa03bb7934bbb0dbe, 0x2)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/lib/builder.go:229 +0x420
furan_1      | github.com/dollarshaveclub/furan/lib.(*GrpcServer).syncBuild(0xc4200c9bc0, 0x7f03ad986eb0, 0xc42032ae00, 0xc420207550, 0xc400000000)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/lib/grpc.go:313 +0x3f4
furan_1      | github.com/dollarshaveclub/furan/lib.(*GrpcServer).buildWorker(0xc4200c9bc0, 0x7f03ad986eb0, 0xc4202c3600)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/lib/grpc.go:166 +0x2bd
furan_1      | github.com/dollarshaveclub/furan/lib.(*GrpcServer).runWorkers.func1(0xc4200c9bc0, 0x7f03ad986eb0, 0xc4202c3600)
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/lib/grpc.go:153 +0x6c
furan_1      | created by github.com/dollarshaveclub/furan/lib.(*GrpcServer).runWorkers
furan_1      | 	/go/src/github.com/dollarshaveclub/furan/lib/grpc.go:153 +0x14d
```

Which is ultimately caused by an erroneous newline in the oauth2 token, which causes `net/http` to throw this error: `net/http: invalid header field value "Bearer xxxxxxxxxxxxxxxxxxxx\n" for key Authorization`

TLDR: ``newResponse()`` should check for a nil *http.Response before calling ``response.populatePageValues()``

bkeroackdsc:
Fixes #659

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -316,6 +316,9 @@ type Response struct {
 
 // newResponse creates a new Response for the provided http.Response.
 func newResponse(r *http.Response) *Response {
+	if r == nil {
+		return nil
+	}
 	response := &Response{Response: r}
 	response.populatePageValues()
 	response.Rate = parseRate(r)


shurcooL:
This change should not be merged as is. See my comment https://github.com/google/go-github/issues/659#issuecomment-311488297.

I don't think the issue exists on `master`.

However, we can make an improvement to `newResponse` documentation as follows:

```diff
 // newResponse creates a new Response for the provided http.Response.
+// r must not be nil.
 func newResponse(r *http.Response) *Response {
```

Then it's just a matter of making sure we never call `newResponse(nil)`. Which, as far as I can tell, we do not in latest `master`.

gmlewis:
SGTM.  Thank you, @shurcooL!

bkeroackdsc:
It seems to me that the nil check is a better solution than a code comment, as it protects against future mistakes.

shurcooL:
The nil check and code comment serve different purposes. The nil check changes the behavior of the `newResponse` function. The code comments simply helps better describe the current behavior of the `newResponse` function.

I don't think we want to change the behavior of `newResponse`. Currently, it expects a non-nil `*http.Response`, and always returns a non-nil `*github.Response`. That's how it's designed, and that's how it's being used. If we change its behavior to return `nil` values sometimes, then we need to update other code, otherwise it would panic (e.g., [this line](https://github.com/google/go-github/blob/7a51fb928f52a196d5f31daefb8a489453ef54ff/github/github.go#L438) would panic if `newResponse` returns `nil`). That means adding more `if resp == nil { /* handle nil case */ }` elsewhere. That's not helpful.

It's a private function for use within `go-github`. It's used only in 3 places in all of `go-github`, so it's very easy to ensure we don't make a mistake and give it a `nil` value.

Thanks for wanting to contribute. I made the suggestion of improving documentation of `newResponse`. If you want to update this PR to do that, we can accept it. If you don't want to do that, it's not a problem. Let us know how you want to proceed. Thanks.

PR #662, opened 2017-06-30
Add OrganizationService.GetByID method.

s111:
This pull request adds the `GetByID` method to `OrganizationsService` in the same fashion as is done for `UsersService`  in #333 and `RepositoriesService` in #332.

I've not created a separate issue as this method was added in the same spirit as described in #329.

### DIFF
--- a/github/orgs.go
+++ b/github/orgs.go
@@ -154,6 +154,25 @@ func (s *OrganizationsService) Get(ctx context.Context, org string) (*Organizati
 	return organization, resp, nil
 }
 
+// GetByID fetches an organization.
+//
+// Note: GetByID uses the undocumented GitHub API endpoint /organizations/:id.
+func (s *OrganizationsService) GetByID(ctx context.Context, id int) (*Organization, *Response, error) {
+	u := fmt.Sprintf("organizations/%d", id)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	organization := new(Organization)
+	resp, err := s.client.Do(ctx, req, organization)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return organization, resp, nil
+}
+
 // Edit an organization.
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/#edit-an-organization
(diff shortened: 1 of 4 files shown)


github/orgs_test.go
```diff
@@ -109,6 +109,26 @@ func TestOrganizationsService_Get_invalidOrg(t *testing.T) {
...
+	defer teardown()
+
+	mux.HandleFunc("/organizations/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, `{"id":1, "login":"l", "url":"u", "avatar_url": "a", "location":"l"}`)
+	})
+
+	org, _, err := client.Organizations.GetByID(context.Background(), 1)
+	if err != nil {
+		t.Errorf("Organizations.GetByID returned error: %v", err)
```
shurcooL: Change `t.Errorf` to `t.Fatalf` here. If we get non-nil error, it doesn't make sense to continue and try to look at result.

(If you don't mind, make a similar improvement to `TestUsersService_GetByID` and `TestRepositoriesService_GetByID`.)

github/orgs.go
```diff
@@ -154,6 +154,25 @@ func (s *OrganizationsService) Get(ctx context.Context, org string) (*Organizati
 	return organization, resp, nil
 }
 
+// GetByID fetches an organization.
+//
+// Note: GetByID uses the undocumented GitHub API endpoint /organizations/:id.
```
gmlewis: Maybe add "... and may break without notice." ?
Your call.

shurcooL: I'd suggest we don't do that for 2 reasons:

1. I have seen increasingly convincing evidence that this is a part of the GitHub REST API v3, albeit not explicitly/consistently documented, and it would not break because GitHub employees actively encourage its use and a lot of code relies on it.
2. The current version is consistent with documentation of `UsersService.GetByID` and `RepositoriesService.GetByID`.

gmlewis: OK, SGTM. Thanks, @shurcooL.

s111:
@shurcooL I've addressed your comment.

Issue #622, opened 2017-04-21
rate limit exceeded: funky error message


kevinburke:
The error message for "rate limit exceeded" has both a period and a semicolon in it which looks weird.

```
GET https://api.github.com/repos/golang/go/issues?direction=desc&page=1&per_page=100&sort=updated&state=all: 403 API rate limit of 5000 
still exceeded until 2017-04-21 08:38:57 -0700 PDT, not making remote request.; rate reset 
in 26m8.032328588s
```

We could also probably round the duration to the nearest second (if greater than a minute) or tenth of a second (if less than a minute) to avoid 8 useless units of precision after the decimal.

gmlewis:
Thanks, @kevinburke... excellent points.

shurcooL:
For anyone who wants to work on this issue, please discuss proposed fix here before sending a PR.

dotbugfix:
Newbie here - hoping to kick off contributions with this one.

The error message originates from `github/github.go:479` - removing the extra period is trivial.

I'm not sure where the "rate reset in ..." is printed, though. There's a test at `github_test.go::TestDo_rateLimit_rateLimitError` but it simply checks the actual members in `RateLimitError`. Where is this extra part concatenated to the error message?

kevinburke:
Ah, that might have come from our error wrapper


--
Kevin Burke
925.271.7005 | kev.inburke.com

shurcooL:
> Newbie here - hoping to kick off contributions with this one.

Thanks for looking at this!

> The error message originates from `github/github.go:479` - removing the extra period is trivial.
>
> I'm not sure where the "rate reset in ..." is printed, though.

One part of the error message is indeed from [`github/github.go:479`](https://github.com/google/go-github/blob/2966f2579cd93bc62410f55ba6830b3925e7629d/github/github.go#L479). The other part comes from the `Error` method of that `RateLimitError` type, at [`github/github.go:530-532`](https://github.com/google/go-github/blob/2966f2579cd93bc62410f55ba6830b3925e7629d/github/github.go#L530-L532).

So, to resolve this, we should figure out what we want the final message to be, and then decide which part of it should go into `Message`, and what part should be in `RateLimitError.Error` method.

[...]

PR #663, opened 2017-06-30
Changed rate limit error message string

korovaisdead:
#622 
Old message: ... rate reset in 26m8.032328588s

New messages:
... [rate reset in 12h26m00s]
... [rate reset in 26m08s]
... [rate reset in 8s]

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -532,7 +532,7 @@ type RateLimitError struct {
 func (r *RateLimitError) Error() string {
 	return fmt.Sprintf("%v %v: %d %v %v",
 		r.Response.Request.Method, sanitizeURL(r.Response.Request.URL),
-		r.Response.StatusCode, r.Message, formatRateLimitResetDuration(r.Rate.Reset.Time.Sub(time.Now())))
+		r.Response.StatusCode, r.Message, formatRateReset(r.Rate.Reset.Time.Sub(time.Now())))
 }
 
 // AcceptedError occurs when GitHub returns 202 Accepted response with an
@@ -883,9 +883,8 @@ func cloneRequest(r *http.Request) *http.Request {
 // formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
 // "[rate reset in 87m02s]" for the positive durations. And like "[rate limit was reset 87m02s ago]"
 // for the negative cases.
-func formatRateLimitResetDuration(d time.Duration) string {
+func formatRateReset(d time.Duration) string {
 	isNegative := d < 0
-	var timeString string
 	if isNegative {
 		d *= -1
 	}
@@ -893,6 +892,7 @@ func formatRateLimitResetDuration(d time.Duration) string {
 	minutes := int((secondsTotal) / 60)
 	seconds := int(secondsTotal - minutes*60)
 
+	var timeString string
 	if minutes > 0 {
 		timeString = fmt.Sprintf("%dm%02ds", minutes, seconds)
 	} else {
--- a/github/github.go
+++ b/github/github.go
@@ -530,9 +530,9 @@ type RateLimitError struct {
 }
 
 func (r *RateLimitError) Error() string {
-	return fmt.Sprintf("%v %v: %d %v [rate reset in %v]",
+	return fmt.Sprintf("%v %v: %d %v %v",
 		r.Response.Request.Method, sanitizeURL(r.Response.Request.URL),
-		r.Response.StatusCode, r.Message, formatDuration(r.Rate.Reset.Time.Sub(time.Now())))
+		r.Response.StatusCode, r.Message, formatRateLimitResetDuration(r.Rate.Reset.Time.Sub(time.Now())))
 }
 
 // AcceptedError occurs when GitHub returns 202 Accepted response with an
@@ -880,21 +880,30 @@ func cloneRequest(r *http.Request) *http.Request {
 	return r2
 }
 
-func formatDuration(d time.Duration) string {
-	secondsTotal := int(d.Seconds())
-	hours := int(secondsTotal / 60 / 60)
-	minutes := int((secondsTotal - hours*60*60) / 60)
-	seconds := int(secondsTotal - hours*60*60 - minutes*60)
-
-	if hours > 0 {
-		return fmt.Sprintf("%dh%02dm%02ds", hours, minutes, seconds)
+// formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
+// "[rate reset in 87m02s]" for the positive durations. And like "[rate limit was reset 87m02s ago]"
+// for the negative cases.
+func formatRateLimitResetDuration(d time.Duration) string {
+	isNegative := d < 0
+	var timeString string
+	if isNegative {
+		d *= -1
 	}
+	secondsTotal := int(d.Seconds())
+	minutes := int((secondsTotal) / 60)
+	seconds := int(secondsTotal - minutes*60)
 
 	if minutes > 0 {
-		return fmt.Sprintf("%dm%02ds", minutes, seconds)
+		timeString = fmt.Sprintf("%dm%02ds", minutes, seconds)
+	} else {
+		timeString = fmt.Sprintf("%ds", seconds)
 	}
 
-	return fmt.Sprintf("%ds", seconds)
+	if isNegative {
+		return fmt.Sprintf("[rate limit was reset %v ago]", timeString)
+	}
+
+	return fmt.Sprintf("[rate reset in %v]", timeString)
 }
 
 // Bool is a helper routine that allocates a new bool value
--- a/github/github.go
+++ b/github/github.go
@@ -530,9 +530,9 @@ type RateLimitError struct {
 }
 
 func (r *RateLimitError) Error() string {
-	return fmt.Sprintf("%v %v: %d %v; rate reset in %v",
+	return fmt.Sprintf("%v %v: %d %v [rate reset in %v]",
 		r.Response.Request.Method, sanitizeURL(r.Response.Request.URL),
-		r.Response.StatusCode, r.Message, r.Rate.Reset.Time.Sub(time.Now()))
+		r.Response.StatusCode, r.Message, formatDuration(r.Rate.Reset.Time.Sub(time.Now())))
 }
 
 // AcceptedError occurs when GitHub returns 202 Accepted response with an
@@ -878,6 +878,23 @@ func cloneRequest(r *http.Request) *http.Request {
 		r2.Header[k] = append([]string(nil), s...)
 	}
 	return r2
+}
+
+func formatDuration(d time.Duration) string{
+	secondsTotal := int(d.Seconds())
+	hours := int(secondsTotal / 60 / 60)
+	minutes := int((secondsTotal - hours * 60 * 60) / 60)
+	seconds := int(secondsTotal - hours * 60 * 60 - minutes * 60)
+	
+	if hours > 0 {
+		return fmt.Sprintf("%dh%02dm%02ds", hours, minutes, seconds)
+	}
+
+	if minutes > 0 {
+		return fmt.Sprintf("%dm%02ds", minutes, seconds)
+	}
+
+	return fmt.Sprintf("%ds", seconds)
 }
 
 // Bool is a helper routine that allocates a new bool value
--- a/github/github.go
+++ b/github/github.go
@@ -880,7 +880,7 @@ func cloneRequest(r *http.Request) *http.Request {
 	return r2
 }
 
-// formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
+// formatRateReset formats d to look like "[rate reset in 2s]" or
 // "[rate reset in 87m02s]" for the positive durations. And like "[rate limit was reset 87m02s ago]"
 // for the negative cases.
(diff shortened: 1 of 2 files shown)


shurcooL:
I think this looks like a reasonable solution for #622. Thanks for working on this @korovaisdead.

I have some suggestions on how we can improve the code, I'll post a review.

github/github.go
```diff
@@ -880,6 +880,23 @@ func cloneRequest(r *http.Request) *http.Request {
...
 
+func formatDuration(d time.Duration) string {
+	secondsTotal := int(d.Seconds())
+	hours := int(secondsTotal / 60 / 60)
+	minutes := int((secondsTotal - hours*60*60) / 60)
+	seconds := int(secondsTotal - hours*60*60 - minutes*60)
+
+	if hours > 0 {
+		return fmt.Sprintf("%dh%02dm%02ds", hours, minutes, seconds)
+	}
```
shurcooL: As far as I know, GitHub rate limit resets at most 1 hour away from current time. So, I think we can get rid of this case for `hours > 0` because it'll never be true.

Even if it's true in some rare case, it's fine for us to print that rate limit is going to be reset in 120 minutes.

github/github.go
```diff
@@ -880,6 +880,23 @@ func cloneRequest(r *http.Request) *http.Request {
 	return r2
 }
 
+func formatDuration(d time.Duration) string {
```
shurcooL: formatDuration could use a comment describing what it's meant to do.

```Go
// formatDuration formats d to look like "26m08s" or "8s".
```

Without a comment saying what the function does, one can only infer it from the implementation, and that makes it hard/impossible to ever improve the implementation, or figure out if a certain behavior is a bug or intended.

shurcooL: Also, what will `formatDuration` return on negative durations? Does it handle them properly?

It's possible for a `RateLimitError` to be stored for some time, and its `Error` method can be called after `r.Rate.Reset` is in the past. Then, `r.Rate.Reset.Time.Sub(time.Now())` will be a negative duration, so `formatDuration` needs to be able to handle it.

Perhaps it's a good idea to add a few test cases for `formatDuration`, including a negative duration.

korovaisdead: Will take a look into it. Thank for sharing the knowledge about the negative values. I've seen them in the test. But thought it's some kind of obsolete. 
So I'll improve the code to support negative numbers + will write some tests. 

Is it possible to assign the issue to me?

shurcooL: Sounds good, thanks.

> Is it possible to assign the issue to me?

@gmlewis would have to invite you to org, otherwise we can't do it because of weird GitHub limitations (/cc @kytrinyx FYI, this isn't very friendly to open source projects):

image

I'll just post a comment in the issue to point to this PR, so people will be aware that you're working on it here.

gmlewis: Sending invite.

korovaisdead: I'm in. You can assign me now. Thanks!

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
 	return r2
 }
 
+// formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
+// "[rate reset in 87m02s]" for the positive durations. And like "[rate limit was reset 87m02s ago]"
+// for the negative cases.
+func formatRateLimitResetDuration(d time.Duration) string {
```
shurcooL: It's good that the func name is descriptive and specific to the task (since it's only being used in one place), but I think we can shorten it to just `formatRateReset`.

It happens to accept a `time.Duration` parameter, but that doesn't mean we need to put "Duration" in its name. Rate reset should be descriptive enough, since it's being passed `r.Rate.Reset` and it's what we call it in the string.

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
...
+	isNegative := d < 0
+	var timeString string
```
shurcooL: In general, it's a good practice not to do unnecessary things, which includes declaring variables way earlier than necessary.

In this case, `var timeString string` is declared here, but not used/needed until 7 lines below.

Let's move it so it's declared just before it's used, before `if minutes > 0` line.

That should improve readability of this function.

github/github_test.go
```diff
@@ -911,3 +911,41 @@ func TestBasicAuthTransport_transport(t *testing.T) {
...
+
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
+	}
+
+	d = 14*time.Minute + 2*time.Second
+	got = formatRateLimitResetDuration(d)
+	want = "[rate reset in 14m02s]"
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
```
shurcooL: Here too.

github/github_test.go
```diff
@@ -911,3 +911,41 @@ func TestBasicAuthTransport_transport(t *testing.T) {
...
 	}
 }
+
+func TestFormatRateLimitResetDuration(t *testing.T) {
+	d := 120*time.Minute + 12*time.Second
+	got := formatRateLimitResetDuration(d)
+	want := "[rate reset in 120m12s]"
+
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
```
shurcooL: To be consistent with the most prevalent style, the order should always be "got", then "want". That makes test output more readable. So make this:

```Go
t.Errorf("Format is wrong. got: %v, want: %v", got, want)
```

And in other lines as well.

github/github_test.go
```diff
@@ -911,3 +911,41 @@ func TestBasicAuthTransport_transport(t *testing.T) {
...
+	want = "[rate reset in 14m02s]"
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
+	}
+
+	d = 2*time.Minute + 2*time.Second
+	got = formatRateLimitResetDuration(d)
+	want = "[rate reset in 2m02s]"
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
```
shurcooL: Here too.

github/github_test.go
```diff
@@ -911,3 +911,41 @@ func TestBasicAuthTransport_transport(t *testing.T) {
...
+	}
+
+	d = 12 * time.Second
+	got = formatRateLimitResetDuration(d)
+	want = "[rate reset in 12s]"
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
```
shurcooL: Here too.

github/github_test.go
```diff
@@ -911,3 +911,41 @@ func TestBasicAuthTransport_transport(t *testing.T) {
...
+	}
+
+	d = -1 * (2*time.Hour + 2*time.Second)
+	got = formatRateLimitResetDuration(d)
+	want = "[rate limit was reset 120m02s ago]"
+	if got != want {
+		t.Errorf("Format is wrong. want: %v, got: %v", want, got)
```
shurcooL: Here too.

korovaisdead: Thank you so much for noticing that. 
Just came from the another language world where all the time I've seen "Expected: bla-bla, but got: bla-bla-bla".

Improved my code :)

github/github_test.go
```diff
@@ -911,3 +911,41 @@ func TestBasicAuthTransport_transport(t *testing.T) {
 		t.Errorf("Expected custom transport to be used.")
 	}
 }
+
+func TestFormatRateLimitResetDuration(t *testing.T) {
+	d := 120*time.Minute + 12*time.Second
+	got := formatRateReset(d)
+	want := "[rate reset in 120m12s]"
+
```
shurcooL: Nitpick, remove this blank line, it's not needed.

shurcooL:
@gmlewis, how does this look to you?

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
...
 
+// formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
+// "[rate reset in 87m02s]" for the positive durations. And like "[rate limit was reset 87m02s ago]"
+// for the negative cases.
+func formatRateReset(d time.Duration) string {
+	isNegative := d < 0
+	if isNegative {
+		d *= -1
+	}
+	secondsTotal := int(d.Seconds())
```
gmlewis: nit: rounding might be nice, here:

`secondsTotal := int(0.5 + d.Seconds())`

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
...
+	minutes := int((secondsTotal) / 60)
```
gmlewis: It seems like there is unnecessary casting here.  How about just:

`minutes := secondsTotal / 60`

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
...
+	seconds := int(secondsTotal - minutes*60)
```
gmlewis: Same here:

`seconds := secondsTotal - minutes*60`

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
 	return r2
 }
 
+// formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
+// "[rate reset in 87m02s]" for the positive durations. And like "[rate limit was reset 87m02s ago]"
+// for the negative cases.
+func formatRateReset(d time.Duration) string {
+	isNegative := d < 0
```
gmlewis: One thought hit me that it might make this simpler and easier to read if `isNegative` was checked only once (and then didn't actually need to be created at all...  something like:

```go
fmtStr := "[rate reset in %v]"
if d < 0 {
	d = -d
	fmtStr = "[rate limit was reset %v ago]"
}
secondsTotal := int(0.5 + d.Seconds())
minutes := secondsTotal / 60
seconds := secondsTotal - minutes*60

var timeString string
if minutes > 0 {
	timeString = fmt.Sprintf("%dm%02ds", minutes, seconds)
} else {
	timeString = fmt.Sprintf("%ds", seconds)
}

return fmt.Sprintf(fmtStr, timeString)
```

shurcooL: > Please let me know if you disagree.

In my personal opinion, I find code that uses variables for format string (to printf calls) harder to read than code that uses separate paths and constant format strings. I find having a `isNegative` boolean which does what it says more clear.

When looking at `return fmt.Sprintf(fmtStr, timeString)`, it's hard to immediately see what it'll print (without having to execute an entire block of code in your head). The following is much more clear to me:

```Go
if isNegative {
	return fmt.Sprintf("[rate limit was reset %v ago]", timeString)
}
return fmt.Sprintf("[rate reset in %v]", timeString)
```

github/github.go
shurcooL: Yep, this is much better.

[`unconvert`](https://github.com/mdempsky/unconvert) tool by @mdempsky helps finds opportunities like this.

github/github.go
gmlewis: OK, SGTM.

korovaisdead: Totally agree with @shurcooL here. Much more easier to read.

github/github.go
```diff
@@ -888,9 +888,9 @@ func formatRateReset(d time.Duration) string {
 	if isNegative {
 		d *= -1
 	}
-	secondsTotal := int(d.Seconds())
-	minutes := int((secondsTotal) / 60)
-	seconds := int(secondsTotal - minutes*60)
+	secondsTotal := int(0.5 + d.Seconds())
+	minutes := (secondsTotal) / 60
```
gmlewis: Please remove the unnecessary parens.

korovaisdead: Already. You did the review much faster than I typed `git commit --amend` in terminal.

korovaisdead:
@shurcooL Any updates here ? :)

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
 	return r2
 }
 
+// formatRateLimitResetDuration formats d to look like "[rate reset in 2s]" or
```
shurcooL: Since you've renamed the function, this comment should also be updated to begin with `// formatRateReset formats d to ...`.

github/github.go
```diff
@@ -880,6 +880,32 @@ func cloneRequest(r *http.Request) *http.Request {
...
+	if minutes > 0 {
+		timeString = fmt.Sprintf("%dm%02ds", minutes, seconds)
+	} else {
+		timeString = fmt.Sprintf("%ds", seconds)
+	}
+
+	if isNegative {
+		return fmt.Sprintf("[rate limit was reset %v ago]", timeString)
+	}
+
```
shurcooL: I'd suggest removing this blank line, it does not aid readability. That way, both return statements are closer to each other.

github/github_test.go
```diff
@@ -911,3 +911,40 @@ func TestBasicAuthTransport_transport(t *testing.T) {
 		t.Errorf("Expected custom transport to be used.")
 	}
 }
+
+func TestFormatRateLimitResetDuration(t *testing.T) {
```
shurcooL: Also rename this to `TestFormatRateReset` to match the new function name.

korovaisdead:
Done @shurcooL

PR #665, opened 2017-07-03
Document that newResponse expects non-nil input.

shurcooL:
The current behavior of `newResponse` is that it expects non-nil `r`, and always returns non-nil `*github.Response`. This behavior is relied on by all callers of `newResponse`.

This change improves the documentation of `newResponse` to reflect that.

Closes #660.


gmlewis:
LGTM.  Merging.

PR #668, opened 2017-07-08
Add missing fields to ProjectCard struct.

### LINKED ISSUE #667: ProjectCard missing attributes
First of all great job with this lib 👍 

Next, I've noticed that the [ProjectCard](https://github.com/google/go-github/blob/master/github/projects.go#L263) is missing some fields, checking the github [documentation](https://developer.github.com/v3/projects/cards/#response) I detect that: `url` and `creator` are missing. But when it comes from a webhook of [project_card](https://developer.github.com/v3/projects/cards/#response) it has also `column_id` and it's missing also.

The `creator` filed matches the struct for [`User`](https://github.com/google/go-github/blob/master/github/users.go#L20)

XescuGC:
Add fields that were missing in the ProjectCard struct.

Fixes #667.

### DIFF
--- a/github/projects.go
+++ b/github/projects.go
@@ -269,7 +269,9 @@ type ProjectCard struct {
 	Creator    *User      `json:"creator,omitempty"`
 	CreatedAt  *Timestamp `json:"created_at,omitempty"`
 	UpdatedAt  *Timestamp `json:"updated_at,omitempty"`
-	ColumnID   *int       `json:"column_id,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	ColumnID *int `json:"column_id,omitempty"`
 }
 
 // ListProjectCards lists the cards in a column of a GitHub Project.
--- a/github/projects.go
+++ b/github/projects.go
@@ -267,6 +267,9 @@ type ProjectCard struct {
 	Note       *string    `json:"note,omitempty"`
 	CreatedAt  *Timestamp `json:"created_at,omitempty"`
 	UpdatedAt  *Timestamp `json:"updated_at,omitempty"`
+	Url        *string    `json:"url,omitempty"`
+	Creator    *User      `json:"creator,omitempty"`
+	ColumnID   *int       `json:"column_id,omitempty"`
 }
 
 // ListProjectCards lists the cards in a column of a GitHub Project.
--- a/github/projects.go
+++ b/github/projects.go
@@ -259,7 +259,7 @@ func (s *ProjectsService) MoveProjectColumn(ctx context.Context, columnID int, o
 
 // ProjectCard represents a card in a column of a GitHub Project.
 //
-// GitHub API docs: https://developer.github.com/v3/repos/projects/
+// GitHub API docs: https://developer.github.com/v3/projects/cards/#get-a-project-card
 type ProjectCard struct {
 	URL        *string    `json:"url,omitempty"`
 	ColumnURL  *string    `json:"column_url,omitempty"`
--- a/github/projects.go
+++ b/github/projects.go
@@ -261,14 +261,14 @@ func (s *ProjectsService) MoveProjectColumn(ctx context.Context, columnID int, o
 //
 // GitHub API docs: https://developer.github.com/v3/repos/projects/
 type ProjectCard struct {
+	URL        *string    `json:"url,omitempty"`
 	ColumnURL  *string    `json:"column_url,omitempty"`
 	ContentURL *string    `json:"content_url,omitempty"`
 	ID         *int       `json:"id,omitempty"`
 	Note       *string    `json:"note,omitempty"`
+	Creator    *User      `json:"creator,omitempty"`
 	CreatedAt  *Timestamp `json:"created_at,omitempty"`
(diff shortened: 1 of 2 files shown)


github/projects.go
```diff
@@ -267,6 +267,9 @@ type ProjectCard struct {
 	Note       *string    `json:"note,omitempty"`
 	CreatedAt  *Timestamp `json:"created_at,omitempty"`
 	UpdatedAt  *Timestamp `json:"updated_at,omitempty"`
+	Url        *string    `json:"url,omitempty"`
```
shurcooL: This field should be `URL` to match Go style, see https://golang.org/s/style#initialisms.

shurcooL: Also, let's be consistent with https://developer.github.com/v3/projects/cards/#get-a-project-card example and place `URL` on top of `ColumnURL`, so it's next to other URL-related fields.

github/projects.go
```diff
@@ -267,6 +267,9 @@ type ProjectCard struct {
...
+	Creator    *User      `json:"creator,omitempty"`
```
shurcooL: Let's move `Creator` below `Note`. It's more important.

github/projects.go
```diff
@@ -267,6 +267,9 @@ type ProjectCard struct {
...
+	ColumnID   *int       `json:"column_id,omitempty"`
```
shurcooL: I asked in https://github.com/google/go-github/issues/667#issuecomment-313971934, but do you have a source for this field that you can share?

github/projects.go
XescuGC: Totally agree, it was my mistake on the `URL` part

XescuGC:
I've readed all the [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md#submitting-a-patch) before the PR and run the commands.

But now that I take a look at them carefully I comited a mistake, i've executed exactly the commands that are listed ther: `go generate github.com/google/go-github/...`, `go test github.com/google/go-github/...` and `go vet github.com/google/go-github/...`. But the mistake is that as it's a **fork** they should be runed with my GH name instead of `googel`. Now I've executed them correctly.

Could be a suggestion to change this docs to be `go generate ./...` and so on for the others? As basically they are mostrly run from insidhe the project.

github/projects.go
XescuGC: Answered https://github.com/google/go-github/issues/667#issuecomment-314026190 👍

github/projects.go
```diff
@@ -261,12 +261,15 @@ func (s *ProjectsService) MoveProjectColumn(ctx context.Context, columnID int, o
 //
 // GitHub API docs: https://developer.github.com/v3/repos/projects/
```
shurcooL: This GitHub API docs link is pretty generic and out of date (the URL has changed). Mind updating it to something more relevant, like https://developer.github.com/v3/projects/cards/#get-a-project-card?

github/projects.go
```diff
@@ -261,12 +261,15 @@ func (s *ProjectsService) MoveProjectColumn(ctx context.Context, columnID int, o
...
 type ProjectCard struct {
+	URL        *string    `json:"url,omitempty"`
 	ColumnURL  *string    `json:"column_url,omitempty"`
 	ContentURL *string    `json:"content_url,omitempty"`
 	ID         *int       `json:"id,omitempty"`
 	Note       *string    `json:"note,omitempty"`
+	Creator    *User      `json:"creator,omitempty"`
 	CreatedAt  *Timestamp `json:"created_at,omitempty"`
 	UpdatedAt  *Timestamp `json:"updated_at,omitempty"`
+	ColumnID   *int       `json:"column_id,omitempty"`
```
shurcooL: Since `ColumnID` will be a field populated in Webhook payloads only, let's be consistent with the pattern in `event_types.go` and separate it out from the rest of the fields with a comment:

```Go
	...
	Creator    *User      `json:"creator,omitempty"`
	CreatedAt  *Timestamp `json:"created_at,omitempty"`
	UpdatedAt  *Timestamp `json:"created_at,omitempty"`

	ColumnID *int `json:"column_id,omitempty"`
}
```

XescuGC: Sorry I did not see the `event_types.go` did it now 👍

github/projects.go
XescuGC: I did see the url mistake and I forgot to commit the new one hehe.

XescuGC:
Totally agree that the one with full path can be used from any directory but only if you are working on the main one, not in a fork (because you have to fork to contribute, except for the collaborators ofc), which may be the use case of the people reading the [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md#submitting-a-patch)
Another option would be to have a `Makefile` with all this so it's forced to be on the dir to run it, but as no `Makefile` is present right now in this project I thik it would be a good idea to add the `./...` in the documentation mentioning if you are in a fork 😄

gmlewis:
LGTM. Thank you, @XescuGC and @shurcooL!
Merging.

Issue #670, opened 2017-07-17
Support preview Repository API - read and replace topics for a repository.


shurcooL:
GitHub Announcement: https://developer.github.com/changes/2017-07-17-update-topics-on-repositories/

It looks like there are 4 changes/additions in this announcement:

- [ ] [Replace all the topics for a repository](https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository).
- [ ] [List the topics for a repository](https://developer.github.com/v3/repos#list-all-topics-for-a-repository) specifically.
- [ ] Topics are also available as part of the payload in the [get repository](https://developer.github.com/v3/repos/#get) endpoint.
- [ ] Topics are also available as part of the payload in the [search repositories](https://developer.github.com/v3/search/#search-repositories) endpoint.

~~The get repository payload may be problematic for the same reason as https://github.com/google/go-github/pull/656#issuecomment-313720527, so it's completely okay to address it later/in a separate PR.~~ This has been resolved (see comments below).

/cc @kytrinyx FYI.

kytrinyx:
I've asked some colleagues what happens if you need multiple different preview headers for a call.

kytrinyx:
Confirmed: at the moment making a request with multiple preview headers is not supported. You'd have to make the same request with multiple accept headers.

kytrinyx:
Actually, I take that back. It looks like we may have released a fix that should let you make a call with multiple preview headers:

```
curl -H "Accept: application/vnd.github.mockingbird-preview" -H "Accept: application/vnd.github.scarlet-witch-preview+json"
```

I just confirmed with this call:

curl -v -XGET "https://api.github.com/repos/:org/:repo" -H "Authorization: token $TOKEN" -H "Accept: application/vnd.github.mercy-preview+json" -H "Accept:  application/vnd.github.drax-preview+json"

The response came back with both `licenses` and `topics`.

shurcooL:
Awesome, thank you @kytrinyx! This is super helpful, because it would not be viable for us to make more than 1 remote API call per method, since each method is supposed to correspond to a single API call.

RaviTezu:
Hi @shurcooL 
Can I tackle this one for my first contribution? Thanks in advance.

bradleyfalzon:
Let me know how you go @RaviTezu, I'm interested in this functionality. I have a work around, so no pressure.

shurcooL:
Feel free to go for it, @RaviTezu, and thanks!

[...]

PR #672, opened 2017-07-18
Add preview support to Topics API

RaviTezu:
Hi, 

This is my first PR to go-github and please feel free to let me know, if there is anything that needs to be changed or can be improved. Thank you!

Fixes https://github.com/google/go-github/issues/670

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -55,7 +55,7 @@ type Repository struct {
 	AllowRebaseMerge *bool            `json:"allow_rebase_merge,omitempty"`
 	AllowSquashMerge *bool            `json:"allow_squash_merge,omitempty"`
 	AllowMergeCommit *bool            `json:"allow_merge_commit,omitempty"`
-	Topics           *[]string        `json:"topics,omitempty"`
+	Topics           []string         `json:"topics,omitempty"`
 
 	// Only provided when using RepositoriesService.Get while in preview
 	License *License `json:"license,omitempty"`
@@ -974,9 +974,9 @@ func (s *RepositoriesService) RemoveAdminEnforcement(ctx context.Context, owner,
 	return s.client.Do(ctx, req, nil)
 }
 
-// Topics represents a repository topics to list or replace.
+// Topics represents a collection of repository topics.
 type Topics struct {
-	Names *[]string `json:"names,omitempty"`
+	Names []string `json:"names,omitempty"`
 }
 
 // ListAllTopics lists topics for a repository.
--- a/github/repos.go
+++ b/github/repos.go
@@ -55,6 +55,7 @@ type Repository struct {
 	AllowRebaseMerge *bool            `json:"allow_rebase_merge,omitempty"`
 	AllowSquashMerge *bool            `json:"allow_squash_merge,omitempty"`
 	AllowMergeCommit *bool            `json:"allow_merge_commit,omitempty"`
+	Topics           *[]string        `json:"topics,omitempty"`
 
 	// Only provided when using RepositoriesService.Get while in preview
 	License *License `json:"license,omitempty"`
@@ -295,7 +296,7 @@ func (s *RepositoriesService) Get(ctx context.Context, owner, repo string) (*Rep
 
 	// TODO: remove custom Accept header when the license support fully launches
 	// https://developer.github.com/v3/licenses/#get-a-repositorys-license
-	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeSquashPreview, mediaTypeCodesOfConductPreview}
+	acceptHeaders := []string{mediaTypeLicensesPreview, mediaTypeSquashPreview, mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
 	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	repository := new(Repository)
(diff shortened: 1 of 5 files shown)


github/repos.go
```diff
@@ -55,6 +55,7 @@ type Repository struct {
 	AllowRebaseMerge *bool            `json:"allow_rebase_merge,omitempty"`
 	AllowSquashMerge *bool            `json:"allow_squash_merge,omitempty"`
 	AllowMergeCommit *bool            `json:"allow_merge_commit,omitempty"`
+	Topics           *[]string        `json:"topics,omitempty"`
```
gmlewis: There's no need to make it a pointer to a slice, as the slice is a reference type already, so please make this type a `[]string` instead of a `*[]string`.

github/repos.go
```diff
@@ -972,3 +973,52 @@ func (s *RepositoriesService) RemoveAdminEnforcement(ctx context.Context, owner,
 
 	return s.client.Do(ctx, req, nil)
 }
+
+// Topics represents a repository topics to list or replace.
```
gmlewis: Maybe `represents a collection of repository topics` ? I think that reads a bit better.

github/repos.go
```diff
@@ -972,3 +973,52 @@ func (s *RepositoriesService) RemoveAdminEnforcement(ctx context.Context, owner,
...
+type Topics struct {
+	Names *[]string `json:"names,omitempty"`
```
gmlewis: Same here: `[]string`.

shurcooL:
@RaviTezu, you should edit your PR description (or commit message) to include `Fixes #670.` instead of just `Issue Link: #670`, since you intend for this PR to resolve the entire issue. That way, when the PR is merged, that issue will be closed (and linked to this PR). This is [special GitHub syntax](https://github.com/blog/1506-closing-issues-via-pull-requests).

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
...
+		fmt.Fprint(w, `{"names":["go", "go-github", "github"]}`)
+	})
+
+	want := &Topics{Names: &[]string{"go", "go-github", "github"}}
+	topics, _, err := client.Repositories.ListAllTopics(context.Background(), "o", "r")
+	if err != nil {
+		t.Errorf("Repositories.ReplaceAllTopics returned error: %v", err)
+	}
+
+	if !reflect.DeepEqual(topics, want) {
```
shurcooL: Move `want` just above this line. It's pretty common to get the value of `got`, then define `want` and check if they're not equal. See tests above.

```Go
want := ...
if !reflect.DeepEqual(topics, want) {
```

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
...
+		fmt.Fprint(w, `{"names":["go", "go-github", "github"]}`)
+	})
+
+	want := &Topics{Names: &[]string{"go", "go-github", "github"}}
+	topics, _, err := client.Repositories.ReplaceAllTopics(context.Background(), "o", "r", want)
+	if err != nil {
+		t.Errorf("Repositories.ReplaceAllTopics returned error: %v", err)
+	}
+
+	if !reflect.DeepEqual(topics, want) {
```
shurcooL: Also here.

RaviTezu:
@gmlewis Made the changes and also, for the https://developer.github.com/v3/search/#search-repositories API, I have added the custom header https://github.com/google/go-github/pull/672/files#diff-6597fafa50e71487cf22a6edaa68b48a Please feel free to let me know, if this is wrong. 

@shurcooL updated the PR body to have the github special syntax. Thanks.

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
...
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/topics", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		fmt.Fprint(w, `{"names":["go", "go-github", "github"]}`)
+	})
+
+	got, _, err := client.Repositories.ListAllTopics(context.Background(), "o", "r")
+	if err != nil {
+		t.Errorf("Repositories.ReplaceAllTopics returned error: %v", err)
```
shurcooL: I think you meant "Repositories. ListAllTopics" in the error message, not "Repositories.ReplaceAllTopics".

Also, use `t.Fatalf` instead of `t.Errorf` here. If `err != nil`, that's a fatal error, we don't want to continue the test, since comparing the result of `got` is meaningless.

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
...
+	}
+
+	want := &Topics{Names: []string{"go", "go-github", "github"}}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Repositories.ReplaceAllTopics returned %+v, want %+v", got, want)
```
shurcooL: Also here, "Repositories.ReplaceAllTopics" should be "Repositories.ListAllTopics".

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
...
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/topics", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		fmt.Fprint(w, `{"names":["go", "go-github", "github"]}`)
+	})
+
+	got, _, err := client.Repositories.ReplaceAllTopics(context.Background(), "o", "r", &Topics{Names: []string{"go", "go-github", "github"}})
+	if err != nil {
+		t.Errorf("Repositories.ReplaceAllTopics returned error: %v", err)
```
shurcooL: Use `t.Fatalf` here as well, for same reason as above.

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
 		t.Errorf("PullRequestReviewsEnforcementRequest.MarshalJSON returned %+v, want %+v", string(json), want)
 	}
 }
+
+func TestRepositoriesService_ListAllTopics(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/topics", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
```
shurcooL: You could add `testHeader(t, r, "Accept", mediaTypeTopicsPreview)` here, similar to other tests.

github/repos_test.go
```diff
@@ -926,3 +926,43 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
...
+		t.Errorf("Repositories.ReplaceAllTopics returned %+v, want %+v", got, want)
+	}
+}
+
+func TestRepositoriesService_ReplaceAllTopics(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/topics", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
```
shurcooL: Also here.

RaviTezu:
@shurcooL Thank you for reviewing the commit. Made the changes.

github/search.go
```diff
@@ -192,6 +192,10 @@ func (s *SearchService) search(ctx context.Context, searchType string, query str
 		// Accept header defaults to "application/vnd.github.v3+json"
 		// We change it here to fetch back text-match metadata
 		req.Header.Set("Accept", "application/vnd.github.v3.text-match+json")
+	case searchType == "repositories":
+		// Accept header for search repositories based on topics preview endpoint
+		// TODO: remove custom Accept header when this API fully launches.
+		req.Header.Set("Accept", mediaTypeTopicsPreview)
```
shurcooL: Minor, but I think it'd be more consistent/logical to move this new case under the `searchType == "commits"` case. It's more similar to that one.

RaviTezu:
@shurcooL Moved the repositories case under commits case. Thanks.

github/repos_test.go
gmlewis: This one LGTM... no need to change this one.

shurcooL:
@gmlewis The reason the generated methods disappeared is because they're only generated for pointers, but those fields were changed to slices. CI is passing.

github/repos_test.go
shurcooL: This is a github bug... 😞 I left this comment in the right place, and after a new commit was pushed which resolved the issue, github incorrectly associated my comment with this wrong place instead of marking it as outdated.

shurcooL: You can see the comment was in the right place on the [original commit where I left it](https://github.com/google/go-github/pull/672/commits/43efbcca65900900ec7cda9ccf16dcbf8d47364e).

RaviTezu:
@gmlewis Yes, I have followed the step-5. Tests passed and vet didn't give any news, which is good I assume. As @shurcooL said, those methods were add when I used the pointer to slice and when it was replaced with just the slice,`generate` removed those added methods.

RaviTezu:
Thank you @gmlewis @shurcooL 
I would like to take up more. I will leave a comment on the issues.  Thanks again.

PR #673, opened 2017-07-18
Rename examples, tests directories to use singular form.

### LINKED ISSUE #671: examples, tests directories should use singular form.
This is a very minor/nitpicky issue that I'd like to fix, because it's bugging me, and because it's setting a bad example for new gophers.

The directories `examples`, `tests` in the repository root are containers of multiple packages. They should use singular form to follow idiomatic Go patterns. That way, it will be consistent with other Go projects, and `./example/basicauth`, `./test/integration`, `./test/fields` import paths read better.

See https://dmitri.shuralyov.com/idiomatic-go#use-singular-form-for-collection-repo-folder-name for rationale/proof. Also note that https://github.com/golang/go/tree/master/test directory is `test`, not `tests`, even though it contains many tests.

If this has approval from @willnorris and @gmlewis, I'll gladly work on a fix and make sure everything works after the rename.

shurcooL:
This is for consistency with prior art.

Also rename `./test/integration` package from `tests` to `integration`. It makes no difference what the package is called, since it's only used as a container for tests. But it's more common to name packages after the directory they're in, so do that.

Resolves #671.

I've noticed that the `integration` package doesn't build, because it hasn't been updated for the change to add `context.Context` support. To make things easier to review, I'll submit that as a separate followup PR.


gmlewis:
LGTM. Thank you, @shurcooL!
Please merge at your convenience.

PR #674, opened 2017-07-18
test/integration: Update for API changes to fix build.

shurcooL:
Update code with the latest breaking API changes so that it builds without errors.

All tests other than `TestRepositories_EditBranches` are passing. To fix that test, it requires someone with expertise in protection requests, and is out of scope of this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 language: go
 go:
-  - 1.7.x
   - 1.8.x
+  - 1.7.x
   - master
 matrix:
   allow_failures:
@@ -16,3 +16,4 @@ script:
   - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; (exit $code) # Check that go generate ./... produces a zero diff; clean up any changes afterwards.
   - go tool vet .
   - go test -v -race ./...
+  - go test -v -tags=integration -run=^$ ./test/integration # Check that integration test builds successfully, but don't run any of the tests (they hit live GitHub API).
--- a/test/integration/activity_test.go
+++ b/test/integration/activity_test.go
@@ -8,6 +8,7 @@
 package integration
 
 import (
+	"context"
 	"testing"
 
 	"github.com/google/go-github/github"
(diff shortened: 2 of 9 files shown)


shurcooL:
Note, CI is meaningless for this change. It doesn't do anything about `./test/integration` because it's an empty package unless `integration` build tag is manually specified.

Hmm, it might be possible to make it so that the integration test is verified to build successfully, but not run the tests. Let me do that.

shurcooL:
> Hmm, it might be possible to make it so that the integration test is verified to build successfully, but not run the tests. Let me do that.

I originally tried to just test that the package builds:

```yml
- go install -v -tags=integration ./test/integration # Test that it builds successfully, but don't run the tests.
```

However, that doesn't verify that the _test.go files contain valid Go code that builds successfully.

I think we'll need to use `go test`, but set `-run` flag such that no tests are matched.

.travis.yml
```diff
@@ -16,3 +16,4 @@ script:
   - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; (exit $code) # Check that go generate ./... produces a zero diff; clean up any changes afterwards.
   - go tool vet .
   - go test -v -race ./...
+  - go test -v -tags=integration -run=^$ ./test/integration # Check that integration test builds successfully, but don't run any of the tests (they hit live GitHub API).
```
gmlewis: Curious... would `go build` be useful here?

shurcooL: Have you seen my comment above at https://github.com/google/go-github/pull/674#issuecomment-316196325? Does it answer your question, or should I elaborate?

Note that `go test` is a superset of `go build`.

gmlewis: I saw that, but guess I didn't fully understand it.  Also, I didn't know why `go install` was used instead of `go build`. Oh, OK.  I think I get it now. No need to elaborate, thanks.

shurcooL: All right.

> Also, I didn't know why `go install` was used instead of `go build`.

For posterity, `go install` and `go build` effectively do the same thing, but the latter tends to throw away its results (unless a single command is provided). In this context, it doesn't matter, but in general, I always prefer to use `go install` since it saves the results of building packages to disk.

PR #677, opened 2017-07-27
Explicitly drop unused variables

alrs:
I found a couple of places where variables were being set but silently swallowed. This explicitly assigns to '_'.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -248,7 +248,7 @@ func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo st
 	}
 	var resp *http.Response
 	// Use http.DefaultTransport if no custom Transport is configured
-	ctx, req = withContext(ctx, req)
+	_, req = withContext(ctx, req)
 	if s.client.client.Transport == nil {
 		resp, err = http.DefaultTransport.RoundTrip(req)
 	} else {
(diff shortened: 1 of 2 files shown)


alrs:
I've signed the CLA.

shurcooL:
Thanks for working on this @alrs. However, this is not a change we should apply. I'll close the PR so it's not accidentally merged, but we can discuss it further and re-open if we decide to approach it differently.

It's pretty subtle, but let me explain what's going on.

We should definitely comment the `withContext` helper, because using it is prone to this type of situation.

Basically, `withContext` has two different implementations depending on whether you're on Google App Engine or normal Go:

```Go
// +build appengine

func withContext(ctx context.Context, req *http.Request) (context.Context, *http.Request) {
	return appengine.WithContext(ctx, req), req
}
```

```Go
// +build !appengine

func withContext(ctx context.Context, req *http.Request) (context.Context, *http.Request) {
	return ctx, req.WithContext(ctx)
}
```

So depending on what environment go-github is used in, it may either update the value of ctx to attach the HTTP request to it, or update req to attach context to it.

As a result, the correct usage of `withContext` is always exactly this:

```Go
ctx, req = withContext(ctx, req)
```

Even if one of `ctx` or `req` aren't used _right now_, they may be used later, and you wouldn't want it to have an inconsistent state:

```Go
_, req = withContext(ctx, req)
...
// added in a future change:
err = someHelper(ctx, ...) // bad, uses incorrect ctx without req attached!
```

gmlewis:
Ah, I see what you are saying, @shurcooL... the moment someone adds a line using ctx, everything would compile and look fine, but then be incorrect and cause crashes during runtime.

Yes, that makes sense.  Thank you for catching that!

PR #679, opened 2017-08-01
github: add IsPullRequest helper

kevinburke:
Instead of defining a pull request in terms of a property that may
or may not be set, define it with a helper method. This way if the
definition changes, the implementation of IsPullRequest can change
without breaking user apps.

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -21,8 +21,8 @@ type IssuesService service
 //
 // Note: As far as the GitHub API is concerned, every pull request is an issue,
 // but not every issue is a pull request. Some endpoints, events, and webhooks
-// may also return pull requests via this struct. If PullRequestLinks is nil,
-// this is an issue, and if PullRequestLinks is not nil, this is a pull request.
+// may also return pull requests via this struct. Call issue.IsPullRequest to
+// determine if the issue is a pull request.
 type Issue struct {
 	ID               *int              `json:"id,omitempty"`
 	Number           *int              `json:"number,omitempty"`
@@ -53,6 +53,13 @@ type Issue struct {
 
 func (i Issue) String() string {
 	return Stringify(i)
+}
+
+// IsPullRequest reports whether the issue is also a pull request. It uses the
+// method recommended by Github's API documentation, which is to check whether
+// PullRequestLinks is non-nil.
+func (i Issue) IsPullRequest() bool {
+	return i.PullRequestLinks != nil
 }
 
 // IssueRequest represents a request to create/edit an issue.
@@ -97,7 +104,7 @@ type IssueListOptions struct {
 }
 
 // PullRequestLinks object is added to the Issue object when it's an issue included
-// in the IssueCommentEvent webhook payload, if the webhooks is fired by a comment on a PR
+// in the IssueCommentEvent webhook payload, if the webhook is fired by a comment on a PR
 type PullRequestLinks struct {
 	URL      *string `json:"url,omitempty"`
 	HTMLURL  *string `json:"html_url,omitempty"`
--- a/github/issues_test.go
+++ b/github/issues_test.go
@@ -275,3 +275,14 @@ func TestIssuesService_Unlock(t *testing.T) {
 		t.Errorf("Issues.Unlock returned error: %v", err)
 	}
 }
+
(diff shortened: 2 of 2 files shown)


github/issues.go
```diff
@@ -55,6 +55,11 @@ func (i Issue) String() string {
 	return Stringify(i)
 }
 
+// IsPullRequest returns true if the issue is also a pull request.
```
shurcooL: s/returns true if/reports whether/

github/issues.go
```diff
@@ -21,8 +21,8 @@ type IssuesService service
 //
 // Note: As far as the GitHub API is concerned, every pull request is an issue,
 // but not every issue is a pull request. Some endpoints, events, and webhooks
-// may also return pull requests via this struct. If PullRequestLinks is nil,
-// this is an issue, and if PullRequestLinks is not nil, this is a pull request.
+// may also return pull requests via this struct. The issue is a pull request
+// if IsPullRequest returns true.
```
shurcooL: Let's not change this. Revert the doc change, but add a sentence that follows:

```
// The IsPullRequest helper makes it easier to check that.
```

Basically, I don't think we want to make existing users of `go-github` have to change their existing code. They don't. They should only use this helper if they wish, if it's helpful, and feel okay with not using it.

github/issues.go
shurcooL: In fact, we might want to document what `IsPullRequest` does, so that people don't get stuck in a situation where they're not sure which of the 2 approaches they should use to check if an issue is a PR (to check `PullRequestLinks` as GitHub docs tell them to, or to use `IsPullRequest` method which they don't know what it does).

By documenting how it operates (i.e., it also uses `PullRequestLinks` value), people can then feel more confident using it because they know it does the same thing.

kevinburke:
Done

github/issues_test.go
```diff
@@ -275,3 +275,14 @@ func TestIssuesService_Unlock(t *testing.T) {
 		t.Errorf("Issues.Unlock returned error: %v", err)
 	}
 }
+
+func TestIsPullRequest(t *testing.T) {
+	i := new(Issue)
+	if i.IsPullRequest() == true {
```
bradleyfalzon: Just curious about thoughts on removing `== true` (and similar `== false` below, but also negating it)? I often refer to tests as an example of how to use a function, and `if i.IsPullRequest() { ... }` if probably preferable than `if i.IsPullRequest() == true { ... }`?

I know both you and @shurcooL who already reviewed this knows you're doing, so curious why this wasn't flagged?

shurcooL: > I often refer to tests as an example of how to use a function

Tests usually do weird things to test functionality and not optimize for readability. Examples are for that.

The `== true` part to me is basically a `== dontWant`, with `true` inlined. We could also be more explicit and write the more verbose version thereof:

```Go
if got, want := i.IsPullRequest(), true; got != want {
```

Either way is fine with me.

I don't think we need to worry about `if i.IsPullRequest() == notWant {` being misinterpreted as the usage of this method.

github/issues.go
shurcooL: Did you see this comment @kevinburke?

You said "Done", but the "If PullRequestLinks is nil, this is an issue, and if PullRequestLinks is not nil, this is a pull request." sentence is still missing.

Do you feel strongly about removing it? I think it'd be better not to, but if you have reasons to, let me know.

shurcooL:
This looks good to me, except I want closure on the one remaining https://github.com/google/go-github/pull/679#discussion_r130783540. Let me know what you think there.

@gmlewis, what are your thoughts on this?

github/issues.go
gmlewis: I agree, @shurcooL, with [your comment](https://github.com/google/go-github/pull/679#discussion_r130531558).

shurcooL:
Friendly ping @kevinburke. See my comment above on what this PR needs to make progress.

ungureanuvladvictor:
Ping on this one, really want to use the helper rather than re-implementing it myself.

e-beach:
I might as well re-submit this PR under my name and make the 1-sentence change requested.

shurcooL:
Closing this in favor of #737.

PR #682, opened 2017-08-06
Add Preview Support for Team Review Requests

### LINKED ISSUE #678: Support preview Team Review Requests API
GitHub Announcement:
https://developer.github.com/changes/2017-07-26-team-review-request-thor-preview/

GitHub API Docs:
https://developer.github.com/v3/pulls/review_requests/

Note that the announcement describes a breaking change which may involve a bump of this package's version number.

RaviTezu:
Adding preview support for team review requests. 
Fixes #678

### DIFF
--- a/github/pulls_reviewers.go
+++ b/github/pulls_reviewers.go
@@ -13,18 +13,22 @@ import (
 // RequestReviewers creates a review request for the provided GitHub users for the specified pull request.
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#create-a-review-request
-func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo string, number int, logins []string) (*PullRequest, *Response, error) {
+func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo string, number int, logins []string, teams []string) (*PullRequest, *Response, error) {
 	u := fmt.Sprintf("repos/%s/%s/pulls/%d/requested_reviewers", owner, repo, number)
 
 	reviewers := struct {
-		Reviewers []string `json:"reviewers,omitempty"`
+		Reviewers     []string `json:"reviewers,omitempty"`
+		TeamReviewers []string `json:"team_reviewers,omitempty"`
 	}{
-		Reviewers: logins,
+		Reviewers:     logins,
+		TeamReviewers: teams,
 	}
 	req, err := s.client.NewRequest("POST", u, &reviewers)
 	if err != nil {
 		return nil, nil, err
 	}
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeTeamReviewPreview)
 
 	r := new(PullRequest)
 	resp, err := s.client.Do(ctx, req, r)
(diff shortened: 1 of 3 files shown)


RaviTezu:
As @gmlewis mentioned in the issue, this involves a breaking change and we may need to bump the package version number. 
@shurcooL

RaviTezu:
@gmlewis @shurcooL 
I see github.go has a const `libraryVersion = "8"`. Shall I change this to `libraryVersion = "9"` ?

shurcooL:
> I see github.go has a const `libraryVersion = "8"`. Shall I change this to `libraryVersion = "9"` ? 

Yes, please do. That'll help us communicate that there are breaking API changes being done (as a result of GitHub preview API changing).

github/pulls_reviewers.go
```diff
@@ -13,18 +13,22 @@ import (
 // RequestReviewers creates a review request for the provided GitHub users for the specified pull request.
```
shurcooL: > for the provided GitHub users

This was true before the change, but we'll need to update docs now to describe the new API.

github/pulls_reviewers.go
```diff
@@ -13,18 +13,22 @@ import (
...
 //
 // GitHub API docs: https://developer.github.com/v3/pulls/review_requests/#create-a-review-request
-func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo string, number int, logins []string) (*PullRequest, *Response, error) {
+func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo string, number int, logins []string, teams []string) (*PullRequest, *Response, error) {
```
shurcooL: @gmlewis, what do you think about this API decision?

An alternative to consider (that I can think of) is to create a `ReviewersRequest` struct which contains the two string slices, and have the user pass that in.

I'm pretty torn between the two, it seems they tradeoffs are pretty much equivalent and I'm not strongly in favor of either of them.

shurcooL:
@RaviTezu Thanks for your work on this PR! It's looking very solid so far. I've only briefly looked at it. First, I want to think about the API and decide which direction we'll want to go. Afterwards, I can review the implementation details more closely.

gmlewis:
LGTM
This seems like a good way to go (to me) for the API.
If @shurcooL comes up with a better alternative, I'm happy to discuss the tradeoffs.

RaviTezu:
@shurcooL Any thoughts on how to move forward with this API?

gmlewis:
@shurcooL's [suggestions](https://github.com/google/go-github/pull/682#pullrequestreview-54752451) SGTM.

RaviTezu:
@shurcooL I agree with you.
It increases the readability and the recent commit has the suggested API changes. 
Take a look. @gmlewis

github/pulls_reviewers.go
```diff
@@ -10,22 +10,31 @@ import (
...
+type Reviewers struct {
+	Users []*User `json:"users,omitempty"`
+	Teams []*Team `json:"teams,omitempty"`
+}
+
+// ReviewersRequest specifies users and teams for a pull request review request.
+type ReviewersRequest struct {
+	Reviewers     []string `json:"reviewers,omitempty"`
+	TeamReviewers []string `json:"team_reviewers,omitempty"`
+}
```
shurcooL: Minor, swap the order of these two structs, so `ReviewersRequest`, the input, is on top, and `Reviewers`, the output, is on bottom.

RaviTezu: @shurcooL Done.

PR #688, opened 2017-08-10
Add cryptographer-preview header to include GPG signed response in Re…

### LINKED ISSUE #687: RepositoriesService.ListCommits could also benefit from Preview support for Git signing.
While investigating #685, I noticed that `RepositoriesService.ListCommits` is also missing this header. Setting it would populate the `Verification` fields:

```
$ curl -s "https://api.github.com/repos/shurcooL/home/commits?per_page=1" | grep verification
$ curl -s "https://api.github.com/repos/shurcooL/home/commits?per_page=1" -H "Accept: application/vnd.github.cryptographer-preview" | grep verification
      "verification": {
```

/cc @RaviTezu FYI.

RaviTezu:
…positoryService's ListCommits method.

Fixes #687

### DIFF
--- a/github/repos_commits.go
+++ b/github/repos_commits.go
@@ -127,6 +127,9 @@ func (s *RepositoriesService) ListCommits(ctx context.Context, owner, repo strin
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeGitSigningPreview)
+
 	var commits []*RepositoryCommit
 	resp, err := s.client.Do(ctx, req, &commits)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


RaviTezu:
Yes! The builds aren't started yet. 
Probably something wrong at their end.

RaviTezu:
@shurcooL I am sorry, I don't get it. 
I think I am missing something, I don't see how https://godoc.org/github.com/google/go-github/github#Repository.License field is impacted with this change.

shurcooL:
It's not the field, but rather its documentation. It currently says:

```Go
// Only provided when using RepositoriesService.Get while in preview
License *License `json:"license,omitempty"`
```

That's no longer true.

RaviTezu:
@shurcooL ok. But, how is this change makes the License field's doc invalid?
I am sorry again, I am having hard time seeing the reason.

shurcooL:
Sorry @RaviTezu, I was mistaken. For some reason, I thought this PR and PR #686 were returning `Repository` and that `License` would now be populated after setting this preview header.

Now I see these two methods return `RepositoryCommits` and this change affects [`Commit.Verification`](https://godoc.org/github.com/google/go-github/github#Commit.Verification) field. Its documentation is fine and doesn't need to be updated.

Sorry about that! I can see how what I said made no sense at all. License isn't even related to this. :)

RaviTezu:
No problem :) 
@shurcooL Thanks.

PR #692, opened 2017-08-11
Add error for no trailing slash

178inaba:
This pull request provides logic to return an error if there is no slash at the end of `BaseURL` or `UploadURL`.

Related comment:
https://github.com/google/go-github/pull/690#issuecomment-321879628

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -242,7 +242,7 @@ func NewClient(httpClient *http.Client) *Client {
 // request body.
 func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Request, error) {
 	if c.BaseURL.Path != "" && !strings.HasSuffix(c.BaseURL.Path, "/") {
-		return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL.String())
+		return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL)
 	}
 	u, err := c.BaseURL.Parse(urlStr)
 	if err != nil {
@@ -278,7 +278,7 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
 // Relative URLs should always be specified without a preceding slash.
 func (c *Client) NewUploadRequest(urlStr string, reader io.Reader, size int64, mediaType string) (*http.Request, error) {
 	if c.UploadURL.Path != "" && !strings.HasSuffix(c.UploadURL.Path, "/") {
-		return nil, fmt.Errorf("UploadURL must have a trailing slash, but %q does not", c.UploadURL.String())
+		return nil, fmt.Errorf("UploadURL must have a trailing slash, but %q does not", c.UploadURL)
 	}
 	u, err := c.UploadURL.Parse(urlStr)
 	if err != nil {
--- a/github/github.go
+++ b/github/github.go
@@ -241,7 +241,7 @@ func NewClient(httpClient *http.Client) *Client {
 // specified, the value pointed to by body is JSON encoded and included as the
 // request body.
 func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Request, error) {
-	if c.BaseURL.Path != "" && !strings.HasSuffix(c.BaseURL.Path, "/") {
+	if !strings.HasSuffix(c.BaseURL.Path, "/") {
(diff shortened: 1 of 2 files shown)


github/github.go
```diff
@@ -241,13 +241,14 @@ func NewClient(httpClient *http.Client) *Client {
 // specified, the value pointed to by body is JSON encoded and included as the
 // request body.
 func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Request, error) {
-	rel, err := url.Parse(urlStr)
+	if c.BaseURL.Path != "" && !strings.HasSuffix(c.BaseURL.Path, "/") {
+		return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL)
+	}
+	u, err := c.BaseURL.Parse(urlStr)
```
shurcooL: Nice work with using `Parse` method to simplify the code. 👍

github/github.go
```diff
@@ -241,13 +241,14 @@ func NewClient(httpClient *http.Client) *Client {
 // specified, the value pointed to by body is JSON encoded and included as the
 // request body.
 func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Request, error) {
-	rel, err := url.Parse(urlStr)
+	if c.BaseURL.Path != "" && !strings.HasSuffix(c.BaseURL.Path, "/") {
+		return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL)
```
shurcooL: I appreciate that you were thoughtful and considered the `c.BaseURL.Path` being empty (i.e., a URL like `https://api.example.com`) case. You're also handling it correctly, because for root URLs, the trailing slash is actually not neccessary. (I've confirmed, it works correctly.)

However, I will suggest we don't make this special case treatment of the root path, even though it's possible.

Removing it has the following advantages:

- It makes the API simpler; for every type of URL, it must always have a trailing slash. This is simpler to communicate than "all URLs must have a trailing slash, except if the path is root, then it's optional."
- It eliminates ambiguity of whether the trailing slash should be included for root URLs. It would always need to be included.
- It makes the code simpler (less special cases).

So, my suggestion is to do this:

```Go
if !strings.HasSuffix(c.BaseURL.Path, "/") {
	return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL)
}
```

That'll mean root URLs will need to be provided as `https://api.example.com/` instead of `https://api.example.com`, which is completely okay.

What do you think?

178inaba: Because I got an error in the test, I checked the sky of Path.
Since the test allowed `http://127.0.0.1:50300`, I thought that an empty pass was allowed.
I think your proposal is correct.
I will fix the test.

178inaba: Done.

shurcooL: I see.

Yeah, fixing the test to use a trailing slash, as the docs require is great. Thanks!

github/github_test.go
```diff
@@ -248,6 +248,52 @@ func TestNewRequest_emptyBody(t *testing.T) {
 	}
 }
 
+func TestNewRequest_errorForNoTrailingSlash(t *testing.T) {
+	tests := []struct {
+		rawurl  string
+		isError bool
+	}{
+		{"https://example.com/api/v3", true},
+		{"https://example.com/api/v3/", false},
```
shurcooL: I'd suggest using keyed fields here to make the boolean values more readable. Just seeing `true` and `false` on their own is not as clear as:

```Go
{rawurl: "https://example.com/api/v3", wantError: true},
{rawurl: "https://example.com/api/v3/", wantError: false},
```

I'd also rename `isError` to `wantError`, using "want" for expected values is more consistent with style in Go tests.

178inaba: Thank you so much for review!
Fixed the test.
PTAL 🙇

gmlewis:
It looks like we made need to revert this change based on the [comment in #690](https://github.com/google/go-github/pull/690#issuecomment-323851845).
Thoughts, @shurcooL?

shurcooL:
@gmlewis See my reply at https://github.com/google/go-github/pull/690#issuecomment-323884954. I don't think we need to revert it yet, because this change is no-op for correct programs. If could cause other problems if the trailing slash is not included, so reporting an error is helpful to catch those kinds of issues.

If we have a few other reports of this causing issues, we can reconsider or look for a better way to handle the situation. I think we should wait and not act based one report, since there's not enough information to make a good decision. I've asked for more info there so we can see what caused it.

PR #693, opened 2017-08-15
Accept application/x-www-form-urlencoded webhooks

lkysow:
- Enable `ValidatePayload` to work for webhooks with Content-Type `application/x-www-form-urlencoded` in addition to `application/json`.
- Reject any Content-Type that is not `application/x-www-form-urlencoded` or `application/json`.

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -136,9 +136,6 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 //     }
 //
 func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err error) {
-	// payloadFormParam is the name of the form parameter that the JSON payload
-	// will be in if a webhook has its content type set to application/x-www-form-urlencoded.
-	const payloadFormParam = "payload"
 	var body []byte // Raw body that GitHub uses to calculate the signature.
 
 	switch ct := r.Header.Get("Content-Type"); ct {
@@ -148,11 +145,16 @@ func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err err
 		if err != nil {
 			return nil, err
 		}
+
 		// If the content type is application/json,
 		// the JSON payload is just the original body.
 		payload = body
 
 	case "application/x-www-form-urlencoded":
+		// payloadFormParam is the name of the form parameter that the JSON payload
+		// will be in if a webhook has its content type set to application/x-www-form-urlencoded.
+		const payloadFormParam = "payload"
+
 		var err error
 		body, err = ioutil.ReadAll(r.Body)
 		if err != nil {
--- a/github/messages.go
+++ b/github/messages.go
@@ -141,8 +141,7 @@ func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err err
 	switch ct := r.Header.Get("Content-Type"); ct {
 	case "application/json":
 		var err error
-		body, err = ioutil.ReadAll(r.Body)
-		if err != nil {
+		if body, err = ioutil.ReadAll(r.Body); err != nil {
 			return nil, err
 		}
 
@@ -156,8 +155,7 @@ func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err err
 		const payloadFormParam = "payload"
 
 		var err error
-		body, err = ioutil.ReadAll(r.Body)
-		if err != nil {
+		if body, err = ioutil.ReadAll(r.Body); err != nil {
 			return nil, err
 		}
 
--- a/github/messages.go
+++ b/github/messages.go
@@ -126,7 +126,8 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 
 // ValidatePayload validates an incoming GitHub Webhook event request
 // and returns the (JSON) payload.
-// The payload can be of Content-Type application/json or application/x-www-form-urlencoded
+// The Content-Type header of the payload can be "application/json" or "application/x-www-form-urlencoded".
+// If the Content-Type is neither then an error is returned.
 // secretKey is the GitHub Webhook secret message.
 //
 // Example usage:
@@ -143,7 +144,8 @@ func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
 	// rawPayload is the raw body that GitHub uses to calculate the signature
 	var rawPayload []byte
 
-	if r.Header.Get("Content-Type") == "application/x-www-form-urlencoded" {
+	switch c := r.Header.Get("Content-Type"); c {
+	case "application/x-www-form-urlencoded":
 		// If the content type is application/x-www-form-urlencoded
 		// the json payload will be under the "payload" form param
 		jsonPayload = []byte(r.FormValue(payloadFormParam))
@@ -151,7 +153,7 @@ func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
 		// GitHub calculates the signature based on the query-escaped
 		// post body. In order to validate, we need to reconstruct the raw body.
 		rawPayload = []byte(fmt.Sprintf("%s=%s", payloadFormParam, url.QueryEscape(string(jsonPayload))))
-	} else {
+	case "application/json":
 		defer r.Body.Close()
 		var err error
 		jsonPayload, err = ioutil.ReadAll(r.Body)
@@ -161,6 +163,8 @@ func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
+	default:
+		return nil, fmt.Errorf("Webhook request has unsupported Content-Type %q", c)
 	}
 
 	sig := r.Header.Get(signatureHeader)
--- a/github/messages.go
+++ b/github/messages.go
@@ -36,9 +36,6 @@ const (
 	eventTypeHeader = "X-Github-Event"
 	// deliveryIDHeader is the GitHub header key used to pass the unique ID for the webhook event.
 	deliveryIDHeader = "X-Github-Delivery"
-	// payloadFormParam is the name of the form parameter that the JSON payload
-	// will be in if a webhook has its content type set to application/x-www-form-urlencoded
-	payloadFormParam = "payload"
 )
 
 var (
@@ -138,39 +135,47 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 //       // Process payload...
 //     }
 //
-func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
-	var (
-		jsonPayload []byte // JSON webhook data.
-		rawPayload  []byte // Raw body that GitHub uses to calculate the signature.
-	)
-	switch ct := r.Header.Get("Content-Type"); ct {
-	case "application/x-www-form-urlencoded":
-		// If the content type is application/x-www-form-urlencoded
-		// the json payload will be under the "payload" form param.
-		jsonPayload = []byte(r.FormValue(payloadFormParam))
+func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err error) {
+	// payloadFormParam is the name of the form parameter that the JSON payload
+	// will be in if a webhook has its content type set to application/x-www-form-urlencoded.
+	const payloadFormParam = "payload"
+	var body []byte // Raw body that GitHub uses to calculate the signature.
 
-		// GitHub calculates the signature based on the query-escaped
-		// post body. In order to validate, we need to reconstruct the raw body.
-		rawPayload = []byte(payloadFormParam + "=" + url.QueryEscape(string(jsonPayload)))
+	switch ct := r.Header.Get("Content-Type"); ct {
 	case "application/json":
 		var err error
-		jsonPayload, err = ioutil.ReadAll(r.Body)
+		body, err = ioutil.ReadAll(r.Body)
 		if err != nil {
 			return nil, err
 		}
-		// If the content type is application/json then the payload
-		// used to calculate the signature is just the original body.
-		rawPayload = jsonPayload
+		// If the content type is application/json,
+		// the JSON payload is just the original body.
+		payload = body
+
+	case "application/x-www-form-urlencoded":
+		var err error
+		body, err = ioutil.ReadAll(r.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		// If the content type is application/x-www-form-urlencoded,
+		// the JSON payload will be under the "payload" form param.
+		form, err := url.ParseQuery(string(body))
+		if err != nil {
+			return nil, err
+		}
+		payload = []byte(form.Get(payloadFormParam))
 
 	default:
 		return nil, fmt.Errorf("Webhook request has unsupported Content-Type %q", ct)
 	}
 
 	sig := r.Header.Get(signatureHeader)
-	if err := validateSignature(sig, rawPayload, secretKey); err != nil {
+	if err := validateSignature(sig, body, secretKey); err != nil {
 		return nil, err
 	}
-	return jsonPayload, nil
+	return payload, nil
 }
 
 // validateSignature validates the signature for the given payload.
--- a/github/messages.go
+++ b/github/messages.go
@@ -20,6 +20,7 @@ import (
 	"hash"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 )
 
@@ -35,6 +36,9 @@ const (
 	eventTypeHeader = "X-Github-Event"
 	// deliveryIDHeader is the GitHub header key used to pass the unique ID for the webhook event.
 	deliveryIDHeader = "X-Github-Delivery"
+	// payloadFormParam is the name of the form parameter that the JSON payload
+	// will be in if a webhook has its content type set to application/x-www-form-urlencoded
+	payloadFormParam = "payload"
 )
 
 var (
@@ -122,6 +126,7 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 
 // ValidatePayload validates an incoming GitHub Webhook event request
 // and returns the (JSON) payload.
+// The payload can be of Content-Type application/json or application/x-www-form-urlencoded
 // secretKey is the GitHub Webhook secret message.
 //
 // Example usage:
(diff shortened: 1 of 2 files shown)


github/messages_test.go
```diff
@@ -82,6 +83,54 @@ func TestValidatePayload(t *testing.T) {
 	}
 }
 
+func TestValidateFormPayloadGET(t *testing.T) {
```
lkysow: I looked at adding this test to the original `TestValidatePayload` test cases but thought the additional complexity of adding it to that test was too much since you need to change how the test is constructing the request.

lkysow:
I signed it!

github/messages.go
```diff
@@ -132,17 +137,37 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
+	if r.Header.Get("Content-Type") == "application/x-www-form-urlencoded" {
+		// If the content type is application/x-www-form-urlencoded
+		// the json payload will be under the "payload" form param
+		jsonPayload = []byte(r.FormValue(payloadFormParam))
+
+		// GitHub calculates the signature based on the query-escaped
+		// post body. In order to validate, we need to reconstruct the raw body.
+		rawPayload = []byte(fmt.Sprintf("%s=%s", payloadFormParam, url.QueryEscape(string(jsonPayload))))
+	} else {
+		defer r.Body.Close()
```
shurcooL: `r.Body.Close()` wasn't being done before. Why is it done now?

github/messages.go
```diff
@@ -122,6 +126,7 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 
 // ValidatePayload validates an incoming GitHub Webhook event request
 // and returns the (JSON) payload.
+// The payload can be of Content-Type application/json or application/x-www-form-urlencoded
```
shurcooL: Nitpicks.

This sentence is missing a period at the end.

Also, I'd rephrase it as:

```
// The payload can have content type of "application/json" or "application/x-www-form-urlencoded".
```

`Content-Type` is the name of the header, but it's pretty natural to use the phrase "content type" or "type of content" in English. E.g., [`ProjectCardOptions.ContentType`](https://godoc.org/github.com/google/go-github/github#ProjectCardOptions.ContentType) is documented as "The type of content to associate with this card."

github/messages_test.go
shurcooL: That's a good call. It's better to add a separate test, then we can be more confident the old one still passes without changes.

lkysow:
@shurcooL regarding your question, the two different types of webhooks are documented here:
https://developer.github.com/webhooks/creating/#content-type

If you click "Add webhook" this is the screen you'll see where by default the formencoded Content-Type is selected. In our application that asks users to add webhooks we found that sometimes users would add the `formencoded` content-type (even though our docs say use json), so we wanted to support both webhook types.

image

shurcooL:
Awesome, thanks, that answers my question. Since GitHub supports both types, it would be a good enhancement for go-github to support them too. Thanks for sending the PR. My next step will be to review the code, I will get to that in a bit.

github/messages.go
lkysow: I think it was a bug with the original implementation. If you look at the docs for `http`: https://golang.org/pkg/net/http/ it's in the example. If you don't close it then the connection remains open: https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body-in-golang

image

github/messages_test.go
```diff
@@ -82,6 +83,54 @@ func TestValidatePayload(t *testing.T) {
...
+	secretKey := []byte("0123456789abcdef")
+
+	req, err := http.NewRequest("GET", "http://localhost/event", nil)
+	q := req.URL.Query()
+	q.Add("payload", payload)
+	req.URL.RawQuery = q.Encode()
+	if err != nil {
+		t.Fatalf("NewRequest: %v", err)
+	}
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
shurcooL: Since you only ever want one value for Content-Type, use `req.Header.Set` rather than `req.Header.Add`. It doesn't make a difference as long as you do it only once, but still, using `Set` is better form and more readable and consistent code.

github/messages_test.go
```diff
@@ -82,6 +83,54 @@ func TestValidatePayload(t *testing.T) {
...
+	secretKey := []byte("0123456789abcdef")
+
+	form := url.Values{}
+	form.Set("payload", payload)
+	buf := bytes.NewBufferString(form.Encode())
+	req, err := http.NewRequest("POST", "http://localhost/event", buf)
+	if err != nil {
+		t.Fatalf("NewRequest: %v", err)
+	}
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
shurcooL: Use `req.Header.Set("Content-Type", ...)` here as well.

github/messages_test.go
shurcooL: Rename the test to `TestValidatePayload_FormGet` for more consistent test naming.

github/messages_test.go
```diff
@@ -82,6 +83,54 @@ func TestValidatePayload(t *testing.T) {
...
+	got, err := ValidatePayload(req, secretKey)
+	if err != nil {
+		t.Errorf("ValidatePayload(%#v): err = %v, want nil", payload, err)
+	}
+	if string(got) != payload {
+		t.Errorf("ValidatePayload = %q, want %q", got, payload)
+	}
+}
+
+func TestValidateFormPayloadPOST(t *testing.T) {
```
shurcooL: And `TestValidatePayload_FormPost` here.

github/messages.go
```diff
@@ -132,17 +137,37 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
-	payload, err = ioutil.ReadAll(r.Body)
-	if err != nil {
-		return nil, err
+func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
+	// jsonPayload is the json webhook data
+	var jsonPayload []byte
+	// rawPayload is the raw body that GitHub uses to calculate the signature
+	var rawPayload []byte
+
+	if r.Header.Get("Content-Type") == "application/x-www-form-urlencoded" {
```
shurcooL: Since "application/x-www-form-urlencoded" and "application/json" are the only two content types we expect from GitHub, and support, let's explicitly check it's one of those two and return an error otherwise. It wouldn't make sense to treat it as JSON if the content type is not actually "application/json".

This is a good place to use a `switch`, like this:

```Go
switch r.Header.Get("Content-Type") {
case "application/x-www-form-urlencoded":
    ...
case "application/json":
    ...
default:
    return nil, fmt.Errorf("Webhook payload has unsupported Content-Type %q", r.Header.Get("Content-Type"))
}
```

github/messages_test.go
```diff
@@ -82,6 +83,54 @@ func TestValidatePayload(t *testing.T) {
...
+
+	got, err := ValidatePayload(req, secretKey)
+	if err != nil {
+		t.Errorf("ValidatePayload(%#v): err = %v, want nil", payload, err)
+	}
+	if string(got) != payload {
+		t.Errorf("ValidatePayload = %q, want %q", got, payload)
+	}
+}
+
```
shurcooL: Since the primary purpose of `ValidatePayload` is to reject invalid payloads, can you also add a test that for an invalid payload that correctly returns a non-nil error?

github/messages.go
lkysow: Totally agree. FYI though that this will be a backwards incompatible change since previously we never checked the content-type header and now we're rejecting anything without `application/json`. If people are using this as expected then GitHub would always send that header, however it is technically breaking BC.

lkysow:
@shurcooL updated

github/messages.go
shurcooL: I see.

However, this function is meant to be used on in HTTP handlers working on server HTTP requests, not client ones.

Specifically:

```Go
func (s *GitHubEventMonitor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
    payload, err := github.ValidatePayload(r, s.webhookSecretKey)
    if err != nil { ... }
    // Process payload...
}
```

HTTP handlers do not need to close the response body.

See documentation here:

https://godoc.org/net/http#Request.Body

> For server requests the Request Body is always non-nil
> but will return EOF immediately when no body is present.
> **The Server will close the request body. The ServeHTTP
> Handler does not need to.**

So, the original behavior was correct, and you should revert that change.

github/messages.go
shurcooL: I think that's fine, because GitHub will always be setting that header.

github/messages.go
```diff
@@ -132,17 +138,40 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
 //
-func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err error) {
-	payload, err = ioutil.ReadAll(r.Body)
-	if err != nil {
-		return nil, err
+func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
+	// jsonPayload is the json webhook data
+	var jsonPayload []byte
+	// rawPayload is the raw body that GitHub uses to calculate the signature
+	var rawPayload []byte
```
shurcooL: Style wise, it might be nicer to do this:

```Go
var (
    jsonPayload []byte // JSON webhook data.
    rawPayload  []byte // Raw body that GitHub uses to calculate the signature.
)
```

But it's up to you, use what you think looks nicer.

github/messages.go
```diff
@@ -132,17 +138,40 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
+		// post body. In order to validate, we need to reconstruct the raw body.
+		rawPayload = []byte(fmt.Sprintf("%s=%s", payloadFormParam, url.QueryEscape(string(jsonPayload))))
+	case "application/json":
+		defer r.Body.Close()
+		var err error
+		jsonPayload, err = ioutil.ReadAll(r.Body)
+		// If the content type is application/json then the raw body
+		// used to calculate the signature is just the original body.
+		rawPayload = jsonPayload
+		if err != nil {
```
shurcooL: This `if err != nil` check is out of place. Did you mean to place it directly under `jsonPayload, err = ioutil.ReadAll(r.Body)` line?

github/messages.go
```diff
@@ -132,17 +138,40 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
+
+	switch c := r.Header.Get("Content-Type"); c {
+	case "application/x-www-form-urlencoded":
+		// If the content type is application/x-www-form-urlencoded
+		// the json payload will be under the "payload" form param
+		jsonPayload = []byte(r.FormValue(payloadFormParam))
+
+		// GitHub calculates the signature based on the query-escaped
+		// post body. In order to validate, we need to reconstruct the raw body.
+		rawPayload = []byte(fmt.Sprintf("%s=%s", payloadFormParam, url.QueryEscape(string(jsonPayload))))
```
shurcooL: It feels to me that using `fmt.Sprintf` here is not being helpful, and the code would be simpler with just plain string concatenation.

How about:

```Go
rawPayload = []byte(payloadFormParam + "=" + url.QueryEscape(string(jsonPayload)))
```

github/messages.go
```diff
@@ -132,17 +138,40 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
-	payload, err = ioutil.ReadAll(r.Body)
-	if err != nil {
-		return nil, err
+func ValidatePayload(r *http.Request, secretKey []byte) ([]byte, error) {
+	// jsonPayload is the json webhook data
+	var jsonPayload []byte
+	// rawPayload is the raw body that GitHub uses to calculate the signature
+	var rawPayload []byte
+
+	switch c := r.Header.Get("Content-Type"); c {
```
shurcooL: I think `ct` would be a better, more readable variable name for Content-Type.

github/messages.go
lkysow: :facepalm: nice catch, will do

github/messages.go
lkysow: yes, moving

github/messages.go
lkysow: I like that better too

lkysow:
@shurcooL thanks for the review! Updated again based on comments.

lkysow:
@shurcooL @gmlewis any chance you guys can look at this again? I've made all requested changed.

github/messages.go
```diff
@@ -132,17 +138,39 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
+	)
+	switch ct := r.Header.Get("Content-Type"); ct {
+	case "application/x-www-form-urlencoded":
+		// If the content type is application/x-www-form-urlencoded
+		// the json payload will be under the "payload" form param.
+		jsonPayload = []byte(r.FormValue(payloadFormParam))
+
+		// GitHub calculates the signature based on the query-escaped
+		// post body. In order to validate, we need to reconstruct the raw body.
+		rawPayload = []byte(payloadFormParam + "=" + url.QueryEscape(string(jsonPayload)))
```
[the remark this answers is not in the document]
lkysow: @shurcooL agreed this is more straight-forward. My only hesitation is around your suggestion to limit the POST body to 10MB. Wouldn't we have to copy the code from https://golang.org/src/net/http/request.go#L1101

```
		var reader io.Reader = r.Body
		maxFormSize := int64(1<<63 - 1)
		if _, ok := r.Body.(*maxBytesReader); !ok {
			maxFormSize = int64(10 << 20) // 10 MB is a lot of text.
			reader = io.LimitReader(r.Body, maxFormSize+1)
		}
		b, e := ioutil.ReadAll(reader)
		if e != nil {
			if err == nil {
				err = e
			}
			break
		}
		if int64(len(b)) > maxFormSize {
			err = errors.New("http: POST too large")
			return
		}
		vs, e = url.ParseQuery(string(b))
		if err == nil {
			err = e
		}
```

Then I'm just worried that we're re-implementing a lot of logic from the core.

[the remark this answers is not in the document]
shurcooL: We don't have to copy all of that code, nor do we have to worry about keeping it in sync. We just need to consider putting in a similar 10 MB limit.

If we decide to do it, we can look at the net/http implementation for inspiration, but I think ours can be simpler:

```Go
const maxFormSize = int64(10 << 20)

var err error
body, err = ioutil.ReadAll(io.LimitReader(r.Body, maxFormSize+1))
if err != nil {
	return nil, err
}
if int64(len(body)) > maxFormSize {
	return nil, errors.New("POST too large")
}
```

Also, we should think about why there's a limit for POST but not for JSON. Does JSON need one too? Or is it okay not to have one for either?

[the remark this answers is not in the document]
lkysow: I've pushed your suggested changes *without* the 10MB limits. I've tested this locally with/without the secret being specified and with both Content-Types and it works as expected.

RE: 10MB limit, I'm still not sure what the right answer is. Can we get this in and then cut another issue to track it as it seems like a separate concern? I'm happy to create that.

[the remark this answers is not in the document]
shurcooL: I'm okay with making it a separate issue. This affects only the new feature, not the existing functionality anyway. I think we should gather information in that issue and only add the limit once we have a provable situation for when it's needed. Thanks.

lkysow:
Sounds good to me 👍  thanks for your excellent code review and explaining why you request certain changes. Really appreciate it.

lkysow:
Created #735 to track cap on body size

lkysow:
@gmlewis 
>  I'll ask @gmlewis to confirm the latest changes are okay with him too.
from above

gmlewis:
Sorry for the delay... taking a look now.

github/messages_test.go
```diff
@@ -82,6 +85,80 @@ func TestValidatePayload(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("ValidatePayload(%#v): err = %v, want nil", payload, err)
+	}
+	if string(got) != payload {
+		t.Errorf("ValidatePayload = %q, want %q", got, payload)
+	}
+
+	// check that if payload is invalid we get error
+	req.Header.Set(signatureHeader, "invalid signature")
+	_, err = ValidatePayload(req, nil)
```
gmlewis: nit:

```go
if _, err = ValidatePayload(req, nil); err == nil {
```

for Go style.

github/messages_test.go
```diff
@@ -82,6 +85,80 @@ func TestValidatePayload(t *testing.T) {
...
```
gmlewis: nit - same here..

```go
if _, err = ValidatePayload(req, nil); err == nil {
```

github/messages_test.go
```diff
@@ -82,6 +85,80 @@ func TestValidatePayload(t *testing.T) {
...
+	}
+}
+
+func TestValidatePayload_InvalidContentType(t *testing.T) {
+	req, err := http.NewRequest("POST", "http://localhost/event", nil)
+	if err != nil {
+		t.Fatalf("NewRequest: %v", err)
+	}
+	req.Header.Set("Content-Type", "invalid content type")
+	_, err = ValidatePayload(req, nil)
```
gmlewis: nit - and here...

```go
if _, err = ValidatePayload(req, nil); err == nil {
```

github/messages.go
```diff
@@ -133,13 +136,45 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
+		// the JSON payload is just the original body.
+		payload = body
+
+	case "application/x-www-form-urlencoded":
+		// payloadFormParam is the name of the form parameter that the JSON payload
+		// will be in if a webhook has its content type set to application/x-www-form-urlencoded.
+		const payloadFormParam = "payload"
+
+		var err error
+		body, err = ioutil.ReadAll(r.Body)
```
gmlewis: nit:

```go
if body, err = ioutil.ReadAll(r.Body); err != nil {
```

github/messages.go
```diff
@@ -133,13 +136,45 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
...
 func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err error) {
-	payload, err = ioutil.ReadAll(r.Body)
-	if err != nil {
-		return nil, err
+	var body []byte // Raw body that GitHub uses to calculate the signature.
+
+	switch ct := r.Header.Get("Content-Type"); ct {
+	case "application/json":
+		var err error
+		body, err = ioutil.ReadAll(r.Body)
```
gmlewis: nit:

```go
if body, err = ioutil.ReadAll(r.Body); err != nil {
```

gmlewis:
Please ignore the [googletbot comment](https://github.com/google/go-github/pull/693#issuecomment-332898713) as I have verified that the CLAs all LGTM.

lkysow:
@gmlewis done!

github/messages.go
gmlewis: Oh, OK.  I thought it was more idiomatic Go, but that's OK.  Feel free to switch it back... I don't feel that strongly about it.  Sorry for the switch... I forgot you preferred the other way.

shurcooL: > I thought it was more idiomatic Go

As far as I know, both are valid idiomatic Go. Which one is better to use depends on the context/situation.

I'll leave it as is, I don't feel very strongly either.

PR #697, opened 2017-08-16
Fix Google App Engine support and provide working example

### LINKED ISSUE #578: Using go-github on Google App Engine with recent context changes - any interest?

gmlewis:
Fixes #578.

### DIFF
--- a/example/appengine/app.go
+++ b/example/appengine/app.go
@@ -37,6 +37,7 @@ func handler(w http.ResponseWriter, r *http.Request) {
 	commits, _, err := client.Repositories.ListCommits(ctx, "google", "go-github", nil)
 	if err != nil {
 		log.Errorf(ctx, "ListCommits: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
--- a/example/appengine/app.go
+++ b/example/appengine/app.go
@@ -1,0 +1,47 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The demo app shows how to use the github package on Google App Engine.
+package demo
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/log"
+)
+
+func init() {
+	http.HandleFunc("/", handler)
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	if r.URL.Path != "/" {
+		http.NotFound(w, r)
+		return
+	}
+
+	ctx := appengine.NewContext(r)
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: os.Getenv("GITHUB_AUTH_TOKEN")},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+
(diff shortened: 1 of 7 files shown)


github/with_appengine.go
```diff
@@ -17,10 +17,8 @@ package github
 import (
 	"context"
 	"net/http"
-
-	"google.golang.org/appengine"
 )
 
 func withContext(ctx context.Context, req *http.Request) (context.Context, *http.Request) {
-	return appengine.WithContext(ctx, req), req
+	return ctx, req
```
shurcooL: Given this change, I think you can simplify `withContext` signature to return only `*http.Request` then. Can you?

shurcooL: Also, now that this func is a no-op for `appengine` case, I think you should add a comment explaining that. E.g.:

```Go
// No-op because appengine doesn't support adding context to a request.
return req
```

example/appengine/app.go
```diff
@@ -0,0 +1,47 @@
...
+		http.NotFound(w, r)
+		return
+	}
+
+	ctx := appengine.NewContext(r)
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: os.Getenv("GITHUB_AUTH_TOKEN")},
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
```
shurcooL: So a new client must be made for every incoming request? That seems pretty unfortunate.

example/appengine/app.go
```diff
@@ -0,0 +1,47 @@
...
+
+	commits, _, err := client.Repositories.ListCommits(ctx, "google", "go-github", nil)
+	if err != nil {
+		log.Errorf(ctx, "ListCommits: %v", err)
+		return
```
shurcooL: You haven't written anything to `w` yet, so you could also do `http.Error(w, err.Error(), http.Status...)` here.

github/github.go
```diff
@@ -399,7 +399,7 @@ func parseRate(r *http.Response) Rate {
 // The provided ctx must be non-nil. If it is canceled or times out,
 // ctx.Err() will be returned.
 func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
-	ctx, req = withContext(ctx, req)
+	req = withContext(ctx, req)
```
shurcooL: This is one of those rare times where I feel it might be worth breaking the "context is always the first parameter" pattern.

Here, `withContext` is basically a method-as-a-function that adds `ctx` to `req`. I think seeing it like this:

```Go
req = withContext(req, ctx)
```

Might read better...

Well, after writing it up and looking at it for a few minutes, I'm not so sure it looks better. It seems about about the same. 😑

example/appengine/app.go
gmlewis: This was expedient for this demo, but I believe it is possible to make a single client as a package global during initialization.

One caveat is that you can't call `os.Getenv` on App Engine within an `init` function due to the environment not being fully passed-in on app startup. The details are quite hairy.

So in cases like that, you could have a `sync.Once` block get the environment variable value and create the client before its first use... but I wanted to have a simple example here and not complicate things too much.

github/github.go
gmlewis: Agreed.  I'll leave it alone.

gmlewis:
Thank you, @shurcooL! Merging.

PR #700, opened 2017-08-17
Removed Inconsistency in documentation for ProjectCardOptions(#696)

### LINKED ISSUE #696: Inconsistency in documentation for ProjectCardOptions.ContentType field.
https://github.com/google/go-github/blob/cf4400336af2282a4879072b0d1c17b7d8c0207e/github/projects.go#L326-L337

The documentation for `ContentType` field says:

> Possible values are: "Issue", "PullRequest".

I noticed at https://developer.github.com/v3/projects/cards/, it says:

> Can only be "Issue" at this time.

This is an inconsistency, which can be investigated and resolved.

I've checked, and our documentation is from 11 months ago when projects support was initially added in #438 (/cc @gmlewis who created it, FYI).

I suspect it's likely that GitHub has changed things on their end since 11 months ago, hence our side being slightly out of date and inconsistent. This appears to be limited to documentation at first glance, but further investigation can confirm that.

scriptonist:
Fixes #696
Inconsistent documentation was found in file github/projects.go
Edited the doc line to match corresponding [Github docs](https://developer.github.com/v3/projects/cards/).

### DIFF
--- a/github/projects.go
+++ b/github/projects.go
@@ -332,7 +332,7 @@ type ProjectCardOptions struct {
 	// The ID (not Number) of the Issue or Pull Request to associate with this card.
 	// Note and ContentID are mutually exclusive.
 	ContentID int `json:"content_id,omitempty"`
-	// The type of content to associate with this card. Possible values are: "Issue", "PullRequest".
+	// The type of content to associate with this card. Possible values are: "Issue".
 	ContentType string `json:"content_type,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


github/projects.go
```diff
@@ -332,7 +332,7 @@ type ProjectCardOptions struct {
 	// The ID (not Number) of the Issue or Pull Request to associate with this card.
```
shurcooL: I think this line also needs to be updated. It says:

```
// The ID (not Number) of the Issue or Pull Request to associate with this card.
```

But GitHub documentation says:

```
The id of the issue to associate with this card.
```

scriptonist: Soory missed it.Should I close this request and open a new one with change ?

scriptonist: I'm sorry, Missed it.I've fixed it now. Hope it's okay now.

PR #701, opened 2017-08-20
search: fix url encoding

nikhita:
Github API uses a search syntax with colon(s), etc in the URL (as mentioned in https://developer.github.com/v3/search/).

When `params.Encode()` is used, it replaces the characters outside the ASCII set with their encoded value. Example: `:` is replaced by `%3A`.

Using `url.QueryUnescape()` will prevent this conversion.

**Example**:

- Query before this change: `q=is%3Apr%2Brepo%3Akubernetes%2Fkubernetes%2Bauthor%3Anikhita`.
    - This results in an incorrect response: https://api.github.com/search/issues?q=is%3Apr%2Brepo%3Akubernetes%2Fkubernetes%2Bauthor%3Anikhita.
- Query after this change: `q=is:pr+repo:kubernetes/kubernetes+author:nikhita`.
    - This results in a correct response: https://api.github.com/search/issues?q=is:pr+repo:kubernetes/kubernetes+author:nikhita.

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"net/url"
 
 	qs "github.com/google/go-querystring/query"
 )
@@ -176,7 +177,13 @@ func (s *SearchService) search(ctx context.Context, searchType string, query str
 		return nil, err
 	}
 	params.Set("q", query)
-	u := fmt.Sprintf("search/%s?%s", searchType, params.Encode())
+
+	encodedParams, err := url.QueryUnescape(params.Encode())
+	if err != nil {
+		return nil, err
+	}
+
+	u := fmt.Sprintf("search/%s?%s", searchType, encodedParams)
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {


gmlewis:
Thank you, @nikhita! Would you mind adding a unit test to cover this change?

shurcooL:
Thanks for reporting the issue and working on a fix.

I'd like to understand this issue better. I think there's a good chance the code can be simplified.

The examples you provided seem to confirm that the current encoding isn't right. I'm a bit surprised that this wasn't already covered by tests and that it took this long to see an issue report, given that `search` unexported method is used by all higher level methods. I wonder if it's because not all search queries are affected, or because few people use Search API.

My main initial question is, can you tell me what would be the input string that you'd expect to be encoded as `q=is:pr+repo:kubernetes/kubernetes+author:nikhita`? (I think this question would also be answered by a test case.)

nikhita:
@shurcooL @gmlewis Sorry, it's my mistake here!

> My main initial question is, can you tell me what would be the input string that you'd expect to be encoded as `q=is:pr+repo:kubernetes/kubernetes+author:nikhita`?

I used the input query string as `is:pr+repo:kubernetes/kubernetes+author:nikhita` (which is wrong).
- the query is encoded as `q=is%3Apr%2Brepo%3Akubernetes%2Fkubernetes%2Bauthor%3Anikhita`.
- which results in an incorrect response - https://api.github.com/search/issues?q=is%3Apr%2Brepo%3Akubernetes%2Fkubernetes%2Bauthor%3Anikhita.

But the input string should have been `is:pr repo:kubernetes/kubernetes author:nikhita`.
- the query is encoded as `q=is%3Apr+repo%3Akubernetes%2Fkubernetes+author%3Anikhita`.
- which results in a correct response - https://api.github.com/search/issues?q=is%3Apr+repo%3Akubernetes%2Fkubernetes+author%3Anikhita.

Again, sorry for the trouble! Closing.

shurcooL:
No problem, glad we got to the bottom of it!

There are 2 other improvements I think we can make as a result of this.

1. The godoc can be improved to give an example query. Right now, https://godoc.org/github.com/google/go-github/github#SearchService doesn't say much. I think it'd be helpful to give an example query that can be done. Something similar to the query you posted here.

2. The test coverage of our search funcs didn't catch a regression in escaping behavior. We have test coverage, but it only covers simple queries. We should add another test case that covers a query with spaces and colons. That way, this functionality won't accidentally regress in the future.

I'll opened a new issue #707 to track it.

PR #703, opened 2017-08-20
Fix #664. Docs for empty DismissalRestrictions were incorrect.  [] -> {}

### LINKED ISSUE #664: PullRequestReviewsEnforcementRequest.DismissalRestrictionsRequest set to nil results in 422 from GH

```
&github.ProtectionRequest{
	}
```
results in:
```
```

beiriannydd:
I've driveby added the require owner commit attribute as an omitmissing as it
was missing from the API.  Because it is omitmissing it is not a breaking change.

Added spew library which helped me discover that we were comparing returned URLs from the
AdminEnforcement object.  None of the other objects in the API care about their URLs so
I just removed that.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -7,7 +7,6 @@ package github
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 )
@@ -559,9 +558,11 @@ type RequiredStatusChecks struct {
 // PullRequestReviewsEnforcement represents the pull request reviews enforcement of a protected branch.
 type PullRequestReviewsEnforcement struct {
 	// Specifies which users and teams can dismiss pull requets reviews.
-	DismissalRestrictions DismissalRestrictions `json:"dismissal_restrictions"`
+	DismissalRestrictions DismissalRestrictions `json:"dismissal_restrictions,omitempty"`
 	// Specifies if approved reviews are dismissed automatically, when a new commit is pushed.
-	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
+	DismissStaleReviews bool `json:"dismiss_stale_reviews,omitempty"`
+	// Specifies if review by the code owner is required
+	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews,omitempty"`
 }
 
 // PullRequestReviewsEnforcementRequest represents request to set the pull request review
@@ -569,32 +570,11 @@ type PullRequestReviewsEnforcement struct {
 // because the request structure is different from the response structure.
 type PullRequestReviewsEnforcementRequest struct {
 	// Specifies which users and teams should be allowed to dismiss pull requets reviews. Can be nil to disable the restrictions.
-	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
 	// Specifies if approved reviews can be dismissed automatically, when a new commit is pushed. (Required)
-	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
-}
-
-// MarshalJSON implements the json.Marshaler interface.
-// Converts nil value of PullRequestReviewsEnforcementRequest.DismissalRestrictionsRequest to empty array
-func (req PullRequestReviewsEnforcementRequest) MarshalJSON() ([]byte, error) {
-	if req.DismissalRestrictionsRequest == nil {
-		newReq := struct {
-			R []interface{} `json:"dismissal_restrictions"`
-			D bool          `json:"dismiss_stale_reviews"`
-		}{
-			R: []interface{}{},
-			D: req.DismissStaleReviews,
-		}
-		return json.Marshal(newReq)
-	}
-	newReq := struct {
-		R *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
-		D bool                          `json:"dismiss_stale_reviews"`
-	}{
-		R: req.DismissalRestrictionsRequest,
-		D: req.DismissStaleReviews,
-	}
-	return json.Marshal(newReq)
+	DismissStaleReviews bool `json:"dismiss_stale_reviews,omitempty"`
+	// Specifies if review by the code owner is required
+	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews,omitempty"`
 }
 
 // PullRequestReviewsEnforcementUpdate represents request to patch the pull request review
@@ -609,8 +589,7 @@ type PullRequestReviewsEnforcementUpdate struct {
 
 // AdminEnforcement represents the configuration to enforce required status checks for repository administrators.
 type AdminEnforcement struct {
-	URL     *string `json:"url,omitempty"`
-	Enabled bool    `json:"enabled"`
+	Enabled bool `json:"enabled"`
 }
 
 // BranchRestrictions represents the restriction that only certain users or
--- a/github/repos.go
+++ b/github/repos.go
@@ -569,7 +569,9 @@ type PullRequestReviewsEnforcement struct {
 // enforcement of a protected branch. It is separate from PullRequestReviewsEnforcement above
 // because the request structure is different from the response structure.
 type PullRequestReviewsEnforcementRequest struct {
-	// Specifies which users and teams should be allowed to dismiss pull requets reviews. Can be nil to disable the restrictions.
+	// Specifies which users and teams should be allowed to dismiss pull requets reviews.
+	// MUST be nil for non-organization repositories, may be empty on an organization repo
+	// to remove existing restrictions
 	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
 	// Specifies if approved reviews can be dismissed automatically, when a new commit is pushed. (Required)
 	DismissStaleReviews bool `json:"dismiss_stale_reviews,omitempty"`
@@ -626,9 +628,11 @@ type DismissalRestrictions struct {
 // different from the response structure.
 type DismissalRestrictionsRequest struct {
 	// The list of user logins who can dismiss pull request reviews. (Required; use []string{} instead of nil for empty list.)
-	Users []string `json:"users"`
+	// If both are nil, then this will be an empty object which is what we want to pass for
+	// no restrictions.
+	Users []string `json:"users,omitempty"`
 	// The list of team slugs which can dismiss pull request reviews. (Required; use []string{} instead of nil for empty list.)
-	Teams []string `json:"teams"`
+	Teams []string `json:"teams,omitempty"`
 }
 
 // ListBranches lists branches for the specified repository.
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -543,7 +543,6 @@ func TestRepositoriesService_GetBranchProtection(t *testing.T) {
 			},
 		},
 		EnforceAdmins: &AdminEnforcement{
-			URL:     String("/repos/o/r/branches/b/protection/enforce_admins"),
 			Enabled: true,
 		},
 		Restrictions: &BranchRestrictions{
@@ -868,7 +867,6 @@ func TestRepositoriesService_GetAdminEnforcement(t *testing.T) {
 	}
 
 	want := &AdminEnforcement{
-		URL:     String("/repos/o/r/branches/b/protection/enforce_admins"),
 		Enabled: true,
 	}
 
@@ -893,7 +891,6 @@ func TestRepositoriesService_AddAdminEnforcement(t *testing.T) {
 	}
 
 	want := &AdminEnforcement{
-		URL:     String("/repos/o/r/branches/b/protection/enforce_admins"),
 		Enabled: true,
 	}
 	if !reflect.DeepEqual(enforcement, want) {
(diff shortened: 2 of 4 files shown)


beiriannydd:
So, tests fail, looks as though that's because I added spew.  It really helped me debug, if you want me to remove then I can do that.

github/repos.go
```diff
@@ -559,42 +558,23 @@ type RequiredStatusChecks struct {
...
 	// Specifies which users and teams should be allowed to dismiss pull requets reviews. Can be nil to disable the restrictions.
-	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
 	// Specifies if approved reviews can be dismissed automatically, when a new commit is pushed. (Required)
-	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
-}
-
-// MarshalJSON implements the json.Marshaler interface.
-// Converts nil value of PullRequestReviewsEnforcementRequest.DismissalRestrictionsRequest to empty array
-func (req PullRequestReviewsEnforcementRequest) MarshalJSON() ([]byte, error) {
```
beiriannydd: No longer necessary as we don't need to hack an empty object into an empty array.

github/repos.go
```diff
@@ -609,8 +589,7 @@ type PullRequestReviewsEnforcementUpdate struct {
 
 // AdminEnforcement represents the configuration to enforce required status checks for repository administrators.
 type AdminEnforcement struct {
-	URL     *string `json:"url,omitempty"`
```
beiriannydd: This is the only time I've seen the returned URL from github be used in a structure in this code.  It is hard to test too.  It was causing my test to fail while correct because URL was being compared as nil.

github/repos_test.go
```diff
@@ -917,20 +914,6 @@ func TestRepositoriesService_RemoveAdminEnforcement(t *testing.T) {
 	}
 }
 
-func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctions(t *testing.T) {
```
beiriannydd: No need to test this function which we've removed.

test/integration/repos_test.go
```diff
@@ -151,9 +150,7 @@ func TestRepositories_EditBranches(t *testing.T) {
 		Restrictions: nil,
 	}
 	if !reflect.DeepEqual(protection, want) {
-		// Using spew helped me see that we were comparing an unwanted URL as part of
-		// AdminEnforcement structure.  Maybe add to other %+v tests?
-		t.Error(spew.Sprintf("Repositories.UpdateBranchProtection() returned %+v, want %+v", protection, want))
+		t.Errorf("Repositories.UpdateBranchProtection() returned %+v, want %+v", protection, want)
```
shurcooL: We have something similar to spew already available, you can/should use it instead.

See https://godoc.org/github.com/google/go-github/github#Stringify.

shurcooL:
Thanks for working on this PR @beiriannydd.

Before proceeding, we should look at the reason `[]` was being used in the first place and confirm that the same reason is no longer valid.

I was hasty in applying the bug label to issue #664, that's my fault. I think we first need to confirm this is a bug. As you wrote in a code comment, it might be that we're getting 422 error because of an invalid precondition, like trying to use team/user permissions on a non-organization repo.

Basically, how we confirm that the previous behavior is not valid and the new one is valid? How did you test this PR?

github/repos.go
shurcooL: Can you explain why it's no longer needed to be done?

At https://developer.github.com/v3/repos/branches/#update-branch-protection, it still says:

> You can disable dismissal restrictions by passing the `dismissal_restrictions` object as an empty array.

Which is why we did this in the first place in #637.

beiriannydd: >Ivan Žužak <support@github.com> | Ivan Žužak <support@github.com> | Aug 10 (12 days ago)
>Thanks so much for reporting this, Faye -- I believe that's a typo in the docs. Instead of >passing an empty array [], you should pass in an empty object {}. Can you give that a try >and let us know if it help?

Yep, that is the issue alright. I tested your suggestion and it worked. Thanks!

test/integration/repos_test.go
beiriannydd: Cool I will look at that.

beiriannydd:
I needed to make a change to make my code which operates on our organization github repos and protects branches work.  I originally fixed it by replacing []interface{}{} with map[string]interface{}{} but then it dawned on me that this was all to work around the normal behavior.  The updateBranch test was failing integration and after the patch it doesn't fail which means it is happy with what we sent, and we get back what we expect in this situation.

github/repos.go
beiriannydd: Perhaps I should have made a PR against their API docs before this one? ;)

github/repos.go
```diff
@@ -559,42 +558,23 @@ type RequiredStatusChecks struct {
...
+	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews,omitempty"`
 }
 
 // PullRequestReviewsEnforcementRequest represents request to set the pull request review
 // enforcement of a protected branch. It is separate from PullRequestReviewsEnforcement above
 // because the request structure is different from the response structure.
 type PullRequestReviewsEnforcementRequest struct {
 	// Specifies which users and teams should be allowed to dismiss pull requets reviews. Can be nil to disable the restrictions.
-	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
```
beiriannydd: Your point has made me question if I am being too liberal with omitempty here.

test/integration/repos_test.go
```diff
@@ -108,20 +108,16 @@ func TestRepositories_EditBranches(t *testing.T) {
...
 			Strict:   true,
 			Contexts: []string{"continuous-integration"},
 		},
 		RequiredPullRequestReviews: &github.PullRequestReviewsEnforcementRequest{
-			DismissalRestrictionsRequest: &github.DismissalRestrictionsRequest{
-				Users: []string{},
-				Teams: []string{},
-			},
-			DismissStaleReviews: true,
+			// Nil is perfectly acceptable here now and it proves the test.
```
beiriannydd: Although I guess the question is whether when the branch was protected with users in this list and we want to remove them, if that will happen with this configuration or if it would be a noop.

github/repos.go
shurcooL: Yes, it looks out of place. The comment says:

```
Can be nil to disable the restrictions.
```

And I suspect it needs to be without `omitempty` to function. That way, using a `nil` value for `DismissalRestrictionsRequest` field has a chance of disabling the restrictions.

However, I can't be certain until I understand the change at large.

github/repos.go
shurcooL: Can you please provide more information about this? It looks like you've pasted a portion of a reply from github support. Can you post more context?

I _think_ I understand what's happening, but I want to confirm. My understanding is that github support has confirmed on August 10th that the GitHub documentation has a typo in:

> You can disable dismissal restrictions by passing the `dismissal_restrictions` object as an empty array.

And they meant to write:

> You can disable dismissal restrictions by passing the `dismissal_restrictions` object as an empty object.

Is that what's happening here?

beiriannydd: Yes that is exactly right.

github/repos.go
beiriannydd: My question is whether there is a tristate behavior:
present and {} = disable existing setting
not present = leave existing setting? (hard to test on non-org repo)
present and {users:[], groups:[]} = configuring new settings.

github/repos.go
```diff
@@ -559,42 +558,25 @@ type RequiredStatusChecks struct {
...
 }
 
 // PullRequestReviewsEnforcementRequest represents request to set the pull request review
 // enforcement of a protected branch. It is separate from PullRequestReviewsEnforcement above
 // because the request structure is different from the response structure.
 type PullRequestReviewsEnforcementRequest struct {
-	// Specifies which users and teams should be allowed to dismiss pull requets reviews. Can be nil to disable the restrictions.
-	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+	// Specifies which users and teams should be allowed to dismiss pull requets reviews.
+	// MUST be nil for non-organization repositories, may be empty on an organization repo
```
beiriannydd: This clarification is based on additional work I did to make it possible to pass an empty DismissalRestrictionsRequest{} and have that convert to a json empty object.  Returning it to that state causes it to 422 again (as expected), but it works when used on an org repo (internal application testing)

beiriannydd:
Hopefully someone can add the $COVERALLS_TOKEN to the build box soon.

willnorris:
> Hopefully someone can add the $COVERALLS_TOKEN to the build box soon.

Strange, it was working before.  I'll drop the coveralls code for now so it doesn't break Travis test reporting.

willnorris:
try rebasing onto master, and travis should hopefully work now.

beiriannydd:
Thanks @willnorris

test/integration/repos_test.go
```diff
@@ -108,28 +108,25 @@ func TestRepositories_EditBranches(t *testing.T) {
...
 			Strict:   true,
 			Contexts: []string{"continuous-integration"},
 		},
 		RequiredPullRequestReviews: &github.PullRequestReviewsEnforcementRequest{
-			DismissalRestrictionsRequest: &github.DismissalRestrictionsRequest{
-				Users: []string{},
-				Teams: []string{},
-			},
-			DismissStaleReviews: true,
+			// The way to disable the setting is to send: &github.DismissalRestrictionsRequest{},
```
beiriannydd:
```
2017/08/27 07:44:17 repo name: ManagementGateway, release branch name: release/8.9
2017/08/27 07:44:17   updated protection: &{RequiredStatusChecks:0xc4201b1c00 RequiredPullRequestReviews:0xc420192ec0 EnforceAdmins:0xc4200fca46 Restrictions:0xc42019f3e0}
2017/08/27 07:44:17   [gh resp]: github.Rate{Limit:5000, Remaining:4881, Reset:github.Timestamp{2017-08-27 08:33:00 +0000 UTC}}
faye-salwin:delivery faye.salwin$ grep -5 DismissalRestrictionsRequest *
delivery.go-			RequiredStatusChecks: &github.RequiredStatusChecks{
delivery.go-				Strict:        true,
delivery.go-				Contexts:      r.Contexts,
delivery.go-			},
delivery.go-			RequiredPullRequestReviews: &github.PullRequestReviewsEnforcementRequest{
delivery.go:				DismissalRestrictionsRequest: &github.DismissalRestrictionsRequest{},
delivery.go-				DismissStaleReviews: true,
delivery.go-				RequireCodeOwnerReviews: false,
delivery.go-			},
delivery.go-			Restrictions: &github.BranchRestrictionsRequest{
delivery.go-				Users: []string{},

```
This is the code I've tested the empty object with.

github/repos.go
shurcooL: Did GitHub support say if they're planning to update/fix the documentation on their side? It's been 12 days and https://developer.github.com/v3/repos/branches/#update-branch-protection still says empty array rather than empty object.

shurcooL:
Unfortunately, I'm finding this PR pretty hard to review and make progress. I have a general understanding of the changes, but I don't have them completely fully grasped to be confident.

It makes many changes in important places that I'd want to discuss and confirm (e.g., changing `omitempty`s). There are some typos and inconsistencies in documentation that we'll need to resolve.

I wish it were smaller, but I'm not sure viable to split this PR into smaller changes (may be it is, maybe not).

@gmlewis Do you have any thoughts on this or ideas for what we can do to make it easier to make progress here?

gmlewis:
I've looked at this PR and am also finding it a bit overwhelming... so I decided to look at one tiny detail - the removal of the `URL` field from the `AdminEnforcement` struct around line 612.

I found that this field was added in #610 and then looked up the API in the GitHub Developer docs and found this:
https://developer.github.com/v3/repos/branches/#get-admin-enforcement-of-protected-branch
which makes it appear as though `URL` is still fully supported.

So I went back through the comments of this PR and saw that the only reason for removing this field was that it made testing more difficult and that it was not seen in the wild.

At the very least, I would say that to make progress on this PR, we really need to strip out changes such as this one (that appear to be unrelated to the root issue) into separate PRs where they can be discussed individually on their own merits.

Ideally, it would be nice to have detailed instructions on the steps needed to test out these API changes... so that the maintainers can confirm that the PR changes are necessary.

(And while I'm here, I thought the whole point of `Stringify` was to be able to add a `%v` (or `%s`) to a format string so that the object could be printed (via its `String` method) so that the `Stringify` method did not need to be called directly... but that's a separate issue.)

beiriannydd:
I'm going to close this PR, I'll continue to patch to have this functionality locally.  I don't have time to go back and re-implement it in smaller pieces.  Thanks for your consideration.  Things to bear in mind: when the URL is returned, it is for a randomly created repository.  You'll need to match this in your tests.  I didn't have time to do this, but wanted passing tests to mean something.  You'll notice that I did not remove any tests on the URL field.  Someone re-adding URL field should also make the working test, which I think would be the right way to do this.  I'm sorry that this PR is hard to follow, but that's mainly because there have been so many changes approved while this was an acceptable failure.

shurcooL:
Thanks for the update @beiriannydd. I understand your position. It's completely okay to withdraw from working on a PR if it's no longer a priority and you're unable to take it further. This is a hard change to understand, to write code for, and to review.

Thank you for contributing anyway, and I'm sure the information and code available in this PR will still be useful to us.

We'll try to figure out how to best proceed from here.

PR #702, opened 2017-08-20
README: add rate limit info for search API

nikhita:
The Search API has a separate rate limit. This is not mentioned in the README and could be misleading.

Link to the Github API docs: https://developer.github.com/v3/search/#rate-limit.

PR #706, opened 2017-08-23
[#647] Add 'affiliation' to ListCollaborators.

### LINKED ISSUE #647: feat: add support for affiliation in the Repositories.ListCollaborators opts
As per the [apidoc](https://developer.github.com/v3/repos/collaborators/#list-collaborators), at some point a request option was added for filtering the returned collaborators by `affiliation`. It would be useful if that was exposed as part of a custom ListCollaboratorOptions struct

e-beach:
Fixes #647.

Let me know if there is anything that needs to be changed or can be improved. 
Thank you!

### DIFF
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -16,7 +16,8 @@ type ListCollaboratorsOptions struct {
 	// Affiliation specifies how collaborators should be filtered by their affiliation.
 	// Possible values are:
 	//     outside - All outside collaborators of an organization-owned repository
-	//     direct - All collaborators with permissions to an organization-owned repository, regardless of organization membership status
+	//     direct - All collaborators with permissions to an organization-owned repository,
+	//     regardless of organization membership status
 	//     all - All collaborators the authenticated user can see
 	//
 	// Default value is "all".
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -13,8 +13,6 @@ import (
 // ListCollaboratorsOptions specifies the optional parameters to the
 // RepositoriesService.ListCollaborators method.
 type ListCollaboratorsOptions struct {
-	ListOptions
-
 	// Affiliation specifies how collaborators should be filtered by their affiliation.
 	// Possible values are:
 	//     outside - All outside collaborators of an organization-owned repository
@@ -23,6 +21,8 @@ type ListCollaboratorsOptions struct {
 	//
 	// Default value is "all".
 	Affiliation string `url:"affiliation,omitempty"`
+
+	ListOptions
 }
 
 // ListCollaborators lists the GitHub users that have access to the repository.
--- a/github/repos_collaborators.go
+++ b/github/repos_collaborators.go
@@ -10,10 +10,25 @@ import (
 	"fmt"
 )
 
+// ListCollaboratorsOptions specifies the optional parameters to the
+// RepositoriesService.ListCollaborators method.
+type ListCollaboratorsOptions struct {
+	ListOptions
+
+	// Affiliation specifies how collaborators should be filtered by their affiliation.
+	// Possible values are:
+	//     outside - All outside collaborators of an organization-owned repository
+	//     direct - All collaborators with permissions to an organization-owned repository, regardless of organization membership status
+	//     all - All collaborators the authenticated user can see
+	//
+	// Default value is "all".
+	Affiliation string `url:"affiliation,omitempty"`
+}
+
 // ListCollaborators lists the GitHub users that have access to the repository.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/collaborators/#list
-func (s *RepositoriesService) ListCollaborators(ctx context.Context, owner, repo string, opt *ListOptions) ([]*User, *Response, error) {
+func (s *RepositoriesService) ListCollaborators(ctx context.Context, owner, repo string, opt *ListCollaboratorsOptions) ([]*User, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/collaborators", owner, repo)
 	u, err := addOptions(u, opt)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


github/repos_collaborators.go
```diff
@@ -10,10 +10,25 @@ import (
 	"fmt"
 )
 
+// ListCollaboratorsOptions specifies the optional parameters to the
+// RepositoriesService.ListCollaborators method.
+type ListCollaboratorsOptions struct {
+	ListOptions
```
shurcooL: Let me know if you find evidence to the contrary, but look at other similar Options structs, I think we tend to put `ListOptions` at the very bottom. E.g.:

https://github.com/google/go-github/blob/78072f320213a7e011b5fb2700bcb9eb90efec37/github/repos.go#L123-L156

Let's do that here too for consistency.

e-beach: @shurcooL edited.

e-beach:
Hmmm... 
This pull request, as well as #702 and #706, are build-failing because of an error in the coveralls, see this [gist](https://gist.github.com/e-beach/d1926dcd03f384f11f1337821e69556e). The error occurs because encrypted environment variable $COVERALLS_TOKEN is not set in build.

Coveralls support and this environment variable were added in 89e28f7f1be 8 days ago, and no PRs since then have had passing builds.

Something is wrong with the coveralls, but I can't debug further due to the environment variable being secured.

e-beach:
Another potential issue is that this paragraph, which better explains the `Affiliation` parameter, is not in the docs:

> For organization-owned repositories, the list of collaborators includes outside collaborators, organization members with access through team memberships, organization members with access through default organization permissions, and organization owners.

Should it be? Should the API be completely understandable only through by the github-go documentation, or should the developer be expected to read the API docs?

beiriannydd:
@e-beach you should be good to go now if you rebase your change against master.  It worked for my PR.

github/repos_collaborators_test.go
```diff
@@ -24,7 +24,35 @@ func TestRepositoriesService_ListCollaborators(t *testing.T) {
...
+		t.Errorf("Repositories.ListCollaborators returned error: %v", err)
+	}
+
+	want := []*User{{ID: Int(1)}, {ID: Int(2)}}
+	if !reflect.DeepEqual(users, want) {
+		t.Errorf("Repositori es.ListCollaborators returned %+v, want %+v", users, want)
+	}
+}
+
+func TestRepositoriesService_ListCollaborators_With_Affiliation(t *testing.T) {
```
shurcooL: Minor. The next test is called `TestRepositoriesService_ListCollaborators_invalidOwner`. Let's make this `TestRepositoriesService_ListCollaborators_withAffiliation` for consistency.

This pattern comes from pattern for naming examples, as described at the bottom of https://godoc.org/testing#hdr-Examples section. We're using it for tests too here.

github/repos_collaborators.go
```diff
@@ -10,10 +10,25 @@ import (
 	"fmt"
 )
 
+// ListCollaboratorsOptions specifies the optional parameters to the
+// RepositoriesService.ListCollaborators method.
+type ListCollaboratorsOptions struct {
+	// Affiliation specifies how collaborators should be filtered by their affiliation.
+	// Possible values are:
+	//     outside - All outside collaborators of an organization-owned repository
+	//     direct - All collaborators with permissions to an organization-owned repository, regardless of organization membership status
```
shurcooL: It might be nice to wrap this line, like so:

```Go
// Possible values are:
//     outside - All outside collaborators of an organization-owned repository
//     direct - All collaborators with permissions to an organization-owned repository,
//              regardless of organization membership status
//     all - All collaborators the authenticated user can see
```

Precedent:

https://github.com/google/go-github/blob/5f3187e17912ee2e3fbe2fa5bdada9d577d25e83/github/orgs_members.go#L26-L30

github/repos_collaborators_test.go
```diff
@@ -24,7 +24,35 @@ func TestRepositoriesService_ListCollaborators(t *testing.T) {
 		fmt.Fprintf(w, `[{"id":1}, {"id":2}]`)
 	})
 
-	opt := &ListOptions{Page: 2}
+	opt := &ListCollaboratorsOptions{
+		ListOptions: ListOptions{Page: 2},
+		Affiliation: "",
```
shurcooL: You could leave `Affiliation` out, the zero value is already the empty string.

shurcooL:
> Should it be? Should the API be completely understandable only through by the github-go documentation, or should the developer be expected to read the API docs?

It's fine to leave it out. We don't expect the entire GitHub API to be understood only by looking at our godoc. We provide a subset of the docs, just the basics. Users are expected to look at GitHub API docs for in depth details.

e-beach:
I believe I addressed @shurcooL's comments. I suggest that you might merge #709 together with this one, as both PRs introduce breaking changes.

github/repos_collaborators.go
```diff
@@ -16,7 +16,8 @@ type ListCollaboratorsOptions struct {
 	// Affiliation specifies how collaborators should be filtered by their affiliation.
 	// Possible values are:
 	//     outside - All outside collaborators of an organization-owned repository
-	//     direct - All collaborators with permissions to an organization-owned repository, regardless of organization membership status
+	//     direct - All collaborators with permissions to an organization-owned repository,
+	//     regardless of organization membership status
```
shurcooL: Did you do this differently than I suggested on purpose, or was it unintentional?

I still think the version I suggested looks a bit nicer, and it's definitely more consistent with the rest of the repo (see "precedent" snippet I included).

e-beach: It was unintentional - I'm sorry for that mistake on my part.

shurcooL: No worries, thanks for clearing it up!

github/repos_collaborators.go
```diff
@@ -10,10 +10,26 @@ import (
...
+	//
+	// Default value is "all".
+	Affiliation string `url:"affiliation,omitempty"`
+
+	ListOptions
+}
+
 // ListCollaborators lists the GitHub users that have access to the repository.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/collaborators/#list
```
shurcooL: I'll just quickly fixup this link to the more up-to-date https://developer.github.com/v3/repos/collaborators/#list-collaborators before merging.

PR #709, opened 2017-08-30
[#605] Use more appropriate field names

e-beach:
This PR renames two fields as per discussion in #605:
`Authorizations.UpdateAt` => `UpdatedAt`
`DeploymentStatus.UpdatedAt` => `PushedAt`

As this is a breaking API change, would it be appropriate to merge these changes as a single version bump with #706, also a breaking change?

### DIFF
--- a/github/repos_deployments.go
+++ b/github/repos_deployments.go
@@ -137,7 +137,7 @@ type DeploymentStatus struct {
 	Description   *string    `json:"description,omitempty"`
 	TargetURL     *string    `json:"target_url,omitempty"`
 	CreatedAt     *Timestamp `json:"created_at,omitempty"`
-	UpdatedAt     *Timestamp `json:"pushed_at,omitempty"`
+	PushedAt      *Timestamp `json:"pushed_at,omitempty"`
 	DeploymentURL *string    `json:"deployment_url,omitempty"`
 	RepositoryURL *string    `json:"repository_url,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


shurcooL:
> As this is a breaking API change, would it be appropriate to merge these changes as a single version bump with #706, also a breaking change?

We can merge them soon one after another, but I don't think it should be a goal to merge unrelated breaking API changes just because they both break the API. In fact, it's better not to, because then people can upgrade one version at a time if they wish. Also, it makes reverting one of them easier, if necessary (not that reverting is likely to happen, but there's always a non-zero chance). They're separate changes, let's treat them that way.

github/repos_deployments.go
```diff
@@ -137,7 +137,7 @@ type DeploymentStatus struct {
 	Description   *string    `json:"description,omitempty"`
 	TargetURL     *string    `json:"target_url,omitempty"`
 	CreatedAt     *Timestamp `json:"created_at,omitempty"`
-	UpdatedAt     *Timestamp `json:"pushed_at,omitempty"`
+	PushedAt      *Timestamp `json:"pushed_at,omitempty"`
```
shurcooL: Now that I've looked, are we sure `pushed_at` is the right JSON field?

It doesn't look to be. I think it might be `updated_at` instead.

Take a look at https://developer.github.com/v3/repos/deployments/. There are 7 hits for "updated_at", 0 for "pushed_at".

e-beach:
@shurcooL This should be good-to-go now.

shurcooL:
Thanks again @e-beach.

Issue #707, opened 2017-08-23
Include more advanced search query in godoc and test case.


shurcooL:
From #701, there are 2 improvements I think we can make for our search endpoints:

- [ ] The godoc can be improved to give an example query. Right now, https://godoc.org/github.com/google/go-github/github#SearchService.Issues doesn't say much. I think it'd be helpful to give an example query that can be done. For example, maybe a query like this:

    `"gopher is:issue label:bug language:go"`
    with
    `opt := &github.SearchOptions{Sort: "created", Order: "asc"}`

    ```Go
    result, _, err := cl.Search.Issues(ctx, "gopher is:issue label:bug language:go", opt)
    ```

- [ ] We should add another test case for `search` that covers a query with spaces and colons. That way, the query parameter escaping functionality won't accidentally regress in the future.

See https://github.com/google/go-github/pull/701#issuecomment-324362880 for more context.

e-beach:
Can this be assigned to me?

PR #711, opened 2017-09-01
Document search api

e-beach:
This PR addresses #707 by adding some example documentation on search queries and a unit test for what GitHub terms "search qualifiers", e.g., `"is:issue"`.

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -15,12 +15,13 @@ import (
 // SearchService provides access to the search related functions
 // in the GitHub API.
 //
-// Each method takes a query string defining the search keywords and any search qualifiers. For example,
-// when searching issues, the query "gopher is:issue language:go" will search for issues containing the word "gopher"
-// in go repositories. The method call
+// Each method takes a query string defining the search keywords and any search qualifiers.
+// For example, when searching issues, the query "gopher is:issue language:go" will search
+// for issues containing the word "gopher" in Go repositories. The method call
 //   opts :=  &github.SearchOptions{Sort: "created", Order: "asc"}
-//   cl.Search.Issues(ctx, "gopher is:issue label:bug language:go", opts)
-// will search for such issues, sorting by creation date in ascending order (i.e., oldest first).
+//   cl.Search.Issues(ctx, "gopher is:issue language:go", opts)
+// will search for such issues, sorting by creation date in ascending order
+// (i.e., oldest first).
 //
 // GitHub API docs: https://developer.github.com/v3/search/
 type SearchService service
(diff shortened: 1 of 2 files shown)


e-beach:
Apologies for the flurry of commits. I think this is good now.

github/search.go
```diff
@@ -15,6 +15,13 @@ import (
 // SearchService provides access to the search related functions
 // in the GitHub API.
 //
+// Each method takes a query string defining the search keywords and any search qualifiers. For example,
+// when searching issues, the query "gopher is:issue language:go" will search for issues containing the word "gopher"
```
shurcooL: Mind wrapping these lines at around 90 characters for better readability? It's not related to terminal widths, just that really long lines are harder to read/scan with eyes.

github/search.go
```diff
@@ -15,6 +15,13 @@ import (
...
+// in go repositories. The method call
```
shurcooL: s/in go repositories/in Go repositories/ since the language is called "Go", not "go".

github/search.go
shurcooL: Also, since the code below does `cl.Search.Issues(ctx, "gopher is:issue label:bug language:go", opts)`, I think you should use the same search query in the text here. Otherwise it might be confusing. It's currently missing `label:bug`. Either add it here, or remove it in the code below.

github/search.go
```diff
@@ -15,6 +15,13 @@ import (
...
+//   opts :=  &github.SearchOptions{Sort: "created", Order: "asc"}
+//   cl.Search.Issues(ctx, "gopher is:issue label:bug language:go", opts)
+// will search for such issues, sorting by creation date in ascending order (i.e., oldest first).
+//
```
shurcooL: I understand the idea of placing this doc section under `SearchService`, it applies to all its methods. But I worry it might be harder to see/find if it's here.

It might be better to just move it to the `Issues` or `Repositories` methods. I feel people are more likely to spot it then.

No need to have it everyone, just the most prominent method is fine.

What do you think?

e-beach: It's true that searching issues is the most common use case, but I don't see how `Issues` is the most prominent method from either the godoc or the source code itself, as it is in the middle of the module's methods. An engineer will still have to scan all the methods of `SearchService` for the example section if they start searching by code or repositories. 

It might make more sense to me if we could choose the order of methods in godoc, but the godoc alphabetical ordering makes it hard to mark a particular method as more important.

shurcooL: They show up almost entirely [on one page](https://user-images.githubusercontent.com/1924134/30095214-2d004e78-929f-11e7-9e13-42f6c8676b6b.png) for me.

All right, it's good as is. We can't guarantee everyone will see it. We just want to have it somewhere so that the people who do look can find it.

e-beach:
@shurcooL ping as Travis has completed

PR #716, opened 2017-09-09
(#715) Add test for anomalous timestamp format from Projects API.

### LINKED ISSUE #715: Support breaking changes to the Projects API preview
GitHub Announcement:
https://developer.github.com/changes/2017-09-01-breaking-changes-to-projects-api-preview-and-webhooks-date-format/

To resolve this issue, it appears that unit tests will need to be updated for the time parsing functions that process webhook payloads. It is possible that the new format is already supported, but it is more likely that the parsing format needs to be changed slightly.

I personally think that it might be nice to support both formats (old and new) in the code base with a "TODO" to remove support for the old format once it is no longer being sent by GitHub, but this is completely up for discussion and I don't have my heart set on this. (One nice serendipity of this would be that no breaking changes would be made to the API.)

Thank you in advance for contributing to this open source project!
Your assistance is greatly appreciated.

e-beach:
This PR adds test logic verifying that timestamp strings in the `"2006-01-02T15:04:05.000Z"`  format are converted properly into `Timestamps` by JSON unmarshalling.
Fixes #715.

### DIFF
--- a/github/timestamp_test.go
+++ b/github/timestamp_test.go
@@ -15,6 +15,7 @@ import (
 const (
 	emptyTimeStr         = `"0001-01-01T00:00:00Z"`
 	referenceTimeStr     = `"2006-01-02T15:04:05Z"`
+	referenceTimeStrFractional = `"2006-01-02T15:04:05.000Z"` // This format was returned by the Projects API before October 1, 2017.
 	referenceUnixTimeStr = `1136214245`
 )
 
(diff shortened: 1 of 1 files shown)


e-beach:
@shurcooL Any word on this? :grinning:

github/timestamp_test.go
```diff
@@ -13,9 +13,10 @@ import (
 )
 
 const (
-	emptyTimeStr         = `"0001-01-01T00:00:00Z"`
-	referenceTimeStr     = `"2006-01-02T15:04:05Z"`
-	referenceUnixTimeStr = `1136214245`
+	emptyTimeStr               = `"0001-01-01T00:00:00Z"`
+	referenceTimeStr           = `"2006-01-02T15:04:05Z"`
+	referenceTimeStrFractional = `"2006-01-02T15:04:05.000Z"` // This format was returned by the Projects API before October 1, 2017.
```
shurcooL: GitHub announcement says:

> Starting on Monday October 2nd, 2017

So, wouldn't it be more accurate to say this?

```diff
-before October 1, 2017.
+before October 2, 2017.
```

At least I parse "before day X" to mean "before and not including day X" rather than "before and also on day X".

But this is extremely insignificant.

Issue #714, opened 2017-09-07
Support preview Nested Teams API


gmlewis:
GitHub Announcement:
https://developer.github.com/changes/2017-08-30-preview-nested-teams/

This involves adding new endpoints and changing existing endpoints according to the blog post.

Successful completion and closing of this issue involves addressing all the changes described in the GitHub announcement.  Note that it doesn't have to all be done in a single PR.  It is always a good idea to carefully document what portions are being addressed in any given PR and if there is still work yet to be done.

Thank you in advance for contributing to this open source project!
Your assistance is greatly appreciated.

e-beach:
I'd like to start working on this.
Currently I'm assigned to #715 but that is quite an insubstantial change.

shurcooL:
Awesome, thanks @e-beach!

shurcooL:
As Glenn said, this doesn't have to be resolved in a single PR if that would make it large. If so, it'd be better to break it into a few smaller independent changes, as long as that's possible and doing so makes sense.

e-beach:
I will be creating a `NewTeam` struct  for the params passed to `CreateTeam` and `EditTeam`, which now differ from the `Team` response because of the `parent`(in response) and `parent_team_id`(in request) fields. https://developer.github.com/v3/orgs/teams/#parameters

I also noticed that the `repo_names` parameter to `CreateTeam` is not supported by `go-github` currently, so it will also go in that struct.

alindeman:
@e-beach Is there anything left to implement? I'm wondering when this branch can be [merged to master](https://github.com/google/go-github/pull/718#pullrequestreview-64129775), and I'm glad to jump in to help if needed.

e-beach:
@alindeman Thanks for the ping. Yes, #741 finished up support for the API, and this can be merged.
/cc @shurcooL @gmlewis .

gmlewis:
OK, merging [NestedTeamsPreview branch](https://github.com/google/go-github/tree/NestedTeamsPreview) to master.

gmlewis:
Merging got sidetracked - checking on status with @elliott-beach.

Meanwhile, there is a new GitHub API announcement:
https://developer.github.com/changes/2017-10-13-update-the-nested-teams-preview/

This means that around line 34 of `orgs_teams.go`:

```go
// Default is "secret".
```

should be changed to:

```go
// Default is "closed".
```

This documentation change can happen in a separate PR.

PR #718, opened 2017-09-10
Request parent team from endpoints that return a team

e-beach:
This PR address part of #714 by
1. Adding a `parent` field to the `Team` struct.
2. Requesting this field to be populated from each endpoint that returns a team, by adding the API preview header to the request.

Initially, I had added the preview header to the `ListTeamMembers`, `IsTeamMember`, and `GetTeamMembership` endpoints, but dropped those changes because they aren't relevant to the goal of this PR.

### DIFF
--- a/github/orgs_teams_test.go
+++ b/github/orgs_teams_test.go
@@ -48,7 +48,8 @@ func TestOrganizationsService_GetTeam(t *testing.T) {
 
 	mux.HandleFunc("/teams/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		fmt.Fprint(w, `{"id":1, "name":"n", "description": "d", "url":"u", "slug": "s", "permission":"p", "ldap_dn":"cn=n,ou=groups,dc=example,dc=com"}`)
+		fmt.Fprint(w, `{"id":1, "name":"n", "description": "d", "url":"u", "slug": "s", "permission":"p", "ldap_dn":"cn=n,ou=groups,dc=example,dc=com", "parent":null}`)
+
 	})
 
 	team, _, err := client.Organizations.GetTeam(context.Background(), 1)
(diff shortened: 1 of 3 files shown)


github/orgs_teams_test.go
```diff
@@ -62,6 +64,31 @@ func TestOrganizationsService_GetTeam(t *testing.T) {
...
+func TestOrganizationService_GetTeam_nestedTeams(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/teams/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeNestedTeamsPreview)
+		fmt.Fprint(w, `{"id":1, "name":"n", "description": "d", "url":"u", "slug": "s", "permission":"p",
+		"parent": {"id":2, "name":"n", "description": "d", "parent": null}}`)
+
```
shurcooL: Minor, unneeded blank line here.

e-beach:
Moving the changes to a separate branch sounds logical to me - it doesn't make sense to add potentially breaking changes piecemeal. The impact of a change such as listing members of child teams, instead of just the parent team, should be handled carefully.

I guess we are awaiting @gmlewis's opinion, then.

I guess we are awaiting @gmlewis's opinion, then.

gmlewis:
A separate feature branch seems totally reasonable to me for this use case. I say "go for it".
Thanks!

shurcooL:
Okay.

I've created a [`NestedTeamsPreview`](https://github.com/google/go-github/tree/NestedTeamsPreview) branch. I'll close this PR, @e-beach please re-send it against that brach rather than `master`. Feel free to address the one minor review comment I posted (extraneous blank line). Thanks!

e-beach:
I just discovered that it is possible to edit a PR to be against a different branch.
See: https://stackoverflow.com/a/24159161/5749914
Oh well, I sent a new one.

shurcooL:
Oh, my bad, I forgot that it's possible now. All right.

PR #723, opened 2017-09-22
Request parent team when endpoints return a team

e-beach:
Re-submit of #718.

This PR address part of #714 by
1. Adding a parent field to the Team struct.
2. Requesting this field to be populated from each endpoint that returns a team, by adding the API preview header to the request.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -102,6 +102,9 @@ const (
 
 	// https://developer.github.com/changes/2017-07-26-team-review-request-thor-preview/
 	mediaTypeTeamReviewPreview = "application/vnd.github.thor-preview+json"
+
+	// https://developer.github.com/changes/2017-08-30-preview-nested-teams/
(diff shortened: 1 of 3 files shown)


e-beach:
As @shurcooL mentioned, another side effect of this PR is stated in the docs:

> Endpoints that return team hashes will now include the parent field representing the parent team. Additionally, the members_count and repositories_count will include all members and repositories according to the nested structure.

Those fields are a part of the Team struct. The endpoints that have the application/vnd.github.hellcat-preview+json header set will return different numbers than other endpoints where application/vnd.github.hellcat-preview+json isn't yet set.

shurcooL:
@e-beach, this PR is currently targeting `master` branch. Was that a mistake? Can you update it to target `NestedTeamsPreview` instead?

e-beach:
@shurcooL Fixed.

PR #724, opened 2017-09-23
Add parent team id param

e-beach:
This replaces `Team` with `NewTeam` in function that accept a team for creating or editing a team.
The `NewTeam` struct is similar to the existing `NewPullRequest` for creating pull requests.

We need a new struct because the set of parameters for creating or editing a team contains some parameters that are not in the parameter for teams, and most properties of teams cannot be used for editing.

The specific change is that Teams are returned with a `parent` field, but are created/edited with a `parent_id` field.

### DIFF
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -114,10 +114,39 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
 	return t, resp, nil
 }
 
+// NewTeam represents a team to be created or modified.
+type NewTeam struct {
+	ID           *int      `json:"id,omitempty"`
+	Name         *string   `json:"name,omitempty"`
+	Description  *string   `json:"description,omitempty"`
+	Maintainers  []*string `json:"maintainers,omitempty"`
+	RepoNames    []*string `json:"repo_names,omitempty"`
+	ParentTeamID *string   `json:"parent_team_id,omitempty"`
+
+	// Permission is deprecated when creating or editing a team in an org
+	// using the new GitHub permission model. It no longer identifies the
+	// permission a team has on its repos, but only specifies the default
+	// permission a repo is initially added with. Avoid confusion by
+	// specifying a permission value when calling AddTeamRepo.
+	Permission *string `json:"permission,omitempty"`
+
+	// Privacy identifies the level of privacy this team should have.
+	// Possible values are:
+	//     secret - only visible to organization owners and members of this team
+	//     closed - visible to all members of this organization
+	// Default is "secret".
+	Privacy *string `json:"privacy,omitempty"`
+
+	// TODO: decide whether to delete
+	// LDAPDN is only available in GitHub Enterprise and when the team
+	// membership is synchronized with LDAP.
+	LDAPDN *string `json:"ldap_dn,omitempty"`
+}
+
 // CreateTeam creates a new team within an organization.
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/teams/#create-team
-func (s *OrganizationsService) CreateTeam(ctx context.Context, org string, team *Team) (*Team, *Response, error) {
+func (s *OrganizationsService) CreateTeam(ctx context.Context, org string, parent_team_id int, team *Team) (*Team, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/teams", org)
 	req, err := s.client.NewRequest("POST", u, team)
 	if err != nil {
@@ -139,7 +168,7 @@ func (s *OrganizationsService) CreateTeam(ctx context.Context, org string, team 
 // EditTeam edits a team.
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/teams/#edit-team
-func (s *OrganizationsService) EditTeam(ctx context.Context, id int, team *Team) (*Team, *Response, error) {
+func (s *OrganizationsService) EditTeam(ctx context.Context, id, parent_team_id int, team *Team) (*Team, *Response, error) {
 	u := fmt.Sprintf("teams/%v", id)
 	req, err := s.client.NewRequest("PATCH", u, team)
 	if err != nil {
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -136,11 +136,6 @@ type NewTeam struct {
 	//     closed - visible to all members of this organization
 	// Default is "secret".
 	Privacy *string `json:"privacy,omitempty"`
-
-	// TODO: decide whether to delete
-	// LDAPDN is only available in GitHub Enterprise and when the team
-	// membership is synchronized with LDAP.
-	LDAPDN *string `json:"ldap_dn,omitempty"`
 }
 
 // CreateTeam creates a new team within an organization.
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -119,8 +119,8 @@ type NewTeam struct {
 	ID           *int      `json:"id,omitempty"`
 	Name         *string   `json:"name,omitempty"`
 	Description  *string   `json:"description,omitempty"`
-	Maintainers  []*string `json:"maintainers,omitempty"`
-	RepoNames    []*string `json:"repo_names,omitempty"`
+	Maintainers  []string `json:"maintainers,omitempty"`
+	RepoNames    []string `json:"repo_names,omitempty"`
 	ParentTeamID *string   `json:"parent_team_id,omitempty"`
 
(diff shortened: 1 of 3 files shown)


github/orgs_teams.go
```diff
@@ -114,10 +114,34 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
 	return t, resp, nil
 }
 
+// NewTeam represents a team to be created or modified.
+type NewTeam struct {
+	ID           *int      `json:"id,omitempty"`
+	Name         *string   `json:"name,omitempty"`
+	Description  *string   `json:"description,omitempty"`
+	Maintainers  []*string `json:"maintainers,omitempty"`
+	RepoNames    []*string `json:"repo_names,omitempty"`
```
shurcooL: These two fields can be simplified to `[]string`.

You can look at other structures for example. We use pointers to basic types so it’s possible to tell if the field was missing or zero. Slices are already pointer types and can be nil if missing.

github/orgs_teams.go
```diff
@@ -114,10 +114,34 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
 	return t, resp, nil
 }
 
+// NewTeam represents a team to be created or modified.
+type NewTeam struct {
+	ID           *int      `json:"id,omitempty"`
+	Name         *string   `json:"name,omitempty"`
+	Description  *string   `json:"description,omitempty"`
```
shurcooL: Are any of these fields required rather than optional?

If so, since this struct is used for providing input only, we can use values for mandatory fields.

github/orgs_teams_test.go
```diff
@@ -73,6 +73,7 @@ func TestOrganizationService_GetTeam_nestedTeams(t *testing.T) {
 		testHeader(t, r, "Accept", mediaTypeNestedTeamsPreview)
 		fmt.Fprint(w, `{"id":1, "name":"n", "description": "d", "url":"u", "slug": "s", "permission":"p",
 		"parent": {"id":2, "name":"n", "description": "d", "parent": null}}`)
+
```
shurcooL: Was this line added by accident?

shurcooL:
Thanks for providing the rationale for `NewTeam` struct, it sounds justified. 👍

github/orgs_teams.go
e-beach: Yes, the `ID` field is required. What you are suggesting makes sense.

shurcooL: Great.

If you want to see some examples, see [`RequiredStatusChecks`](https://godoc.org/github.com/google/go-github/github#RequiredStatusChecks) and grep for "Required" and "Optional" in comments.

alindeman:
Related: #724 

Looking forward to having this. I will plan to use it to implement `parent_team_id` in the Terraform GitHub provider unless someone else beats me to it :)

e-beach:
I noticed and fixed a mistake I made - `Name` is the required parameter, while ID should not even be a field of `NewTeam`.

github/orgs_teams.go
```diff
@@ -139,7 +164,7 @@ func (s *OrganizationsService) CreateTeam(ctx context.Context, org string, team
 // EditTeam edits a team.
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/teams/#edit-team
-func (s *OrganizationsService) EditTeam(ctx context.Context, id int, team *Team) (*Team, *Response, error) {
+func (s *OrganizationsService) EditTeam(ctx context.Context, id int, team *NewTeam) (*Team, *Response, error) {
```
e-beach: It's odd that the `EditTeam` endpoint requires both a name and an id for the team, but that is the documented behavior.

github/orgs_teams.go
```diff
@@ -136,11 +136,6 @@ type NewTeam struct {
 	//     closed - visible to all members of this organization
 	// Default is "secret".
 	Privacy *string `json:"privacy,omitempty"`
-
-	// TODO: decide whether to delete
-	// LDAPDN is only available in GitHub Enterprise and when the team
-	// membership is synchronized with LDAP.
-	LDAPDN *string `json:"ldap_dn,omitempty"`
```
alindeman: @e-beach I don't think `LDAPDN` should be deleted. On GitHub Enterprise, it's allowed to add or update a team with this parameter. We use it in the Terraform GitHub provider too.

e-beach: Thanks for that feedback! I'll add it back in today, whoops.

alindeman: Thanks @e-beach!

github/orgs_teams.go
```diff
@@ -114,10 +114,39 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
 	return t, resp, nil
 }
 
+// NewTeam represents a team to be created or modified.
+type NewTeam struct {
+
+	// The name of the team (Required.)
+	Name         string   `json:"name"`
```
shurcooL: Since there's no blank line after `Name` field, and `Description` doesn't have a comment of its own, this reads as if the comment above is for the whole block. You can make it inline to make it more readable. Also, maybe drop the "The" to shorten it.

```Go
Name         string   `json:"name"` // Name of the team. (Required.)
Description  *string  `json:"description,omitempty"`
...
```

github/orgs_teams.go
```diff
@@ -114,10 +114,39 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
 	return t, resp, nil
 }
 
+// NewTeam represents a team to be created or modified.
+type NewTeam struct {
+
```
shurcooL: Is this blank line helpful? If not, let's remove it.

alindeman:
LGTM, but 511f540 being in the commit list is a bit odd. Maybe rebased from the wrong spot?

e-beach:
Yes. Sorry - I was occupied and just got to trying to fix this. I rebased from master, but I think that is actually unnecessary. Not sure why there are conflicting files.

e-beach:
This should be good now.

I rebased off of master, realized that was unnecessary and caused conflicts, so rebased off of NestedTeamsPreview and pushed again.

@shurcooL your style suggestions should be addressed in 94ed0c7.

github/orgs_teams.go
```diff
@@ -114,10 +114,37 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
...
+	Description  *string  `json:"description,omitempty"`
+	Maintainers  []string `json:"maintainers,omitempty"`
+	RepoNames    []string `json:"repo_names,omitempty"`
+	ParentTeamID *string  `json:"parent_team_id,omitempty"`
+
+	// Permission is deprecated when creating or editing a team in an org
+	// using the new GitHub permission model. It no longer identifies the
+	// permission a team has on its repos, but only specifies the default
+	// permission a repo is initially added with. Avoid confusion by
+	// specifying a permission value when calling AddTeamRepo.
```
shurcooL: @e-beach Are you familiar with the [convention Go uses for deprecated symbols](https://github.com/golang/blog/commit/257114af91a0defb1fc2c16c7f4ae2429b8a4e0f)?

I think it would be nice to use here.

github/orgs_teams.go
```diff
@@ -114,10 +114,37 @@ func (s *OrganizationsService) GetTeam(ctx context.Context, team int) (*Team, *R
 	return t, resp, nil
 }
 
+// NewTeam represents a team to be created or modified.
+type NewTeam struct {
+	Name         string   `json:"name"` // Name of the team (Required.)
```
shurcooL: Nitpicky, but I'd add a period at the end of the sentence. "Required" is another sentence inside parenthesis.

```Go
// Name of the team. (Required.)
```

shurcooL: This how it's written elsewhere. See:

https://github.com/google/go-github/blob/511f540f1887d30b88cee4a2fcd1f2922754acf4/github/repos.go#L552

github/orgs_teams.go
e-beach: That is copied from `Team` so I'll fix it there as well:
https://github.com/google/go-github/blob/511f540f1887d30b88cee4a2fcd1f2922754acf4/github/orgs_teams.go#L23-L28

e-beach:
@shurcooL I believe I addressed your review.

shurcooL:
Thanks! Those adjustments look great. My LGTM from before still applies. Merging this to `NestedTeamsPreview` branch.

Issue #179, opened 2015-04-08
Add support for Enterprise interfaces


moutons:
in the [godocs](https://godoc.org/github.com/google/go-github/github) for go-github there are a few references to using it with GitHub Enterprise instances, but I don't see reference to the management tasks other than [PromoteSiteAdmin](https://github.com/google/go-github/blob/master/github/users_administration.go#L13). It would be nice to be able to use go-github to work with some interfaces like /enterprise/settings/license or /enterprise/stats/

gmlewis:
6 years later, it looks like `/enterprise/settings/...` endpoints are still missing, but closing issue for lack of interest/activity.
Feel free to reopen if anyone is still interested.

BlackVegetable:
I'd like this! Can I re-open.

naikrovek:
I would also like to see the enterprise settings endpoints become supported.  I was very surprised when I discovered that they weren't.

marcellodesales:
Wow, no Github Enterprise support?

gmlewis:
> Wow, no Github Enterprise support?

Have you looked at the repo or the docs?
I don't think "no GitHub Enterprise support" is quite accurate, as I have personally used this repo extensively within GitHub Enterprise environments to great success.
It might be limited in some aspects such as those listed above, but PRs are welcome.

marcellodesales:
@gmlewis my bad... I just found `github.NewEnterpriseClient` ... I think this should be closed!

gmlewis:
OK, good point.
Thank you, @marcellodesales .
Closing issue as obsolete.

naikrovek:
feel free to leave this closed, i just wanted to say that while enterprise API endpoints are supported, support for all of them is not present, and I would like to see that.  I'm not going to demand that, as I have no right to, and I don't expect you to provide me anything that you have not already provided, just so I am being clear. 

*I am thankful for this repo and what it has allowed me to do.*

I would also like to add that there is a response above that is probably almost never intended to be rude or dismissive, but in effect, it _always_ is:

> PRs are welcome

You don't need to say that.  PRs are always welcome, it should be assumed, unless there is some declaration that they are not welcomed.  Saying this equates to "do it yourself if you want it so bad" and that is rude and dismissive.  So saying this is not only unneeded, it can potentially turn a willing contributor into a non-contributor.

[...]

PR #725, opened 2017-09-25
Add admin stats

defsprite:
This adds support for "Admin Stats" part of the GitHub Enterprise API, as per https://developer.github.com/v3/enterprise-admin/admin_stats. It should implement part of #179 

At this point, only getting "all" stats is supported, thought it should be straightforward to add  others at some point. 

I'm really not sure about the naming here, the api prefixes for the GHE Admin API are all over the place with `/admin`, `/staff`, `/setup` and `/enterprise`. I see three options here:

- Stick to the pattern (In that case I'd need to move this implementation to `EnterpriseService`)
- Collect all of them in the existing `AdminService` 
- Collect all of them in a newly created `EnterpriseAdminService` (migration & deprecation fun)


defsprite:
I signed it!

github/admin_stats.go
```diff
@@ -0,0 +1,99 @@
...
+type RepoStats struct {
+	TotalRepos  *int `json:"total_repos,omitempty"`
+	RootRepos   *int `json:"root_repos,omitempty"`
+	ForkRepos   *int `json:"fork_repos,omitempty"`
+	OrgRepos    *int `json:"org_repos,omitempty"`
+	TotalPushes *int `json:"total_pushes,omitempty"`
+	TotalWikis  *int `json:"total_wikis,omitempty"`
+}
+
+func (s *AdminService) GetAdminStats(ctx context.Context) (*AdminStats, *Response, error) {
```
e-beach: I anticipate the maintainers will want you to add doc-sentences for this function and the structs, as in https://github.com/google/go-github/blob/61bf8a6a2009b7fb88173ed454c77fb8013b8852/github/licenses.go#L61-L64.

shurcooL:
Thanks for the PR @defsprite.

As @e-beach mentioned, we'll definitely want the exported top-level identifiers to have documentation. See the other .go files for precedent and preferred format.

> I'm really not sure about the naming here, the api prefixes for the GHE Admin API are all over the place with `/admin`, `/staff`, `/setup` and `/enterprise`. I see three options here:
>
> - Stick to the pattern (In that case I'd need to move this implementation to `EnterpriseService`)
> - Collect all of them in the existing `AdminService` 
> - Collect all of them in a newly created `EnterpriseAdminService` (migration & deprecation fun)

Good question.

I think in this case there's an easy answer. Since we already have an existing `AdminService`, this feels like it would fit nicely there.

If there was no equivalent non-GHE service for a GHE endpoint, we might want to create an `EnterpriseService` and put all the stragglers there. I definitely don't think we ever need to go down the `EnterpriseAdminService`, there's just not that many GHE-only endpoints to justify it.

Thanks!

shurcooL:
Thanks, looks good.

This is relatively minor, but would be nice to fix: many of the sentences of the documentations do not end with a period. They should, to be consistent with the rest of our documentation. Our goal is to use correct grammar, spelling, and have complete sentences with punctuation.

github/admin_stats_test.go
```diff
@@ -0,0 +1,142 @@
...
+	})
+
+	stats, _, err := client.Admin.GetAdminStats(context.Background())
+	if err != nil {
+		t.Errorf("AdminService.GetAdminStats returned error: %v", err)
+	}
+
+	want := &AdminStats{
+		Repos: &RepoStats{
+			Int(212),
```
gmlewis: I'm not sure we need to test the value of every field, indeed we do not do elsewhere in this package if I recall correctly, as it becomes more challenging to maintain as fields are added.

However, I'm not adverse to it either if you explicitly use the field names here in the test and not rely solely on ordering. That way, a new field can more easily be added, or even if it is not, the unit test would not suddenly break.

defsprite: Absolutely, relying on order is super brittle!

defsprite:
All done now. 

Please re-check @gmlewis

PR #726, opened 2017-09-25
GitHub requires 'reviewers' and 'team_reviewers' to be sent, even if empty

alindeman:
Otherwise an error is returned. Extracted from our logs:

```
level=error msg="POST https://X/api/v3/repos/X/Y/pulls/2234/requested_reviewers: 422 Invalid request.\n\n\"reviewers\" wasn't supplied. []" 
```

### DIFF
--- a/github/pulls_reviewers.go
+++ b/github/pulls_reviewers.go
@@ -12,7 +12,7 @@ import (
 
 // ReviewersRequest specifies users and teams for a pull request review request.
 type ReviewersRequest struct {
-	Reviewers     []string `json:"reviewers,omitempty"`
+	Reviewers     []string `json:"reviewers"`
 	TeamReviewers []string `json:"team_reviewers,omitempty"`
 }
 
--- a/github/pulls_reviewers_test.go
+++ b/github/pulls_reviewers_test.go
@@ -35,6 +35,28 @@ func TestRequestReviewers(t *testing.T) {
 	}
 }
 
+func TestRequestReviewers_withEmptyReviewers(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/pulls/1/requested_reviewers", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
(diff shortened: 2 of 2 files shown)


github/pulls_reviewers.go
```diff
@@ -12,8 +12,8 @@ import (
 
 // ReviewersRequest specifies users and teams for a pull request review request.
 type ReviewersRequest struct {
-	Reviewers     []string `json:"reviewers,omitempty"`
-	TeamReviewers []string `json:"team_reviewers,omitempty"`
```
e-beach: The error message indicates that `Reviewers` is required. Is it also the case that `TeamReviewers` is required?
My intuition would be `TeamReviewers` would not be required because it's a new parameter that most requests won't have.

shurcooL:
I think we should try to match GitHub documentation, and if there's a discrepancy, report it to GitHub so they either update documentation, or fix a bug on their side. Otherwise, we're going to diverge from documented behavior and have a library that works unpredictably and is harder to develop and maintain.

Currently, at https://developer.github.com/v3/pulls/review_requests/#create-a-review-request, neither `reviewers` nor `team_reviewers` are listed as required parameters. Contrast that with, for example, https://developer.github.com/v3/pulls/comments/#create-a-comment, where all 4 input fields are documented as required.

The `team_reviewers` field is a part of an API preview, which further means this could be indicative of an issue on GitHub API side.

That's why I don't think we should try to work around it on our side, but rather contact GitHub and let them know about the discrepancy in behavior.

Also see this warning:

image

alindeman:
Thanks @e-beach @shurcooL. On further investigation, `team_reviewers` doesn't appear to be required, so I've updated my diff to reflect that. I've also contacted GitHub via their support email address to see about updating the documentation. I'll let you know if they get back to me.

alindeman:
Got a response from GitHub :tada:

> Hey Andy,
>
> The team talked about this and they tweaked the behavior so that neither parameter is required (in the sense that it needs to be present in every request), but at least one needs to be provided (either reviewers or team_reviewers).
>
> Hope this helps. And if you notice anything else that looks strange -- let us know!
>
> Cheers,
> Ivan

Contacting GitHub was definitely the right choice. Thanks for nudging me that way @shurcooL. No need for this code anymore.

PR #731, opened 2017-09-27
Change ListPendingOrgInvitations method to take string org name.

### LINKED ISSUE #730: OrganizationsService.ListPendingOrgInvitations does not work
The argument of `ListPendingOrgInvitations` is to pass an integer as `org`, but this is a mistake and I think you should pass the string correctly.

shurcooL:
The `/orgs/:org` GitHub API endpoint expects the string name of the organization, not its integer ID (that would be `/organizations/:id` endpoint, which we're generally not using).

Change the org parameter type from int to string to make it possible to specify the organization via its name.

This method wasn't functional before.

Fixes #730.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	libraryVersion = "11"
+	libraryVersion = "12"
 	defaultBaseURL = "https://api.github.com/"
 	uploadBaseURL  = "https://uploads.github.com/"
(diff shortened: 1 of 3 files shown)


shurcooL:
Friendly ping @gmlewis, are you able to review this? Someone sent a duplicate PR #738.

gmlewis:
Sorry for the delay... reviewing...

Issue #720, opened 2017-09-20
Support Marketplace (preview) API.


shurcooL:
GitHub API Docs: https://developer.github.com/v3/apps/marketplace/.

Not seeing an announcement blog post mentioning this, but there are new API endpoints available, as documented at the URL above.

This issue is to track their implementation.

A sub-task here is supporting the `MarketplacePurchaseEvent` event, which is tracked in a separate issue #704.

lackerman:
Hi @shurcooL, I would love to get involved but I would be remiss if I didn't say that I'm very new to Golang. My understanding from the Contribution guide and the issue you raised is that this requires the implementation of the `marketplace_listing` endpoints in perhaps a `marketplace.go` file. Any suggestions or resources to get me underway would be greatly appreciated.

shurcooL:
Sounds great, thanks @lackerman!

The contribution guide should have most of the things covered. For a general idea of what it looks like, I suggest looking at a similar issue and the PR that it took to resolve it in the past. Those serve as great examples/references. For example, see issue #569 and PR #620.

You can start your PR by implementing just one endpoint and sending a PR for review first. We can give feedback, and once everything looks good, you can implement the remaining endpoints. That way you're less likely to spend time on something that needs to be changed.

> this requires the implementation of the `marketplace_listing` endpoints in perhaps a `apps_marketplace.go` file

A new file named `apps_marketplace.go` sounds like a great fit to me. 👍

lackerman:
@shurcooL thanks for the information. I've created a PR and would appreciate any feedback.

PR #732, opened 2017-09-28
#720 Initial commit for listing Plans and Plan Accounts

lackerman:
* Includes basic testing for Listing Plans and the Accounts linked to a Plan (including stubs)

**NOTE** This is currently awaiting scrutiny as it's my first contribution to the project. I would appreciate any feedback (good or bad) and the 👍  if it's ok to carry on with the rest of the implementation for #720

### DIFF
--- a/github/apps_marketplace.go
+++ b/github/apps_marketplace.go
@@ -23,8 +23,6 @@ type MarketplaceService struct {
 	//
 	// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
 	Stubbed bool
-
-	ListOptions
 }
 
 // MarketplacePlan represents a GitHub Apps Marketplace Listing Plan.
--- a/github/apps_marketplace.go
+++ b/github/apps_marketplace.go
@@ -32,10 +32,11 @@ type MarketplacePlan struct {
 
 // MarketplacePurchase represents a GitHub Apps Marketplace Purchase.
 type MarketplacePurchase struct {
-	BillingCycle    *string          `json:"billing_cycle,omitempty"`
-	NextBillingDate *string          `json:"next_billing_date,omitempty"`
-	UnitCount       *int             `json:"unit_count,omitempty"`
-	Plan            *MarketplacePlan `json:"plan,omitempty"`
+	BillingCycle           *string                 `json:"billing_cycle,omitempty"`
+	NextBillingDate        *string                 `json:"next_billing_date,omitempty"`
+	UnitCount              *int                    `json:"unit_count,omitempty"`
+	Plan                   *MarketplacePlan        `json:"plan,omitempty"`
+	MarketplacePlanAccount *MarketplacePlanAccount `json:"account,omitempty"`
 }
 
 // MarketplacePlanAccount represents a GitHub Account (user or organization) on a specific plan.
@@ -75,10 +76,10 @@ func (s *MarketplaceService) ListPlans(ctx context.Context, stubbed bool, opt *L
 	return i, resp, nil
 }
 
-// ListPlanAccounts lists all GitHub accounts (user or organization) on a specific plan.
+// ListPlanAccountsForPlan lists all GitHub accounts (user or organization) on a specific plan.
 //
 // GitHub API docs: https://developer.github.com/v3/apps/marketplace/#list-all-github-accounts-user-or-organization-on-a-specific-plan
-func (s *MarketplaceService) ListPlanAccounts(ctx context.Context, planID int, stubbed bool, opt *ListOptions) ([]*MarketplacePlanAccount, *Response, error) {
+func (s *MarketplaceService) ListPlanAccountsForPlan(ctx context.Context, planID int, stubbed bool, opt *ListOptions) ([]*MarketplacePlanAccount, *Response, error) {
 	u, err := addOptions(fmt.Sprintf("%v/plans/%v/accounts", marketplaceURI(stubbed), planID), opt)
 	if err != nil {
 		return nil, nil, err
@@ -93,6 +94,58 @@ func (s *MarketplaceService) ListPlanAccounts(ctx context.Context, planID int, s
 	req.Header.Set("Accept", mediaTypeMarketplacePreview)
 
 	var i []*MarketplacePlanAccount
+	resp, err := s.client.Do(ctx, req, &i)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return i, resp, nil
+}
+
+// ListPlanAccountsForAccount lists all GitHub accounts (user or organization) associated with an account.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/marketplace/#check-if-a-github-account-is-associated-with-any-marketplace-listing
+func (s *MarketplaceService) ListPlanAccountsForAccount(ctx context.Context, accountID int, stubbed bool, opt *ListOptions) ([]*MarketplacePlanAccount, *Response, error) {
+	u, err := addOptions(fmt.Sprintf("%v/accounts/%v", marketplaceURI(stubbed), accountID), opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMarketplacePreview)
+
+	var i []*MarketplacePlanAccount
+	resp, err := s.client.Do(ctx, req, &i)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return i, resp, nil
+}
+
+// ListMarketplacePurchasesForUser lists all GitHub marketplace purchases made by a user.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/marketplace/#get-a-users-marketplace-purchases
+func (s *MarketplaceService) ListMarketplacePurchasesForUser(ctx context.Context, stubbed bool, opt *ListOptions) ([]*MarketplacePurchase, *Response, error) {
+	u, err := addOptions("apps/user/marketplace_purchases", opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMarketplacePreview)
+
+	var i []*MarketplacePurchase
 	resp, err := s.client.Do(ctx, req, &i)
 	if err != nil {
 		return nil, resp, err
--- a/github/apps_marketplace.go
+++ b/github/apps_marketplace.go
@@ -1,4 +1,4 @@
-// Copyright 2013 The go-github AUTHORS. All rights reserved.
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
@@ -20,8 +20,6 @@ type MarketplaceService struct {
 	// instead of production endpoints. Stubbed data is fake data that's useful
 	// for testing your GitHub Apps. Stubbed data is hard-coded and will not
 	// change based on actual subscriptions.
-	//
-	// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
 	Stubbed bool
 }
 
@@ -41,17 +39,17 @@ type MarketplacePlan struct {
 
 // MarketplacePurchase represents a GitHub Apps Marketplace Purchase.
 type MarketplacePurchase struct {
-	BillingCycle           *string                 `json:"billing_cycle,omitempty"`
-	NextBillingDate        *string                 `json:"next_billing_date,omitempty"`
-	UnitCount              *int                    `json:"unit_count,omitempty"`
-	Plan                   *MarketplacePlan        `json:"plan,omitempty"`
-	MarketplacePlanAccount *MarketplacePlanAccount `json:"account,omitempty"`
+	BillingCycle    *string                 `json:"billing_cycle,omitempty"`
+	NextBillingDate *string                 `json:"next_billing_date,omitempty"`
+	UnitCount       *int                    `json:"unit_count,omitempty"`
+	Plan            *MarketplacePlan        `json:"plan,omitempty"`
+	Account         *MarketplacePlanAccount `json:"account,omitempty"`
 }
 
 // MarketplacePlanAccount represents a GitHub Account (user or organization) on a specific plan.
 type MarketplacePlanAccount struct {
 	URL                      *string              `json:"url,omitempty"`
-	AccountType              *string              `json:"type,omitempty"`
+	Type                     *string              `json:"type,omitempty"`
 	ID                       *int                 `json:"id,omitempty"`
 	Login                    *string              `json:"login,omitempty"`
 	Email                    *string              `json:"email,omitempty"`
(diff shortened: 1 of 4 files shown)


github/apps_marketplace.go
```diff
@@ -0,0 +1,109 @@
...
+
+// MarketplacePurchase represents a GitHub Apps Marketplace Purchase.
+type MarketplacePurchase struct {
+	BillingCycle    *string          `json:"billing_cycle"`
+	NextBillingDate *string          `json:"next_billing_date"`
+	UnitCount       *int             `json:"unit_count"`
+	Plan            *MarketplacePlan `json:"plan"`
+}
+
+// MarketplacePlanAccount represents a GitHub Account (user or organization) on a specific plan
```
shurcooL: In general, I think we should strive to have all sentences in Go documentation to have no spelling or grammar mistakes, and to include periods at ends of sentences. So, I'll point out that you're missing a period at the end of this sentence.

github/apps_marketplace.go
```diff
@@ -0,0 +1,109 @@
...
+	URL                      *string              `json:"url"`
+	AccountType              *string              `json:"type"`
+	ID                       *int                 `json:"id"`
+	Login                    *string              `json:"login"`
+	Email                    *string              `json:"email"`
+	OrganizationBillingEmail *string              `json:"organization_billing_email"`
+	MarketplacePurchase      *MarketplacePurchase `json:"marketplace_purchase"`
+}
+
+// ListPlans lists all plans for your Marketplace listing
```
shurcooL: Also missing a period at end of this sentence.

github/apps_marketplace.go
```diff
@@ -0,0 +1,109 @@
...
+	var i []*MarketplacePlan
+	resp, err := s.client.Do(ctx, req, &i)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return i, resp, nil
+}
+
+// ListPlanAccounts lists all GitHub accounts (user or organization) on a specific plan
```
shurcooL: And this one.

github/apps_marketplace.go
```diff
@@ -0,0 +1,109 @@
...
+	URL                 *string   `json:"url"`
+	AccountsURL         *string   `json:"accounts_url"`
+	ID                  *int      `json:"id"`
+	Name                *string   `json:"name"`
+	Description         *string   `json:"description"`
+	MonthlyPriceInCents *int      `json:"monthly_price_in_cents"`
+	YearlyPriceInCents  *int      `json:"yearly_price_in_cents"`
+	PriceModel          *string   `json:"price_model"`
+	UnitName            *string   `json:"unit_name"`
+	Bullets             *[]string `json:"bullets"`
```
shurcooL: For consistency with other structs that we map to GitHub responses, all of these should have `,omitempty` JSON tag option. E.g., see:

https://github.com/google/go-github/blob/511f540f1887d30b88cee4a2fcd1f2922754acf4/github/issues.go#L26-L52

It doesn't make a difference for unmarshaling JSON, but it does for marshaling. I don't think these will be marshaled (since we're receiving data from GitHub API, not sending these upstream). So, the only reason to do it is to maintain consistency.

See issue #537 for some more background information on this.

lackerman:
I signed it!

github/apps_marketplace.go
lackerman: I was wondering why they were there. I didn't realise that they had no impact on unmarshalling (but it makes perfect sense). Happy to maintain the consistency.

lackerman:
@shurcooL Regarding
> ...I think we should factor out that stubbed bool argument from each method. Perhaps, instead, it should be a new option that we add to github.Client. What do you think about that?...

I'm so happy you said that. The flag felt like suck a hack. Not sure I have enough context or experience to comment on the preferred solution, but I'm happy to implement it ;).

lackerman:
Hi @shurcooL, could you perhaps tell me why the build is failing? I see it's during the `go generate` step, but I'm not sure how the code I've written has impacted.

e-beach:
@lackerman You need to run the command `go generate` and commit and push, this will generate some accessors for structs.

shurcooL:
Nice work on resolving the build failure.

> I'm so happy you said that. The flag felt like suck a hack. Not sure I have enough context or experience to comment on the preferred solution, but I'm happy to implement it ;).

Great! The next step here, as far as I can tell, is to factor out that `stubbed bool` argument from all the new methods.

Looking at it a bit closer, I think we can move it into `MarketplaceService` struct rather than into the [`github.Client`](https://godoc.org/github.com/google/go-github/github#Client) struct. Perhaps it can look something like this:

```Go
// MarketplaceService handles communication with the marketplace related
// methods of the GitHub API.
//
// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
type MarketplaceService struct {
	client *Client

	// Stubbed controls whether endpoints that return stubbed data are used
	// instead of production endpoints. Stubbed data is fake data that's useful
	// for testing your GitHub Apps. Stubbed data is hard-coded and will not
	// change based on actual subscriptions.
	//
	// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
	Stubbed bool
}
```

Then, users can set it to true if they wish:

```Go
client.MarketplaceService.Stubbed = true
// ...
```

lackerman:
Hi @shurcooL what about using a Options wrapper?

```go
// MarketplaceOptions specifies the optional parameters to the Marketplace service methods
type MarketplaceOptions struct {
	// Stubbed controls whether endpoints that return stubbed data are used
	// instead of production endpoints. Stubbed data is fake data that's useful
	// for testing your GitHub Apps. Stubbed data is hard-coded and will not
	// change based on actual subscriptions.
	//
	// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
	Stubbed bool

	ListOptions
}
```

And then only passing `ListOptions` to `addOptions`

```
uri := marketplaceURI(opt, "plans")
u, err := addOptions(uri, opt.ListOptions)
```

Or is this pattern specifically used for optional query parameters in the Url?

shurcooL:
`MarketplaceOptions` would work too. So it's just a design decision that's up to us to make.

I think it largely depends on how often one wants to set Stubbed to true. From my current understanding how it's meant to be used, it makes more sense to me to set it once for the entire service while doing testing, rather than for each individual endpoint via the options.

So, unless I'm not understanding the use case properly, I still prefer my `client.MarketplaceService.Stubbed = true` suggestion above.

@gmlewis @elliott-beach Do you have thoughts/preferences on how to deal with the stubbed parameter?

lackerman:
Hi @shurcooL, this is probably due to my own ignorance of the language, but when I try your suggested implementation, I get:
```
./github.go:231:39: cannot convert &c.common (type *service) to 
type *MarketplaceService
```
since the Service is of type `service` by default but now it's being declared as type `struct`.  Is this because I'm missing something that will allow it to _implement_ the service interface?

shurcooL:
@lackerman You wouldn't be able to reuse `&c.common` struct anymore, because `*MarketplaceService` will have an extra field. You'll have to create a new `*MarketplaceService` struct, like so:

```diff
-c.Marketplace = (*MarketplaceService)(&c.common)
+c.Marketplace = &MarketplaceService{client: c}
```

lackerman:
Thanks for the help @shurcooL. Please could I ask you to explain the difference between the following two calling conventions:
`(*MarketplaceService)(&c.common)` and `&MarketplaceService{client: c}`
The second one makes obvious sense but I'm not seeing how (*XXX)(...) creates a valid instance.

shurcooL:
@lackerman It works for other services because they're defined as `service`, e.g.:

```Go
type ActivityService service
```

So `*service` is convertible to `*ActivityService`. E.g., see [here](https://play.golang.org/p/mWxWK_RPhG).

You can find more information about this in #389 and #390 that introduced it. Note, this is a pretty advanced trick, it's not commonly done, and the benefits it offers are nice but not critical. It's a nice to have.

lackerman:
Ah, a colleague pointed out that it's just a type cast. Due to the syntax `(*type)(...)`, I thought it was some kind or function call/syntactic sugar for new or &.

shurcooL:
Yes, it is. Although if you want to be precise about terms, it's a "type conversion". See https://golang.org/ref/spec#Conversions:

> Conversions are expressions of the form T(x) where T is a type and x is an expression that can be converted to type T.
>
> If the type starts with the operator * or <-, or if the type starts with the keyword func and has no result list, it must be parenthesized when necessary to avoid ambiguity

Hence `(*T)(x)`.

lackerman:
Hi @shurcooL, I've completed the unit tests and performed a basic test to my account on one of the stubbed endpoints and it appears to work as expected.

shurcooL:
Good to hear!

I'm not seeing any new commits. Did you mean to push anything that implements the changes we discussed above?

lackerman:
@shurcooL, sad panda...yeah, I forgot to push. It's pushed now.

github/apps_marketplace.go
```diff
@@ -14,7 +14,18 @@ import (
...
+	client *Client
+	// Stubbed controls whether endpoints that return stubbed data are used
+	// instead of production endpoints. Stubbed data is fake data that's useful
+	// for testing your GitHub Apps. Stubbed data is hard-coded and will not
+	// change based on actual subscriptions.
+	//
+	// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
+	Stubbed bool
+
+	ListOptions
```
shurcooL: `ListOptions` probably doesn't belong here in a service struct, does it? I'm guessing you forgot to remove it from back when we considered a `MarketplaceOptions` wrapper, yeah?

lackerman: Yurp, pushed a fix.

lackerman:
@shurcooL fixed the issue you pointed out.

gmlewis:
Sorry for the delay - yes, I like the `MarketplaceService.Stubbed` solution that @shurcooL proposed in the [comment above](https://github.com/google/go-github/pull/732#issuecomment-333726743).
I'll now review the PR.

github/apps_marketplace.go
```diff
@@ -0,0 +1,180 @@
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMarketplacePreview)
+
+	var i []*MarketplacePlanAccount
```
gmlewis: s/i/accounts/
or "accts" if you prefer.

github/apps_marketplace.go
```diff
@@ -0,0 +1,180 @@
...
```
gmlewis: s/i/accounts/

github/apps_marketplace.go
```diff
@@ -0,0 +1,180 @@
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMarketplacePreview)
+
+	var i []*MarketplacePurchase
```
gmlewis: s/i/purchases/

github/apps_marketplace_test.go
```diff
@@ -0,0 +1,202 @@
+// Copyright 2013 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2013/2017/

github/apps_marketplace_test.go
```diff
@@ -0,0 +1,202 @@
...
+		testHeader(t, r, "Accept", mediaTypeMarketplacePreview)
+		testFormValues(t, r, values{
+			"page":     "1",
+			"per_page": "2",
+		})
+		fmt.Fprint(w, `[{"id":1}]`)
+	})
+
+	opt := &ListOptions{Page: 1, PerPage: 2}
+	client.Marketplace.Stubbed = false
```
gmlewis: Hmmm... I wish we didn't use a package global `client` for unit testing, as this will prevent unit tests from all being run in parallel in the future.

Ideally, we would have something like:

```go
client, teardown := setup()
defer teardown()
```

for each test.  No action is needed in this PR, but I'll file a new issue and point here for motivation.

github/github.go
```diff
@@ -143,6 +146,7 @@ type Client struct {
 	Licenses       *LicensesService
 	Migrations     *MigrationService
 	Reactions      *ReactionsService
+	Marketplace    *MarketplaceService
```
gmlewis: Let's also move this one two lines up to keep this list sorted.

lackerman:
Thanks @gmlewis for the extra pair of eyes, really appreciate the suggestions.

github/github.go
lackerman: Have no problem moving it up, but note that there are items higher in the list which are not sorted.

gmlewis: Would you mind sorting them while you are here, please?

lackerman: Done

github/apps_marketplace.go
```diff
@@ -162,13 +160,13 @@ func (s *MarketplaceService) ListMarketplacePurchasesForUser(ctx context.Context
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeMarketplacePreview)
 
-	var i []*MarketplacePurchase
-	resp, err := s.client.Do(ctx, req, &i)
+	var purchaces []*MarketplacePurchase
```
gmlewis: nit: s/purchaces/purchases/

github/github.go
```diff
@@ -136,16 +139,17 @@ type Client struct {
 	Gists          *GistsService
 	Git            *GitService
 	Gitignores     *GitignoresService
+	Licenses       *LicensesService
```
gmlewis: `Licenses` after `Issues`, please. :smile:

lackerman: Will it ever end... :P Done

github/apps_marketplace.go
```diff
@@ -0,0 +1,178 @@
...
+// methods of the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
+type MarketplaceService struct {
+	client *Client
+	// Stubbed controls whether endpoints that return stubbed data are used
+	// instead of production endpoints. Stubbed data is fake data that's useful
+	// for testing your GitHub Apps. Stubbed data is hard-coded and will not
+	// change based on actual subscriptions.
+	Stubbed bool
```
shurcooL: This `Stubbed` field is quite unusual and may cause people to seek more information about it. I think it's worth it to include a link to GitHub API documentation for it as well here, as I suggested in https://github.com/google/go-github/pull/732#issuecomment-333726743:

```diff
 // Stubbed controls whether endpoints that return stubbed data are used
 // instead of production endpoints. Stubbed data is fake data that's useful
 // for testing your GitHub Apps. Stubbed data is hard-coded and will not
 // change based on actual subscriptions.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
 Stubbed bool
```

github/apps_marketplace.go
```diff
@@ -0,0 +1,180 @@
...
+// methods of the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/marketplace/
+type MarketplaceService struct {
+	client *Client
+	// Stubbed controls whether endpoints that return stubbed data are used
+	// instead of production endpoints. Stubbed data is fake data that's useful
+	// for testing your GitHub Apps. Stubbed data is hard-coded and will not
+	// change based on actual subscriptions.
+	//
```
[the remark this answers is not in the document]
shurcooL: @gmlewis See https://github.com/google/go-github/pull/732/files#r148168072 for why I think it's still helpful to include them. It just so happens the URL matches the one for the service, I wish we could link to a more specific section, but I don't see one...

github/apps_marketplace.go
shurcooL: @gmlewis Does that sound good to you?

If so, I can apply it and merge.

lackerman:
@shurcooL, updated

lackerman:
@shurcooL not sure why the build failed after adding a comment line

shurcooL:
We have some issues with gen-accessors right now, I've filed #778 for it. But fixing that is out of scope of this PR, so just re-generate the package, and it should be fine.

lackerman:
@shurcooL ready to be merged

shurcooL:
Merged. Thank you @lackerman!

Issue #597, opened 2017-03-16
int64 IDs


bradfitz:
On 32-bit platforms, a Go `int` is an `int32`.

I worry about the use of `int` as the type for Github IDs in the go-github package.

Is there documentation to suggest that Github IDs are limited to 2 billion items? Seems unlikely. But I can't find anything.

I propose we keep "Number" fields as int, but change all the `ID` fields to be `int64`.

Thoughts?

/cc @gmlewis @willnorris @shurcooL

gmlewis:
I totally agree. Absolutely.

shurcooL:
I've noticed that too.

By now, there are definitely some IDs that exceed the capacity of `int32`. I have the following example. Consider the endpoint that returns all public events:

https://api.github.com/events

The first event I got from a query just now was:

```
  {
    "id": "5510765310",
    "type": "WatchEvent",
    "actor": {
      "id": 12230841,
      ...
    },
    "repo": {
      "id": 76620213,
      ...
    },
    "payload": {
      "action": "started"
    },
    "public": true,
    "created_at": "2017-03-17T01:19:24Z"
  }
```

The ID of that event exceeds 2^31 - 1.

So far we've been using `int` IDs for consistency with existing fields. But it definitely makes sense to address all of them at once in a future change.

shurcooL:
Actually, that's not a great example because https://godoc.org/github.com/google/go-github/github#Event.ID type is string, not int.

shurcooL:
The fact that both events and notifications (types of objects that are very likely to have IDs surpass 2 billion) encode their IDs as JSON strings rather than JSON numbers suggests that other IDs may be much smaller.

> Is there documentation to suggest that Github IDs are limited to 2 billion items? Seems unlikely. But I can't find anything.

I've looked, and couldn't find anything either. But this is something that's important for the API docs to specify.

Perhaps it'd be useful to send GitHub support an email to inquire about this. If they can assure us that those IDs are guaranteed to fit into int32, we won't have to make a change. If they can confirm that int32 may not be enough, we can feel better about making the change.

bradfitz:
> If they can assure us that those IDs are guaranteed to fit into int32

It's a global ID, right? It's not an ID scoped to just a single org or repo, right? If so, we don't need their confirmation. 2 billion is too small.

> because https://godoc.org/github.com/google/go-github/github#Event.ID type is string

That is its own question. Why is it a string and not some integer type? Can it actually be non-numeric?

[...]

PR #733, opened 2017-09-29
Add missing apps methods

### LINKED ISSUE #734: Add missing apps methods
Currently, only a subset of the GitHub Apps API is supported: https://github.com/google/go-github/blob/master/github/apps.go

Compare with: https://developer.github.com/v3/apps/

(As a side note, [the spreadsheet ](https://docs.google.com/spreadsheets/d/1wlHnoJSAN01nXUUF1JobsaKO4mJjAwZNWK8L_4ZSlls/edit#gid=0) doesn't mention the apps API at all.)

mandrean:
This pull request provides the missing [apps](https://developer.github.com/v3/apps/) methods and tests.

Fixes #734.

### DIFF
--- a/github/apps.go
+++ b/github/apps.go
@@ -5,13 +5,68 @@
 
 package github
 
-import "context"
+import (
+	"context"
+	"fmt"
+	"time"
+)
 
 // AppsService provides access to the installation related functions
 // in the GitHub API.
 //
 // GitHub API docs: https://developer.github.com/v3/apps/
 type AppsService service
+
+// App represents a GitHub App.
+type App struct {
+	ID          *int       `json:"id,omitempty"`
+	Owner       *User      `json:"owner,omitempty"`
+	Name        *string    `json:"name,omitempty"`
+	Description *string    `json:"description,omitempty"`
+	ExternalURL *string    `json:"external_url,omitempty"`
+	HTMLURL     *string    `json:"html_url,omitempty"`
+	CreatedAt   *time.Time `json:"created_at,omitempty"`
+	UpdatedAt   *time.Time `json:"updated_at,omitempty"`
+}
+
+// InstallationToken represents an installation token.
+type InstallationToken struct {
+	Token     *string    `json:"token,omitempty"`
+	ExpiresAt *time.Time `json:"expires_at,omitempty"`
+}
+
+// Get a single GitHub App. Passing the empty string will get
+// the authenticated GitHub App.
+//
+// Note: appSlug is just the URL-friendly name of your GitHub App.
+// You can find this on the settings page for your GitHub App
+// (e.g., https://github.com/settings/apps/:app_slug).
+//
+// GitHub API docs: https://developer.github.com/v3/apps/#get-a-single-github-app
+func (s *AppsService) Get(ctx context.Context, appSlug string) (*App, *Response, error) {
+	var u string
+	if appSlug != "" {
+		u = fmt.Sprintf("apps/%v", appSlug)
+	} else {
+		u = "app"
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
+
+	app := new(App)
+	resp, err := s.client.Do(ctx, req, app)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return app, resp, nil
+}
 
 // ListInstallations lists the installations that the current GitHub App has.
 //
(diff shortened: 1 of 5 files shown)


mandrean:
Note: I accidentally pushed an incorrect squash, but fixed it. However, that caused `googlebot` to remove the `cla: yes` label. But I have signed the `cla`. Just FYI.

github/apps.go
```diff
@@ -5,14 +5,65 @@
...
+type InstallationToken struct {
+	Token     *string    `json:"token,omitempty"`
+	ExpiresAt *time.Time `json:"expires_at,omitempty"`
+}
+
+// Get a single GitHub App. Passing the empty string will get
+// the authenticated GitHub App.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/#get-a-single-github-app
+func (s *AppsService) Get(ctx context.Context, slug string) (*App, *Response, error) {
```
gmlewis: Can you please change `slug` to `appSlug` to match the docs and also add the following comment so that people don't have to follow the link to figure out what an `appSlug` is?

```go
// Note: appSlug is just the URL-friendly name of your GitHub App.
// You can find this on the settings page for your GitHub App
// (e.g., https://github.com/settings/apps/:app_slug).
```

github/apps.go
```diff
@@ -5,14 +5,65 @@
...
 
 // AppsService provides access to the installation related functions
 // in the GitHub API.
 //
 // GitHub API docs: https://developer.github.com/v3/apps/
 type AppsService service
 
+// App represents a GitHub App.
+type App struct {
+	ID          *int       `json:"id,omitempty"`
```
gmlewis: This is actually more of a question for @shurcooL or @willnorris...

In the light of #597, do you think we should start using `int64` for numeric IDs for new APIs in order to make the implementation of #597 easier (by providing an example and reducing the number of changes needed)?

Or do you think that we should remain internally consistent and just change everything all at once when #597 is addressed?

github/apps.go
```diff
@@ -38,3 +89,77 @@ func (s *AppsService) ListInstallations(ctx context.Context, opt *ListOptions) (
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
+
+	var i struct {
```
gmlewis: @shurcooL - what do you think about making this a new `type ... struct` and include the `TotalCount` field, then returning a pointer to the struct in the return value of this method?

I'm asking because it seems more consistent with the rest of the package and also allows GitHub to add/tweak their response and we could more flexibly support it however they decide to tweak it.

github/apps_installation.go
```diff
@@ -47,3 +50,61 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
+
+	var r struct {
```
gmlewis: @shurcooL - Same for this one... make a new `type ... struct` and return it?

github/apps_installation.go
```diff
@@ -47,3 +50,61 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
...
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
```
gmlewis: Is this needed for this endpoint?  I didn't see it in the link on line 57.

github/apps.go
shurcooL: IMO it's unclear what the state of #597 is right now. The issue has been opened a while ago, but I haven't seen any reports of issues from production use. I'm not convinced that we should act right away, or how we should act. I think that we need to collect more data in that issue (or, wait until there's a concrete need to act).

Because of that, I think we should not try to incorporate that issue into this one. So, I suggest we stay consistent, and deal with `int` -> `int64` changes (or another solution), if at all, later and separately from this PR.

github/apps.go
shurcooL: I agree, I think we should have this method return a pointer to a struct that corresponds to the JSON response in the documentation.

Returning `[]*Installation` would've made sense if the GitHub response was a JSON array of installation objects, but it's a JSON object containing `installations` field with the array.

The only question is what to name it, if we make it exported (I think we have to, don't we?). Perhaps `ListUserInstallationsResponse`?

shurcooL: Then again, it's unfortunate it's quite different from the existing `ListInstallations` method (https://developer.github.com/v3/apps/#find-installations) which _does_ return a JSON array of installations...

I can't help but think GitHub might've made the API inconsistent unintentionally and might want to change it while it's still in preview, so it'd be worth shooting them an email about it.

mandrean: @gmlewis @shurcooL I did this in my original contribution, but later removed it because I didn't see this pattern being used in other places in `go-github` where the same type of JSON structure was being handled.

github/apps.go
mandrean: Done!

github/apps.go
shurcooL: > I didn't see this pattern being used in other places in `go-github` where the same type of JSON structure was being handled.

Mind sharing examples of that? It would be helpful.

mandrean: @shurcooL: Same file. This is currently in `master`: https://github.com/google/go-github/blob/master/github/apps_installation.go#L40

But as you can see in the GitHub API docs, the response actually has the `total_count` field.

shurcooL: Thanks.

@gmlewis What do you think about this finding?

gmlewis: OK, looks like we do this elsewhere... I suppose that if the GitHub API starts expanding these return values to more than just `total_count` that we can then revisit making a named structure.
So I'm fine with proceeding as written.
Thanks, @mandrean and @shurcooL!

github/apps_installation.go
mandrean: Removed!

itay:
Is there any update on this PR in terms of reviewing the updates from @mandrean ? It would be great to have those missing APIs, and it's hard to add them outside the library.

gmlewis:
@mandrean - would you mind resolving the merge conflicts?

shurcooL:
With the help of @itay and PR #769, I'll take care of the merge conflicts here by force pushing the squashed and rebased commit (with merge conflicts resolved) to the PR branch. I've confirmed that it's good and doesn't differ from the contents of this PR (aside from conflict resolution).

(For reference, the previous head commit of this PR was 0142ac9807a45f57f16c63f6ea682b9a3cc9e87a.)

mandrean:
Thanks everyone (and @shurcooL for fixing the merge conflicts - I totally missed @gmlewis question)

❤️

PR #738, opened 2017-10-03
orgs: Fixup ListPendingOrgInvitations.

thorduri:
The API doesn't take an organization ID, but rather the name.

This is obviously a breaking change to the API -- and I'm unsure how to handle that,
given that it doesn't function as is, compare:

```
curl -vvv -H "Authorization: token $GITHUB_TOKEN" https://api.github.com/orgs/<id>/invitations
curl -vvv -H "Authorization: token $GITHUB_TOKEN" https://api.github.com/orgs/<name>/invitations
```


shurcooL:
THanks for the PR. This looks like it fixes issue #730, is that right? We already have a PR open for it, see #731, it’s waiting on review.

Your change is nearly identical, which serves as positive validation.

thorduri:
@shurcooL  Indeed!

Apologies for not checking open Issues/PRs -- I'll hang back until #731 lands in master.

PR #740, opened 2017-10-03
Add missing URL`s data to Issue type

### LINKED ISSUE #739: No repository data for retrieved issues
Trying to get a single issue using Get from IssuesServices[0] return it with no errors but Repository field is nil. Checking GitHub API[1] for the same issue, it seems that no "repository" data is returned, instead "repository_url" is. I am not able to find any documented API change about this

[0] https://github.com/google/go-github/blob/master/github/issues.go#L234
[1] https://api.github.com/repos/google/go-github/issues/735

artmello:
Retrieve following data fields returned for Issues from GitHub API  (Fix #739) :
- repository_url;
- comments_url;
- events_url;
- labels_url.

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -41,6 +41,10 @@ type Issue struct {
 	ClosedBy         *User             `json:"closed_by,omitempty"`
 	URL              *string           `json:"url,omitempty"`
 	HTMLURL          *string           `json:"html_url,omitempty"`
+	CommentsURL      *string           `json:"comments_url,omitempty"`
+	EventsURL        *string           `json:"events_url,omitempty"`
+	LabelsURL        *string           `json:"labels_url,omitempty"`
+	RepositoryURL    *string           `json:"repository_url,omitempty"`
 	Milestone        *Milestone        `json:"milestone,omitempty"`
 	PullRequestLinks *PullRequestLinks `json:"pull_request,omitempty"`
 	Repository       *Repository       `json:"repository,omitempty"`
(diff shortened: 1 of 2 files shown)


e-beach:
@artmello You also need to run `go generate` as per the contributing guide - that should make the build pass.

I'll note that `repository_url` is part of the documented reponse at https://developer.github.com/v3/issues/#get-a-single-issue,  but it seems we are missing several other fields, including `labels_url`, `comments_url`, and `events_url`.

github/issues.go
```diff
@@ -44,6 +44,7 @@ type Issue struct {
 	Milestone        *Milestone        `json:"milestone,omitempty"`
 	PullRequestLinks *PullRequestLinks `json:"pull_request,omitempty"`
 	Repository       *Repository       `json:"repository,omitempty"`
+	RepositoryURL    *string           `json:"repository_url,omitempty"`
```
shurcooL: Please move it under `HTMLURL` so that it’s grouped together with all `url` fields.

shurcooL:
> but it seems we are missing several other fields, including `labels_url`, `comments_url`, and `events_url`.

That sounds accurate. We can either add those too, or defer doing so until someone runs into a concrete need for them. Either way is fine. I’ll leave it to PR author to decide.

artmello:
Thanks for reviewing it. I added the other 3 missing URLs  and executed "go generate". Should I take some action related with "Repository" field (remove it)? On my test case, getting issues from search or from IssuesService, this field is always nil, but I am not that familiar with the code base to know if it can be populated somewhere else

shurcooL:
It probably is populated elsewhere. We don't need to deal with it as part of this PR.

PR #741, opened 2017-10-05
Add NestedTeamsPreview header to remaining endpoints

e-beach:
This adds the preview header to the rest of the endpoints mentioned in https://developer.github.com/changes/2017-08-30-preview-nested-teams/.

In summary, supporting the preview API required adding a parameter for `CreateTeam` and `EditTeam`, creating the `NewPullRequest` struct, and adding the preview header to every header involving a team (this PR).

This should finish up support for the API.

### DIFF
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -242,8 +242,8 @@ func (s *OrganizationsService) ListTeamRepos(ctx context.Context, team int, opt 
 	}
 
 	// TODO: remove custom Accept header when topics API fully launches.
-	headers := strings.Join([]string{mediaTypeTopicsPreview, mediaTypeNestedTeamsPreview}, ", ")
-	req.Header.Set("Accept", headers)
+	headers := []string{mediaTypeTopicsPreview, mediaTypeNestedTeamsPreview}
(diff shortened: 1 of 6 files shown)


github/orgs_teams.go
```diff
@@ -237,7 +242,8 @@ func (s *OrganizationsService) ListTeamRepos(ctx context.Context, team int, opt
 	}
 
 	// TODO: remove custom Accept header when topics API fully launches.
-	req.Header.Set("Accept", mediaTypeTopicsPreview)
+	headers := strings.Join([]string{mediaTypeTopicsPreview, mediaTypeNestedTeamsPreview}, ", ")
```
gmlewis: nit: I prefer the way you did this at lines 269-270 below where you first make the string slice then join them on the next line, and if you wish to make the two consistent with each other, I would prefer that you change this instance...
(looking at the code, I see that the two-step method is already being done in `repos.go` and `repos_test.go`)...
but I'm fine either way and will wait to merge until I hear your opinion.

Thanks!

e-beach: @gmlewis Thanks - I fixed it.

gmlewis:
Thank you, @e-beach!
LGTM.
Merging.

PR #744, opened 2017-10-05
Adds RequireCodeOwnerReviews to PullRequestReviewsEnforcement

denniswebb:
While working on terraform-providers/terraform-provider-github#50 I realized that this was missing.

GitHub.com support page about feature https://developer.github.com/v3/repos/branches/#update-branch-protection

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -562,7 +562,7 @@ type PullRequestReviewsEnforcement struct {
 	DismissalRestrictions DismissalRestrictions `json:"dismissal_restrictions"`
 	// Specifies if approved reviews are dismissed automatically, when a new commit is pushed.
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
-	// Specifies if an approved review is required in pull requests including files with a designated code owner.
+	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews"`
 }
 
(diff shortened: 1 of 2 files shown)


elliott-beach:
It feels somewhat despotic to have the power to review PRs, but will proceed to do so, @gmlewis.

github/repos.go
```diff
@@ -562,6 +562,8 @@ type PullRequestReviewsEnforcement struct {
 	DismissalRestrictions DismissalRestrictions `json:"dismissal_restrictions"`
 	// Specifies if approved reviews are dismissed automatically, when a new commit is pushed.
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
+	// Specifies if an approved review is required in pull requests including files with a designated code owner.
```
elliott-beach: To maintain the same style as the current API, this should be `RequireCodeOwnerReviews specifies...`

github/repos.go
```diff
@@ -572,6 +574,8 @@ type PullRequestReviewsEnforcementRequest struct {
 	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
 	// Specifies if approved reviews can be dismissed automatically, when a new commit is pushed. (Required)
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
+	// Specifies if an approved review is required in pull requests including files with a designated code owner.
```
elliott-beach: Same here.

gmlewis:
Hi Elliott, I apologize for not discussing this with you first... are you open to reviewing PRs for this repo? You come highly recommended.  😄 
Also, just FYI, I'm about to go on vacation for 9 days and may not be very active on this repo during this time period. Others may step in to fill the gap, but if not, I'll try to catch up when I return.

gmlewis:
Thank you!
Merging.

github/repos.go
shurcooL: (That is a hard requirement for top level identifiers, but for struct fields, I think both ways are acceptable. This is currently consistent with the documentation for fields above.)

PR #746, opened 2017-10-05
Adds state to project options per github's docs

seemethere:
State was missing, useful for updating projects per:
https://developer.github.com/v3/projects/#update-a-project

Adds the ability to `open`/`close` projects based on `ProjectOptions` with `ProjectUpdate`

### DIFF
--- a/github/projects.go
+++ b/github/projects.go
@@ -65,7 +65,11 @@ type ProjectOptions struct {
 	Name string `json:"name,omitempty"`
 	// The body of the project. (Optional.)
 	Body string `json:"body,omitempty"`
-	// State of the project. Either open or closed
+
+	// The following field(s) are only applicable for update.
+	// They should be left with zero values for creation.
+
+	// State of the project. Either "open" or "closed". (Optional.)
 	State string `json:"state,omitempty"`
 }
 
--- a/github/projects.go
+++ b/github/projects.go
@@ -65,6 +65,8 @@ type ProjectOptions struct {
 	Name string `json:"name,omitempty"`
 	// The body of the project. (Optional.)
 	Body string `json:"body,omitempty"`
+	// State of the project. Either open or closed
+	State string `json:"state,omitempty"`
 }
 
 // UpdateProject updates a repository project.
--- a/github/projects_test.go
+++ b/github/projects_test.go
@@ -18,7 +18,7 @@ func TestProjectsService_UpdateProject(t *testing.T) {
 	setup()
 	defer teardown()
 
-	input := &ProjectOptions{Name: "Project Name", Body: "Project body."}
+	input := &ProjectOptions{Name: "Project Name", Body: "Project body.", State: "open"}
 
 	mux.HandleFunc("/projects/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "PATCH")


github/projects.go
```diff
@@ -65,6 +65,8 @@ type ProjectOptions struct {
 	Name string `json:"name,omitempty"`
 	// The body of the project. (Optional.)
 	Body string `json:"body,omitempty"`
+	// State of the project. Either open or closed
```
shurcooL: Add a missing period to the end of the last sentence.

Also, let's figure if this is optional or required, and document that too, just like the two fields above.

shurcooL: From looking over https://developer.github.com/v3/projects/, it seems that this field is optional and only applicable to the `UpdateProject` endpoint. Let's write that:

```Go
// State of the project. Either "open" or "closed". (Only for update; optional.)
State string `json:"state,omitempty"`
```

I wonder if it's okay that we're reusing the same `ProjectOptions` for updating and creating projects, given that creating projects doesn't have `State` field...

seemethere: @shurcooL The `state` field is invalid for Project Creation.

It is optional for Project Updates though.

In this case I think it'd warrant creating its own `ProjectUpdateOptions` struct since it's not valid for Project Creation to have the `state` field.

wdyt?

shurcooL: Perhaps we should organize the fields like this:

```Go
Name string `json:"name,omitempty"` // The name of the project. (Required for creation; optional for update.)
Body string `json:"body,omitempty"` // The body of the project. (Optional.)

// The following fields are only applicable for update.
// They should be left with zero values for creation.
State string `json:"state,omitempty"` // State of the project. Either "open" or "closed". (Optional.)
```

shurcooL: > In this case I think it'd warrant creating its own `ProjectUpdateOptions` struct since it's not valid for Project Creation to have the `state` field.

Yes, it's possible for us to do that too. It would be a breaking API change.

I'm not quite sure which of the two approaches would be better, need to think about it.

shurcooL: We do have a history of reusing same struct for creating and updating certain resources (e.g., [issues](https://godoc.org/github.com/google/go-github/github#IssueRequest), [comments](https://godoc.org/github.com/google/go-github/github#IssueComment)).

I think let's go with my suggested approach of documenting the separate update-only fields for now, and reusing the same struct. They don't deviate very strongly at this point. Should that change in the future, we can make the breaking API change _then_.

github/projects.go
```diff
@@ -65,6 +65,10 @@ type ProjectOptions struct {
 	Name string `json:"name,omitempty"`
 	// The body of the project. (Optional.)
 	Body string `json:"body,omitempty"`
+
+	// The following fields are only applicable for update.
```
gmlewis: nit: It seems slightly odd to me to describe everything here in the plural when it is just a single field, but not a big deal.
another nit: we typically comment on the line before as to what values a field can have, but this is not a big deal either.

seemethere: Yeah a bit of an oversight by me, I've amended the commit to include those changes as I think they are valuable as well.

shurcooL: Unlike Gerrit, on GitHub, I can’t see what has changed if you amend a commit and force push it. Can you tell me what you changed please?

shurcooL: > another nit: we typically comment on the line before as to what values a field can have, but this is not a big deal either.

Again, it’s hard for me to be sure, but wasn’t this already the case before? The `Either "open" or "closed".` sentence. I wish I could see what changed... :(

seemethere:
Any updates on this?

shurcooL:
Merged. Thanks @seemethere!

github/projects.go
```diff
@@ -65,6 +65,12 @@ type ProjectOptions struct {
 	Name string `json:"name,omitempty"`
 	// The body of the project. (Optional.)
 	Body string `json:"body,omitempty"`
+
+	// The following field(s) are only applicable for update.
+	// They should be left with zero values for creation.
+
+	// State of the project. Either "open" or "closed". (Optional.)
+	State string `json:"state,omitempty"`
```
gmlewis: Looking now at #969, I'm thinking this optional field should be changed to `State *string ...` so that it is not passed if it is not provided. (In other words, to let `omitempty` do its thing.)
If there are no objections, I'll comment about this in #969.

PR #748, opened 2017-10-10
parent_team_id is an integer identifier, not a string

alindeman:
Using the code as written in the `NestedTeamsPreview` branch produces an error when talking to the GitHub API:

```
panic: PATCH https://api.github.com/teams/2514825: 422 Invalid request.

For 'properties/parent_team_id', "2514824" is not a number or null. []
```

The [documentation](https://developer.github.com/v3/orgs/teams/#create-team) states that this parameter is an `id` type which--for the GitHub API--appears to mean an integer, not a string.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -2965,9 +2965,9 @@ func (n *NewTeam) GetLDAPDN() string {
 }
 
 // GetParentTeamID returns the ParentTeamID field if it's non-nil, zero value otherwise.
-func (n *NewTeam) GetParentTeamID() string {
+func (n *NewTeam) GetParentTeamID() int {
 	if n == nil || n.ParentTeamID == nil {
-		return ""
+		return 0
 	}
 	return *n.ParentTeamID
 }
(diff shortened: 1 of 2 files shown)


alindeman:
:eyes: @elliott-beach @gmlewis

elliott-beach:
LGTM, 

but I think GitHub should fix their docs. This may be the first JSON parameter in the API that has the type of id, and when the API returns ids, the type is an integer, not a string. It seems the documentation is ambiguous on the argument type - I will email GitHub.

elliott-beach:
Response from GitHub:
>Thanks for writing back to clarify! I see what you mean now and agree it would be helpful to update the type there. I've opened an issue internally for our Documentation team to take a look and consider updating for future iterations. Please let us know if you spot anything else like this and we'd be happy to take another look!

So, we will have to wait for this to be fixed, or not.

shurcooL:
So the response doesn't really confirm that the type is an integer, does it?

elliott-beach:
I got another email that confirmed that the type is an integer. I also realized that my thinking on this had been backward the whole time - it makes perfect sense for ID to be an int, it should not have required emailing GitHub at all.

shurcooL:
This has my and Elliott's approvals. I'll merge this PR into `NestedTeamsPreview` branch and we can continue discussion/review there. Thanks @alindeman!

PR #749, opened 2017-10-11
Enhancement/add remove repository

MorrisLaw:
Added support for preview installations API for the following:
- adding a repository to an installation
- removing a repository from an installation


shurcooL:
> I've attempted to tackle this issue but my changes have failed in travis checks. Can someone please point me in the right direction, the changes are located in PR #749. I am new to Go and this github client in general, so I may be missing something obvious. Thank you everyone for your patience and advice! (: @shurcooL @gmlewis 

@MorrisLaw, you can click on the icon here to see the CI results:

image

It links to the [CI page](https://travis-ci.org/google/go-github/builds/286640168). There, you can see all 4 jobs failed. Pick the first one ([Go 1.9.x](https://travis-ci.org/google/go-github/jobs/286640169) job), and look at the log. If you scan towards the bottom, you'll see something in red, that's the first failure:

```
$ go test -v -race ./...
import cycle not allowed in test
package github.com/google/go-github/github (test)
	imports github.com/google/go-github/github
FAIL	github.com/google/go-github/github [setup failed]
?   	github.com/google/go-github/example/appengine	[no test files]
?   	github.com/google/go-github/example/basicauth	[no test files]
?   	github.com/google/go-github/test/fields	[no test files]
?   	github.com/google/go-github/test/integration	[no test files]
The command "go test -v -race ./..." exited with 1.
```

The command that failed was `go test -v -race ./...`. You will need to fix that issue.

github/apps_installation_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"net/http"
 	"reflect"
 	"testing"
+
+	"github.com/google/go-github/github"
```
shurcooL: This is an internal package test .go file. Therefore, you don't need to import that package - you're already in it.

Instead of writing `github.ErrorResponse` below, you can access it with just `ErrorResponse`.

That will fix the problem.

You can read more about internal/external tests at https://golang.org/cmd/go/#hdr-Test_packages.

github/apps_installation.go
```diff
@@ -47,3 +50,38 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
...
+// GitHub API docs: https://developer.github.com/v3/apps/installations/#add-repository-to-installation
+func (s *AppsService) AddRepo(ctx context.Context, instID, repoID int) (*Repository, *Response, error) {
+	u := fmt.Sprintf("/apps/installations/%v/repositories/%v", instID, repoID)
+	req, err := s.client.NewRequest("PUT", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeV3)
```
shurcooL: `mediaTypeV3` is already the default:

https://github.com/google/go-github/blob/0f6d3ce15ec23c92c74d014303a167a9a374dd7e/github/github.go#L269

So no need to do this, you can delete the two lines above.

github/apps_installation.go
```diff
@@ -47,3 +50,38 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
 
 	return r.Repositories, resp, nil
 }
+
+// AddRepo adds a single repository to an installation.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/installations/#add-repository-to-installation
+func (s *AppsService) AddRepo(ctx context.Context, instID, repoID int) (*Repository, *Response, error) {
```
shurcooL: I think we should call this method `AddRepository` rather than `AddRepo`. We already use the full "Repository" word in many other method names, e.g.:

- [`ActivityService.GetRepositorySubscription`](https://godoc.org/github.com/google/go-github/github#ActivityService.GetRepositorySubscription)
- [`ActivityService.DeleteRepositorySubscription`](https://godoc.org/github.com/google/go-github/github#ActivityService.DeleteRepositorySubscription)
- [`IssuesService.ListRepositoryEvents`](https://godoc.org/github.com/google/go-github/github#IssuesService.ListRepositoryEvents)

github/apps_installation.go
```diff
@@ -47,3 +50,38 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
...
+		return nil, nil, err
+	}
+
+	return r, resp, nil
+}
+
+// RemoveRepo removes a single repository from an installation.
+//
+// GitHub docs: https://developer.github.com/v3/apps/installations/#remove-repository-from-installation
+func (s *AppsService) RemoveRepo(ctx context.Context, instID, repoID int) (*Response, error) {
```
shurcooL: Similarly, this would be `RemoveRepository`.

github/apps_installation.go
```diff
@@ -47,3 +50,38 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
...
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeV3)
+
+	r := new(Repository)
+	resp, err := s.client.Do(ctx, req, r)
+	if err != nil {
+		return nil, nil, err
```
shurcooL: Even if err != nil, you should return `resp` rather than `nil`:

```Go
if err != nil {
    return nil, resp, err
}
```

This is what all other methods do.

github/apps_installation_test.go
```diff
@@ -38,3 +38,39 @@ func TestAppsService_ListRepos(t *testing.T) {
...
+
+	mux.HandleFunc("/apps/installations/1/repositories/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		testHeader(t, r, "Accept", mediaTypeV3)
+		fmt.Fprint(w, `{"id":1,"name":"n","description":"d","owner":{"login":"l"},"license":{"key":"mit"}}`)
+	})
+
+	repo, _, err := client.Apps.AddRepository(context.Background(), 1, 1)
+	if err != nil {
+		t.Errorf("Apps.AddRepo returned error: %v", err)
```
shurcooL: It'd be nice to rename the method name here too.

github/apps_installation_test.go
```diff
@@ -38,3 +38,39 @@ func TestAppsService_ListRepos(t *testing.T) {
...
+	}
+
+	want := &Repository{ID: Int(1), Name: String("n"), Description: String("d"), Owner: &User{Login: String("l")}, License: &License{Key: String("mit")}}
+	if !reflect.DeepEqual(repo, want) {
+		t.Errorf("AddRepo returned %+v, want %+v", repo, want)
```
shurcooL: And here.

github/apps_installation_test.go
```diff
@@ -38,3 +38,39 @@ func TestAppsService_ListRepos(t *testing.T) {
...
+	defer teardown()
+
+	mux.HandleFunc("/apps/installations/1/repositories/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	_, err := client.Apps.RemoveRepository(context.Background(), 1, 1)
+	if err != nil {
+		t.Errorf("Apps.RemoveRepo returned error: %v", err)
```
shurcooL: As well as here.

github/apps_installation.go
```diff
@@ -47,3 +50,35 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
 
 	return r.Repositories, resp, nil
 }
+
+// AddRepository adds a single repository to an installation.
+//
+// GitHub API docs: https://developer.github.com/v3/apps/installations/#add-repository-to-installation
+func (s *AppsService) AddRepository(ctx context.Context, instID, repoID int) (*Repository, *Response, error) {
+	u := fmt.Sprintf("/apps/installations/%v/repositories/%v", instID, repoID)
```
shurcooL: We need to not include the starting slash in the URLs, otherwise enterprise GH URLs will not work.

This should be:

```Go
u := fmt.Sprintf("apps/installations/%v/repositories/%v", instID, repoID)
```

github/apps_installation.go
```diff
@@ -47,3 +50,35 @@ func (s *AppsService) ListRepos(ctx context.Context, opt *ListOptions) ([]*Repos
...
+	}
+
+	return r, resp, nil
+}
+
+// RemoveRepository removes a single repository from an installation.
+//
+// GitHub docs: https://developer.github.com/v3/apps/installations/#remove-repository-from-installation
+func (s *AppsService) RemoveRepository(ctx context.Context, instID, repoID int) (*Response, error) {
+	u := fmt.Sprintf("/apps/installations/%v/repositories/%v", instID, repoID)
```
shurcooL: Same here, please remove the first slash in the URL. You'll notice all other endpoints don't include it either.

MorrisLaw:
Thank you for the thorough reviewing and for providing all of the useful feedback! @shurcooL

PR #751, opened 2017-10-13
remove ListOptions from RepositoryListAllOptions (they have no effect)

beyang:
The standard pagination options have no effect in the `/repositories` endpoint, both empirically and according to the documentation: https://developer.github.com/v3/repos/#list-all-public-repositories

> Note: Pagination is powered exclusively by the since parameter. Use the Link header to get the URL for the next page of repositories.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	libraryVersion = "12"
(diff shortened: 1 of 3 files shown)


shurcooL:
I'll bump `libraryVersion` and merge this.

PR #753, opened 2017-10-13
Make tests catch invalid absolute URLs in methods.

### LINKED ISSUE #752: Make tests catch invalid absolute URLs in methods.
When adding support for new endpoints, it's possible to make a mistake and use an absolute URL in the method implementation, rather than a relative one.

We need to use relative URLs so that the `Client.BaseURL` field works properly in `NewRequest` method:

https://github.com/google/go-github/blob/0f6d3ce15ec23c92c74d014303a167a9a374dd7e/github/github.go#L112-L115

https://github.com/google/go-github/blob/0f6d3ce15ec23c92c74d014303a167a9a374dd7e/github/github.go#L238-L247

E.g., this is invalid:

```Go
u := fmt.Sprintf("/admin/users/%v/authorizations", username)
```

This is correct:

```Go
u := fmt.Sprintf("admin/users/%v/authorizations", username)
```

Including that leading slash has no negative effect on GitHub API clients, but it typically breaks GitHub Enterprise clients. That makes this type of mistake easy to miss during code review.

Right now, our tests do not catch this, so we rely on human reviewers to catch it (e.g., this recently happened in #749).

I've thought about it and I think I found a way to make it so that our tests catch this. That will reduce the amount of work the human reviewers have to do, by automating more work. :) I'll send a PR soon.

P.S. Thanks to @MorrisLaw for giving me this idea.

shurcooL:
This PR consists of 4 commits, each commit has a commit message describing what was done. Please review this PR by looking at the 4 commits individually and reading their commit messages (as well as the CI result of each commit).

Fixes #752.

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -33,17 +33,39 @@ var (
 	server *httptest.Server
 )
 
+const (
+	// baseURLPath is a non-empty Client.BaseURL path to use during tests,
+	// to ensure relative URLs are used for all endpoints. See issue #752.
+	baseURLPath = "/api-v3"
+)
+
 // setup sets up a test HTTP server along with a github.Client that is
 // configured to talk to that test server. Tests should register handlers on
 // mux which provide mock responses for the API method being tested.
 func setup() {
 	// test server
 	mux = http.NewServeMux()
-	server = httptest.NewServer(mux)
+
+	// We want to ensure that tests catch mistakes where the endpoint URL is
+	// specified as absolute rather than relative. It only makes a difference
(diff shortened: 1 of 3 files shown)


shurcooL:
Thanks for the quick review. I will leave out the "break UsersService.ListAll method" and "unbreak UsersService.ListAll method" commits, and merge the other two as they are (i.e., separately).

shurcooL:
Merged as 10a2b148cba999c248d9f7cea42308b24f390519 and 7de811820d2b3b6ef66f466bdae21e494a52b633.

PR #754, opened 2017-10-18
add optional User parameter when forking

jstrachan:
so that we can support forking into user accounts instead of just organisations


shurcooL:
I don't see this parameter documented at https://developer.github.com/v3/repos/forks/#create-a-fork. Is it documented somewhere else?

As far as I can tell, the `organization` parameter is optional, and if left out, it makes a fork into the authenticated user account. Can you fork into another user's account?

jstrachan:
Thanks for the super fast response!

So I tried putting the user account name into the `organisation` parameter and got an error that this was a user account not an organisation so I kinda assumed we'd need a `user` parameter - I added it and it worked. 

It turns out you can just omit the `organisation` parameter when doing user forks; there's no need for a new `user` parameter - DOH! Apologies, lets close this!

PR #758, opened 2017-10-21
NestedTeamsPreview: Add the ListChildTeams endpoint

elliott-beach:
As per discussion in https://github.com/google/go-github/pull/747#issuecomment-336780818, this endpoint is missing.

### DIFF
--- a/github/orgs_teams.go
+++ b/github/orgs_teams.go
@@ -211,6 +211,32 @@ type OrganizationListTeamMembersOptions struct {
 	ListOptions
 }
 
+// ListChildTeams lists child teams for a team.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/teams/#list-child-teams
+func (s *OrganizationsService) ListChildTeams(ctx context.Context, id int, opt *ListOptions) ([]*Team, *Response, error) {
+	u := fmt.Sprintf("teams/%v/teams", id)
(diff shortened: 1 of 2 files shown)


elliott-beach:
I signed it (before). Also, I updated my CLA contact info with my new username, which I hope will prevent this problem from re-occurring in the future.

github/orgs_teams.go
```diff
@@ -211,6 +211,32 @@ type OrganizationListTeamMembersOptions struct {
 	ListOptions
 }
 
+// ListChildTeams lists child teams for a team.
+//
+// GitHub API docs: https://developer.github.com/v3/orgs/teams/#list-child-teams
+func (s *OrganizationsService) ListChildTeams(ctx context.Context, id int, opt *ListOptions) ([]*Team, *Response, error) {
```
shurcooL: Given the name of the receiver type (`OrganizationsService`) and method (`ListChildTeams`), it might not be 100% clear what `id int` refers to. It's quite clear in the endpoint URL itself, being `/teams/:id/teams`, which is why `id` is fine there.

What do you think about renaming the parameter to `teamID` to make it more clear?

elliott-beach: Good idea, will do.

elliott-beach:
Should be good.

Issue #756, opened 2017-10-20
Add convenience helper for creating client with custom (enterprise) URL.


chriskingnet:
I am trying to write a little app that connects to a Github Enterprise instance. I would like to be able to pass in the base enterprise URL, and have it override the default one.

As far as I can see (from the tests and research I have done) the API for Github Enterprise includes all of the regular functionality, plus some more Enterprise-only features (but perhaps someone knows different?). Because of this, just adding the ability to pass in the URL would already be of great benefit.

I have created this issue to discuss, but have also already made an attempt at the changes and will submit a PR, in case this issue is accepted.

shurcooL:
`go-github` already has support for GitHub Enterprise. See the `BaseURL` field of `Client` struct:

https://godoc.org/github.com/google/go-github/github#Client

chriskingnet:
Hi @shurcooL, thanks for getting back to me. You're right, I totally didn't notice this! I don't have any better argument for doing it my way - I missed this mostly because of my relative inexperience with Go, I think. 

I will close the issue - thanks for your work on this lib :)

shurcooL:
No problem.

I think we can still put up the `NewEnterpriseClient` from #757 for consideration as an optional helper. It could be rewritten to use the existing API, but accept all URLs and add trailing slash if neccessary. E.g., something like:

```Go
// NewEnterpriseClient returns a new GitHub API client with provided
// base URL and upload URL.
// If either URL does not have a trailing slash, one is added automatically.
// If a nil httpClient is provided, http.DefaultClient will be used.
//
// Note that NewEnterpriseClient is a convenience helper only;
// its behavior is equivalent to using NewClient, followed by setting
// the BaseURL and UploadURL fields.
func NewEnterpriseClient(baseURL, uploadURL string, httpClient *http.Client) (*Client, error) {
	base, err := url.Parse(baseURL)
	if err != nil {
		return nil, err
	}
	if !strings.HasSuffix(base.Path, "/") {
		base.Path += "/"
	}
	upload, err := url.Parse(uploadURL)
	if err != nil {
		return nil, err
	}
	if !strings.HasSuffix(upload.Path, "/") {
		upload.Path += "/"
	}
	client := NewClient(httpClient)
	client.BaseURL = base
	client.UploadURL = upload
	return client, nil
}
```

I'm not completely sure if it's worth adding such a helper, but let's consider/discuss it. What do you think? /cc @gmlewis @elliott-beach

chriskingnet:
[...]

PR #763, opened 2017-10-26
NewEnterpriseClient method with custom urls

chriskingnet:
With this PR I have completed the following, addressing the issues discussed in #756:

- Introduced new func to create a new enterprise client
- Ensured base and upload urls are formatted with a trailing slash
- Added appropriate comment documentation
- Extended test cases to cover this new creation method

Did I cover everything? :)

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -235,21 +235,35 @@ func NewClient(httpClient *http.Client) *Client {
 	return c
 }
 
-// NewEnterpriseClient returns a new GitHub API client (same as NewClient),
-// except it allows injection of custom base and upload urls. These urls might
-// be the same, but for flexibility both can be set separately.
-func NewEnterpriseClient(httpClient *http.Client, baseURL string, uploadURL string) *Client {
-	if !strings.HasSuffix(baseURL, "/") {
-		baseURL = baseURL + "/"
+// NewEnterpriseClient returns a new GitHub API client with provided
+// base URL and upload URL (often the same URL).
+// If either URL does not have a trailing slash, one is added automatically.
+// If a nil httpClient is provided, http.DefaultClient will be used.
+//
+// Note that NewEnterpriseClient is a convenience helper only;
+// its behavior is equivalent to using NewClient, followed by setting
+// the BaseURL and UploadURL fields.
+func NewEnterpriseClient(httpClient *http.Client, baseEndpointStr string, uploadEndpointStr string) (*Client, error) {
+	baseURL, err := url.Parse(baseEndpointStr)
+	if err != nil {
+		return nil, err
 	}
-	if !strings.HasSuffix(uploadURL, "/") {
-		uploadURL = uploadURL + "/"
+	if !strings.HasSuffix(baseURL.Path, "/") {
+		baseURL.Path += "/"
+	}
+
+	uploadURL, err := url.Parse(uploadEndpointStr)
+	if err != nil {
+		return nil, err
+	}
+	if !strings.HasSuffix(uploadURL.Path, "/") {
+		uploadURL.Path += "/"
 	}
 
 	c := NewClient(httpClient)
-	c.BaseURL, _ = url.Parse(baseURL)
-	c.UploadURL, _ = url.Parse(uploadURL)
-	return c
+	c.BaseURL = baseURL
+	c.UploadURL = uploadURL
+	return c, nil
 }
 
 // NewRequest creates an API request. A relative URL can be provided in urlStr,
--- a/github/github.go
+++ b/github/github.go
@@ -243,8 +243,8 @@ func NewClient(httpClient *http.Client) *Client {
 // Note that NewEnterpriseClient is a convenience helper only;
 // its behavior is equivalent to using NewClient, followed by setting
 // the BaseURL and UploadURL fields.
-func NewEnterpriseClient(baseEndpointStr, uploadEndpointStr string, httpClient *http.Client) (*Client, error) {
-	baseEndpoint, err := url.Parse(baseEndpointStr)
+func NewEnterpriseClient(baseURL, uploadURL string, httpClient *http.Client) (*Client, error) {
+	baseEndpoint, err := url.Parse(baseURL)
 	if err != nil {
 		return nil, err
 	}
@@ -252,7 +252,7 @@ func NewEnterpriseClient(baseEndpointStr, uploadEndpointStr string, httpClient *
 		baseEndpoint.Path += "/"
 	}
 
-	uploadEndpoint, err := url.Parse(uploadEndpointStr)
+	uploadEndpoint, err := url.Parse(uploadURL)
 	if err != nil {
 		return nil, err
 	}
--- a/github/github.go
+++ b/github/github.go
@@ -235,6 +235,23 @@ func NewClient(httpClient *http.Client) *Client {
 	return c
 }
 
+// NewEnterpriseClient returns a new GitHub API client (same as NewClient),
+// except it allows injection of custom base and upload urls. These urls might
+// be the same, but for flexibility both can be set separately.
+func NewEnterpriseClient(httpClient *http.Client, baseURL string, uploadURL string) *Client {
+	if !strings.HasSuffix(baseURL, "/") {
+		baseURL = baseURL + "/"
+	}
+	if !strings.HasSuffix(uploadURL, "/") {
+		uploadURL = uploadURL + "/"
+	}
+
+	c := NewClient(httpClient)
+	c.BaseURL, _ = url.Parse(baseURL)
+	c.UploadURL, _ = url.Parse(uploadURL)
+	return c
+}
+
 // NewRequest creates an API request. A relative URL can be provided in urlStr,
 // in which case it is resolved relative to the BaseURL of the Client.
 // Relative URLs should always be specified without a preceding slash. If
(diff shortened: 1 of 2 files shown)


github/github_test.go
```diff
@@ -186,6 +186,34 @@ func TestNewClient(t *testing.T) {
...
+	c := NewEnterpriseClient(nil, baseURL, uploadURL)
+	if got, want := c.BaseURL.String(), baseURL; got != want {
+		t.Errorf("NewClient BaseURL is %v, want %v", got, want)
+	}
+	if got, want := c.UploadURL.String(), uploadURL; got != want {
+		t.Errorf("NewClient UploadURL is %v, want %v", got, want)
+	}
+}
+
+func TestNewEnterpriseClient_adssTrailingSlashToURLs(t *testing.T) {
```
gmlewis: s/adss/adds/

github/github.go
```diff
@@ -235,6 +235,23 @@ func NewClient(httpClient *http.Client) *Client {
...
+	if !strings.HasSuffix(baseURL, "/") {
+		baseURL = baseURL + "/"
+	}
+	if !strings.HasSuffix(uploadURL, "/") {
+		uploadURL = uploadURL + "/"
+	}
+
+	c := NewClient(httpClient)
+	c.BaseURL, _ = url.Parse(baseURL)
+	c.UploadURL, _ = url.Parse(uploadURL)
```
shurcooL: Can you provide rationale for why you decided to ignore errors instead of returning them, as I did in the sample in https://github.com/google/go-github/issues/756#issuecomment-338877501?

I don't necessarily disagree with the decision, I just want it to be explicit so we can understand it better and agree on.

github/github.go
```diff
@@ -235,6 +235,23 @@ func NewClient(httpClient *http.Client) *Client {
 	return c
 }
 
+// NewEnterpriseClient returns a new GitHub API client (same as NewClient),
+// except it allows injection of custom base and upload urls. These urls might
+// be the same, but for flexibility both can be set separately.
```
shurcooL: I think there's an issue here that we need to address. As is, the documentation makes it confusing/unclear as to whether the previous code of creating `NewClient` and setting its `BaseURL`, `UploadURL` fields. I think we should make it clear to users that this is an optional convenience helper, and using it (or changing old code) is not mandatory.

Also, I wouldn't use the word "injection" here, that sounds odd, how about "returns a new GitHub API client with the provided base URL and upload URL"?

The documentation I wrote in https://github.com/google/go-github/issues/756#issuecomment-338877501 did not have these issues IMO, what do you think about using it?

I can also incorporate the "These urls might be the same, but for flexibility both can be set separately" message to it as well, making it:

```Go
// NewEnterpriseClient returns a new GitHub API client with provided
// base URL and upload URL (often the same URL).
// If either URL does not have a trailing slash, one is added automatically.
// If a nil httpClient is provided, http.DefaultClient will be used.
//
// Note that NewEnterpriseClient is a convenience helper only;
// its behavior is equivalent to using NewClient, followed by setting
// the BaseURL and UploadURL fields.
```

github/github.go
```diff
@@ -235,6 +235,23 @@ func NewClient(httpClient *http.Client) *Client {
...
+func NewEnterpriseClient(httpClient *http.Client, baseURL string, uploadURL string) *Client {
+	if !strings.HasSuffix(baseURL, "/") {
+		baseURL = baseURL + "/"
+	}
+	if !strings.HasSuffix(uploadURL, "/") {
+		uploadURL = uploadURL + "/"
+	}
```
shurcooL: How about modifying the `Path` field of the URL after parsing it?

Also, why not use `+=` operator?

```Go
if !strings.HasSuffix(client.BaseURL.Path, "/") {
	client.BaseURL.Path += "/"
}
if !strings.HasSuffix(client.UploadURL.Path, "/") {
	client.UploadURL.Path += "/"
}
```

github/github_test.go
chriskingnet: Nice spot, will fix this!

github/github.go
chriskingnet: I agree, this documentation is clearer. I wasn't sure how much detail to go into about what the method did (should have looked back at your suggestion more closely!) I prefer what you have written here, and like the idea of adding in the part about the equivalence to setting the fields directly.

github/github.go
chriskingnet: Totally agree about the +=, is much better (my golang inexperience coming through :D).

No particular reason I did it this way around. However, given the discussion about returning the error, perhaps it makes more sense to parse as early as possible and edit the Path only after it is known to be valid.

github/github.go
chriskingnet: My rationale wasn't super strong, I did this to remain consistent with `NewClient`. But thinking about it now, I don't completely agree with that because we are now allowing the passing in of params that could be malformed. I would have nothing against returning an error to the user if parsing fails. I think this would be friendlier and useful.

Thanks for your in depth review! @shurcooL :)

chriskingnet:
Added the changes as requested in the feedback!

github/github.go
```diff
@@ -235,6 +235,37 @@ func NewClient(httpClient *http.Client) *Client {
...
 
+// NewEnterpriseClient returns a new GitHub API client with provided
+// base URL and upload URL (often the same URL).
+// If either URL does not have a trailing slash, one is added automatically.
+// If a nil httpClient is provided, http.DefaultClient will be used.
+//
+// Note that NewEnterpriseClient is a convenience helper only;
+// its behavior is equivalent to using NewClient, followed by setting
+// the BaseURL and UploadURL fields.
+func NewEnterpriseClient(httpClient *http.Client, baseEndpointStr string, uploadEndpointStr string) (*Client, error) {
```
shurcooL: When writing idiomatic Go code (see references at the end of this comment), it's common to think about how an identifier will appear in `godoc` and prioritize the readability of the documentation and signature, even if sometimes it requires making small sacrifices to the internal implementation details.

In this case, we have very nice and readable internal variable names `baseURL` and `uploadURL` of type `*url.URL`. Since the function parameter types are different, those names can't be used, so `baseEndpointStr` and `uploadEndpointStr` are used instead.

I suggest we flip that around, and use the nicer `baseURL` and `uploadURL` names in the function signature instead. First of all, it simply reads better. Second, it matches the fields `BaseURL` and `UploadURL` of `Client` struct, so it should be easier to understand they're related. Third, it matches what the documentation says more closely.

Another point, there is a related Go package for accessing the GitHub API v4 [`githubql`](https://github.com/shurcooL/githubql), and it has a similar function already:

https://godoc.org/github.com/shurcooL/githubql#NewEnterpriseClient

I think it'd be better to stay consistent with it and put the `baseURL, uploadURL` parameters first, `httpClient` last. Especially since `httpClient` can be left `nil`, it looks nicer if it's the at the end of the parameter list (but this isn't a big deal). If you have arguments for `httpClient` being better placed first, let me know. We can consider changing `githubql` to match.

So, this is the signature I'd suggest we should go with:

```Go
func NewEnterpriseClient(baseURL, uploadURL string, httpClient *http.Client) (*Client, error)
```

(You can also simplify `baseURL string, uploadURL string` to just `baseURL, uploadURL string`, since they're the same type.)

References:

- https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters
- https://golang.org/doc/effective_go.html#commentary

chriskingnet: Thanks for the detailed comment, this is a great way for me to learn. I completely agree with your points, also interesting about `godoc` and keeping it in mind when we write our APIs. This is something I will try and do in the future!

I have no strong arguments for keeping `httpClient` first, happy to change this and keep it consistent with the `githubql ` package.

Have added a commit to the branch with the requested changes. Thanks!

shurcooL: The parameter names still have previous names rather than `baseURL, uploadURL`. Are you planning on pushing that change in a separate commit?

chriskingnet: Sorry, you're right, I got ahead of myself there! Added another commit :)

github/github_test.go
```diff
@@ -186,6 +186,34 @@ func TestNewClient(t *testing.T) {
 	}
 }
 
+func TestNewEnterpriseClient(t *testing.T) {
+	baseURL := "https://custom-url/"
+	uploadURL := "https://custom-upload-url/"
+	c, _ := NewEnterpriseClient(baseURL, uploadURL, nil)
```
shurcooL: I think we should check for errors in tests. We expect it to be `nil`, let's ensure that's the case with:

```Go
c, err := NewEnterpriseClient(baseURL, uploadURL, nil)
if err != nil {
	t.Fatalf("NewEnterpriseClient returned unexpected error: %v", err)
}
```

Otherwise, it might not be `nil`, and tests wouldn't catch that.

github/github_test.go
```diff
@@ -186,6 +186,34 @@ func TestNewClient(t *testing.T) {
...
+	}
+}
+
+func TestNewEnterpriseClient_addsTrailingSlashToURLs(t *testing.T) {
+	baseURL := "https://custom-url"
+	uploadURL := "https://custom-upload-url"
+	formattedBaseURL := baseURL + "/"
+	formattedUploadURL := uploadURL + "/"
+
+	c, _ := NewEnterpriseClient(baseURL, uploadURL, nil)
```
shurcooL: Also here.

chriskingnet: Good suggestion, have included this check.

shurcooL:
Merged. Thank you @chriskingnet and @gmlewis (for reviewing)!

Issue #762, opened 2017-10-26
Unit tests should use local client instead of package-global client


gmlewis:
In https://github.com/google/go-github/pull/732 it became clear that having a package-global `client` for unit testing will prevent the unit tests from being run in parallel in the future.  Ideally, each test will have its own client... something like this:

```go
client, teardown := setup()
defer teardown()
```

and the global `client` in `github_test.go` will be removed.

This would make for a very large PR, but could possibly be done with the assistance of [`gofmt -r`](https://blog.golang.org/go-fmt-your-code) or your favorite text editor. :smile: 

This would also be a great project for a new contributor to the repo. Let it be known that all help is greatly appreciated! Thank you in advance.

chriskingnet:
I decided to have a go at this issue, sorry I forgot to comment here first. This was quite a nice one to do as a new contributor :)

shurcooL:
@gmlewis I don't think it'll be sufficient to have `setup` return only `client`. I think it'll need to return `server` and `mux` as well. Otherwise, it's not sufficient to achieve the goal of avoiding global shared state between tests.

chriskingnet:
Makes sense, thanks for input @shurcooL. From a quick check, there does actually seem to be a couple of tests using the global `server` in order to get its URL. 

So in order to not change any of the tests, we would have setup() always return the four things (I like explicitly returning teardown rather than just having every test call server.Close):

```go
client, mux, server, teardown := setup()
```

Or we could do something like (naming tbc):

```go
client, mux, globalConfig, teardown := setup()
```
where global config contains configuration details that might be needed in tests - i.e. the server base URL value. 

Any thoughts?

shurcooL:
How about this:

```Go
server, mux, client := setup()
defer server.Close()
```

Since everything that `teardown` does fits on one line, I think we can just get rid of it.

chriskingnet:
That's true. Although I quite liked that it is explicit, and that to "forget" to teardown I have to purposefully ignore the variable and use `_`, otherwise I will get an error for not using it. If we rely on calling server.Close(), I think that this step might be more easily missed.

That said, I imagine almost all tests written will just copy a previous one, which will have the code you mentioned in it. So in practice I'm not sure how much of a problem that is.

chriskingnet:
Ah, sorry, I replied before I saw your updated comment!

>Is the server base URL value the only thing used by tests? If so, maybe just return that url instead of globalConfig?

Yes, that sound like a good approach. So we're now at:

```go
[...]

PR #766, opened 2017-10-27
Removes global test client and calls setup for each client test

chriskingnet:
This is my attempt at removing the global client, as outlined in #762. 

As I am a new contributor to this lib, I thought this might be a good issue to have a go at as suggested - as you get to see quite a lot of the test code. Also put my find and replace skills to the test! :D

This seemed to work, all the test keep running and it was achieved I think pretty much exactly as outlined in the issue.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ again.
 
 [forking]: https://help.github.com/articles/fork-a-repo
 [golint]: https://github.com/golang/lint
-[golint readme]: https://github.com/golang/lint/blob/master/README
+[golint readme]: https://github.com/golang/lint/blob/master/README.md
 [gocov]: https://github.com/axw/gocov
 [gocov-html]: https://github.com/matm/gocov-html
 [well-formed commit messages]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ again.
      issue that you are planning to work on that bug or feature so that it can
      be assigned to you.
 
-  1. Follow the normal process of [forking][] the project, and setup a new
+  1. Follow the normal process of [forking][] the project, and setup() a new
      branch to work in. It's important that each group of changes be done in
      separate branches in order to ensure that a pull request only includes the
      commits related to that bug or feature.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ again.
      issue that you are planning to work on that bug or feature so that it can
      be assigned to you.
 
-  1. Follow the normal process of [forking][] the project, and a new
+  1. Follow the normal process of [forking][] the project, and setup a new
      branch to work in. It's important that each group of changes be done in
      separate branches in order to ensure that a pull request only includes the
      commits related to that bug or feature.
@@ -55,7 +55,7 @@ again.
 
 [forking]: https://help.github.com/articles/fork-a-repo
 [golint]: https://github.com/golang/lint
-[golint readme]: https://github.com/golang/lint/blob/master/README.md
+[golint readme]: https://github.com/golang/lint/blob/master/README
 [gocov]: https://github.com/axw/gocov
 [gocov-html]: https://github.com/matm/gocov-html
 [well-formed commit messages]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ again.
      issue that you are planning to work on that bug or feature so that it can
      be assigned to you.
 
-  1. Follow the normal process of [forking][] the project, and setup() a new
+  1. Follow the normal process of [forking][] the project, and a new
      branch to work in. It's important that each group of changes be done in
      separate branches in order to ensure that a pull request only includes the
      commits related to that bug or feature.
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -172,6 +172,41 @@ func TestNewClient(t *testing.T) {
 	}
 }
 
+func TestNewEnterpriseClient(t *testing.T) {
+	baseURL := "https://custom-url/"
+	uploadURL := "https://custom-upload-url/"
+	c, err := NewEnterpriseClient(baseURL, uploadURL, nil)
+	if err != nil {
+		t.Fatalf("NewEnterpriseClient returned unexpected error: %v", err)
+	}
+
+	if got, want := c.BaseURL.String(), baseURL; got != want {
+		t.Errorf("NewClient BaseURL is %v, want %v", got, want)
+	}
+	if got, want := c.UploadURL.String(), uploadURL; got != want {
+		t.Errorf("NewClient UploadURL is %v, want %v", got, want)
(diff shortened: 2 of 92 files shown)


shurcooL:
This is a great start @chriskingnet, thanks! But see https://github.com/google/go-github/issues/762#issuecomment-340133632.

chriskingnet:
Have made a few changes, based on the discussion from #762, and now also removed the `mux` and `server` objects from being package global.

CONTRIBUTING.md
```diff
@@ -25,7 +25,7 @@ again.
      issue that you are planning to work on that bug or feature so that it can
      be assigned to you.
 
-  1. Follow the normal process of [forking][] the project, and setup a new
+  1. Follow the normal process of [forking][] the project, and setup() a new
```
shurcooL: This change looks unintentional.

shurcooL:
On the the one hand, I think it'd be nice if this PR actually made use of the fact that tests are independent and made them parallel. That is the reason to be making these changes; if we don't make tests parallel, these changes have no positive effect. Also, making the tests run in parallel may uncover unexpected issues, ones that might result in us deciding to abort this change. It'd better to find that out sooner rather than later.

On the other hand, we should merge this PR sooner, because it'll keep getting merge conflicts when other PRs get merged.

What do you think we should aim for, @gmlewis and @chriskingnet?

shurcooL:
Ok, I see the original motivation for doing this in https://github.com/google/go-github/pull/732#discussion_r147207659...

So even if making tests in parallel won't work, it's still worth to make the tests not share globals. In that case, I'm okay with this PR not making tests run in parallel, that can be a separate PR.

CONTRIBUTING.md
chriskingnet: Thanks for spotting, I have fixed this now.

chriskingnet:
Thanks for the comments, @shurcooL. I agree with the two step approach here - first remove the dependencies on the global variables, and then do the work to parallelise the tests. 

I am not sure what is going on here, though, I have tried to merge and resolve the conflicts, but it is still reporting them. I must be doing something wrong.

github/github_test.go
```diff
@@ -42,9 +31,9 @@ const (
 // setup sets up a test HTTP server along with a github.Client that is
 // configured to talk to that test server. Tests should register handlers on
 // mux which provide mock responses for the API method being tested.
```
gmlewis: I would like to capture the removed comments from lines 26, 29, 32, and 73 into the description of `setup` so that it is clear what the purpose of the return values are.

CONTRIBUTING.md
```diff
@@ -25,7 +25,7 @@ again.
      issue that you are planning to work on that bug or feature so that it can
      be assigned to you.
 
-  1. Follow the normal process of [forking][] the project, and setup a new
+  1. Follow the normal process of [forking][] the project, and a new
```
gmlewis: Maybe revert all changes to CONTRIBUTING.md ?

github/github_test.go
chriskingnet: Thanks for spotting, @gmlewis. Have re-inserted in the new places they are built :)

chriskingnet:
Sorry for the delay on this. I have now merged master, fixed the newly added tests and reverted all comments that were caught in the find-replace.

github/github_test.go
```diff
@@ -42,9 +31,9 @@ const (
...
-func setup() {
-	// test server
-	mux = http.NewServeMux()
+func setup() (string, *http.ServeMux, *Client, func()) {
```
gmlewis: Thanks, @chriskingnet! I was actually thinking about putting the comments into the `setup` func comments immediately following line 44, but that's OK.  As an alternative, I propose the following...

@shurcooL - for documentation purposes, what do you think about using named return values here?
I personally think it would make the code (and godocs) easier to understand:

```go
func setup() (url string, mux *http.ServeMux, client *Client, teardown func()) {
```

shurcooL: Agreed, named returned values would help readability. Seeing `string` and `func()` doesn't say much.

chriskingnet:
Have named return values. Good idea, thanks @gmlewis!

github/github_test.go
chriskingnet: Thanks for the input on this, @shurcooL. Have now included this :)

github/repos_contents_test.go
```diff
@@ -91,14 +91,14 @@ func TestRepositoriesService_GetReadme(t *testing.T) {
 }
 
 func TestRepositoriesService_DownloadContents_Success(t *testing.T) {
-	setup()
+	serverBaseURL, mux, client, teardown := setup()
```
shurcooL: Minor, but I'd rename `serverBaseURL` to just `serverURL` for consistency with the return variable name in `setup`. At least I don't see a good reason to give it a different, more verbose name.

github/github_test.go
```diff
@@ -61,18 +50,17 @@ func setup() {
 		http.Error(w, "Client.BaseURL path prefix is not preserved in the request URL.", http.StatusInternalServerError)
 	})
 
-	server = httptest.NewServer(apiHandler)
+	// server is a test HTTP server used to provide mock API responses.
+	server := httptest.NewServer(apiHandler)
 
-	// github client configured to use test server
+	// client is the GitHub client being tested and is
+	// configured to use test server
```
shurcooL: Minor, missing period at end of sentence.

github/github_test.go
```diff
@@ -22,17 +22,6 @@ import (
...
-var (
-	// mux is the HTTP request multiplexer used with the test server.
-	mux *http.ServeMux
-
-	// client is the GitHub client being tested.
-	client *Client
-
-	// server is a test HTTP server used to provide mock API responses.
-	server *httptest.Server
-)
```
shurcooL: 👌

github/github_test.go
```diff
@@ -42,8 +31,8 @@ const (
 // setup sets up a test HTTP server along with a github.Client that is
 // configured to talk to that test server. Tests should register handlers on
 // mux which provide mock responses for the API method being tested.
-func setup() {
-	// test server
+func setup() (serverURL string, mux *http.ServeMux, client *Client, teardown func()) {
```
shurcooL: I've looked over the PR, and noticed that the vast majority of instances ignore `serverURL`. In fact, it's only used in 1 place.

`client` and teardown are the most often used parameters, used in all instances.

`mux` is used in most, but missing in about 20-30% of them.

Since `setup` is going to be called in so many places and in all future tests, let's optimize their order based on frequency of use. I think the following order would be better (most often used parameters first, except `teardown` is always last):

```Go
func setup() (client *Client, mux *http.ServeMux, serverURL string, teardown func())
```

I just think that's going to look nicer than the `_, mux, client, teardown := setup()` lines that we get now.

chriskingnet: Thanks for the feedback, @shurcooL. This is a very valid point, and something that is going to be used so often deserves this kind of depth of designing for sure. 

I will make this change and also merge in any new tests again. Unfortunately I am unable to work on this until next week - but I will endeavour to get it done then.

chriskingnet:
Hi @shurcooL, I have merged master and I think I have changed everything. Sorry it took a few days for me to get around from it. I'm pleased with how this looks now - this parameter ordering is much better. Let me know if there is anything else you spot :)

PR #770, opened 2017-11-01
gen-accessors: add support for named structs

### LINKED ISSUE #587: Support named structs in gen-accessors.go.
@gmlewis, I wanted to ask you a question regarding what you said in https://github.com/google/go-github/pull/577#discussion_r104304110:

> (Note that named structs themselves don't get accessors [intentionally]... there is no `GetUser` method, for example, because you can just do something like `resp.User.GetName()` for example and it will just work... even if `resp.User == nil`... does that answer the question?)

Did you take into account that the above won't work if accessing more than 1 field in a single expression?

For example, if the caller wants to get the name of user that created an issue, this will be okay as you mentioned:

```Go
name := issue.User.GetName()
```

Because of the `if u == nil { return "" }` check in `(u *User) GetName()`.

But this will panic:

```Go
planName := issue.User.Plan.GetName()

// panic: runtime error: invalid memory address or nil pointer dereference
```

Because `issue.User` is `(*github.User)(nil)`, and accessing `Plan` field of that causes a run-time panic:

> The following rules apply to selectors:
> [...]
> 5\. If `x` is of pointer type and has the value `nil` and `x.f` denotes a struct field, assigning to or evaluating `x.f` causes a run-time panic.

(Source: https://golang.org/ref/spec#Selectors.)

Did you take that into account when making the decision to omit the named structs? If so, what's the expected work around? If not, do we want to make a change, or is it fine?

This is just a question at this point, I don't have any suggestions for next steps yet.

sahildua2305:
Add support for automatically adding accessor methods for named structs
in github-accessors.go. Add a new flag to struct type Getter - NamedStruct
which is a flag to decide if this getter is for a named struct or not.
If it's a named struct, we make some changes in the way its
accessor method looks like finally in github-accessors.go.

Another thing to note here is that named structs are always going to end
up of type *ast.Ident from AST point of view so we don't need to check
for named structs for any other cases (like array/map/selector expression).

Fixes #587 
/cc @gmlewis

### DIFF
--- a/github/gen-accessors.go
+++ b/github/gen-accessors.go
@@ -155,7 +155,7 @@ func (t *templateData) dump() error {
 	return ioutil.WriteFile(t.filename, clean, 0644)
 }
 
-func newGetter(receiverType, fieldName, fieldType, zeroValue string) *getter {
+func newGetter(receiverType, fieldName, fieldType, zeroValue string, namedStruct bool) *getter {
 	return &getter{
 		sortVal:      strings.ToLower(receiverType) + "." + strings.ToLower(fieldName),
 		ReceiverVar:  strings.ToLower(receiverType[:1]),
@@ -163,6 +163,7 @@ func newGetter(receiverType, fieldName, fieldType, zeroValue string) *getter {
 		FieldName:    fieldName,
 		FieldType:    fieldType,
 		ZeroValue:    zeroValue,
+		NamedStruct:  namedStruct,
 	}
 }
 
@@ -176,11 +177,12 @@ func (t *templateData) addArrayType(x *ast.ArrayType, receiverType, fieldName st
 		return
 	}
 
-	t.Getters = append(t.Getters, newGetter(receiverType, fieldName, "[]"+eltType, "nil"))
+	t.Getters = append(t.Getters, newGetter(receiverType, fieldName, "[]"+eltType, "nil", false))
 }
 
 func (t *templateData) addIdent(x *ast.Ident, receiverType, fieldName string) {
 	var zeroValue string
+	var namedStruct = false
 	switch x.String() {
 	case "int":
 		zeroValue = "0"
@@ -190,11 +192,12 @@ func (t *templateData) addIdent(x *ast.Ident, receiverType, fieldName string) {
 		zeroValue = "false"
 	case "Timestamp":
 		zeroValue = "Timestamp{}"
-	default: // other structs handled by their receivers directly.
-		return
+	default:
+		zeroValue = `nil`
+		namedStruct = true
 	}
 
-	t.Getters = append(t.Getters, newGetter(receiverType, fieldName, x.String(), zeroValue))
+	t.Getters = append(t.Getters, newGetter(receiverType, fieldName, x.String(), zeroValue, namedStruct))
 }
 
 func (t *templateData) addMapType(x *ast.MapType, receiverType, fieldName string) {
@@ -218,7 +221,7 @@ func (t *templateData) addMapType(x *ast.MapType, receiverType, fieldName string
 
 	fieldType := fmt.Sprintf("map[%v]%v", keyType, valueType)
 	zeroValue := fmt.Sprintf("map[%v]%v{}", keyType, valueType)
-	t.Getters = append(t.Getters, newGetter(receiverType, fieldName, fieldType, zeroValue))
+	t.Getters = append(t.Getters, newGetter(receiverType, fieldName, fieldType, zeroValue, false))
 }
 
 func (t *templateData) addSelectorExpr(x *ast.SelectorExpr, receiverType, fieldName string) {
@@ -243,7 +246,7 @@ func (t *templateData) addSelectorExpr(x *ast.SelectorExpr, receiverType, fieldN
 		if xX == "time" && x.Sel.Name == "Duration" {
 			zeroValue = "0"
 		}
-		t.Getters = append(t.Getters, newGetter(receiverType, fieldName, fieldType, zeroValue))
+		t.Getters = append(t.Getters, newGetter(receiverType, fieldName, fieldType, zeroValue, false))
 	default:
 		logf("addSelectorExpr: xX %q, type %q, field %q: unknown x=%+v; skipping.", xX, receiverType, fieldName, x)
 	}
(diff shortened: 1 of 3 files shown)


github/gen-accessors.go
```diff
@@ -190,11 +192,12 @@ func (t *templateData) addIdent(x *ast.Ident, receiverType, fieldName string) {
 		zeroValue = "false"
 	case "Timestamp":
 		zeroValue = "Timestamp{}"
-	default: // other structs handled by their receivers directly.
-		return
+	default:
+		zeroValue = `nil`
```
shurcooL: Style nitpick, use an interpreted string literal `"nil"` here rather than raw string literal, for consistency with the code above.

shurcooL:
We could also add a test that doing something like `issue.GetUser().GetPlan().GetName()` doesn't panic, when `issue` is a `nil` `*github.Issue`.

gmlewis:
Uh, wow, that was fast, @sahildua2305! Thanks for doing that!
Yes, I like @shurcooL's idea of adding a no-panic test with a null receiver, and then we can squash and merget his PR.

sahildua2305:
Indeed, a good idea! @shurcooL 
Should I add a test to just one of the files like for `Issue` and leave all others as it is? @gmlewis

gmlewis:
Yes, I think one is sufficient, since they all follow the same pattern.  In fact, @shurcooL's example of `issue.GetUser().GetPlan().GetName()` sounds like a great one to test.

sahildua2305:
@gmlewis I'm not sure where should the test for `issue.GetUser().GetPlan().GetName()` go in the code. Should it go to `issues_test.go`? I don't see any other tests about accessors.

gmlewis:
How about a brand new `gen-accessors_test.go` ?

shurcooL:
We could also just add it somewhere in `github_test.go`. That file already contains many general go-github tests. Creating a new file `gen-accessors_test.go` seems like it's too much for just 1 little test. Besides, that name almost makes it sound like the entire file is generated, but it's not.

gmlewis:
OK, I'm fine with that.

sahildua2305:
@gmlewis @shurcooL just pushed a test in `github_test.go` to check the no-panic situation. 😄

github/github_test.go
```diff
@@ -1054,3 +1054,21 @@ func TestFormatRateReset(t *testing.T) {
...
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/issues/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeReactionsPreview)
+		fmt.Fprint(w, `{"number":1}`)
+	})
+
+	issue, _, _ := client.Issues.Get(context.Background(), "o", "r", 1)
```
shurcooL: Let’s simplify this test. It’s not necessary to test `Issues.Get` method here, we already have dedicated tests for that. The scope of this test is to ensure the chained accessor methods on a nil do not panic only. So, you can replace all of the code above with just 1 line:

```Go
issue := &Issue{User: nil}
```

sahildua2305: Great! Will do that.

sahildua2305:
@shurcooL I updated (rebased/squashed) my commit with the changes you suggested. Please review 😃

github/github_test.go
```diff
@@ -1054,3 +1054,12 @@ func TestFormatRateReset(t *testing.T) {
 		t.Errorf("Format is wrong. got: %v, want: %v", got, want)
 	}
 }
+
+func TestNamedStructAccessor_noPanic(t *testing.T) {
```
shurcooL: Optional suggestion, but `TestNestedStructAccessorNoPanic` might be a better name for the test. There's no need for an underscore, since the text before the underscore doesn't refer to any method or type.

I also added "Nested" in there since that's really the most important part of this test, the fact that _nested_ accessors work without panic. A single accessor wouldn't panic even before this change.

See https://godoc.org/testing#hdr-Examples for where the underscore naming convention comes from.

sahildua2305: Thanks @shurcooL! I have made the change.

> A single accessor wouldn't panic even before this change.

However, a single accessor for a named struct would give error that the method doesn't exist before this change. That was my idea behind the name. I understand with the convention however.

github/github_test.go
```diff
@@ -1054,3 +1054,12 @@ func TestFormatRateReset(t *testing.T) {
 		t.Errorf("Format is wrong. got: %v, want: %v", got, want)
 	}
 }
+
+func TestNestedStructAccessorNoPanic(t *testing.T) {
+	issue := &Issue{User: nil}
```
gmlewis: @shurcooL - what do you think about replacing this line with:

```go
var issue *Issue = nil
```

?  This is what I was envisioning in order to test the full pipeline of accessors.

gmlewis: Doh.  That will crash.  Never mind.  😄

sahildua2305:
PS: I'll fix the conflicts later just before the merging of this PR because this file will keep having conflicts.

github/github_test.go
sahildua2305: @gmlewis I see that you had commented the following but I don't see that comment now -
> Doh. That will crash. Never mind.

Is that still valid? The only benefit I see by doing it like this is that this way, we will be testing the `nil` case even for `GetUser()` and not just for `GetPlan()`.

shurcooL: > what do you think about replacing this line with `var issue *Issue = nil`

@gmlewis Both ways are equivalent. Either one will each give us full coverage.

gmlewis: Yeah, I'm fine either way, actually.  I think that "Doh. That will crash. Never mind." was incorrect... so I deleted the comment... I actually think my first comment is correct, in that it tests one more nil, but I'm fine keeping it as is.  Thanks, and sorry for the noise.  :smile:

sahildua2305: So, if we are good with this here, should I go ahead and resolve the merge conflicts in `github-accessors.go` so that we are good to merge?

gmlewis: SGTM.

sahildua2305: Cool! Resolved the conflicts. 👍

gmlewis:
Would you mind running gen-accessors so that this PR also checks in all the new goodness?

sahildua2305:
Good point, @gmlewis. Pushed.

gmlewis:
Great, thanks, @sahildua2305!
LGTM. Merging.

PR #771, opened 2017-11-01
Removing omitifempty on topics.Names to be able to remove all topics

jpbelanger-mtl:
Based on GitHub documentation:

> Required. An array of topics to add to the repository. Pass one or more topics to replace the set of existing topics. Send an empty array ([]) to clear all topics from the repository.

https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository 

right now if passing an empty array, the Names attribute is omitted resulting in API validation error on GitHub side.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -991,7 +991,7 @@ type Topics struct {
 // ListAllTopics lists topics for a repository.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/#list-all-topics-for-a-repository
-func (s *RepositoriesService) ListAllTopics(ctx context.Context, owner, repo string) (*Topics, *Response, error) {
+func (s *RepositoriesService) ListAllTopics(ctx context.Context, owner, repo string) ([]string, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/topics", owner, repo)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


shurcooL:
> right now if passing an empty array, the Names attribute is omitted resulting in API validation error on GitHub side.

I want to clarify, do you mean passing `nil`? Because I'd expect passing an actual empty slice to work successfully. I mean:

```Go
client.Repositories.ReplaceAllTopics(ctx, "o", "r", &Topics{Names: []string{}})
```

gmlewis:
Ah, good point, @shurcooL!

shurcooL:
Before merging this, let's come to a decision in #773, because the outcome there might influence what needs to be done here.

shurcooL:
We've come to a decision in #773. @jpbelanger-mtl, would you like to modify this PR to resolve that issue?

jpbelangerupgrade:
@shurcooL Pushed the changes for ReplaceAllTopics but also did the ListAllTopics so they have the same behavior. Also added tests for both.

github/repos.go
```diff
@@ -1007,27 +1007,31 @@ func (s *RepositoriesService) ListAllTopics(ctx context.Context, owner, repo str
...
 //
 // GitHub API docs: https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository
-func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo string, topics *Topics) (*Topics, *Response, error) {
+func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo string, topics []string) ([]string, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/topics", owner, repo)
-	req, err := s.client.NewRequest("PUT", u, topics)
+	t := new(Topics)
+	if t.Names = topics; topics == nil {
+		t.Names = []string{}
+	}
```
shurcooL: I think the above 4 lines can be expressed in a more readable way:

```Go
t := &Topics{
	Names: topics,
}
if t.Names == nil {
	t.Names = []string{}
}
```

jpbelanger-mtl: Removed ternary operator.

github/repos.go
```diff
@@ -985,13 +985,13 @@ func (s *RepositoriesService) RemoveAdminEnforcement(ctx context.Context, owner,
 
 // Topics represents a collection of repository topics.
 type Topics struct {
-	Names []string `json:"names,omitempty"`
+	Names []string `json:"names"`
 }
```
shurcooL: Since this type is replaced in favor of `[]string` in the exported method signatures, it's no longer used, and should be either made unexported, or removed (replaced by ad-hoc type definitions in the methods that need it).

github/repos.go
```diff
@@ -984,14 +984,14 @@ func (s *RepositoriesService) RemoveAdminEnforcement(ctx context.Context, owner,
...
 
 // Topics represents a collection of repository topics.
```
shurcooL: The type name in the comment should be updated to match the new type name:

```diff
-// Topics ...
+// repositoryTopics ...
type repositoryTopics struct {
```

gmlewis:
Merging.

PR #772, opened 2017-11-01
readme: add section on versioning

### LINKED ISSUE #376: versioning and tagging releases

willnorris:
This does not yet address some of the additional workflows and policies we'll need to adopt as a result.  For example, when do we cut new releases, do we attempt to maintain a formal changelog, or at least release notes when we cut a new release, etc.  This also requires that we update go docs to call out which methods are exposing preview functionality from the GitHub API.

Fixes #376

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ straightforward.
 In general, go-github follows [semver](http://semver.org/) as closely as we can
 for tagging releases of the package. For self-contained libraries, the
 application of semantic versioning is relatively straightforward and generally
-understood.  But because go-github is a client library for the GitHub API,
+understood. But because go-github is a client library for the GitHub API,
 which itself changes behavior, and because we are typically pretty aggressive
 about implementing preview features of the GitHub API, we've adopted the
 following versioning policy:
@@ -258,9 +258,8 @@ following versioning policy:
 * We increment the **patch version** with any backwards-compatible bug fixes.
 
 Preview functionality may take the form of entire methods or simply additional
-data returned from an otherwise non-preview method.  We do our best to call out
-which portions of a method's behavior and/or data are part of a GitHub Preview
-in the appropriate Go doc.
+data returned from an otherwise non-preview method. Refer to the GitHub API
+documentation for details on preview functionality.
 
 ## License ##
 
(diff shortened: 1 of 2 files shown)


gmlewis:
This sounds good to me.

One thing I'm left wondering is... what does `libraryVersion` now become?
Do we want to add a section in `CONTRIBUTING.md` that describes how contributors are expected to bump the `libraryVersion` on each PR? Or do we only use GitHub tagged releases for this new versioning?  (Or both?  ...probably not.)

This may all fall under "additional workflows and policies" that you called out in the description above, so maybe it can all be addressed later.

shurcooL:
Nit: The rest of README.md, and Go style at large uses 1 space between sentences, not two.

Everything I read sounds reasonable and SGTM, up until this part:

> We do our best to call out which portions of a method's behavior and/or data are part of a GitHub Preview in the appropriate Go doc.

I don't think we should do that, we should have people look that information up in GitHub documentation. I think it would cause too much churn to deal with in our Go package, and it'd end up being too inaccurate and out of date more often than not. It'd cause merge conflicts, delays in PRs being merged, etc. In the end, I see not nearly enough benefit, but huge downsides.

gmlewis:
Do you want to proactively add a tagged version *before* this PR gets merged and another one afterward?

README.md
```diff
@@ -238,6 +238,29 @@ straightforward.
 [roadmap]: https://docs.google.com/spreadsheet/ccc?key=0ApoVX4GOiXr-dGNKN1pObFh6ek1DR2FKUjBNZ1FmaEE&usp=sharing
 [contributing]: CONTRIBUTING.md
 
+## Versioning ##
+
+In general, go-github follows [semver](http://semver.org/) as closely as we can
```
shurcooL: s/http/https

willnorris: fixed.

willnorris:
I've created historical tags for v2.0.0 through v15.0.0, corresponding to each time the `libraryVersion` was bumped.  Going forward, we'll need to decide on a release schedule and tag the release using a version number that reflects whatever changes are included in that release.

PR #774, opened 2017-11-04
fix: RemoveReviewers json unmarshall error

ldez:
Use a pointer when execute the request.

```
error: json: Unmarshal(non-pointer github.ReviewersRequest)
```

I think the best way is to change method signature to use a pointer but it's breaking.
```go
func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo string, number int, reviewers *ReviewersRequest) (*PullRequest, *Response, error) {
```

### DIFF
--- a/github/pulls_reviewers.go
+++ b/github/pulls_reviewers.go
@@ -84,5 +84,5 @@ func (s *PullRequestsService) RemoveReviewers(ctx context.Context, owner, repo s
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeTeamReviewPreview)
 
-	return s.client.Do(ctx, req, reviewers)
+	return s.client.Do(ctx, req, nil)
 }


ldez:
I signed it!

github/pulls_reviewers.go
```diff
@@ -84,5 +84,5 @@ func (s *PullRequestsService) RemoveReviewers(ctx context.Context, owner, repo s
 	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeTeamReviewPreview)
 
-	return s.client.Do(ctx, req, reviewers)
+	return s.client.Do(ctx, req, &reviewers)
```
shurcooL: I believe this should be:

```Go
```

`reviewers` is only used in the request body, the response from [this endpoint](https://developer.github.com/v3/pulls/review_requests/#delete-a-review-request) does not return anything, just 200 OK.

See a similar endpoint like `UsersService.DeleteEmails`:

https://github.com/google/go-github/blob/464f5a91396a85da262abcac0b0a0aaf5b8f3c02/github/users_emails.go#L63-L71

ldez: You're right 👍

PR #779, opened 2017-11-15
Bug fix PUT /repos/:owner/:repo/branches/:branch/protection

mspiegel:
Empty dismissal restrictions should be a JSON object not a JSON array. See https://developer.github.com/v3/repos/branches/#update-branch-protection. I think this is pretty straightforward? Providing a JSON array for the 'dismissal_restrictions' the GitHub API returns an error response.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -583,11 +583,11 @@ type PullRequestReviewsEnforcementRequest struct {
 func (req PullRequestReviewsEnforcementRequest) MarshalJSON() ([]byte, error) {
 	if req.DismissalRestrictionsRequest == nil {
 		newReq := struct {
-			R []interface{} `json:"dismissal_restrictions"`
-			D bool          `json:"dismiss_stale_reviews"`
-			O bool          `json:"require_code_owner_reviews"`
+			R map[string]string `json:"dismissal_restrictions"`
+			D bool              `json:"dismiss_stale_reviews"`
+			O bool              `json:"require_code_owner_reviews"`
 		}{
-			R: []interface{}{},
+			R: make(map[string]string),
 			D: req.DismissStaleReviews,
 			O: req.RequireCodeOwnerReviews,
 		}
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -929,7 +929,7 @@ func TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctio
 		t.Errorf("PullRequestReviewsEnforcementRequest.MarshalJSON returned error: %v", err)
 	}
 
-	want := `{"dismissal_restrictions":[],"dismiss_stale_reviews":false,"require_code_owner_reviews":false}`
+	want := `{"dismissal_restrictions":{},"dismiss_stale_reviews":false,"require_code_owner_reviews":false}`
 	if want != string(json) {
 		t.Errorf("PullRequestReviewsEnforcementRequest.MarshalJSON returned %+v, want %+v", string(json), want)
 	}


shurcooL:
This is closely related to #664 (the issue tracking this) and #703 (an attempt to resolve this in the past, containing relevant information).

> I think this is pretty straightforward?

This issue has complicated history, but it looks like GitHub has finally updated the documentation, which will make it possible (and much easier) to resolve #664.

In #703, on Aug 22, I noted that:

> At https://developer.github.com/v3/repos/branches/#update-branch-protection, it still says:
>
> > You can disable dismissal restrictions by passing the dismissal_restrictions object as an empty array.

Today, it no longer says that there.

github/repos.go
```diff
@@ -583,11 +583,11 @@ type PullRequestReviewsEnforcementRequest struct {
 func (req PullRequestReviewsEnforcementRequest) MarshalJSON() ([]byte, error) {
```
shurcooL: Now that it's no longer neccessary to send empty arrays to GitHub, this entire `MarshalJSON` method and a lot of other related code can be deleted. The end result will be much simpler.

I suggest looking at #703 for more information. That PR was on the right track, we just couldn't review it property at the time because GitHub hadn't updated their documentation back then.

mspiegel:
I was experimenting with the GitHub API a few hours ago. If I recall correctly the API requires sending an empty JSON object. A 400 response is returned if you: use a null value for the key; or if you use a value that's a JSON object with 'users' and 'keys' both set to null. So I think the MarshalJSON is still needed.

I'll take a fresh look at it tomorrow.

shurcooL:
> the API requires sending an empty JSON object

If so, it might be possible to achieve in a simpler way by removing the `,omitempty` option.

> I'll take a fresh look at it tomorrow.

👍

mspiegel:
Also if you omit the key 'dismissal_restrictions' the response is a 4xx error. At least for org repositories.

mspiegel:
I think the omitempty strategy won't work. If we want to disable the dismissal_restrictions then the strategy works because it generates an empty JSON object (that's neat I like the approach). The problem is that if dismissal_restrictions has either 'users' or 'teams' then it *must* have both 'users' and 'teams'. The GitHub API will barf if one of those keys is missing.

We have a separate problem which is that for personal repositories the 'dismissal_restrictions' key must be absent. It's not allowed to be an empty JSON object. I think one way to fix this is to change the implementation of UpdateBranchProtection() to first make a call to RepositoriesService.Get() to determine whether the repository belongs to an organization. It's a hack but I think it's necessary?

shurcooL:
> I think the omitempty strategy won't work. If we want to disable the dismissal_restrictions then the strategy works because it generates an empty JSON object (that's neat I like the approach). The problem is that if dismissal_restrictions has either 'users' or 'teams' then it must have both 'users' and 'teams'. The GitHub API will barf if one of those keys is missing.

Fair enough. If it can't be made simpler, then it's fine as is. Also, if we do find a way to make it simpler later, we can always apply it then. First priority is to get to be correct and functional.

> We have a separate problem which is that for personal repositories the 'dismissal_restrictions' key must be absent. It's not allowed to be an empty JSON object. I think one way to fix this is to change the implementation of UpdateBranchProtection() to first make a call to RepositoriesService.Get() to determine whether the repository belongs to an organization. It's a hack but I think it's necessary?

Hmm. I think we need to approach this differently. It should be a parameter that the user supplies. In general, each method of `go-github` is meant to map to a single API call. This is not a high level library, it's a low level wrapper around GitHub REST V3 API, so multiple API calls are not allowed.

mspiegel:
I'm comfortable with the approach of adding a parameter. Wasn't sure if it
was ok to make a breaking change.

On Nov 15, 2017 3:38 PM, "Dmitri Shuralyov" <notifications@github.com>
wrote:

I think the omitempty strategy won't work. If we want to disable the
dismissal_restrictions then the strategy works because it generates an
empty JSON object (that's neat I like the approach). The problem is that if
dismissal_restrictions has either 'users' or 'teams' then it must have both
'users' and 'teams'. The GitHub API will barf if one of those keys is
missing.

Fair enough. If it can't be made simpler, then it's fine as is. Also, if we
do find a way to make it simpler later, we can always apply it then. First
priority is to get to be correct and functional.

We have a separate problem which is that for personal repositories the
'dismissal_restrictions' key must be absent. It's not allowed to be an
empty JSON object. I think one way to fix this is to change the
implementation of UpdateBranchProtection() to first make a call to
RepositoriesService.Get() to determine whether the repository belongs to an
organization. It's a hack but I think it's necessary?

Hmm. I think we need to approach this differently. It should be a parameter
that the user supplies. In general, each method of go-github is meant to
map to a single API call. This is not a high level library, it's a low
level wrapper around GitHub REST V3 API, so multiple API calls are not
allowed.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/google/go-github/pull/779#issuecomment-344720825>, or mute
the thread

mspiegel:
The bug fix for personal repos I think will involve a change to the MarshalJSON() function because we need to omit a key only for personal repositories (this is the worst GitHub API ever). So should we add a boolean field 'Personal' to the PullRequestReviewsEnforcementRequest struct with `json:"-"` and then use the new field in the MarshalJSON() implementation? Alternatively we could name the field 'Org'. If we name if 'Personal' then the default value false is what I think most use cases will want. But that's subjective I happen to use this endpoint predominantly for org repositories.

shurcooL:
> I think the omitempty strategy won't work. If we want to disable the dismissal_restrictions then the strategy works because it generates an empty JSON object (that's neat I like the approach). The problem is that if dismissal_restrictions has either 'users' or 'teams' then it *must* have both 'users' and 'teams'. The GitHub API will barf if one of those keys is missing.
>
> We have a separate problem which is that for personal repositories the 'dismissal_restrictions' key must be absent. It's not allowed to be an empty JSON object. I think one way to fix this is to change the implementation of UpdateBranchProtection() to first make a call to RepositoriesService.Get() to determine whether the repository belongs to an organization. It's a hack but I think it's necessary?

After your last comment, I re-read the above, and https://developer.github.com/v3/repos/branches/#update-branch-protection, and started to wonder the following...

Isn't it now possible to create a very simple mapping of Go structs to JSON parameters of the GitHub API endpoint? I am not 100% certain, but my understanding the reason `PullRequestReviewsEnforcementRequest.MarshalJSON` came up was because of the unusual previous behavior/documentation where it said that an empty _array_–rather than object–should be used. I highly suspect it was a typo in the documentation, that lead us astray. Mapping it to an object makes much more sense and should work with normal `encoding/json` mapping rules without the need for custom encoding behaviors.

Basically, what I'm proposing is creating a simple mapping to the GitHub API as it is documented at https://developer.github.com/v3/repos/branches/#update-branch-protection, and letting the user take care of setting the right parameters that satisfy the GitHub API requirements:

image

In other words, `go-github` is a low level wrapper around the GitHub API, it should map to the GitHub API as closely as possible. We don't need to create a higher level API on top of it, if we don't have to.

Does that make sense? Do you think it's possible to simplify the solution as a result?

shurcooL:
I've prototyped how it can look like (this is not necessarily the final version and can be adjusted further) for dismissal_restrictions, which has one of 3 modes: empty JSON object to disable, JSON object containing both users and teams for org repos, and omit parameter for personal repos:

image

See https://play.golang.org/p/ZwKhQxLNan.

mspiegel:
I like the approach but the proposal doesn't quite work. If "users" is populated and "teams" is empty (or vice versa) the GitHub API requires that both keys be present in the JSON request body. It's quite a terrible API. See https://play.golang.org/p/MBdTjDvL7h for an example.

shurcooL:
Good catch. I think we can still work around it. The slices will need to be changed to pointers to slices, keeping `omitempty`. Then `nil` means omit, but empty slice is preserved.

See https://play.golang.org/p/kaeh7FbDb7.

mspiegel:
Yes! This works. I can close this pull request

mspiegel:
Should I submit a new pull request? Or modify this one?

shurcooL:
Since the planned approach has changed significantly, sending a new PR would be better. Thanks!

PR #782, opened 2017-11-15
repos: add archived boolean field to Repository struct

### LINKED ISSUE #776: Support new API features for Archived Repositories
GitHub Developer API Announcement:
https://developer.github.com/changes/2017-11-08-archiving-repositories/

This issue consists of one main change (along with updates to the comments):

    *   Add the new `archived bool` to `Repository` responses.

Since the `Repository` struct is used both in the responses and in the [`RepositoriesService.Edit` method](https://github.com/google/go-github/blob/master/github/repos.go#L359), 
this will also enable API users to archive a repo through this client library by setting `archived: true`.

Note that the `libraryVersion` will not need changing as these minor changes are backward-compatible.

This is a great issue for a new contributor to work on since the scope is small. Check out the [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) doc to get started. Let me know if you would like to be assigned to this issue (by commenting below) and I will send you an invite to contribute to this repo that you will then need to accept.

Thank you for your contributions!  All assistance is greatly appreciated!

sahildua2305:
Add Archived flag to Repository struct as per changes announced by GitHub
- https://developer.github.com/changes/2017-11-08-archiving-repositories/

Fixes #776 
/cc @gmlewis @shurcooL

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -68,6 +68,7 @@ type Repository struct {
 	HasDownloads      *bool   `json:"has_downloads"`
 	LicenseTemplate   *string `json:"license_template,omitempty"`
 	GitignoreTemplate *string `json:"gitignore_template,omitempty"`
+	Archived          *bool   `json:"archived,omitempty"`
 
 	// Creating an organization repository. Required for non-owners.
 	TeamID *int `json:"team_id"`
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -7412,6 +7412,14 @@ func (r *Repository) GetAllowSquashMerge() bool {
 	return *r.AllowSquashMerge
 }
 
+// GetArchived returns the Archived field if it's non-nil, zero value otherwise.
+func (r *Repository) GetArchived() bool {
+	if r == nil || r.Archived == nil {
+		return false
+	}
+	return *r.Archived
+}
+
 // GetArchiveURL returns the ArchiveURL field if it's non-nil, zero value otherwise.
 func (r *Repository) GetArchiveURL() string {
 	if r == nil || r.ArchiveURL == nil {


github/repos.go
```diff
@@ -68,6 +68,7 @@ type Repository struct {
 	HasDownloads      *bool   `json:"has_downloads"`
 	LicenseTemplate   *string `json:"license_template,omitempty"`
 	GitignoreTemplate *string `json:"gitignore_template,omitempty"`
+	Archived          *bool   `json:"archived"`
```
shurcooL: Why doesn't this field have `,omitempty` like the other fields above?

sahildua2305: Honestly, I wasn't sure about that and got confused by looking at the existing fields of type `*bool` some of which have `,omitempty` while others don't have. What's the logic/reasoning behind that?

shurcooL: No worries.

> got confused by looking at the existing fields (in `Repository`) of type `*bool` some of which have `,omitempty` while others don't.

The reason is very relevant. It's primarily to have consistency, so other people don't run into such situation where it's unclear which of the two styles to use. As far as I can tell, it's much more common to use `,omitempty` on pointer fields than not. We should fix the other fields too (probably in a separate issue/PR).

There's very little effect of using/not using `,omitempty` here. Including it makes it so that fields with `nil` pointer values get omitted if the struct happens to get JSON marshaled.

sahildua2305: Thanks! Updated my commit. Please review.

gmlewis:
Thank you, @sahildua2305 and @shurcooL!
LGTM.
Merging.

Issue #786, opened 2017-11-16
Add `omitempty` to pointer type fields' json schema


sahildua2305:
As per https://github.com/google/go-github/pull/782#discussion_r151298759, we should add `,omitempty` to all pointer fields so as to make things consistent and not making new contributors confused about whether to add it or not to new fields.

sahildua2305:
@shurcooL @gmlewis feel free to assign it to me. 😸

PR #787, opened 2017-11-16
Add missing omitempty to the field json schema

sahildua2305:
Adds missing ",omitempty" to the field json schema of the pointer
fields.

Fixes: #786

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -39,7 +39,7 @@ type Repository struct {
 	SSHURL           *string          `json:"ssh_url,omitempty"`
 	SVNURL           *string          `json:"svn_url,omitempty"`
 	Language         *string          `json:"language,omitempty"`
-	Fork             *bool            `json:"fork"`
+	Fork             *bool            `json:"fork,omitempty"`
 	ForksCount       *int             `json:"forks_count,omitempty"`
 	NetworkCount     *int             `json:"network_count,omitempty"`
 	OpenIssuesCount  *int             `json:"open_issues_count,omitempty"`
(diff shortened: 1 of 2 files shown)


sahildua2305:
This is work in progress. I'm looking for more files to fix using the following regex in the entire package.
```
\*[a-z]+\s+`json:"[a-z_]+"`
```

sahildua2305:
@shurcooL this is ready for a review.

sahildua2305:
@shurcooL do we need any changes here?

shurcooL:
> do we need any changes here?

I'll let you know when I get to reviewing this PR.

Thanks for your work, I'll get to it soon.

github/repos.go
```diff
@@ -61,18 +61,18 @@ type Repository struct {
...
-	HasIssues         *bool   `json:"has_issues"`
-	HasWiki           *bool   `json:"has_wiki"`
-	HasPages          *bool   `json:"has_pages"`
+	Private           *bool   `json:"private,omitempty"`
+	HasIssues         *bool   `json:"has_issues,omitempty"`
+	HasWiki           *bool   `json:"has_wiki,omitempty"`
+	HasPages          *bool   `json:"has_pages,omitempty"`
+	HasDownloads      *bool   `json:"has_downloads,omitempty"`
 	HasProjects       *bool   `json:"has_projects,omitempty"`
-	HasDownloads      *bool   `json:"has_downloads"`
```
shurcooL: What's the reason you re-arranged the order of the fields?

It's not alphabetical, and I can't quite see the pattern. If there's a good reason, that's fine. If not, it'd be better to keep their order unchanged and focus this PR on just adding `,omitempty`s.

sahildua2305: No reason. That's wrong, indeed. I mistakenly changed the order of fields while resolving the merge conflict with that new field (`HasProjects`).

shurcooL:
Merged. Please let us know if any issues come up!

sahildua2305:
@shurcooL even I was surprised to know that. Good though :D

Issue #773, opened 2017-11-01
RepositoriesService.ReplaceAllTopics has *Topics parameter, it'd be better made a value.


shurcooL:
I just noticed the signature of [`RepositoriesService.ReplaceAllTopics`](https://godoc.org/github.com/google/go-github/github#RepositoriesService.ReplaceAllTopics) has a `*Topic` parameter.

The parameter is not optional, it's mandatory. See https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository:

![image](https://user-images.githubusercontent.com/1924134/32302727-9388673a-bf3a-11e7-8caa-771a70b31af0.png)

There's no value in making it a pointer rather than value, only harm, because it makes it possible to pass `nil` value.

I suggest we making a breaking API change to improve the API of `go-github` and change the signature of the method to:

```Go
func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo string,
    topics Topics) (*Topics, *Response, error)
```

This is a new method, still under a preview API, so a breaking change is acceptable IMO, as long as it leads to an improve `go-github` API.

/cc @gmlewis

shurcooL:
Actually, better yet, we can also consider making it a `topics []string` instead:

```Go
func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo string,
    topics []string) (*Topics, *Response, error)
```

shurcooL:
If we go with `topics []string`, then #771 becomes no longer neccessary, since `Topics` won't necessarily be sent. It might instead be:

```Go
if topics == nil {
	topics = []string{}
}
t := struct {
	Names []string `json:"names"`
}{Names: topics}
resp, err := s.client.Do(ctx, req, t)
```

jpbelangerupgrade:
I do like the simplicity of passing `[]string` instead of the `*Topics` struct which in the end only contains the string array. Also since this is a preview API, who knows what changes could be made by github on this, more breaking change could be coming anyway.

jpbelanger-mtl:
Just let me know what you decide. Breaking API change is not a deal breaker on this, still a very simple preview feature from github.

gmlewis:
Yes, I very much like changing the API to use `topics []string` in the parameter list, as [suggested by @shurcooL above](https://github.com/google/go-github/issues/773#issuecomment-341273316).

Thank you, @shurcooL and @jpbelanger-mtl!

sahildua2305:
[...]

PR #788, opened 2017-11-16
Add support for the repository transfer API

### LINKED ISSUE #777: Support new Repository Transfer API
GitHub Developers API announcement:
https://developer.github.com/changes/2017-11-09-repository-transfer-api-preview/

GitHub v3 API developer docs:
https://developer.github.com/v3/repos/#transfer-a-repository

This would be a great project for a new contributor to this repo!
The new endpoint should fit nicely into one of the existing source files, possibly `repos.go` unless a better alternative is found and we can discuss it.

All contributors and contributions are greatly appreciated. Thank you for your help in keeping this GitHub API client library for Go up-to-date!

juliaferraioli:
As announced here:
https://developer.github.com/changes/2017-11-09-repository-transfer-api-preview/

And detailed here:
https://developer.github.com/v3/repos/#transfer-a-repository

Fixes #777

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -1053,10 +1053,10 @@ type TransferRequest struct {
 // in a successful request.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/#transfer-a-repository
-func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string, transfer *TransferRequest) (*Repository, *Response, error) {
+func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string, transfer TransferRequest) (*Repository, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/transfer", owner, repo)
 
-	req, err := s.client.NewRequest("POST", u, transfer)
+	req, err := s.client.NewRequest("POST", u, &transfer)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/repos.go
+++ b/github/repos.go
@@ -1037,3 +1037,38 @@ func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo 
 
 	return t.Names, resp, nil
 }
+
+// TransferRequest represents a request to transfer a repository.
+type TransferRequest struct {
+	NewOwner string `json:"new_owner,omitempty"`
+	TeamID   []int  `json:"team_id,omitempty"`
+}
+
+// Transfer transfers a repository from one account or organization to another.
+//
+// This method might return an *AcceptedError and a status code of
+// 202. This is because this is the status that GitHub returns to signify that
+// it has now scheduled the transfer of the repository in a background task.
+// A follow up request, after a delay of a second or so, should result
+// in a successful request.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/#transfer-a-repository
+func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string, transfer *TransferRequest) (*Repository, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/transfer", owner, repo)
+
+	req, err := s.client.NewRequest("POST", u, transfer)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeRepositoryTransferPreview)
+
+	r := new(Repository)
+	resp, err := s.client.Do(ctx, req, r)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, nil
+}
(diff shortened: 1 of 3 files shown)


github/repos.go
```diff
@@ -1037,3 +1037,38 @@ func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo
 
 	return t.Names, resp, nil
 }
+
+// TransferRequest represents a request to transfer a repository.
+type TransferRequest struct {
+	NewOwner string `json:"new_owner,omitempty"`
```
shurcooL: According to GitHub API docs, `new_owner` is required, so I don't think `omitempty` is appropriate here.

github/repos.go
```diff
@@ -1037,3 +1037,38 @@ func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo
...
+// Transfer transfers a repository from one account or organization to another.
+//
+// This method might return an *AcceptedError and a status code of
+// 202. This is because this is the status that GitHub returns to signify that
+// it has now scheduled the transfer of the repository in a background task.
+// A follow up request, after a delay of a second or so, should result
+// in a successful request.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/#transfer-a-repository
+func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string, transfer *TransferRequest) (*Repository, *Response, error) {
```
shurcooL: Why is `transfer` a `*TransferRequest` pointer rather than value?

That means user can pass `nil`, but that wouldn't be meaningful, would it?

github/repos.go
juliaferraioli: Excellent catch. Fixed.

github/repos.go
juliaferraioli: I understand your point here. And while I tend to agree, there's precedent for this approach even in repos.create being passed a pointer to a Repo. I didn't want to introduce code that broke with that model.

Thoughts?

shurcooL: > there's precedent for this approach even in repos.create being passed a pointer to a Repo. I didn't want to introduce code that broke with that model.

I see. However, that API decision was made [5 years ago](https://github.com/google/go-github/commit/d3d9655fad081711f66546dbd439814a8ef1d5c7#diff-49ba3de83fa99f3d8bae2ece519a449fR153), and is not very representative of the best practices regarding API decisions this repo has evolved toward since that time.

In fact, we've ran into possible panics in the library caused due to such parameters being passed as pointers in situations where nil values made no sense. We dealt with that in #531 and #539, by adding `if param == nil { return errors.New(...) }` guards.

We went that route because it was less disruptive than breaking an API for a very minor improvement. Given two equivalent choices, we prefer consistency.

However, in this case, it's a new API, so we can go with the best API design we can come up with. Using a pointer isn't meaningful and doesn't add value (does it?), but it create a liability for dealing with users potentially passing `nil` values.

I understand your point about wanting to be consistent with existing code, but in this case, I don't think the consistency argument applies, because using a pointer has tangible downsides compared to using a value. They're not equivalent choices. When adding new APIs to this library, we try to follow the best known practices, and we sometimes even break APIs of preview endpoints when improvements are discovered.

If you want to see more recent precedent, see #773 which came up during review of #771. The original version started by using pointers, but we came to agreement that value parameters were better.

(One could argue that `TransferRequest` contains just 2 fields that could be inlined into the method signature. However, I think using a dedicated struct for it as you've done makes sense, because it maps more closely to GitHub API as [documented](https://developer.github.com/v3/repos/#input-3), and will be easy to update should GitHub add extra fields to that input object.)

What do you think?

juliaferraioli: That sounds reasonable to me. Thanks for the thorough background! I'll go ahead and work on the updates today.

I had thought about your parenthetical initially, but decided against it for exactly the reasons you enumerated.

github/repos.go
```diff
@@ -1037,3 +1037,38 @@ func (s *RepositoriesService) ReplaceAllTopics(ctx context.Context, owner, repo
 
 	return t.Names, resp, nil
 }
+
+// TransferRequest represents a request to transfer a repository.
+type TransferRequest struct {
+	NewOwner string `json:"new_owner"`
+	TeamID   []int  `json:"team_id,omitempty"`
```
shurcooL: I just noticed now, I think `json:"team_id,omitempty"` should've been `json:"team_ids,omitempty"`, isn't that right?

From https://developer.github.com/v3/repos/#transfer-a-repository:

image

(FWIW, https://developer.github.com/v3/orgs/members/#create-organization-invitation also uses `team_ids`.)

gmlewis: Good catch, @shurcooL!

gmlewis: Thank you, @juliaferraioli! No rush... this is not urgent.

PR #785, opened 2017-11-16
Support MarketplacePurchaseEvent support

scriptonist:
this PR is intended to solve #704

PR #789, opened 2017-11-19
Support MarketeplacePurchaseEvent

scriptonist:
Fixes #785

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -274,11 +274,14 @@ type LabelEvent struct {
 //
 // Github API docs: https://developer.github.com/v3/activity/events/types/#marketplacepurchaseevent
 type MarketplacePurchaseEvent struct {
-	Action                      *string              `json:"action"`
+	Action *string `json:"action,omitempty"`
+
+	// The following fields are only populated by Webhook events.
 	EffectiveDate               *Timestamp           `json:"effective_date"`
 	MarketplacePurchase         *MarketplacePurchase `json:"marketplace_purchase"`
 	PreviousMarketplacePurchase *MarketplacePurchase `json:"previous_marketplace_purchase"`
 	Sender                      *User                `json:"sender,omitempty"`
+	Installation                *Installation        `json:"installation,omitempty"`
 }
 
 // MemberEvent is triggered when a user is added as a collaborator to a repository.
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -277,9 +277,9 @@ type MarketplacePurchaseEvent struct {
 	Action *string `json:"action,omitempty"`
 
 	// The following fields are only populated by Webhook events.
-	EffectiveDate               *Timestamp           `json:"effective_date"`
-	MarketplacePurchase         *MarketplacePurchase `json:"marketplace_purchase"`
-	PreviousMarketplacePurchase *MarketplacePurchase `json:"previous_marketplace_purchase"`
+	EffectiveDate               *Timestamp           `json:"effective_date,omitempty"`
+	MarketplacePurchase         *MarketplacePurchase `json:"marketplace_purchase,omitempty"`
+	PreviousMarketplacePurchase *MarketplacePurchase `json:"previous_marketplace_purchase,omitempty"`
 	Sender                      *User                `json:"sender,omitempty"`
 	Installation                *Installation        `json:"installation,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


github/event_types.go
```diff
@@ -268,6 +270,19 @@ type LabelEvent struct {
...
 }
 
+// MarketplacePurchaseEvent is triggered when a user purchases, cancels, or changes
+// their GitHub Marketplace plan.
+// Webhook event name "marketplace_purchase".
+//
+// Github API docs: https://developer.github.com/v3/activity/events/types/#marketplacepurchaseevent
+type MarketplacePurchaseEvent struct {
+	Action                      *string              `json:"action"`
+	EffectiveDate               *time.Time           `json:"effective_date"`
```
shurcooL: We don't use `time.Time` directly in go-github, we have a custom [`Timestamp`](https://godoc.org/github.com/google/go-github/github#Timestamp) type defined that handles some inconsistencies in time formatting on GitHub API side:

 We should use it here, primarily for consistency. For example, see:

https://github.com/google/go-github/blob/471e67c93cdd36a5f7c9bba5d6d11665ba12147d/github/event_types.go#L583-L585

https://github.com/google/go-github/blob/471e67c93cdd36a5f7c9bba5d6d11665ba12147d/github/event_types.go#L670-L671

github/event_types.go
```diff
@@ -268,6 +270,19 @@ type LabelEvent struct {
...
+	MarketplacePurchase         *MarketplacePurchase `json:"marketplace_purchase"`
+	PreviousMarketplacePurchase *MarketplacePurchase `json:"previous_marketplace_purchase"`
+	Sender                      *User                `json:"sender,omitempty"`
+}
```
shurcooL: You'll need to break the fields into two sections. The event API payload fields (of which, there's only one, `action`):

image

And Webhook payload fields, which are all the rest. You should put them under a `// The following fields are only populated by Webhook events.` comment. Something like this:

```Go
type MarketplacePurchaseEvent struct {
	// Action is the action that was performed. Possible values are:
	// "purchased", "cancelled", or "changed".

	...
}
```

See structs right above and below the one you added for reference.

shurcooL: Additionally, you'll need to add an implicit `Installation *Installation `json:"installation,omitempty"` field. Its existence is documented at https://developer.github.com/webhooks/#payloads. Put it at the very bottom; see other events in this file as reference.

scriptonist: Okay @shurcooL, will fix these. thanks :+1:  .

github/event_types.go
```diff
@@ -268,6 +268,22 @@ type LabelEvent struct {
...
+// Webhook event name "marketplace_purchase".
+//
+// Github API docs: https://developer.github.com/v3/activity/events/types/#marketplacepurchaseevent
+type MarketplacePurchaseEvent struct {
+	Action *string `json:"action,omitempty"`
+
+	// The following fields are only populated by Webhook events.
+	EffectiveDate               *Timestamp           `json:"effective_date"`
+	MarketplacePurchase         *MarketplacePurchase `json:"marketplace_purchase"`
+	PreviousMarketplacePurchase *MarketplacePurchase `json:"previous_marketplace_purchase"`
```
shurcooL: The 3 fields above are missing `,omitempty` suffix. The rest of the fields have it.

You should add them for consistency (unless there's a good reason not to include them; I don't see one).

shurcooL:
The CI is failing on the latest commit. It looks like you'll need to regenerate the package with `go generate` again.

scriptonist:
@shurcooL It looks good right ?

github/event_types.go
```diff
@@ -268,6 +268,22 @@ type LabelEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// MarketplacePurchaseEvent is triggered when a user purchases, cancels, or changes
+// their GitHub Marketplace plan.
+// Webhook event name "marketplace_purchase".
+//
+// Github API docs: https://developer.github.com/v3/activity/events/types/#marketplacepurchaseevent
+type MarketplacePurchaseEvent struct {
+	Action *string `json:"action,omitempty"`
```
shurcooL: It'd be nice to add a comment here documenting possible action values, like in the events above and below.

```
// Action is the action that was performed. Possible values are:
// "purchased", "cancelled", "changed".
```

PR #791, opened 2017-11-23
Bug fix PUT /repos/:owner/:repo/branches/:branch/protection

### LINKED ISSUE #664: PullRequestReviewsEnforcementRequest.DismissalRestrictionsRequest set to nil results in 422 from GH

mspiegel:
The 'dismissal_restrictions' key in the 'required_pull_request_reviews' object must either be omitted (for the case of personal repositories), an empty JSON object, or a JSON object with exactly two keys 'users' and 'teams'. See https://developer.github.com/v3/repos/branches/#update-branch-protection. Closes #664

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -571,7 +571,9 @@ type PullRequestReviewsEnforcement struct {
 // enforcement of a protected branch. It is separate from PullRequestReviewsEnforcement above
 // because the request structure is different from the response structure.
 type PullRequestReviewsEnforcementRequest struct {
-	// Specifies which users and teams should be allowed to dismiss pull request reviews. Can be nil to disable the restrictions.
+	// Specifies which users and teams should be allowed to dismiss pull request reviews.
+	// User and team dismissal restrictions are only available for
+	// organization-owned repositories. Must be nil for personal repositories.
 	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
 	// Specifies if approved reviews can be dismissed automatically, when a new commit is pushed. (Required)
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
@@ -629,10 +631,11 @@ type DismissalRestrictions struct {
 // restriction to allows only specific users or teams to dimiss pull request reviews. It is
 // separate from DismissalRestrictions above because the request structure is
 // different from the response structure.
+// Note: Both Users and Teams must be nil, or both must be non-nil.
 type DismissalRestrictionsRequest struct {
-	// The list of user logins who can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or []string{} otherwise.)
+	// The list of user logins who can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or &[]string{} otherwise.)
 	Users *[]string `json:"users,omitempty"`
-	// The list of team slugs which can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or []string{} otherwise.)
+	// The list of team slugs which can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or &[]string{} otherwise.)
 	Teams *[]string `json:"teams,omitempty"`
 }
 
--- a/github/repos.go
+++ b/github/repos.go
@@ -7,7 +7,6 @@ package github
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


mspiegel:
Closes #664

mspiegel:
ping. Any changes needed?

test/integration/repos_test.go
```diff
@@ -118,8 +118,8 @@ func TestRepositories_EditBranches(t *testing.T) {
 		},
 		RequiredPullRequestReviews: &github.PullRequestReviewsEnforcementRequest{
 			DismissalRestrictionsRequest: &github.DismissalRestrictionsRequest{
-				Users: []string{},
-				Teams: []string{},
+				Users: &[]string{},
+				Teams: &[]string{},
```
elliott-beach: This integration test (`TestRepositories_EditBranches`) still fails. I figure it could be addressed in this PR, as it covers the request being fixed. Since the test is using a personal repo rather than an org, `DismissalRestrictionsRequest` can be omitted, and then GitHub stops barfing.

FWIW, I got it to pass by making that change and also adding an enforcement URL for the `want` object:

```
enforcementURL :=  "https://api.github.com/repos/"+ *repo.Owner.Login + "/" + *repo.Name + "/branches/master/protection/enforce_admins"

...

EnforceAdmins: &github.AdminEnforcement{
			URL: &enforcementURL,
			Enabled: true,
		},
```

mspiegel:
ping. I fixed the integration test

gmlewis:
It sounds like we are waiting to hear back from GitHub support according to [this comment](https://github.com/google/go-github/issues/664#issuecomment-340520319), correct?

mspiegel:
Sorry, I didn't notice the date of that comment. So possibly GitHub introduced a bug recently when an empty array is passed for `dismissal_restrictions`. This pull request assumes the current behavior is the intended behavior. But we're waiting to see if that's true. Ugh.

mspiegel:
Ping. I don't think GitHub is going to fix their mismatch between their documentation and their current behavior. I have users giving me feedback that the functionality isn't working. So for now, I have to rely on my fork that includes this pull request (https://github.com/mspiegel/go-github/) :(

github/repos.go
```diff
@@ -658,10 +630,10 @@ type DismissalRestrictions struct {
 // separate from DismissalRestrictions above because the request structure is
 // different from the response structure.
 type DismissalRestrictionsRequest struct {
-	// The list of user logins who can dismiss pull request reviews. (Required; use []string{} instead of nil for empty list.)
-	Users []string `json:"users"`
-	// The list of team slugs which can dismiss pull request reviews. (Required; use []string{} instead of nil for empty list.)
-	Teams []string `json:"teams"`
+	// The list of user logins who can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or []string{} otherwise.)
+	Users *[]string `json:"users,omitempty"`
```
gmlewis: Note that a slice is already a reference type, so having a pointer to a slice here is not necessary.
Can you please try your local patches without this `*` and only with the addition of `omitempty`?
You should get identical results.
Please report back with your findings, and if it it does indeed still work, please remove these `*`s.

mspiegel: For personal repositories, the 'dismissal_restrictions' must be omitted. For repositories in an organization, the 'dismissal_restrictions' must either be an empty JSON object or JSON object with exactly two keys 'users' and 'teams'. Any other combination of repository and 'dismissal_restrictions' is rejected by the GitHub API.

If I add 'omitempty' to those fields, and also leave 'omitempty' on dismissal_restrictions, it no longer works.

gmlewis: I apologize for not being clear.

I'm not asking you to change `omitempty` at all.

I'm asking you to remove two asterisks:

```go
Users []string `json:"users,omitempty"`
//...
Teams []string `json:"teams,omitempty"`
```

mspiegel: Ah oops, no the error is on me. I read your comment too quickly. Take a look at these two configurations from the go playground. Here is the configuration using the pointers on the inner arrays: https://play.golang.org/p/nShoOQw1Fd_A. It allows us to specify 3 configurations. When we remove the pointers on the arrays: https://play.golang.org/p/OfPxg539U14. We can no longer generate this JSON object: `{"dismissal_restrictions":{"users":[],"teams":[]}}`

shurcooL: That is good justification for why it needs to be a `*[]string` with `omitempty`. It's to enable the JSON marshaling that is needed for GitHub API.

github/repos.go
```diff
@@ -658,10 +630,10 @@ type DismissalRestrictions struct {
 // separate from DismissalRestrictions above because the request structure is
 // different from the response structure.
 type DismissalRestrictionsRequest struct {
-	// The list of user logins who can dismiss pull request reviews. (Required; use []string{} instead of nil for empty list.)
-	Users []string `json:"users"`
-	// The list of team slugs which can dismiss pull request reviews. (Required; use []string{} instead of nil for empty list.)
-	Teams []string `json:"teams"`
+	// The list of user logins who can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or []string{} otherwise.)
```
shurcooL: `&[]string{}` here, not `[]string{}`.

Also, it might be worth adding a note saying that either both `Users` and `Teams` must be nil, or both must be non-nil.

test/integration/repos_test.go
```diff
@@ -135,19 +131,17 @@ func TestRepositories_EditBranches(t *testing.T) {
...
 		},
 		RequiredPullRequestReviews: &github.PullRequestReviewsEnforcement{
-			DismissalRestrictions: github.DismissalRestrictions{
-				Users: []*github.User{},
-				Teams: []*github.Team{},
-			},
 			DismissStaleReviews: true,
 		},
 		EnforceAdmins: &github.AdminEnforcement{
+			URL:     &enforcementURL,
```
shurcooL: You can use `String` method to create a `*string` inline here. I think it'd be simpler/more readable that way.

github/repos.go
```diff
@@ -573,40 +572,13 @@ type PullRequestReviewsEnforcement struct {
 // because the request structure is different from the response structure.
 type PullRequestReviewsEnforcementRequest struct {
 	// Specifies which users and teams should be allowed to dismiss pull request reviews. Can be nil to disable the restrictions.
-	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions"`
+	DismissalRestrictionsRequest *DismissalRestrictionsRequest `json:"dismissal_restrictions,omitempty"`
```
shurcooL: You said:

> For personal repositories, the 'dismissal_restrictions' must be omitted.

I think it's worth documenting here as well. We can take the phrasings from GitHub API docs:

> User and team dismissal restrictions are only available for organization-owned repositories. Omit this parameter for personal repositories.

github/repos.go
gmlewis: Ah, thank you for the playground example!  That makes sense now.

mspiegel:
I think we're in good shape. The GitHub API documentation _used_ to be inaccurate. It previously said "You can disable dismissal restrictions by passing the dismissal_restrictions object as an empty array." https://web.archive.org/web/20170812204213/https://developer.github.com/v3/repos/branches/. Now it says "Pass an empty dismissal_restrictions object to disable"

test/integration/repos_test.go
```diff
@@ -141,7 +138,7 @@ func TestRepositories_EditBranches(t *testing.T) {
 			DismissStaleReviews: true,
 		},
 		EnforceAdmins: &github.AdminEnforcement{
-			URL:     &enforcementURL,
+			URL:     github.String(enforcementURL),
```
shurcooL: The reason I suggested `github.String` is so that you could inline `enforcementURL `:

```Go
URL: github.String("https://api.github.com/repos/" + *repo.Owner.Login + "/" + *repo.Name + "/branches/master/protection/enforce_admins"),
```

mspiegel:
How's it look? I feel like we have momentum and this is nearly complete.

mspiegel:
Thanks! Merge? Is there a release cycle?

gmlewis:
Merging.

joebowbeer:
@dmitshur has anyone tested dismissal_restrictions on org-owned repos?

Based on what I'm seeing, setting `dismissal_restrictions` to an empty object does *not* disable, but rather enables _no one_ to dismiss.

In the GitHub branch protection settings I see dismissal restrictions enabled, but no one listed as being able to dismiss:

```
Restrict who can dismiss pull request reviews
Specify people or teams allowed to dismiss pull request reviews. 
```

If, furthermore, the restrictions apply to admins (set further down), then absolutely no one except the reviewer will be able to dismiss.

Perhaps not passing the key is the correct way to disable on org-owned repos?

gmlewis:
Hi @joebowbeer, could you please run your question by support@github.com and report back here what their answer is?

joebowbeer:
I asked GitHub support for clarification. Watch this space.

PR #792, opened 2017-11-23
Added 'review_requested' and 'review_request_removed' to PullRequestEvent action

### LINKED ISSUE #790: Add 'review_requested' and 'review_request_removed' to PullRequestEvent action
[PullRequestEvent](https://godoc.org/github.com/google/go-github/github#PullRequestEvent) action is missing 'review_requested' and 'review_request_removed'.

It would be an enhancement to add them.

https://developer.github.com/v3/activity/events/types/#pullrequestevent

scriptonist:
Fixes #790

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -486,7 +486,7 @@ type PullRequestEvent struct {
 	// the pull request was closed with unmerged commits. If the action is "closed"
 	// and the merged key is true, the pull request was merged.
 	Action      *string      `json:"action,omitempty"`
-	Number      *int         `json:"n	umber,omitempty"`
+	Number      *int         `json:"number,omitempty"`
 	PullRequest *PullRequest `json:"pull_request,omitempty"`
 
 	// The following fields are only populated by Webhook events.
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -479,9 +479,9 @@ type PublicEvent struct {
 //
 // GitHub API docs: https://developer.github.com/v3/activity/events/types/#pullrequestevent
 type PullRequestEvent struct {
-	// Action is the action that was performed. Possible values are: "assigned",
-	// "unassigned","review_requested","review_request_removed" "labeled", "unlabeled",
-	// "opened", "closed", or "reopened","synchronize", "edited".
+	// Action is the action that was performed. Possible values are:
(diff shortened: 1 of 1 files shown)


github/event_types.go
```diff
@@ -480,19 +480,24 @@ type PublicEvent struct {
...
-	Repo         *Repository   `json:"repository,omitempty"`
-	Sender       *User         `json:"sender,omitempty"`
-	Installation *Installation `json:"installation,omitempty"`
+	Changes *EditChange `json:"changes,omitempty"`
+	Repo    *Repository `json:"repository,omitempty"`
+	Sender  *User       `json:"sender,omitempty"`
+	// According to Github API Docs,
+	// Deliveries for review_requested and review_request_removed events
+	// will have an additional field called requested_reviewers.
+	RequestedReviewers []*User       `json:"requested_reviewers,omitempty"`
```
shurcooL: Get rid of "According to Github API Docs", and the rest can be simplified. We should say "RequestedReviewers" rather than "requested_reviewers", since the Go field is named `RequestedReviewers`. How about just this:

```Go
RequestedReviewers []*User `json:"requested_reviewers,omitempty"` // For "review_requested", "review_request_removed" events.
```

github/event_types.go
```diff
@@ -480,19 +480,24 @@ type PublicEvent struct {
 // GitHub API docs: https://developer.github.com/v3/activity/events/types/#pullrequestevent
 type PullRequestEvent struct {
 	// Action is the action that was performed. Possible values are: "assigned",
-	// "unassigned", "labeled", "unlabeled", "opened", "closed", or "reopened",
-	// "synchronize", "edited". If the action is "closed" and the merged key is false,
+	// "unassigned","review_requested","review_request_removed" "labeled", "unlabeled",
+	// "opened", "closed", or "reopened","synchronize", "edited".
```
shurcooL: Add space between `"reopened","synchronize"`, and get rid of `or`, it's not needed.

github/event_types.go
shurcooL: Also, let's move this field up by two, so it's under `RequestedReviewers` and before the more generic `Repo`, `Sender`. It feels out of place here.

scriptonist:
@shurcooL I've made the proposed changes.

PR #798, opened 2017-11-29
Set accept header for function Apps.ListUserRepos

### LINKED ISSUE #797: Apps.ListUserRepos may return message "specify a custom media type in the 'Accept' header".
When I am trying to use `Apps.ListUserRepos`, I got message like this:
image

I know this can be fix by adding following accept header:
```go
// in github/github.go line 85~86
// https://developer.github.com/changes/2016-09-14-Integrations-Early-Access/
mediaTypeIntegrationPreview = "application/vnd.github.machine-man-preview+json"
```

And it can be fix by add following to `ListUserRepos` function in `github/apps_installation.go`
```go
req.Header.Set("Accept", mediaTypeIntegrationPreview)
```

I can provide a PR to fix this.

I noticed that there is a TODO comment in other function says that `TODO: remove custom Accept header when this API fully launches.`, so I will keep this in my fix.

lingsamuel:
Fixes #797

### DIFF
--- a/github/apps_installation.go
+++ b/github/apps_installation.go
@@ -67,6 +67,9 @@ func (s *AppsService) ListUserRepos(ctx context.Context, id int, opt *ListOption
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
+
 	var r struct {
 		Repositories []*Repository `json:"repositories"`
 	}
--- a/github/apps_installation_test.go
+++ b/github/apps_installation_test.go
@@ -45,6 +45,7 @@ func TestAppsService_ListUserRepos(t *testing.T) {
 
 	mux.HandleFunc("/user/installations/1/repositories", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeIntegrationPreview)
 		testFormValues(t, r, values{
 			"page":     "1",
 			"per_page": "2",
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -479,9 +479,10 @@ type PublicEvent struct {
 //
 // GitHub API docs: https://developer.github.com/v3/activity/events/types/#pullrequestevent
 type PullRequestEvent struct {
-	// Action is the action that was performed. Possible values are: "assigned",
-	// "unassigned", "labeled", "unlabeled", "opened", "closed", or "reopened",
-	// "synchronize", "edited". If the action is "closed" and the merged key is false,
+	// Action is the action that was performed. Possible values are:
+	// "assigned", "unassigned", "review_requested", "review_request_removed", "labeled", "unlabeled",
+	// "opened", "closed", "reopened", "synchronize", "edited".
+	// If the action is "closed" and the merged key is false,
 	// the pull request was closed with unmerged commits. If the action is "closed"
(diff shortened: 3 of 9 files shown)


lingsamuel:
I signed it!

sahildua2305:
Just tested, I'm fine with the changes. @lingsamuel thanks for proactively submitting a PR to fix the issue. LGTM.

@shurcooL however, this actually exposes the weakness of our tests when it comes to header restrictions on some endpoints. For example - test for `Apps.ListUserRepos` passes even without the header restrictions. Maybe we can enforce the headers in `mux` requests setup.

github/apps_installation.go
```diff
@@ -67,6 +67,9 @@ func (s *AppsService) ListUserRepos(ctx context.Context, id int, opt *ListOption
 		return nil, nil, err
 	}
 
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeIntegrationPreview)
+
```
shurcooL: This part looks good.

sahildua2305:
@shurcooL if I remove this line in test for `ListRepos` method, I'll expect the test to fail because it has the header restriction. https://github.com/google/go-github/blob/a89ea1cdf79929726a9416663609269ada774da0/github/apps_installation_test.go#L22

However, on my local setup, the tests still pass. Are you able to reproduce this?

shurcooL:
The `testHeader` is a helper that checks that a header is set. If you remove that line, you're removing the check, so the test will of course pass.

The way you can check that `testHeader` works is by removing the `req.Header.Set("Accept", mediaTypeIntegrationPreview)` line from `AppsService.ListRepos` method, then its test (with `testHeader`) will fail.

sahildua2305:
Ah okay! Thanks for explaining, @shurcooL. My understanding of the `testHeader` method was wrong.

lingsamuel:
@shurcooL All done.

By the way, I think @sahildua2305 means the test should be failed if we can't get right data. Maybe he thought that the test is passed so the function works correctly.

sahildua2305:
@lingsamuel yes, that’s what I was thinking. Moreover, I assumed that testHeader was confirming that the request is successful only with the header mentioned.

shurcooL:
> Moreover, I assumed that testHeader was confirming that the request is successful only with the header mentioned.

It does do that. It's just many tests don't call that helper.

Specifically, `testHeader` checks for an exact header value match:

```Go
func testHeader(t *testing.T, r *http.Request, header string, want string) {
	if got := r.Header.Get(header); got != want {
		t.Errorf("Header.Get(%q) returned %q, want %q", header, got, want)
	}
}
```

So, if you want, you can use it like this in a test to ensure that the default `Accept` header value is set:

```Go
testHeader(t, r, "Accept", mediaTypeV3)
```

I'm not sure if there are benefits to doing that, but I haven't thought about it too much. If you think there are worthwhile improvements possible, feel free to open a new issue about it.

(P.S. This is a good example of where helpers make code less readable. I'm sure you wouldn't misunderstand what `testHeader` does if its code was written directly in the test, rather than being in a helper.)

lingsamuel:
Oh no, I think @sahildua2305 means, `testHeader` should ensure that we can't get the **right data** if we don't set the header.
Not just check if the header exists.

shurcooL:
Sorry, I don’t understand. If you think there’s a worthwhile opportunity to improve tests, feel free to open a new issue and describe it in more detail. I am satisfied with the current tests as they are, and don’t see any ways to make them significantly better.

PR #805, opened 2017-12-06
RoundTrip: avoid modifying the original request

### LINKED ISSUE #556: UnauthenticatedRateLimitedTransport.RoundTrip breaks http.RoundTripper contract (mostly harmless).
`UnauthenticatedRateLimitedTransport.RoundTrip` does this:

```Go
// To set extra querystring params, we must make a copy of the Request so
// that we don't modify the Request we were given. This is required by the
// specification of http.RoundTripper.
req = cloneRequest(req)
q := req.URL.Query()
q.Set("client_id", t.ClientID)
q.Set("client_secret", t.ClientSecret)
req.URL.RawQuery = q.Encode()
```

But `cloneRequest` doesn't make a copy of the `URL` field, it still points to the same `url.URL` struct that the original request points to. So doing `req.URL.RawQuery = q.Encode()` actually modifies the `URL` of original request.

This issue is mostly harmless, but it's a bug.

This is similar to https://go-review.googlesource.com/c/36483/.

sahildua2305:
Avoid modifying the original request as per http.RoundTripper contract.
In UnauthenticatedRateLimitedTransport.RoundTrip, we need to modify the
URL of the request only, while in BasicAuthTransport.RoundTrip, we need
to modify only the headers.

Hence, we get rid of cloneRequest method which wasn't working good for
the needs of UnauthenticatedRateLimitedTransport.RoundTrip. Instead, now
we have the implementation of cloneRequest inline in both the RoundTrip
interfaces.

We decided to make the cloneRequest implementation inline because of its
use at only these two places and we think we won't gain much by making a
generic implementation of cloneRequest as a function. To follow the
discussion on this, check the GitHub issue - https://github.com/google/go-github/issues/556.

Fixes #556

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -859,10 +859,8 @@ func (t *UnauthenticatedRateLimitedTransport) RoundTrip(req *http.Request) (*htt
 	//
 	// Since we are going to modify only req.URL here, we only need a deep copy
 	// of req.URL.
-	// shallow copy of request
 	req2 := new(http.Request)
 	*req2 = *req
-	// deep copy of request URL
 	req2.URL = new(url.URL)
 	*req2.URL = *req.URL
 
@@ -910,10 +908,8 @@ func (t *BasicAuthTransport) RoundTrip(req *http.Request) (*http.Response, error
 	//
 	// Since we are going to modify only req.Header here, we only need a deep copy
 	// of req.Header.
-	// shallow copy of the struct
 	req2 := new(http.Request)
 	*req2 = *req
-	// deep copy of the Header
 	req2.Header = make(http.Header, len(req.Header))
 	for k, s := range req.Header {
 		req2.Header[k] = append([]string(nil), s...)
--- a/github/github.go
+++ b/github/github.go
@@ -912,7 +912,7 @@ func (t *BasicAuthTransport) RoundTrip(req *http.Request) (*http.Response, error
 	*req2 = *req
 	req2.Header = make(http.Header, len(req.Header))
 	for k, s := range req.Header {
-		req2.Header[k] = append([]string{}, s...)
+		req2.Header[k] = append([]string(nil), s...)
 	}
 
 	req2.SetBasicAuth(t.Username, t.Password)
--- a/github/github.go
+++ b/github/github.go
@@ -856,14 +856,23 @@ func (t *UnauthenticatedRateLimitedTransport) RoundTrip(req *http.Request) (*htt
 	// To set extra querystring params, we must make a copy of the Request so
 	// that we don't modify the Request we were given. This is required by the
 	// specification of http.RoundTripper.
-	req = cloneRequest(req)
-	q := req.URL.Query()
+	//
+	// Since we are going to modify only req.URL here, we only need a deep copy
+	// of req.URL.
+	// shallow copy of request
+	req2 := new(http.Request)
+	*req2 = *req
+	// deep copy of request URL
+	req2.URL = new(url.URL)
+	*req2.URL = *req.URL
+
+	q := req2.URL.Query()
 	q.Set("client_id", t.ClientID)
 	q.Set("client_secret", t.ClientSecret)
-	req.URL.RawQuery = q.Encode()
+	req2.URL.RawQuery = q.Encode()
 
 	// Make the HTTP request.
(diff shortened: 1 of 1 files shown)


sahildua2305:
@shurcooL I am struggling with tests for this change. I tried this -
```
tp := &UnauthenticatedRateLimitedTransport{
    ClientID: "id",
    ClientSecret: "secret",
}
req := httptest.NewRequest("GET", "/foo/bar", nil)
want := req.URL
_, err := tp.RoundTrip(req)
if got := req.URL; got != want {
    t.Errorf("RoundTrip should not modify the original request")
}
```

However, that doesn't work because RoundTrip request itself fails.

shurcooL:
> I am struggling with tests for this change.

I don't think this change needs a test. We'll just review it carefully to ensure that the `URL` field of original request is not being modified, and the existing package tests will ensure that functionality is still working.

sahildua2305:
> I don't think this change needs a test.

Okay, then this PR is ready for review.

sahildua2305:
@shurcooL did you have a chance to look at this? I know this is a low priority pull request. Just a gentle reminder.

github/github.go
```diff
@@ -856,14 +856,23 @@ func (t *UnauthenticatedRateLimitedTransport) RoundTrip(req *http.Request) (*htt
...
 	// specification of http.RoundTripper.
-	req = cloneRequest(req)
-	q := req.URL.Query()
+	//
+	// Since we are going to modify only req.URL here, we only need a deep copy
+	// of req.URL.
+	// shallow copy of request
+	req2 := new(http.Request)
+	*req2 = *req
+	// deep copy of request URL
```
shurcooL: The "shallow copy of request" and "deep copy of request URL" phrases seem out of place here, since there's a sentence explaining it on top. Let's get rid of them.

github/github.go
```diff
@@ -895,12 +904,26 @@ type BasicAuthTransport struct {
...
+	// To set extra headers, we must make a copy of the Request so
+	// that we don't modify the Request we were given. This is required by the
+	// specification of http.RoundTripper.
+	//
+	// Since we are going to modify only req.Header here, we only need a deep copy
+	// of req.Header.
+	// shallow copy of the struct
+	req2 := new(http.Request)
+	*req2 = *req
+	// deep copy of the Header
```
shurcooL: Also get rid of "shallow copy of the struct" and "deep copy of the Header" here.

github/github.go
```diff
@@ -895,12 +902,24 @@ type BasicAuthTransport struct {
...
+	// that we don't modify the Request we were given. This is required by the
+	// specification of http.RoundTripper.
+	//
+	// Since we are going to modify only req.Header here, we only need a deep copy
+	// of req.Header.
+	req2 := new(http.Request)
+	*req2 = *req
+	req2.Header = make(http.Header, len(req.Header))
+	for k, s := range req.Header {
+		req2.Header[k] = append([]string(nil), s...)
```
gmlewis: Why not `[]string{}` instead of `[]string(nil)`?
I see it was written that way before in `cloneRequest` but I'm not sure I understand the difference.
`[]string{}` seems more common (to me) but maybe it's not a big deal?

sahildua2305: @gmlewis good one! I didn't think about it. Definitely, `[]string{}` is more common one. I'll change that in next commit. Thanks for reviewing!

shurcooL: > Why not `[]string{}` instead of `[]string(nil)`?

@gmlewis I think it's because that's the style used in [Go slice tricks document](https://github.com/golang/go/wiki/SliceTricks#copy), where this "copy slice" pattern is taken from. I think it's better to stay consistent with that.

If you think it should be changed, I suggest you make the proposal to change Go slice tricks, then we can update our style here.

gmlewis: Oh, weird!  Thanks for the pointer, @shurcooL. OK, I'm fine with keeping it that way.  I just thought it was odd-looking... looks odd there too, but that's fine... no need to change it. I'll shut up now.  :smile:

sahildua2305: Thanks for the pointer, @shurcooL. Changed it back 😄

shurcooL:
Thank you @sahildua2305 and @gmlewis!

PR #807, opened 2017-12-10
Upgrade RepoStatus.ID from int to int64

### LINKED ISSUE #806: RepoStatus.id needs to be int64
[RepoStatus.ID](https://godoc.org/github.com/google/go-github/github#RepoStatus) is *int but it must be int64. When calling [RepositoriesService.CreateStatus()](https://godoc.org/github.com/google/go-github/github#RepositoriesService.CreateStatus), the call fails with:
```
json: cannot unmarshal number 4396447431 into Go struct field RepoStatus.id of type int
```

It broke [gohci](https://github.com/periph/gohci) which is used for https://periph.io.

maruel:
This is needed on 32 bits platforms. Right now it is impossible to call
CreateStatus() on 32 bits executable because the ID returned by Github
overflows.

This is an API breaking change.

Fixes #806

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -8564,12 +8564,12 @@ func (r *RepoStatus) GetDescription() string {
 	return *r.Description
 }
 
-// GetID returns the ID field if it's non-nil, zero value otherwise.
-func (r *RepoStatus) GetID() int64 {
-	if r == nil || r.ID == nil {
-		return 0
+// GetID returns the ID field.
+func (r *RepoStatus) GetID() *int64 {
+	if r == nil {
+		return nil
 	}
-	return *r.ID
+	return r.ID
 }
 
 // GetState returns the State field if it's non-nil, zero value otherwise.
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -8565,7 +8565,7 @@ func (r *RepoStatus) GetDescription() string {
 }
 
 // GetID returns the ID field if it's non-nil, zero value otherwise.
-func (r *RepoStatus) GetID() int {
+func (r *RepoStatus) GetID() int64 {
 	if r == nil || r.ID == nil {
 		return 0
 	}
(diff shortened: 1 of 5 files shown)


maruel:
I do not address the other IDs as I confirmed this change is sufficient to get gohci back into working state and this reduces the amount of API breakage.

sahildua2305:
> I'm not sure why GetID() exists, it's not useful.

@maruel answer to your comment in commit 84280894f9f6e7c343ce9ccaa46e4451b5d877c2 - Since RepoStatus is an exported struct, it will have automatically generated accessors for all of its exported fields, unless the struct itself is blacklisted or some of its fields are blacklisted.

shurcooL:
We should not merge this PR until we make a decision in #597. Merging it sooner would make a decision implicitly, cause an API breakage, and put the library in an inconsistent state (`int`s in some places, `int64` in others), and make resolving this issue even harder.

This library has a lot of users, and the vast majority of them are on 64-bit architectures, so they're not negatively affected (their `int`s are 64-bit).  For those on 32-bit systems that need this, they should apply the change ad-hoc until we can resolve the issue upstream in a way that's viable, which is what @maruel said he has already done in https://github.com/google/go-github/issues/806#issuecomment-350561012.

Thanks for sending this PR @maruel, but I'll close it for now to avoid this PR being merged by accident (it can easily be re-opened when needed). We should continue further discussion and come up with a plan of action in #597.

github/github-accessors.go
```diff
@@ -8564,12 +8564,12 @@ func (r *RepoStatus) GetDescription() string {
 	return *r.Description
 }
 
-// GetID returns the ID field if it's non-nil, zero value otherwise.
-func (r *RepoStatus) GetID() int {
-	if r == nil || r.ID == nil {
-		return 0
+// GetID returns the ID field.
+func (r *RepoStatus) GetID() *int64 {
```
shurcooL: For posterity (no need to resolve this now), the generator is not doing the right thing for `int64` type. It should be generating the following signature:

```Go
func (r *RepoStatus) GetID() int64
```

But it's treating `int64` as if it were a struct.

PR #808, opened 2017-12-11
bugfix: pr comment field should be in_reply_to_id

jordanlewis:
It was previously and incorrectly `in_reply_to`.
See https://developer.github.com/v3/pulls/comments/

### DIFF
--- a/github/pulls_comments.go
+++ b/github/pulls_comments.go
@@ -14,7 +14,7 @@ import (
 // PullRequestComment represents a comment left on a pull request.
 type PullRequestComment struct {
 	ID               *int       `json:"id,omitempty"`
-	InReplyTo        *int       `json:"in_reply_to,omitempty"`
+	InReplyTo        *int       `json:"in_reply_to_id,omitempty"`
 	Body             *string    `json:"body,omitempty"`
 	Path             *string    `json:"path,omitempty"`
 	DiffHunk         *string    `json:"diff_hunk,omitempty"`


jordanlewis:
Ah, I didn't realize that this struct was used for both input and output.

Perhaps there should be a `PullRequestCommentRequest` struct that represents the input to `CreateComment`? Otherwise, seeing those duplicate-looking fields might be confusing for users.

By the way, are you aware of any way to submit a comment reply as a draft? That appears to be missing from the GitHub API itself, but perhaps you're aware of a workaround.

shurcooL:
> Ah, I didn't realize that this struct was used for both input and output.

Unfortunately, a handful of structs are, including `IssueComment`, `PullRequest`, and some others. These decisions were made years ago and are hard to change now. I prefer separate structs for input for new GitHub APIs, especially when they have very little in common with output (and we do this in a few places where it was unavoidable).

> Perhaps there should be a `PullRequestCommentRequest` struct that represents the input to `CreateComment`? Otherwise, seeing those duplicate-looking fields might be confusing for users.

That would be a potential enhancement, but a breaking API change. I suggest we consider it separately from the simple issue of a single missing field. No need to conflate the two together.

> By the way, are you aware of any way to submit a comment reply as a draft? That appears to be missing from the GitHub API itself, but perhaps you're aware of a workaround.

It's not something I have direct experience with yet, so I can't say for sure. But I suspect you might be able to do it by creating a draft review. I don't think standalone comments can have drafts.

gmlewis:
Closing as resolved.

PR #810, opened 2017-12-14
Added examples to the README

### LINKED ISSUE #809: Adding a simple working program under the Usage section of the README
Hello,

I am writing a simple Go program which uses this package to make some API calls to GitHub as part of learning Go. It took me some redirections to get started with the package to make native Unauthenticated and Basic authenticated calls; probably due to my naivety at reading/writing Go code.

Hence, I thought of mentioning this so as to help beginners (like me) pick the package up quicker in the future. I would like to suggest having at least one complete working code snippet which can be used exclusively as a copy-paste program. This can be achieved in two ways:
1. Adding a link to the `examples` directory in the "Usage" section.
2. Modify the already present examples to be inclusive enough for direct use as recipes.

Since I am totally new to Go development environment, I beg pardon if my suggestions seem rather ingenuous to the more experienced community members.

kshitij10496:
The modifications performed in this commit:
1.    A working code snippet has been added to the README
2.    The "examples" folder has been explicitly hyperlinked

Resolves #809

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Construct a new GitHub client, then use the various services on the client to
 access different parts of the GitHub API. For example:
 
 ```go
+ctx = context.Background()
 client := github.NewClient(nil)
 
 // list all organizations for user "willnorris"
@@ -31,6 +32,7 @@ orgs, _, err := client.Organizations.List(ctx, "willnorris", nil)
 Some API methods have optional parameters that can be passed. For example:
 
 ```go
+ctx = context.Background()
 client := github.NewClient(nil)
 
 // list public repositories for org "github"
@@ -43,30 +45,12 @@ the structure of the GitHub API documentation at
 https://developer.github.com/v3/.
 
 
-A basic unauthenticated program might look something like:
-
-```go
-import (
-	"context"
-	"github.com/google/go-github/github"
-)
-
-func FetchOrganizations(username string) ([]*Authorization, error) {
-	ctx := context.Background()
-	client := github.NewClient(nil)
-
-	// list all organizations for user "willnorris"
-	orgs, _, err := client.Organizations.List(ctx, "willnorris", nil)
-
-	if err != nil {
-		return nil, err
-	}
-	return orgs, err
-}
-```
+NOTE: The [context](https://godoc.org/context) package is being used to provide
+request scoped values to the API calls. If case there is no context available,
+then `context.Background()` can be used as a starting point.
 
 For more sample code snippets, head over to the
-[examples](https://github.com/google/go-github/tree/master/example) directory.
+[example](https://github.com/google/go-github/tree/master/example) directory.
 
 
 ### Authentication ###
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@ The services of a client divide the API into logical chunks and correspond to
 the structure of the GitHub API documentation at
 https://developer.github.com/v3/.
 
-
 NOTE: Using the [context](https://godoc.org/context) package, one can easily
 pass cancelation signals and deadlines to various services of the client for
 handling a request. In case there is no context available, then `context.Background()`
@@ -50,7 +49,6 @@ can be used as a starting point.
 
 For more sample code snippets, head over to the
 [example](https://github.com/google/go-github/tree/master/example) directory.
-
 
 ### Authentication ###
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,33 @@ The services of a client divide the API into logical chunks and correspond to
 the structure of the GitHub API documentation at
 https://developer.github.com/v3/.
 
+
+A basic unauthenticated program might look something like:
+
+```go
+import (
+	"context"
+	"github.com/google/go-github/github"
+)
+
+func FetchOrganizations(username string) ([]*Authorization, error) {
+	ctx := context.Background()
+	client := github.NewClient(nil)
+
+	// list all organizations for user "willnorris"
+	orgs, _, err := client.Organizations.List(ctx, "willnorris", nil)
+
+	if err != nil {
+		return nil, err
+	}
+	return orgs, err
+}
+```
+
+For more sample code snippets, head over to the
+[examples](https://github.com/google/go-github/tree/master/example) directory.
+
+
 ### Authentication ###
 
 The go-github library does not directly handle authentication. Instead, when
--- a/README.md
+++ b/README.md
@@ -22,22 +22,20 @@ Construct a new GitHub client, then use the various services on the client to
 access different parts of the GitHub API. For example:
 
 ```go
-ctx = context.Background()
 client := github.NewClient(nil)
 
 // list all organizations for user "willnorris"
-orgs, _, err := client.Organizations.List(ctx, "willnorris", nil)
+orgs, _, err := client.Organizations.List(context.Background(), "willnorris", nil)
 ```
 
 Some API methods have optional parameters that can be passed. For example:
 
 ```go
-ctx = context.Background()
 client := github.NewClient(nil)
 
 // list public repositories for org "github"
 opt := &github.RepositoryListByOrgOptions{Type: "public"}
-repos, _, err := client.Repositories.ListByOrg(ctx, "github", opt)
+repos, _, err := client.Repositories.ListByOrg(context.Background(), "github", opt)
 ```
 
 The services of a client divide the API into logical chunks and correspond to
@@ -45,9 +43,10 @@ the structure of the GitHub API documentation at
 https://developer.github.com/v3/.
 
 
-NOTE: The [context](https://godoc.org/context) package is being used to provide
-request scoped values to the API calls. If case there is no context available,
-then `context.Background()` can be used as a starting point.
+NOTE: Using the [context](https://godoc.org/context) package, one can easily
+pass cancelation signals and deadlines to various services of the client for
+handling a request. In case there is no context available, then `context.Background()`
+can be used as a starting point.
 
 For more sample code snippets, head over to the
 [example](https://github.com/google/go-github/tree/master/example) directory.
--- a/example/simple/main.go
+++ b/example/simple/main.go
@@ -6,26 +6,20 @@
 // The simple command demonstrates a simple functionality which
 // prompts the user for a GitHub username and lists all the public
 // organization memberships of the specified username.
-
 package main
 
 import (
 	"context"
 	"fmt"
+
 	"github.com/google/go-github/github"
 )
 
 // Fetch all the public organizations' membership of a user.
 //
 func FetchOrganizations(username string) ([]*github.Organization, error) {
-	ctx := context.Background()
 	client := github.NewClient(nil)
-
-	orgs, _, err := client.Organizations.List(ctx, username, nil)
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -42,6 +42,33 @@ The services of a client divide the API into logical chunks and correspond to
...
+
+	// list all organizations for user "willnorris"
+	orgs, _, err := client.Organizations.List(ctx, "willnorris", nil)
+
+	if err != nil {
+		return nil, err
+	}
+	return orgs, err
+}
+```
```
shurcooL: This program, contrary to what the objective of #809 was, is not actually complete. It's missing a `package main` clause, and it's missing `func main() { ... }`. I think we should fix that and make it a complete, valid program.

However, I think the README is not the best place for it. It becomes a long snippet and feels out of place, given the above paragraphs already cover this.

Instead, I really like the idea of linking to the example directory. So let's make this a complete basic program and move it to the `example` directory.

README.md
```diff
@@ -42,6 +42,33 @@ The services of a client divide the API into logical chunks and correspond to
...
+
+For more sample code snippets, head over to the
+[examples](https://github.com/google/go-github/tree/master/example) directory.
```
shurcooL: The directory is named `example`, not `examples`, you should refer to it accordingly. ([This](https://dmitri.shuralyov.com/idiomatic-go#use-singular-form-for-collection-repo-folder-name) explains why.)

kshitij10496: Thanks a lot for the resource @shurcooL 
I will make the necessary changes.

I read through the [Contributing Guide](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) and I had a question regarding pushing a new commit(with the desired changes) so as to update the PR.

Should I push both the commits or should I squash them into a single commit on my end?

README.md
kshitij10496: In this example, I attempted to create a simple generic function which can be placed directly into any package. Hence, I thought of excluding the `package` name and `func main` as I wanted to separate the caller from the functionality.

I agree with you that adding a complete program might render the README look clumsy and dense. However, there are two issues that we need to address before we migrate this code to the `example` directory:

1. Adding information regarding the need to import `context` package and defining `ctx` variable in the snippets preceding this example code.
2. Move the example to a `noauth` subdirectory (maybe?) under the `example` directory. 

If we can agree to these suggestions, I can make the required modifications.

P.S: I notice two bugs in the example snippet:
1. use `username` instead of the hardcoded string
2. `[]*Organization` should be the type of the first return value.

README.md
shurcooL: We should add a mention for it there, but for a contributor, you should prefer to always push separate, new commits to PRs, rather than squashing and force pushing existing ones.

The reason to do so is because it makes reviewing the PR easier. Each change can be seen separately, so once a PR is reviewed partially, only the new changes need to be reviewed, rather than starting from scratch each time.

When we merge your PR, we will squash all your commits into a single clean logical commit.

README.md
shurcooL: > Adding information regarding the need to import context package and defining ctx variable in the snippets preceding this example code.

I think this can be resolved by having a comment in the program. It can be a few lines long and explain that [`context`](https://godoc.org/context) package is being used. If there's no context available, then `context.Background()` can be used as a starting point. But it should point reader to the documentation of [`context`](https://godoc.org/context) package for more info.

> Move the example to a noauth subdirectory (maybe?) under the example directory.

I think we should have all examples under `example` directory, let's just give them descriptive names.

I would suggest we call the example you're suggesting `simple` or `hello` or something like that.

We should also link to it right after the `// list all organizations for user "willnorris"` snippet in README.

I'm not completely convinced that having a general function would be helpful in this context. General functions make sense when used multiple times, otherwise they add unnecessary complexity. I'd imagine a simpler example of a complete program would be a better starting point. For example, this is how I imagine `simple` can look like:

```Go
// TODO: Explain what this command does, etc.
package main

import (
	"context"
	"fmt"
	"log"

)

func main() {
	// Create a GitHub API client without authentication.

	// List all organizations for user "willnorris".
	orgs, _, err := client.Organizations.List(context.Background(), "willnorris", nil)
	if err != nil {
		log.Fatalln(err)
	}

	// Print organization names.
	for _, org := range orgs {
		fmt.Println(org.Name)
	}
}
```

kshitij10496:
@shurcooL Does this look better now?

example/simple/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Fetch all the public organizations' membership of a user.
+//
+func FetchOrganizations(username string) ([]*github.Organization, error) {
+	ctx := context.Background()
+	client := github.NewClient(nil)
+
+	orgs, _, err := client.Organizations.List(ctx, username, nil)
+
+	if err != nil {
```
shurcooL: I'd suggest dropping this `if` statement, just do:

```Go
orgs, _, err := client.Organizations.List(ctx, username, nil)
return orgs, err
```

It's simpler code and functionally the same. If `err` is non-nil, we can just pass `orgs` through unmodified. There's no need to explicitly return `nil` instead.

example/simple/main.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/google/go-github/github"
+)
+
+// Fetch all the public organizations' membership of a user.
+//
+func FetchOrganizations(username string) ([]*github.Organization, error) {
+	ctx := context.Background()
```
shurcooL: Inline `ctx` into the only call where it's needed. It's not being used in any other way, so inlined is simpler I think, and works better for this example. I.e.:

```Go
orgs, _, err := client.Organizations.List(context.Background(), username, nil)
```

example/simple/main.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+//
+// The simple command demonstrates a simple functionality which
+// prompts the user for a GitHub username and lists all the public
+// organization memberships of the specified username.
+
```
shurcooL: This package comment is not formatted correctly and won't show up in `godoc`.

The package comment needs to be directly above the `package` clause, without a blank line in between. It should also be separate from the license comments above.

See https://golang.org/doc/effective_go.html#commentary for more information about the package comment. See [here](https://github.com/google/go-github/blob/0c3b302de2a6de84a2511db47ea1bb2ff8146830/example/basicauth/main.go#L1-L9) for another example that does this right.

README.md
```diff
@@ -22,6 +22,7 @@ Construct a new GitHub client, then use the various services on the client to
 access different parts of the GitHub API. For example:
 
 ```go
+ctx = context.Background()
```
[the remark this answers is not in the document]
kshitij10496: Inline value makes the code succinct. I will make this change throughout the examples.

README.md
```diff
@@ -42,6 +44,15 @@ The services of a client divide the API into logical chunks and correspond to
 the structure of the GitHub API documentation at
 https://developer.github.com/v3/.
 
+
+NOTE: The [context](https://godoc.org/context) package is being used to provide
+request scoped values to the API calls. If case there is no context available,
+then `context.Background()` can be used as a starting point.
```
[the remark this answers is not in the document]
kshitij10496: > This is not an accurate statement.

I'm sorry about the first statement;  I should be researched better. I learned about the importance and usage of the `context` package only recently after pushing the changes.

How about we rephrase the comment as follows:

> Using the `context` package, one can easily pass cancelation signals and deadlines to various services of the client for handling a request. In case there is no context available, then `context.Background()` can be used as a starting point.

example/simple/main.go
kshitij10496: Got it! Thanks.

example/simple/main.go
kshitij10496: I used to write such naive code 10 days ago!
I still do but I am definitely learning a lot.

kshitij10496:
@shurcooL Can you give this another review?

README.md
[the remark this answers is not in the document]
shurcooL: Sounds good.

README.md
```diff
@@ -35,13 +35,23 @@ client := github.NewClient(nil)
...
 opt := &github.RepositoryListByOrgOptions{Type: "public"}
-repos, _, err := client.Repositories.ListByOrg(ctx, "github", opt)
+repos, _, err := client.Repositories.ListByOrg(context.Background(), "github", opt)
 ```
 
 The services of a client divide the API into logical chunks and correspond to
 the structure of the GitHub API documentation at
 https://developer.github.com/v3/.
 
+
```
shurcooL: No need for two blank lines here, just make it one for consistency.

README.md
```diff
@@ -35,13 +35,23 @@ client := github.NewClient(nil)
...
+NOTE: Using the [context](https://godoc.org/context) package, one can easily
+pass cancelation signals and deadlines to various services of the client for
+handling a request. In case there is no context available, then `context.Background()`
+can be used as a starting point.
+
+For more sample code snippets, head over to the
+[example](https://github.com/google/go-github/tree/master/example) directory.
+
+
```
shurcooL: Also here, one line is enough.

shurcooL: Additionally, you should update `doc.go` with the same changes to keep it in sync with README.

example/simple/main.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+//
```
shurcooL: The copyright header and the package comment should be separate comments. Like this:

```Go
// Copyright 2017 The go-github AUTHORS. All rights reserved.
//

// The simple command demonstrates a simple functionality which
// prompts the user for a GitHub username and lists all the public
// organization memberships of the specified username.
package main
```

Notice the blank line separating the comments.

Otherwise the entire comment is treated as the package comment; that's not what we want to see in godoc.

example/simple/main.go
```diff
@@ -0,0 +1,40 @@
...
+	fmt.Scanf("%s", &username)
+
+	organizations, err := FetchOrganizations(username)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+		return
+	}
+
+	for i, organization := range organizations {
+		fmt.Printf("%v. %v\n", i+1, *organization.Login)
```
gmlewis: How about changing `*organization.Login` to `organization.GetLogin()` in order to demonstrate the use of accessors?

example/simple/main.go
kshitij10496: Got it. Thanks for the explanation.

README.md
kshitij10496: Sure, I will make the changes.

example/simple/main.go
kshitij10496: Sure!

kshitij10496:
Thanks, @shurcooL, and @gmlewis for the review.
I have tried to address your suggestions in the last commit.

gmlewis:
Sorry for the delay... Merging.

Issue #524, opened 2017-01-19
Support preview Organizations Membership API and new filter


gmlewis:
Announcement: https://developer.github.com/changes/2017-01-16-audit-outside-collaborators-for-2fa/

glapul:
Hi, I'd like to work on it.

gmlewis:
Thank you, @glapul - I'll send you an invite so that I can assign it to you.

gmlewis:
You are correct.  The new Organizations Membership API needs to be supported too.

Announcement: https://developer.github.com/changes/2016-11-28-preview-org-membership/
I'll change the label for this issue, and it can address both.

glapul:
#538 I've implemented the support for listing outside collaborators. Feedback?

e-beach:
Should this be closed, as #538 was closed and merged?

gmlewis:
From a quick glance at this announcement: https://developer.github.com/changes/2016-11-28-preview-org-membership/ it appears that #538 did not fully support all the new API endpoints. For example, I don't think [Remove an outside collaborator](https://developer.github.com/v3/orgs/outside_collaborators/#remove-outside-collaborator) was addressed.  There may be more, and part of the ownership of an issue (in my opinion) is to investigate what all is needed to completely resolve the issue.

If @glapul would like to complete this issue, then @glapul is welcome to... otherwise if someone else would like to read the announcements and add the missing endpoints, that is fine with me too.

PR #812, opened 2017-12-16
Add Outside Collaborators API for OrganizationsService

kshitij10496:
I have implemented the two methods on `OrganizationsService` corresponding to missing
endpoints of the Outside Collaborators API, namely:
1. [`RemoveOutsideCollaborator`](https://developer.github.com/v3/orgs/outside_collaborators/#remove-outside-collaborator)
2. [`ConvertMemberToOutsideCollaborator`](https://developer.github.com/v3/orgs/outside_collaborators/#convert-member-to-outside-collaborator)

### TODO
- [x] Add unit tests

This was my first attempt at wrapping endpoints so it is likely that the code can be improved upon.
Meanwhile, I will work on adding unit tests for these endpoints.

Refer #524

### DIFF
--- a/github/orgs_outside_collaborators_test.go
+++ b/github/orgs_outside_collaborators_test.go
@@ -64,6 +64,38 @@ func TestOrganizationsService_RemoveOutsideCollaborator(t *testing.T) {
 	}
 }
 
+func TestOrganizationsService_RemoveOutsideCollaborator_NonMember(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusNotFound)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.RemoveOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusNotFound {
+		t.Errorf("Organizations.RemoveOutsideCollaborator_NonMember returned error: %v", err)
+	}
+}
+
+func TestOrganizationsService_RemoveOutsideCollaborator_Member(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusUnprocessableEntity)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.RemoveOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusUnprocessableEntity {
+		t.Errorf("Organizations.RemoveOutsideCollaborator_Member returned error: %v", err)
+	}
+}
+
 func TestOrganizationsService_ConvertMemberToOutsideCollaborator(t *testing.T) {
 	client, mux, _, teardown := setup()
 	defer teardown()
@@ -76,5 +108,37 @@ func TestOrganizationsService_ConvertMemberToOutsideCollaborator(t *testing.T) {
 	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(context.Background(), "o", "u")
 	if err != nil {
 		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator returned error: %v", err)
+	}
+}
+
+func TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMember(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		w.WriteHeader(http.StatusForbidden)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusForbidden {
+		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator_NonMember returned error: %v", err)
+	}
+}
+
+func TestOrganizationsService_ConvertMemberToOutsideCollaborator_LastOwner(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		w.WriteHeader(http.StatusForbidden)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusForbidden {
+		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator_LastOwner returned error: %v", err)
 	}
 }
--- a/github/orgs_outside_collaborators_test.go
+++ b/github/orgs_outside_collaborators_test.go
@@ -111,7 +111,7 @@ func TestOrganizationsService_ConvertMemberToOutsideCollaborator(t *testing.T) {
 	}
 }
 
-func TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMember(t *testing.T) {
+func TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMemberOrLastOwner(t *testing.T) {
 	client, mux, _, teardown := setup()
 	defer teardown()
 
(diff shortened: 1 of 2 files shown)


kshitij10496:
I have added two tests corresponding to the accepted behavior of the endpoints.
However, there are other possibilities for writing unit tests which I would like to address:

- `RemoveOutsideCollaborator`
  - [x] Removing a outside collaborator from the organization: 204
  - [ ] Removing a member from the organization: 422
  - [ ] Removing a non-member from the organization: 404 (from: Manual experimentation)

- `ConvertMemberToOutsideCollaborator`
  - [x] Converting a member of the organization to outside collaborator: 204
  - [ ] Converting a non-member to outside collaborator: 403
  - [ ] Converting the last owner of the organization to outside collaborator: 403

Refer: [Outside Collaborators](https://developer.github.com/v3/orgs/outside_collaborators/#remove-outside-collaborator)

sahildua2305:
Very nice, thoughtful PR, @kshitij10496.

I feel that it makes sense to include the test cases for those "invalid" scenarios as well. 👍

kshitij10496:
> Very nice, thoughtful PR, @kshitij10496.

Thank you, yet again for the appreciation. 😄 

> I feel that it makes sense to include the test cases for those "invalid" scenarios as well. 

I am working on adding those cases. Since this is my first time writing unit tests in Go, they will need a thorough review.

kshitij10496:
I would like to share a question which irked me while I was writing test cases for the methods.

_What are the guidelines we use to decide between returning a non-nil `error` value or an "expected" unsuccessful(302, 404, 403, 422) response with a `nil` error?_

For example:
- [Get Team Member](https://developer.github.com/v3/orgs/teams/#get-team-member) endpoint implemented via [`OrganizationsService.IsTeamMember`](https://github.com/google/go-github/blob/fbfee053c26dab3772adfc7799d995eed379133e/github/orgs_teams.go#L273) method.
This endpoint returns a 404 response in certain cases which are passed as `nil` error values by our client. I understand that this makes sense because a boolean nature response is expected here.

However, I found it difficult to formulate a general approach to writing code for an endpoint which can return multiple types of response.

Currently, the [endpoints](https://developer.github.com/v3/orgs/outside_collaborators/#remove-outside-collaborator) I have implemented in this PR return only the 204 response with a `nil` error. In all other cases,(including the 403 mentioned in the spec) some non-nil error value is returned. The reason behind this particular approach is the correspondence between `ErrorResponse` type and the API response received in such cases.(In particular, the fields `Message` and `DocumentUrl` resonate with the `message` and `document_url` JSON fields of the response from the API.)

Should the implementation return all the different responses as a `*Response` value as opposed to an `error` value? In general, how can one decide which approach to take?

kshitij10496:
This PR is good for a review. 
Ping @gmlewis @shurcooL

github/orgs_outside_collaborators_test.go
```diff
@@ -48,3 +48,97 @@ func TestOrganizationsService_ListOutsideCollaborators_invalidOrg(t *testing.T)
...
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusForbidden {
+		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator_NonMember returned error: %v", err)
+	}
+}
+
+func TestOrganizationsService_ConvertMemberToOutsideCollaborator_LastOwner(t *testing.T) {
```
gmlewis: AFAICT, `TestOrganizationsService_ConvertMemberToOutsideCollaborator_LastOwner` and `TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMember` are identical, so I don't see the value in keeping them both.

Can you please delete one and maybe just add a comment that the two cases appear identical?

kshitij10496: > AFAICT, TestOrganizationsService_ConvertMemberToOutsideCollaborator_LastOwner and TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMember are identical, so I don't see the value in keeping them both.

I agree with you that the current implementation of the tests is identical in nature. My motivation behind writing two separate tests was to illustrate the fact that both the cases are handled differently by the API. The HTTP response codes are the same (403) but the `message` fields are different, based on the scenario.

kshitij10496: This was the reason I asked for some guideline in the preceding comment.

gmlewis: OK, I think a comment is sufficient to explain the situation. We don't need tests to show every possible response from the server, so please delete the duplicated test. As for how the remaining test is written, it LGTM as-is. An added comment can explain what is going on... it can be quick and short... no need to go into great detail.

Thanks, @kshitij10496!

kshitij10496:
@shurcooL I would love to have your feedback on this.

github/orgs_outside_collaborators_test.go
```diff
@@ -48,3 +48,81 @@ func TestOrganizationsService_ListOutsideCollaborators_invalidOrg(t *testing.T)
...
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusNotFound)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.RemoveOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusNotFound {
+		t.Errorf("Organizations.RemoveOutsideCollaborator_NonMember returned error: %v", err)
```
shurcooL: The `t.Errorf` error text here looks copied/pasted from above. But it doesn't accurately state what happened. It should be something like:

```Go
t.Errorf("Organizations.RemoveOutsideCollaborator did not return an error, or the status code wasn't 404 Not Found")
```

github/orgs_outside_collaborators_test.go
```diff
@@ -48,3 +48,81 @@ func TestOrganizationsService_ListOutsideCollaborators_invalidOrg(t *testing.T)
...
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusUnprocessableEntity)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.RemoveOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusUnprocessableEntity {
+		t.Errorf("Organizations.RemoveOutsideCollaborator_Member returned error: %v", err)
```
shurcooL: Same here, the `t.Errorf` text is misleading.

github/orgs_outside_collaborators_test.go
```diff
@@ -48,3 +48,81 @@ func TestOrganizationsService_ListOutsideCollaborators_invalidOrg(t *testing.T)
...
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		w.WriteHeader(http.StatusForbidden)
+	}
+	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
+
+	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(context.Background(), "o", "u")
+	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusForbidden {
+		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator_NonMember returned error: %v", err)
```
shurcooL: Also here.

kshitij10496:
> In general, we only add tests for the success case. It's pretty rare for GitHub API docs to document the error codes for error cases, but since they do, having unit tests for that seems okay.

I wrote the tests with the intention of ensuring that the library is able to handle the different cases where the GitHub API is returning 403s, 422s rather than returning a 404 to the end-user. 

A year ago, I was trying to write a similar wrapper library around an external API in Python. The client provided by my library used to return a 404 response whenever the API response was not 200. Since that incident, I have been over cautious about writing tests for all the different possible responses.

```
if api_res.status_code != 200: # BUG
    return ClientResponse("Not Found", 404)
``` 

Back to the PR: I have split the conditionals into `if-else` blocks and returning the corresponding error message. Let me know how we can improve this further.

github/orgs_outside_collaborators_test.go
```diff
@@ -122,7 +126,9 @@ func TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMemberOrLast
 	mux.HandleFunc("/orgs/o/outside_collaborators/u", handler)
 
 	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(context.Background(), "o", "u")
-	if err, ok := err.(*ErrorResponse); !ok || err.Response.StatusCode != http.StatusForbidden {
-		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator_NonMember returned error: %v", err)
+	if err, ok := err.(*ErrorResponse); !ok {
+		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator did not return an error")
+	} else if err.Response.StatusCode != http.StatusForbidden {
+		t.Errorf("Organizations.ConvertMemberToOutsideCollaborator did not return 403 Status code")
```
shurcooL: Nit; capital "S" in "Status code" here but not anyone else. It should be lowercase for consistency.

shurcooL:
There is a CI failure after I merged this PR. It's harmless and has nothing to do with the content of this PR, just that this happened to be the first PR merged in 2018. I've filed issue #823 about it.

PR #816, opened 2017-12-23
Migrate IDs to int64 type

kshitij10496:
Through this PR, I attempt to complete the migration of all the struct IDs(which refer to GitHub IDs) from platform dependent `int` type to the signed `int64`.

Also, a helper function `Int64` is implemented to assist the migration of unit tests.

### Checklist
- [x] Update `ID`s to `int64`
- [x] Updated unit tests
- [x] Updated Integration tests 
- [ ] Documentation 

Reference: #597


shurcooL:
> The issue is with the `go generate` command making some unwanted changes to [github/github-accessors.go](github/github-accessors.go).

I think the fix will involve turning `case "int"` into `case "int", "int64":` on [this line](https://github.com/google/go-github/blob/0c3b302de2a6de84a2511db47ea1bb2ff8146830/github/gen-accessors.go#L207).

anubhakushwaha:
@shurcooL  @kshitij10496 I am facing the same travis fail error in #817 and I tried your suggestion [link](https://github.com/anubhakushwaha/go-github/blob/6e4b5f4f2d41a56372b6efb7874dda17a8e78929/github/gen-accessors.go#L207) but it's still failing, any inputs?

anubhakushwaha:
It got resolved by fixing the formatting as instructed by @kshitij10496 .

kshitij10496:
@shurcooL I have made the necessary changes based on your suggestion.
Also, shouldn't we mention the fact that the client now uses `int64` ID values in the documentation?

shurcooL:
> Also, shouldn't we mention the fact that the client now uses `int64` ID values in the documentation?

Sure, if there's a good place for it. Can you think of one?

I don't think it's critical for getting this in. We can take care of it later if desired. In any case, it'll b something for reviewers to look out when doing future reviews (because some contributors can miss the documentation).

gmlewis:
Oh!  Let's please not forget to bump `libraryVersion` in this PR, @kshitij10496,  since this is a rather major change.

shurcooL:
> I can't remember now if you also wanted to add a type ID int64 to the code or leave it like this?

@gmlewis There was relevant discussion in the issue (#597), specifically this https://github.com/google/go-github/issues/597#issuecomment-353181253.

kshitij10496:
I've updated the `libraryVersion` in the `github.go` file.

github/event_types.go
```diff
@@ -486,7 +486,7 @@ type PullRequestEvent struct {
 	// the pull request was closed with unmerged commits. If the action is "closed"
 	// and the merged key is true, the pull request was merged.
 	Action      *string      `json:"action,omitempty"`
-	Number      *int         `json:"number,omitempty"`
+	Number      *int64       `json:"number,omitempty"`
```
shurcooL: Why is this `Number` field changed to `int64`? It's not an ID. I'm guessing this was an accident, right?

shurcooL:
This is a large PR. Thank you again @kshitij10496!

I've reviewed the fields that were modified first. They match up with the table in https://github.com/google/go-github/issues/597#issuecomment-353373725, with the exception of `PullRequestEvent.Number` that I asked about above, and 4 other ID fields that are valid but were missing in that table (I've added them now).

Then I reviewed the method parameters that were modified. This was harder to do, but the ones I saw modified look right. I can't be sure if any are missing though (but that's easier to fix in the future, if needed).

Overall, once we come to a resolution about `PullRequestEvent.Number`, I think this looks good. At least I'm not seeing anything wrong.

It's still a bit scary because this is going to be a relatively large breaking change.

@bradfitz Can you look this over and confirm if this PR looks as you'd expect it to for resolving #597?

github/event_types.go
kshitij10496: My bad; this is a mistake on my part. The `Number` field should not be changed.
Let me rectify it.

bradfitz:
I'd prefer if fixes within this PR didn't accumulate in new commits (like https://github.com/google/go-github/pull/816/commits/4c02471a559afd6741fef4587040c70253a150ba).

I'd just rebase/squash mistakes and have commits that stand alone. It'll be make git archaeology in the future easier.

shurcooL:
@bradfitz We (almost always) merge PRs via squash strategy, so the entire diff will be just one commit once it lands to `master` (e.g., see https://github.com/google/go-github/pull/812/commits, which was 6 commits that got merged as a single commit 218ecc238cf77bbaa2e16a54cada252ec87e7d6e).

We encourage contributors not to squash during the course of the PR, because that makes it harder to review and see what has changed since last review (this isn't Gerrit; you can't see old versions of a PR).

Does that address your concern?

github/event_types.go
```diff
@@ -486,7 +486,7 @@ type PullRequestEvent struct {
 	// the pull request was closed with unmerged commits. If the action is "closed"
 	// and the merged key is true, the pull request was merged.
 	Action      *string      `json:"action,omitempty"`
-	Number      *int64       `json:"number,omitempty"`
+	Number      *int         `json:"number,omitempty"`
```
shurcooL: @kshitij10496 You'll need to re-run `go generate` after this change to make CI pass.

kshitij10496: Thanks for the heads up.

kshitij10496:
@shurcooL I have updated the PR after generating the code. Thanks for the suggestion.
@bradfitz If you have any other suggestions related to this PR, I would be more than happy to work on it. 😄

bradfitz:
> We (almost always) merge PRs via squash strategy

Ah, great. You can tell I don't use GitHub PRs much, huh? :)

kshitij10496:
@shurcooL Is there anything else that we need to address in this PR related to the issue?

shurcooL:
One extra thing I can do to increase confidence before merging is apply this PR locally and see how it looks on all my projects that import go-github. There’s a chance it’ll help catch something we can’t see just by looking at the diff.

shurcooL:
Checked it out locally, it went well. Out of the 7 packages I have that [directly import go-github](https://godoc.org/github.com/google/go-github/github?importers), I only needed to update [one type conversion in one package](https://github.com/shurcooL/users/blob/3500fa32fc1c1bce424c831ce06357dd3b2ddab5/githubapi/githubapi.go#L54). I really expected there to be more (but of course other packages may vary).

It's probably nice to wait until Monday (so people who don't vendor their dependencies can enjoy their weekend without breaking API changes), but otherwise we can merge this anytime if there are no further objections.

shurcooL:
Ping @gmlewis. Shall we proceed?

gmlewis:
Yes, LGTM.
Merging.

kshitij10496:
I want to thank everyone involved in the discussion on #597 which immensely helped me with the implementation.

gmlewis:
Thank you for the PR, @kshitij10496!

PR #820, opened 2017-12-29
Add PullRequest.MergeableState field

ains:
This isn't a well documented field but it's present in API responses and mentioned here 
https://github.com/octokit/octokit.rb/issues/626

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -5676,6 +5676,14 @@ func (p *PullRequest) GetMergeable() bool {
 	return *p.Mergeable
 }
 
+// GetMergeableState returns the MergeableState field if it's non-nil, zero value otherwise.
+func (p *PullRequest) GetMergeableState() string {
+	if p == nil || p.MergeableState == nil {
+		return ""
+	}
+	return *p.MergeableState
+}
+
 // GetMergeCommitSHA returns the MergeCommitSHA field if it's non-nil, zero value otherwise.
 func (p *PullRequest) GetMergeCommitSHA() string {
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @ains. I have a couple questions for you.

According to the documentation here: https://developer.github.com/v3/pulls/#get-a-single-pull-request
the `mergeable` field ought to be sufficient, and `mergeable_state` is nowhere to be found (in the docs) but you say you have seen it in responses.

Have you tried using the `mergeable` field to see if it satisfies your needs?

If you have tried it, but it doesn't work, then I think we need to contact support@github.com and ask for clarification regarding these two fields (one documented, one not).
If this is the case, can you please provide a JSON response showing both the `mergeable` and `mergeable_state` fields and show how `mergeable` is not sufficient? This will help the GitHub support team figure out what is going on.

Thanks again!

ains:
No problem @gmlewis 

You can find a bit more context on the field `mergeable_state` in this issue https://github.com/octokit/octokit.rb/issues/626, 

and in this thread on the github platform forum (where they are planning to document it) 
https://platform.github.community/t/documentation-about-mergeable-state/4259

Unfortunately mergeable doesn't satisfy the needs we have, if a PR can cleanly merge into a branch but fails a check (set when protecting that branch) `mergeable` is still `true` however `mergeable_state` will be `blocked`, indicating it actually shouldn't be merged.

gmlewis:
Ah, thank you for the forum link, @ains! That is sufficient for me.
LGTM.

PR #832, opened 2018-01-14
Make use of testMethod consistent

### LINKED ISSUE #829: Make use of testHeader and testMethod consistent
As found out while reviewing https://github.com/google/go-github/pull/817, we have some inconsistencies in our tests where some of the tests don't use the `testHeader` and `testMethod` functions to test request header and method.

One of the defaulter is https://github.com/google/go-github/blob/master/github/git_trees_test.go#L22.

This is completely harmless. Though it's important to make sure it doesn't cause further confusion for the new contributors as it did in that PR initially.

Via this issue, I'd like to point it out and mention that I'd like to fix this inconsistency.

sahildua2305:
Use testMethod to test the request method instead of manually comparing
it and testing. This change is to make things consistent through out the
code base.

Fixes #829

### DIFF
--- a/github/git_trees_test.go
+++ b/github/git_trees_test.go
@@ -19,9 +19,7 @@ func TestGitService_GetTree(t *testing.T) {
 	defer teardown()
 
 	mux.HandleFunc("/repos/o/r/git/trees/s", func(w http.ResponseWriter, r *http.Request) {
-		if m := "GET"; m != r.Method {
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @shurcooL!
Merging.

sahildua2305:
Thanks @gmlewis and @shurcooL

Issue #814, opened 2017-12-21
Support preview GraphQL API v4 Node IDs in REST API v3 resources


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2017-12-19-graphql-node-id/

To complete this issue:

* a new media type (`application/vnd.github.jean-grey-preview+json`) must be added to `github.go`,
* this media type must be added to the header of calls to all affected endpoints (see announcement),
* a new `node_id` field must be added to the appropriate responses structs that are returned.

This issue presents a great learning opportunity for a new contributor to participate and interact with this community of passionate and enthusiastic Go developers. Any and all contributions are greatly appreciated!

If you would like to work on this issue, please comment below and our team will send you an invite so we can assign the issue to you. Make sure to check out the [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) documentation for help on getting started. Thank you!

anubhakushwaha:
@gmlewis Can I take this up?

gmlewis:
Definitely!  Thank you, @anubhakushwaha!  I sent you an invite.

joho:
I'm currently blocked on a project as I need `node_id` on PullRequest. 

I'd be happy to submit a PR adding it, but don't want to duplicate any of @anubhakushwaha's work in case there's a second PR in progress (btw thanks a lot for the first one). Let me know if it's wanted/needed.

parkr:
`node_id`'s are also present in webhook payloads, e.g. the `push` event has a `repository.node_id`, `sender.node_id`, etc.

palash25:
Hey @gmlewis I see that the following resources were not covered by the PR #817 
Projects
Pull Requests
Reactions
Releases
Repositories
Statuses
Teams
Users

I would like to take this up to complete the rest of the work :smile:

vaibhavsingh97:
@gmlewis Since @palash25 haven't sent PR, I would love to add resources mentioned by him.

gmlewis:
Let's give @palash25 a day or two to respond and afterward you can take this, @vaibhavsingh97.

palash25:
@gmlewis I am busy preparing for gsoc currently so @vaibhavsingh97 can take this up

gmlewis:
Good luck with Google Summer of Code, @palash25!

OK, this is yours, @vaibhavsingh97.

vaibhavsingh97:
[...]

PR #833, opened 2018-01-21
Add NodeID for grapqhl preview to PullRequest

joho:
A supporting PR for #814 adding `NodeID` for `PullRequest` and `NewPullRequest`.

Not trying to steal @anubhakushwaha's thunder, but did it in a local fork this morning anyway so I could make progress on another project.


joho:
I signed it!

anubhakushwaha:
@joho I was struck with something, and it's open source, we build together, thanks for doing the work, I will also send the rest of the work in a day or two for sure :smile:

github/pulls.go
```diff
@@ -174,6 +181,7 @@ type NewPullRequest struct {
 	Body                *string `json:"body,omitempty"`
 	Issue               *int    `json:"issue,omitempty"`
 	MaintainerCanModify *bool   `json:"maintainer_can_modify,omitempty"`
+	NodeID              *string `json:"node_id,omitempty"`
```
shurcooL: I don't think `NewPullRequest` struct should get the `NodeID` field, should it?

It's a struct we use to pass parameters to [`PullRequests.Create`](https://godoc.org/github.com/google/go-github/github#PullRequestsService.Create) method. You can't specify the `node_id` when creating a new PR (i.e., https://developer.github.com/v3/pulls/#create-a-pull-request).

joho: Ah, sorry about that, I'd misread the flow in Create - `NewPullRequest` goes in, regular `PullRequest` comes out. Fixed.

github/pulls_test.go
```diff
@@ -59,6 +60,7 @@ func TestPullRequestsService_Get(t *testing.T) {
 
 	mux.HandleFunc("/repos/o/r/pulls/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", mediaTypeGraphQLNodeIDPreview)
```
shurcooL: Maybe also do this for `TestPullRequestsService_Create` and `TestPullRequestsService_Edit` below, unless there's a reason not to.

joho: Couldn't think of a reason not to, so I did.

PR #837, opened 2018-01-25
Support for custom media types for Repository contents

### LINKED ISSUE #727: How to render readme content as HTML?
/* func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string, opt *RepositoryContentGetOptions) (*RepositoryContent, *Response, error) */
this func cannot set http header "application/vnd.github.VERSION.html+json".

kshitij10496:
This is a proof-of-concept of how we can extend support for custom mime types across all services.
Reviews, suggestions, and discussions on the implementation are welcome. 

- [x] `GetReadme` supports `html`, `raw` and `object`(default) media types
- [ ] `GetContents` supports custom media types
- [ ] `DownloadContents` supports custom media types
- [x] `GetArchiveLink` should support only the `ref` parameter.

Fixes #727 

**NOTE:** This change breaks the backward compatibility of `GetArchiveLink` as the new `opt` is of type `RepositoryContentsGetParameters` instead of `RepositoryContentsGetOptions`. One way to overcome this issue is to revert the naming change and rename the new `RepositoryContentsGetOptions` to something else.

Ping @gmlewis @sahildua2305

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -9,6 +9,7 @@
 package github
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
@@ -52,6 +53,14 @@ type RepositoryContentFileOptions struct {
 	Author    *CommitAuthor `json:"author,omitempty"`
 	Committer *CommitAuthor `json:"committer,omitempty"`
 }
+
+type RepositoryContentMediaType uint8
+
+const (
+	Object RepositoryContentMediaType = iota
+	Raw
+	HTML
+)
 
 // RepositoryContentGetOptions represents an optional ref parameter, which can be a SHA,
 // branch, or tag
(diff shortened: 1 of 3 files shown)


github/repos_contents.go
```diff
@@ -53,12 +54,26 @@ type RepositoryContentFileOptions struct {
...
-type RepositoryContentGetOptions struct {
+type RepositoryContentParameters struct {
 	Ref string `url:"ref,omitempty"`
 }
 
+// RepositoryContentGetOptions represents the optional custom media type and ref parameters.
+type RepositoryContentGetOptions struct {
+	Params    RepositoryContentParameters
+	MediaType RepositoryContentMediaType
+}
```
sahildua2305: @kshitij10496 how about having this instead?

```
type RepositoryContentGetOptions struct {
 	Ref       string `url:"ref,omitempty"`
 	MediaType RepositoryContentMediaType
 }
```

This will ensure the API's backward compatibility and any new options (in future) can still be added to `RepositoryContentGetOptions` struct instead of `RepositoryContentParameters`. What do you think?

github/repos_contents.go
```diff
@@ -106,6 +121,46 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
 	return readme, resp, nil
 }
 
+// GetReadme gets the Readme file for the repository.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/contents/#get-the-readme
+func (s *RepositoriesService) GetReadmeCustom(ctx context.Context, owner, repo string, opt *RepositoryContentGetOptions) (string, *Response, error) {
```
sahildua2305: 👍 for `GetReadmeCustom` instead of having different methods like `GetReadmeRaw`, `GetReadmeHTML`, etc. I read your comment about this suggestion on the related issue.

github/repos_contents.go
```diff
@@ -106,6 +121,46 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
...
+	case HTML:
+		req.Header.Set("Accept", mediaTypeV3HTML)
+	case Object:
+		req.Header.Set("Accept", mediaTypeV3Object)
+		repoContent, resp, err := s.GetReadme(ctx, owner, repo, opt)
+		if err != nil {
+			return "", nil, err
+		}
+		content, err := repoContent.GetContent()
+		return content, resp, err
```
sahildua2305: L145-L151 isn't very readable. In my opinion, it will confuse any future contributors to understand what's going on in this usual piece of code.

I'll need to give it a thought about how to improve this. Let me know if you have any suggestions.

github/repos_contents.go
sahildua2305: I see that it was suggested in your approach #2 on the issue. However, I don't understand the cons of this approach.

kshitij10496: The issues we might face if on proceeding with Approach #02:
1. The `addOptions` function expects a value of struct type with url tags. Thus, we will have to declare another type to handle this for us.
2. As a result of the above impediment, adding a new parameter to the `RepositoryContentsGetOptions` in the future will require a lot of corresponding changes to be made.

At a fundamental level, is it a good practice to encapsulate the request parameters and the request headers into a single type value as raw data rather than composing them from simpler structs?

gmlewis:
Closing as there did not seem to be much discussion or demand for this from contributors or users (in addition to having several open questions).

PR #844, opened 2018-02-05
Remove preview media type for Team Review Requests

### LINKED ISSUE #835: Team Review Requests API becomes an official part of API v3
GitHub Announcement:
https://developer.github.com/changes/2018-01-24-end-thor-preview/

The preview period has ended, and we will no longer need to pass the `mediaTypeTeamReviewPreview` custom media type. Instead, GitHub recommends that we specify `v3` as the version in the Accept header. However, note that some GitHub Enterprise versions do not yet support the Team Review Requests API, thus, will still require the custom media type.

In continuation to #678

### LINKED ISSUE #840: Automating the sync between GitHub API Blog and our library
I noticed that we currently have to manually create a New Issue every time there is an announcement from [GitHub API Blog](https://developer.github.com/changes/) (usually on a weekly basis). How about we automate this bookkeeping process so as to save some effort?

I would like to share a way we can implement this:
Step 1: Using an RSS feed parser (such as [gofeed](https://github.com/mmcdole/gofeed)) to notify about updates 
Step 2: Using `go-github` itself to create new issues every time there is an update.

Currently, I intend to implement this a simple program or script which one can run as an hourly cronjob. We can create a bot account similar to `googlebot` which reports these updates on our Issues tracker.

This will ease the load of the project maintainers, especially @gmlewis from having to manually keep track of updates from GitHub. I might have overlooked some important details in the Proof-Of-Concept above but I would love to listen to what the developers think about this proposal and whether this would be a useful addition to the project.

Ping @willnorris @gmlewis @shurcooL @sahildua2305 @anubhakushwaha

kshitij10496:
As the Team Review Requests API is fully supported by GitHub API v3, we
have removed the preview(custom) media type: `thor-preview`.

Fixes #835 

-------------

1. How should we proceed here given that some GitHub enterprise versions may not yet support the Team Review Requests API?
2. According to the new versioning guideline, under which category should be removing preview functionality in favor of GitHub API support fall?

Ping @willnorris @gmlewis @sahildua2305

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -99,9 +99,6 @@ const (
 	// https://developer.github.com/changes/2017-07-17-update-topics-on-repositories/
 	mediaTypeTopicsPreview = "application/vnd.github.mercy-preview+json"
 
-	// https://developer.github.com/changes/2017-07-26-team-review-request-thor-preview/
-	mediaTypeTeamReviewPreview = "application/vnd.github.thor-preview+json"
-
 	// https://developer.github.com/v3/apps/marketplace/
 	mediaTypeMarketplacePreview = "application/vnd.github.valkyrie-preview+json"
 
--- a/github/pulls_reviewers.go
+++ b/github/pulls_reviewers.go
@@ -32,9 +32,6 @@ func (s *PullRequestsService) RequestReviewers(ctx context.Context, owner, repo 
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeTeamReviewPreview)
-
 	r := new(PullRequest)
 	resp, err := s.client.Do(ctx, req, r)
 	if err != nil {
@@ -59,9 +56,6 @@ func (s *PullRequestsService) ListReviewers(ctx context.Context, owner, repo str
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeTeamReviewPreview)
-
 	reviewers := new(Reviewers)
 	resp, err := s.client.Do(ctx, req, reviewers)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


gmlewis:
Re: question 1: GitHub Enterprise customers must now pin specific tagged releases of this package and we can move forward without hesitation due to #376.

Re: question 2: According to our [new versioning strategy](https://github.com/google/go-github/issues/376#issuecomment-283110291), the tagged minor version would be bumped in this case.

gmlewis:
In your first comment above, you wrote Closes #840 - but did you mean "Fixes #835" instead?

kshitij10496:
> In your first comment above, you wrote Closes #840 - but did you mean "Fixes #835" instead?

Yes, I meant to refer #835  instead of #840. Thanks for pointing this out.
I will be more mindful while writing the commit message.
I think I should rectify the commit message accordingly and force push the change, shouldn't I?

gmlewis:
I think you can simply edit that first message.  A force push should not be necessary because I can make sure the commit message has the right text when merging... as long as I remember to do that correctly. :smile:

kshitij10496:
Sure! That's awesome.

kshitij10496:
> I'll let this sit for a day in case there are any other comments, then merge tomorrow (unless another reviewer LGTM's and merges before then). Feel free to ping me if I forget to merge tomorrow.

This makes sense. 
How about we formulate a guideline to merge any change(as trivial as this PR) after at least 24 hours have passed from the moment of the patch being sent, so as to give everyone time to review the work? I remember @shurcooL raising a related point in a [comment](https://github.com/google/go-github/pull/816#issuecomment-359204097). Thus, I think we can mention this formally in the Contributing document.

gmlewis:
I think it is a good idea in general to wait for comments/conversation on a PR, but am not sure that we need to formalize it.

Sometimes there are good reasons to merge things ASAP... I think it might keep things a bit more flexible if we leave it up to the maintainers of the repo to decide on a PR-by-PR basis.

gmlewis:
Sorry for the delay... Merging.

PR #847, opened 2018-02-07
Remove unnecessary draining the `resp.Body`

### LINKED ISSUE #843: [question] Drain up to 512 bytes and close the body to let the Transport reuse the connection
Hi, do you think we still need following lines today (with current and newer golang version)?

https://github.com/google/go-github/blob/master/github/github.go#L481-L485

Thanks!

diorahman:
The behavior on connection reuse is introduced in golang/go@18072ad.

Fixes #843

---
This change is [Reviewable](https://reviewable.io/reviews/google/go-github/847)

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -477,8 +477,8 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 		return nil, err
 	}
-
 	defer resp.Body.Close()
+
 	response := newResponse(resp)
 
 	c.rateMu.Lock()
--- a/github/github.go
+++ b/github/github.go
@@ -478,7 +478,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	resp.Body.Close()
+	defer resp.Body.Close()
 	response := newResponse(resp)
 
 	c.rateMu.Lock()
--- a/github/github.go
+++ b/github/github.go
@@ -478,12 +478,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	defer func() {
-		// Drain up to 512 bytes and close the body to let the Transport reuse the connection
-		io.CopyN(ioutil.Discard, resp.Body, 512)
-		resp.Body.Close()
-	}()
-
+	resp.Body.Close()
 	response := newResponse(resp)
 
 	c.rateMu.Lock()


github/github.go
```diff
@@ -478,12 +478,6 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	defer func() {
-		// Drain up to 512 bytes and close the body to let the Transport reuse the connection
-		io.CopyN(ioutil.Discard, resp.Body, 512)
-		resp.Body.Close()
```
gmlewis: I believe that we should go back to how this looked prior to the addition of the drain:
https://github.com/google/go-github/commit/e0ff7111b024fda991e457d975e0b73e1380f05f

Specifically, this should now be:

```go
defer resp.Body.Close()
```

Note that this is an HTTP client, and [its documentation](https://godoc.corp.google.com/pkg/net/http) says:

```
The client must close the response body when finished with it:
...<example elided>...
```

gmlewis:
This also means that the title of the PR is not correct. The response body does indeed need to be closed.

github/github.go
```diff
@@ -478,12 +478,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
-	}()
-
+	resp.Body.Close()
```
sahildua2305: @diorahman This still needs a correction.

As @gmlewis mentioned, this should be same as how this looked prior to the addition of the drain:
https://github.com/google/go-github/commit/e0ff7111b024fda991e457d975e0b73e1380f05f

That means, you should `defer` this statement so that it executes only when the relevant routine returns.

Please change it to:

```go
defer resp.Body.Close()
```

github/github.go
```diff
@@ -478,12 +478,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	defer func() {
-		// Drain up to 512 bytes and close the body to let the Transport reuse the connection
-		io.CopyN(ioutil.Discard, resp.Body, 512)
-		resp.Body.Close()
-	}()
-
+	defer resp.Body.Close()
```
gmlewis: nit: I would like to see a blank line between the `defer` and the `newResponse`... since they aren't closely related.

...and while you're add it, I would like to tightly bind the defer to the `c.client.Do` above, so if you would like to get rid of the blank line above (line 480), that would be fantastic, IMHO.

Thanks, @diorahman!

diorahman: Ah OK. Thanks @gmlewis for the review. I updated it on https://github.com/google/go-github/pull/847/commits/d285239f10debdeece9710b1b15449b03bf9e41c.

Issue #822, opened 2018-01-04
CLI tool in examples


FenwickElliott:
### Use Case
My main usage of this project is with a small CLI tool I use to (primarily) create a Github repo which I can push to without leaving my IDE.

For example (within working directory):

```
git init
go-git-cli new example
git remote add origin https://github.com/FenwickElliott/example.git
```
### Questions
- Would the community liked this use case added to the examples directory?
- For authorization I use a package that gets an OAuth2 token and saves it locally (in Application Storage). Is that acceptable or is there another method of authorization is should use?

gmlewis:
I like having more examples, and this sounds like a useful one!
The OAuth2 solution sounds good to me from the description.
Thank you for the offer, @FenwickElliott!

FenwickElliott:
@gmlewis I finally got around to writing this up. Would you mind glancing over it before I submit it as  a pull request?

https://github.com/FenwickElliott/go-github/blob/cli/example/newrepo/main.go

(getting the OAuth2 token involved too much tangental material so I am just importing it as a environment variable)

Thanks regardless

gmlewis:
Thanks, @FenwickElliott!

I've looked it over and have a bunch of code-review comments, but can't make them with that link... overall, I think that this will be a good addition to have... so if you are willing to go through the process of a code review via a PR, then I'm happy to review it and get it into this repo after we've made some changes.

Thanks again!

FenwickElliott:
@gmlewis, 

Great, thank you. I would be very grateful for your review.

I'll submit the PR when I get back to my desk.


Thanks again,

PR #848, opened 2018-02-07
CLI newrepo example

FenwickElliott:
A basic use case utilizing go-github as a CLI tool with a single executable: newrepo, which takes an Auth token as an environment variable and creates a new repo on the owners GitHub account.

Fixes: #822


example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The newrepo command utilizes go-github as a cli tool for
+// creating new repositories.
+
```
gmlewis: Please remove this blank line so that Godoc can show the above comment as the package's main documentation.

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+	"os"
+	"strings"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+func main() {
+	token := os.Getenv("GITHUB_AUTH_TOKEN")
+	name := strings.Join(os.Args[1:], "-")
```
gmlewis: This seems odd... why would a person call this with space-separated strings and not just hyphenate the single string themselves if that is the repo name they wanted?

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+	if token == "" {
+		fmt.Println("Unauthorized: No token present")
```
gmlewis: Why not use `log.Fatal` here and get rid of the `return` on the next line?
In addition to being more concise, another nice feature is that it will return an error code that could be tested in shell scripts or other calling programs.

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+		return
+	}
+	if name == "" {
+		fmt.Println("No name: New repos must be given a name")
```
gmlewis: Same here.

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+	repo := &github.Repository{Name: github.String(name)}
+
+	repo, _, err := client.Repositories.Create(ctx, "", repo)
+	if err != nil {
+		fmt.Println("Error:", err)
+	}
+	fmt.Println("Sucsesfully created new repo: ", *repo.Name)
```
gmlewis: s/Sucsesfully/Successfully/

Also, why not:

```go
fmt.Printf("Successfully created new repo: %v\n", repo.GetName())
```

?

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+		fmt.Println("No name: New repos must be given a name")
+		return
+	}
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+	repo := &github.Repository{Name: github.String(name)}
+
+	repo, _, err := client.Repositories.Create(ctx, "", repo)
```
gmlewis: Please add a comment (either here or in the top-level comment describing this CLI tool) that the `""` signifies that the GitHub owner of the new repo will be the authorized user's GitHub account.

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	if name == "" {
+		fmt.Println("No name: New repos must be given a name")
+		return
+	}
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+	repo := &github.Repository{Name: github.String(name)}
```
gmlewis: Oh, one more thing... using `repo` here and then again below might be slightly confusing to someone... What do you think about making this `r` instead?  Your call.

example/newrepo/main.go
FenwickElliott: I agree, ideally users would only call this with a single argument but joining arguments felt like the most elegant way of handling cases where:
- There are no command line arguments (will panic if not handled),
- There are more than one command line arguments (subsequent arguments will be disregarded),

Do you recommend keeping it simple with `name := os.Args[1:][0]` and removing `if name == "" {
		log.Fatal("No name: New repos must be given a name")
	}
`? I could of course check `len(os.Args) > 1` but would have to do so before accessing `os.Args[1:][0]` so joining felt neater, albeit imperfect.

Let me know what you think. I really appreciate your input as I am still very new to the environment.

FenwickElliott:
@gmlewis,

Thanks for all your input. I have incorporated all your suggestions with the exception of joining the argument strings which I have answered separately.

One further question. To make this cli practice one can export their auth token from their path. Should I mention that somewhere or should I trust users to know how best to use this tool by examining it?

example/newrepo/main.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	if name == "" {
+		log.Fatal("No name: New repos must be given a name")
+	}
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+	r := &github.Repository{Name: github.String(name)}
+
```
gmlewis: Since `r` is tightly coupled to the `Create` call on line 38, please delete this blank line (line 37).
It would make total sense to me to move the blank line above line 36 since all the earlier stuff is setup.

example/newrepo/main.go
gmlewis: If it were me, I would make it a flag.

Something like:

```go
var repoName = flag.String("repo", "", "Name of repo to create in authenticated user's GitHub account.")

func main() {
  flag.Parse()
  if token == "" {...}
  if *repoName == "" {
...
```

and then get rid of `name` and `os.Args` entirely.

example/newrepo/main.go
```diff
@@ -11,30 +11,36 @@ package main
...
 	"fmt"
 	"log"
 	"os"
-	"strings"
 
 	"github.com/google/go-github/github"
 	"golang.org/x/oauth2"
 )
 
+var name = flag.String("name", "", "Name of repo to create in authenticated user's GitHub account.")
```
gmlewis: nit: to follow Go style, please place these in a `var` block...

```go
var (
  name = ...
  description = ...
  private = ...
)
```

example/newrepo/main.go
```diff
@@ -11,30 +11,36 @@ package main
...
+	if *name == "" {
 		log.Fatal("No name: New repos must be given a name")
 	}
 	ctx := context.Background()
 	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
 	tc := oauth2.NewClient(ctx, ts)
 	client := github.NewClient(tc)
-	r := &github.Repository{Name: github.String(name)}
 
+	r := &github.Repository{Name: github.String(*name)}
```
gmlewis: Since the flags are already pointers, `github.String` and `github.Bool` in these 3 lines don't need to be used at all.

```go
r := &github.Repository{Name: name, Private: private, Description: description}
```

FenwickElliott:
Thanks @gmlewis,

Sorry for missing such obvious things, as I say, I'm new to Golang.

gmlewis:
Not obvious at all, @FenwickElliott!  These points are rather subtle. Thanks for doing that... you are doing great and your help is really appreciated.  I'll take a look.

PR #850, opened 2018-02-13
Support preview Team Discussions API 

### LINKED ISSUE #849: Support preview Team Discussions API
https://developer.github.com/changes/2018-02-07-team-discussions-api/

Feel free to volunteer for any issue, and we can send you an invite to contribute to the repo (which you then accept) and the issue can be assigned to you so that others don't attempt to duplicate the work.

Make sure you have two-factor authentication enabled on GitHub... as we will be requiring it soon to be a contributor. It's a good idea anyway... in general... not just for GitHub!  :smile: 

Thank you!

parkhyukjun89:
Added new API introduced in https://developer.github.com/changes/2018-02-07-team-discussions-api/

Moved all Teams API from OrganizationsService per https://developer.github.com/changes/2018-02-07-team-discussions-api/#updates-to-the-teams-api-documentation

Fixes #849


github/github.go
```diff
@@ -155,6 +158,7 @@ type Client struct {
 	Repositories   *RepositoriesService
 	Search         *SearchService
 	Users          *UsersService
+	Teams          *TeamsService
```
gmlewis: Please move this above `Users` and after `Search` to keep this list sorted.

github/github.go
```diff
@@ -245,6 +249,7 @@ func NewClient(httpClient *http.Client) *Client {
 	c.Repositories = (*RepositoriesService)(&c.common)
 	c.Search = (*SearchService)(&c.common)
 	c.Users = (*UsersService)(&c.common)
+	c.Teams = (*TeamsService)(&c.common)
```
gmlewis: Same here... please move up one line.

github/teams.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2013 The go-github AUTHORS. All rights reserved.
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
```
gmlewis: Please revert the copyright date back to the original `2013` value.
Any new files that are added to the repo would get a new `2018` date, but since this file moved, it retains its original date. If people are curious, git can show them the origins (and complete change history) of this file.

github/teams_discussion_comments_test.go
```diff
@@ -0,0 +1,203 @@
...
+	})
+
+	comment, _, err := client.Teams.GetComment(context.Background(), 2, 3, 4)
+	if err != nil {
+		t.Errorf("Teams.GetComment returned error: %v", err)
+	}
+
+	want := &DiscussionComment{Number: Int64(4)}
+	if !reflect.DeepEqual(comment, want) {
+		t.Errorf("ITeams.GetComment returned %+v, want %+v", comment, want)
```
gmlewis: s/ITeams/Teams/

github/teams_discussions_test.go
```diff
@@ -0,0 +1,212 @@
...
+	})
+
+	discussion, _, err := client.Teams.GetDiscussion(context.Background(), 2, 3)
+	if err != nil {
+		t.Errorf("Teams.GetDiscussion returned error: %v", err)
+	}
+
+	want := &TeamDiscussion{Number: Int64(3)}
+	if !reflect.DeepEqual(discussion, want) {
+		t.Errorf("ITeams.GetDiscussion returned %+v, want %+v", discussion, want)
```
gmlewis: s/ITeams/Teams/

github/teams_members.go
```diff
@@ -0,0 +1,174 @@
...
+	t := new(Membership)
+	resp, err := s.client.Do(ctx, req, t)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return t, resp, nil
+}
+
+// TeamAddTeamMembershipOptions does stuff specifies the optional
```
gmlewis: s/does stuff//
(I realize you didn't do this... I see it in the original, but now would be a great time to fix it.  :smile: ... thank you!)

parkhyukjun89:
@gmlewis Thanks for the review! I have made changes that you pointed out.

github/teams_discussions_test.go
```diff
@@ -0,0 +1,212 @@
...
+					},
+					"body": "test",
+					"body_html": "<p>test</p>",
+					"body_version": "version",
+					"comments_count": 1,
+					"comments_url": "https://api.github.com/teams/2/discussions/3/comments",
+					"created_at": "2018-01-01T00:00:00Z",
+					"last_edited_at": null,
+					"html_url": "https://github.com/orgs/1/teams/2/discussions/3",
+					"node_id": "node",
```
elliott-beach: I think this is too much boilerplate - the practice in other portions of the API is to only spell out a minimal response in unit tests.  Example: https://github.com/google/go-github/blob/12126fbfe000a09047ebdeb2b5160be3af88ab9e/github/pulls_test.go#L225-L228 Mainly for consistency, I would recommend only specifying a minimal few fields to get working tests. Also, writing out every field can increase the cost of maintenance. Does that make sense to you?

github/teams_memebers_test.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
```
elliott-beach: minor type: Rename this to `team_members_test.go`

github/teams_discussion_comments_test.go
```diff
@@ -0,0 +1,203 @@
...
+				ID:                Int64(0),
+				AvatarURL:         String("https://avatars1.githubusercontent.com/u/0?v=4"),
+				GravatarID:        String(""),
+				URL:               String("https://api.github.com/users/author"),
+				HTMLURL:           String("https://github.com/author"),
+				FollowersURL:      String("https://api.github.com/users/author/followers"),
+				FollowingURL:      String("https://api.github.com/users/author/following{/other_user}"),
+				GistsURL:          String("https://api.github.com/users/author/gists{/gist_id}"),
+				StarredURL:        String("https://api.github.com/users/author/starred{/owner}{/repo}"),
+				SubscriptionsURL:  String("https://api.github.com/users/author/subscriptions"),
```
elliott-beach: Same point here.

github/teams_discussions_test.go
parkhyukjun89: Shouldn't newly added struct be tested for its ability to parse a full response? I made sure to do it only once, and that is why Get* tests only have minimal fields. I can remove Author field as that should be covered by users_test.

github/teams_memebers_test.go
parkhyukjun89: I will fix `memebers`, but actual service name is `teams`. Shouldn't the file name start with `teams` then?

parkhyukjun89:
@gmlewis changes are pushed.
FYI, orgs_members.go is not a part of team discussion change set. It was included in the commit d6d7f62 due to pulling upstream to resolve merge conflicts.

shurcooL:
Thank you for working on this. This is a large change. From a brief high-level look, it looks right.

Due to its large size and the fact it's a breaking change, it's quite hard to review and feel confident about merging. This is especially so because the new Team Discussions API is considered in preview, and hence may change, and be less stable.

However, as I understand, the Team API that is being moved from `OrganizationsService` was also a preview API rather than stable.

One possibility I want to propose is to split this PR into 2 parts that are easier to review, if possible.

Part 1 would add the new endpoints but keep the previous API (under `OrganizationsService`) around, but have it call out to the new `TeamsService` methods. Also mark them as deprecated.

That way, the diff would be mostly additive and it would not be a breaking change.

Part 2 would be to remove the deprecated methods from `OrganizationsService`.

So, instead of:

```diff
 // ListTeams lists all of the teams for an organization.
 //
-// GitHub API docs: https://developer.github.com/v3/orgs/teams/#list-teams
-func (s *OrganizationsService) ListTeams(ctx context.Context, org string, opt *ListOptions) ([]*Team, *Response, error) {
+// GitHub API docs: https://developer.github.com/v3/teams/#list-teams
+func (s *TeamsService) ListTeams(ctx context.Context, org string, opt *ListOptions) ([]*Team, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/teams", org)
 	u, err := addOptions(u, opt)
 	if err != nil {
```

It would be:

```diff
 // ListTeams lists all of the teams for an organization.
+// Deprecated: Use TeamsService.ListTeams instead.
 //
 // GitHub API docs: https://developer.github.com/v3/orgs/teams/#list-teams
 func (s *OrganizationsService) ListTeams(ctx context.Context, org string, opt *ListOptions) ([]*Team, *Response, error) {
-	u := fmt.Sprintf("orgs/%v/teams", org)
-	u, err := addOptions(u, opt)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	req, err := s.client.NewRequest("GET", u, nil)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeNestedTeamsPreview)
-
-	var teams []*Team
-	resp, err := s.client.Do(ctx, req, &teams)
-	if err != nil {
-		return nil, resp, err
-	}
-
-	return teams, resp, nil
+	return s.teamsService.ListTeams(ctx, org, opt)
 }
```

What do you think of such an approach?

If you think this is viable and a good idea, give it a shot.

If you don't think it's viable, or not worth the extra effort because this breaking API change affects only a preview API, then we can keep the PR as is and I'll try to review it as is.

Thanks again.

shurcooL:
About the idea above, I'm starting to think it's not worth it. The extra overhead of doing the change and then a 2nd change to remove deprecated methods is probably too much and not justified, given these are all preview API endpoints.

parkhyukjun89:
It would depend on API deprecation approach, but I was thinking that preview API wouldn't adhere to it anyways.

Though I think it'd be better to remove migrated APIs while they are in preview mode, I agree that this PR is large. Would it make more sense to split this PR into 2, one for team discussions API and one for API migration?

shurcooL:
> Though I think it'd be better to remove migrated APIs while they are in preview mode,

I had that in mind. The goal of deprecating was just to split the large PR into smaller, easier to review ones, not to keep the deprecated old preview API.

> Would it make more sense to split this PR into 2, one for team discussions API and one for API migration?

If you think it'll make it easier to review and you don't mind doing so, feel free. Thanks.

parkhyukjun89:
Closing this PR to split it into smaller PRs

Issue #851, opened 2018-02-14
Return app_id for an Installation


sridharavinash:
I noticed that `get single installation` endpoint query here https://github.com/google/go-github/blob/master/github/apps.go#L100 returns an `Installation` but without an `app_id`. The `app_id` is returned in the call to https://developer.github.com/v3/apps/#get-a-single-installation. 

I'd like to propose adding an `AppID` to the `Installation` struct here https://github.com/google/go-github/blob/master/github/apps_installation.go#L14 . 
I also plan on updating the https://github.com/google/go-github/blob/master/github/apps_test.go#L84 test to include an `app_id` to test.

I have a PR ready to go with the addition of the field to the struct. Is that something that would be considered for merging? 

🙇 Thanks for reading.

kshitij10496:
Thanks for raising this issue.
After some digging, I found that the `Installation` struct could use some new fields for wrapping a response.

Reference: [Get a Single Installation](https://developer.github.com/v3/apps/#get-a-single-installation)
```json
{
  "id": 1,
  "account": {
    "login": "github",
    "id": 1,
    "url": "https://api.github.com/orgs/github",
    "repos_url": "https://api.github.com/orgs/github/repos",
    "events_url": "https://api.github.com/orgs/github/events",
    "hooks_url": "https://api.github.com/orgs/github/hooks",
    "issues_url": "https://api.github.com/orgs/github/issues",
    "members_url": "https://api.github.com/orgs/github/members{/member}",
    "public_members_url": "https://api.github.com/orgs/github/public_members{/member}",
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    "description": "A great organization"
  },
  "access_tokens_url": "https://api.github.com/installations/1/access_tokens",
  "repositories_url": "https://api.github.com/installation/repositories",
  "html_url": "https://github.com/organizations/github/settings/installations/1",
  "app_id": 1,
  "target_id": 1,
  "target_type": "Organization",
  "permissions": {
    "metadata": "read",
    "contents": "read",
    "issues": "write",
    "single_file": "write"
  },
  "events": [
    "push",
    "pull_request"
  ],
  "single_file_name": "config.yml",
  "repository_selection": "selected"
}
```
Currently, we support only few fields of the response in the [Installation struct](https://github.com/google/go-github/blob/master/github/apps_installation.go#L14-L20).

```
type Installation struct {
	ID              *int64  `json:"id,omitempty"`
	Account         *User   `json:"account,omitempty"`
[...]

PR #852, opened 2018-02-14
Add app_id to the Installation struct

sridharavinash:
Implements https://github.com/google/go-github/issues/851

An InstallationID is different from an AppID. This PR adds an additional field
so that we can query the GitHub application ID given its installation ID.

The `app_id` field is returned during installation and creation events.
It is also returned in the https://developer.github.com/v3/apps/#get-a-single-installation endpoint.

### DIFF
--- a/github/apps_installation.go
+++ b/github/apps_installation.go
@@ -13,6 +13,7 @@ import (
 // Installation represents a GitHub Apps installation.
 type Installation struct {
 	ID              *int64  `json:"id,omitempty"`
+	AppID           *int64  `json:"app_id,omitempty"`
 	Account         *User   `json:"account,omitempty"`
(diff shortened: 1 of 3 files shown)


sridharavinash:
I signed it!

kshitij10496:
@sridharavinash Thanks for the PR and your patience with the review.
Can you also add a unit test as already mentioned in #851?

sridharavinash:
@kshitij10496  Thanks for the ping. Updated the tests to include the `app_id` as part of the response.

kshitij10496:
Ping @gmlewis 
Since the other fields will be addressed in a companion PR, I think we can proceed with this one for the time being. Can you give it a review?

PR #857, opened 2018-02-28
Corrects create org invitation return value.

1995parham:


1995parham:
I signed it!

gmlewis:
According to the example in the GitHub API docs:
https://developer.github.com/v3/orgs/members/#create-organization-invitation
this endpoint returns a list (their example shows a list of one item).

Why is this change needed, @1995parham ?

Using the API as it is today, a user would access `invitation[0]`.

1995parham:
I use the following code based on [this](https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership) API and got an error for using an array of invitation.

```go
email := "email@email.com"
client.Organizations.CreateOrgInvitation(ctx, "ce104", &CreateOrgInvitationOptions{
    Email: &email,
}) 
```

gmlewis:
Thank you for the clarification, @1995parham.

In that case, I think you need to use a different endpoint.

In the [link you provided](https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership), note that the REST endpoint is: `PUT /orgs/:org/memberships/:username`.

In the function you modified in this PR, the endpoint is:

```go
	u := fmt.Sprintf("orgs/%v/invitations", org)
 	req, err := s.client.NewRequest("POST", u, opt)
```

which translates to: `POST /orgs/:org/invitations`.

So to use the first endpoint, you will want to call:
[`client.Organizations.EditOrgMembership`](https://github.com/google/go-github/blob/master/github/orgs_members.go#L240-L244)
and make sure you provide a non-empty `user` argument to the call.

Please let me know if that information resolves the issue for you or if I have misunderstood.

1995parham:
Thanks for your description and sorry for my mistake about API URLs. But when I calling `CreateOrgInvitation` I get the following error: `json: cannot unmarshal object into Go value of type []*github.Invitation`.

gmlewis:
In cases like these, I like to see the full response from the server in order to debug the problem. You can either construct a `curl` request similar to the one being made by this package, or just temporarily instrument your copy of the package to dump out the contents of the response body, something like:

```go
data, err := ioutil.ReadAll(r.Body)
log.Errorf("dump: %s", data)
```

somewhere around https://github.com/google/go-github/blob/master/github/github.go#L480

If you share the response body, maybe we can help you figure out why it isn't working. Maybe our package is at fault or the GitHub API docs, or both.

1995parham:
```
dump: {"id":5701266,"login":null,"email":"parham.alvani@gmail.com","role":"direct_member","created_at":"2018-03-02T07:00:20Z","inviter":{"login":"1995parham","id":8181240,"avatar_url":"https://avatars1.githubusercontent.com/u/8181240?v=4","gravatar_id":"","url":"https://api.github.com/users/1995parham","html_url":"https://github.com/1995parham","followers_url":"https://api.github.com/users/1995parham/followers","following_url":"https://api.github.com/users/1995parham/following{/other_user}","gists_url":"https://api.github.com/users/1995parham/gists{/gist_id}","starred_url":"https://api.github.com/users/1995parham/starred{/owner}{/repo}","subscriptions_url":"https://api.github.com/users/1995parham/subscriptions","organizations_url":"https://api.github.com/users/1995parham/orgs","repos_url":"https://api.github.com/users/1995parham/repos","events_url":"https://api.github.com/users/1995parham/events{/privacy}","received_events_url":"https://api.github.com/users/1995parham/received_events","type":"User","site_admin":false},"team_count":0,"invitation_teams_url":"https://api.github.com/organizations/35544286/invitations/5701266/teams"}
```

1995parham:
I did as you say and attached the logs

gmlewis:
Ah HAH!!!  Thank you so much, @1995parham! Now I understand.
I apologize for my prior confusion.
Your reply matches this API endpoint:

https://developer.github.com/v3/orgs/members/#create-organization-invitation

...which does indeed match this PR exactly!

Can you please update the corresponding test as reported by the Travis CI build failure?

```
--- FAIL: TestOrganizationsService_CreateOrgInvitation (0.00s)
	orgs_members_test.go:480: Organizations.CreateOrgInvitation returned error: json: cannot unmarshal array into Go value of type github.Invitation
	orgs_members_test.go:485: Organizations.ListPendingOrgInvitations returned <nil>, want [github.Invitation{Email:"octocat@github.com"}]
```

We will then merge after you fix this and we get a second LGTM.

1995parham:
thank you very much for your attention.

gmlewis:
LGTM.  Thanks again, @1995parham!
Merging.

shurcooL:
> According to the example in the GitHub API docs:
> https://developer.github.com/v3/orgs/members/#create-organization-invitation
> this endpoint returns a list (their example shows a list of one item).

Looks like GitHub has fixed the incorrect documentation on their side. The linked endpoint example used to display an array of objects, but now it displays a single object.

PR #861, opened 2018-03-02
Add remaining `app` fields

sridharavinash:
This is the companion PR for adding the remaining fields of the `app` response. 
context: https://github.com/google/go-github/issues/851

This PR adds the remaining fields for the `app` endpoints, specifically adding support for 
`target`
`target_type`
`repository_selection`
`single_file_name`
`permissions`
`events`

### Tests
* I've added unit tests for `target`(s) field for all the tests
* Added a more detailed response test for `ListInstallations` only (would like feedback if I can follow the same pattern for the remaining tests).

/cc @kshitij10496 @gmlewis

### DIFF
--- a/github/apps_installation.go
+++ b/github/apps_installation.go
@@ -12,14 +12,16 @@ import (
 
 // Installation represents a GitHub Apps installation.
 type Installation struct {
-	ID              *int64  `json:"id,omitempty"`
-	AppID           *int64  `json:"app_id,omitempty"`
-	TargetID        *int64  `json:"target_id,omitempty"`
-	Account         *User   `json:"account,omitempty"`
-	AccessTokensURL *string `json:"access_tokens_url,omitempty"`
-	RepositoriesURL *string `json:"repositories_url,omitempty"`
-	HTMLURL         *string `json:"html_url,omitempty"`
-	TargetType      *string `json:"target_type,omitempty"`
+	ID                  *int64  `json:"id,omitempty"`
+	AppID               *int64  `json:"app_id,omitempty"`
+	TargetID            *int64  `json:"target_id,omitempty"`
+	Account             *User   `json:"account,omitempty"`
+	AccessTokensURL     *string `json:"access_tokens_url,omitempty"`
+	RepositoriesURL     *string `json:"repositories_url,omitempty"`
+	HTMLURL             *string `json:"html_url,omitempty"`
+	TargetType          *string `json:"target_type,omitempty"`
+	SingleFileName      *string `json:"single_file,omitempty"`
+	RepositorySelection *string `json:"repository_selection,omitempty"`
 }
 
 func (i Installation) String() string {
--- a/github/apps_installation.go
+++ b/github/apps_installation.go
@@ -10,9 +10,9 @@ import (
 	"fmt"
 )
 
-// InstallationPermissions Lists the permissions for metadata, contents, issues and single file for an installation.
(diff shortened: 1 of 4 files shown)


github/apps_installation.go
```diff
@@ -10,14 +10,28 @@ import (
 	"fmt"
 )
 
+// InstallationPermissions Lists the permissions for metadata, contents, issues and single file for an installation.
+type InstallationPermissions struct {
+	MetaData   *string `json:"metadata,omitempty"`
```
gmlewis: To be consistent with our naming policies, please change this field name from `MetaData` to `Metadata` since the corresponding JSON field is not `meta_data`.

github/apps_installation.go
```diff
@@ -10,14 +10,28 @@ import (
...
+	AppID               *int64                   `json:"app_id,omitempty"`
+	TargetID            *int64                   `json:"target_id,omitempty"`
+	Account             *User                    `json:"account,omitempty"`
+	AccessTokensURL     *string                  `json:"access_tokens_url,omitempty"`
+	RepositoriesURL     *string                  `json:"repositories_url,omitempty"`
+	HTMLURL             *string                  `json:"html_url,omitempty"`
+	TargetType          *string                  `json:"target_type,omitempty"`
+	SingleFileName      *string                  `json:"single_file_name,omitempty"`
+	RepositorySelection *string                  `json:"repository_selection,omitempty"`
+	Events              *[]string                `json:"events,omitempty"`
```
gmlewis: Since a Go slice is already a reference type, please change this from `*[]string` to `[]string`.

github/apps_installation.go
```diff
@@ -10,14 +10,28 @@ import (
 	"fmt"
 )
 
+// InstallationPermissions Lists the permissions for metadata, contents, issues and single file for an installation.
```
gmlewis: I'm noticing that the only file in our repo that uses type `Installation` is [apps.go](https://github.com/google/go-github/blob/master/github/apps.go#L71-L123), so I'm thinking that we should move `InstallationPermissions` and `Installation` into apps.go.

Also, for Godoc purposes, please change `Lists` to `lists` here so that this reads as a normal English sentence.

github/apps_installation.go
sridharavinash: 👍

github/apps_installation.go
sridharavinash: 👍

github/apps_installation.go
sridharavinash: >I'm noticing that the only file in our repo that uses type Installation is apps.go, so I'm thinking that we should move InstallationPermissions and Installation into apps.go.

@gmlewis no strong opinions here. I can move it over to `apps.go` if that makes more sense.

sridharavinash: >Also, for Godoc purposes, please change Lists to lists here so that this reads as a normal English sentence.

👍

gmlewis: Yes, please move those two structs to `apps.go` in this PR if you don't mind.
Thank you!

sridharavinash: done in https://github.com/google/go-github/pull/861/commits/ba66189c58f247ec0277709cbb2f3b90f9c4dd82

sridharavinash:
@gmlewis Thanks for review. I think i've addressed all the feedback. Let me know how this is looking. 🙇

gmlewis:
Merging.

PR #865, opened 2018-03-07
detect abuse rate limits on github enterprise

bbatha:
On github enteprise the documenation URLs are different for instance the public documentation url for abuse limits is https://developer.github.com/v3/#abuse-rate-limits where as an enterprise version will return a version specific URL such as https://developer.github.com/enterprise/2.12/v3/#abuse-rate-limits

This changes the error type deserialization to test for the error in a github deployment agnostic way.


gmlewis:
Thank you, @bbatha!

One other possibility would be to search for:

```go
strings.HasSuffix(errorResponse.DocumentationURL, "/#abuse-rate-limits")
```

This might be more stable in the long run in case GitHub changes their response wording.  Maybe other contributors could comment on which might be a better approach. (I'm leaning toward my suggestion, but I'm biased.)

Regardless, please add a test case that shows both Enterprise and non-Enterprise responses... partly as a means of documenting the behavior.

Feel free to wait for consensus, though, before you add the test so as to not make extra work for yourself.

shurcooL:
The documentation at https://developer.github.com/enterprise/2.12/v3/#abuse-rate-limits actually says:

> If your application triggers this rate limit, you'll receive an informative response:
>
> ```
> HTTP/1.1 403 Forbidden
> Content-Type: application/json; charset=utf-8
> Connection: close
> {
>   "message": "You have triggered an abuse detection mechanism and have been temporarily blocked from content creation. Please retry your request again later.",
>   "documentation_url": "https://developer.github.com/v3/#abuse-rate-limits"
> }
> ```

It does not say `"documentation_url": "https://developer.github.com/enterprise/2.12/v3/#abuse-rate-limits"`. Is their documentation page wrong?

If so, I agree with @gmlewis's suggestion of checking for `"/#abuse-rate-limits"` suffix. Maybe even look for `"/v3/#abuse-rate-limits"` suffix since that's the largest common suffix. I don't imagine there would be false positives.

At the same time, I don't expect GitHub would change the exact message text very often either.

sahildua2305:
I don't think there's a big difference in checking for `message` or the `documentation_url`. Maybe we can have both checks and if at least one of the two triggers, we take that as a rate limits abuse?

bbatha:
Its looks like in the official octokit libraries the way that the abuse errors are detected varies. In the ruby library, is just [looks for the string "abuse" in the body](https://github.com/octokit/octokit.rb/blob/ad74f84e818ee15d2e49422d51978a40b2d957a0/lib/octokit/error.rb#L66). In the C# library it checks body for ["abuse-rate-limits" or "abuse detection mechanism"](https://github.com/octokit/octokit.net/blob/4e804f61a69f41757acbf2a1e4e96dd288e59a57/Octokit/Http/Connection.cs#L684).

> It does not say "documentation_url": "https://developer.github.com/enterprise/2.12/v3/#abuse-rate-limits". Is their documentation page wrong?

Yes it is. All errors return documentation urls specific to your github version. This behavior in my memory has been consistent across at least the last 4 releases of github enterprise. It seems all of the example urls in responses on that page are copied from their .com documentation, see [1](https://developer.github.com/enterprise/2.12/v3/#link-header) [2](https://developer.github.com/enterprise/2.12/v3/#rate-limiting) [3](https://developer.github.com/enterprise/2.12/v3/#graphql-global-relay-ids)

bbatha:
I have added a test case with the enterprise documentation url for 2.12 (latest). And made the checks for abuse limits depend on either the message or the common suffix in the documentation url. Please let me know if you would like this to change.

github/github.go
```diff
@@ -687,7 +687,7 @@ func CheckResponse(r *http.Response) error {
 			Response: errorResponse.Response,
 			Message:  errorResponse.Message,
 		}
-	case r.StatusCode == http.StatusForbidden && errorResponse.DocumentationURL == "https://developer.github.com/v3/#abuse-rate-limits":
+	case r.StatusCode == http.StatusForbidden && (strings.HasPrefix(errorResponse.Message, "You have triggered an abuse detection mechanism") || strings.HasSuffix(errorResponse.DocumentationURL, "#abuse-rate-limits")):
```
shurcooL: I think it'd be better to simplify the detection check to just:

```Go
case r.StatusCode == http.StatusForbidden && strings.HasSuffix(errorResponse.DocumentationURL, "/v3/#abuse-rate-limits"):
```

It would handle both enterprise and not situations. False positives and false negatives are unlikely (unless GitHub changes things on their end, but then we can update our code accordingly). It would be easier to understand and maintain.

bbatha:
Updated to simplify the limit check as requested.

bbatha:
I don't have permissions to restart the travis build, but it looks like the failure for go 1.9 is spurious: the coverage tool failed to `go get`.

bbatha:
Hi just checking in, is there anything blocking the merge of this pull request?

shurcooL:
Thanks for the ping @bbatha, we just forgot about this PR. I've requested a review from @gmlewis and we can merge once it looks good to him as well.

One more thing, have you had a chance to report the documentation bug to GitHub support? If not, would you mind doing that? When GitHub fixes the documentation mismatch (and confirms it's a docs bug rather than behavior bug), it'll be easier for us to have confidence this behavior is correct and won't break in the future.

I checked the newer version (2.13) at https://developer.github.com/enterprise/2.13/v3/#abuse-rate-limits and it still says:

> If your application triggers this rate limit, you'll receive an informative response:
>
> ```
> HTTP/1.1 403 Forbidden
> Content-Type: application/json; charset=utf-8
> Connection: close
> {
>   "message": "You have triggered an abuse detection mechanism and have been temporarily blocked from content creation. Please retry your request again later.",
>   "documentation_url": "https://developer.github.com/v3/#abuse-rate-limits"
> }
> ```

bbatha:
I've filed a ticket under our support portal for the documentation url.

bbatha:
I have fixed the commit message to respect GitHub branding.

github/github_test.go
```diff
@@ -674,6 +674,40 @@ func TestDo_rateLimit_abuseRateLimitError(t *testing.T) {
 	}
 }
 
+// Ensure *AbuseRateLimitError is returned when the response indicates that
+// the client has triggered an abuse detection mechanism on github enterprise.
```
shurcooL: I believe Glenn was referring to the string here, which can be made "GitHub Enterprise".

github/github_test.go
```diff
@@ -674,6 +674,40 @@ func TestDo_rateLimit_abuseRateLimitError(t *testing.T) {
...
+func TestDo_rateLimit_abuseRateLimitErrorEnterprise(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		w.WriteHeader(http.StatusForbidden)
+		// When the abuse rate limit error is of the "temporarily blocked from content creation" type,
+		// there is no "Retry-After" header.
+		// This response returns a documentation url like the one returned for github enterprise, this
```
shurcooL: And on this line too.

Issue #856, opened 2018-02-22
Support new preview Label API improvements


DirtyF:
Following GitHub announcement https://github.com/blog/2505-label-improvements-emoji-descriptions-and-more, it would be useful to be able to define and access to label descriptions.

gmlewis:
Thanks, @DirtyF, for filing the issue.

As far as I can tell, the GitHub API does not expose the description of each label:
https://developer.github.com/v3/pulls/#response

It only exposes the following fields:

    * ID     *int64  `json:"id,omitempty"`
    * URL    *string `json:"url,omitempty"`
    * Name   *string `json:"name,omitempty"`
    * Color  *string `json:"color,omitempty"`
    * NodeID *string `json:"node_id,omitempty"`

Indeed, check out this response:
https://api.github.com/repos/google/go-github/pulls/853

```js
  "labels": [
    {
      "id": 698400531,
      "url": "https://api.github.com/repos/google/go-github/labels/cla:%20yes",
      "name": "cla: yes",
      "color": "0e8a16",
      "default": false
    }
  ],
```

The name of the label is provided, but not the description of the label.
(Looking at it more closely, I don't know what `NodeID` represents or why we don't support `Default`, but that is a separate issue :smile:)

If you wish, you could contact support@github.com and request that they add the description to their API, at which point we could add support for it.  Thanks again.

shurcooL:
@DirtyF Historically, API access to new features comes some time after the feature is released on github.com. The linked announcement doesn't say anything about API access; you'll need to watch https://developer.github.com/changes/ for that.

DirtyF:
https://developer.github.com/changes/2018-02-22-label-description-search-preview/

gmlewis:
Wow, that was quick!  OK, I'm going to hijack this issue and change the title now that the preview API has been announced.

GitHub Developer API announcement:
https://developer.github.com/changes/2018-02-22-label-description-search-preview/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue, and we can send you an invite to contribute to the repo (which you then accept) and the issue can be assigned to you so that others don't attempt to duplicate the work.

Make sure you have two-factor authentication enabled on GitHub.

Thank you!

gjacquet:
That's something I would be interested in contributing to.
I was looking for that feature too.

MonicB14:
Is this issue still available for contribution?

[...]

PR #866, opened 2018-03-11
Label api improvements

gjacquet:
Add support for the new label API improvements as described in https://developer.github.com/changes/2018-02-22-label-description-search-preview/ and addresses #856

This PR adds:
* support for the Description and Default fields of the labels
* search label

N.B. search label result text highlighting is no implemented yet. I can add it to this PR or open another one but since this is my first contribution to this project (and my first public go PR),  I would like go get feedback early.

### DIFF
--- a/github/issues_labels.go
+++ b/github/issues_labels.go
@@ -8,15 +8,18 @@ package github
 import (
 	"context"
 	"fmt"
+	"strings"
 )
 
 // Label represents a GitHub label on an Issue
 type Label struct {
-	ID     *int64  `json:"id,omitempty"`
-	URL    *string `json:"url,omitempty"`
-	Name   *string `json:"name,omitempty"`
-	Color  *string `json:"color,omitempty"`
-	NodeID *string `json:"node_id,omitempty"`
+	ID          *int64  `json:"id,omitempty"`
+	URL         *string `json:"url,omitempty"`
+	Name        *string `json:"name,omitempty"`
+	Color       *string `json:"color,omitempty"`
+	Description *string `json:"description,omitempty"`
+	Default     *bool   `json:"default,omitempty"`
+	NodeID      *string `json:"node_id,omitempty"`
 }
 
 func (l Label) String() string {
@@ -39,7 +42,8 @@ func (s *IssuesService) ListLabels(ctx context.Context, owner string, repo strin
 	}
 
 	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeGraphQLNodeIDPreview)
+	acceptHeaders := []string{mediaTypeGraphQLNodeIDPreview, mediaTypeLabelDescriptionSearchPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	var labels []*Label
 	resp, err := s.client.Do(ctx, req, &labels)
@@ -61,7 +65,8 @@ func (s *IssuesService) GetLabel(ctx context.Context, owner string, repo string,
 	}
 
 	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeGraphQLNodeIDPreview)
+	acceptHeaders := []string{mediaTypeGraphQLNodeIDPreview, mediaTypeLabelDescriptionSearchPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	label := new(Label)
 	resp, err := s.client.Do(ctx, req, label)
@@ -83,7 +88,8 @@ func (s *IssuesService) CreateLabel(ctx context.Context, owner string, repo stri
 	}
 
 	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeGraphQLNodeIDPreview)
+	acceptHeaders := []string{mediaTypeGraphQLNodeIDPreview, mediaTypeLabelDescriptionSearchPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	l := new(Label)
 	resp, err := s.client.Do(ctx, req, l)
@@ -105,7 +111,8 @@ func (s *IssuesService) EditLabel(ctx context.Context, owner string, repo string
 	}
 
 	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeGraphQLNodeIDPreview)
+	acceptHeaders := []string{mediaTypeGraphQLNodeIDPreview, mediaTypeLabelDescriptionSearchPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	l := new(Label)
 	resp, err := s.client.Do(ctx, req, l)
@@ -144,7 +151,8 @@ func (s *IssuesService) ListLabelsByIssue(ctx context.Context, owner string, rep
 	}
 
 	// TODO: remove custom Accept header when this API fully launches.
-	req.Header.Set("Accept", mediaTypeGraphQLNodeIDPreview)
+	acceptHeaders := []string{mediaTypeGraphQLNodeIDPreview, mediaTypeLabelDescriptionSearchPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	var labels []*Label
 	resp, err := s.client.Do(ctx, req, &labels)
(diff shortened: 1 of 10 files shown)


github/issues_labels.go
```diff
@@ -183,9 +192,14 @@ func (s *IssuesService) AddLabelsToIssue(ctx context.Context, owner string, repo
 func (s *IssuesService) RemoveLabelForIssue(ctx context.Context, owner string, repo string, number int, label string) (*Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/issues/%d/labels/%v", owner, repo, number, label)
 	req, err := s.client.NewRequest("DELETE", u, nil)
+
```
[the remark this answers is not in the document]
gjacquet: Yup, my bad...

github/search.go
```diff
@@ -172,18 +178,52 @@ func (c CodeResult) String() string {
...
+}
+
+func (l LabelResult) String() string {
+	return Stringify(l)
+}
+
+// Labels searches labels via various criteria.
+//
+// GitHub API docs: https://developer.github.com/v3/search/#search-labels
+func (s *SearchService) Labels(ctx context.Context, repositoryID string, query string, opt *SearchOptions) (*LabelsSearchResult, *Response, error) {
```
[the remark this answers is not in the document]
shurcooL: Great!

> Should the enum `repositoryID` have been renamed to `repoID` or is OK to keep it more similar to what is in the Gihub API?

Since `searchParameters` is unexported, users won't ever see it (and we can easily change it at any time without a breaking API change). I'm okay with both, so go with whichever you think is nicer.

github/search.go
```diff
@@ -172,18 +179,52 @@ func (c CodeResult) String() string {
...
 func (s *SearchService) Code(ctx context.Context, query string, opt *SearchOptions) (*CodeSearchResult, *Response, error) {
 	result := new(CodeSearchResult)
-	resp, err := s.search(ctx, "code", query, opt, result)
+	resp, err := s.search(ctx, "code", &searchParameters{Query: query}, opt, result)
+	return result, resp, err
+}
+
+// LabelsSearchResult represents the result of a code search.
+type LabelsSearchResult struct {
+	Total             *int          `json:"total_count,omitempty"`
```
gmlewis: s/Total/TotalCount/

github/search.go
```diff
@@ -172,18 +179,52 @@ func (c CodeResult) String() string {
...
+	IncompleteResults *bool         `json:"incomplete_results,omitempty"`
+	Labels            []LabelResult `json:"items,omitempty"`
```
gmlewis: Two things:
s/Labels/Items/
s/[]LabelResult/[]*LabelResult/

github/search.go
```diff
@@ -172,18 +179,52 @@ func (c CodeResult) String() string {
...
+	Score       *float64 `json:"score,omitempty"`
+}
+
+func (l LabelResult) String() string {
+	return Stringify(l)
+}
+
+// Labels searches labels via various criteria.
+//
+// GitHub API docs: https://developer.github.com/v3/search/#search-labels
```
gmlewis: We need a comment here about the significance of `repoID==0`.

github/search.go
```diff
@@ -200,6 +241,10 @@ func (s *SearchService) search(ctx context.Context, searchType string, query str
 		// Accept header for search repositories based on topics preview endpoint
 		// TODO: remove custom Accept header when this API fully launches.
 		req.Header.Set("Accept", mediaTypeTopicsPreview)
+	case searchType == "labels":
+		// Accept header for search labels based on label description preview endpoint
```
gmlewis: nit: add period to end of comment. I see that the other comments don't, but Go style typically uses complete sentences for comments.

github/search.go
shurcooL: @gmlewis Please consider that this is consistent with the other search result structs. `RepositoriesSearchResult`, `RepositoriesSearchResult`, `CommitsSearchResult`, etc., all have `Total`, not `TotalCount`. We've had this discussion before and agreed it's better to keep it that way.

github/search.go
shurcooL: Agree with `s/[]LabelResult/[]*LabelResult/`.

However, I don't think `s/Labels/Items/` should be done for the same reason as above. The other `*SearchResult` structs all use renamed Go fields, not `Items`.

github/search.go
gmlewis: OK, SGTM. Sorry I forgot about this exception to keeping the JSON tag consistent with the Go name.

github/search.go
gmlewis: OK, SGTM.

github/search.go
shurcooL: As I understand, `repoID` value of `0` is invalid. This endpoint requires a repository to be specified, and there is no repo with ID 0.

I haven't tried it, but I imagine that GitHub API would return an error.

We can still document it to make it more clear.

github/search.go
shurcooL: No problem! It's unfortunate we have to keep remembering these exceptions to the rules, but alas. Luckily, new types of search results aren't added all that often. :)

github/search.go
gmlewis: Sorry I wasn't clear.  My comment was based on the code I saw below:

```go
	if parameters.RepositoryID != 0 {
		params.Set("repository_id", strconv.FormatInt(parameters.RepositoryID, 10))
	}
```

This shows that a `repoID` of `0` will not be sent at all.  That's all I meant by commenting it.
I was thinking of something like `// A repoID of 0 will not be sent.` or some such.

Also, @shurcooL - what are your thoughts about using `fmt.Sprintf` here instead of `strconv.FormatInt` ?
I prefer the former, but could go either way.

gjacquet: Actually for that one I meant to have `repoID` As an optional parameter...
I guess that anything `<= 0` does not make sense for the API.

Maybe having that parameter as an `*int64` and checking for `nil` would make it clearer.

I have no opinion on `Sprintf` vs `FormatInt`. I’m d be curious to know what makes one better than the other.

shurcooL: > I was thinking of something like `// A repoID of 0 will not be sent.` or some such.

A comment like that in `searchParameters` struct is okay with me.

However, it shouldn't be added to the `SearchService.Labels` endpoint because it's not a part of the API. `SearchService.Labels` always needs to be called with a valid `repoID`, and 0 is not a valid one.

An alternative to make things more clear could be to modify `searchParameters` to:

```Go
type searchParameters struct {
	Query        string
	RepositoryID *int64 // Sent if non-nil.
}
```

And `SearchService.Labels` would do:

```Go
&searchParameters{RepositoryID: &repoID, Query: query}
```

gjacquet: I have turned `repoID` into an `*int64` I think it makes more sense since it’s an optional parameter in this API.

github/search_test.go
```diff
@@ -285,16 +285,17 @@ func TestSearchService_Labels(t *testing.T) {
 		fmt.Fprint(w, `{"total_count": 4, "incomplete_results": false, "items": [{"id": 1234, "name":"bug", "description": "some text"},{"id": 4567, "name":"feature"}]}`)
 	})
 
+	repoID := int64(1234)
 	opts := &SearchOptions{Sort: "updated", Order: "desc", ListOptions: ListOptions{Page: 2, PerPage: 2}}
-	result, _, err := client.Search.Labels(context.Background(), 1234, "blah", opts)
+	result, _, err := client.Search.Labels(context.Background(), &repoID, "blah", opts)
```
shurcooL: We have [`github.Int64`](https://godoc.org/github.com/google/go-github/github#Int64) helper for this, but I'm not yet sure if this change should be done.

github/search.go
shurcooL: I agree that it's optional in the context of `search` method (which is used for searching labels, repositories, issues, etc.), but is it optional for `Labels` method too?

From reading documentation at https://developer.github.com/v3/search/#search-labels, it sounded to me that it was mandatory to specify a repository when searching labels:

> Find labels in a repository with names or descriptions that match search keywords.

Have you tried if the endpoint works if repository ID is not specified?

gjacquet: You are absolutely right. The doc does not explicitly states it is required like `q` for instance so I assumed it was not required.
I did get a 422 when querying without repo id. I’ll fix that in a minute.

gjacquet:
Should be all good now!

gmlewis:
@gjacquet - it looks like your commits are coming from two different email addresses, and the `cla/google` check might not be happy until both email addresses have signed CLAs or are from the same GitHub account.

So you may be able to associate both email addresses with your GitHub account by visiting https://github.com/settings/emails and clicking `Add email address`.

If that doesn't fix the problem, then you probably will have to follow the CLA link (see comments from @googlebot above) in an incognito window from both email addresses and sign the CLA. If that doesn't work, then we may need to pull in the expert.

shurcooL:
> You are absolutely right. The doc does not explicitly states it is required like q for instance so I assumed it was not required.
> I did get a 422 when querying without repo id. I’ll fix that in a minute.

Thanks for checking.

Did you see my earlier https://github.com/google/go-github/pull/866#discussion_r178421939? Do you think it's a good idea?

gjacquet:
@shurcooL now that we’ve established that repoID is mandatory, your suggestion makes a lot of sense.

gjacquet:
@shurcooL I have just updated the comment based on your suggestion.

gmlewis:
OK, thanks, @gjacquet... @willnorris can clear the @googlebot issue for us... possibly just by squashing and merging for us... I'll leave it up to him.

PR #867, opened 2018-03-12
Add the ability to read a label from labeled events

cflewis:
This PR adds the ability to read labels from 'labeled' pull request events. This isn't present in any other event, hence the comment.

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -495,6 +495,7 @@ type PullRequestEvent struct {
 	Repo               *Repository   `json:"repository,omitempty"`
 	Sender             *User         `json:"sender,omitempty"`
 	Installation       *Installation `json:"installation,omitempty"`
+	Label              *Label        `json:"label,omitempty"` // Populated in "labeled" event deliveries.
 }
 
 // PullRequestReviewEvent is triggered when a review is submitted on a pull


cflewis:
I have no idea what to make of the travis failure :/

willnorris:
This is not documented in the GitHub API docs, but Chris is seeing it in the wild.

github/event_types.go
```diff
@@ -495,6 +495,7 @@ type PullRequestEvent struct {
 	Repo               *Repository   `json:"repository,omitempty"`
 	Sender             *User         `json:"sender,omitempty"`
 	Installation       *Installation `json:"installation,omitempty"`
+	Label              *Label        `json:"label,omitempty"` // Populated in "labeled" event deliveries.
```
shurcooL: To help keep the code here more consistent, it would be a good idea to move `Label` up so it's between `RequestedReviewers` and `Repo`.

If you look at all the events above and below this one, they all follow a pattern. There are normal event fields, then a blank line, then a "// The following fields are only populated by Webhook events." comment, then all the fields that are Webhook-only.

However, most events have generic `Repo`, `Sender`, `Installation` fields that are not specific to the event type, and they're placed last, after event-specific Webhook fields.

PR #871, opened 2018-03-16
Update orgs_members.go

mlbright:
role filter 'member' means return only users that are not owners

### DIFF
--- a/github/orgs_members.go
+++ b/github/orgs_members.go
@@ -59,7 +59,7 @@ type ListMembersOptions struct {
 	// Possible values are:
 	//     all - all members of the organization, regardless of role
 	//     admin - organization owners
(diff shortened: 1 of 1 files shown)


github/orgs_members.go
```diff
@@ -59,7 +59,7 @@ type ListMembersOptions struct {
 	// Possible values are:
 	//     all - all members of the organization, regardless of role
 	//     admin - organization owners
-	//     member - non-organization members
+	//     member - non-owner members
```
shurcooL: The documentation of role values at https://developer.github.com/v3/orgs/members/#members-list says:

* `member` - Non-owner organization members.

Would using that phrasing be better?

mlbright: Perfect

PR #872, opened 2018-03-17
Fix incorrect marshalling of DismissalRestrictionsRequest field.

### LINKED ISSUE #664: PullRequestReviewsEnforcementRequest.DismissalRestrictionsRequest set to nil results in 422 from GH

ethomas2:
Previously, an empty DismissalRestrictionsRequest was being marshalled
to an empty array [] instead of an empty object {}. This is an invalid
api call according to github's documentation, and caused github to
respond with a 422.

Fixes #664

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -585,11 +585,11 @@ type PullRequestReviewsEnforcementRequest struct {
 func (req PullRequestReviewsEnforcementRequest) MarshalJSON() ([]byte, error) {
 	if req.DismissalRestrictionsRequest == nil {
 		newReq := struct {
-			R []interface{} `json:"dismissal_restrictions"`
-			D bool          `json:"dismiss_stale_reviews"`
-			O bool          `json:"require_code_owner_reviews"`
+			R struct{} `json:"dismissal_restrictions"`
+			D bool     `json:"dismiss_stale_reviews"`
+			O bool     `json:"require_code_owner_reviews"`
 		}{
-			R: []interface{}{},
+			R: struct{}{},
 			D: req.DismissStaleReviews,
 			O: req.RequireCodeOwnerReviews,
 		}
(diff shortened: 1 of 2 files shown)


ethomas2:
According to github's [docs](https://developer.github.com/v3/repos/branches/#update-branch-protection), to disable dismissal_restrictions, you are supposed to pass an empty object. go-github was passing an empty array, which caused github to respond with a 422

It looks like this mistake was made in go-github because there used to be a typo in github's documentation that said to pass in an empty array. That typo has since been fixed in docs. I _believe_ passing an array instead of an object has always resulted in a 422, even when the docs were incorrect.

ethomas2:
Somewhat related to this pull this pull request: 

There are several array-fields in structs for branch protection requests that get marshaled to nil instead of  `[]`. Github responds with a when these fields are nil. For example
```go
	_, _, err := ghclient.Repositories.UpdateBranchProtection(
		ctx, "MyOrg", "myrepo", "master",
		&github.ProtectionRequest{
			RequiredPullRequestReviews: &github.PullRequestReviewsEnforcementRequest{},
			Restrictions:               &github.BranchRestrictionsRequest{},
		})
```
gets this response from github

```
PUT https://git.dev.box.net/api/v3/repositories/3333/branches/master/protection: 422 Invalid request.

For 'anyOf/1', {"users"=>nil, "teams"=>nil} is not a null. []
```

Instead you must explicitly set Users and Teams to be an empty array instead of a nil array
```go
	_, _, err := ghclient.Repositories.UpdateBranchProtection(
		ctx, "ProductivitySandbox", "evan-test", "master",
		&github.ProtectionRequest{
			RequiredPullRequestReviews: &github.PullRequestReviewsEnforcementRequest{},
			Restrictions: &github.BranchRestrictionsRequest{
				Users: []string{},
				Teams: []string{},
			},
		})
```

There are more structs other than `github.BranchRestrictionsRequest` which follow this pattern. I've read go-github's docs for these fields and I realize this behavior is intentional, but I think it provides a poor experience. In my own project, i've resorted to writing my own branch protection structs that marshall `nil` arrays to `[]`. Would you guys be open to me opening another pull request that marshals nil arrays to `[]`?

shurcooL:
This overlaps with #791.

ethomas2:
I think #791 makes this pr unnecessary. Closing

PR #874, opened 2018-03-21
Remove mediaTypeGraphQLNodeIDPreview accept header from DownloadReleaseAsset.

### LINKED ISSUE #870: Incorrect Accept header in DownloadReleaseAsset.
https://github.com/google/go-github/blob/d2b80df1e3eeccae96eb1cb489e65372cc9550c8/github/repos_releases.go#L256

I was able to confirm with curl, that removing `mediaTypeGraphQLNodeIDPreview` from the Accept header will result in the server returning the correct response.

With the `mediaTypeGraphQLNodeIDPreview` Accept header the v3 json response is returned.

rhysd:
As I commented at https://github.com/google/go-github/issues/870#issuecomment-375071816, a downloaded binary file with `DownloadReleaseAsset()` is broken due to the additional accept header. I'm facing this problem on [go-github-selfupdate](https://github.com/rhysd/go-github-selfupdate) library.

https://travis-ci.org/rhysd/go-github-selfupdate/jobs/356272200#L218

I confirmed that this fix solved the problem on my local machine (macOS 10.12, Go 1.10).

Fixes #870.

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -251,8 +251,6 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
 	if err != nil {
 		return nil, "", err
 	}
-
-	// TODO: remove custom Accept header when APIs fully launch.
 	req.Header.Set("Accept", defaultMediaType)
 
 	s.client.clientMu.Lock()
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -253,8 +253,7 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
 	}
 
 	// TODO: remove custom Accept header when APIs fully launch.
-	acceptHeaders := []string{defaultMediaType, mediaTypeGraphQLNodeIDPreview}
-	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
+	req.Header.Set("Accept", defaultMediaType)
 
 	s.client.clientMu.Lock()
 	defer s.client.clientMu.Unlock()
--- a/github/repos_releases_test.go
+++ b/github/repos_releases_test.go
@@ -220,10 +220,9 @@ func TestRepositoriesService_DownloadReleaseAsset_Stream(t *testing.T) {
 	client, mux, _, teardown := setup()
 	defer teardown()
 
-	acceptHeaders := []string{defaultMediaType, mediaTypeGraphQLNodeIDPreview}
 	mux.HandleFunc("/repos/o/r/releases/assets/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		testHeader(t, r, "Accept", strings.Join(acceptHeaders, ", "))
+		testHeader(t, r, "Accept", defaultMediaType)
 		w.Header().Set("Content-Type", "application/octet-stream")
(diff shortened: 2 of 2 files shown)


github/repos_releases.go
```diff
@@ -253,7 +253,7 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
 	}
 
 	// TODO: remove custom Accept header when APIs fully launch.
-	acceptHeaders := []string{defaultMediaType, mediaTypeGraphQLNodeIDPreview}
+	acceptHeaders := []string{defaultMediaType}
 	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
```
shurcooL: This is a good start, but let's simplify the code further. We can make it as simple as it was [before](https://github.com/google/go-github/commit/25b0481acbe9f2123f2b39c3235b156cff6296cd#diff-75909ea88d89029fe24198471fc539aeR228) the `mediaTypeGraphQLNodeIDPreview` header was added to it:

```Go
req.Header.Set("Accept", defaultMediaType)
```

This function operates on binary data, and it's very unlikely its media type will ever need to change. (But if so, it'll be easy to do so then.)

The `// TODO: remove custom Accept header when APIs fully launch.` should be removed too, since it's not applicable (there isn't any preview API we're waiting to fully launch).

github/repos_releases_test.go
```diff
@@ -220,7 +220,7 @@ func TestRepositoriesService_DownloadReleaseAsset_Stream(t *testing.T) {
 	client, mux, _, teardown := setup()
 	defer teardown()
 
-	acceptHeaders := []string{defaultMediaType, mediaTypeGraphQLNodeIDPreview}
+	acceptHeaders := []string{defaultMediaType}
 	mux.HandleFunc("/repos/o/r/releases/assets/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", strings.Join(acceptHeaders, ", "))
```
shurcooL: Similarly, we can make this just:

```Go
testHeader(t, r, "Accept", defaultMediaType)
```

And so on below.

rhysd:
Sure, thank you @shurcooL for your review. I'll fix my patch tonight :+1:

rhysd:
@shurcooL I updated the patch at 6e52e87. Could you review again?

github/repos_releases.go
```diff
@@ -253,8 +253,7 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
 	}
 
 	// TODO: remove custom Accept header when APIs fully launch.
```
shurcooL: Remove this comment, since it's no longer applicable. Also remove the blank line above it, there's no need to separate the `req.Header.Set` call from the code above.

shurcooL:
Also, there's no need to squash commits when working on a PR. You can keep them separate, which makes reviewing individual changes easier (i.e., old commits don't need to be reviewed again). We'll squash everything into one commit when merging this PR.

github/repos_releases.go
rhysd: I did not care about the comment. I'll do it.

rhysd:
@shurcooL @gmlewis Hi, any updates on this? I believe this branch is ready for merge.

gmlewis:
Merging.

rhysd:
@gmlewis Thank you!

Issue #862, opened 2018-03-05
No way to inspect the body after receiving a 202


SamWhited:
When a 202 is returned, an `AcceptedError` is returned by `client.Do` and the `resp.Body` is closed. The body is never read or decoded into the result parameter, so there is no way to inspect the body. However, I am beta testing an experimental endpoint that returns URLs you need to check the status of your request so I need access to the body.

I am not sure if any existing endpoints return a body in 202s, but some will in the future so it would be nice if we could access the body.

shurcooL:
> I am beta testing an experimental endpoint that returns URLs

Can you clarify what you mean? Is it an experimental GitHub REST API v3 endpoint?

If/when GitHub does add an endpoint that returns meaningful data with 202 status code, then we can make changes to support reading the body. I don't think we should do it before that.

We would need to know the schema/encoding of the body to be able to parse it, otherwise the best go-github could do is return it as a byte slice. But that's not very friendly.

shurcooL:
> I went and looked, some of their existing APIs do too. For example: https://developer.github.com/v3/repos/#transfer-a-repository

Thanks, that’s helpful.

I agree, we should support that endpoint returning the repository data when response status code is 202.

chughpiyush:
@gmlewis This too can be closed as the PR is already merged.

gmlewis:
Closing as fixed by #878.

PR #878, opened 2018-03-26
In CreateFork, return the repo info when 202 Accepted occurs.

dsymonds:
Per the docs
(https://developer.github.com/v3/repos/forks/#create-a-fork), there is
still a useful body, and it is helpful for automatic fork creation to
get that response.

I'm not sure if this approach is okay. An alternate is to do the decoding inside Do, so this works for the ~dozen other places that might get an AcceptedError, but I didn't want to go check the docs for every single case to see whether the response body is meant to have useful data.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -487,18 +487,25 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 	err = CheckResponse(resp)
 	if err != nil {
-		// even though there was an error, we still return the response
-		// in case the caller wants to inspect it further
-		return response, err
+		// Even though there was an error, we still return the response
+		// in case the caller wants to inspect it further.
+		// However, if the error is AcceptedError, decode it below before
+		// returning from this function and closing the response body.
+		if _, ok := err.(*AcceptedError); !ok {
+			return response, err
+		}
 	}
 
 	if v != nil {
 		if w, ok := v.(io.Writer); ok {
 			io.Copy(w, resp.Body)
 		} else {
-			err = json.NewDecoder(resp.Body).Decode(v)
-			if err == io.EOF {
-				err = nil // ignore EOF errors caused by empty response body
+			decErr := json.NewDecoder(resp.Body).Decode(v)
+			if decErr == io.EOF {
+				decErr = nil // ignore EOF errors caused by empty response body
+			}
+			if decErr != nil {
+				err = decErr
 			}
 		}
 	}
--- a/github/repos_forks.go
+++ b/github/repos_forks.go
@@ -58,7 +58,8 @@ type RepositoryCreateForkOptions struct {
 //
 // This method might return an *AcceptedError and a status code of
 // 202. This is because this is the status that GitHub returns to signify that
-// it is now computing creating the fork in a background task.
+// it is now computing creating the fork in a background task. In this event,
+// the Repository value will be returned, which includes the details about the pending fork.
 // A follow up request, after a delay of a second or so, should result
 // in a successful request.
 //
@@ -77,6 +78,9 @@ func (s *RepositoriesService) CreateFork(ctx context.Context, owner, repo string
 
 	fork := new(Repository)
 	resp, err := s.client.Do(ctx, req, fork)
+	if _, ok := err.(*AcceptedError); ok {
+		return fork, resp, err
+	}
 	if err != nil {
 		return nil, resp, err
 	}
--- a/github/repos_forks_test.go
+++ b/github/repos_forks_test.go
@@ -72,6 +72,30 @@ func TestRepositoriesService_CreateFork(t *testing.T) {
 	}
 }
 
+func TestRepositoriesService_CreateFork_deferred(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/forks", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+		testFormValues(t, r, values{"organization": "o"})
+		// This response indicates the fork will happen asynchronously.
+		w.WriteHeader(http.StatusAccepted)
+		fmt.Fprint(w, `{"id":1}`)
+	})
+
+	opt := &RepositoryCreateForkOptions{Organization: "o"}
+	repo, _, err := client.Repositories.CreateFork(context.Background(), "o", "r", opt)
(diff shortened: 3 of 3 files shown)


shurcooL:
This is related to #862. (Also #486, #488 where `AcceptedError` was initially designed and implemented.) We can use information there when reviewing this.

dsymonds:
I'm not clear what's happening, or who this is blocked on. #862 hasn't had any activity in a month.

dsymonds:
cc @gmlewis

gmlewis:
Sorry I missed this.  Ideally, we would like to do something similar to this for all endpoints that potentially return 202, but my opinion is to go ahead and move forward with this, since there is an immediate need.
Then, if it becomes clear that more endpoints need processing like this, we can address that then.

gmlewis:

github/repos_forks.go
```diff
@@ -77,6 +79,10 @@ func (s *RepositoriesService) CreateFork(ctx context.Context, owner, repo string
 
 	fork := new(Repository)
 	resp, err := s.client.Do(ctx, req, fork)
+	if _, ok := err.(*AcceptedError); ok {
+		json.NewDecoder(resp.Body).Decode(fork)
+		return fork, resp, err
+	}
 	if err != nil {
```
shurcooL: I'd suggest merging this with the if statement above, so it's a little easier to see that this if statement deals with the same `err` value from line 81:

```Go
if _, ok := err.(*AcceptedError); ok {
	...
} else if err != nil {
	...
}
```

dsymonds:
Good call on adding a test. It turned up a bug in Client.Do that would close the response body on error, preventing the caller from being able to process it at all.

github/repos_forks.go
dsymonds: I think it's fine keeping them separate. They are very different code paths; the `AcceptedError` case is only nominally an error.

github/github.go
```diff
@@ -487,8 +487,13 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
 	if err != nil {
-		// even though there was an error, we still return the response
-		// in case the caller wants to inspect it further
+		// Even though there was an error, we still return the response
+		// in case the caller wants to inspect it further.
+		// Since we close the original response body upon returning from this
+		// function, read its contents now so the caller can use it,
+		// and wrap it such that it doesn't matter if they close it.
+		body, _ := ioutil.ReadAll(response.Body)
+		response.Body = ioutil.NopCloser(bytes.NewReader(body))
```
shurcooL: Doing this makes sense, but I don't think it's the right place.

We don't want it to run in _all_ cases when `err != nil`. If you look at the code of `CheckResponse`, it does the following things:

1. if status code is accepted, it immediately returns `&AcceptedError{}`
2. if status code is in 2xx range, it returns `nil`
3. it reads the response body into a byte slice (`data, err := ioutil.ReadAll(r.Body)`)
4. it returns some specific error type or a general `&ErrorResponse{Response: r}`

We don't want this code to run after `ioutil.ReadAll(r.Body)` already happened in `CheckResponse`.

I think a good fix is to move this code into `CheckResponse`'s first case:

```Go
func CheckResponse(r *http.Response) error {
	if r.StatusCode == http.StatusAccepted {
		// Since Do closes the original response body upon returning,
		// read its contents now so the caller can use it,
		// and wrap it such that it doesn't matter if they close it.
		body, _ := ioutil.ReadAll(r.Body)
		r.Body = ioutil.NopCloser(bytes.NewReader(body))
		return &AcceptedError{}
	}
	...
}
```

shurcooL: This gives me a followup idea. Instead of using `r.Body` as a way of transmitting the body from here into `CreateFork`, we can put it inside `AcceptedError`:

```Go
type AcceptedError struct {
	body []byte
}

func CheckResponse(r *http.Response) error {
	if r.StatusCode == http.StatusAccepted {
		// Since Do closes the original response body upon returning,
		// read its contents now so that CreateFork can use it.
		body, _ := ioutil.ReadAll(r.Body)
		return &AcceptedError{body: body}
	}
	...
}
```

Then `CreateFork` code becomes more clear:

```diff
 	fork := new(Repository)
 	resp, err := s.client.Do(ctx, req, fork)
+	if ae, ok := err.(*AcceptedError); ok {
+		json.NewDecoder(bytes.NewReader(ae.body)).Decode(fork)
+		return fork, resp, err
+	}
 	if err != nil {
```

I think this has the benefits of not changing externally visible behavior of the library, and it's easier to read and be sure is correct. If `resp.Body` is modified as currently done, then all other endpoints which don't specifically handle `AcceptedError` will be returning a resp with a body. Users may start to rely on that, and we don't neccessary want to promise it'll always work that way.

In the future, we can consider either exporting `AcceptedError.body` so users of this package can use it, or implement a different solution if we find a better one.

What do you think about this idea?

shurcooL:
One more thing. When making changes to this PR, you don't have to squash and force push. Feel free to send distinct followup commits into this PR. We will squash them into a single commit when merging this PR.

GitHub PRs don't have a way of seeing history, so keeping changes as separate commits makes them easier to review.

github/github.go
dsymonds: Doing that in `CheckResponse` merely hides the issue from the non-trivial control flow in `Do` (which is where the body closing is happening). Right now, `CheckResponse` doesn't touch the body in the HTTP 202 case, just like in any of the other 2xx cases. Your suggestion would mean that `Do` needs to know that `CheckResponse` does that transformation and therefore closing the body is okay.

Now `Do` currently handles all the non-error decoding, so perhaps it should handle this case too. It has the destination object there already. Then the body closing is moot. Let me take that approach; I think it may be the best option.

dsymonds:
PTAL. I think this is much more straightforward now.

github/github.go
dsymonds: In regard to your follow-up idea: I think that's unnecessary. The AcceptedError flows are generally all places where the response body is some subset of the regular response (that's why it is a 2xx code rather than a different code), so forcing the callers to do the parsing seems like unnecessary work. I think my latest revision (having Do do the decoding when there's an AcceptedError) is the simplest option, and it still preserves existing code behaviour unless the caller is doing something truly odd.

gmlewis:
Very nice!  Thank you, @dsymonds and @shurcooL!  I like this a lot.
LGTM.
Merging.

Issue #876, opened 2018-03-22
Support preview Protected Branches API


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2018-03-16-protected-branches-required-approving-reviews/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue, and we can send you an invite to contribute to the repo (which you then accept after you enable two-factor authentication) and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

scriptonist:
@gmlewis Is it okay take up this issue? I've already made a couple of contributions in the past.

gmlewis:
Sounds great, @scriptonist! Thank you.

scriptonist:
I guess the preview period for  [protected branches](https://developer.github.com/changes/2015-11-11-protected-branches-api/)  has ended which is detailed [here](https://developer.github.com/changes/2017-09-06-protected-branches-preview-end/).

gmlewis:
Fixed by #880.

Issue #879, opened 2018-03-27
Extending branch protection to include number of reviewers


victoryNap:
Currently the PullRequestReviewsEnforcement struct doesn't contain the `required_approving_review_count` object that is available through the API. Working on a PR to add it.

https://github.com/google/go-github/blob/master/github/repos.go#L562-L568

https://developer.github.com/v3/repos/branches/#get-branch-protection

victoryNap:
@gmlewis would you mind taking a look at this?

gmlewis:
Thanks, @victoryNap. I commented on the PR.

victoryNap:
Reviewing and making changes now. Thanks for the feedback.

victoryNap:
@gmlewis @shurcooL When you get a minute, I've updated the PR for this.

wesleimp:
@gmlewis I think we can close this issue, right?

gmlewis:
Thank you, @wesleimp, I believe you are correct.

If anyone disagrees, feel free to reopen and explain why.  Thanks.

Closing.

PR #880, opened 2018-03-28
Add RequiredApprovingReviewCount to PullRequestReviewsEnforcement

victoryNap:
Added `RequiredApprovingReviewCount` to `PullRequestReviewsEnforcement`

Related to the following issue - https://github.com/google/go-github/issues/879

This count represents the number of approvals required on a pull request before it can be merged.

This will complete the model for the `required_pull_request_reviews` object which is returned by the GitHub API. 

https://developer.github.com/v3/repos/branches/#get-branch-protection

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -111,8 +111,16 @@ const (
 	// https://developer.github.com/changes/2018-01-25-organization-invitation-api-preview/
 	mediaTypeOrganizationInvitationPreview = "application/vnd.github.dazzler-preview+json"
 
+
 	// https://developer.github.com/changes/2018-03-16-protected-branches-required-approving-reviews/
 	mediaTypeRequiredApprovingReviewsPreview = "application/vnd.github.luke-cage-preview+json"
+
+	// https://developer.github.com/changes/2018-02-22-label-description-search-preview/
+	mediaTypeLabelDescriptionSearchPreview = "application/vnd.github.symmetra-preview+json"
+
+	// https://developer.github.com/changes/2018-02-07-team-discussions-api/
+	mediaTypeTeamDiscussionsPreview = "application/vnd.github.echo-preview+json"
+
 )
 
 // A Client manages communication with the GitHub API.
@@ -154,6 +162,7 @@ type Client struct {
 	Reactions      *ReactionsService
 	Repositories   *RepositoriesService
 	Search         *SearchService
+	Teams          *TeamsService
 	Users          *UsersService
 }
 
@@ -244,6 +253,7 @@ func NewClient(httpClient *http.Client) *Client {
 	c.Reactions = (*ReactionsService)(&c.common)
 	c.Repositories = (*RepositoriesService)(&c.common)
 	c.Search = (*SearchService)(&c.common)
+	c.Teams = (*TeamsService)(&c.common)
 	c.Users = (*UsersService)(&c.common)
 	return c
 }
@@ -487,18 +497,25 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 	err = CheckResponse(resp)
 	if err != nil {
-		// even though there was an error, we still return the response
-		// in case the caller wants to inspect it further
-		return response, err
+		// Even though there was an error, we still return the response
+		// in case the caller wants to inspect it further.
+		// However, if the error is AcceptedError, decode it below before
+		// returning from this function and closing the response body.
+		if _, ok := err.(*AcceptedError); !ok {
+			return response, err
+		}
 	}
 
 	if v != nil {
 		if w, ok := v.(io.Writer); ok {
 			io.Copy(w, resp.Body)
 		} else {
-			err = json.NewDecoder(resp.Body).Decode(v)
-			if err == io.EOF {
-				err = nil // ignore EOF errors caused by empty response body
+			decErr := json.NewDecoder(resp.Body).Decode(v)
+			if decErr == io.EOF {
+				decErr = nil // ignore EOF errors caused by empty response body
+			}
+			if decErr != nil {
+				err = decErr
 			}
 		}
 	}
@@ -687,7 +704,7 @@ func CheckResponse(r *http.Response) error {
 			Response: errorResponse.Response,
 			Message:  errorResponse.Message,
 		}
-	case r.StatusCode == http.StatusForbidden && errorResponse.DocumentationURL == "https://developer.github.com/v3/#abuse-rate-limits":
+	case r.StatusCode == http.StatusForbidden && strings.HasSuffix(errorResponse.DocumentationURL, "/v3/#abuse-rate-limits"):
 		abuseRateLimitError := &AbuseRateLimitError{
 			Response: errorResponse.Response,
 			Message:  errorResponse.Message,
--- a/github/github.go
+++ b/github/github.go
@@ -111,7 +111,6 @@ const (
 	// https://developer.github.com/changes/2018-01-25-organization-invitation-api-preview/
 	mediaTypeOrganizationInvitationPreview = "application/vnd.github.dazzler-preview+json"
 
-
 	// https://developer.github.com/changes/2018-03-16-protected-branches-required-approving-reviews/
 	mediaTypeRequiredApprovingReviewsPreview = "application/vnd.github.luke-cage-preview+json"
 
@@ -120,7 +119,6 @@ const (
 
 	// https://developer.github.com/changes/2018-02-07-team-discussions-api/
 	mediaTypeTeamDiscussionsPreview = "application/vnd.github.echo-preview+json"
-
 )
 
 // A Client manages communication with the GitHub API.
--- a/github/github.go
+++ b/github/github.go
@@ -51,9 +51,6 @@ const (
 	// https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/
 	mediaTypeStarringPreview = "application/vnd.github.v3.star+json"
 
-	// https://developer.github.com/changes/2015-11-11-protected-branches-api/
-	mediaTypeProtectedBranchesPreview = "application/vnd.github.loki-preview+json"
-
 	// https://help.github.com/enterprise/2.4/admin/guides/migrations/exporting-the-github-com-organization-s-repositories/
 	mediaTypeMigrationsPreview = "application/vnd.github.wyandotte-preview+json"
 
@@ -113,6 +110,9 @@ const (
 
 	// https://developer.github.com/changes/2018-01-25-organization-invitation-api-preview/
 	mediaTypeOrganizationInvitationPreview = "application/vnd.github.dazzler-preview+json"
+
+	// https://developer.github.com/changes/2018-03-16-protected-branches-required-approving-reviews/
+	mediaTypeRequiredApprovingReviewsPreview = "application/vnd.github.luke-cage-preview+json"
 )
 
 // A Client manages communication with the GitHub API.
--- a/github/repos.go
+++ b/github/repos.go
@@ -566,6 +566,8 @@ type PullRequestReviewsEnforcement struct {
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged
+	RequiredApprovingReviewCount int `json:"required_approving_review_count"`
 }
 
 // PullRequestReviewsEnforcementRequest represents request to set the pull request review
@@ -578,6 +580,8 @@ type PullRequestReviewsEnforcementRequest struct {
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged
+	RequiredApprovingReviewCount int `json:"required_approving_review_count"`
 }
 
 // MarshalJSON implements the json.Marshaler interface.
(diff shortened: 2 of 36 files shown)


gmlewis:
Thanks, @victoryNap. Some general feedback before a full review.

First, we have to be very careful with "omitempty" and also with the documentation here. In the GitHub Developer API documentation, there is one place where the field "must be" provided and another place where the field "can be" provided. The latter case needs to have "omitempty" and must be a *pointer* to an int so that the field can intentionally not be sent.

Also, the documentation needs to state here that the field must be a value between 1 and 6 for it to be valid.

Finally, all comments should be complete sentences with a period at the end so that the resulting Godocs read correctly.  For example, see:
https://godoc.org/github.com/google/go-github/github

I don't have time for a full review at the moment, but saw these issues at a quick glance.  Please address these first and make sure your documentation also specifies when the field is required and when it is optional... then please ping me and I'll do a full review.

Thanks again!

github/repos.go
```diff
@@ -579,6 +581,9 @@ type PullRequestReviewsEnforcementRequest struct {
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
+	// Can be omitted.
+	RequiredApprovingReviewCount *int `json:"required_approving_review_count,omitempty"`
```
shurcooL: This parameter appears to be mandatory, not optional, here.

https://developer.github.com/v3/repos/branches/#update-branch-protection lists `required_approving_review_count` in a table labeled as:

> The `required_pull_request_reviews` object **must** have the following keys

Emphasis on _must_.

So, it should be:

```Go
RequiredApprovingReviewCount int `json:"required_approving_review_count"`
```

github/repos_test.go
```diff
@@ -555,7 +555,8 @@ func TestRepositoriesService_GetBranchProtection(t *testing.T) {
 					{Slug: String("t"), ID: Int64(4)},
 				},
 			},
-			RequireCodeOwnerReviews: true,
+			RequireCodeOwnerReviews:      true,
+			RequiredApprovingReviewCount: 0,
```
shurcooL: Why is the value 3 in the JSON above:

```JSON
"required_approving_review_count":3
```

But `RequiredApprovingReviewCount` is 0 here?

github/repos_test.go
```diff
@@ -629,7 +630,8 @@ func TestRepositoriesService_UpdateBranchProtection(t *testing.T) {
 					{Slug: String("tt"), ID: Int64(4)},
 				},
 			},
-			RequireCodeOwnerReviews: true,
+			RequireCodeOwnerReviews:      true,
+			RequiredApprovingReviewCount: 0,
```
shurcooL: Let's give it a value in the valid 1-6 range, since it's documented that way:

> Use a number between 1 and 6.

github/repos_test.go
```diff
@@ -776,7 +778,8 @@ func TestRepositoriesService_GetPullRequestReviewEnforcement(t *testing.T) {
 				{Slug: String("t"), ID: Int64(2)},
 			},
 		},
-		RequireCodeOwnerReviews: true,
+		RequireCodeOwnerReviews:      true,
+		RequiredApprovingReviewCount: 0,
```
shurcooL: Also here.

github/repos_test.go
```diff
@@ -851,7 +855,8 @@ func TestRepositoriesService_DisableDismissalRestrictions(t *testing.T) {
 			Users: []*User{},
 			Teams: []*Team{},
 		},
-		RequireCodeOwnerReviews: true,
+		RequireCodeOwnerReviews:      true,
+		RequiredApprovingReviewCount: 0,
```
shurcooL: Also here.

github/repos_test.go
```diff
@@ -532,7 +532,7 @@ func TestRepositoriesService_GetBranchProtection(t *testing.T) {
 
 		testMethod(t, r, "GET")
 		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
-		fmt.Fprintf(w, `{"required_status_checks":{"strict":true,"contexts":["continuous-integration"]},"required_pull_request_reviews":{"dismissal_restrictions":{"users":[{"id":3,"login":"u"}],"teams":[{"id":4,"slug":"t"}]},"dismiss_stale_reviews":true,"require_code_owner_reviews":true},"enforce_admins":{"url":"/repos/o/r/branches/b/protection/enforce_admins","enabled":true},"restrictions":{"users":[{"id":1,"login":"u"}],"teams":[{"id":2,"slug":"t"}]}}`)
+		fmt.Fprintf(w, `{"required_status_checks":{"strict":true,"contexts":["continuous-integration"]},"required_pull_request_reviews":{"dismissal_restrictions":{"users":[{"id":3,"login":"u"}],"teams":[{"id":4,"slug":"t"}]},"dismiss_stale_reviews":true,"require_code_owner_reviews":true},"enforce_admins":{"url":"/repos/o/r/branches/b/protection/enforce_admins","enabled":true},"restrictions":{"users":[{"id":1,"login":"u"}],"teams":[{"id":2,"slug":"t"}]},"required_approving_review_count":3}`)
```
shurcooL: I see why it's 0 below. It's because the `"required_approving_review_count":3` entry is added in the wrong place here. It should be added inside the `"required_pull_request_reviews":{...}` object.

github/repos.go
victoryNap: That was my initial thought as well, according to the docs its required. That said, the request has worked thus far without having it?

github/repos_test.go
victoryNap: If the value hasn't been set, by default, it comes back as 0. I'll update the documents.

github/repos.go
shurcooL: The documentation at https://developer.github.com/v3/repos/branches/#update-branch-protection has this note (which links to https://developer.github.com/changes/2018-03-16-protected-branches-required-approving-reviews/):

image

It doesn't look like we're setting that `application/vnd.github.luke-cage-preview+json` custom Accept header, are we?

github/repos.go
```diff
@@ -579,6 +582,10 @@ type PullRequestReviewsEnforcementRequest struct {
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
+	// Valid values are 1-6.
+	// Can be omitted.
+	RequiredApprovingReviewCount int `json:"required_approving_review_count,omitempty"`
```
gmlewis: If it can be omitted, then I think we need to make this type `*int` instead of `int`.

github/repos.go
```diff
@@ -591,6 +598,10 @@ type PullRequestReviewsEnforcementUpdate struct {
 	DismissStaleReviews *bool `json:"dismiss_stale_reviews,omitempty"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews,omitempty"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
+	// Valid values are 1 - 6.
+	// Can be omitted.
+	RequiredApprovingReviewCount int `json:"required_approving_review_count,omitempty"`
```
gmlewis: Same here: `*int`.

github/repos.go
victoryNap: Issue with the comment, I forgot to update it. Its a required field.

github/github.go
```diff
@@ -51,9 +51,6 @@ const (
 	// https://developer.github.com/changes/2014-12-09-new-attributes-for-stars-api/
 	mediaTypeStarringPreview = "application/vnd.github.v3.star+json"
 
-	// https://developer.github.com/changes/2015-11-11-protected-branches-api/
-	mediaTypeProtectedBranchesPreview = "application/vnd.github.loki-preview+json"
```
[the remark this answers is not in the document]
victoryNap: I also didn't see an update saying it was no longer a preview, but thats the assumption that I made, their documentation makes no reference to the previous preview header, only to the new one that I've added.

screen shot 2018-05-10 at 7 58 52 pm

github/repos.go
victoryNap: Same as above

victoryNap:
I tested this on github, and it worked! What kind of logs would you like to see?

github/repos.go
gmlewis: Then in that case, should we get rid of the `,omitempty` here and below?

gmlewis:
Thanks for testing it, @victoryNap! I don't need to see the logs.

github/repos_test.go
```diff
@@ -531,8 +533,37 @@ func TestRepositoriesService_GetBranchProtection(t *testing.T) {
 		json.NewDecoder(r.Body).Decode(v)
 
 		testMethod(t, r, "GET")
-		testHeader(t, r, "Accept", mediaTypeProtectedBranchesPreview)
-		fmt.Fprintf(w, `{"required_status_checks":{"strict":true,"contexts":["continuous-integration"]},"required_pull_request_reviews":{"dismissal_restrictions":{"users":[{"id":3,"login":"u"}],"teams":[{"id":4,"slug":"t"}]},"dismiss_stale_reviews":true,"require_code_owner_reviews":true},"enforce_admins":{"url":"/repos/o/r/branches/b/protection/enforce_admins","enabled":true},"restrictions":{"users":[{"id":1,"login":"u"}],"teams":[{"id":2,"slug":"t"}]}}`)
+		// TODO: remove custom Accept header when this API fully launches
+		testHeader(t, r, "Accept", mediaTypeRequiredApprovingReviewsPreview)
+		fmt.Fprintf(w, `{
+							"required_status_checks":{
```
gmlewis: I'm thinking we don't need the large amount of indentation here (and below).
What about un-indenting this entire region?  Something like:

```go
		fmt.Fprintf(w, `{
"required_status_checks":{
	"strict":true,
	"contexts":["continuous-integration"]
},
"required_pull_request_reviews":{
...
```
?

github/repos.go
```diff
@@ -579,6 +582,9 @@ type PullRequestReviewsEnforcementRequest struct {
 	DismissStaleReviews bool `json:"dismiss_stale_reviews"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
+	// Valid values are 1-6.
+	RequiredApprovingReviewCount int `json:"required_approving_review_count,omitempty"`
```
gmlewis: What are your thoughts, @victoryNap on this `,omitempty`?
I'm thinking it should be removed since it is a required parameter.

github/repos.go
```diff
@@ -591,6 +597,9 @@ type PullRequestReviewsEnforcementUpdate struct {
 	DismissStaleReviews *bool `json:"dismiss_stale_reviews,omitempty"`
 	// RequireCodeOwnerReviews specifies if an approved review is required in pull requests including files with a designated code owner.
 	RequireCodeOwnerReviews bool `json:"require_code_owner_reviews,omitempty"`
+	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
+	// Valid values are 1 - 6.
+	RequiredApprovingReviewCount int `json:"required_approving_review_count,omitempty"`
```
gmlewis: ...and this one too.

victoryNap:
@gmlewis I don't know that having the block totally unindented looks makes sense, but I agree, I was overly aggressive with the tabs...

gmlewis:
It looks like we have a CI failure and some tests need updating:

```
--- FAIL: TestPullRequestReviewsEnforcementRequest_MarshalJSON_nilDismissalRestirctions (0.00s)
	repos_test.go:1034: PullRequestReviewsEnforcementRequest.MarshalJSON returned {"dismiss_stale_reviews":false,"require_code_owner_reviews":false,"required_approving_review_count":0}, want {"dismiss_stale_reviews":false,"require_code_owner_reviews":false}
	repos_test.go:1048: PullRequestReviewsEnforcementRequest.MarshalJSON returned {"dismissal_restrictions":{},"dismiss_stale_reviews":false,"require_code_owner_reviews":false,"required_approving_review_count":0}, want {"dismissal_restrictions":{},"dismiss_stale_reviews":false,"require_code_owner_reviews":false}
	repos_test.go:1065: PullRequestReviewsEnforcementRequest.MarshalJSON returned {"dismissal_restrictions":{"users":[],"teams":[]},"dismiss_stale_reviews":false,"require_code_owner_reviews":false,"required_approving_review_count":0}, want {"dismissal_restrictions":{"users":[],"teams":[]},"dismiss_stale_reviews":false,"require_code_owner_reviews":false}
```

Can you please look into this, @victoryNap?

victoryNap:
@gmlewis I'll take a look. Probably a result of my indentation changes

victoryNap:
@gmlewis Fixed :)

@shurcooL Mind taking a look when you get a chance?

victoryNap:
@gmlewis who else can approve this PR before it needs to be merged?

gmlewis:
I don't believe that there is any strict rule for code reviewers for this repo... we have typically relied on active contributors but I know we lost some people when the repo required two-factor auth to be enabled, unfortunately.

For non-trivial changes, we like to have a second set of eyes review the code and check that things look good with respect to the GitHub Developer API docs.

So I believe we welcome all contributors to check for the `NeedsReview` label and provide any feedback, including full reviews and LGTMs.

@willnorris can correct me if my understanding is incorrect.

victoryNap:
@willnorris @dmitshur Can I get another review on this?

victoryNap:
@gmlewis I really need this feature added in. Any thoughts on getting another review?

gmlewis:
Sorry about that... Let's see if @juliaferraioli has time.

gmlewis:
I'm hearing no objections.
Merging.

gmlewis:
Fixes #876.

tadityar:
I think this also Fixes #905.

PR #883, opened 2018-03-28
Replace `int` id with `int64` in OrganizationsService

mspiegel:
I think the OrganizationsService didn't get converted in the int -> int64 id migration.

Updates #597.


shurcooL:
You're right.

There's a few extra I found:

- `IssuesService.GetComment`
- `IssuesService.EditComment`
- `IssuesService.DeleteComment`
- `RepositoriesService.EditKey`
- `RepositoriesService.DeleteKey`

Would you like to expand the scope of your PR and fix those instances as well? If not, np, we can do it as part of a separate PR.

mspiegel:
Yeah I can add those too

gmlewis:
Thank you, @mspiegel and @shurcooL!
LGTM.
Merging.

PR #886, opened 2018-03-29
Rename id parameter to commentID in IssuesService comment methods.

### LINKED ISSUE #885: Rename "id" parameter to "commentID" in IssuesService.
The endpoints that touch comments in `IssuesService` are very different from ones that touch issues. Comment-related endpoints take the comment ID (an `int64` now), while issue-related endpoints use the issue number (an `int`):

```Go
IssuesService.Get       (ctx context.Context, owner string, repo string, number int) (*Issue, *Response, error)
IssuesService.GetComment(ctx context.Context, owner string, repo string, id int64)   (*IssueComment, *Response, error)
```

Using `id` can be misleading, people can think it refers to the issue ID.

`GistsService` sets a good example by calling the parameter `commentID` in its `GetComment` method:

```
GistsService.GetComment(ctx context.Context, gistID string, commentID int64) (*GistComment, *Response, error)
```

I think we should rename `id` to `commentID` in `IssuesService` to make it more clear.

This problem is exacerbated in the `EditComment` endpoints because they take an `id int64` and the `*github.IssueComment` itself also contains an `ID int64` (but setting it has no effect; it's there for the Get/List operations, not Edit).

(I've also noticed a similar issue in `PullRequestsService` and its comment-related methods, but I'll make that a separate issue.)

shurcooL:
Also document which `IssueComment` fields need to be set in `EditComment` method.

This should help improve clarity of the API, and reduce the chance to mix up the issue vs comment IDs.

Resolves #885.

### DIFF
--- a/github/issues_comments.go
+++ b/github/issues_comments.go
@@ -79,8 +79,8 @@ func (s *IssuesService) ListComments(ctx context.Context, owner string, repo str
 // GetComment fetches the specified issue comment.
 //
 // GitHub API docs: https://developer.github.com/v3/issues/comments/#get-a-single-comment
-func (s *IssuesService) GetComment(ctx context.Context, owner string, repo string, id int64) (*IssueComment, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/issues/comments/%d", owner, repo, id)
+func (s *IssuesService) GetComment(ctx context.Context, owner string, repo string, commentID int64) (*IssueComment, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/issues/comments/%d", owner, repo, commentID)
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


github/issues_comments.go
```diff
@@ -118,10 +118,11 @@ func (s *IssuesService) CreateComment(ctx context.Context, owner string, repo st
 }
 
 // EditComment updates an issue comment.
+// A non-nil comment.Body must be provided. Other comment fields should be left nil.
```
shurcooL: An additional step we can consider doing is checking if any of the `comment` fields other than `Body` are set (especially `ID`), and return an error with helpful text if so.

GitHub API documents that the only input field is `body`, and it is required. See https://developer.github.com/v3/issues/comments/#edit-a-comment:

image

A breaking API change to replace `IssueComment` parameter with a new `IssueCommentEdit` struct containing only the `Body` field is another option, but that's a breaking API change.

PR #888, opened 2018-03-30
Change number int to commentID int64 in PRsService comment methods.

shurcooL:
**This is a breaking API change that is a part of issue #597.** It should've been applied earlier, but we missed it because the parameter was misleadingly named `number` rather than `id` or `commentID`.

Rename the parameter to `commentID` to make it more clear that it's the comment ID and not the PR ID nor PR number. This should help improve clarity of the API, and reduce the chance to mix up the IDs.

Also document which `PullRequestComment` fields need to be set in `EditComment` method.

Similar to #886.
Updates #885.
Updates #597.


shurcooL:
Thanks for reviewing.

Since it's a breaking change, I'll wait for Monday before merging (so people can have a nice weekend).

PR #889, opened 2018-04-03
Support preview Team Discussions API

### LINKED ISSUE #849: Support preview Team Discussions API

parkhyukjun89:
- Add a new service TeamsService to support API related to Organization Teams https://developer.github.com/v3/teams/
- Add methods to support new preview API for team discussions https://developer.github.com/v3/teams/discussions/
- Add methods to support new preview API for team discussion comments https://developer.github.com/v3/teams/discussion_comments/

Fixes #849 
https://developer.github.com/changes/2018-02-07-team-discussions-api/#new-rest-api-endpoints

Second part of the change https://developer.github.com/changes/2018-02-07-team-discussions-api/#updates-to-the-teams-api-documentation will be done in a separate PR, once this PR is merged.

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -1,0 +1,7 @@
+package github
+
+// TeamsService provides access to the team related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/
+type TeamsService service
--- a/github/teams.go
+++ b/github/teams.go
@@ -1,6 +1,6 @@
 package github
 
-// TeamsService provides access to the team related functions
+// TeamsService provides access to the team-related functions
 // in the GitHub API.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/
--- a/github/teams_discussion_comments.go
+++ b/github/teams_discussion_comments.go
@@ -7,7 +7,6 @@ package github
 
 import (
 	"context"
-	"errors"
 	"fmt"
 )
 
@@ -43,8 +42,8 @@ type DiscussionCommentListOptions struct {
 // Authenticated user must grant read:discussion scope.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#list-comments
-func (s *TeamsService) ListComments(ctx context.Context, team, discussion int64, options *DiscussionCommentListOptions) ([]*DiscussionComment, *Response, error) {
-	u := fmt.Sprintf("teams/%v/discussions/%v/comments", team, discussion)
+func (s *TeamsService) ListComments(ctx context.Context, teamID int64, discussionNumber int, options *DiscussionCommentListOptions) ([]*DiscussionComment, *Response, error) {
+	u := fmt.Sprintf("teams/%v/discussions/%v/comments", teamID, discussionNumber)
 	u, err := addOptions(u, options)
 	if err != nil {
 		return nil, nil, err
@@ -71,8 +70,8 @@ func (s *TeamsService) ListComments(ctx context.Context, team, discussion int64,
 // Authenticated user must grant read:discussion scope.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#get-a-single-comment
-func (s *TeamsService) GetComment(ctx context.Context, team, discussion, id int64) (*DiscussionComment, *Response, error) {
-	u := fmt.Sprintf("teams/%v/discussions/%v/comments/%v", team, discussion, id)
+func (s *TeamsService) GetComment(ctx context.Context, teamID int64, discussionNumber, commentNumber int) (*DiscussionComment, *Response, error) {
+	u := fmt.Sprintf("teams/%v/discussions/%v/comments/%v", teamID, discussionNumber, commentNumber)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
@@ -94,12 +93,8 @@ func (s *TeamsService) GetComment(ctx context.Context, team, discussion, id int6
 // Authenticated user must grant write:discussion scope.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#create-a-comment
-func (s *TeamsService) CreateComment(ctx context.Context, team, discussion int64, comment *DiscussionComment) (*DiscussionComment, *Response, error) {
-	if comment == nil {
-		return nil, nil, errors.New("comment must be provided")
-	}
-
-	u := fmt.Sprintf("teams/%v/discussions/%v/comments", team, discussion)
+func (s *TeamsService) CreateComment(ctx context.Context, teamID int64, discsusionNumber int, comment DiscussionComment) (*DiscussionComment, *Response, error) {
+	u := fmt.Sprintf("teams/%v/discussions/%v/comments", teamID, discsusionNumber)
 	req, err := s.client.NewRequest("POST", u, comment)
 	if err != nil {
 		return nil, nil, err
@@ -122,12 +117,8 @@ func (s *TeamsService) CreateComment(ctx context.Context, team, discussion int64
 // User is allowed to edit body of a comment only.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#edit-a-comment
-func (s *TeamsService) EditComment(ctx context.Context, team, discussion, id int64, comment *DiscussionComment) (*DiscussionComment, *Response, error) {
-	if comment == nil {
-		return nil, nil, errors.New("comment must be provided")
-	}
-
-	u := fmt.Sprintf("teams/%v/discussions/%v/comments/%v", team, discussion, id)
+func (s *TeamsService) EditComment(ctx context.Context, teamID int64, discussionNumber, commentNumber int, comment DiscussionComment) (*DiscussionComment, *Response, error) {
+	u := fmt.Sprintf("teams/%v/discussions/%v/comments/%v", teamID, discussionNumber, commentNumber)
 	req, err := s.client.NewRequest("PATCH", u, comment)
 	if err != nil {
 		return nil, nil, err
@@ -149,8 +140,8 @@ func (s *TeamsService) EditComment(ctx context.Context, team, discussion, id int
 // Authenticated user must grant write:discussion scope.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#delete-a-comment
-func (s *TeamsService) DeleteComment(ctx context.Context, team, discussion, id int64) (*Response, error) {
-	u := fmt.Sprintf("teams/%v/discussions/%v/comments/%v", team, discussion, id)
+func (s *TeamsService) DeleteComment(ctx context.Context, teamID int64, discussionNumber, commentNumber int) (*Response, error) {
+	u := fmt.Sprintf("teams/%v/discussions/%v/comments/%v", teamID, discussionNumber, commentNumber)
 	req, err := s.client.NewRequest("DELETE", u, nil)
 	if err != nil {
 		return nil, err
--- a/github/teams_discussion_comments.go
+++ b/github/teams_discussion_comments.go
@@ -7,10 +7,11 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
 )
 
-// DiscussionComment represents a GitHub dicussion in a team
+// DiscussionComment represents a GitHub dicussion in a team.
 type DiscussionComment struct {
 	Author        *User      `json:"author,omitempty"`
 	Body          *string    `json:"body,omitempty"`
@@ -31,7 +32,7 @@ func (c DiscussionComment) String() string {
 }
 
 // DiscussionCommentListOptions specifies optional parameters to the
-// TeamServices.ListComments method
+// TeamServices.ListComments method.
 type DiscussionCommentListOptions struct {
 	// Sorts the discussion comments by the date they were created.
 	// Accepted values are asc and desc. Default is desc.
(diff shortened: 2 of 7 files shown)


github/teams_discussion_comments.go
```diff
@@ -0,0 +1,154 @@
...
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// DiscussionComment represents a GitHub dicussion in a team
```
gmlewis: nit: end comment with a period.

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,154 @@
...
+	UpdatedAt     *Timestamp `json:"updated_at,omitempty"`
+	URL           *string    `json:"url,omitempty"`
+}
+
+func (c DiscussionComment) String() string {
+	return Stringify(c)
+}
+
+// DiscussionCommentListOptions specifies optional parameters to the
+// TeamServices.ListComments method
```
gmlewis: nit: end comment with a period.

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return discussionComment, resp, nil
+}
+
+// CreateComment creates a new discussion post on a team discussion.
+// Authenticated user must grant write:discussion scope.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#create-a-comment
+func (s *TeamsService) CreateComment(ctx context.Context, team, discussion int64, comment *DiscussionComment) (*DiscussionComment, *Response, error) {
+	u := fmt.Sprintf("teams/%v/discussions/%v/comments", team, discussion)
```
gmlewis: I'm thinking we might want to check for `comment == nil` here and return an error if so.

github/teams.go
```diff
@@ -0,0 +1,7 @@
+package github
+
+// TeamsService provides access to the team related functions
```
gmlewis: nit: s/team related/team-related/

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return discussionComment, resp, nil
+}
+
+// EditComment edits the body text of a discussion comment.
+// Authenticated user must grant write:discussion scope.
+// User is allowed to edit body of a comment only.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#edit-a-comment
+func (s *TeamsService) EditComment(ctx context.Context, team, discussion, id int64, comment *DiscussionComment) (*DiscussionComment, *Response, error) {
```
gmlewis: We might also want to check for `comment == nil` here as well.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// TeamDiscussion represents a GitHub dicussion in a team
```
gmlewis: nit: end comment with a period.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+	UpdatedAt     *Timestamp `json:"updated_at,omitempty"`
+	URL           *string    `json:"url,omitempty"`
+}
+
+func (d TeamDiscussion) String() string {
+	return Stringify(d)
+}
+
+// DiscussionListOptions specifies optional parameters to the
+// TeamServices.ListDiscussions method
```
gmlewis: nit: end comment with a period.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+type DiscussionListOptions struct {
+	// Sorts the discussion by the date they were created.
+	// Accepted values are asc and desc. Default is desc.
+	Direction string `url:"direction,omitempty"`
```
gmlewis: Do we also need to include `ListOptions` here?  I haven't yet checked the docs.

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,154 @@
...
+func (c DiscussionComment) String() string {
+	return Stringify(c)
+}
+
+// DiscussionCommentListOptions specifies optional parameters to the
+// TeamServices.ListComments method
+type DiscussionCommentListOptions struct {
+	// Sorts the discussion comments by the date they were created.
+	// Accepted values are asc and desc. Default is desc.
+	Direction string `url:"direction,omitempty"`
```
gmlewis: Do we also need to include `ListOptions` here?  I haven't yet checked the docs.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return teamDiscussion, resp, nil
+}
+
+// CreateDiscussion creates a new discussion post on a team's page.
+// Authenticated user must grant write:discussion scope.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/discussions/#create-a-discussion
+func (s *TeamsService) CreateDiscussion(ctx context.Context, team int64, discussion *TeamDiscussion) (*TeamDiscussion, *Response, error) {
```
gmlewis: We might want to check for `discussion == nil` here and return an error if so.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return teamDiscussion, resp, nil
+}
+
+// EditDiscussion edits the title and body text of a discussion post.
+// Authenticated user must grant write:discussion scope.
+// User is allowed to change Title and Body of a discussion only.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/discussions/#edit-a-discussion
+func (s *TeamsService) EditDiscussion(ctx context.Context, team, id int64, discussion *TeamDiscussion) (*TeamDiscussion, *Response, error) {
```
gmlewis: We might want to check for `discussion == nil` here and return an error if so.

github/teams_discussion_comments.go
parkhyukjun89: Fixed.

github/teams_discussion_comments.go
parkhyukjun89: Fixed.

github/teams_discussion_comments.go
parkhyukjun89: Added a nil check

github/teams.go
parkhyukjun89: Fixed.

github/teams_discussion_comments.go
parkhyukjun89: Added a nil check

github/teams_discussions.go
parkhyukjun89: Fixed.

github/teams_discussions.go
parkhyukjun89: Fixed.

github/teams_discussions.go
parkhyukjun89: doc mentions directions only https://developer.github.com/v3/teams/discussions/#parameters

github/teams_discussion_comments.go
parkhyukjun89: doc mentions directions only https://developer.github.com/v3/teams/discussion_comments/#parameters

github/teams_discussions.go
parkhyukjun89: Added a nil check

github/teams_discussions.go
parkhyukjun89: Added a nil check

parkhyukjun89:
Travis build failed at 'go get' stage. 

> package golang.org/x/tools/cmd/cover: unrecognized import path "golang.org/x/tools/cmd/cover" (https fetch: Get https://golang.org/x/tools/cmd/cover?go-get=1: net/http: TLS handshake timeout)

Is there a way to kick off another build?

shurcooL:
I've restarted the Travis build, it's passing now.

gmlewis:
It looks like merge conflicts need resolving, @parkhyukjun89. Either you can do that, or I can take a look at it later today.

parkhyukjun89:
Resolved merge conflict @gmlewis

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return comments, resp, nil
+}
+
+// GetComment gets a specific comment on a team discussion.
+// Authenticated user must grant read:discussion scope.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#get-a-single-comment
+func (s *TeamsService) GetComment(ctx context.Context, team, discussion, id int64) (*DiscussionComment, *Response, error) {
```
shurcooL: `team, discussion, id int64`

I think we'll need to take care of this (here, and elsewhere).

From looking over the API, it does seem they use "number" instead of "id" to identify discussions and comments within said discussions.

However, I haven't actually used team discussions, so I can't confirm if this matches the reality.

From https://developer.github.com/v3/teams/discussion_comments/#get-a-single-comment, I'm seeing:

```
GET /teams/:team_id/discussions/:discussion_number/comments/:comment_number
```

That tells me that the first parameter is a team ID (which we want to use `int64` for), second one is the discussion number (not ID), which we want to map to `int`, and the comment also seems to be a number rather than ID (hence `int`).

I suggest we follow that closely and therefore use:

```Go
teamID int64, discussionNumber, commentNumber int
```

Let me know what you think, or if you have any concerns.

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// CreateComment creates a new discussion post on a team discussion.
+// Authenticated user must grant write:discussion scope.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/discussion_comments/#create-a-comment
+func (s *TeamsService) CreateComment(ctx context.Context, team, discussion int64, comment *DiscussionComment) (*DiscussionComment, *Response, error) {
+	if comment == nil {
+		return nil, nil, errors.New("comment must be provided")
+	}
```
shurcooL: We did the `if pull == nil` change for existing endpoints where a pointer was already used (see PR #539), in order to avoid making a breaking API change.

Since this is a new API, I think we can simplify things and just make `comment` a value rather than pointer. After all, it's required.

It would be a little inconsistent with previous endpoints that use pointers, but I think it's a beneficial tradeoff. What do you think?

github/teams_discussion_comments.go
shurcooL: Similarly, what do you think of just making `comment` a value rather than pointer here as well?

github/teams_discussion_comments.go
parkhyukjun89: Agreed. Discussion & comment numbers are incrementing numbers.

github/teams_discussion_comments.go
parkhyukjun89: Understood. But, what if user decides to pass in an empty struct. Should that be considered an error?

shurcooL: > But, what if user decides to pass in an empty struct. Should that be considered an error and need to be guarded against?

I think it's okay to let the GitHub API handle such an incorrect request (with a descriptive error message). We shouldn't, because it adds code, and it's not worth it (most people will not try to create an invalid comment, other than during testing). We don't try to catch empty input in all other endpoints either.

Making it a value rather than pointer is more about eliminating an unnecessary choice and making the API more intuitive to use. Since there's no reason to use a pointer (other than being consistent with prior unfortunate API choices, which isn't a strong enough reason I think), we shouldn't use it.

parkhyukjun89:
@shurcooL made changes accordingly

github/teams_discussion_comments.go
```diff
@@ -0,0 +1,154 @@
...
+	Author        *User      `json:"author,omitempty"`
+	Body          *string    `json:"body,omitempty"`
+	BodyHTML      *string    `json:"body_html,omitempty"`
+	BodyVersion   *string    `json:"body_version,omitempty"`
+	CreatedAt     *Timestamp `json:"created_at,omitempty"`
+	LastEditedAt  *Timestamp `json:"last_edited_at,omitempty"`
+	DiscussionURL *string    `json:"discussion_url,omitempty"`
+	HTMLURL       *string    `json:"html_url,omitempty"`
+	NodeID        *string    `json:"node_id,omitempty"`
+	Number        *int64     `json:"number,omitempty"`
```
shurcooL: I just noticed this, it’s not an ID, so it should be `int`. We use `int` type for all numbers other than IDs.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+)
+
+// TeamDiscussion represents a GitHub dicussion in a team.
+type TeamDiscussion struct {
+	Author        *User      `json:"author,omitempty"`
+	Body          *string    `json:"body,omitempty"`
+	BodyHTML      *string    `json:"body_html,omitempty"`
+	BodyVersion   *string    `json:"body_version,omitempty"`
+	CommentsCount *int64     `json:"comments_count,omitempty"`
```
shurcooL: Also here, should be `int`.

github/teams_discussions.go
```diff
@@ -0,0 +1,159 @@
...
+	CommentsURL   *string    `json:"comments_url,omitempty"`
+	CreatedAt     *Timestamp `json:"created_at,omitempty"`
+	LastEditedAt  *Timestamp `json:"last_edited_at,omitempty"`
+	HTMLURL       *string    `json:"html_url,omitempty"`
+	NodeID        *string    `json:"node_id,omitempty"`
+	Number        *int64     `json:"number,omitempty"`
```
shurcooL: Also here.

gmlewis:
Great catches, @shurcooL!  Thank you!

parkhyukjun89:
I will add those fixes in the next pr for https://developer.github.com/changes/2018-02-07-team-discussions-api/#updates-to-the-teams-api-documentation

Issue #875, opened 2018-03-22
Support preview Hovercard API


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2018-03-21-hovercard-api-preview/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue, and we can send you an invite to contribute to the repo (which you then accept after you enable two-factor authentication) and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

MonicB14:
@gmlewis I will be happy to take this issue. Please feel free to assign it to me.

gmlewis:
Thank you, @MonicB14!  I assigned it to you.

PR #890, opened 2018-04-08
Support  Preview hovercard API (#875)

MonicB14:
Add support for new preview Hovercard API as described in:
https://developer.github.com/changes/2018-03-21-hovercard-api-preview/

This PR adds:

- A new method to support the new [Hovercard API](https://developer.github.com/v3/users/#get-contextual-information-about-a-user).

Fixes:  #875


MonicB14:
I signed it!

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	//TODO : remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeHovercardPreview)
+
+	contextResp := new(UserContextualInfoResponse)
```
shurcooL: Give this variable a shorter name. Like `uci` (for User Contextual Info).

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+}
+
+// UserContext represents the contextual information about user.
+type UserContext struct {
+	Message *string `url:"message,omitempty"`
+	Octicon *string `url:"octicon,omitempty"`
+}
+
+// UserContextualInfoResponse holds the parsed response from GetContextualInfo.
+type UserContextualInfoResponse struct {
```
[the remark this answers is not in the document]
shurcooL: Actually, what do you think about calling it `Hovercard`? It's what the API endpoint is called:

```
GET /users/:username/hovercard
```

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
 	return uResp, resp, nil
 }
 
+// UserContextualInfoOptions specifies optional parameters to the UsersService.GetContextualInfo
+// method.
+type UserContextualInfoOptions struct {
```
shurcooL: This will be `HovercardOptions` then.

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+	// SubjectType specifies the additional information to be received about the hovercard.
+	// Possible values are : organization, repository, issue, pull_request.
+	SubjectType string `url:"subject_type,omitempty"`
+
+	// SubjectID specifies the ID for the SubjectType.
+	SubjectID string `url:"subject_id,omitempty"`
```
shurcooL: Both the parameters are required, not optional. So remove `,omitempty` and add `(Required.)` at the end of comment (like [here](https://gotools.org/github.com/google/go-github/github?rev=88eb4e9dc91cb2db1bbc56c936f1ee39bbe72ba5#orgs-teams.go-L118)).

github/users.go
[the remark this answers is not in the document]
MonicB14: This came to my mind but then I thought to keep the names consistent and hence changed it to <code>UserContextualInfoResponse</code>, but <code>Hovercard</code> seems much better name considering the API endpoint.

MonicB14:
Thanks for the suggestions @shurcooL .
PR has been updated to address the suggested changes.

MonicB14:
To use Hovercard api, authentication needs to be in place. 
Should a comment describing this be added in the new method?

shurcooL:
> Should a comment describing this be added in the new method?

Good call, let's do that.

GitHub API v3 already documents that requirement at https://developer.github.com/v3/users/#get-contextual-information-about-a-user, but it will be nice if we mention it here too.

[`UsersService.GetGPGKey`](https://godoc.org/github.com/google/go-github/github#UsersService.GetGPGKey) method serves as a good template for how to do it:

https://github.com/google/go-github/blob/89a446efe88e1636068183a91ca8845683683477/github/users_gpg_keys.go#L77-L81

Also, given we already renamed `UserContextualInfoResponse` to `Hovercard` and the endpoint is `GET /users/:username/hovercard`, I think we should apply the same change to the method name. It can be renamed to `GetHovercard`. What do you think?

So, it can look something like this:

```Go
// GetHovercard gets hovercard information for specified user. It requires authentication
// via Basic Auth or via OAuth with the repo scope.
//
// GitHub API docs: https://developer.github.com/v3/users/#get-contextual-information-about-a-user
func (s *UsersService) GetHovercard(ctx context.Context, user string, opt *HovercardOptions) (*Hovercard, *Response, error)
```

Note that you should return the entire `*Hovercard` from that method, not just `[]*UserContext`. That way, if the hovercard gets extra fields in the future, they'll be a part of the response.

github/users.go
```diff
@@ -171,16 +171,16 @@ func (s *UsersService) GetContextualInfo(ctx context.Context, user string, opt *
 		return nil, nil, err
 	}
 
-	//TODO : remove custom Accept header when this API fully launches.
+	// TODO: remove custom Accept header when this API fully launches.
 	req.Header.Set("Accept", mediaTypeHovercardPreview)
 
-	contextResp := new(UserContextualInfoResponse)
-	resp, err := s.client.Do(ctx, req, contextResp)
+	uci := new(Hovercard)
```
shurcooL: `uci` is no longer applicable. Make it `hc` or `hovercard`.

MonicB14: Oops Sorry! missed it.
I will make the suggested improvements.

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+
+// Hovercard holds the parsed response from GetContextualInfo.
+type Hovercard struct {
+	Contexts []*UserContext `json:"contexts,omitempty"`
+}
+
+// GetContextualInfo fetches contextual information about user.
+//
+// GitHub API docs: https://developer.github.com/v3/users/#get-contextual-information-about-a-user
+func (s *UsersService) GetContextualInfo(ctx context.Context, user string, opt *HovercardOptions) ([]*UserContext, *Response, error) {
```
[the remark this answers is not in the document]
MonicB14: @shurcooL I revisited the API documentation and it says "The <code>subject_type</code> and <code>subject_id</code> parameters provide context for the person's hovercard, which returns more information than without the parameters". 
I tried using the API service without parameters and the response is empty <code>contexts</code> array. (HTTP response code is 200)

Response without parameters:
<code>
{
  "contexts": [  ]
}
</code>

So I was thinking whether to keep this parameters optional or mandatory ?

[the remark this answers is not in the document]
MonicB14: Great! Will keep the <code>opt</code> optional and update the documentation. Should be done by today. Thanks @shurcooL

MonicB14:
I have made the suggested changes.
Apologies for taking so long.

MonicB14:
Hey @shurcooL , any views on this changes ? Let me know if this looks good to you

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+	SubjectID string `url:"subject_id"`
+}
+
+// UserContext represents the contextual information about user.
+type UserContext struct {
+	Message *string `json:"message,omitempty"`
+	Octicon *string `json:"octicon,omitempty"`
+}
+
+// Hovercard holds the parsed response from GetHovercard.
```
shurcooL: It'd be nice if we could better describe what this struct is, rather than saying it's "what GetHovercard returns". Perhaps something like this?

```
// Hovercard represents hovercard information about a user.
```

It's not great, so you can go either way here.

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+type Hovercard struct {
+	Contexts []*UserContext `json:"contexts,omitempty"`
+}
+
+// GetHovercard fetches contextual information about user. It requires authentication
+// via Basic Auth or via OAuth with the repo scope.
+// GitHub API docs: https://developer.github.com/v3/users/#get-contextual-information-about-a-user
```
shurcooL: The `// GitHub API docs: ...` line needs an empty line above, otherwise it gets parsed by godoc as part of the same paragraph, making it less readable. See here for example:

https://github.com/google/go-github/blob/d4c2343a938988db2745d0ed89035deb9200a789/github/users.go#L118-L120

github/users.go
```diff
@@ -134,6 +134,55 @@ func (s *UsersService) Edit(ctx context.Context, user *User) (*User, *Response,
...
+// UserContext represents the contextual information about user.
+type UserContext struct {
+	Message *string `json:"message,omitempty"`
+	Octicon *string `json:"octicon,omitempty"`
+}
+
+// Hovercard holds the parsed response from GetHovercard.
+type Hovercard struct {
+	Contexts []*UserContext `json:"contexts,omitempty"`
+}
```
shurcooL: It would be good to place `Hovercard` on top of `UserContext`. The general pattern is higher level identifiers should come first. They are easier to understand, and they set context for the lower level identifiers that follow. I.e.:

```Go
// Hovercard ...
type Hovercard struct {
    ...
}

// UserContext ...
type UserContext struct {
    ...
}
```

Issue #658, opened 2017-06-19
Raw GetBlob() support


obno:
Current implementation of `Git.GetBlob()` API only supports the JSON media type. This can be inconvenient for vey large git blobs since they are stored completely in memory in the corresponding `github.Blob` struct. 

The Github `GET /repos/:owner/:repo/git/blobs/:sha` API also allows for blobs to be fetched in raw rather than json format using the `application/vnd.github.VERSION.raw` media type. This gets around the above limitation for very large blobs. 

You can abuse the current API like below but it would be nice to have this feature built into the API.

```go
func DownloadFromGithub(accessToken, owner, repo, sha, toFileName string) error{
	//create github client
	httpClient := oauth2.NewClient(context.Background(), oauth2.StaticTokenSource(&oauth2.Token{
		AccessToken: accessToken,
	}))
	ghClient := github.NewClient(httpClient)

	out, err := os.Create(toFileName)
	if err != nil  {
		return err
	}
	defer out.Close()

	//download the git blob as raw data
	u := fmt.Sprintf("repos/%v/%v/git/blobs/%v", owner, repo, sha)
	req, err := ghClient.NewRequest("GET", u, nil)
	if err != nil{
		return err
	}
	req.Header.Del("Accept")
	req.Header.Set("Accept", "application/vnd.github.VERSION.raw")
	req.Header.Del("Content-Type")
	_, err = ghClient.Do(context.Background(), req, out)
	if err != nil{
		return err
	}
	return nil

}
```

Maybe provide something like this?

`func (s *GitService) GetRawBlob(ctx context.Context, owner string, repo string, sha string, to io.Writer) (*Response, error)`

shurcooL:
There was a related issue/PR to add support for fetching raw content of a pull request at https://github.com/google/go-github/pull/481.

I think a solution to this issue is most likely to follow the same pattern as that PR.

obno:
Excellent - thanks!

wpstmehs:
thx

shurcooL:
Another similar change that was recently added was #767. It was for downloading commits in diff or patch format.

kshitij10496:
I think this issue shares a fundamental similarity with #727 
Both revolve around the support for custom media types.

PR #767, opened 2017-10-29
Add RepositoriesService.GetCommitRaw method.

shurcooL:
Follows the same high-level pattern as set out in PR #481 with `PullRequestsService.GetRaw` method. (/cc @huydx FYI.)

Improve documentation wording.

Improve style of `PullRequestsService.GetRaw` and its tests to be more consistent.

Fix minor typo in GitHub documentation URL.

PR #891, opened 2018-04-11
Add GitService.GetBlobRaw.

dsymonds:
This addresses #658.

The name isn't great (I would have preferred `GetRawBlob`), but it matches the previously added `RepositoriesService.GetCommitRaw`. Happy to switch if you'd prefer the better name and don't mind the mild inconsistency.

### DIFF
--- a/github/git_blobs.go
+++ b/github/git_blobs.go
@@ -6,6 +6,7 @@
 package github
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 )
@@ -20,7 +21,7 @@ type Blob struct {
 	NodeID   *string `json:"node_id,omitempty"`
 }
 
-// GetBlob fetchs a blob from a repo given a SHA.
+// GetBlob fetches a blob from a repo given a SHA.
 //
 // GitHub API docs: https://developer.github.com/v3/git/blobs/#get-a-blob
 func (s *GitService) GetBlob(ctx context.Context, owner string, repo string, sha string) (*Blob, *Response, error) {
@@ -36,6 +37,23 @@ func (s *GitService) GetBlob(ctx context.Context, owner string, repo string, sha
 	blob := new(Blob)
 	resp, err := s.client.Do(ctx, req, blob)
 	return blob, resp, err
+}
+
+// GetBlobRaw fetches a blob's contents from a repo.
+// Unlike GetBlob, it returns the raw bytes rather than the base64-encoded data.
+//
+// GitHub API docs: https://developer.github.com/v3/git/blobs/#get-a-blob
+func (s *GitService) GetBlobRaw(ctx context.Context, owner, repo, sha string) ([]byte, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/git/blobs/%v", owner, repo, sha)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	req.Header.Set("Accept", "application/vnd.github.v3.raw")
+
+	var buf bytes.Buffer
+	resp, err := s.client.Do(ctx, req, &buf)
+	return buf.Bytes(), resp, err
 }
 
 // CreateBlob creates a blob object.
--- a/github/git_blobs_test.go
+++ b/github/git_blobs_test.go
@@ -6,6 +6,7 @@
 package github
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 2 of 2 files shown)


github/git_blobs.go
```diff
@@ -38,6 +39,23 @@ func (s *GitService) GetBlob(ctx context.Context, owner string, repo string, sha
...
 
+// GetBlobRaw fetches a blob's contents from a repo.
+// Unlike GetBlob, it returns the raw bytes rather than the base64-encoded data.
+func (s *GitService) GetBlobRaw(ctx context.Context, owner, repo, sha1 string) ([]byte, error) {
+	u := fmt.Sprintf("repos/%v/%v/git/blobs/%v", owner, repo, sha1)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "application/vnd.github.v3.raw")
```
gmlewis: Do you want to make `"application/vnd.github.v3.raw"` a `const` in `github.go`?
No big deal... either way is fine with me.

dsymonds: I considered it, but it didn't seem necessary for this initial case. If any other code starts using it then that's probably a good enough time to define it as a constant.

shurcooL:
I was hoping that this would follow the same pattern as in the previous PRs #481 and #767, but I see now that the choice of supported media types by the blob endpoints is not the same.

According to https://developer.github.com/v3/media/#commits-commit-comparison-and-pull-requests:

> The Commit, Commit Comparison, and Pull Request resources support diff and patch formats:
>
> - diff
> - patch
> - sha

(It's missing JSON, but it is also supported.)

But the blob endpoint are described at https://developer.github.com/v3/media/#git-blob-properties:

> The following media types are allowed when getting a blob:
>
> - JSON
> - raw

I was going to suggest we have a `opt RawOptions` parameter here like in others, but this complicates things.

The reason to have a parameter like that is in case GitHub adds additional media types. If there's a `opt RawOptions`-like parameter, it's possible to add support without requiring breaking API changes.

Blob endpoints currently supports only 1 non-JSON raw type, so not having an option works today. It feels like this won't change soon, but it's hard to predict with certainty...

If you think the current approach makes sense, I'm okay with this API (no `RawOptions`-like parameter), with the hope it will last for some time without needing to change. But it's not an easy call trying to predict how it'll play out.

github/git_blobs.go
```diff
@@ -38,6 +39,23 @@ func (s *GitService) GetBlob(ctx context.Context, owner string, repo string, sha
...
+
+	var buf bytes.Buffer
+	if _, err = s.client.Do(ctx, req, &buf); err != nil {
```
shurcooL: I would suggest having `_, err = s.client.Do(ctx, req, &buf)` be on a separate line, like this:

```Go
_, err = s.client.Do(ctx, req, &buf)
if err != nil {
    ...
}
```

Making it inline in the `if` statement makes it look less important, and it's not consistent with the style in the hundreds of other endpoints we have. I think it's a good idea to try to keep the code in this library as consistent as possible, and not deviate unnecessarily. Inconsistent style makes it harder for new contributors to figure out which style to use.

github/git_blobs.go
```diff
@@ -38,6 +39,23 @@ func (s *GitService) GetBlob(ctx context.Context, owner string, repo string, sha
 	return blob, resp, err
 }
 
+// GetBlobRaw fetches a blob's contents from a repo.
+// Unlike GetBlob, it returns the raw bytes rather than the base64-encoded data.
```
shurcooL: Please add a GitHub API link. It's the same as the one above:

```
//
// GitHub API docs: https://developer.github.com/v3/git/blobs/#get-a-blob
```

github/git_blobs.go
```diff
@@ -38,6 +39,23 @@ func (s *GitService) GetBlob(ctx context.Context, owner string, repo string, sha
...
+func (s *GitService) GetBlobRaw(ctx context.Context, owner, repo, sha1 string) ([]byte, error) {
```
shurcooL: This method doesn't return `*Response`. Is there a specific reason you left it out?

Unless there's a reason for it, I believe it's needed for consistency and for callers to be able to find out more details about the response.

shurcooL: The endpoint above, and the GitHub API documentation (https://developer.github.com/v3/git/blobs/#get-a-blob) uses `sha` as the name of the parameter (rather than `sha1`). I suspect it's better if we stay consistent.

gmlewis:
I get the impression that GitHub is focused more upon their GraphQL API for the long-term from the way they were talking at their San Francisco get-together last year... (but that certainly doesn't mean this won't change). So I think the current approach makes sense.

shurcooL:
Yes, that's the main reason I'm okay going without an `RawType`-like enum parameter. It's hard to justify asking for it now, when it would have a single `Raw` value, and it feels like it would never get a second value.

github/git_blobs.go
dsymonds: seems like unnecessary consistency, but sure.

github/git_blobs.go
dsymonds: done

github/git_blobs.go
dsymonds: I don't know what one would do with the Response apart from what is in the []byte and error return values, but sure.

dsymonds: done

dsymonds:
I think I addressed everything. PTAL.

github/git_blobs_test.go
```diff
@@ -51,6 +52,28 @@ func TestGitService_GetBlob_invalidOwner(t *testing.T) {
...
+	mux.HandleFunc("/repos/o/r/git/blobs/s", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testHeader(t, r, "Accept", "application/vnd.github.v3.raw")
+
+		fmt.Fprint(w, `raw contents here`)
+	})
+
+	blob, _, err := client.Git.GetBlobRaw(context.Background(), "o", "r", "s")
+	if err != nil {
+		t.Errorf("Git.GetBlob returned error: %v", err)
```
shurcooL: Minor, should be `GetBlobRaw` in the error text.

dsymonds: whoops. fixed.

shurcooL:
Thank you!

dsymonds:
Cool. I'm just glad to upstream some of my custom code.  ;-)

github/git_blobs.go
shurcooL: It might be helpful, for example, to log the response status code from GitHub API in case an error happens.

It's mostly for consistency. All other endpoints return it, so users would expect to be able to access the `*github.Response` struct in this case too.

PR #893, opened 2018-04-13
Add an example for PullRequestsService.Create

aerostitch:
This PR is about providing an example on how to create a pull request (it's always nice to have examples in the godoc).

### DIFF
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -93,5 +93,5 @@ func ExamplePullRequestsService_Create() {
 		return
 	}
 
-	fmt.Printf("PR created: %s\n", *pr.HTMLURL)
+	fmt.Printf("PR created: %s\n", pr.GetHTMLURL())
 }
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -92,5 +92,6 @@ func ExamplePullRequestsService_Create() {
 		fmt.Println(err)
 		return
 	}
+
 	fmt.Printf("PR created: %s\n", *pr.HTMLURL)
 }
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -77,6 +77,12 @@ func ExampleUsersService_ListAll() {
 
 func ExamplePullRequestsService_Create() {
 	// In this example we're creating a PR and displaying the HTML url at the end.
+
+	// Note that authentication is needed here as you are performing a modification
+	// so you will need to modify the example to provide an oauth client to
+	// github.NewClient() instead of nil. See the following documentation for more
+	// information on how to authenticate with the client:
+	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
 	client := github.NewClient(nil)
 
 	newPR := &github.NewPullRequest{
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -29,11 +29,7 @@ func ExampleClient_Markdown() {
 }
 
 func ExampleRepositoriesService_GetReadme() {
-	ctx := context.Background()
-	token := "MyGitHubTokenThatIDontStoreInMyScripts"
-	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
-	tc := oauth2.NewClient(ctx, ts)
-	client := github.NewClient(tc)
+	client := github.NewClient(nil)
 
 	readme, _, err := client.Repositories.GetReadme(context.Background(), "google", "go-github", nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


aerostitch:
I signed it! :smile_cat:

github/examples_test.go
```diff
@@ -74,3 +74,23 @@ func ExampleUsersService_ListAll() {
...
+		Body:                github.String("This is the description of the PR created with the package `github.com/google/go-github/github`"),
+		MaintainerCanModify: github.Bool(true),
+	}
+
+	pr, _, err := client.PullRequests.Create(context.Background(), "myOrganization", "myRepository", newPR)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	fmt.Printf("PR created: %s\n", *pr.HTMLURL)
```
gmlewis: One more tiny nit... I would leave the blank line you had before this one to separate it a bit from the error above... but not a big deal either way... feel free to ignore that.

Also, when contributing to this repo in the future, which we hope you do, you don't need to force push... you can just keep adding commit after commit during the review process (which makes the changes easier for the reviewers to review), and then we squash and merge at the very end into a single commit.

aerostitch: Got it. Thanks a lot!
Empty line added ;)

github/examples_test.go
```diff
@@ -74,3 +74,24 @@ func ExampleUsersService_ListAll() {
...
+		MaintainerCanModify: github.Bool(true),
+	}
+
+	pr, _, err := client.PullRequests.Create(context.Background(), "myOrganization", "myRepository", newPR)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	fmt.Printf("PR created: %s\n", *pr.HTMLURL)
```
gmlewis: Sorry I didn't see this before... I meant to use the accessor here:
`pr.GetHTMLURL()`
instead of `*pr.HTMLURL`

github/examples_test.go
```diff
@@ -74,3 +74,24 @@ func ExampleUsersService_ListAll() {
 		// Process users...
 	}
 }
+
+func ExamplePullRequestsService_Create() {
+	// In this example we're creating a PR and displaying the HTML url at the end.
+	client := github.NewClient(nil)
```
shurcooL: Not dealing with authentication keeps the example shorter and simpler, but I worry it might be misleading.

In reality, this code would never work as is, because creating a PR requires an authenticated user.

I think it'd be good if we added a comment noting that, and perhaps linking to https://godoc.org/github.com/google/go-github/github#hdr-Authentication for more information on how to do authentication.

What do you think?

shurcooL:
Thanks, but I don't think you should make that change. The other examples do not need authentication because they perform read operations on public data. They'll work okay without auth.

Only the new `PullRequest.Create` example you're adding needs authentication.

Additionally, I think all that auth code detracts from the point of the example. I think adding a comment saying that authentication is needed and linking to the documentation at https://godoc.org/github.com/google/go-github/github#hdr-Authentication for more information is a better way to go.

We want to keep the quality of the examples high.

If you'd like to add the example code for doing auth via oauth2 package, we can make that a separate dedicated example covering authentication only. But that should be a separate PR. For this one, please revert these changes and add a comment instead.

github/examples_test.go
```diff
@@ -77,7 +78,11 @@ func ExampleUsersService_ListAll() {
 
 func ExamplePullRequestsService_Create() {
 	// In this example we're creating a PR and displaying the HTML url at the end.
-	client := github.NewClient(nil)
+	ctx := context.Background()
+	token := "MyGitHubTokenThatIDontStoreInMyScripts"
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
```
shurcooL: Here's an idea, in addition to adding the comment saying authentication is needed and linking to its documentation, you can keep this auth code, just factor it out into a helper:

```Go
client := newGitHubClientWithAuth()
```

And below `ExamplePullRequestsService_Create`, you can create that helper:

```
func newGitHubClientWithAuth() *github.Client {
		&oauth2.Token{AccessToken: "... your access token ..."},
	)
	tc := oauth2.NewClient(context.Background(), ts)
	return github.NewClient(tc)
}
```

That way the code is realistic, and the details of authentication don't need to detract from the main point of the `ExamplePullRequestsService_Create` example.

aerostitch: But you can't see the code of `newGitHubClientWithAuth` in the final godoc if you do that right?
So that would be a bit misleading for the user that thinks the `newGitHubClientWithAuth` is an unexported helper from the package right?

aerostitch:
@shurcooL I reverted the auth changes and added a comment about authentication.
Not sure if you saw my comment on the outdated change so here it is:
> But you can't see the code of `newGitHubClientWithAuth` in the final godoc if you do that right?
> So that would be a bit misleading for the user that thinks the newGitHubClientWithAuth is an unexported helper from the package right?

github/examples_test.go
shurcooL: You're right, it shouldn't show up. To see the actual source code of `newGitHubClientWithAuth`, it would require people to jump into the source code.

aerostitch:
Thanks for the review!
Yes, I think that adding `newGitHubClientWithAuth` (while useful in real-life programs) would be more confusing than it would provide help for new users reading the documentation on godoc.

shurcooL:
Thank you for improving the package documentation @aerostitch!

The example is visible at https://godoc.org/github.com/google/go-github/github#example-PullRequestsService-Create now.

PR #894, opened 2018-04-16
Add an example of using the contents API to push a file

aerostitch:
This PR provides an example on how to use `client.Repositories.CreateFile` to push a new file to a repository.

### DIFF
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -73,7 +73,7 @@ func ExampleRepositoriesService_CreateFile() {
 	ctx := context.Background()
 	fileContent := []byte("This is the content of my file\n\nand the 2nd line of it")
 
-	// Note: the files need to be absent from the repository as you are not
+	// Note: the file need to be absent from the repository as you are not
 	// specifying a SHA reference here
 	opts := github.RepositoryContentFileOptions{
 		Message:   github.String("This is my commit message"),
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -59,6 +59,35 @@ func ExampleRepositoriesService_List() {
 	fmt.Printf("Recently updated repositories by %q: %v", user, github.Stringify(repos))
 }
 
+func ExampleRepositoriesService_CreateFile() {
+	// In this example we're creating a new file in a repository using the
+	// Contents API. Only 1 file per commit can be managed through that API
+
+	// Note that authentication is needed here as you are performing a modification
+	// so you will need to modify the example to provide an oauth client to
+	// github.NewClient() instead of nil. See the following documentation for more
+	// information on how to authenticate with the client:
+	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
+	client := github.NewClient(nil)
+
+	ctx := context.Background()
+	fileContent := []byte("This is the content of my file\n\nand the 2nd line of it")
+
+	// Note: the files need to be absent from the repository as you are not
+	// specifying a SHA reference here
+	opts := github.RepositoryContentFileOptions{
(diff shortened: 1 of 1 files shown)


github/examples_test.go
```diff
@@ -73,7 +73,7 @@ func ExampleRepositoriesService_CreateFile() {
 	ctx := context.Background()
 	fileContent := []byte("This is the content of my file\n\nand the 2nd line of it")
 
-	// Note: the files need to be absent from the repository as you are not
+	// Note: the file need to be absent from the repository as you are not
```
gmlewis: How about `the file needs` ?
Also, how about adding a period after `reference here.`?

aerostitch:
Good catch. Thanks @gmlewis ! :)
Sorry for the late reply.
Fixed.

Joseph

github/examples_test.go
```diff
@@ -59,6 +59,35 @@ func ExampleRepositoriesService_List() {
...
+
+	// Note that authentication is needed here as you are performing a modification
+	// so you will need to modify the example to provide an oauth client to
+	// github.NewClient() instead of nil. See the following documentation for more
+	// information on how to authenticate with the client:
+	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
+	client := github.NewClient(nil)
+
+	ctx := context.Background()
+	fileContent := []byte("This is the content of my file\n\nand the 2nd line of it")
```
gmlewis: nit: with two newlines, that is actually the 3rd line of it... so I would simply remove one of the two `\n`.

github/examples_test.go
```diff
@@ -59,6 +59,35 @@ func ExampleRepositoriesService_List() {
 	fmt.Printf("Recently updated repositories by %q: %v", user, github.Stringify(repos))
 }
 
+func ExampleRepositoriesService_CreateFile() {
+	// In this example we're creating a new file in a repository using the
+	// Contents API. Only 1 file per commit can be managed through that API
```
gmlewis: nit: end sentence with a period.

github/examples_test.go
```diff
@@ -59,6 +59,35 @@ func ExampleRepositoriesService_List() {
...
+	// information on how to authenticate with the client:
+	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
+	client := github.NewClient(nil)
+
+	ctx := context.Background()
+	fileContent := []byte("This is the content of my file\n\nand the 2nd line of it")
+
+	// Note: the file needs to be absent from the repository as you are not
+	// specifying a SHA reference here.
+	opts := github.RepositoryContentFileOptions{
```
gmlewis: nit: since `opts` is going to be only used as a pointer anyway, you might as well make it that way up-front.
So this would be:
`opts := &github.RepositoryContentFileOptions{`

github/examples_test.go
```diff
@@ -59,6 +59,35 @@ func ExampleRepositoriesService_List() {
...
+		Message:   github.String("This is my commit message"),
+		Content:   fileContent,
+		Branch:    github.String("master"),
+		Committer: &github.CommitAuthor{Name: github.String("FirstName LastName"), Email: github.String("user@example.com")},
+	}
+	_, _, err := client.Repositories.CreateFile(ctx, "myOrganization", "myRepository", "myNewFile.md", &opts)
```
gmlewis: ...and this would be just `opts` instead of `&opts`.

aerostitch:
Thanks for the feedback @gmlewis. I updated the PR accordingly. :)

aerostitch:
Thanks! :)

PR #896, opened 2018-04-18
Add example to get a team ID from its name using ListTeams

aerostitch:
This PR adds an example on how to use ListTeams to get the ID of a team based on its name.

### DIFF
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -103,13 +103,13 @@ func ExamplePullRequestsService_Create() {
 }
 
 func ExampleOrganizationsService_ListTeams() {
-	// This example shows how to get a team ID corresponding to a given team name
+	// This example shows how to get a team ID corresponding to a given team name.
 
 	// Note that authentication is needed here as you are performing a lookup on
-	// an organization's administrative configurationC, so you will need to modify
+	// an organization's administrative configuration, so you will need to modify
 	// the example to provide an oauth client to github.NewClient() instead of nil.
-	// See the following documentation for more // information on how to
-	// authenticate with the client:
+	// See the following documentation for more information on how to authenticate
+	// with the client:
 	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
 	client := github.NewClient(nil)
 
@@ -135,6 +135,10 @@ func ExampleOrganizationsService_ListTeams() {
 		}
 		opts.Page++
 	}
-	fmt.Printf("Team \"%s\" has ID %d", teamName, teamID)
-	return
+
+	if teamID == -1 {
+		fmt.Printf("Team \"%s\" was not found", teamName)
+	} else {
+		fmt.Printf("Team \"%s\" has ID %d", teamName, teamID)
+	}
 }
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -113,7 +113,6 @@ func ExampleOrganizationsService_ListTeams() {
 	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
 	client := github.NewClient(nil)
 
-	teamID := int64(-1)
 	teamName := "Developpers team"
 	ctx := context.Background()
 	opts := &github.ListOptions{}
@@ -129,16 +128,12 @@ func ExampleOrganizationsService_ListTeams() {
 		}
 		for _, t := range teams {
 			if *t.Name == teamName {
-				teamID = *t.ID
-				break
+				fmt.Printf("Team \"%s\" has ID %d", teamName, *t.ID)
+				return
 			}
 		}
 		opts.Page++
 	}
 
-	if teamID == -1 {
-		fmt.Printf("Team \"%s\" was not found", teamName)
-	} else {
-		fmt.Printf("Team \"%s\" has ID %d", teamName, teamID)
-	}
+	fmt.Printf("Team \"%s\" was not found", teamName)
 }
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -113,18 +113,15 @@ func ExampleOrganizationsService_ListTeams() {
 	// https://godoc.org/github.com/google/go-github/github#hdr-Authentication
 	client := github.NewClient(nil)
 
-	teamName := "Developpers team"
+	teamName := "Developers team"
 	ctx := context.Background()
 	opts := &github.ListOptions{}
 
 	for {
-		teams, _, err := client.Organizations.ListTeams(ctx, "myOrganization", opts)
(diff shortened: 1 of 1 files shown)


github/examples_test.go
```diff
@@ -101,3 +101,40 @@ func ExamplePullRequestsService_Create() {
...
+	ctx := context.Background()
+	opts := &github.ListOptions{}
+
+	for {
+		teams, _, err := client.Organizations.ListTeams(ctx, "myOrganization", opts)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		if len(teams) == 0 {
```
gmlewis: Hmmm... I'm not sure that this is an excellent example for how to paginate because it makes one extra call to the API when it could know up front how many pages are expected by parsing the results of the first call.

See the `Link:` information located here: https://developer.github.com/v3/teams/#list-teams
and here: https://github.com/google/go-github/blob/master/github/github.go#L358-L366

github/examples_test.go
```diff
@@ -101,3 +101,40 @@ func ExamplePullRequestsService_Create() {
...
+		}
+		for _, t := range teams {
+			if *t.Name == teamName {
+				teamID = *t.ID
+				break
+			}
+		}
+		opts.Page++
+	}
+	fmt.Printf("Team \"%s\" has ID %d", teamName, teamID)
```
gmlewis: This prints `... has ID -1` if the team name is not found. It might be nice to print a different message, like the team name was not found.

github/examples_test.go
```diff
@@ -101,3 +101,40 @@ func ExamplePullRequestsService_Create() {
...
+	return
```
gmlewis: This return statement is unnecessary.

github/examples_test.go
```diff
@@ -101,3 +101,40 @@ func ExamplePullRequestsService_Create() {
 
 	fmt.Printf("PR created: %s\n", pr.GetHTMLURL())
 }
+
+func ExampleOrganizationsService_ListTeams() {
+	// This example shows how to get a team ID corresponding to a given team name
```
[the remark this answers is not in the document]
aerostitch: Sorry seems I fixed that while you were already reviewing. That one's fixed.

github/examples_test.go
```diff
@@ -101,3 +101,40 @@ func ExamplePullRequestsService_Create() {
...
+
+	// Note that authentication is needed here as you are performing a lookup on
+	// an organization's administrative configurationC, so you will need to modify
```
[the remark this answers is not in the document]
aerostitch: Sorry seems I fixed that while you were already reviewing. That one's fixed.

github/examples_test.go
```diff
@@ -101,3 +101,40 @@ func ExamplePullRequestsService_Create() {
...
+	// the example to provide an oauth client to github.NewClient() instead of nil.
+	// See the following documentation for more // information on how to
```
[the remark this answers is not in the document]
aerostitch: Sorry seems I fixed that while you were already reviewing. That one's fixed too.

github/examples_test.go
aerostitch: Sorry seems I fixed that while you were already reviewing. That one's fixed.

github/examples_test.go
aerostitch: Sorry seems I fixed that while you were already reviewing. That one's fixed.

gmlewis:
Ah, OK, @aerostitch... thanks for the info.  Typically, we develop PRs locally or in our own forked repos and don't actually create a PR on the main repo until we are happy with all the changes. That way, when a PR comes in, anyone can start reviewing it. No big deal... I just want to help you out with the typical "flow". I'll take another look.

github/examples_test.go
```diff
@@ -101,3 +101,39 @@ func ExamplePullRequestsService_Create() {
...
+	opts := &github.ListOptions{}
+
+	for {
+		teams, resp, err := client.Organizations.ListTeams(ctx, "myOrganization", opts)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		for _, t := range teams {
+			if *t.Name == teamName {
```
gmlewis: In general, I recommend using the provided accessors:
`if t.GetName() == teamName {`

github/examples_test.go
```diff
@@ -101,3 +101,39 @@ func ExamplePullRequestsService_Create() {
...
+				fmt.Printf("Team \"%s\" has ID %d", teamName, *t.ID)
```
gmlewis: Use accessors and take advantage of `%q` to make the code easier to read (also add a newline):

```go
fmt.Printf("Team %q has ID %d\n", teamName, t.GetID())
```

github/examples_test.go
```diff
@@ -101,3 +101,39 @@ func ExamplePullRequestsService_Create() {
...
+				return
+			}
+		}
+		if opts.Page == resp.LastPage {
+			break
+		}
+		opts.Page = resp.NextPage
+	}
+
+	fmt.Printf("Team \"%s\" was not found", teamName)
```
gmlewis:
```go
fmt.Printf("Team %q was not found.\n", teamName)
```

aerostitch:
Thanks for the review @gmlewis 

> Typically, we develop PRs locally or in our own forked repos and don't actually create a PR on the main repo until we are happy with all the changes.

Yes, that's what I usually try to stick to but I got second thoughts after reading the PR again. Sorry for that.

Good catch for the usage of `%q`, I must say I never used it before! :sweat_smile:

I know for the accessors, you told me last time and I forgot again. Sorry for that. :disappointed: 

Anyway, I updated the PR based on your feedback. Thanks! :)
Joseph

aerostitch:
Thanks. New lines added! :)

github/examples_test.go
```diff
@@ -101,3 +101,39 @@ func ExamplePullRequestsService_Create() {
...
+		}
+		for _, t := range teams {
+			if t.GetName() == teamName {
+				fmt.Printf("Team %q has ID %d\n", teamName, t.GetID())
+				return
+			}
+		}
+		if opts.Page == resp.LastPage {
+			break
+		}
```
shurcooL: There are many valid ways of doing pagination and detecting last page. This is one of them.

But I think we should be consistent with the example provided at https://godoc.org/github.com/google/go-github/github#hdr-Pagination, and use the same style here.

aerostitch: That totally makes sense.

aerostitch:
Thanks for the review @shurcooL !
PR updated accordingly! :)

aerostitch:
Thanks! :)

PR #899, opened 2018-04-23
change org id to int64 in OrganizationsListOptions

qinfchen:
Change org ID in OrganizationsListOptions to be consistent with the [org id type](https://github.com/google/go-github/blob/master/github/orgs.go#L23)


qinfchen:
I signed it!

gmlewis:
@googlebot - please rescan.

shurcooL:
Thanks @qinfchen!

Issue #913, opened 2018-05-09
Support the new Checks API (beta)


sridharavinash:
The new Checks API was recently announced in https://developer.github.com/changes/2018-05-07-new-checks-api-public-beta/

While it's currently in beta and supports only GithHubApps, it might be worthwhile adding support.

I'm happy to get an initial PR going for this.

gmlewis:
Here are some recent updates to the Checks API:

GitHub Developer API announcement:
https://developer.github.com/changes/2018-05-23-request-actions-on-checks/

gmlewis:
...and some more recent updates to the Checks API:
https://developer.github.com/changes/2018-07-11-graphql-checks-preview/#changes-to-the-checks-rest-api

PR #914, opened 2018-05-10
Add checks api

sridharavinash:
~~This is a work in progress~~

Adding support for the checks api as mentioned in #913 

Plan on supporting: 

## [CheckRuns API](https://developer.github.com/v3/checks/runs/)
- [x]     Create a check run
- [x]     Update a check run
- [x]     List check runs for a specific ref
- [x]     List check runs in a check suite
- [x]     Get a single check run
- [x]     List annotations for a check run

## [CheckSuite API](https://developer.github.com/v3/checks/suites/)

- [x]    Get a single check suite
- [x]     List check suites for a specific ref
- [x]     Set preferences for check suites on a repository
- [x]     Create a check suite
- [x]     Request check suites

# [Events](https://developer.github.com/v3/activity/events/types/)
- [x] [CheckRunEvent](https://developer.github.com/v3/activity/events/types/#checkrunevent)
- [x] [CheckSuiteEvent](https://developer.github.com/v3/activity/events/types/#checksuiteevent)

### DIFF
--- a/github/checks.go
+++ b/github/checks.go
@@ -96,16 +96,16 @@ func (s *ChecksService) GetCheckRun(ctx context.Context, owner string, repo stri
 
 // CreateCheckRunOptions sets up parameters need to create a CheckRun.
 type CreateCheckRunOptions struct {
-	Name        string          `json:"name"`        // The name of the check (e.g., "code-coverage").(Required.)
-	HeadBranch  string          `json:"head_branch"` // The name of the branch to perform a check against.(Required.)
-	HeadSHA     string          `json:"head_sha"`    // The SHA of the commit.(Required.)
-	DetailsURL  *string         `json:"details_url,omitempty"`
-	ExternalID  *int64          `json:"external_id,omitempty"`
-	Status      *string         `json:"status,omitempty"`
-	Conclusion  *string         `json:"conclusion,omitempty"` // Required if you provide a status of completed.
-	StartedAt   *Timestamp      `json:"started_at,omitempty"`
-	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // Required if you provide conclusion.
-	Output      *CheckRunOutput `json:"output,omitempty"`       // Optionally provide descriptive details about the run.
+	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage").(Required.)
+	HeadBranch  string          `json:"head_branch"`            // The name of the branch to perform a check against.(Required.)
+	HeadSHA     string          `json:"head_sha"`               // The SHA of the commit.(Required.)
+	DetailsURL  *string         `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
+	ExternalID  *int64          `json:"external_id,omitempty"`  // A reference for the run on the integrator's system. (Optional.)
+	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of queued, in_progress, or completed. Default: queued. (Optional.)
+	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of success, failure, neutral, cancelled, timed_out, or action_required.(Optional. Required if you provide a status of completed.)
+	StartedAt   *Timestamp      `json:"started_at,omitempty"`   // The time that the check run began in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.(Optional.)
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run.(Optional)
 }
 
 // CreateCheckRun Creates a check run for repository.
--- a/github/checks.go
+++ b/github/checks.go
@@ -241,7 +241,7 @@ type ListCheckRunsResults struct {
 
 // ListCheckRunsForRef lists check runs for a specific ref.
 //
-// GitHub API docs: https://developer.github.com/v3/checks/runs/list-check-runs-for-a-specific-ref
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#list-check-runs-for-a-specific-ref
 func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner, repo, ref string, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-runs", owner, repo, ref)
 	u, err := addOptions(u, opt)
@@ -294,7 +294,7 @@ func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner, repo
 // ListCheckSuiteOptions represents parameters to list check suites.
 type ListCheckSuiteOptions struct {
 	CheckName *string `url:"check_name,omitempty"` // Filters checks suites by the name of the check run.
-	Filter    *string `url:"filter,omitempty"`     // Filters check suites by GitHub App id.
+	AppID     *int    `url:"app_id,omitempty"`     // Filters check suites by GitHub App id.
 
 	ListOptions
 }
@@ -350,7 +350,7 @@ type CheckSuitePreferenceResults struct {
 
 // PreferenceList represents a list of auto trigger checks for  repository
 type PreferenceList struct {
-	AutoTriggerChecks []AutoTriggerCheck `json:"auto_trigger_checks,omitempty"` // A slice of auto trigger checks that can be set for a check suite in a repository.
+	AutoTriggerChecks []*AutoTriggerCheck `json:"auto_trigger_checks,omitempty"` // A slice of auto trigger checks that can be set for a check suite in a repository.
 }
 
 // SetCheckSuitePreferences Changes the default automatic flow when creating check suites.
--- a/github/checks.go
+++ b/github/checks.go
@@ -90,7 +90,7 @@ func (c CheckSuite) String() string {
 // GetCheckRun gets a check-run for a repository.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/#get-a-single-check-run
-func (s *ChecksService) GetCheckRun(ctx context.Context, owner string, repo string, checkRunID int64) (*CheckRun, *Response, error) {
+func (s *ChecksService) GetCheckRun(ctx context.Context, owner, repo string, checkRunID int64) (*CheckRun, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/check-runs/%v", owner, repo, checkRunID)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
@@ -108,10 +108,10 @@ func (s *ChecksService) GetCheckRun(ctx context.Context, owner string, repo stri
 	return checkRun, resp, nil
 }
 
-// GetCheckSuite Gets a single check suite.
+// GetCheckSuite gets a single check suite.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/suites/#get-a-single-check-suite
-func (s *ChecksService) GetCheckSuite(ctx context.Context, owner string, repo string, checkSuiteID int64) (*CheckSuite, *Response, error) {
+func (s *ChecksService) GetCheckSuite(ctx context.Context, owner, repo string, checkSuiteID int64) (*CheckSuite, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/check-suites/%v", owner, repo, checkSuiteID)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
@@ -131,22 +131,22 @@ func (s *ChecksService) GetCheckSuite(ctx context.Context, owner string, repo st
 
 // CreateCheckRunOptions sets up parameters needed to create a CheckRun.
 type CreateCheckRunOptions struct {
-	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage").(Required.)
-	HeadBranch  string          `json:"head_branch"`            // The name of the branch to perform a check against.(Required.)
-	HeadSHA     string          `json:"head_sha"`               // The SHA of the commit.(Required.)
+	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage"). (Required.)
+	HeadBranch  string          `json:"head_branch"`            // The name of the branch to perform a check against. (Required.)
+	HeadSHA     string          `json:"head_sha"`               // The SHA of the commit. (Required.)
 	DetailsURL  *string         `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
 	ExternalID  *string         `json:"external_id,omitempty"`  // A reference for the run on the integrator's system. (Optional.)
-	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of queued, in_progress, or completed. Default: queued. (Optional.)
-	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of success, failure, neutral, cancelled, timed_out, or action_required.(Optional. Required if you provide a status of completed.)
-	StartedAt   *Timestamp      `json:"started_at,omitempty"`   // The time that the check run began in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.(Optional.)
-	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. (Optional. Required if you provide conclusion.)
-	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run.(Optional)
+	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of "queued", "in_progress", or "completed". Default: "queued". (Optional.)
+	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of "success", "failure", "neutral", "cancelled", "timed_out", or "action_required". (Optional. Required if you provide a status of "completed".)
+	StartedAt   *Timestamp      `json:"started_at,omitempty"`   // The time that the check run began. (Optional.)
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run. (Optional)
 }
 
-// CreateCheckRun Creates a check run for repository.
+// CreateCheckRun creates a check run for repository.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/#create-a-check-run
-func (s *ChecksService) CreateCheckRun(ctx context.Context, owner string, repo string, opt CreateCheckRunOptions) (*CheckRun, *Response, error) {
+func (s *ChecksService) CreateCheckRun(ctx context.Context, owner, repo string, opt CreateCheckRunOptions) (*CheckRun, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/check-runs", owner, repo)
 	req, err := s.client.NewRequest("POST", u, opt)
 	if err != nil {
@@ -166,21 +166,21 @@ func (s *ChecksService) CreateCheckRun(ctx context.Context, owner string, repo s
 
 // UpdateCheckRunOptions sets up parameters needed to update a CheckRun.
 type UpdateCheckRunOptions struct {
-	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage").(Required.)
-	HeadBranch  *string         `json:"head_branch"`            // The name of the branch to perform a check against.(Optional.)
-	HeadSHA     *string         `json:"head_sha"`               // The SHA of the commit.(Optional.)
+	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage"). (Required.)
+	HeadBranch  *string         `json:"head_branch,omitempty"`  // The name of the branch to perform a check against. (Optional.)
+	HeadSHA     *string         `json:"head_sha,omitempty"`     // The SHA of the commit. (Optional.)
 	DetailsURL  *string         `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
 	ExternalID  *string         `json:"external_id,omitempty"`  // A reference for the run on the integrator's system. (Optional.)
-	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of queued, in_progress, or completed. Default: queued. (Optional.)
-	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of success, failure, neutral, cancelled, timed_out, or action_required.(Optional. Required if you provide a status of completed.)
-	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. (Optional. Required if you provide conclusion.)
-	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run.(Optional)
+	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of "queued", "in_progress", or "completed". Default: "queued". (Optional.)
+	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of "success", "failure", "neutral", "cancelled", "timed_out", or "action_required". (Optional. Required if you provide a status of "completed".)
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run. (Optional)
 }
 
-// UpdateCheckRun Updates a check run for a specific commit in a repository.
+// UpdateCheckRun updates a check run for a specific commit in a repository.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/#update-a-check-run
-func (s *ChecksService) UpdateCheckRun(ctx context.Context, owner string, repo string, checkRunID int64, opt UpdateCheckRunOptions) (*CheckRun, *Response, error) {
+func (s *ChecksService) UpdateCheckRun(ctx context.Context, owner, repo string, checkRunID int64, opt UpdateCheckRunOptions) (*CheckRun, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/check-runs/%v", owner, repo, checkRunID)
 	req, err := s.client.NewRequest("PATCH", u, opt)
 	if err != nil {
@@ -198,10 +198,10 @@ func (s *ChecksService) UpdateCheckRun(ctx context.Context, owner string, repo s
 	return checkRun, resp, nil
 }
 
-// ListCheckRunAnnotations List the annotations for a check run.
+// ListCheckRunAnnotations lists the annotations for a check run.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/#list-annotations-for-a-check-run
-func (s *ChecksService) ListCheckRunAnnotations(ctx context.Context, owner string, repo string, checkRunID int64, opt *ListOptions) ([]*CheckRunAnnotation, *Response, error) {
+func (s *ChecksService) ListCheckRunAnnotations(ctx context.Context, owner, repo string, checkRunID int64, opt *ListOptions) ([]*CheckRunAnnotation, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/check-runs/%v/annotations", owner, repo, checkRunID)
 	u, err := addOptions(u, opt)
 	if err != nil {
@@ -224,25 +224,25 @@ func (s *ChecksService) ListCheckRunAnnotations(ctx context.Context, owner strin
 	return checkRunAnnotations, resp, nil
 }
 
-// ListCheckRunsOptions Optional parameters to list check runs.
+// ListCheckRunsOptions represents parameters to list check runs.
 type ListCheckRunsOptions struct {
-	CheckName *string `url:"check_name,omitempty"` //Returns check runs with the specified name.
-	Status    *string `url:"status,omitempty"`     //Returns check runs with the specified status. Can be one of queued, in_progress, or completed.
-	Filter    *string `url:"filter,omitempty"`     //Filters check runs by their completed_at timestamp. Can be one of latest (returning the most recent check runs) or all. Default: latest
+	CheckName *string `url:"check_name,omitempty"` // Returns check runs with the specified name.
+	Status    *string `url:"status,omitempty"`     // Returns check runs with the specified status. Can be one of "queued", "in_progress", or "completed".
+	Filter    *string `url:"filter,omitempty"`     // Filters check runs by their completed_at timestamp. Can be one of "latest" (returning the most recent check runs) or "all". Default: "latest"
 
 	ListOptions
 }
 
 // ListCheckRunsResults represents the result of a check run list.
 type ListCheckRunsResults struct {
-	Total     *int       `json:"total_count,omitempty"`
-	CheckRuns []CheckRun `json:"check_runs,omitempty"`
+	Total     *int        `json:"total_count,omitempty"`
+	CheckRuns []*CheckRun `json:"check_runs,omitempty"`
 }
 
-// ListCheckRunsForRef List check runs for a specific ref.
+// ListCheckRunsForRef lists check runs for a specific ref.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/list-check-runs-for-a-specific-ref
-func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner string, repo string, ref string, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
+func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner, repo string, ref string, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-runs", owner, repo, ref)
 	u, err := addOptions(u, opt)
 	if err != nil {
@@ -265,10 +265,10 @@ func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner string, r
 	return checkRunResults, resp, nil
 }
 
-// ListCheckRunsCheckSuite Lists check runs for a check suite.
+// ListCheckRunsCheckSuite lists check runs for a check suite.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/#list-check-runs-in-a-check-suite
-func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner string, repo string, checkSuiteID int64, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
+func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner, repo string, checkSuiteID int64, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/check-suites/%v/check-runs", owner, repo, checkSuiteID)
 	u, err := addOptions(u, opt)
 	if err != nil {
@@ -291,24 +291,24 @@ func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner strin
 	return checkRunResults, resp, nil
 }
 
-// ListCheckSuiteOptions Optional parameters to list check suites.
+// ListCheckSuiteOptions represents  parameters to list check suites.
 type ListCheckSuiteOptions struct {
-	CheckName *string `url:"check_name,omitempty"` //Filters checks suites by the name of the check run``.
-	Filter    *string `url:"filter,omitempty"`     //Filters check suites by GitHub App id.
+	CheckName *string `url:"check_name,omitempty"` // Filters checks suites by the name of the check run.
+	Filter    *string `url:"filter,omitempty"`     // Filters check suites by GitHub App id.
 
 	ListOptions
 }
 
 // ListCheckSuiteResults represents the result of a check run list.
 type ListCheckSuiteResults struct {
-	Total       *int         `json:"total_count,omitempty"`
-	CheckSuites []CheckSuite `json:"check_suites,omitempty"`
+	Total       *int          `json:"total_count,omitempty"`
+	CheckSuites []*CheckSuite `json:"check_suites,omitempty"`
 }
 
-// ListCheckSuitesForRef List check suite for a specific ref.
+// ListCheckSuitesForRef lists check suite for a specific ref.
 //
 // GitHub API docs: https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref
-func (s *ChecksService) ListCheckSuitesForRef(ctx context.Context, owner string, repo string, ref string, opt *ListCheckSuiteOptions) (*ListCheckSuiteResults, *Response, error) {
+func (s *ChecksService) ListCheckSuitesForRef(ctx context.Context, owner, repo string, ref string, opt *ListCheckSuiteOptions) (*ListCheckSuiteResults, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-suites", owner, repo, ref)
 	u, err := addOptions(u, opt)
 	if err != nil {
@@ -331,15 +331,15 @@ func (s *ChecksService) ListCheckSuitesForRef(ctx context.Context, owner string,
 	return checkSuiteResults, resp, nil
 }
 
-// AutoTriggerCheck Enables or disables automatic creation of CheckSuite events upon pushes to the repository.
+// AutoTriggerCheck enables or disables automatic creation of CheckSuite events upon pushes to the repository.
 type AutoTriggerCheck struct {
 	AppID   *int64 `json:"app_id,omitempty"`  // The id of the GitHub App. (Required.)
-	Setting *bool  `json:"setting,omitempty"` // Set to true to enable automatic creation of CheckSuite events upon pushes to the repository, or false to disable them. Default: true (Required.)
+	Setting *bool  `json:"setting,omitempty"` // Set to "true" to enable automatic creation of CheckSuite events upon pushes to the repository, or "false" to disable them. Default: "true" (Required.)
 }
 
 // CheckSuitePreferenceOptions Set options for check suite preferences for a repository.
 type CheckSuitePreferenceOptions struct {
-	PreferenceList *PreferenceList `json:"auto_trigger_checks,omitempty"` // An array of auto trigger checks that can be set for a check suite in a repository.
+	PreferenceList *PreferenceList `json:"auto_trigger_checks,omitempty"` // A list of auto trigger checks that can be set for a check suite in a repository.
 }
 
 // CheckSuitePreferenceResults Represents the results of the preference set operation.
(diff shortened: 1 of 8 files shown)


github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+type CheckRun struct {
+	ID           *int64          `json:"id,omitempty"`
+	HeadSHA      *string         `json:"head_sha,omitempty"`
+	ExternalID   *int64          `json:"external_id,omitempty"`
+	URL          *string         `json:"url,omitempty"`
+	HTMLURL      *string         `json:"html_url,omitempty"`
+	Status       *string         `json:"status,omitempty"`
+	Conclusion   *string         `json:"conclusion,omitempty"`
+	StartedAt    *time.Time      `json:"started_at,omitempty"`
+	CompletedAt  *time.Time      `json:"completed_at,omitempty"`
```
shurcooL: I believe we want to use our `Timestamp` type instead of `time.Time` for consistency. Look at other structs, e.g.:

https://github.com/google/go-github/blob/8ea2e2657df890db8fb434a9274799d641bd698c/github/repos.go#L31-L33

(I noticed we do use `time.Time` sometimes, but that looks like an unintentional issue to me.)

github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+// GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/
+type ChecksService service
+
+// CheckRun represents a GitHub check run on a repository associated with a GitHub app
+type CheckRun struct {
+	ID           *int64          `json:"id,omitempty"`
+	HeadSHA      *string         `json:"head_sha,omitempty"`
+	ExternalID   *int64          `json:"external_id,omitempty"`
```
shurcooL: Using `int64` for ID fields is good.

github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+	App          *App            `json:"app,omitempty"`
+	PullRequests []*PullRequest  `json:"pull_requests,omitempty"`
+}
+
+// CheckRunOutput represents the output of a CheckRun
+type CheckRunOutput struct {
+	Title            *string `json:"title,omitempty"`
+	Summary          *string `json:"summary,omitempty"`
+	Text             *string `json:"text,omitempty"`
+	AnnotationsCount *int64  `json:"annotations_count,omitempty"`
```
shurcooL: But all other integer fields should be `int`, not `int64`. Look at other structs for reference, e.g.:

https://github.com/google/go-github/blob/8ea2e2657df890db8fb434a9274799d641bd698c/github/repos.go#L42-L48

github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+// CheckSuite represents a suite of check runs
+type CheckSuite struct {
+	ID *int64 `json:"id,omitempty"`
+}
+
+func (c CheckRun) String() string {
+	return Stringify(c)
+}
+
+// GetCheckRun gets a check-run for a repository
```
shurcooL: Go style suggests having proper grammar, so sentences should end with a period.

github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#get-a-single-check-run
+func (s *ChecksService) GetCheckRun(ctx context.Context, owner string, repo string, id int64) (*CheckRun, *Response, error) {
```
shurcooL: This is okay, but I'm tempted to suggest using `checkRunID` instead of `id` as the last parameter name, for clarity.

github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+	"time"
+)
+
+// ChecksService provides access to the the Checks API in the
+// GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/
+type ChecksService service
+
+// CheckRun represents a GitHub check run on a repository associated with a GitHub app
```
shurcooL: Period at end of sentence here as well, please.

github/checks.go
```diff
@@ -0,0 +1,76 @@
...
+
+	acceptHeaders := []string{mediaTypeCheckRunsPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
+
+	checkRun := new(CheckRun)
+	resp, err := s.client.Do(ctx, req, checkRun)
+	if err != nil {
+		return nil, resp, err
+	}
+	return checkRun, resp, nil
```
shurcooL: To make the code more consistent with everywhere else, I think there should be a blank line above the `return` line. E.g.:

https://github.com/google/go-github/blob/8ea2e2657df890db8fb434a9274799d641bd698c/github/repos.go#L252-L258

github/github.go
```diff
@@ -311,7 +316,6 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
 			return nil, err
 		}
 	}
-
```
shurcooL: Was this line removed intentionally?

It doesn't look like an improvement, so unless you have a reason, I'd suggest reverting it.

github/checks.go
sridharavinash: 👍 . Thanks I'll make that change.

github/checks.go
sridharavinash: makes sense 👍

github/github.go
sridharavinash: yup, stray line kill . Thanks for catching that!

github/checks.go
sridharavinash: 👍

github/checks.go
```diff
@@ -0,0 +1,133 @@
...
+
+package github
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+// ChecksService provides access to the the Checks API in the
```
shurcooL: s/the the/the

github/checks.go
```diff
@@ -0,0 +1,133 @@
...
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#get-a-single-check-run
+func (s *ChecksService) GetCheckRun(ctx context.Context, owner string, repo string, checkRunID int64) (*CheckRun, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/check-runs/%v", owner, repo, checkRunID)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	acceptHeaders := []string{mediaTypeCheckRunsPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
```
shurcooL: There's only one custom Accept header right now, so you can simplify this to just:

```Go
req.Header.Set("Accept", mediaTypeCheckRunsPreview)
```

When/if there's a need for more accept headers, this form can be used at that time.

github/checks.go
```diff
@@ -0,0 +1,133 @@
...
+	Conclusion  *string         `json:"conclusion,omitempty"`
+	StartedAt   *Timestamp      `json:"started_at,omitempty"`
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"`
+	Output      *CheckRunOutput `json:"output,omitempty"`
+}
+
+// CreateCheckRun Creates a check run for repository.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#create-a-check-run
+func (s *ChecksService) CreateCheckRun(ctx context.Context, owner string, repo string, opt *CreateCheckRunOptions) (*CheckRun, *Response, error) {
```
shurcooL: According to https://developer.github.com/v3/checks/runs/#create-a-check-run, some of the parameters are required. They're not all optional. So, let's use a value `opt CreateCheckRunOptions ` rather than pointer `opt * CreateCheckRunOptions` here, to avoid having to deal with the bad input `nil`.

github/checks.go
```diff
@@ -0,0 +1,133 @@
...
+	HeadBranch  *string         `json:"head_branch,omitempty"`
+	HeadSHA     *string         `json:"head_sha,omitempty"`
+	DetailsURL  *string         `json:"details_url,omitempty"`
+	ExternalID  *int64          `json:"external_id,omitempty"`
+	Status      *string         `json:"status,omitempty"`
+	Conclusion  *string         `json:"conclusion,omitempty"`
+	StartedAt   *Timestamp      `json:"started_at,omitempty"`
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"`
+	Output      *CheckRunOutput `json:"output,omitempty"`
+}
```
shurcooL: And this struct should be modified so the required parameters are values, not pointers. They also don't need `,omitempty` option. So:

```Go
// CreateCheckRunOptions sets up parameters need to create a CheckRun.
type CreateCheckRunOptions struct {
	Name        string  `json:"name"`                  // The name of the check (e.g., "code-coverage"). (Required.)
	HeadBranch  string  `json:"head_branch"`           // The name of the branch to perform a check against. (Required.)
	HeadSHA     string  `json:"head_sha"`              // The SHA of the commit. (Required.)
	DetailsURL  *string `json:"details_url,omitempty"` // The URL of the integrator's site that has the full details of the check. (Optional.)
	ExternalID  *int64  `json:"external_id,omitempty"` // A reference for the run on the integrator's system. (Optional.)
	...
}
```

Search the codebase for "(Required.)" or "(Optional.)" to see examples of this pattern. E.g.:

https://github.com/google/go-github/blob/8ea2e2657df890db8fb434a9274799d641bd698c/github/orgs_teams.go#L116-L122

sridharavinash: @shurcooL Thanks for the extra 👀 and pointers.  Any recommended pattern to follow when a `required` field is dependent on another field. For example the `conclusion` field has
> Required if you provide a status of completed.

```

conclusion | string | Required. The final conclusion of the check. Can be one of success, failure, neutral, cancelled, timed_out, or action_required. When the conclusion is action_required, additional details should be provided on the site specified by details_url. Required if you provide a status of completed.
```

The same for the `completed_at` field
>Required if you provide conclusion.

Ideally these would be an `omitempty` by default but shouldn't be when certain other fields are present.

shurcooL: > Any recommended pattern to follow when a required field is dependent on another field.

Yes, we have no choice but to make them a pointer with `,omitempty`, and just rely on documentation.

See here for precedent:

https://github.com/google/go-github/blob/8ea2e2657df890db8fb434a9274799d641bd698c/github/projects.go#L67-L68

(It doesn't use a pointer... but that entire struct uses values for optional fields and expects them to be zero value when unset. Unfortunate that it's not consistent.)

sridharavinash: Thanks for that. I updated the struct to include required/optional as comments along side the field.

github/checks.go
```diff
@@ -0,0 +1,156 @@
...
+// GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/
+type ChecksService service
+
+// CheckRun represents a GitHub check run on a repository associated with a GitHub app.
+type CheckRun struct {
+	ID           *int64          `json:"id,omitempty"`
+	HeadSHA      *string         `json:"head_sha,omitempty"`
+	ExternalID   *int64          `json:"external_id,omitempty"`
```
haya14busa: ExternalID should be "string". 

https://developer.github.com/v3/checks/runs/#parameters

sridharavinash: 👍

sridharavinash:
I think this is now complete and implements all the API endpoints including handling of 2 additional event types structs for the `CheckRun` and `CheckSuite`.

sridharavinash:
friendly bump for review/merge.  🙇

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+func (c CheckSuite) String() string {
+	return Stringify(c)
+}
+
+// GetCheckRun gets a check-run for a repository.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#get-a-single-check-run
+func (s *ChecksService) GetCheckRun(ctx context.Context, owner string, repo string, checkRunID int64) (*CheckRun, *Response, error) {
```
gmlewis: nit: `owner, repo string` instead of `owner string, repo string`.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil, resp, err
+	}
+
+	return checkRun, resp, nil
+}
+
+// GetCheckSuite Gets a single check suite.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/suites/#get-a-single-check-suite
+func (s *ChecksService) GetCheckSuite(ctx context.Context, owner string, repo string, checkSuiteID int64) (*CheckSuite, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkSuite, resp, nil
+}
+
+// CreateCheckRunOptions sets up parameters needed to create a CheckRun.
+type CreateCheckRunOptions struct {
+	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage").(Required.)
```
gmlewis: nit: add a space between `). (`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	HeadBranch  string          `json:"head_branch"`            // The name of the branch to perform a check against.(Required.)
```
gmlewis: `against. (Required.)` (Add a space after the period.)

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	HeadSHA     string          `json:"head_sha"`               // The SHA of the commit.(Required.)
```
gmlewis: Another space after the period.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	DetailsURL  *string         `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
+	ExternalID  *string         `json:"external_id,omitempty"`  // A reference for the run on the integrator's system. (Optional.)
+	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of queued, in_progress, or completed. Default: queued. (Optional.)
+	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of success, failure, neutral, cancelled, timed_out, or action_required.(Optional. Required if you provide a status of completed.)
```
gmlewis: Add space after period. Also, it might be nice to quote each of the possible values.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	StartedAt   *Timestamp      `json:"started_at,omitempty"`   // The time that the check run began in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.(Optional.)
```
gmlewis: Add space after period.  Note that the comment about format doesn't apply here since the JSON decoder handles the conversion to a `Timestamp` for us.
Here and below too.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run.(Optional)
```
gmlewis: Add space after period.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// CreateCheckRun Creates a check run for repository.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#create-a-check-run
+func (s *ChecksService) CreateCheckRun(ctx context.Context, owner string, repo string, opt CreateCheckRunOptions) (*CheckRun, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	resp, err := s.client.Do(ctx, req, checkRun)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRun, resp, nil
+}
+
+// UpdateCheckRunOptions sets up parameters needed to update a CheckRun.
+type UpdateCheckRunOptions struct {
```
gmlewis: Add space after period. I'm going to stop commenting on these now, but below as well.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage").(Required.)
+	HeadBranch  *string         `json:"head_branch"`            // The name of the branch to perform a check against.(Optional.)
```
gmlewis: Add `omitempty` to optional parameter.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	HeadSHA     *string         `json:"head_sha"`               // The SHA of the commit.(Optional.)
```
gmlewis: Add `omitempty`.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	DetailsURL  *string         `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
+	ExternalID  *string         `json:"external_id,omitempty"`  // A reference for the run on the integrator's system. (Optional.)
+	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of queued, in_progress, or completed. Default: queued. (Optional.)
+	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of success, failure, neutral, cancelled, timed_out, or action_required.(Optional. Required if you provide a status of completed.)
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run.(Optional)
```
gmlewis: This `struct` looks almost identical to `CreateCheckRunOptions` above... can they be shared?
I suppose the trade-off is that the documentation might need more clarification.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// UpdateCheckRun Updates a check run for a specific commit in a repository.
```
gmlewis: s/Updates/updates/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#update-a-check-run
+func (s *ChecksService) UpdateCheckRun(ctx context.Context, owner string, repo string, checkRunID int64, opt UpdateCheckRunOptions) (*CheckRun, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil, resp, err
+	}
+
+	return checkRun, resp, nil
+}
+
+// ListCheckRunAnnotations List the annotations for a check run.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#list-annotations-for-a-check-run
+func (s *ChecksService) ListCheckRunAnnotations(ctx context.Context, owner string, repo string, checkRunID int64, opt *ListOptions) ([]*CheckRunAnnotation, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	DetailsURL  *string         `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
+	ExternalID  *string         `json:"external_id,omitempty"`  // A reference for the run on the integrator's system. (Optional.)
+	Status      *string         `json:"status,omitempty"`       // The current status. Can be one of queued, in_progress, or completed. Default: queued. (Optional.)
+	Conclusion  *string         `json:"conclusion,omitempty"`   // Can be one of success, failure, neutral, cancelled, timed_out, or action_required.(Optional. Required if you provide a status of completed.)
+	StartedAt   *Timestamp      `json:"started_at,omitempty"`   // The time that the check run began in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.(Optional.)
+	CompletedAt *Timestamp      `json:"completed_at,omitempty"` // The time the check completed in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput `json:"output,omitempty"`       // Provide descriptive details about the run.(Optional)
+}
+
+// CreateCheckRun Creates a check run for repository.
```
gmlewis: s/Creates/creates/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	checkRun := new(CheckRun)
+	resp, err := s.client.Do(ctx, req, checkRun)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRun, resp, nil
+}
+
+// GetCheckSuite Gets a single check suite.
```
gmlewis: s/Gets/gets/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	var checkRunAnnotations []*CheckRunAnnotation
+	resp, err := s.client.Do(ctx, req, &checkRunAnnotations)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRunAnnotations, resp, nil
+}
+
+// ListCheckRunsOptions Optional parameters to list check runs.
```
gmlewis: Add a verb for the godocs... something like:
`// ListCheckRunsOptions represents optional parameters to list check runs.`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+type ListCheckRunsOptions struct {
+	CheckName *string `url:"check_name,omitempty"` //Returns check runs with the specified name.
```
gmlewis: nit: add space after `//` before first letter... `// Returns...`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	Status    *string `url:"status,omitempty"`     //Returns check runs with the specified status. Can be one of queued, in_progress, or completed.
+	Filter    *string `url:"filter,omitempty"`     //Filters check runs by their completed_at timestamp. Can be one of latest (returning the most recent check runs) or all. Default: latest
+
+	ListOptions
+}
+
+// ListCheckRunsResults represents the result of a check run list.
+type ListCheckRunsResults struct {
+	Total     *int       `json:"total_count,omitempty"`
+	CheckRuns []CheckRun `json:"check_runs,omitempty"`
```
gmlewis: `[]*CheckRun`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// ListCheckRunsForRef List check runs for a specific ref.
```
gmlewis: s/List/lists/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/list-check-runs-for-a-specific-ref
+func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner string, repo string, ref string, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
```
gmlewis: s/owner, repo string/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	var checkRunResults *ListCheckRunsResults
+	resp, err := s.client.Do(ctx, req, &checkRunResults)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRunResults, resp, nil
+}
+
+// ListCheckRunsCheckSuite Lists check runs for a check suite.
```
gmlewis: s/Lists/lists/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/#list-check-runs-in-a-check-suite
+func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner string, repo string, checkSuiteID int64, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	var checkRunResults *ListCheckRunsResults
+	resp, err := s.client.Do(ctx, req, &checkRunResults)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRunResults, resp, nil
+}
+
+// ListCheckSuiteOptions Optional parameters to list check suites.
```
gmlewis: s/Optional/represents optional/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+type ListCheckSuiteOptions struct {
+	CheckName *string `url:"check_name,omitempty"` //Filters checks suites by the name of the check run``.
```
gmlewis: Space after `//` here and elsewhere.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	Filter    *string `url:"filter,omitempty"`     //Filters check suites by GitHub App id.
+
+	ListOptions
+}
+
+// ListCheckSuiteResults represents the result of a check run list.
+type ListCheckSuiteResults struct {
+	Total       *int         `json:"total_count,omitempty"`
+	CheckSuites []CheckSuite `json:"check_suites,omitempty"`
```
gmlewis: `[]*CheckSuite`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// ListCheckSuitesForRef List check suite for a specific ref.
```
gmlewis: s/List/lists/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref
+func (s *ChecksService) ListCheckSuitesForRef(ctx context.Context, owner string, repo string, ref string, opt *ListCheckSuiteOptions) (*ListCheckSuiteResults, *Response, error) {
```
gmlewis: `owner, repo, ref string,`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	var checkSuiteResults *ListCheckSuiteResults
+	resp, err := s.client.Do(ctx, req, &checkSuiteResults)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkSuiteResults, resp, nil
+}
+
+// AutoTriggerCheck Enables or disables automatic creation of CheckSuite events upon pushes to the repository.
```
gmlewis: s/Enables/enables/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+type AutoTriggerCheck struct {
+	AppID   *int64 `json:"app_id,omitempty"`  // The id of the GitHub App. (Required.)
```
gmlewis:
```go
AppID   int64 `json:"app_id"`
```

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	Setting *bool  `json:"setting,omitempty"` // Set to true to enable automatic creation of CheckSuite events upon pushes to the repository, or false to disable them. Default: true (Required.)
```
gmlewis:
```go
Setting bool  `json:"setting"`
```

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// CheckSuitePreferenceOptions Set options for check suite preferences for a repository.
```
gmlewis: s/Set/sets/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+type CheckSuitePreferenceOptions struct {
+	PreferenceList *PreferenceList `json:"auto_trigger_checks,omitempty"` // An array of auto trigger checks that can be set for a check suite in a repository.
+}
+
+// CheckSuitePreferenceResults Represents the results of the preference set operation.
```
gmlewis: s/Represents/represents/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+type CheckSuitePreferenceResults struct {
+	Preferences *PreferenceList `json:"preferences,omitempty"`
+	Repository  *Repository     `json:"repository,omitempty"`
+}
+
+// PreferenceList Represents a list of auto trigger checks for  repository
+type PreferenceList struct {
+	AutoTriggerChecks []AutoTriggerCheck `json:"auto_trigger_checks,omitempty"` // An array of auto trigger checks that can be set for a check suite in a repository.
```
gmlewis: `[]*AutoTriggerCheck`

Also, maybe change `array` to `slice` to fit Go nomenclature better (here and elsewhere).

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// SetCheckSuitePreferences Changes the default automatic flow when creating check suites.
```
gmlewis: s/Changes/changes/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/suites/#set-preferences-for-check-suites-on-a-repository
+func (s *ChecksService) SetCheckSuitePreferences(ctx context.Context, owner string, repo string, opt CheckSuitePreferenceOptions) (*CheckSuitePreferenceResults, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	var checkSuitePrefResults *CheckSuitePreferenceResults
+	resp, err := s.client.Do(ctx, req, &checkSuitePrefResults)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkSuitePrefResults, resp, nil
+}
+
+// CreateCheckSuiteOptions  sets up parameters to manually create a check suites
```
gmlewis: nix double space and add period to end of sentence.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+type CreateCheckSuiteOptions struct {
+	HeadSHA    string  `json:"head_sha"`              //The sha of the head commit. (Required.)
+	HeadBranch *string `json:"head_branch,omitempty"` //The name of the head branch where the code changes are implemented.
+}
+
+// CreateCheckSuite Manually create a check suite for repository.
```
gmlewis: s/Manually create/manually creates/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/suites/#create-a-check-suite
+func (s *ChecksService) CreateCheckSuite(ctx context.Context, owner string, repo string, opt CreateCheckSuiteOptions) (*CheckSuite, *Response, error) {
```
gmlewis: `owner, repo string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkSuite, resp, nil
+}
+
+// RequestCheckSuiteOptions sets up the parameters for a request check suite endpoint.
+type RequestCheckSuiteOptions struct {
+	HeadSHA string `url:"head_sha"` // The sha of the head commit.(Required.)
```
gmlewis: s/url/json/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+// RequestCheckSuite Triggers GitHub to create a new check suite, without pushing new code to a repository.
```
gmlewis: s/Triggers/triggers/

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/checks/suites/#request-check-suites
+func (s *ChecksService) RequestCheckSuite(ctx context.Context, owner string, repo string, opt RequestCheckSuiteOptions) (*Response, error) {
```
gmlewis: `owner, repo string`

github/event_types.go
```diff
@@ -7,6 +7,38 @@
 
 package github
 
+// CheckRunEvent is triggered when a check run is created, updated, or rerequested.
```
gmlewis: Maybe `re-requested` ?

github/event_types.go
```diff
@@ -7,6 +7,38 @@
...
+// The Webhook event name is "check_run".
+//
+// GitHub API docs: https://developer.github.com/v3/activity/events/types/#checkrunevent
+type CheckRunEvent struct {
+	CheckRun *CheckRun `json:"check_run,omitempty"`
+	//The action performed. Can be created, updated or rerequested.
```
gmlewis: `// The` (add space.)
Also maybe quote the options.

github/checks.go
sridharavinash: Yeah I chose to keep them separate so as to keep the documentation clearer. I noticed there were a few other places where the options were backed by a dedicated struct even if they shared similar fields. But I don't have any strong opinions either way. Documenting a shared struct could introduce some verbose comments.

sridharavinash:
@gmlewis Thanks for the extra 👀 . Hopefully I got all them 🤞. Also noticed that the `Request Check Suite` was a `POST` and not a `GET`. Fixed that up as well. 

Let me know how this is looking now with the changes. 🙇

github/checks.go
```diff
@@ -291,24 +291,24 @@ func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner strin
 	return checkRunResults, resp, nil
 }
 
-// ListCheckSuiteOptions Optional parameters to list check suites.
+// ListCheckSuiteOptions represents  parameters to list check suites.
```
gmlewis: nix double space.

github/checks.go
```diff
@@ -350,13 +350,13 @@ type CheckSuitePreferenceResults struct {
 
 // PreferenceList Represents a list of auto trigger checks for  repository
 type PreferenceList struct {
-	AutoTriggerChecks []AutoTriggerCheck `json:"auto_trigger_checks,omitempty"` // An array of auto trigger checks that can be set for a check suite in a repository.
+	AutoTriggerChecks []AutoTriggerCheck `json:"auto_trigger_checks,omitempty"` // A slice of auto trigger checks that can be set for a check suite in a repository.
```
gmlewis: `[]*AutoTriggerCheck`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+// ListCheckRunsResults represents the result of a check run list.
+type ListCheckRunsResults struct {
+	Total     *int        `json:"total_count,omitempty"`
+	CheckRuns []*CheckRun `json:"check_runs,omitempty"`
+}
+
+// ListCheckRunsForRef lists check runs for a specific ref.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/runs/list-check-runs-for-a-specific-ref
+func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner, repo string, ref string, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
```
gmlewis: `owner, repo, ref string`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRunResults, resp, nil
+}
+
+// ListCheckSuiteOptions represents  parameters to list check suites.
+type ListCheckSuiteOptions struct {
+	CheckName *string `url:"check_name,omitempty"` // Filters checks suites by the name of the check run.
```
gmlewis: `json:`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	Filter    *string `url:"filter,omitempty"`     // Filters check suites by GitHub App id.
```
gmlewis: `json:`

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkRunAnnotations, resp, nil
+}
+
+// ListCheckRunsOptions represents parameters to list check runs.
+type ListCheckRunsOptions struct {
+	CheckName *string `url:"check_name,omitempty"` // Returns check runs with the specified name.
```
gmlewis: `json:` here and the next two lines

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+// ListCheckSuiteResults represents the result of a check run list.
+type ListCheckSuiteResults struct {
+	Total       *int          `json:"total_count,omitempty"`
+	CheckSuites []*CheckSuite `json:"check_suites,omitempty"`
+}
+
+// ListCheckSuitesForRef lists check suite for a specific ref.
+//
+// GitHub API docs: https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref
+func (s *ChecksService) ListCheckSuitesForRef(ctx context.Context, owner, repo string, ref string, opt *ListCheckSuiteOptions) (*ListCheckSuiteResults, *Response, error) {
```
gmlewis: `owner, repo, ref string`

github/checks.go
sridharavinash: @gmlewis I think these need to be `url` since they are URL parameters passed to the `GET` query in https://developer.github.com/v3/checks/runs/#list-check-runs-for-a-specific-ref. If I switch this back to `json`, I fail validation in testing the `form values` here https://github.com/google/go-github/pull/914/files#diff-bd0176fc496a77ad3d32d734f27d98e4R348 ?

github/checks.go
sridharavinash: This field is now changed to `AppID` since the API has changed since the last time I looked at it or I missed it entirely on the first run. But I think the https://github.com/google/go-github/pull/914#discussion_r191454875 comment is still valid?

github/checks.go
sridharavinash: same as https://github.com/google/go-github/pull/914#discussion_r191454875 comment. These are `url` params to the `GET` query so should probably stay `url` instead of `json`?

github/checks.go
gmlewis: This change can still be made.

github/checks.go
```diff
@@ -291,10 +291,10 @@ func (s *ChecksService) ListCheckRunsCheckSuite(ctx context.Context, owner, repo
 	return checkRunResults, resp, nil
 }
 
-// ListCheckSuiteOptions represents  parameters to list check suites.
+// ListCheckSuiteOptions represents parameters to list check suites.
 type ListCheckSuiteOptions struct {
 	CheckName *string `url:"check_name,omitempty"` // Filters checks suites by the name of the check run.
-	Filter    *string `url:"filter,omitempty"`     // Filters check suites by GitHub App id.
+	AppID     *int    `url:"app_id,omitempty"`     // Filters check suites by GitHub App id.
```
gmlewis: I'm confused by this change.
I'm looking at this:
https://developer.github.com/v3/checks/runs/#list-check-runs-for-a-specific-ref
and it appears to take the `filter` argument, not `app_id`.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	Preferences *PreferenceList `json:"preferences,omitempty"`
+	Repository  *Repository     `json:"repository,omitempty"`
+}
+
+// PreferenceList represents a list of auto trigger checks for  repository
+type PreferenceList struct {
+	AutoTriggerChecks []*AutoTriggerCheck `json:"auto_trigger_checks,omitempty"` // A slice of auto trigger checks that can be set for a check suite in a repository.
+}
+
+// SetCheckSuitePreferences Changes the default automatic flow when creating check suites.
```
gmlewis: Somehow GitHub hid my comment.  s/Changes/changes/ can still be fixed.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	return checkSuitePrefResults, resp, nil
+}
+
+// CreateCheckSuiteOptions sets up parameters to manually create a check suites
+type CreateCheckSuiteOptions struct {
+	HeadSHA    string  `json:"head_sha"`              // The sha of the head commit. (Required.)
+	HeadBranch *string `json:"head_branch,omitempty"` // The name of the head branch where the code changes are implemented.
+}
+
+// CreateCheckSuite Manually creates a check suite for a repository.
```
gmlewis: s/Manually/manually/
The goal of all these changes is to make the resulting Godocs read naturally.
See: https://godoc.org/github.com/google/go-github/github

github/checks.go
sridharavinash: @gmlewis This one is for `check suites` :) . https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref

github/checks.go
sridharavinash: 👍

github/checks.go
sridharavinash: 👍

sridharavinash:
Hope I got all of them @gmlewis. Thanks for the extra 👀 . 🙇

WillAbides:
Anybody else want to give a 👍 ?

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// ChecksService provides access to the Checks API in the
+// GitHub API.
```
haya14busa: [super nit] no need to break line.

github/checks.go
```diff
@@ -0,0 +1,428 @@
...
+	resp, err := s.client.Do(ctx, req, checkSuite)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return checkSuite, resp, nil
+}
+
+// CreateCheckRunOptions sets up parameters needed to create a CheckRun.
+type CreateCheckRunOptions struct {
```
haya14busa: GitHub recently added new "action" field.
https://developer.github.com/v3/checks/runs/#parameters
https://developer.github.com/changes/2018-05-23-request-actions-on-checks/

github/event_types.go
```diff
@@ -7,6 +7,38 @@
 
 package github
 
+// CheckRunEvent is triggered when a check run is "created", "updated", or "re-requested".
+// The Webhook event name is "check_run".
+//
+// GitHub API docs: https://developer.github.com/v3/activity/events/types/#checkrunevent
+type CheckRunEvent struct {
+	CheckRun *CheckRun `json:"check_run,omitempty"`
+	// The action performed. Can be "created", "updated" or "re-requested".
```
haya14busa: requested_action is added.
https://developer.github.com/changes/2018-05-23-request-actions-on-checks/
https://developer.github.com/v3/activity/events/types/#checkrunevent

chrsmith:
Hi! I've patched this PR into my local enlistment to start using it in (eager) anticipation of it being merged. It appears there is a small thing missing from this PR that would be nice to add before it gets committed.

The PR doesn't work with the `ParseWebHook` method. This is due to a map not containing the new entries. In `messages.go` there is a map associating web hook type names to their corresponding structs.

`messages.go:40`
```
var (
	// eventTypeMapping maps webhooks types to their corresponding go-github struct types.
	eventTypeMapping = map[string]string{
		// MISSING: "check_suite": "CheckSuiteEvent", and others.
		"commit_comment":              "CommitCommentEvent",
		"create":                      "CreateEvent",
		"delete":                      "DeleteEvent",
```

And then later, in `activity_events.go`, where the event payload is actually parsed.

`activity_events.go:31`
```
// ParsePayload parses the event payload. For recognized event types,
func (e *Event) ParsePayload() (payload interface{}, err error) {
	switch *e.Type {
		// MISSING: "CheckSuiteEvent" and others.
	case "CommitCommentEvent":
		payload = &CommitCommentEvent{}
	case "CreateEvent":
		...
```

By adding the new hooks into those method, a consumer (like me) can use `ParseWebHook` to build a switch statement keyed to handle incoming webhooks.

```
		webhookType := github.WebHookType(r)
		event, err := github.ParseWebHook(webhookType, payload)
		if err != nil {
				...
		}

		switch event := event.(type) {
		case *github.CheckSuiteEvent:
			...
			break
		case *github.PullRequestEvent:
			...
			break
		...
```

@sridharavinash , I'll start working on a pull request to your fork at https://github.com/sridharavinash/go-github/tree/add-checks-api. Though I must admit I'm not sure how "pull requests to forks that became pull requests" is supposed to work.

sridharavinash:
@chrsmith  That's a great catch. Thanks for fixing it  ❤️ .

>Though I must admit I'm not sure how "pull requests to forks that became pull requests" is supposed to work.

I'm not sure either 😆 , but i'll merge your PR and see what happens 🤞 .

chrsmith:
Hi @googlebot , I am OK with my commits being contributed to the project.

haya14busa:
Any updates on this p-r? I'd like to use check API with go-github, and I used go-github with this p-r from about 2 month ago, but i cannot merge my dev branch to master...

Considering this slow updates, I cannot help but considering to stop go-github...
Are there anything i can help???

gmlewis:
Yes, @haya14busa - how about giving this PR a full Go code review and if you are happy with it, then given it an LGTM? That would satisfy the requirement for two LGTMs on non-trivial PRs and we could then merge this.  Thank you!

jamiehannaford:
@gmlewis This LGTM, could we get this merged? 🙇

gmlewis:
Thank you, @jamiehannaford, @sridharavinash, and @chrsmith!
Merging.

jamiehannaford:
Thanks @gmlewis!

PatWie:
Sorry to chime in. But did anyone tested these functions in the wild? (Not just running the unit tests.)

```go
_, _, err = client.Checks.CreateCheckRun(ctx, "owner", "repo", opt)
if err != nil {
    panic(err)
}
```

gives me

```
panic: POST https://api.github.com/repos/owner/repo/check-runs: 415 If you would like to help us test the Checks API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details. []
```

As far as I can tell, it is [there](https://github.com/google/go-github/blob/b93adc1c7febaf7dad1bed965e5310a86bc2623e/github/checks.go#L156)

jamiehannaford:
@PatWie I was able to create a check run with the new functionality successfully today.

sridharavinash:
@PatWie Check Runs requires authentication via `GitHub Apps` 

>Note: The Checks API is currently in public beta and only available for use with GitHub Apps.

Here's a [gist example](https://gist.github.com/sridharavinash/4515e3bedd79d5fb4f2ffc6ab8bf704f) of how I was able to get it to work after creating a GitHub App and installing it on the repo I wanted to create the `check-run` on.

haya14busa:
btw, why my review was ignored? https://github.com/google/go-github/pull/914#pullrequestreview-129356983

I think it was totally no problem to merge the p-r and create another p-r to add the missing feature, but do you have any plan to work on this??? @sridharavinash

sridharavinash:
>create another p-r to add the missing feature, but do you have any plan to work on this??? @sridharavinash

@haya14busa I haven't got to it yet, but feel free create an issue and add a PR.  🙇

PatWie:
@sridharavinash, @jamiehannaford I executed the exact same code today (I had implemented a GitHub app exactly like the gist but without `Status`) and it works right away without any changes. I guess this is was it meant to be in "BETA".

PR #916, opened 2018-05-29
Revert pull request number from int64 to int

mspiegel:
In a few instances the pull request number was accidentally promoted from an int to an int64.


gmlewis:
Thank you for catching this, @mspiegel!

Issue #918, opened 2018-06-01
Support additional endpoints for the Migrations API


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2018-05-24-user-migration-api/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue, and we can send you an invite to contribute to the repo (which you then accept after you enable two-factor authentication) and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our CONTRIBUTING.md guide to get started.

Thank you!

sanketpayghan:
Can you send me an invite. I am new to golang but Would like to work on it.

willnorris:
@sanketpayghan: invite sent.

sanketpayghan:
Thanks @willnorris

sanketpayghan:
Working on this

mohanprasaths:
@sanketpayghan  If you are not working on this  I would like to work on this

PR #927, opened 2018-06-06
support for user migration API (#918)

sanketpayghan:


github/github-accessors.go
```diff
@@ -10708,6 +10708,70 @@ func (u *UserLDAPMapping) GetURL() string {
 	return *u.URL
 }
 
+// GetCreatedAt returns the CreatedAt field if it's non-nil, zero value otherwise.
+func (u *UserMigration) GetCreatedAt() string {
+	if u == nil || u.CreatedAt == nil {
```
petar-dambovaliev: is it possible for "u" to be nil?
i don't think you can invoke a function on a nil.

gmlewis: Yes indeed it is possible.

Check this out: https://play.golang.org/p/ZgOImESxgz_X

petar-dambovaliev: Ohh, ok.

sanketpayghan:
@gmlewis Please review

example/migrations/main.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"fmt"
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+func FetchAllUserMigrations() ([]*github.UserMigration, error) {
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: "b453839afb5b7871b9dabfe8184b314747f6523a"},
```
gmlewis: Please don't add access tokens to the repo... and if this is a valid one, you will now want to go and delete it as it has been exposed to the world.

`&oauth2.Token{AccessToken: "b45...access-token-here...3a"},`

example/migrations/main.go
```diff
@@ -0,0 +1,41 @@
...
+	)
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
+
+	user_migrations, _, err := client.Migrations.ListUserMigrations(ctx)
+
```
gmlewis: nit: I would nix this blank line to more tightly couple the call with the resulting return value.

example/migrations/main.go
```diff
@@ -0,0 +1,41 @@
...
+	return user_migrations, err
+}
+
+func main() {
+	user_migrations, err := FetchAllUserMigrations()
```
gmlewis: Go style comment:
s/user_migrations/userMigrations/
or just `migrations`.

example/migrations/main.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		fmt.Printf("Error %v\n", err)
+		return
+	}
+
+	for i, user_migration := range user_migrations {
```
gmlewis: Go style comment:

```go
for i, userMigration := range userMigrations {
```

or

```go
for i, m := range migrations {
```

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2016/2018/

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// ExcludeAttachments indicates whether attachments should be excluded from
+	// the migration (to reduce migration archive file size).
+	ExcludeAttachments *bool `json:"exclude_attachments,omitempty"`
+}
+
+// StartUserMigration starts the generation of a migration archive.
+// repos is a slice of repository names to migrate.
+//
+// GitHub API docs: https://developer.github.com/v3/migration/migrations/#start-a-migration
```
gmlewis: https://developer.github.com/v3/migrations/orgs/#start-a-migration

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return m, resp, nil
+}
+
+// ListUserMigrations lists the most recent migrations.
+//
+// GitHub API docs: https://developer.github.com/v3/migration/migrations/#get-a-list-of-migrations
```
gmlewis: https://developer.github.com/v3/migrations/orgs/#get-a-list-of-migrations

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+	m := &UserMigration{}
+	resp, err := s.client.Do(ctx, req, m)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return m, resp, nil
+}
+
+// UserMigrationArchiveURL get the URL for specific migration archive.
```
gmlewis: s/get/gets/
s/for specific/for a specific/

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+	}()
+	resp, err := s.client.Do(ctx, req, m)
+	if err == nil {
+		return "", errors.New("expected redirect, none provided")
+	}
+	loc = resp.Header.Get("Location")
+	return loc, nil
+}
+
+// DeleteUserMigration will delete previous migration archive
```
gmlewis:
```go
// DeleteUserMigration will delete a previous migration archive.
```

Please note the period at the end of the sentence.

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+// id is the migration ID
```
gmlewis: Add period to end of sentence.

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMigrationsPreview)
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// UnlockUserRepository will unlock repo that was locked for migration.
+// id is migration ID
```
gmlewis: Add period to end of sentence.

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+		return nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeMigrationsPreview)
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// UnlockUserRepository will unlock repo that was locked for migration.
```
gmlewis: s/unlock repo/unlock a repo/

github/migrations_user_test.go
```diff
@@ -0,0 +1,197 @@
+// Copyright 2016 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2016/2018/

example/migrations/main.go
sanketpayghan: Yes. I have already deleted this access token. Now removed from code too.

example/migrations/main.go
```diff
@@ -0,0 +1,41 @@
...
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+func FetchAllUserMigrations() ([]*github.UserMigration, error) {
```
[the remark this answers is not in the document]
sanketpayghan: did not get this. Please elaborate.

github/migrations_user.go
sanketpayghan: should it be
https://developer.github.com/v3/migrations/users/#start-a-user-migration?

github/migrations_user.go
sanketpayghan: should it be https://developer.github.com/v3/migrations/users/#get-a-list-of-user-migrations?

example/migrations/main.go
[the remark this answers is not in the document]
sanketpayghan: oh...yeah got it.

sanketpayghan:
@gmlewis Thanks fo review. I have made changes as suggested.

github/migrations_user.go
```diff
@@ -0,0 +1,214 @@
...
+
+// StartUserMigration starts the generation of a migration archive.
+// repos is a slice of repository names to migrate.
+//
+// GitHub API docs: https://developer.github.com/v3/migrations/users/#start-a-user-migration
+func (s *MigrationService) StartUserMigration(ctx context.Context, repos []string, opt *UserMigrationOptions) (*UserMigration, *Response, error) {
+	u := "user/migrations"
+
+	body := &startUserMigration{Repositories: repos}
+	if opt != nil {
```
juliaferraioli: Would it be better to use the [`addOptions` function](https://github.com/google/go-github/blob/master/github/github.go#L205) instead?

You can see examples of this being done before across the codebase, including [here](https://github.com/google/go-github/blob/b93adc1c7febaf7dad1bed965e5310a86bc2623e/github/activity_events.go#L152)

sanketpayghan: Looks like addOptions function is used to add query parameters to url.
May be name `UserMigrationOptions` is confusing. Can change `StartUserMigration` methods parameter to accept `startUserMigration` stuct directly, where repos list will be mandatory.

@juliaferraioli Please correct me if I am wrong and also suggest about above approach or any other approach.

juliaferraioli: @sanketpayghan No, you're totally right! I missed that. The `...Options` name does seem misleading.  I think that might be a good change to make. WDYT @gmlewis?

gmlewis: I'm thinking that `...Options` makes sense from the perspective of the user of this API because it modifies the behavior of the call despite not going through the internal `addOptions` mechanism.

I've looked through the code, and we do have precedent for using the name `...Options` even when `addOptions` is not involved... for example:

* `Client.Markdown`
* `MigrationService.StartMigration`
* `OrganizationsService.CreateOrgInvitation`
* `ProjectsService.CreateProjectColumn`
* `ProjectsService.MoveProjectColumn`
* `ProjectsService.UpdateProject`
* `ProjectsService.UpdateProjectColumn`
* `PullRequestsService.GetRaw`
* `PullRequestsService.Merge`
* `RepositoriesService.AddCollaborator`
* `RepositoriesService.CreateProject`
* `RepositoriesService.GetArchiveLink`
* `RepositoriesService.GetCommitRaw`
* `SearchService.*`

This was just a cursory glance... I may have got some of these wrong, but I think it is OK to move forward with the current naming scheme, as the typical end-user of this repo probably doesn't care about the internals, but the word `Options` should make sense to them for tweaking the behavior of the endpoint. At least, that's my $0.02. :smile:

juliaferraioli: SGTM!

gmlewis: OK, great. Thank you, @juliaferraioli! Do you mind giving this an official LGTM approval so we can go ahead and merge?

juliaferraioli: @gmlewis was just fishing out my laptop :-) Done.

gmlewis:
Thank you, @sanketpayghan and @juliaferraioli!
Merging.

PR #931, opened 2018-06-15
remove addition of preview Accept headers in RepositoriesService.ListByOrg (repos.go)

### LINKED ISSUE #930: addition of preview Accept headers in RepositoriesService.ListByOrg causes errors with GHE
Preview headers cause server 500 Errors with GHE (observed with GHE version 2.11 - in particular, it is the `application/vnd.github.scarlet-witch-preview+json` header that is causing it).  It happens rarely - only on some requests with some specific `page` parameter value - but when it does, it happens consistently (always for the same request / URL) thus preventing the listing of public repositories.

Currently the several preview Accept types are added to ALL `RepositoriesService.ListByOrg` calls - even when no preview features are needed or wanted, with no possibility to turn the header addition off.  Maybe this can be made configurable so that clients could at least turn off the Accept header addition - for example, through an additional field in `RepositoryListByOrgOptions` or some other way.  Alternatively, we could make the users who do need to use the preview features to specify corresponding options.

dmitris:
Fixes #930.

The change in this PR removes the mediaTypeLicensesPreview, mediaTypeCodesOfConductPreview,
and mediaTypeTopicsPreview in RepositoriesService.ListByOrg calls to prevent the GHE errors.  It's fine if you reject it in case it is too drastic or breaks other users' code, but it would be great to find a solution.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -216,10 +216,6 @@ func (s *RepositoriesService) ListByOrg(ctx context.Context, org string, opt *Re
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeCodesOfConductPreview, mediaTypeTopicsPreview}
-	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
-
 	var repos []*Repository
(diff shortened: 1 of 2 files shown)


gmlewis:
Fixed by other PR. Closing.

Issue #617, opened 2017-04-13
Update Protected Branches API preview


gmlewis:
GitHub Announcement:
https://developer.github.com/changes/2017-04-11-protected-branch-dismissal-restrictions/

Note that the [provided link in the announcement](https://developer.github.com/v3/repos/branches/#get-pull-request-review-enforcement-of-protected-branch) is somewhat misleading in that this change affects potentially more than just this one API endpoint on this page.

I think the point is to ensure that `dismissal_restrictions` is handled properly.

navarrano:
I can work on this. @gmlewis, please assign it to me

gmlewis:
Please also check out the latest GitHub Developer announcement about this API:
https://developer.github.com/changes/2017-04-20-protected-branches-dismiss-stale-review/
If you can incorporate these changes as well, it seems reasonable to me to do so within the same PR since it is the same preview API.

If not, though, that is fine too and we can open up a new issue for it.

navarrano:
@gmlewis No problem, I'll bring this changes as well, no need to create another issue

gmlewis:
There is yet another GitHub Developer announcement about breaking changes to this API:
https://developer.github.com/changes/2017-05-02-adoption-of-admin-enforced/

gmlewis:
There is another announcement for this API:
https://developer.github.com/changes/2017-06-16-loki-preview-ending-soon/

gmlewis:
I believe that this issue is still open, so I didn't close it with #637.
Please correct me if I'm wrong as I haven't done a thorough investigation at this point.

e-beach:
The Jun 22 announcement did not introduce new changes, but it means we can remove the special accept header for the API preview on Sept 1st.

The functions related to protected branches such as `updateBranchProtection` already work in the way described in the announcement. The update endpoint uses the `PUT` method, and we have a `requiredPullRequestReview ` property of the struct which the client can populate. https://github.com/google/go-github/blob/fe4b6036cb400908b8903d3df9444b9599a60d57/github/repos.go#L773-774.

I can take of care of removing the accept header, although it is a very simple change.

gmlewis:
While here, @e-beach, can you determine if removing the API preview accept header would affect Enterprise API clients?  I seem to recall that if you are not a GitHub Enterprise customer, that you don't have access to the Enterprise API docs, but this was a long time ago and I could be mistaken. We just had a discussion about this in #638.

e-beach:
[...]

PR #934, opened 2018-06-22
Support preview Protected Branches API

### LINKED ISSUE #876: Support preview Protected Branches API

scriptonist:
This PR corresponds to https://github.com/google/go-github/issues/876

I've removed the [protected branches preview](https://developer.github.com/changes/2015-11-11-protected-branches-api/) [media type](https://github.com/google/go-github/blob/a83ae98ad5d09188c49d6056edb60ec9bdf202bd/github/github.go#L55) from the corresponding functions which were required to complete this PR.Please review and suggest fixes if needed, Apologies for the late PR.

Fixes #876


github/repos.go
```diff
@@ -560,11 +560,13 @@ type RequiredStatusChecks struct {
 // PullRequestReviewsEnforcement represents the pull request reviews enforcement of a protected branch.
 type PullRequestReviewsEnforcement struct {
 	// Specifies which users and teams can dismiss pull request reviews.
```
gmlewis: As written. the new `omitempty` tags aren't useful because the fields are not pointers and therefore can not be `nil`.

But looking back at https://github.com/google/go-github/pull/637/files , I'm not seeing why these were not originally made to be pointers with the `omitempty` tag. Maybe @shurcooL remembers? I read through the comments but it looks like I didn't bring up this point before and neither did anyone else.

Unless @shurcooL says otherwise, I'm thinking that these fields need to be turned into pointers and the `omitempty` tags should be added since these are return values.

scriptonist: I'm sorry I didn't get you. Should I make any changes or are we waiting for @shurcooL for a response?

gmlewis: Let's go ahead and change these fields to pointers and add `omitempty` since this struct is part of the return value.

scriptonist: @gmlewis let me make sure I got It correct, you are suggesting that we add `omitempty` to the following structs, Right?

https://github.com/scriptonist/go-github/blob/0120b56d9901fcafe451c9d08ea8f446478ce0d0/github/repos.go#L536

https://github.com/scriptonist/go-github/blob/0120b56d9901fcafe451c9d08ea8f446478ce0d0/github/repos.go#L544

https://github.com/scriptonist/go-github/blob/0120b56d9901fcafe451c9d08ea8f446478ce0d0/github/repos.go#L552

gmlewis: Sorry for the confusion.  It looks like you have already added `omitempty` to all the `PullRequestReviewsEnforcement` struct fields, which is what I was requesting.

I'm thinking that to make the `omitempty` tags effective, we should also make all the fields pointers.

So they would be:
```go
DismissalRestrictions *DismissalRestrictions ...
DismissStaleReviews *bool ...
RequireCodeOwnerReviews *bool ...
...
```

Please let me know if that is not clear.

@jkosecki - maybe you can comment on what you think about these proposed changes since you authored #617?

scriptonist: This may be a noob question but why are we opting pointers for types like `int` and `bool`? @gmlewis

gmlewis: That is an excellent question, @scriptonist, and there have been several very good discussions about it. In a nutshell, it is so that we can tell if the server populated the field that it sent back to us which is especially important for variables like `bool` where you want to know if you are really getting back a `false` or if the server didn't send a value at all.

It is actually more important on the request side than the response side, really... which is why they may not be needed at all in this situation. I think I'm contradicting what I said earlier, unfortunately, and apologize. I said "since this struct is part of the return value". That is not good enough reason to make this change, really...

So I'm thinking now, let's hold off on my most recent request to make these all pointers.
Especially if the docs seem to indicate that these fields will always be sent in the reply.

Hopefully @jkosecki or @shurcooL can provide a code review and then we can get this merged when they are also happy with the changes. As this stands, I'm going to give this my LGTM.

Thanks again, @scriptonist, for bearing with me.

Here is a search of some of the previous great discussions about using pointers in this repo:
https://github.com/google/go-github/issues?utf8=%E2%9C%93&q=is%3Aissue+pointers

gmlewis:
Thank you, @juliaferraioli!
When attempting to resolve conflicts, I see that I've allowed similar changes to be made in other PRs.
I'm going to have to create a new PR for this one to pull in its changes and manually resolve conflicts.

juliaferraioli:
@gmlewis I was worried that would be the case. Thanks :-\

gmlewis:
Actually, now that I'm looking at the final results of the merge, all the important things have been addressed, and the extra `omitempty`s are really not desirable anyway because these fields can't be omitted as they are not pointers.

See #974 for details.

Closing.

PR #938, opened 2018-06-23
Add UpdateRequiredStatusChecks function

### LINKED ISSUE #937: Add UpdateRequiredStatusChecks function
Add function for endpoint described at https://developer.github.com/v3/repos/branches/#update-required-status-checks-of-protected-branch.

nmiyake:
Fixes #937

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -557,6 +557,12 @@ type RequiredStatusChecks struct {
 	Contexts []string `json:"contexts"`
 }
 
+// RequiredStatusChecksRequest represents a request to edit a protected branch's status checks.
+type RequiredStatusChecksRequest struct {
+	Strict   *bool    `json:"strict,omitempty"`
+	Contexts []string `json:"contexts,omitempty"`
+}
+
 // PullRequestReviewsEnforcement represents the pull request reviews enforcement of a protected branch.
 type PullRequestReviewsEnforcement struct {
 	// Specifies which users and teams can dismiss pull request reviews.
@@ -789,6 +795,25 @@ func (s *RepositoriesService) RemoveBranchProtection(ctx context.Context, owner,
 	req.Header.Set("Accept", mediaTypeProtectedBranchesPreview)
 
 	return s.client.Do(ctx, req, nil)
+}
+
+// UpdateRequiredStatusChecks updates the required status checks for a given protected branch.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#update-required-status-checks-of-protected-branch
+func (s *RepositoriesService) UpdateRequiredStatusChecks(ctx context.Context, owner, repo, branch string, sreq *RequiredStatusChecksRequest) (*RequiredStatusChecks, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_status_checks", owner, repo, branch)
+	req, err := s.client.NewRequest("PATCH", u, sreq)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	sc := new(RequiredStatusChecks)
+	resp, err := s.client.Do(ctx, req, sc)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return sc, resp, nil
 }
 
 // License gets the contents of a repository's license if one is detected.
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -9140,6 +9140,14 @@ func (r *RepoStatus) GetURL() string {
 	return *r.URL
 }
 
+// GetStrict returns the Strict field if it's non-nil, zero value otherwise.
+func (r *RequiredStatusChecksRequest) GetStrict() bool {
+	if r == nil || r.Strict == nil {
+		return false
+	}
+	return *r.Strict
+}
+
 // GetName returns the Name field if it's non-nil, zero value otherwise.
 func (s *ServiceHook) GetName() string {
 	if s == nil || s.Name == nil {
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -727,6 +727,41 @@ func TestRepositoriesService_GetRequiredStatusChecks(t *testing.T) {
 	}
 }
 
+func TestRepositoriesService_UpdateRequiredStatusChecks(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	input := &RequiredStatusChecksRequest{
+		Strict:   Bool(true),
+		Contexts: []string{"continuous-integration"},
+	}
+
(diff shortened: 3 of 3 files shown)


github/repos.go
```diff
@@ -557,6 +557,12 @@ type RequiredStatusChecks struct {
 	Contexts []string `json:"contexts"`
 }
 
+// RequiredStatusChecksRequest represents a request to edit a protected branch's status checks.
+type RequiredStatusChecksRequest struct {
```
nmiyake: Adding request object because `UpdateRequiredStatusChecks` is a `PATCH` function so fields can be omitted in requests.

github/repos.go
```diff
@@ -557,6 +557,12 @@ type RequiredStatusChecks struct {
...
+	Strict   *bool     `json:"strict"`
+	Contexts *[]string `json:"contexts"`
```
gmlewis: A slice is already a reference object, so this should be `[]string` and not `*[]string`.
Also, please add `omitempty` to this one as well.

github/repos.go
```diff
@@ -557,6 +557,12 @@ type RequiredStatusChecks struct {
 	Contexts []string `json:"contexts"`
 }
 
+// RequiredStatusChecksRequest represents a request to edit a protected branch's status checks.
+type RequiredStatusChecksRequest struct {
+	Strict   *bool     `json:"strict"`
```
gmlewis: Please add `omitempty` to both fields since they are optional.

github/repos.go
```diff
@@ -791,6 +797,28 @@ func (s *RepositoriesService) RemoveBranchProtection(ctx context.Context, owner,
 	return s.client.Do(ctx, req, nil)
 }
 
+// UpdateRequiredStatusChecks updates the required status checks for a given protected branch.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/branches/#update-required-status-checks-of-protected-branch
+func (s *RepositoriesService) UpdateRequiredStatusChecks(ctx context.Context, owner, repo, branch string, sreq *RequiredStatusChecksRequest) (*RequiredStatusChecks, *Response, error) {
```
gmlewis: Even though the fields are optional, this call does nothing if `sreq` is `nil` so I'm thinking that we should either make `sreq` a value and not a pointer or add a check so that `nil` can not be passed.

github/repos.go
```diff
@@ -791,6 +797,28 @@ func (s *RepositoriesService) RemoveBranchProtection(ctx context.Context, owner,
...
+	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection/required_status_checks", owner, repo, branch)
+	req, err := s.client.NewRequest("PATCH", u, sreq)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
```
gmlewis: nit: add period to end of sentence.

From the GitHub Developer API docs, it is not clear to me that this custom media type is required for this call.

github/repos.go
nmiyake: Will do for `omitempty`.

How important is consistency? I was trying to match the other code already in the repo, and other request objects with optional slices use pointers -- for example, in the same file:

```
type DismissalRestrictionsRequest struct {
	// The list of user logins who can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or &[]string{} otherwise.)
	Users *[]string `json:"users,omitempty"`
	// The list of team slugs which can dismiss pull request reviews. (Required; use nil to disable dismissal_restrictions or &[]string{} otherwise.)
	Teams *[]string `json:"teams,omitempty"`
}
```

github/repos.go
nmiyake: Should this behavior differ from other functions? I modeled the implementation of this after the existing `UpdateBranchProtection` function in the same file:

```
// UpdateBranchProtection updates the protection of a given branch.
//
// GitHub API docs: https://developer.github.com/v3/repos/branches/#update-branch-protection
func (s *RepositoriesService) UpdateBranchProtection(ctx context.Context, owner, repo, branch string, preq *ProtectionRequest) (*Protection, *Response, error) {
	u := fmt.Sprintf("repos/%v/%v/branches/%v/protection", owner, repo, branch)
	req, err := s.client.NewRequest("PUT", u, preq)
	if err != nil {
		return nil, nil, err
	}

	// TODO: remove custom Accept header when this API fully launches
	req.Header.Set("Accept", mediaTypeProtectedBranchesPreview)

	p := new(Protection)
	resp, err := s.client.Do(ctx, req, p)
	if err != nil {
		return nil, resp, err
	}

	return p, resp, nil
}
```

github/repos.go
nmiyake: This is from the other function I used as a base for this one. However, agree that, based on the docs, this media type should not be required for this particular call -- will remove.

github/repos.go
nmiyake: Actually on second glance it looks like the file has both styles:

```go
type TransferRequest struct {
	NewOwner string  `json:"new_owner"`
	TeamID   []int64 `json:"team_id,omitempty"`
}
```

Will remove pointer reference.

gmlewis: Consistency is very important in this repo, but in this case, the `*[]string` is not required I don't believe because an empty list here would just mean no contexts (if I'm reading the docs right). Do you agree?

github/repos.go
gmlewis: Usually in the GitHub Developer API documentation, they mention the custom media headers on an endpoint-by-endpoint basis. So I'm thinking it doesn't apply here.

nmiyake:
Thanks for the review! Addressed most of the feedback -- will wait to hear back on whether I should change the function parameter or keep it consistent with the other ones.

github/repos.go
nmiyake: Yes, agree and removed in latest PR update.

github/repos.go
nmiyake: Yes, agree and updated in latest PR update.

nmiyake:
Ping -- still waiting on response to latest question (all other PR feedback has been addressed in update)

github/repos.go
gmlewis: Sorry for the delay, @nmiyake!
Hmmm... I see you are right about not checking `preq` for `nil`... seems odd to me but shows precedent, so let's run with it. We could always clean all these up later in a subsequent PR. This one has been held up long enough.

nmiyake:
Ping for another reviewer (@juliaferraioli or anyone else actively doing reviews) to unblock

gmlewis:
Thank you, @nmiyake and @juliaferraioli!
Merging.

PR #945, opened 2018-07-11
Add APIs for pre-receive hooks

### LINKED ISSUE #944: Add support for pre-receive hooks API
https://developer.github.com/v3/repos/pre_receive_hooks/

This API is only available to GitHub Enterprise instances, but it's still a published part of the API so I think that adding support for this in this library would make sense.

nmiyake:
Fixes #944

### DIFF
--- a/github/repos_prereceive_hooks.go
+++ b/github/repos_prereceive_hooks.go
@@ -1,0 +1,110 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// PreReceiveHook represents a GitHub pre-receive hook for a repository.
+type PreReceiveHook struct {
+	ID          *int64  `json:"id,omitempty"`
+	Name        *string `json:"name,omitempty"`
+	Enforcement *string `json:"enforcement,omitempty"`
+	ConfigURL   *string `json:"configuration_url,omitempty"`
+}
+
+func (p PreReceiveHook) String() string {
+	return Stringify(p)
+}
+
+// ListPreReceiveHooks lists all pre-receive hooks for the specified repository.
+//
+// GitHub API docs: https://developer.github.com/enterprise/2.13/v3/repos/pre_receive_hooks/#list-pre-receive-hooks
+func (s *RepositoriesService) ListPreReceiveHooks(ctx context.Context, owner, repo string, opt *ListOptions) ([]*PreReceiveHook, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pre-receive-hooks", owner, repo)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypePreReceiveHooksPreview)
+
+	var hooks []*PreReceiveHook
+	resp, err := s.client.Do(ctx, req, &hooks)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return hooks, resp, nil
+}
+
+// GetPreReceiveHook returns a single specified pre-receive hook.
+//
+// GitHub API docs: https://developer.github.com/enterprise/2.13/v3/repos/pre_receive_hooks/#get-a-single-pre-receive-hook
+func (s *RepositoriesService) GetPreReceiveHook(ctx context.Context, owner, repo string, id int64) (*PreReceiveHook, *Response, error) {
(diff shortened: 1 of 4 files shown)


github/repos_prereceive_hooks.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2013 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2013/2018/ as that is the year this particular file is created.

github/repos_prereceive_hooks.go
```diff
@@ -0,0 +1,102 @@
...
+
+// PreReceiveHook represents a GitHub pre-receive hook for a repository.
+type PreReceiveHook struct {
+	ID          *int64  `json:"id,omitempty"`
+	Name        *string `json:"name,omitempty"`
+	Enforcement *string `json:"enforcement,omitempty"`
+	ConfigURL   *string `json:"configuration_url,omitempty"`
+}
+
+func (h PreReceiveHook) String() string {
```
gmlewis: Go style nit: `func (p PreReceiveHook)...`
s/h/p/ to match the receiver variable with the start of the name of its type...
although I see that this Go style recommendation is not followed consistently throughout this repo, so you can ignore this comment safely.  I'll leave it anyway in case someone wants to clean up all receivers in this repo. :smile:

github/repos_prereceive_hooks.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypePreReceiveHooksPreview)
+
+	hook := new(PreReceiveHook)
+	resp, err := s.client.Do(ctx, req, hook)
+	return hook, resp, err
```
gmlewis: I think we would like the error checking to remain consistent, like your function above:

```go
if err != nil {
	return nil, resp, err
}

return hook, resp, nil
```

We think this is easier to read and reason about.

If you copied this pattern from somewhere else in this repo, we should probably change that one too.

github/repos_prereceive_hooks.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypePreReceiveHooksPreview)
+
+	h := new(PreReceiveHook)
+	resp, err := s.client.Do(ctx, req, h)
+	return h, resp, err
```
gmlewis: same here with the error handling

github/repos_prereceive_hooks.go
```diff
@@ -0,0 +1,102 @@
...
+	u := fmt.Sprintf("repos/%v/%v/pre-receive-hooks/%d", owner, repo, id)
+	req, err := s.client.NewRequest("DELETE", u, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypePreReceiveHooksPreview)
+
+	return s.client.Do(ctx, req, nil)
```
gmlewis: This is fine to leave as-is since we are not returning any "interesting" struct pointers here.

github/repos_prereceive_hooks_test.go
```diff
@@ -0,0 +1,133 @@
+// Copyright 2013 The go-github AUTHORS. All rights reserved.
```
gmlewis: s/2013/2018/

github/repos_prereceive_hooks_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestRepositoriesService_ListPreReceiveHooks(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/pre-receive-hooks", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
```
gmlewis: Here and below, please add `testHeader` to check for the custom media type specified in the header.
That helps us to keep this repo cleaner by explicitly calling out when custom media headers are being used.

github/repos_prereceive_hooks.go
nmiyake: Yes, this file was a copy-modify of `repos_hooks.go`, so these issues are also copied from there. Will make the changes in this PR and open a separate PR to fix the occurrences in `repos_hooks.go`.

nmiyake:
Thanks for the review! Updated PR to address feedback.

gmlewis:
Thanks for the changes, @nmiyake.

Just FYI... it makes reviewing easier if you add additional commits (instead of force-pushing over top of an earlier commit). That way, reviewers can see just the changes that were made after the last review.

In the end, we squash and merge the final commit anyway, so feel free to add as many commits as you want.  :smile:

gmlewis:
Thank you, @nmiyake and @juliaferraioli!
Merging.

PR #950, opened 2018-07-13
Fix InReplyToID JSON tag for PullRequestComment

### LINKED ISSUE #948: Invalid PullRequestComment InReplyTo JSON tag
The PullRequestComment InReplyTo JSON tag mismatches the one returned from the API, so the value is always encoded as nil.

The current tag is `in_reply_to`: https://github.com/google/go-github/blob/e96f1f1b07a5be7cba161bc7c1fd636044bfd75c/github/pulls_comments.go#L17

Whereas the correct one is `in_reply_to_id`. Relevant docs: https://developer.github.com/v3/pulls/comments/#get-a-single-comment

emersonwood:
Fixes #948.

The JSON tag mismatches the one returned from the API, so the value is always encoded as nil.

Relevant docs: https://developer.github.com/v3/pulls/comments/#get-a-single-comment

Also have to use an alias since the JSON tag is different between getting a comment and creating a comment.

I'd be keen to see if there's a more elegant way to do this.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -6364,12 +6364,12 @@ func (p *PullRequestComment) GetID() int64 {
 	return *p.ID
 }
 
-// GetInReplyToID returns the InReplyToID field if it's non-nil, zero value otherwise.
-func (p *PullRequestComment) GetInReplyToID() int64 {
-	if p == nil || p.InReplyToID == nil {
+// GetInReplyTo returns the InReplyTo field if it's non-nil, zero value otherwise.
+func (p *PullRequestComment) GetInReplyTo() int64 {
+	if p == nil || p.InReplyTo == nil {
 		return 0
 	}
-	return *p.InReplyToID
+	return *p.InReplyTo
 }
 
 // GetOriginalCommitID returns the OriginalCommitID field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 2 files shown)


gmlewis:
I'm looking at the API for `CreateComment` and I'm not terribly happy with it as it currently stands.

For `CreateComment` to do its thing, we need just a small subset of the fields provided in the `PullRequestComment` struct... and there are two variations of the method in the API docs.

What if we were to split the current `CreateComment` method in two, like this:

```go
func (s *PullRequestsService) CreateComment(ctx context.Context, owner string, repo string, number int, 
comment *PullRequestComment) (*PullRequestComment, *Response, error) { ... }

func (s *PullRequestsService) CreateCommentInReplyTo(ctx context.Context, owner string, repo string, number int, body string, commentID int64) (*PullRequestComment, *Response, error) { ... }
```

That way, we don't break the current API, but add a simpler method to handle the alternate case?

(I was first going to create a `PRCommentRequest` struct that had just the fields needed in the first call, but that would break the API and we try not to do that when not absolutely necessary... and I think this case does not require breaking the API... but I think the new helper function to address the alternate call mechanism would be very nice to have and prevent you from having to make an alias.)

emersonwood:
Thanks for the feedback, I've updated the PR based on your recommendation.

I've updated the original `PullRequestComment` tag since it's currently incorrect for the GET request but I haven't changed the field name to avoiding breaking the existing API.

gmlewis:
Thank you, @juliaferraioli!
Merging.

PR #951, opened 2018-07-13
Make error return style in repos_hooks.go consistent

nmiyake:


nmiyake:
As flagged by @gmlewis in https://github.com/google/go-github/pull/945#discussion_r202214852

PR #954, opened 2018-07-20
fix go vet error with go1.11

dmitris:
currently builds of the package with go1.11 fail due to using `%q` qualifier with `*string` type in github/repos_contents_test.go:
```
$ go1.11beta2 tool vet .
github/repos_contents_test.go:59: T.Errorf format %q has arg tt.encoding of wrong type *string
github/repos_contents_test.go:62: T.Errorf format %q has arg tt.encoding of wrong type *string
```

In Travis, a number of PR builds fail with `Go:master` ("Allowed failures"), ex. https://travis-ci.org/google/go-github/builds/406237120

PR fixes the issue by treating the cases where tt.enconding and tt.content are nil or not nil separately and adding extra varaibles (`encoding` and `content`) for printing in error messages.  It also adds a workaround for gofmt difference between go1.10 and go1.11  (as in https://github.com/golang/go/issues/26228, Go commit: https://github.com/golang/go/commit/542ea5ad91367d2b40589942cc5757a4d5f43f97) by adding an empty line.

### DIFF
--- a/github/repos_contents_test.go
+++ b/github/repos_contents_test.go
@@ -56,15 +56,26 @@ func TestRepositoryContent_GetContent(t *testing.T) {
 		r := RepositoryContent{Encoding: tt.encoding, Content: tt.content}
 		got, err := r.GetContent()
 		if err != nil && !tt.wantErr {
-			t.Errorf("RepositoryContent(%q, %q) returned unexpected error: %v", tt.encoding, tt.content, err)
+			t.Errorf("RepositoryContent(%s, %s) returned unexpected error: %v",
+				stringOrNil(tt.encoding), stringOrNil(tt.content), err)
 		}
 		if err == nil && tt.wantErr {
-			t.Errorf("RepositoryContent(%q, %q) did not return unexpected error", tt.encoding, tt.content)
+			t.Errorf("RepositoryContent(%s, %s) did not return unexpected error",
+				stringOrNil(tt.encoding), stringOrNil(tt.content))
 		}
 		if want := tt.want; got != want {
 			t.Errorf("RepositoryContent.GetContent returned %+v, want %+v", got, want)
 		}
 	}
+}
+
+// stringOrNil converts a potentially null string pointer to string.
+// For non-nil input pointer, the returned string is enclosed in double-quotes.
+func stringOrNil(s *string) string {
+	if s == nil {
+		return "<nil>"
+	}
+	return fmt.Sprintf("%q", *s)
 }
 
 func TestRepositoriesService_GetReadme(t *testing.T) {
--- a/github/misc_test.go
+++ b/github/misc_test.go
@@ -147,6 +147,7 @@ func TestAPIMeta(t *testing.T) {
 		Git:      []string{"g"},
 		Pages:    []string{"p"},
 		Importer: []string{"i"},
+
 		VerifiablePasswordAuthentication: Bool(true),
 	}
 	if !reflect.DeepEqual(want, meta) {


gmlewis:
Since this affects only a test, why not just change the `%q`s to `%v`s?

dmitris:
`%v` does not show the string, only the pointer address - I thought it would be useful to see the actual string value (if not nil) in the error message.  Let me know if you prefer to change it to `%v`, I can change it.

gmlewis:
Oh, right... doh.  It seems like we could come up with a solution that has less repetition... like taking advantage of the same logic in the `github-accessors.go` file... for example:


```go
...
got, err := r.GetContent()

stringOrNil := func(s *string) string {
  if s == nil {
    return "<nil>"
  }
  return *s
}
encoding := stringOrNil(tt.encoding)
content := stringOrNil(tt.content)
...
```

What do you think?

dmitshur:
We'll need to think about a better long term way to deal with gofmt changes.

I think the best future proof solution is to have the `diff -u <(echo -n) <(gofmt -d -s .)` check run only for one of the Go versions, not all of them. It might not be possible to make a single .go file compatible with gofmt of many Go versions (or possible by sacrificing readability).

dmitris:
@gmlewis - changed to use `stringOrNil` as you suggested

github/repos_contents_test.go
```diff
@@ -55,11 +55,19 @@ func TestRepositoryContent_GetContent(t *testing.T) {
...
 		r := RepositoryContent{Encoding: tt.encoding, Content: tt.content}
 		got, err := r.GetContent()
+		stringOrNil := func(s *string) string {
+			if s == nil {
+				return "<nil>"
+			}
+			return *s
+		}
+		encoding := stringOrNil(tt.encoding)
+		content := stringOrNil(tt.content)
```
dmitshur: I don't think it's appropriate to add this much code in between `got, err := ...` and the `if err != nil ...` statements while keeping the variable name just `err` (see [here](https://dmitri.shuralyov.com/idiomatic-go#error-variable-naming)). A future reader may think there's a missing error check after the `got, err := ...` line.

I would suggest factoring out the definition of `stringOrNil` to the very top of `TestRepositoryContent_GetContent`, or even outside to the package scope, below `TestRepositoryContent_GetContent`. And the `stringOrNil(...)` calls can just be inlined into the corresponding `t.Errorf`.

dmitris: agree, done with 2c7b329

github/repos_contents_test.go
```diff
@@ -56,17 +56,27 @@ func TestRepositoryContent_GetContent(t *testing.T) {
...
 		}
 	}
 }
 
+// stringOrNil converts a potentially null string pointer to string
+func stringOrNil(s *string) string {
+	if s == nil {
+		return "<nil>"
+	}
+	return *s
```
gmlewis: random musing - what about making this line `return fmt.Sprintf("%q", *s)` to preserve the quoting?

github/repos_contents_test.go
```diff
@@ -56,17 +56,27 @@ func TestRepositoryContent_GetContent(t *testing.T) {
...
+			t.Errorf("RepositoryContent(%s, %s) did not return unexpected error",
+				stringOrNil(tt.encoding), stringOrNil(tt.content))
 		}
 		if want := tt.want; got != want {
 			t.Errorf("RepositoryContent.GetContent returned %+v, want %+v", got, want)
 		}
 	}
 }
 
+// stringOrNil converts a potentially null string pointer to string
```
gmlewis: Please add period to end of sentence.

dmitris: done

github/repos_contents_test.go
dmitris: if we want the function call to have quotes around the parameters as in a "real" one, I think we should add them to the `t.Errorf`'s format, and change the surrounding quotes markers: 
```
		if err == nil && tt.wantErr {
			t.Errorf(`RepositoryContent("%s", "%s") did not return unexpected error`,
				stringOrNil(tt.encoding), stringOrNil(tt.content))
		}
```
- this would allow to keep `stringOrNil` doing straightforward conversion to string or "<nil>" without adding additional formatting (surrounding quotes).

gmlewis: No, this is not correct because `<nil>` would then be quoted like `"<nil>"` which is not right.
It should either be `<nil>` or `"string-value"`.
But no big deal... we can ignore it.

dmitris: you are right; changed `stringOrNil` to use `fmt.Sprintf("%q", *s)` for quoting

dmitris:
@gmlewis - could you merge this, or do you need the second approval?  Thanks.

gmlewis:
@dmitris - Sorry for the delay. I think this is good-to-go and doesn't need a second LGTM now.
Merging.

PR #955, opened 2018-07-21
Remove unnecessary index.html from a #compare-two-commits url

shuheiktgw:
I always try finding the right go-github method by searching the url of a compatible GitHub API Document URL. When I searched `https://developer.github.com/v3/repos/commits/#compare-two-commits`, I found the same URL is specified as `https://developer.github.com/v3/repos/commits/index.html#compare-two-commits` in go-github GoDoc.

I personally think the former one would be better than later one since the url people normally encounter on GitHub API Document is the former one. So I changed it! Thanks! 👍


shuheiktgw:
I signed it!

PR #965, opened 2018-08-05
Installation: Use *Timestamp instead of *time.Time for the benefit of InstallationEvent

bmoylan:
According to https://developer.github.com/v3/activity/events/types/#installationevent the `created_at` and `updated_at` timestamps in the event payloads are unix-formatted, while the CRUD APIs use RFC3339. Using `*Timestamp` handles the unfortunate fact that the API is inconsistent. Without this change, my application emits the following error
```
failed to unmarshal PullRequestEvent: parsing time "1533435666" as ""2006-01-02T15:04:05Z07:00"": cannot parse "1533435666" as """
```

I understand this might be a breaking change to the public API, so I'm open to feedback if there are alternative ways to proceed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/go-github/965)

### DIFF
--- a/github/apps.go
+++ b/github/apps.go
@@ -57,8 +57,8 @@ type Installation struct {
 	RepositorySelection *string                  `json:"repository_selection,omitempty"`
 	Events              []string                 `json:"events,omitempty"`
 	Permissions         *InstallationPermissions `json:"permissions,omitempty"`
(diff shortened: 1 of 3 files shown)


bmoylan:
@nmiyake for SA

gmlewis:
Thank you, @bmoylan and @juliaferraioli!
Merging.

Issue #884, opened 2018-03-29
Support preview Organization Project Permissions API


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2018-03-28-organization-project-permissions/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue, and we can send you an invite to contribute to the repo (which you then accept after you enable two-factor authentication) and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

ganeshmannamal:
hello @gmlewis I would like to contribute to this issue. Thanks

gmlewis:
Thank you, @ganeshmannamal! I've sent you an invite.

ganeshmannamal:
thanks @gmlewis accepted

gmlewis:
Thanks, @palash25. I'm assigning this issue to you. If, during your investigation of this PR, you find there is nothing left to do, please report back with your findings.

palash25:
@gmlewis need a little help, I am confused about in which files to write the code for this issue. For the 
Organization project collaborators endpoints (which use `projects` endpoint) should the code reside in `orgs_projects.go` or `projects.go`.

I have the same confusion with  Organization projects on teams endpoints would the code go int `teams.go` or `orgs_projects.go`

gmlewis:
Hi @palash25, sure.

But first, can you please elaborate on which exact endpoints need to be written?
Part of resolving this issue is to investigate and determine which ones we still need.

We attempt to keep the code in this package organized as closely as possible to the GitHub Developer API documentation.

So, for example, you can see the endpoints in `orgs_projects.go`, `projects.go`, and `teams.go` we currently have:

```
$ grep fmt.Sprintf github/orgs_projects.go
	u := fmt.Sprintf("orgs/%v/projects", org)
	u := fmt.Sprintf("orgs/%v/projects", org)
$ grep fmt.Sprintf github/projects.go
	u := fmt.Sprintf("projects/%v", id)
	u := fmt.Sprintf("projects/%v", id)
	u := fmt.Sprintf("projects/%v", id)
	u := fmt.Sprintf("projects/%v/columns", projectID)
	u := fmt.Sprintf("projects/columns/%v", id)
	u := fmt.Sprintf("projects/%v/columns", projectID)
[...]

PR #969, opened 2018-08-09
projects: Add missing fields to struct

radeksimko:
The original motivation was to add Project's `HtmlURL` field as we need it in https://github.com/terraform-providers/terraform-provider-github/pull/111
but I took some time to look at all missing fields in both Project & ProjectColumn, just for completeness.

The PR covers only one, small part of https://github.com/google/go-github/issues/884 - there is still more work to do before #884 can actually be closed - specifically around Project Collaborators and Team Projects.

### DIFF
--- a/github/projects.go
+++ b/github/projects.go
@@ -19,15 +19,18 @@ type ProjectsService service
 
 // Project represents a GitHub Project.
 type Project struct {
-	ID        *int64     `json:"id,omitempty"`
-	URL       *string    `json:"url,omitempty"`
-	OwnerURL  *string    `json:"owner_url,omitempty"`
-	Name      *string    `json:"name,omitempty"`
-	Body      *string    `json:"body,omitempty"`
-	Number    *int       `json:"number,omitempty"`
-	CreatedAt *Timestamp `json:"created_at,omitempty"`
-	UpdatedAt *Timestamp `json:"updated_at,omitempty"`
-	NodeID    *string    `json:"node_id,omitempty"`
+	ID         *int64     `json:"id,omitempty"`
+	URL        *string    `json:"url,omitempty"`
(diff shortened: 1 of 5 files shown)


github/projects.go
```diff
@@ -74,6 +77,15 @@ type ProjectOptions struct {
 
 	// State of the project. Either "open" or "closed". (Optional.)
 	State string `json:"state,omitempty"`
```
gmlewis: I believe all three of these optional parameters need to be pointers.
I think I made a mistake in the review of #746 and that we should have used `State *string` here, not `State string`. Can you please change that?

Also, please make the next two parameters pointers as well.

github/projects.go
```diff
@@ -19,15 +19,18 @@ type ProjectsService service
...
-	OwnerURL  *string    `json:"owner_url,omitempty"`
-	Name      *string    `json:"name,omitempty"`
-	Body      *string    `json:"body,omitempty"`
-	Number    *int       `json:"number,omitempty"`
-	CreatedAt *Timestamp `json:"created_at,omitempty"`
-	UpdatedAt *Timestamp `json:"updated_at,omitempty"`
-	NodeID    *string    `json:"node_id,omitempty"`
+	ID         *int64     `json:"id,omitempty"`
+	URL        *string    `json:"url,omitempty"`
+	HtmlURL    *string    `json:"html_url,omitempty"`
```
gmlewis: Let's be consistent with the rest of the naming in this repo and call this `HTMLURL` (for better or worse).

radeksimko:
@gmlewis Thanks for the review. I believe I addressed all your comments. PTAL.

radeksimko:
@gmlewis done, PTAL

PR #970, opened 2018-08-09
Add NodeID to a few structs

parkr:
Based on the `check_run` and `push` events' documentation.

- [x] App
- [x] Organization
- [x] Repository
- [x] User
- [x] PushEventRepository
- [ ] Add PushEventRepository.Owner.NodeID (can use type User, or add NodeID to PushEventRepoOwner)

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -5628,14 +5628,6 @@ func (o *Organization) GetNodeID() string {
 	return *o.NodeID
 }
 
-// GetNodeID returns the NodeID field if it's non-nil, zero value otherwise.
-func (o *Organization) GetNodeID() string {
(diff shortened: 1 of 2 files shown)


parkr:
`PushEventRepository.Owner` is presently of type `PushEventRepoOwner`. It has a `name` and `email`. Looking at [the payload](https://developer.github.com/v3/activity/events/types/#pushevent), `.repository.owner` looks like a full `User`. Changing the type from `PushEventRepoOwner` to `User` is one option. Another is to add the `NodeID` field I need to `PushEventRepoOwner`. The former is a breaking change (easily remedied), the latter may mislead folks since the `.pusher` only has `name` and `email`. 🤷‍♂️ What do you think?

PR #978, opened 2018-08-15
Fix RequestCheckSuite

jnewland:
Previously, `RequestCheckSuite` appended the contents of the `ops RequestCheckSuiteOptions` parameter as query string parameters, resulting in requests like `https://api.github.com/repos/$owner/$repo/check-suite-requests?HeadSHA=$sha`. These sorts of requests result in `422 Invalid request` errors from the GitHub API.

This PR passes the contents of the `ops RequestCheckSuiteOptions` parameter through to `(c *Client) NewRequest` as the body instead, resulting in requests that GitHub's API accepts.

### DIFF
--- a/github/checks_test.go
+++ b/github/checks_test.go
@@ -467,7 +467,7 @@ func TestChecksService_RequestCheckSuite(t *testing.T) {
 	mux.HandleFunc("/repos/o/r/check-suite-requests", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "POST")
 		testHeader(t, r, "Accept", mediaTypeCheckRunsPreview)
-		testBody(t, r, "{\"head_sha\":\"deadbeef\"}"+"\n")
+		testBody(t, r, `{"head_sha":"deadbeef"}`+"\n")
 	})
 	opt := RequestCheckSuiteOptions{
 		HeadSHA: "deadbeef",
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @jnewland!  Your changes LGTM.
When you sign the CLA we can proceed with merging this change.

jnewland:
> When you sign the CLA we can proceed with merging this change.

Sorry for the delay here. I've signed the CLA and this should be good to go! 🖊

github/checks_test.go
```diff
@@ -467,6 +467,7 @@ func TestChecksService_RequestCheckSuite(t *testing.T) {
 	mux.HandleFunc("/repos/o/r/check-suite-requests", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "POST")
 		testHeader(t, r, "Accept", mediaTypeCheckRunsPreview)
+		testBody(t, r, "{\"head_sha\":\"deadbeef\"}"+"\n")
```
gmlewis: Sorry that I didn't see this before, but we can simplify this line and make it easier to read:
```
testBody(t, r, `{"head_sha":"deadbeef"}`+"\n")
```

jnewland: Great suggestion! Applied in f9cebb7.

gmlewis:
Very nice. Thanks, @jnewland!
Merging.

PR #979, opened 2018-08-16
Add FindRepositoryInstallationByID to AppsService

gudmundur:
With a GitHub App that I'm building, I am only ever persisting IDs rather than `owner` and `name` for repositories. For the most part I am using the GraphQL API, but I need to create an installation access token, which I'm using `go-github` for.

This change adds a `FindRepositoryInstallationByID` endpoint which mirrors how `RepositoryService#GetByID` works.


gudmundur:
Thank you for the quick turnaround on this @gmlewis. 🎊

PR #983, opened 2018-08-19
Added support for Dismissed Review (#827)

### LINKED ISSUE #827: Support dismissed_review IssueEvent
https://developer.github.com/changes/2018-01-05-dismiss-event-breaking-change/

This announcement actually confuses me a bit...
From the announcement, the `dismissed_review` IssueEvent is changing, but I don't see any history of it, so I don't know when it was introduced. Nevertheless, we would like to support it, so it needs to be added to the `IssueEvent` struct in `github/issues_events.go`.

I think it would suffice to leave its type as a `string` field, similar to other states returned by this package (as opposed to strictly enumerating the values), unless there are objections.

This would be a great first issue for a new contributor to this repo or new Go developer to work on.
All help is greatly appreciated. Thank you!

mrsimonfletcher:
**Disclaimer:** This is my first real use of the json package and first OS pull so please bear with me 😊

This commit simply adds support for a dismissed_review issue event. I
have added a new struct type called DismissedReview. Additional to this
I have added this type as a a field against the IssueEvent.

Example:

```
func FetchIssueEvents(username, repo string, issueId int) ([]*github.IssueEvent, error) {
	events, _, err := client.Issues.ListIssueEvents(context.Background(), username, repo, issueId, nil)
	return events, err
}

func main() {
	issueEvents, err := FetchIssueEvents("mrsimonfletcher", "test", 4)
	if err != nil {
		return
	}

	for _, v := range issueEvents {
		if *v.Event == "review_dismissed" {
			fmt.Println(*v.DismissedReview)
			fmt.Println(*v.DismissedReview.State)
			fmt.Println(*v.DismissedReview.ReviewID)
			fmt.Println(*v.DismissedReview.DismissalCommitID)
		}
	}
}

```
**Questions:**

- I tested this manually through including the package locally, looking at the test suite there doesn't seems to be existing checks for individual fields on the struct. Is there a test that I should be adding updating with this change?

- `DismissalCommitID` is a conditional field and does not return though the API if it isn't applicable. Do we need to defensively code against this or should the consumer of `go-github` handle this explicitly?

- `DismissalMessage` will return null in some instances this causes the program to panic if you try to access the field when it's null. Do we need to defensively code against this or should the consumer of `go-github` handle this explicitly?

<details>
 <summary>Example response of when the dismissal_commit_id field is ommited</summary>

```json
[
    {
        "id": 1795889201,
        "node_id": "MDIwOlJldmlld0Rpc21pc3NlZEV2ZW50MTc5NTg4OTIwMQ==",
        "url": "https://api.github.com/repos/mrsimonfletcher/test/issues/events/1795889201",
        "actor": {
            "login": "mrsimonfletcher",
            "id": 1643158,
            "node_id": "MDQ6VXNlcjE2NDMxNTg=",
            "avatar_url": "https://avatars1.githubusercontent.com/u/1643158?v=4",
            "gravatar_id": "",
            "url": "https://api.github.com/users/mrsimonfletcher",
            "html_url": "https://github.com/mrsimonfletcher",
            "followers_url": "https://api.github.com/users/mrsimonfletcher/followers",
            "following_url": "https://api.github.com/users/mrsimonfletcher/following{/other_user}",
            "gists_url": "https://api.github.com/users/mrsimonfletcher/gists{/gist_id}",
            "starred_url": "https://api.github.com/users/mrsimonfletcher/starred{/owner}{/repo}",
            "subscriptions_url": "https://api.github.com/users/mrsimonfletcher/subscriptions",
            "organizations_url": "https://api.github.com/users/mrsimonfletcher/orgs",
            "repos_url": "https://api.github.com/users/mrsimonfletcher/repos",
            "events_url": "https://api.github.com/users/mrsimonfletcher/events{/privacy}",
            "received_events_url": "https://api.github.com/users/mrsimonfletcher/received_events",
            "type": "User",
            "site_admin": true
        },
        "event": "review_dismissed",
        "commit_id": null,
        "commit_url": null,
        "created_at": "2018-08-18T07:45:00Z",
        "dismissed_review": {
            "state": "changes_requested",
            "review_id": 147422943,
            "dismissal_message": "You are incorrect"
        }
    }
]
```
</details>

<details>
 <summary>Example response of when the dismissal_message is null</summary>

```json
[
    {
        "id": 1796257888,
        "node_id": "MDIwOlJldmlld0Rpc21pc3NlZEV2ZW50MTc5NjI1Nzg4OA==",
        "url": "https://api.github.com/repos/mrsimonfletcher/test/issues/events/1796257888",
        "actor": {
            "login": "mrsimonfletcher",
            "id": 1643158,
            "node_id": "MDQ6VXNlcjE2NDMxNTg=",
            "avatar_url": "https://avatars1.githubusercontent.com/u/1643158?v=4",
            "gravatar_id": "",
            "url": "https://api.github.com/users/mrsimonfletcher",
            "html_url": "https://github.com/mrsimonfletcher",
            "followers_url": "https://api.github.com/users/mrsimonfletcher/followers",
            "following_url": "https://api.github.com/users/mrsimonfletcher/following{/other_user}",
            "gists_url": "https://api.github.com/users/mrsimonfletcher/gists{/gist_id}",
            "starred_url": "https://api.github.com/users/mrsimonfletcher/starred{/owner}{/repo}",
            "subscriptions_url": "https://api.github.com/users/mrsimonfletcher/subscriptions",
            "organizations_url": "https://api.github.com/users/mrsimonfletcher/orgs",
            "repos_url": "https://api.github.com/users/mrsimonfletcher/repos",
            "events_url": "https://api.github.com/users/mrsimonfletcher/events{/privacy}",
            "received_events_url": "https://api.github.com/users/mrsimonfletcher/received_events",
            "type": "User",
            "site_admin": true
        },
        "event": "review_dismissed",
        "commit_id": null,
        "commit_url": null,
        "created_at": "2018-08-19T08:12:00Z",
        "dismissed_review": {
            "state": "approved",
            "review_id": 147450868,
            "dismissal_message": null,
            "dismissal_commit_id": "0cb225efb81ffdb35032b6148859183cd70f53aa"
        }
    }
]

```
</details>

closes #827

### DIFF
--- a/github/issues_events.go
+++ b/github/issues_events.go
@@ -44,6 +44,10 @@ type IssueEvent struct {
 	//     renamed
 	//       The Actor changed the issue title from Rename.From to Rename.To.
 	//
+	//     dismissed_review
+	//       The Actor dismissed a review.
+	//       DismissedReview holds the state, review_id, dismissal_message (if applicable) & dismissal_commit_id (if applicable)
+	//
 	//     mentioned
 	//       Someone unspecified @mentioned the Actor [sic] in an issue comment body.
 	//
@@ -68,13 +72,14 @@ type IssueEvent struct {
 	Issue     *Issue     `json:"issue,omitempty"`
 
 	// Only present on certain events; see above.
-	Assignee   *User      `json:"assignee,omitempty"`
-	Assigner   *User      `json:"assigner,omitempty"`
-	CommitID   *string    `json:"commit_id,omitempty"`
-	Milestone  *Milestone `json:"milestone,omitempty"`
-	Label      *Label     `json:"label,omitempty"`
-	Rename     *Rename    `json:"rename,omitempty"`
-	LockReason *string    `json:"lock_reason,omitempty"`
+	Assignee        *User            `json:"assignee,omitempty"`
+	Assigner        *User            `json:"assigner,omitempty"`
+	CommitID        *string          `json:"commit_id,omitempty"`
+	Milestone       *Milestone       `json:"milestone,omitempty"`
+	Label           *Label           `json:"label,omitempty"`
+	Rename          *Rename          `json:"rename,omitempty"`
+	LockReason      *string          `json:"lock_reason,omitempty"`
+	DismissedReview *DismissedReview `json:"dismissed_review,omitempty"`
 }
 
 // ListIssueEvents lists events for the specified issue.
@@ -154,5 +159,17 @@ type Rename struct {
 }
 
 func (r Rename) String() string {
+	return Stringify(r)
+}
+
+// DismissedReview contains details for 'dismissed_review' events
+type DismissedReview struct {
+	State             *string `json:"state,omitempty"`
+	ReviewID          *int    `json:"review_id,omitempty"`
+	DismissalMessage  *string `json:"dismissal_message,ommitempty"`
+	DismissalCommitID *string `json:"dismissal_commit_id,omitempty"`
+}
+
+func (r DismissedReview) String() string {
 	return Stringify(r)
 }
--- a/AUTHORS
+++ b/AUTHORS
@@ -154,6 +154,7 @@ Sebastian Mæland Pedersen <sem.pedersen@stud.uis.no>
 Sevki <s@sevki.org>
 Shawn Catanzarite <me@shawncatz.com>
 Shawn Smith <shawnpsmith@gmail.com>
+Simon Fletcher <simonfletcher0@gmail.com>
 sona-tar <sona.zip@gmail.com>
 SoundCloud, Ltd.
 Stian Eikeland <stian@eikeland.se>


github/issues_events.go
```diff
@@ -156,3 +161,15 @@ type Rename struct {
 func (r Rename) String() string {
 	return Stringify(r)
 }
+
+// DismissedReview contains details for 'dismissed_review' events
+type DismissedReview struct {
+	State             *string `json:"state,omitempty"`
```
gmlewis: Please enumerate the possible values of `state` here, found at:
https://developer.github.com/v3/issues/events/#attributes

github/issues_events.go
```diff
@@ -156,3 +161,15 @@ type Rename struct {
...
+	ReviewID          *int    `json:"review_id,omitempty"`
```
gmlewis: Since this is an ID, let's please make its type `*int64`.

github/issues_events.go
mrsimonfletcher: I can see this being done in the code by via comments:

https://github.com/google/go-github/blob/d7732128a00e8e95e8fe896017da18ee20b2180d/github/checks.go#L139

I am wondering if you think this would suffice or whether we should add a custom `UnmarshalJSON` method here?

mrsimonfletcher:
@gmlewis thanks for the reply here

> So to answer your question, yes, it would be nice to have you add tests that exercise the marshaling and unmarshaling of your new fields. It also provides a good example for others as to their common usage.

I agree, I will look to add these. After looking at the `issues_events_test.go` file I can see where these tests are getting done but they are not testing the whole data structure. I can see why I missed this now. 

https://github.com/google/go-github/blob/d7732128a00e8e95e8fe896017da18ee20b2180d/github/issues_events_test.go#L80

Personally, I think that we should update this test file to be more of a true representation of the data structure. Adding these is probably outside the scope of this PR as it is more a test refactor than specific to this issue.

I am happy to make these changes in a separate branch first, we can then merge this branch and continue with this work WDYT?

> Speaking of this, your PR causes tests to fail. Please see the instructions in CONTRIBUTING.md and run the required commands

Absolutely, sorry I must have missed some stuff here, I will do this.

github/issues_events.go
gmlewis: Hmmm... I'm not sure I understand what you are saying here.

I see that line in another file, but I'm feeling that that is unrelated... and I'm not sure why we are talking about `UnmarshalJSON` here at all.

I was just asking for a comment for the `State` field... something like this:

```go
// State represents the state of the dismissed review and can be one of: ...
State             *string `json:"state,omitempty"`
```

gmlewis:
Yes, a separate PR to address updating other tests would be good.  Thanks.
Let's keep this PR focused on the smaller task.

github/issues_events.go
mrsimonfletcher: > I was just asking for a comment for the State field... something like this

Sorry yes, I thought this is what you meant, this was the reasoning for my snippet. Although it wasn't clear, the comment was what I was trying to communicate with the snipped 😆.

> and I'm not sure why we are talking about UnmarshalJSON here at all

This was due to my misunderstanding, I wasn't sure if you wanted a comment like the above or we wanted to say that the field can be one of X values, and so we would need to permit those values instead of allowing any value to come back.

Anyway, it's clear that we just need a comment here so I will add. Thanks 👍

gmlewis:
Closing obsolete PR.

PR #985, opened 2018-08-26
Introduce go modules

marwan-at-work:
This PR is generated by a [tool I wrote](https://github.com/marwan-at-work/mod) that helps migrate Go packages that are tagged +2 to Semantic Import Versioning. 

Note that this change should be backwards compatible and can still be imported without the `/vX` suffix when Go Modules is off.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: go
 go:
+  - "1.11.x"
   - "1.10.x"
   - "1.9.x"
   - "1.8.x"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,13 @@ go:
   - "1.11.x"
   - "1.10.x"
   - "1.9.x"
-  - "1.8.x"
-  - master
+  - tip
+env:
+  global:
+    - GO111MODULE=on
 matrix:
   allow_failures:
-    - go: master
+    - go: tip
   fast_finish: true
 env:
   secure: "IrnPmy/rkIP6Nrbqji+u7MCAibQlA6WvPLEllmDQ2yZP/uIe3wLwwYbTu9BOkgzoLA+f8PA6u3pp/RhY/rtaM4NzHAO2nVfGIv9UHUQ3NGq0DYS6rODjVKhq7vkhELoagRewyqFVN4rE0LnExkknRMgjQfRke6/DA7u7Xm8JyhY=" # COVERALLS_TOKEN
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ If you're interested in using the [GraphQL API v4][], the recommended library is
 ## Usage ##
 
 ```go
-import "github.com/google/go-github/github"
+import "github.com/google/go-github/github/v17"
 ```
 
 Construct a new GitHub client, then use the various services on the client to
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 go-github is a Go client library for accessing the [GitHub API v3][].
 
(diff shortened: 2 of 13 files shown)


marwan-at-work:
@gmlewis seems that gofmt 1.11 has different rules from 1.10 and 1.9?

gmlewis:
Thanks for fixing the Travis CI build, @marwan-at-work!

Since we recently removed support for Go 1.7 and bumped the version number, and this PR now removes support for Go 1.8, can you please make this semantic version be v18.0.0 and we will tag it as such when we merge the PR?

Also, can you please make the necessary change to `README.md` similar to the one I did here: https://github.com/google/go-github/commit/fbd7fb9c7f3d27aa691e8421690e00be916f8d8d
?

gmlewis:
To answer your question about investigating differences in `gofmt` between `v1.11` and `v1.10`, it seems pretty minor and I'm fine to move forward keeping the builds happy, but if you would like to investigate, you are welcome to, obviously.

gmlewis:
Thank you, @juliaferraioli!
Merging then bumping tagged release.

marwan-at-work:
Thank you both :)

tmc:
I think `github.com/google/go-github/github/v18` in the readme is supposed to be `github.com/google/go-github/v18/github`.

tmc:
Perhaps the README should contain instructions for pre-go1.11 users as well?

marwan-at-work:
As for pre go 1.11, the latest Go 1.9 and 1.10 should work fine with the new import path. However they wouldn't work with Go 1.8 or before that. So before 1.8, they can just use v17 or below. But maybe the readme should just say the package requires Go1.9.X or above?

gmlewis:
...and it does:

`go-github requires Go version 1.9 or greater.`

.travis.yml
```diff
@@ -1,13 +1,16 @@
 sudo: false
 language: go
 go:
+  - "1.11.x"
   - "1.10.x"
   - "1.9.x"
-  - "1.8.x"
-  - master
+  - tip
```
dmitshur: @marwan-at-work Can you please help me understand why "master" was replaced with "tip"? I tried reading the PR description and commit messages, and didn't find a rationale there.

The latest [Travis CI documentation](https://docs.travis-ci.com/user/languages/go/#specifying-a-go-version-to-use) says:

> or use `master` to get the latest version from source

That's why this change is unexpected for me and I'd like to understand it better.

README.md
```diff
@@ -4,15 +4,15 @@
...
 
 If you're interested in using the [GraphQL API v4][], the recommended library is
 [shurcooL/githubv4][].
 
 ## Usage ##
 
 ```go
-import "github.com/google/go-github/github"
+import "github.com/google/go-github/github/v18"
 ```
```
dmitshur: Given that modules were released very recently and are still experimental/optional, perhaps it's a good idea to expand this to have two sections: one for traditional GOPATH workspaces (with `import "github.com/google/go-github/github"`), and another section for module-aware environments (with `import "github.com/google/go-github/v18/github"`).

marwan-at-work: @dmitshur yes, and probably should include `go get` part too?

.travis.yml
marwan-at-work: @dmitshur im not sure but it failed for me and switched to tip because we use `tip` at athens. I don't know the difference, but maybe we can open a pr and check?

gmlewis: I can do a quick experiment.

gmlewis: See #1002 for successful run with `master`.

PR #986, opened 2018-08-30
CheckResponse: Allow reuse of response body

radeksimko:
I tried using `CheckResponse` in a custom transport to identify certain errors and implement conditional retries and I kept bumping into a weird error

```
http: ContentLength=157 with Body length 0
```

from another transport I use to log all HTTP requests & responses (incl. bodies).

I noticed (after some debugging) that `ioutil.ReadAll` is in use, which closes the stream and prevents other readers from reading the body.

Attached is a test reproducing described behaviour.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -698,6 +698,9 @@ func CheckResponse(r *http.Response) error {
 	if err == nil && data != nil {
 		json.Unmarshal(data, errorResponse)
 	}
+
+	r.Body = ioutil.NopCloser(bytes.NewReader(data))
+
 	switch {
 	case r.StatusCode == http.StatusUnauthorized && strings.HasPrefix(r.Header.Get(headerOTP), "required"):
 		return (*TwoFactorAuthError)(errorResponse)
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -858,6 +858,29 @@ func TestCheckResponse_errorBody(t *testing.T) {
 	}
 }
 
+func TestGetRepository_errorBody(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/test/test", func(w http.ResponseWriter, req *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprintln(w, `{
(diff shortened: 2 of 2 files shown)


gmlewis:
I'm not convinced this is a safe change.

In https://godoc.org/net/http it says: `The client must close the response body when finished with it`.

I've instrumented this package many times to dump out full requests and responses without having to use this technique. 

Have you taken a look at the methods found in https://godoc.org/net/http/httputil ? I have found those to be incredibly handy.

Additionally, you can make yourself a `RoundTripper` described here: https://godoc.org/net/http#RoundTripper
that can intercept each request and dump any info you might need before and after the request.

I think these alternatives are much safer than this change.

I'll leave this PR open for a few days in case there is any more discussion, but unless I hear a strong argument for why this is needed, I don't think we want to go this route as it can potentially cause a lot of heartache if clients do not close the body of the responses.

radeksimko:
> I've instrumented this package many times to dump out full requests and responses without having to use this technique.

So have I. It worked just fine until I called `CheckResponse` explicitly. Have you looked at the attached test which demonstrates the problem? Just remove that one line and you should see the test failing.

> Additionally, you can make yourself a RoundTripper described here:  https://godoc.org/net/http#RoundTripper
> that can intercept each request and dump any info you might need before and after the request.

That is exactly [what I did](https://github.com/hashicorp/terraform/blob/master/helper/logging/transport.go) as mentioned in my initial post 😉 

While I see your point, [`httputil.DumpResponse`](https://golang.org/src/net/http/httputil/dump.go?s=8166:8231#L271) does exactly what I did here - reads the response body and then resets it w/ `ioutil.NopCloser`, so either we're misinterpreting the comment `The client must close the response body when finished with it` or these two packages are contradicting each other. 🤔

gmlewis:
Oh, I see what you are saying now.  Thanks for the detailed explanation.

OK, I would like to get some discussion going on how users of this package address this situation.

This is an open invitation to all users and contributors to add your thoughts to this PR. Thanks.

gmlewis:
There is no discussion on this PR, so I'll take that as no interest in it.
Closing, but we can reopen if there is demand.

PR #987, opened 2018-08-30
remove v18 postfix

coolaj86:
`go get github.com/google/go-github/github/v18` failed.

`go get github.com/google/go-github/github` worked.

So I changed the import url from "github.com/google/go-github/github/v18" to "github.com/google/go-github/github"

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ If you're interested in using the [GraphQL API v4][], the recommended library is
 ## Usage ##
 
 ```go
-import "github.com/google/go-github/github/v18"
+import "github.com/google/go-github/github"
 ```
 
 Construct a new GitHub client, then use the various services on the client to


gmlewis:
@marwan-at-work - you had added this change in #985 and explained that it was backward compatible.
I think we need to discuss what is going on here.
@coolaj86 - can you please explain why you think this change is needed?

marwan-at-work:
@gmlewis I believe I just had a typo in the import path (fixed with PR 989) -- Yes the new import path should be backwards compatible with the latest Go 1.9, 1.10 and 1.11. But @coolaj86 try importing `github.com/google/go-github/v18/github`

marwan-at-work:
@coolaj86: 
1. You must do `go get github.com/google/go-github/v18/github` and **not** `go get github.com/google/go-github/github/v18` -- Like I mentioned this was a typo in the README. 

2. In your second example, it seems you're doing the correct `go get` but it's still not working for you. This is because if you have GO111MODULE turned off, then you must use the old go get (no v18 in the import path). But if you're inside a go.mod path, then you must have GO111MODULE turned on and must include v18 in the import path. 

For example, running this in a directory with a valid go.mod file works:

```
$ go get github.com/google/go-github/v18/github@latest
go: finding github.com/google/go-github/v18/github latest

$ cat go.mod

module example

require (
	github.com/google/go-github/v18 v18.0.0 // indirect
)
```


3. The support was removed for Go 1.8, and _not_ go-github v18.0.0 

I think this is good feedback on workflow changes to the Go team because it's probably confusing to know when to use GO111MODLE and when not to even though it's backwards compatible.

coolaj86:
I have no idea what `GO111MODLE` or `go.mod` is. I guess I'll... wait for it... go google it!

coolaj86:
Well, that's embarrassing...

screen shot 2018-08-30 at 11 07 26 pm

marwan-at-work:
@coolaj86 you should probably prepend your search with the keyword "golang" since you're searching the whole internet...

Or just check out the wiki for go modules https://github.com/golang/go/wiki/Modules

gwilym:
I realised this is a closed issue but just wanted to mention that I was hit by this today. I'm using Go 1.11 and the README-advertised import path doesn't function, whereas using just `github.com/google/go-github/github` does:

```
package github.com/google/go-github/v18/github: cannot find package "github.com/google/go-github/v18/github" in any of:
	/usr/local/Cellar/go/1.11/libexec/src/github.com/google/go-github/v18/github (from $GOROOT)
	/Users/gwilym.evans/go/src/github.com/google/go-github/v18/github (from $GOPATH)
```

Should the README perhaps have a mention of which import path to use depending on your module usage? That is, given that module support is still an opt-in experimental feature in Go 1.11.

gmlewis:
Did you check out #989?
Feel free to submit another PR to improve the README.md file.

gwilym:
Yes I did, the import path attempt of mine above reflects the typo correction from that PR. At this point I probably won't submit a README correction since don't know what the updated wording should be. That is, I don't know what the intention behind the import path is and I'm not yet familiar enough with Go's new module features to make a call on how it should be worded to suit all cases.

PR #991, opened 2018-09-04
Add NodeID to CheckRun, CheckSuite

thepwagner:
Per: https://developer.github.com/v3/checks/runs/#get-a-single-check-run / https://developer.github.com/v3/checks/suites/#get-a-single-check-suite


gmlewis:
Thank you, @thepwagner... LGTM.
Merging.

PR #995, opened 2018-09-05
Make ValidateSignature public when the original request is unavailable

StevenACoffman:
In an AWS Lambda invoked via API Gateway, I do not have direct access to the original `*http.Request`, but instead will typically have a signature of:

`func Handler(ctx context.Context, request events.APIGatewayProxyRequest) (Response, error) {`

The `events.APIGatewayProxyRequest` contains the original body as well as headers, and I would like to be able to invoke `ValidateSignature` directly.

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -175,19 +175,19 @@ func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err err
 	}
 
 	sig := r.Header.Get(signatureHeader)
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Closed and re-opened to trigger travis to re-run flaky test.

gmlewis:
That seems like a reasonable use-case. Thank you, @StevenACoffman!
LGTM.
Merging.

PR #996, opened 2018-09-06
add missing org fields

jmcampanini:


jmcampanini:
I signed it!

github/orgs.go
```diff
@@ -19,31 +19,33 @@ type OrganizationsService service
...
+	UpdatedAt             *time.Time `json:"updated_at,omitempty"`
+	TotalPrivateRepos     *int       `json:"total_private_repos,omitempty"`
+	OwnedPrivateRepos     *int       `json:"owned_private_repos,omitempty"`
+	PrivateGists          *int       `json:"private_gists,omitempty"`
+	DiskUsage             *int       `json:"disk_usage,omitempty"`
+	Collaborators         *int       `json:"collaborators,omitempty"`
+	BillingEmail          *string    `json:"billing_email,omitempty"`
+	Type                  *string    `json:"type,omitempty"`
+	Plan                  *Plan      `json:"plan,omitempty"`
+	DefaultRepoPermission *string    `json:"default_repository_permission,omitempty"`
```
gmlewis: Can you please add a comment before this field stating what the possible values are?
Also, can you add a comment to the field below stating that its default value is `true`?
Also, for both fields, please note that they are only used in a call to `OrganizationsService.Edit`.
As such, I would separate them by a blank line.

For example:

```go
Plan ...

// DefaultRepoPermission can be one of: "read", "write", "admin", or "none". (Default: "read").
// It is only used in OrganizationsService.Edit.
DefaultRepoPermission...
// MembersCanCreateRepos default value is true and is only used in Organizations.Edit.
MembersCanCreateRepos...
```

jmcampanini: updated :)

jmcampanini:
much thanks!

PR #997, opened 2018-09-07
Restrict request params of CreateRelease and EditRelease to known subset

### LINKED ISSUE #992: Heads up: stricter validation coming soon to the Update a release API endpoint
Hi 👋 ,

I an engineer on the API team at GitHub. A colleague and I are currently going through GitHub's REST API, tightening up validations.

We noticed that google/go-github is passing undocumented parameters to the [Edit a release](https://developer.github.com/v3/repos/releases/#edit-a-release) endpoint.

In particular, we're receiving the following parameters which the endpoint doesn't know about:

- `assets`
- `assets_url`
- `author`
- `created_at`
- `html_url`
- `id`
- `node_id`
- `published_at`
- `tarball_url`
- `url`
- `zipball_url`

https://github.com/google/go-github/blob/29a3681d054b5329b97878453f84a4ea20e54b49/github/repos_releases.go#L20-L40

Currently the backend code is ignoring unknown parameters, but we're shortly going to change the validation to return a [422 Unprocessable Entity](https://httpstatuses.com/422) if an undocumented parameter is passed.

We're still discussing our timeline for this, but wanted to give you a heads up so you're not caught by surprise. If you have an ETA for the fix and an idea of how long people would typically need in order to update their dependencies, that information would be very useful for us to have.

gmlewis:
Fixes #992.

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -125,13 +125,37 @@ func (s *RepositoriesService) getSingleRelease(ctx context.Context, url string) 
 	return release, resp, nil
 }
 
+// repositoryReleaseRequest is a subset of RepositoryRelease and
+// is used internally by CreateRelease and EditRelease to pass
+// only the known fields for these endpoints.
+//
+// See https://github.com/google/go-github/issues/992 for more
+// information.
+type repositoryReleaseRequest struct {
+	TagName         *string `json:"tag_name,omitempty"`
+	TargetCommitish *string `json:"target_commitish,omitempty"`
+	Name            *string `json:"name,omitempty"`
+	Body            *string `json:"body,omitempty"`
+	Draft           *bool   `json:"draft,omitempty"`
+	Prerelease      *bool   `json:"prerelease,omitempty"`
+}
+
 // CreateRelease adds a new release for a repository.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/releases/#create-a-release
 func (s *RepositoriesService) CreateRelease(ctx context.Context, owner, repo string, release *RepositoryRelease) (*RepositoryRelease, *Response, error) {
 	u := fmt.Sprintf("repos/%s/%s/releases", owner, repo)
 
-	req, err := s.client.NewRequest("POST", u, release)
+	releaseReq := &repositoryReleaseRequest{
+		TagName:         release.TagName,
+		TargetCommitish: release.TargetCommitish,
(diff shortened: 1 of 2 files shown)


github/repos_releases.go
```diff
@@ -125,13 +125,37 @@ func (s *RepositoriesService) getSingleRelease(ctx context.Context, url string)
 	return release, resp, nil
 }
 
+// repositoryReleaseRequest is a subset of RepositoryRelease and
+// is used internally by CreateRelease and EditRelease to pass
+// only the known fields for these endpoints.
+//
+// See https://github.com/google/go-github/issues/992 for more
+// information.
+type repositoryReleaseRequest struct {
```
gauntface: Why not make this public and ask users for this type?

gmlewis:
If we go the breaking change route, then users of the package would no longer be able to call:
* `ListReleases`, 
* `GetRelease`, 
* `GetLatestRelease`, or
* `GetReleaseByTag`

before calling:

* `CreateRelease` (after editing the `Name`) or
* `EditRelease`

but would instead have to perform the steps provided by this PR (namely, copy the other fields). (Actually, for `EditRelease` I bet that the unmodified fields would not truly need copying, and by their absense, their values would not change.)

However, on the flip side, how useful is this?!?

I would really like to take a survey of users of this package who frequently employ these endpoints to find out if the above scenario is even useful... but that's not practical.

If it isn't, then we may want to go the breaking-change route.

However, we now have 3 different votes... @dmitshur votes for [Option 4](https://github.com/google/go-github/issues/992#issuecomment-419678668), @gauntface votes for [Option 1](https://github.com/google/go-github/issues/992#issuecomment-419678668) and I vote for [Option 3](https://github.com/google/go-github/issues/992#issuecomment-419678668). It would be nice to have a tie breaker from people who rely on these endpoints.

gmlewis:
I have rearranged RepositoryRelease and added clarifying comments about its usage.

I believe that this is a sustainable pattern for handling all `422` errors in the future.

* It is initially more work for the package maintainers, but:
* it clearly states what is going on with the parameters, and
* it requires no change to package user's code, and
* it does not further pollute the struct namespace with `YetAnotherRequestStruct`, yet
* it does the right thing for honoring the correct parameters being sent to the backend.

PTAL.

gmlewis:
Thank you, @gauntface and @dmitshur.
Merging.

PR #998, opened 2018-09-07
Fix printing for Response instances.

rspier:
Explicitly specify the Rate type so Rate's String() receiver doesn't propagate to Response.

(Otherwise, `fmt.Printf("%+v", resp)` will just output the embedded Rate instead of the whole Response.)

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -375,7 +375,9 @@ type Response struct {
 	FirstPage int
 	LastPage  int
 
-	Rate
+	// Explicitly specify the Rate type so Rate's String() receiver doesn't
+	// propagate to Response.
+	Rate Rate
 }
 
 // newResponse creates a new Response for the provided http.Response.


gmlewis:
Thank you, @rspier!

In this issue comment thread, can you please show an example before/after so that others can see how this change will affect the Stringer output?

rspier:
Hi @gmlewis!

Here's a simplified version:

**Before:**

```
package main

import (
	"fmt"
	"net/http"
)

type Rate struct{}
type Response struct {
	*http.Response
	Rate
}

func (r Rate) String() string {
	return "I'm a Rate"
}

func main() {
	resp := &Response{Response: &http.Response{}, Rate: Rate{}}
	fmt.Printf("%+v", resp)
}
```

returns

```
I'm a Rate
```

**The fixed version:**

```
package main

import (
	"fmt"
	"net/http"
)

type Rate struct{}
type Response struct {
	*http.Response
	Rate Rate
}

func (r Rate) String() string {
	return "I'm a Rate"
}

func main() {
	resp := &Response{Response: &http.Response{}, Rate: Rate{}}
	fmt.Printf("%+v", resp)
}
```

returns

```
&{Response:0x1074a190 Rate:I'm a Rate}
```

gmlewis:
Hi @rspier!  Thanks!
LGTM.
Merging.

PR #1002, opened 2018-09-11
Use master instead of tip for .travis.yml

gmlewis:
This is an experiment to address this question:
https://github.com/google/go-github/pull/985#discussion_r216491123


gmlewis:
Although I don't see the check on this PR (passing or otherwise, which is odd), I see that the build is successful:
https://travis-ci.org/google/go-github/builds/426983699

gmlewis:
Hehe... now it appears.  Yesterday GitHub showed only "1 successful check" (the `cla/google` one)... very strange.

Thanks, @dmitshur!
Merging.

PR #1003, opened 2018-09-13
add requested_teams to PullRequest

jipperinbham:
`requested_teams` was added to the `pull_request` object in the `review_requested` event, see [payload example](https://developer.github.com/v3/activity/events/types/#webhook-payload-example-25) for definition.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -60,6 +60,10 @@ type PullRequest struct {
 	NodeID              *string    `json:"node_id,omitempty"`
 	RequestedReviewers  []*User    `json:"requested_reviewers,omitempty"`
 
+	// RequestedTeams is populated as part of the PullRequestEvent.
+	// See, https://developer.github.com/v3/activity/events/types/#pullrequestevent for an example.
+	RequestedTeams []*Team `json:"requested_teams,omitempty"`
+
 	Head *PullRequestBranch `json:"head,omitempty"`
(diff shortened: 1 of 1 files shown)


jipperinbham:
Yea, the reference to it is not called out anywhere other than the example payload. You can see it [here](https://developer.github.com/v3/activity/events/types/#webhook-payload-example-25). I'll also send a note to GH support to verify its existence.

gmlewis:
Oh, I see! Thanks for the link.
Then we can proceed with your change... but let's add a comment that this field is populated in the `PullRequestEvent` and point to the URL you provided:
https://developer.github.com/v3/activity/events/types/#pullrequestevent

jipperinbham:
@gmlewis not sure what happened with the go 1.9 test, any ideas?

gmlewis:
It looks like this was caused by a difference in `gofmt` in v1.9.
Try putting a blank line before line 62 and see if that fixes the problem.

gmlewis:
LGTM.
Merging.

jipperinbham:
Thanks!

Also, for clarity, here's what I received from GH Support:

```
The requested_teams object looks to exist for review_requested and review_request_removed events.

Creating a Pull Request without Requesting reviewers, commenting on the Pull Request or Closing the Pull Request would send a payload without the requested_teams object.
```

PR #1017, opened 2018-09-25
Restrict request params of Create to known subset

### LINKED ISSUE #1014: (Another) heads up: stricter validation coming soon in the create repo for org endpoint
A few weeks back we discussed an issue where the GitHub API is going to return 422s when incorrect parameters are passed (rather than just ignoring those parameters): https://github.com/google/go-github/issues/992

We found another instance where we get incorrect parameters with the google/go-github header.

Our current plan is to turn on stricter validation on November 1st (we were going to do this earlier, but we've been seeing a lot of mismatches across the board, and are doing some work to let everyone know first!).

The [Create a repository for an organization](https://developer.github.com/v3/repos/#create) endpoint doesn't take an `archived` parameter, but we're seeing `archived` being passed to a number of endpoints.

Would you have a chance to look into whether this is encoded into the library itself, or whether it might be user error?

gmlewis:
Fixes #1014.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -56,6 +56,7 @@ type Repository struct {
 	AllowSquashMerge *bool            `json:"allow_squash_merge,omitempty"`
 	AllowMergeCommit *bool            `json:"allow_merge_commit,omitempty"`
 	Topics           []string         `json:"topics,omitempty"`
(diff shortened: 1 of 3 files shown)


gmlewis:
Thank you, @gauntface!
Merging.

PR #1018, opened 2018-09-26
Restrict request params of CreateHook for repos and orgs to known subset

### LINKED ISSUE #1015: (Yet another) heads up: Stricter validation coming soon to the create organization hook endpoint
We're working on adding stricter validation in the GitHub REST API, and uncovering all sorts of goodies and inconsistencies. ✨ 

In the [Create organization hook](https://developer.github.com/v3/orgs/hooks/#create-a-hook) endpoint we noticed the following errors come through with the `google/go-github` user agent:

```
'"created_at", "updated_at" are not permitted keys.'
```

Would you have a moment to check if this is encoded into the library, or whether this is a user error, as seen in https://github.com/google/go-github/issues/992 ?

gmlewis:
Fixes #1015.

### DIFF
--- a/github/orgs_hooks.go
+++ b/github/orgs_hooks.go
@@ -51,10 +51,21 @@ func (s *OrganizationsService) GetHook(ctx context.Context, org string, id int64
 // CreateHook creates a Hook for the specified org.
 // Name and Config are required fields.
 //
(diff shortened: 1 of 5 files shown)


gmlewis:
Thank you, @gauntface!
Merging.

PR #1019, opened 2018-09-28
add other missing org field

jmcampanini:
related: https://github.com/google/go-github/pull/996

unfortunately i only added the field that matches `default_repository_settings` which is used to GET the org info. according to the docs (https://developer.github.com/v3/orgs/#get-an-organization):

* `default_repository_settings` is used for getting info about the org, using `Get`
* `default_repository_permission` is used to set the value on the org, using `Edit`

ive named the field `GetDefaultRepoPermissionXyz`, using the `Get` and `Edit` suffix to define the field, and using the correct json deserialization in each instance.

### DIFF
--- a/github/orgs.go
+++ b/github/orgs.go
@@ -46,12 +46,12 @@ type Organization struct {
 	Plan                        *Plan      `json:"plan,omitempty"`
 	TwoFactorRequirementEnabled *bool      `json:"two_factor_requirement_enabled,omitempty"`
 
-	// DefaultRepoPermissionEdit can be one of: "read", "write", "admin", or "none". (Default: "read").
+	// DefaultRepoPermission can be one of: "read", "write", "admin", or "none". (Default: "read").
 	// It is only used in OrganizationsService.Edit.
-	DefaultRepoPermissionEdit *string `json:"default_repository_permission,omitempty"`
-	// DefaultRepoPermissionGet can be one of: "read", "write", "admin", or "none". (Default: "read").
+	DefaultRepoPermission *string `json:"default_repository_permission,omitempty"`
+	// DefaultRepoSettings can be one of: "read", "write", "admin", or "none". (Default: "read").
 	// It is only used in OrganizationsService.Get.
-	DefaultRepoPermissionGet *string `json:"default_repository_settings,omitempty"`
+	DefaultRepoSettings *string `json:"default_repository_settings,omitempty"`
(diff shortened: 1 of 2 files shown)


jmcampanini:
@gmlewis this should match the docs and hopefully reduce confusion. let me know if you want any renames or additional godoc

github/orgs.go
```diff
@@ -46,9 +46,13 @@ type Organization struct {
...
 	TwoFactorRequirementEnabled *bool      `json:"two_factor_requirement_enabled,omitempty"`
 
-	// DefaultRepoPermission can be one of: "read", "write", "admin", or "none". (Default: "read").
+	// DefaultRepoPermissionEdit can be one of: "read", "write", "admin", or "none". (Default: "read").
 	// It is only used in OrganizationsService.Edit.
-	DefaultRepoPermission *string `json:"default_repository_settings,omitempty"`
+	DefaultRepoPermissionEdit *string `json:"default_repository_permission,omitempty"`
+	// DefaultRepoPermissionGet can be one of: "read", "write", "admin", or "none". (Default: "read").
+	// It is only used in OrganizationsService.Get.
+	DefaultRepoPermissionGet *string `json:"default_repository_settings,omitempty"`
```
gmlewis: I'm not sure why I let this go 8 hours ago, but why isn't this field called `DefaultRepoSettings` or `Default RepositorySettings` ?

github/orgs.go
```diff
@@ -46,9 +46,13 @@ type Organization struct {
 	Plan                        *Plan      `json:"plan,omitempty"`
 	TwoFactorRequirementEnabled *bool      `json:"two_factor_requirement_enabled,omitempty"`
 
-	// DefaultRepoPermission can be one of: "read", "write", "admin", or "none". (Default: "read").
+	// DefaultRepoPermissionEdit can be one of: "read", "write", "admin", or "none". (Default: "read").
 	// It is only used in OrganizationsService.Edit.
-	DefaultRepoPermission *string `json:"default_repository_settings,omitempty"`
+	DefaultRepoPermissionEdit *string `json:"default_repository_permission,omitempty"`
```
gmlewis: Why isn't this `DefaultRepoPermission` or `DefaultRepositoryPermission` to match the json?

jmcampanini:
ive updated the field names to match the JSON keys

PR #1020, opened 2018-09-30
fix import statement

themousepotato:
Correct import statement and update according to https://godoc.org/github.com/google/go-github/github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ If you're interested in using the [GraphQL API v4][], the recommended library is
 ## Usage ##
 
 ```go
-import "github.com/google/go-github/v18/github"
+import "github.com/google/go-github/v18/github"	// with go modules enabled (GO111MODULE=on or outside GOPATH)
+import "github.com/google/go-github/github" // with go modules disabled
 ```
 
 Construct a new GitHub client, then use the various services on the client to


themousepotato:
I signed it!

themousepotato:
Hey @gmlewis !

I tested the code. It wasn't working. I updated it as https://godoc.org/github.com/google/go-github/github. If this isn't correct, it should be corrected. From my testing experience, I can say that this is correct. Please test it once in the meantime.

gmlewis:
Thanks, @themousepotato.

What version of Go are you using?

Is your environment variable `GO111MODULE` set?
If so, what is its value?

Have you read the docs on Go modules and specifically release information here:
https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher
?

One thing we are trying to do is move toward supporting `Go v1.11` and beyond by adding support for Go modules.

themousepotato:
@gmlewis Sorry for the late reply. Here you go:
`go version go1.11 linux/amd64`

`GO111MODULE` env var is not set. I only tried to keep in sync with the docs on https://godoc.org/github.com/google/go-github/github. Thanks!

gmlewis:
Thanks, @themousepotato, for the extra info.
@marwan-at-work - do you have time to take a look at this PR and comment please?

README.md
```diff
@@ -12,7 +12,7 @@ If you're interested in using the [GraphQL API v4][], the recommended library is
 ## Usage ##
 
 ```go
-import "github.com/google/go-github/v18/github"
+import "github.com/google/go-github/github"
```
marwan-at-work: we should update the docs to say this: 

```
Usage: 

with go modules enabled (GO111MODULE=on or outside GOPATH): import "github.com/google/go-github/v18/github"	

with go modules disable: import "github.com/google/go-github/github"
```

Then also mention that v18 and above is go1.9.x and above. 
For anything below, people can use v17.

PR #1023, opened 2018-10-02
fixes AcceptedError unmarshaling

### LINKED ISSUE #1022: Bug with how AcceptedError is handled
Computing stats, `ListContributorsStats`, will return an `AcceptedError` and a body that is a `JSON` object and not a `JSON` array. However, it seems this will try to decode it as a `JSON` array causing a serialization error.

Commit that caused the bug... this seems to be a breaking change as well; I guess it is now obvious that it is:
https://github.com/google/go-github/commit/88eb4e9dc91cb2db1bbc56c936f1ee39bbe72ba5

My suggestion would be to internally type alias the `[]ContributorStats` and add a `ImmediateAcceptedErrorReturn` interface that can be casted during the `client.Do` function. If it is castable to an `ImmediateAcceptedErrorReturn`, it will immediately return the `AcceptedError` and not try to unmarshal the body. This makes it completely backwards compatible. I'll see if I can get some time to do this myself! :)

xibz:
Fixes #1022 

`AcceptedError` had a change in behavior in https://github.com/google/go-github/commit/88eb4e9dc91cb2db1bbc56c936f1ee39bbe72ba5

This change adds a `Raw` field to `AcceptedError`s. This allows for per API handling of the accepted error.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -502,13 +502,23 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 	err = CheckResponse(resp)
 	if err != nil {
-		// Even though there was an error, we still return the response
-		// in case the caller wants to inspect it further.
-		// However, if the error is AcceptedError, decode it below before
-		// returning from this function and closing the response body.
-		if _, ok := err.(*AcceptedError); !ok {
-			return response, err
+		// Special case for AcceptedErrors. If an AcceptedError
+		// has been encountered, the response's payload will be
+		// added to the AcceptedError and returned.
+		//
+		// Issue #1022
+		aerr, ok := err.(*AcceptedError)
+		if ok {
+			b, readErr := ioutil.ReadAll(resp.Body)
+			if readErr != nil {
+				return response, readErr
+			}
+
+			aerr.Raw = b
+			return response, aerr
 		}
+
+		return response, err
 	}
 
 	if v != nil {
@@ -610,7 +620,10 @@ func (r *RateLimitError) Error() string {
 // Technically, 202 Accepted is not a real error, it's just used to
 // indicate that results are not ready yet, but should be available soon.
 // The request can be repeated after some time.
-type AcceptedError struct{}
+type AcceptedError struct {
+	// Raw contains the response body.
+	Raw []byte
+}
 
 func (*AcceptedError) Error() string {
 	return "job scheduled on GitHub side; try again later"
--- a/github/repos_forks.go
+++ b/github/repos_forks.go
@@ -8,6 +8,8 @@ package github
 import (
 	"context"
 	"fmt"
+
+	"encoding/json"
 )
 
 // RepositoryListForksOptions specifies the optional parameters to the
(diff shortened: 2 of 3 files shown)


gmlewis:
@dsymonds - could you please review this PR when you have time?

xibz:
@dsymonds - Thanks for the feedback in #1022. I went ahead and updated what I think is a much better solution.

github/repos.go
```diff
@@ -1122,6 +1123,14 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string,
 	r := new(Repository)
 	resp, err := s.client.Do(ctx, req, r)
 	if err != nil {
+		// Persist AcceptedError's metadata to the Repository object.
+		if aerr, ok := err.(*AcceptedError); ok {
```
xibz: Hm, I guess transfer was never updated to handle `AcceptedError`, but probably should have since it also contains metadata. Im actually going to remove this change to keep it to be the same behavior.

github/github.go
```diff
@@ -502,13 +502,23 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
 	if err != nil {
-		// Even though there was an error, we still return the response
-		// in case the caller wants to inspect it further.
-		// However, if the error is AcceptedError, decode it below before
-		// returning from this function and closing the response body.
-		if _, ok := err.(*AcceptedError); !ok {
-			return response, err
+		// Special case for AcceptedErrors. If an AcceptedError
+		// has been encountered, the response's payload will be
+		// set on the AcceptedError and returned.
```
dsymonds:
```go
// added to the AcceptedError.
```

github/github.go
```diff
@@ -502,13 +502,23 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
+		// Issue #1022
+		aerr, ok := err.(*AcceptedError)
+		if ok {
+			b, readErr := ioutil.ReadAll(resp.Body)
+			if readErr != nil {
+				return response, readErr
+			}
+
+			aerr.Raw = b
+			err = aerr
```
dsymonds: `return response, aerrr` is enough.

github/github.go
```diff
@@ -610,7 +620,11 @@ func (r *RateLimitError) Error() string {
 // Technically, 202 Accepted is not a real error, it's just used to
 // indicate that results are not ready yet, but should be available soon.
 // The request can be repeated after some time.
-type AcceptedError struct{}
+type AcceptedError struct {
+	// Raw contains the raw payload that was returned with the
+	// error.
+	Raw json.RawMessage
```
dsymonds: this may as well be `[]byte`

github/github.go
```diff
@@ -610,7 +620,11 @@ func (r *RateLimitError) Error() string {
 // Technically, 202 Accepted is not a real error, it's just used to
 // indicate that results are not ready yet, but should be available soon.
 // The request can be repeated after some time.
-type AcceptedError struct{}
+type AcceptedError struct {
+	// Raw contains the raw payload that was returned with the
```
dsymonds:
```go
// Raw contains the response body.
```

dsymonds:
The general shape of this change LGTM.

Back to @gmlewis.

gmlewis:
Thank you, @xibz, and thank you, @dsymonds for taking the time to review and provide feedback... your assistance is greatly appreciated!
LGTM.
Merging.

PR #1024, opened 2018-10-02
Add comment explaining use of examples_test.go

gauntface:
Adds a comment explaining what these examples are used for. This is to clarify how they differ from example/ and test/


gmlewis:
Thank you, @gauntface!
LGTM.
Merging.

Issue #1040, opened 2018-11-03
Support updates to Marketplace-related webhooks and REST APIs


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2018-10-31-improved-experience-for-marketplace-pending-changes/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

srmocher:
@gmlewis New contributor here. Interested in working on this, could you assign this issue to me?

gmlewis:
Thank you, @srmocher!
I'm no longer able to send a repo contributor invite to you, but please consider yourself assigned. :smile:

dreamer-89:
@gmlewis : I am also interested to work on this issue. It seems @srmocher working on this one. Can you please assign me other issues/bugs ?

gmlewis:
Thanks for your interest, @dreamer-89!

My recommendation to you is to go through the list of open issues in this repo and choose one that interests you... then add a comment that you want to work on it. (I just browsed through and there are a [bunch of outstanding ones](https://github.com/google/go-github/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22).)

If someone is already listed on it, chances are high that they have forgotten about it and you can ask to take it over from them.

dreamer-89:
Thank you so much @gmlewis :) :)

dreamer-89:
@gmlewis : Please review
https://github.com/google/go-github/pull/1041

gmlewis:
@dreamer-89 - sorry I wasn't clear before... you apparently misunderstood me. This issue is assigned to @srmocher and so you should not be submitting PRs against this issue. I started to review your PR and came across a number of issues (both with the PR and with the GitHub documentation), then came back here and realized that this issue is not yours (currently) to be working on.

As a result, I'm going to close your PR and let's give @srmocher a chance to submit a PR against this issue.

[...]

PR #1042, opened 2018-11-04
Support updates to Marketplace APIs and webhook payload

srmocher:
Adds support for pending changes to Marketplace plans which were added to the API. (discussed here https://developer.github.com/changes/2018-10-31-improved-experience-for-marketplace-pending-changes/) and referenced here #1040.

### DIFF
--- a/github/apps_marketplace.go
+++ b/github/apps_marketplace.go
@@ -28,22 +28,24 @@ type MarketplaceService struct {
 
 // MarketplacePlan represents a GitHub Apps Marketplace Listing Plan.
 type MarketplacePlan struct {
-	URL                 *string   `json:"url,omitempty"`
-	AccountsURL         *string   `json:"accounts_url,omitempty"`
-	ID                  *int64    `json:"id,omitempty"`
-	Name                *string   `json:"name,omitempty"`
-	Description         *string   `json:"description,omitempty"`
-	MonthlyPriceInCents *int      `json:"monthly_price_in_cents,omitempty"`
-	YearlyPriceInCents  *int      `json:"yearly_price_in_cents,omitempty"`
-	PriceModel          *string   `json:"price_model,omitempty"`
-	UnitName            *string   `json:"unit_name,omitempty"`
-	Bullets             *[]string `json:"bullets,omitempty"`
-	State               *string   `json:"state,omitempty"`
-	HasFreeTrial        *bool     `json:"has_free_trial,omitempty"`
+	URL                 *string `json:"url,omitempty"`
+	AccountsURL         *string `json:"accounts_url,omitempty"`
+	ID                  *int64  `json:"id,omitempty"`
+	Name                *string `json:"name,omitempty"`
+	Description         *string `json:"description,omitempty"`
+	MonthlyPriceInCents *int    `json:"monthly_price_in_cents,omitempty"`
+	YearlyPriceInCents  *int    `json:"yearly_price_in_cents,omitempty"`
+	// The pricing model for this listing.  Can be one of "flat-rate", "per-unit", or "free".
+	PriceModel   *string   `json:"price_model,omitempty"`
+	UnitName     *string   `json:"unit_name,omitempty"`
+	Bullets      *[]string `json:"bullets,omitempty"`
+	State        *string   `json:"state,omitempty"`
+	HasFreeTrial *bool     `json:"has_free_trial,omitempty"`
 }
 
 // MarketplacePurchase represents a GitHub Apps Marketplace Purchase.
 type MarketplacePurchase struct {
+	// BillingCycle can be one of the values "yearly", "monthly" or nil.
 	BillingCycle    *string                 `json:"billing_cycle,omitempty"`
 	NextBillingDate *string                 `json:"next_billing_date,omitempty"`
 	UnitCount       *int                    `json:"unit_count,omitempty"`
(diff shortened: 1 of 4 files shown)


github/apps_marketplace.go
```diff
@@ -46,17 +48,28 @@ type MarketplacePurchase struct {
 	UnitCount       *int                    `json:"unit_count,omitempty"`
 	Plan            *MarketplacePlan        `json:"plan,omitempty"`
 	Account         *MarketplacePlanAccount `json:"account,omitempty"`
+	OnFreeTrial     *bool                   `json:"on_free_trial,omitempty"`
+	FreeTrialEndsOn *string                 `json:"free_trial_ends_on,omitempty"`
+}
+
+// MarketplacePendingChange represents a pending change to a Github Apps Marketplace Plan
```
gmlewis: s/Github/GitHub/
s/Plan/Plan./

github/apps_marketplace.go
```diff
@@ -46,17 +48,28 @@ type MarketplacePurchase struct {
...
+type MarketplacePendingChange struct {
+	EffectiveDate *string          `json:"effective_date,omitempty"`
```
gmlewis: Why not `*Timestamp` here instead of `*string`?

srmocher: It should be a timestamp (not string). I will update accordingly.

srmocher: The reason I kept that as string was in the GitHub docs it is mentioned as such ```

effective_date | string | The date the action becomes effective.
-- | -- | --
```

File: github/apps_marketplace.go [hunk 3]
```diff
@@ -27,36 +28,51 @@ type MarketplaceService struct {
...
+	ID                  *int64  `json:"id,omitempty"`
+	Name                *string `json:"name,omitempty"`
+	Description         *string `json:"description,omitempty"`
+	MonthlyPriceInCents *int    `json:"monthly_price_in_cents,omitempty"`
+	YearlyPriceInCents  *int    `json:"yearly_price_in_cents,omitempty"`
+	// The pricing model for this listing.  Can be one of "flat-rate", "per-unit", or "free".
+	PriceModel   *string   `json:"price_model,omitempty"`
+	UnitName     *string   `json:"unit_name,omitempty"`
+	Bullets      *[]string `json:"bullets,omitempty"`
+	State        *string   `json:"state,omitempty"`
```
gmlewis: What are the possible values of `State`? Can you please add a comment to `State`?

github/apps_marketplace.go
```diff
@@ -27,36 +28,51 @@ type MarketplaceService struct {
...
 // MarketplacePurchase represents a GitHub Apps Marketplace Purchase.
 type MarketplacePurchase struct {
+	// BillingCycle can be one of the values "yearly", "monthly" or nil.
 	BillingCycle    *string                 `json:"billing_cycle,omitempty"`
 	NextBillingDate *string                 `json:"next_billing_date,omitempty"`
 	UnitCount       *int                    `json:"unit_count,omitempty"`
 	Plan            *MarketplacePlan        `json:"plan,omitempty"`
 	Account         *MarketplacePlanAccount `json:"account,omitempty"`
+	OnFreeTrial     *bool                   `json:"on_free_trial,omitempty"`
+	FreeTrialEndsOn *string                 `json:"free_trial_ends_on,omitempty"`
```
gmlewis: This should probably be a `*Timestamp`.

github/apps_marketplace.go
```diff
@@ -27,36 +28,51 @@ type MarketplaceService struct {
...
+	Bullets      *[]string `json:"bullets,omitempty"`
+	State        *string   `json:"state,omitempty"`
+	HasFreeTrial *bool     `json:"has_free_trial,omitempty"`
 }
 
 // MarketplacePurchase represents a GitHub Apps Marketplace Purchase.
 type MarketplacePurchase struct {
+	// BillingCycle can be one of the values "yearly", "monthly" or nil.
 	BillingCycle    *string                 `json:"billing_cycle,omitempty"`
 	NextBillingDate *string                 `json:"next_billing_date,omitempty"`
```
gmlewis: This should probably be a `*Timestamp`.

github/apps_marketplace.go
```diff
@@ -27,36 +28,51 @@ type MarketplaceService struct {
...
 	UnitCount       *int                    `json:"unit_count,omitempty"`
 	Plan            *MarketplacePlan        `json:"plan,omitempty"`
 	Account         *MarketplacePlanAccount `json:"account,omitempty"`
+	OnFreeTrial     *bool                   `json:"on_free_trial,omitempty"`
+	FreeTrialEndsOn *string                 `json:"free_trial_ends_on,omitempty"`
+}
+
+// MarketplacePendingChange represents a pending change to a GitHub Apps Marketplace Plan.
+type MarketplacePendingChange struct {
+	EffectiveDate *time.Time       `json:"effective_date,omitempty"`
```
gmlewis: This should be a `*Timestamp` to be consistent with the repo.

github/apps_marketplace.go
srmocher: I only found sparse description for the possible values here. Seems like it can take either "draft" or "published". https://developer.github.com/marketplace/listing-on-github-marketplace/setting-a-github-marketplace-listing-s-pricing-plan/#creating-pricing-plans

github/apps_marketplace.go
srmocher: Updated.

github/apps_marketplace.go
srmocher: Updated.

github/apps_marketplace.go
srmocher: Updated.

gmlewis:
Thank you, @gauntface!
Merging.

PR #1044, opened 2018-11-06
Restore the resp body to an "unread state"

lukebjerring:
Suggested fix for #1047

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -524,16 +524,21 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return response, err
 	}
 
-	if v != nil {
-		if w, ok := v.(io.Writer); ok {
-			io.Copy(w, resp.Body)
-		} else {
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you for your PR, @lukebjerring, but there are a couple problems with this...
The first is that it doesn't compile:

```go
github/github.go:528:17: no new variables on left side of :=
```

The second is that `gofmt` needs to be run on the file you changed.  Please see our CONTRIBUTING.md file in this repo for more details about how to submit a PR.

lukebjerring:
Hi @gmlewis, sorry for the sloppy PR, I authored it in the GitHub UI. Cleaned up + added test.

While this is a possible "solution" to the issue, I was hoping to discuss whether the need for it justifies potential performance/memory impacts.

Thoughts?

gmlewis:
Hi @lukebjerring - thanks for the cleanup.

Looking at the code now, I'm a bit concerned about the approach.
I think I need to better understand the problem that you are trying to solve.  I'll comment on #1047.

PR #1061, opened 2018-11-29
Add actions to check create/update

### LINKED ISSUE #1060: {Create,Update}CheckRunOptions missing Actions
As per the docs, check runs support optionally exposing some user-invoked actions.
https://developer.github.com/v3/checks/runs/#create-a-check-run
https://developer.github.com/v3/checks/runs/#update-a-check-run

These fields are missing from `CreateCheckRunOptions` and `UpdateCheckRunOptions` structs.

lukebjerring:
Fixes #1060 

Adds a field for the `actions` object array, and a struct definition of that object's fields.

### DIFF
--- a/github/checks.go
+++ b/github/checks.go
@@ -149,7 +149,7 @@ type CreateCheckRunOptions struct {
 // CheckRunAction exposes further actions the integrator can perform, which a user may trigger.
 type CheckRunAction struct {
 	Label       string `json:"label"`       // The text to be displayed on a button in the web UI. The maximum size is 20 characters. (Required.)
-	Description string `json:"description,omitempty"` // A short explanation of what this action would do. The maximum size is 40 characters. (Required.)
+	Description string `json:"description"` // A short explanation of what this action would do. The maximum size is 40 characters. (Required.)
 	Identifier  string `json:"identifier,omitempty"`  // A reference for the action on the integrator's system. The maximum size is 20 characters. (Required.)
 }
 
--- a/github/checks.go
+++ b/github/checks.go
@@ -133,16 +133,24 @@ func (s *ChecksService) GetCheckSuite(ctx context.Context, owner, repo string, c
 
 // CreateCheckRunOptions sets up parameters needed to create a CheckRun.
 type CreateCheckRunOptions struct {
-	Name        string          `json:"name"`                   // The name of the check (e.g., "code-coverage"). (Required.)
-	HeadBranch  string          `json:"head_branch"`            // The name of the branch to perform a check against. (Required.)
-	HeadSHA     string          `json:"head_sha"`               // The SHA of the commit. (Required.)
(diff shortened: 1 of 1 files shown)


gmlewis:
I'm trying to figure out why `github-accessors.go` was not updated, but I have to run out the door... more later.

lukebjerring:
I was wondering that myself; IIRC, arrays get treated specially because they're `nil` when empty/len(0) anyway.
And, all of the fields on the new struct are required (so don't need the non-nil-accessor helpers)

github/checks.go
```diff
@@ -133,16 +133,24 @@ func (s *ChecksService) GetCheckSuite(ctx context.Context, owner, repo string, c
...
+	Conclusion  *string           `json:"conclusion,omitempty"`   // Can be one of "success", "failure", "neutral", "cancelled", "timed_out", or "action_required". (Optional. Required if you provide a status of "completed".)
+	StartedAt   *Timestamp        `json:"started_at,omitempty"`   // The time that the check run began. (Optional.)
+	CompletedAt *Timestamp        `json:"completed_at,omitempty"` // The time the check completed. (Optional. Required if you provide conclusion.)
+	Output      *CheckRunOutput   `json:"output,omitempty"`       // Provide descriptive details about the run. (Optional)
+	Actions     []*CheckRunAction `json:"actions,omitempty"`      // Possible further actions the integrator can perform, which a user may trigger. (Optional.)
+}
+
+// CheckRunAction exposes further actions the integrator can perform, which a user may trigger.
+type CheckRunAction struct {
+	Label       string `json:"label,omitempty"`       // The text to be displayed on a button in the web UI. The maximum size is 20 characters. (Required.)
```
gmlewis: Would you please remove the `,omitempty` on these three required parameters to be consistent with the other required parameters?

lukebjerring:
```suggestion
	Label       string `json:"label"`       // The text to be displayed on a button in the web UI. The maximum size is 20 characters. (Required.)
```

github/checks.go
```diff
@@ -133,16 +133,24 @@ func (s *ChecksService) GetCheckSuite(ctx context.Context, owner, repo string, c
...
+	Description string `json:"description,omitempty"` // A short explanation of what this action would do. The maximum size is 40 characters. (Required.)
```
lukebjerring:
```suggestion
	Description string `json:"description"` // A short explanation of what this action would do. The maximum size is 40 characters. (Required.)
```

github/checks.go
```diff
@@ -133,16 +133,24 @@ func (s *ChecksService) GetCheckSuite(ctx context.Context, owner, repo string, c
...
+	Identifier  string `json:"identifier,omitempty"`  // A reference for the action on the integrator's system. The maximum size is 20 characters. (Required.)
```
lukebjerring:
```suggestion
	Identifier  string `json:"identifier"`  // A reference for the action on the integrator's system. The maximum size is 20 characters. (Required.)
```

lukebjerring:
Done.

PR #1075, opened 2018-12-17
added Body_html export from Github API

nico-bock:
I only changed the header which defines the expected answer json (https://developer.github.com/v3/media/) and added BodyHTML to the RepositoryRelease struct.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -36,7 +36,7 @@ const (
 	headerRateReset     = "X-RateLimit-Reset"
 	headerOTP           = "X-GitHub-OTP"
 
-	mediaTypeV3                = "application/vnd.github.v3+json"
+	mediaTypeV3                = "application/vnd.github.full+json"
 	defaultMediaType           = "application/octet-stream"
 	mediaTypeV3SHA             = "application/vnd.github.v3.sha"
 	mediaTypeV3Diff            = "application/vnd.github.v3.diff"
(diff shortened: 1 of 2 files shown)


gmlewis:
Why?

What are you trying to solve?

gmlewis:
OK, then I think we need to design a better way to do this, as the current proposal would break all current functionality if I understand the proposed changes correctly.

nico-bock:
Ok, can you give me a hint what you mean?

The Change from the Accept Header from v3.json to full.json adds "body_html" and "body_text" to the wandered json.

I compared curl
 -X GET \
  https://api.github.com/repos/adam-p/markdown-here/releases/latest \
  -H 'Accept: application/vnd.github.full+json' \

to 

curl -X GET \
  https://api.github.com/repos/adam-p/markdown-here/releases/latest \
  -H 'Accept: application/vnd.github.v3+json' \

gmlewis:
What I mean is that we shouldn't change the definition of `mediaTypeV3` suddenly because that would affect every user of this package.

This package has been in use for many years by many people and this is the first time anyone has desired to have the `body_html` returned that I'm aware of. We are not going to suddenly increase the returned payload size for every API call arbitrarily.

That is why I wanted to know what your use case it, as it might be possible to solve your problem in a different manner.

Saying

> I want to support the bodyHtml from GitHub.

doesn't really answer my question.

If your original problem can't be solved in another way, then at the very least we would want to gate this change by an option so that the rest of the users of this package are not affected by this change.

nico-bock:
Ah okay. I did't find other parts which seemed to be affected by that change. But you have the better overview.

I wanted to get the bodyHtml property from the GitHub API, because I'd like to send automated RealeaseNews via e-mail and the converting from Markdown to html is more error-prone then get the converted html from GitHub.

gmlewis:
Closing.

PR #1076, opened 2018-12-21
list user repos, list repos by org does not support the code of conduct preview

fejta:
https://developer.github.com/v3/repos/ does not mention any of the List repositories apis supporting  `application/vnd.github.scarlet-witch-preview+json`

And https://developer.github.com/v3/codes_of_conduct/ only mentions it for getting a repo, not listing.

Fixes https://github.com/google/go-github/issues/930

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -179,7 +179,7 @@ func (s *RepositoriesService) List(ctx context.Context, user string, opt *Reposi
 	}
 
 	// TODO: remove custom Accept headers when APIs fully launch.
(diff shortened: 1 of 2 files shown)


fejta:
Thanks! Do you need another LGTM from a repo collaborator or I can I just find someone random on the internet? 😁

fejta:
@ibrasho Maybe leave an approving review? Prow and its `/assign`, etc commands aren't in this repo 😉

ibrasho:
There you go. 👽

gmlewis:
Thank you, @ibrasho!
Merging.

PR #1079, opened 2018-12-21
Adding the new Interactions API support for repositories

waseem18:
Created files based on the `{service}_{api}.go` format i.e `interactions_repos.go`.

API response for repository interactions and organisation interactions is same so I've abstracted the `Interaction` struct in `interactions.go`.

This PR adds the API's to `GetRestrictions`, `UpdateRestrictions` and `RemoveRestrictions` for a given repository.

### DIFF
--- a/github/interactions.go
+++ b/github/interactions.go
@@ -1,0 +1,20 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+// InteractionsService handles communication with the repo and org related
+// methods of the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/
+type InteractionsService service
+
+// Interaction represents the interaction restrictions
+// for repository and organisation
+type Interaction struct {
+	Limit     *string    `json:"limit,omitempty"`
+	Origin    *string    `json:"origin,omitempty"`
+	ExpiresAt *Timestamp `json:"expires_at,omitempty"`
+}
--- a/github/interactions.go
+++ b/github/interactions.go
@@ -11,8 +11,8 @@ package github
 // GitHub API docs: https://developer.github.com/v3/interactions/
 type InteractionsService service
 
-// Interaction represents the interaction restrictions for repository and organization.
-type Interaction struct {
+// InteractionRestriction represents the interaction restrictions for repository and organization.
+type InteractionRestriction struct {
 	// Specifies the group of GitHub users who can
 	// comment, open issues, or create pull requests for the given repository.
 	// Possible values are: "existing_users", "contributors_only" and "collaborators_only".
--- a/github/interactions.go
+++ b/github/interactions.go
@@ -5,16 +5,24 @@
 
 package github
 
-// InteractionsService handles communication with the repo and org related
+// InteractionsService handles communication with the repository and organization related
 // methods of the GitHub API.
 //
 // GitHub API docs: https://developer.github.com/v3/interactions/
 type InteractionsService service
 
-// Interaction represents the interaction restrictions
-// for repository and organisation
+// Interaction represents the interaction restrictions for repository and organization.
 type Interaction struct {
-	Limit     *string    `json:"limit,omitempty"`
-	Origin    *string    `json:"origin,omitempty"`
+	// Specifies the group of GitHub users who can
+	// comment, open issues, or create pull requests for the given repository.
+	// Possible values are: "existing_users", "contributors_only" and "collaborators_only".
+	Limit *string `json:"limit,omitempty"`
+
+	// Origin specifies the type of the resource to interact with.
+	// Possible values are: "repository" and "organization".
+	Origin *string `json:"origin,omitempty"`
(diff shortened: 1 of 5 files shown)


github/interactions.go
```diff
@@ -0,0 +1,20 @@
...
+package github
+
+// InteractionsService handles communication with the repo and org related
+// methods of the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/
+type InteractionsService service
+
+// Interaction represents the interaction restrictions
+// for repository and organisation
```
gmlewis: For the sake of the godocs, please make all doc comments be complete sentences followed by a period.

Also, for consistency, let's please use the American English spellings, for example: `organization`.
If you are unsure, you can copy the spellings from the GitHub Developer API docs.

Additionally, I noticed you did a force push... please do not use those in PRs for this repo, as it makes things easier for reviewers to see only the changes since the last review.  Thank you!

github/interactions.go
```diff
@@ -0,0 +1,20 @@
...
+type Interaction struct {
+	Limit     *string    `json:"limit,omitempty"`
```
gmlewis: Can you please document the possible values each field can assume if when they are enumerated in the GitHub Developer API docs?

waseem18:
@gmlewis  Added API's to update and remove interactions of a repository.

Please review.

Thanks

github/interactions_repos.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+// GetInteractions fetches the interaction restrictions for a repository.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/repos/#get-interaction-restrictions-for-a-repository
+func (s *InteractionsService) GetInteractions(ctx context.Context, owner string, repo string) (*Interaction, *Response, error) {
```
gmlewis: Since this is the `InteractionsService`, what do you think about naming these functions `GetRestrictions`, `UpdateRestrictions`, and `RemoveRestrictions`?

github/interactions_repos.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, resp, err
+	}
+
+	return repositoryInteractions, resp, nil
+}
+
+// UpdateInteractions adds or updates the interaction restrictions for a repository.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/repos/#add-or-update-interaction-restrictions-for-a-repository
+func (s *InteractionsService) UpdateInteractions(ctx context.Context, owner string, repo string, interaction *Interaction) (*Interaction, *Response, error) {
```
gmlewis: Here, and on the other functions as well, let's take advantage of Go's parameter type shorthand:

s/owner string, repo string/owner, repo string/

github/interactions_repos.go
waseem18: I agree with this. Will address.

github/interactions_repos.go
waseem18: I'm aware of the shorthand notation in case all the parameters are of the same type.

But I didn't get this below line.
> s/owner string, repo string/owner, repo string/

waseem18: Addressed in the new commit - https://github.com/google/go-github/pull/1079/commits/a8446a5e03d2118bccaf382d31d5921ef3d58708

github/interactions.go
```diff
@@ -0,0 +1,28 @@
...
+package github
+
+// InteractionsService handles communication with the repository and organization related
+// methods of the GitHub API.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/
+type InteractionsService service
+
+// Interaction represents the interaction restrictions for repository and organization.
+type Interaction struct {
```
gauntface: Should this be `InteractionRestriction`? 

`Interaction` with these fields wouldn't read particularly in my head if I was working with docs / autocomplete.

gmlewis: Excellent point, @gauntface! I agree. Let's please change this, @waseem18.

waseem18: Addressed in a new commit - https://github.com/google/go-github/pull/1079/commits/c93e1de5d098a64104ff0c9d4933bc92e87e3388 👍

gmlewis:
Thank you, @waseem18 and @gauntface!
LGTM.
Merging.

github/interactions_repos.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+// GetRestrictions fetches the interaction restrictions for a repository.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/repos/#get-interaction-restrictions-for-a-repository
+func (s *InteractionsService) GetRestrictions(ctx context.Context, owner, repo string) (*InteractionRestriction, *Response, error) {
```
waseem18: @gmlewis I was wondering if I should rename `GetRestrictions` to `GetRestrictionsForRepo` so that for Organization interactions API, we can name methods as `GetRestrictionsForOrg`?

gmlewis: That sounds like a good idea, @waseem18!

Issue #1077, opened 2018-12-21
Support the new Interactions API


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2018-12-18-interactions-preview/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

waseem18:
@gmlewis I'm really interested in working on this issue.

gmlewis:
Thank you, @waseem18!

This issue is now assigned to you (sorry I can't send you an invite).

Please feel free to work on your PR in your own forked repo and send a pull request when you have tested it out and are ready to have it reviewed.

waseem18:
Sure, Will do that :)

waseem18:
I'm gonna divide this issue into two parts.

1. Interactions API for a repositories
2. Interactions API for organisations

I have put up a [WIP PR](https://github.com/google/go-github/pull/1078) for repo interactions. After finishing repo interactions will hop onto organisation interactions.

waseem18:
Now that #1079 is merged. I'll start on adding Interactions API for Organizations.

PR #1090, opened 2019-01-04
Added Interactions restrictions API for organizations

waseem18:
Issue : #1077

### DIFF
--- a/github/interactions_orgs.go
+++ b/github/interactions_orgs.go
@@ -1,0 +1,73 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// GetRestrictionsForOrg fetches the interaction restrictions for an organization.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/orgs/#get-interaction-restrictions-for-an-organization
+func (s *InteractionsService) GetRestrictionsForOrg(ctx context.Context, organization string) (*InteractionRestriction, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/interaction-limits", organization)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeInteractionRestrictionsPreview)
+
+	organizationInteractions := new(InteractionRestriction)
+
+	resp, err := s.client.Do(ctx, req, organizationInteractions)
+	if err != nil {
+		return nil, resp, err
+	}
+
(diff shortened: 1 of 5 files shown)


github/interactions_orgs.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
```
gmlewis: Would you mind bumping the Copyright year on new files (only) to 2019, please?
All existing files can keep their original Copyright year unchanged.
Thanks!

github/interactions_orgs.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, resp, err
+	}
+
+	return organizationInteractions, resp, nil
+}
+
+// UpdateRestrictionsForOrg adds or updates the interaction restrictions for an organization.
+//
+// GitHub API docs: https://developer.github.com/v3/interactions/orgs/#add-or-update-interaction-restrictions-for-an-organization
+func (s *InteractionsService) UpdateRestrictionsForOrg(ctx context.Context, organization string, interaction *InteractionRestriction) (*InteractionRestriction, *Response, error) {
```
gmlewis: It just dawned on me that this API currently only takes a single parameter (`limit`) and yet we are passing in a full `InteractionRestriction` struct.

GitHub has recently been clamping down on extra parameters and is wanting to return errors when extra parameters are sent.

Therefore, we have to be extremely careful at what is being sent.

I recommend for now that we pass in another string for the required `limit` parameter and that we document what values can be passed.

Let's also go back and double-check your recent additions for repos and make sure that we are not violating this same principle, as I forgot to check that carefully.

So to be precise, I'm thinking of something like this:

```go
// UpdateRestrictionsForOrg adds or updates the interaction restrictions for an organization.
//
// limit specifies the group of GitHub users who can comment, open issues, or create pull requests
// in public repositories for the given organization.
// Possible values are: "existing_users", "contributors_only", "collaborators_only".
//
// GitHub API docs: https://developer.github.com/v3/interactions/orgs/#add-or-update-interaction-restrictions-for-an-organization
func (s *InteractionsService) UpdateRestrictionsForOrg(ctx context.Context, organization, limit string) (*InteractionRestriction, *Response, error) {
```

And then, within the function, you can still use that same struct, but we are then assured of the other fields being empty:

```go
interaction := &InteractionRestriction{limit: String(limit)}
```

github/interactions_orgs_test.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
```
gmlewis: Same here please: s/2018/2019/

github/interactions_orgs.go
waseem18: > GitHub has recently been clamping down on extra parameters and is wanting to return errors when extra parameters are sent.

Thanks, I didn't know that.

Addressed the review in both organization and repository interactions update API's.

gmlewis:
Thank you, @gauntface for the review!
Merging.

PR #1099, opened 2019-01-16
Escape commit ref with percent encoding

bgpat:
If `Repositories.GetCommitSHA1` was called with the branch name which includes non-alphabet character, it occurs an error. It looks to require escaping with percent encoding.

Below is sample code: 

```go
package main

import (
	"context"
	"fmt"

)

func main() {
	sha1, _, err := client.Repositories.GetCommitSHA1(ctx, "bgpat", "test", "cpu100%", "")
	if err != nil {
		panic(err)
	}
	fmt.Println(sha1)
}
```

before: `parse repos/bgpat/test/commits/cpu100%: invalid URL escape "%"`
after: `46736bd30fe65160f7d21e49d8f23e4687abc41b`

### DIFF
--- a/github/checks.go
+++ b/github/checks.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"net/url"
 )
 
 // ChecksService provides access to the Checks API in the
@@ -255,7 +256,7 @@ type ListCheckRunsResults struct {
 //
 // GitHub API docs: https://developer.github.com/v3/checks/runs/#list-check-runs-for-a-specific-ref
 func (s *ChecksService) ListCheckRunsForRef(ctx context.Context, owner, repo, ref string, opt *ListCheckRunsOptions) (*ListCheckRunsResults, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-runs", owner, repo, ref)
+	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-runs", owner, repo, url.QueryEscape(ref))
 	u, err := addOptions(u, opt)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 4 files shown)


gmlewis:
I'm curious if there are other APIs in our repo where we have made the same mistake?

I'm betting that anywhere we pass a `ref` in the URL, we would need to escape it.

While you are here, @bgpat, do you want to take a look?

bgpat:
I found 9 same mistakes. I will fix it.

```console
$ grep -n Sprintf *.go | grep ', ref\W'
checks.go:258:	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-runs", owner, repo, ref)
checks.go:324:	u := fmt.Sprintf("repos/%v/%v/commits/%v/check-suites", owner, repo, ref)
git_refs.go:60:	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, ref)
git_refs.go:91:	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, ref)
git_refs.go:211:	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, ref)
repos_commits.go:192:	u := fmt.Sprintf("repos/%v/%v/commits/%v", owner, repo, ref)
repos_statuses.go:47:	u := fmt.Sprintf("repos/%v/%v/commits/%v/statuses", owner, repo, ref)
repos_statuses.go:72:	u := fmt.Sprintf("repos/%v/%v/statuses/%v", owner, repo, ref)
repos_statuses.go:111:	u := fmt.Sprintf("repos/%v/%v/commits/%v/status", owner, repo, ref)
```

gmlewis:
OK, this is true awesomeness... thank you so much, @bgpat!
LGTM.

Awaiting input from @gauntface as a sanity check to make sure my review makes sense... then we will get this in.

gmlewis:
Excellent point, @gauntface!
I've added issue #1101 to track this.
Merging.

gmlewis:
This is a bug fix. Bumping package version number to capture it.

PR #1102, opened 2019-01-22
Allow media type to be specified in release asset upload

aidenkeating:
Update UploadOptions with a MediaType field. If the MediaType is
provided to UploadReleaseAsset it will take precedence over the type
retrieved from looking at the file extension. The complete order of
precedence is:

- MediaType in UploadOptions
- media type from file extension
- default media type, application/octet-stream

Update unit tests to test the order of precedence, default value and
that the Content-Type header is actually set in the upload request.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -191,7 +191,8 @@ type ListOptions struct {
 
 // UploadOptions specifies the parameters to methods that support uploads.
 type UploadOptions struct {
-	Name string `url:"name,omitempty"`
+	Name      string `url:"name,omitempty"`
+	MediaType string `url:"-"`
 }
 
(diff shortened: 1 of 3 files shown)


gmlewis:
For other Gophers watching, @aidenkeating has taken advantage of a cool feature of the [JSON Marshaling](https://golang.org/pkg/encoding/json/#Marshal) machinery... this doc says:

```
As a special case, if the field tag is "-", the field is always omitted. Note that a field with name "-" can still be generated using the tag "-,".
```

Since we don't want `MediaType` to be sent to the GitHub server, we used the `"-"` field tag, which is super-handy in this case.

Thank you, @gauntface!
Merging.

PR #1104, opened 2019-01-23
Allow label to be specified in release asset upload

aidenkeating:
Update UploadOptions with a Label field. The label is provided as a URI
query parameter to the GitHub API request when uploading the file. The
label is used as a short or alternate display name for the file in the
release.

Update unit tests to test for two scenarios, the label being provided
in UploadOptions and the label being excluded. When the label is
provided it should be a query param in the request.


aidenkeating:
The API docs for uploading a release asset are here: https://developer.github.com/v3/repos/releases/#input-2

gmlewis:
Very nice, @aidenkeating - now that we merged your other PR, would you mind updating this one and resolving the conflicts please?

aidenkeating:
@gmlewis Thanks for taking a look, rebased.

gmlewis:
Thank you, @gauntface!
Merging.

PR #1107, opened 2019-01-24
Add Target field to RepositoryContent

### LINKED ISSUE #1106: RepositoryContent target field for symlinks
Per the [API docs](https://developer.github.com/v3/repos/contents/#response-if-content-is-a-symlink), when retrieving repository contents and the file is a symlink that doesn't point to a normal file (for example, a directory), the target field is set. The target field is currently missing from [RepositoryContent](https://github.com/google/go-github/blob/master/github/repos_contents.go#L23).

I will open a PR shortly to add the Target field.

bkeroackdsc:
When getting file contents and the file is a symlink that points to anything other than a normal file (for example, a directory), the target field is used to indicate this.

This PR adds the Target field and accessor.

Fixes #1106

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -9700,6 +9700,14 @@ func (r *RepositoryContent) GetSize() int {
 	return *r.Size
 }
 
+// GetTarget returns the Target field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @bkeroackdsc!
LGTM.
Merging.

PR #1108, opened 2019-01-27
update broken and old enterprise doc links

JordanSussman:
This PR updates the existing documentation links that are either broken or outdated. You will note that `https://developer.github.com/enterprise/v3/` redirects to `https://developer.github.com/enterprise/<lastest enterprise version>/v3/`, so this will hopefully help from having the links become outdated in the future.

### DIFF
--- a/github/authorizations.go
+++ b/github/authorizations.go
@@ -403,7 +403,7 @@ func (s *AuthorizationsService) DeleteGrant(ctx context.Context, id int64) (*Res
 // you can e.g. create or delete a user's public SSH key. NOTE: creating a
 // new token automatically revokes an existing one.
 //
-// GitHub API docs: https://developer.github.com/enterprise/2.5/v3/users/administration/#create-an-impersonation-oauth-token
+// GitHub API docs: https://developer.github.com/enterprise/v3/enterprise-admin/users/#create-an-impersonation-oauth-token
 func (s *AuthorizationsService) CreateImpersonation(ctx context.Context, username string, authReq *AuthorizationRequest) (*Authorization, *Response, error) {
 	u := fmt.Sprintf("admin/users/%v/authorizations", username)
 	req, err := s.client.NewRequest("POST", u, authReq)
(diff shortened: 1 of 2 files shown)


JordanSussman:
I signed it!

gmlewis:
Actually, there should be no enterprise URLs in this repo at all.

Looking back at the history, I am to blame, as I merged in a PR that included two enterprise links:
https://github.com/google/go-github/commit/07a1059646109986eb08a3f57126f59e1d19f049
These should be removed entirely.

All enterprise users are expected to use the public API documentation, then figure out where the documentation is for their particular enterprise version since it would be impossible for this repo to support all possible enterprise versions in existence.

If you want to fix those two URLs, then they need to be changed to the public API documentation.

Thanks.

JordanSussman:
> Actually, there should be no enterprise URLs in this repo at all.
 Looking back at the history, I am to blame, as I merged in a PR that included two enterprise links:
07a1059
These should be removed entirely.

Are you proposing that no direct documentation urls be provided for any enterprise only endpoints? It seems a bit wonky to create enterprise only functions but not point to the direct documentation that correlates to the endpoints. 

> All enterprise users are expected to use the public API documentation, then figure out where the documentation is for their particular enterprise version since it would be impossible for this repo to support all possible enterprise versions in existence.

I'd argue that endpoints that are only available to the enterprise version (all of the ones that I'm updating) should point to the latest version of the public enterprise documentation. 

> If you want to fix those two URLs, then they need to be changed to the public API documentation.

The public API (github.com) documentation does not have any of these endpoints mentioned, you must navigate to a version of the enterprise documentation. What exact documentation page(s) are you proposing that I update to instead?

gmlewis:
Ah!  My mistake.  I apologize... I did not realize that these URLs are Enterprise-only...
In that case, this makes perfect sense.
I'll take a closer look tomorrow.
Thank you, @JordanSussman!

JordanSussman:
Thanks!

PR #1112, opened 2019-02-03
add UserSuspendOptions for user.Suspend

JordanSussman:
GitHub updated the [suspend a user endpoint](https://developer.github.com/enterprise/v3/enterprise-admin/users/#suspend-a-user) to allow you to send a JSON payload containing the reason that the specified user is being suspended. If you input `""` for the reason, GitHub will default the reason to `Suspended via API by SITE_ADMINISTRATOR` in the audit log.

### DIFF
--- a/github/users_administration.go
+++ b/github/users_administration.go
@@ -40,17 +40,16 @@ func (s *UsersService) DemoteSiteAdmin(ctx context.Context, user string) (*Respo
 
 // userSuspendOptions represents the reason a user is being suspended.
 type userSuspendOptions struct {
-	Reason string `json:"reason,omitempty"`
+	Reason *string `json:"reason,omitempty"`
 }
 
 // Suspend a user on a GitHub Enterprise instance.
 //
 // GitHub API docs: https://developer.github.com/enterprise/v3/enterprise-admin/users/#suspend-a-user
-func (s *UsersService) Suspend(ctx context.Context, user string, reason string) (*Response, error) {
+func (s *UsersService) Suspend(ctx context.Context, user string, opt *userSuspendOptions) (*Response, error) {
 	u := fmt.Sprintf("users/%v/suspended", user)
-	opts := &userSuspendOptions{Reason: reason}
 
-	req, err := s.client.NewRequest("PUT", u, opts)
+	req, err := s.client.NewRequest("PUT", u, opt)
 	if err != nil {
 		return nil, err
 	}
--- a/github/users_administration.go
+++ b/github/users_administration.go
@@ -38,15 +38,15 @@ func (s *UsersService) DemoteSiteAdmin(ctx context.Context, user string) (*Respo
 	return s.client.Do(ctx, req, nil)
 }
 
-// userSuspendOptions represents the reason a user is being suspended.
-type userSuspendOptions struct {
+// UserSuspendOptions represents the reason a user is being suspended.
+type UserSuspendOptions struct {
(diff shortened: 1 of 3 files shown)


github/users_administration.go
```diff
@@ -38,13 +38,19 @@ func (s *UsersService) DemoteSiteAdmin(ctx context.Context, user string) (*Respo
...
+// userSuspendOptions represents the reason a user is being suspended.
+type userSuspendOptions struct {
+	Reason string `json:"reason,omitempty"`
+}
+
 // Suspend a user on a GitHub Enterprise instance.
 //
 // GitHub API docs: https://developer.github.com/enterprise/v3/enterprise-admin/users/#suspend-a-user
-func (s *UsersService) Suspend(ctx context.Context, user string) (*Response, error) {
+func (s *UsersService) Suspend(ctx context.Context, user string, reason string) (*Response, error) {
```
gmlewis: I'm thinking that we want to make this an `opt *UserSuspendOptions` and and allow it to be `nil` to continue the current behavior.

That way, it is more consistent with other endpoints in this package and it allows us to be a bit more flexible by not breaking this package's API in case GitHub wishes to add more options to this command.

github/users_administration.go
```diff
@@ -38,13 +38,19 @@ func (s *UsersService) DemoteSiteAdmin(ctx context.Context, user string) (*Respo
 	return s.client.Do(ctx, req, nil)
 }
 
+// userSuspendOptions represents the reason a user is being suspended.
+type userSuspendOptions struct {
+	Reason string `json:"reason,omitempty"`
```
gmlewis: Let's change this from `string` to `*string` so that the `omitempty` can do its job and not pass it if it is not provided.

github/users_administration_test.go
```diff
@@ -50,7 +50,7 @@ func TestUsersService_Suspend(t *testing.T) {
 		w.WriteHeader(http.StatusNoContent)
 	})
 
-	_, err := client.Users.Suspend(context.Background(), "u")
+	_, err := client.Users.Suspend(context.Background(), "u", "test")
```
gmlewis: Please add a new test that checks that `reason` is passed successfully and keep one current test where `reason` is not passed at all.

JordanSussman:
Thanks for the feedback, @gmlewis. 

The latest commit should address the requested updates.

github/users_administration.go
```diff
@@ -38,13 +38,18 @@ func (s *UsersService) DemoteSiteAdmin(ctx context.Context, user string) (*Respo
 	return s.client.Do(ctx, req, nil)
 }
 
+// userSuspendOptions represents the reason a user is being suspended.
+type userSuspendOptions struct {
```
gmlewis: Please export this type: `UserSuspendOptions`.

github/users_administration_test.go
```diff
@@ -50,7 +52,32 @@ func TestUsersService_Suspend(t *testing.T) {
 		w.WriteHeader(http.StatusNoContent)
 	})
 
-	_, err := client.Users.Suspend(context.Background(), "u")
+	opt := &userSuspendOptions{}
+	_, err := client.Users.Suspend(context.Background(), "u", opt)
```
gauntface: I would make this nil instead of an empty options.

JordanSussman:
Thanks for the feedback!

The last couple of commits should address everything mentioned.

PR #1111, opened 2019-02-01
Removing 3 service hooks endpoints from the code and the test

joshuabezaleel:
Issue: https://github.com/google/go-github/issues/1109

According to this [post](https://developer.github.com/changes/2019-01-29-life-after-github-services/), there will be 3 endpoints affected which will be deprecated: Edit a Hook, Test a Hook, and Ping a Hook. This PR removes the code regarding those 3 endpoints and also from the test file.

cc: @gmlewis

PR #1115, opened 2019-02-09
Revert "Removing 3 service hooks endpoints from the code and the test"

### LINKED ISSUE #1114: Unable to Edit Repository Webhooks
Could you consider reverting https://github.com/google/go-github/pull/1111 please? Those API endpoints happen to be used by both the GitHub Services feature, which is being deprecated (as mentioned in https://github.com/google/go-github/issues/1109), but also by repository webhooks which are still very much a supported feature from what I can tell. https://developer.github.com/v3/guides/replacing-github-services/#replacing-services-with-webhooks mentions that the migration path is to use webhooks, and https://developer.github.com/webhooks/#webhooks still mentions both organization and repository webhooks.  Finally, the [announcement](https://developer.github.com/changes/2019-01-29-life-after-github-services/) specifically states that the particular APIs that were removed in https://github.com/google/go-github/pull/1111 

> affect the following endpoints for _service hooks only_ 

i.e. they're still vaild for repository hooks themselves.

FYI, I stumbled across this whilst trying to update the vendored version of go-github in https://github.com/terraform-providers/terraform-provider-github.

gmlewis:
Reverts google/go-github#1111.
Fixes #1114.

### DIFF
--- a/github/repos_hooks.go
+++ b/github/repos_hooks.go
@@ -169,12 +169,54 @@ func (s *RepositoriesService) GetHook(ctx context.Context, owner, repo string, i
 	return h, resp, nil
 }
 
+// EditHook updates a specified Hook.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/hooks/#edit-a-hook
+func (s *RepositoriesService) EditHook(ctx context.Context, owner, repo string, id int64, hook *Hook) (*Hook, *Response, error) {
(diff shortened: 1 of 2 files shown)


gmlewis:
It wasn't your fault, @gauntface - I take full blame.
Thanks for the review!
Merging.

PR #1117, opened 2019-02-11
Fix Webhook creation.

DominicGunn:
## Changes

Attempting to create webhooks fail unless a name is specified. As per the documentation the only name that is valid is "web", so I haven't opted to allow modification of this but creation seems to fail without it.

### DIFF
--- a/github/repos_hooks.go
+++ b/github/repos_hooks.go
@@ -92,7 +92,7 @@ func (h Hook) String() string {
 // information.
 type createHookRequest struct {
 	// Config is required.
-	Name   string                 `json:"name,omitempty"`
+	Name   string                 `json:"name"`
 	Config map[string]interface{} `json:"config,omitempty"`
 	Events []string               `json:"events,omitempty"`
 	Active *bool                  `json:"active,omitempty"`
--- a/github/repos_hooks.go
+++ b/github/repos_hooks.go
@@ -92,6 +92,7 @@ func (h Hook) String() string {
 // information.
 type createHookRequest struct {
 	// Config is required.
+	Name   string                 `json:"name,omitempty"`
 	Config map[string]interface{} `json:"config,omitempty"`
 	Events []string               `json:"events,omitempty"`
 	Active *bool                  `json:"active,omitempty"`
@@ -108,6 +109,7 @@ func (s *RepositoriesService) CreateHook(ctx context.Context, owner, repo string
 	u := fmt.Sprintf("repos/%v/%v/hooks", owner, repo)
 
 	hookReq := &createHookRequest{
+		Name: 	"web",
 		Events: hook.Events,
 		Active: hook.Active,
 		Config: hook.Config,
--- a/github/repos_hooks.go
+++ b/github/repos_hooks.go
@@ -109,7 +109,7 @@ func (s *RepositoriesService) CreateHook(ctx context.Context, owner, repo string
 	u := fmt.Sprintf("repos/%v/%v/hooks", owner, repo)
 
 	hookReq := &createHookRequest{
-		Name: 	"web",
+		Name:   "web",
 		Events: hook.Events,
 		Active: hook.Active,
 		Config: hook.Config,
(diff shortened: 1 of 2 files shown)


DominicGunn:
I signed it!

gmlewis:
The only documentation I found about this:
https://developer.github.com/v3/repos/hooks/#create-a-hook
says that the `name` field has a default value of `web`.

Do you want to contact GitHub (support@github.com) and ask them to clarify what is going on after describing the situation you discovered? I have found their support team to be incredibly responsive and helpful.

radeksimko:
I have just emailed GitHub support about this, will update here once I get response from them.

radeksimko:
Here's a response I received from helpful GitHub support folk:

----

This field is currently a legacy field, in that we only provide a single option for this: 'web' for webhooks. Previously we had the concept of GitHub Services: https://developer.github.com/changes/2018-10-01-denying-new-github-services/ that allowed you to create hooks for this.

However as mentioned in the above article we have since deprecated and switched off this service. We decided to leave this field in, to prevent issues with current integrations that might still rely on this. In this case it automatically defaults to 'web' if omitted, or if provided, must have the value 'web'.

For instance the following core curl request works:
```
curl -H "Accept: application/vnd.github.v3+json" \
-H "Authorization: token $TOKEN" \
-d "{\"config\":{\"url\":\"$URL\"}}" \
https://api.github.com/repos/mattamorphic/WebhookTest/hooks
```
Where $TOKEN is the OAuth token, and $URL is the resource to send the requests to.

Long term we could choose to add other values to this as we are often iterating on the services that we provide, we could also choose to in the long term officially deprecate and remove this field as part of an ongoing effort to improve our validation rules.

In terms of the PR, https://github.com/google/go-github/pull/1117, the author @DominicGunn mentions that:

> Attempting to create webhooks fail unless a name is specified

As we haven't been able to reproduce this with the above curl I'd be interested to know if the user can reproduce this with a `curl -v`?

TL;DR; This field has a single value, is a legacy field, and has a default value if omitted.

In terms of defining this field in the SDK, I feel like the SDK should emulate the API behaviour:

1. Optional, and if not provided default to 'web'
1. If provided this can only accept 'web'

----

I'm not sure if the SDK has any convention around validation of arguments, which may need to be added, but this PR otherwise looks good to me according to the above response.

github/repos_hooks.go
```diff
@@ -92,6 +92,7 @@ func (h Hook) String() string {
 // information.
 type createHookRequest struct {
 	// Config is required.
+	Name   string                 `json:"name,omitempty"`
```
gmlewis: Please remove the `,omitempty` since it is not a pointer and with this PR, it will always be defined.

gmlewis:
Note that this PR does not implement the recommendation in the [above comment](https://github.com/google/go-github/pull/1117#issuecomment-467513337) but it sounds to me like the API is not changing, so this solution would solve the issue and is fine with me.

We'll see if @gauntface agrees.

github/repos_hooks.go
DominicGunn: Removed as per request.

DominicGunn:
For some deeper clarity, this is against a enterprise version of GHE, I'm not sure how much difference that would make. 

In terms of allowing this to be definable by the user, I'm not against it but it feels superfluous given that the only accepted value will be "web".

gauntface:
I'm ok with this landing, but it does sound like a bug in GHE if you are getting an error without the name field.

radeksimko:
> In terms of allowing this to be definable by the user, I'm not against it but it feels superfluous given that the only accepted value will be "web".

Having it definable would help us a lot in the Terraform GitHub provider as we [still haven't upgraded](https://github.com/terraform-providers/terraform-provider-github/pull/186) SDK since this field was removed and if the field is put back we can deprecate it more gracefully and easily, rather than just pulling the cord and forcing everyone to change their configurations.

sourabh3b:
Hi All,
Adding my thoughts to the conversation.
I am facing the same issue.
Getting this error, while adding a webhook to a repo using go-github client : `"name" wasn't supplied. []`
 It's due to the field "name" in POST `/repos/:owner/:repo/hooks` API. The documentation at https://developer.github.com/v3/repos/hooks/#create-a-hook says that `name` is not required, whereas enterprise documentation at https://developer.github.com/enterprise/2.14/v3/repos/hooks/#create-a-hook says that both `name` and `config` are required. Hence I am able to add webhook using Postman but not using go-github client.

PR #1119, opened 2019-02-12
Allow any Payload during deployment creation

### LINKED ISSUE #1116: Deployment Payload field is double-encoded in responses
Hi all, I've run into slight ergonomics issue.

The Github deployments API has a field called `payload` which is an arbitrary JSON object encoded into a string.
```json
{
    "repo": "foo",
    "owner": "bar",
    "payload": "{\"run_acceptance_tests\": true}"
}
```

If I access this from, for example, Octokit in Ruby, I get the string I put in.

```ruby
payload = client.
  deployments(owner: "tragiclifestories", repo: "my-awesome-ruby-app").
  first.
  payload
# => "{\"run_acceptance_tests\": true}"

payload[0]
# => "{"
```

With go-github, however, the `Payload` field of the returned `github.Deployment` struct is double-encoded.

```go
deployments, _, _ := d.Repositories.ListDeployments(
	ctx,
	"tragiclifestories",
	"my-awesome-go-app",
)

payload := deployments[0].GetPayload()

fmt.Printf("%s\n", string(payload))     // `"{\"run_acceptance_tests\": true}"`
fmt.Printf("%s\n", string(payload[:1])) // `"`
```

This makes unmarshaling a bit more hassly than in comparable clients.

```
type myStruct struct {
	RunAcceptanceTests bool `json:"run_acceptance_tests"`
}

var payloadJSON string

err := json.Unmarshal(payload, payloadJSON)
if err != nil {
	// whatever ...
}

myPayload = new(myStruct)

err := json.Unmarshal(payloadJSON, myPayload)

if err != nil {
	// whatever ...
}
```

As is clear from the above, there is a functional workaround. However, it seems wrong to me that this field has one value in the HTTP API and most other client library implementations, and another in the equivalent Go data structure.

gmlewis:
Fixes #1116.

I tested this change with the following code:

```go
func createDeployment() {
	owner := "gmlewis"
	repo := "my-test-repo"
	foo := struct {
		Foo int    `json:"foo,omitempty"`
		Bar string `json:"bar,omitempty"`
	}{Foo: 1, Bar: "test create deployment with interface{}"}
	req := &github.DeploymentRequest{
		Ref:                  github.String("master"),
		TransientEnvironment: github.Bool(true),
		// Payload:              github.String(`{"foo":1}`),
		Payload: &foo,
	}
	ctx := context.Background()
	deployment, _, err := client.Repositories.CreateDeployment(ctx, owner, repo, req)
	if err != nil {
		log.Fatalf("Repositories.CreateDeployment returned error: %v", err)
	}
	goon.Dump(deployment)
}
```

and when listing the deployments created on multiple runs, I see:

```
2019/02/12 09:55:03 #0: payload={"foo":1,"bar":"test create deployment with interface{}"}
2019/02/12 09:55:03 #1: payload={"Foo":1}
2019/02/12 09:55:03 #2: payload="{\"foo\":1}"
```

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -2132,14 +2132,6 @@ func (d *DeploymentRequest) GetEnvironment() string {
 	return *d.Environment
 }
 
-// GetPayload returns the Payload field if it's non-nil, zero value otherwise.
-func (d *DeploymentRequest) GetPayload() string {
-	if d == nil || d.Payload == nil {
-		return ""
-	}
-	return *d.Payload
-}
-
 // GetProductionEnvironment returns the ProductionEnvironment field if it's non-nil, zero value otherwise.
 func (d *DeploymentRequest) GetProductionEnvironment() bool {
 	if d == nil || d.ProductionEnvironment == nil {
--- a/github/repos_deployments.go
+++ b/github/repos_deployments.go
@@ -32,15 +32,15 @@ type Deployment struct {
 
 // DeploymentRequest represents a deployment request
 type DeploymentRequest struct {
-	Ref                   *string   `json:"ref,omitempty"`
-	Task                  *string   `json:"task,omitempty"`
-	AutoMerge             *bool     `json:"auto_merge,omitempty"`
-	RequiredContexts      *[]string `json:"required_contexts,omitempty"`
-	Payload               *string   `json:"payload,omitempty"`
-	Environment           *string   `json:"environment,omitempty"`
-	Description           *string   `json:"description,omitempty"`
-	TransientEnvironment  *bool     `json:"transient_environment,omitempty"`
-	ProductionEnvironment *bool     `json:"production_environment,omitempty"`
+	Ref                   *string     `json:"ref,omitempty"`
+	Task                  *string     `json:"task,omitempty"`
+	AutoMerge             *bool       `json:"auto_merge,omitempty"`
+	RequiredContexts      *[]string   `json:"required_contexts,omitempty"`
+	Payload               interface{} `json:"payload,omitempty"`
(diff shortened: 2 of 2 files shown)


gmlewis:
Closing due to #1116 closing.

gmlewis:
Reopening for discussion of #1116.

gmlewis:
I think that before we proceed with this PR, it would be great if we could get some volunteers to test out this solution independently and see if it works for them.

@jrossiter - any volunteers?  :smile:

jrossiter:
@gmlewis As a workaround for this (it was a blocker for us) we implemented the above with `interface{}`. It worked for our simple use case of basic string key-value pairs.

gmlewis:
OK, thank you, @jrossiter.  Can you think of any downsides (apart from the breaking API) or regressions this PR might cause?

jrossiter:
Not that I can tell since it seems like developers will only be using `DeploymentRequest` as an argument for `Repositories.CreateDeployment`. The data would be read as raw JSON coming from a GitHub deployment payload afterwards.

gmlewis:
Thank you, @jrossiter!

gmlewis:
@wesleimp - if you have time to review this, that would be great.  Thank you!

For some reason, the GitHub UI is not allowing me to request you as a reviewer, so I'll just ask in a comment.  :smile:

wesleimp:
> 
> 
> @wesleimp - if you have time to review this, that would be great. Thank you!
> 
> For some reason, the GitHub UI is not allowing me to request you as a reviewer, so I'll just ask in a comment. 😄

@gmlewis no problem. 😄

gmlewis:
Thank you, @wesleimp!
Merging.

Issue #1120, opened 2019-02-15
Support Draft Pull Requests API Public Beta


dmitshur:
GitHub Developer API announcement: https://developer.github.com/changes/2019-02-14-draft-pull-requests/.

It looks like we'll need the following for this issue:

-	a new media type (`application/vnd.github.shadow-cat-preview+json`) must be added to `github.go`,
-	this media type must be added to the header of calls to all affected endpoints (list PRs, get a single PR),
-	a new `Draft` field must be added to the [`PullRequest`](https://godoc.org/github.com/google/go-github/github#PullRequest) struct that is returned from list/get PR endpoints,
-	~~and also to [`NewPullRequest`](https://godoc.org/github.com/google/go-github/github#NewPullRequest) struct used for creating a PR,~~ **Edit:** This doesn't need to be done; this isn't a part of the preview API at this time.
-	~~the [`PullRequestsService.Edit`](https://godoc.org/github.com/google/go-github/github#PullRequestsService.Edit) needs to be updated to allow the new `Draft` field to be passed~~ **Edit:** This doesn't need to be done; this isn't a part of the preview API at this time.

This would be a great PR for any new contributor to this repo or a new Go developer. All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

joshuabezaleel:
Hi @dmitshur , I am really new to Go and this repo but I tried to take a shot on this. I am sorry if there's anything wrong. Hope this works.

Looking forward to the review.
Thank you very much.

dmitshur:
> I am really new to Go and this repo but I tried to take a shot on this. I am sorry if there's anything wrong.

@joshuabezaleel Thank you for doing that! You don't have to worry, we have code review exactly so that we can improve the PR and get it into a mergeable state.

PR #1121, opened 2019-02-15
Support draft pull requests api public beta

joshuabezaleel:
Fixes: #1120 

* Adding new media type (`application/vnd.github.shadow-cat-preview+json`) must be added to `github.go`,
* Adding the type to the header of listPRs and get single PR endpoint call,
* Adding new Draft boolean field to the [`PullRequest`](https://godoc.org/github.com/google/go-github/github#PullRequest) struct.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -72,6 +72,10 @@ type PullRequest struct {
 	// ActiveLockReason is populated only when LockReason is provided while locking the pull request.
 	// Possible values are: "off-topic", "too heated", "resolved", and "spam".
 	ActiveLockReason *string `json:"active_lock_reason,omitempty"`
+
+	// IsDraft identifies if the pull request is a draft or not.
+	// https://developer.github.com/changes/2019-02-14-draft-pull-requests/
+	IsDraft             *bool      `json:"is_draft,omitempty"`
 }
 
 func (p PullRequest) String() string {
@@ -146,7 +150,7 @@ func (s *PullRequestsService) List(ctx context.Context, owner string, repo strin
 	}
 
 	// TODO: remove custom Accept header when this API fully launches.
-	acceptHeaders := []string{mediaTypeLabelDescriptionSearchPreview, mediaTypeLockReasonPreview}
+	acceptHeaders := []string{mediaTypeLabelDescriptionSearchPreview, mediaTypeLockReasonPreview, mediaTypeDraftPreview}
 	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
 
 	var pulls []*PullRequest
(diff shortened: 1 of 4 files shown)


dmitshur:
Thanks, this is a great start!

It looks like this PR currently includes commit 31ef9b1 that shouldn't be here. It's removing some code and isn't related to issue #1120.

You should also update the PR to include text such as "Fixes #1120." That way, when we merge this PR, it will automatically close the associated issue. GitHub has documentation on this at https://help.github.com/articles/closing-issues-using-keywords/.

github/pulls.go
```diff
@@ -72,6 +72,10 @@ type PullRequest struct {
 	// ActiveLockReason is populated only when LockReason is provided while locking the pull request.
 	// Possible values are: "off-topic", "too heated", "resolved", and "spam".
 	ActiveLockReason *string `json:"active_lock_reason,omitempty"`
+
+	// IsDraft identifies if the pull request is a draft or not.
+	// https://developer.github.com/changes/2019-02-14-draft-pull-requests/
+	IsDraft             *bool      `json:"is_draft,omitempty"`
```
dmitshur: Based on what I see at https://developer.github.com/v3/pulls/, it seems that the field is named exactly `draft`, not `is_draft`. Do you have another source suggesting it should be `is_draft` instead?

If not, please change the JSON tag to `draft`, and also rename the Go field to `Draft`, to match.

Additionally, this is a pretty simple boolean field, so I think we can just place it in the big group of fields at the top of this struct, rather than at the bottom, separately. How do you feel about placing it between the `User` and `Merged` fields?

github/repos_hooks_test.go
```diff
@@ -161,39 +124,3 @@ func TestRepositoriesService_DeleteHook_invalidOwner(t *testing.T) {
...
-	}
-}
-
-func TestRepositoriesService_TestHook_invalidOwner(t *testing.T) {
-	client, _, _, teardown := setup()
-	defer teardown()
-
-	_, err := client.Repositories.TestHook(context.Background(), "%", "%", 1)
-	testURLParseError(t, err)
-}
```
dmitshur: As mentioned earlier, this is out of place, let's get rid of it.

dmitshur:
Another step that needs to be done is to re-generate one of the files in this Go package by running the `go generate github.com/google/go-github/github` command. This is mentioned in step 5 of [CONTRIBUTING.md#submitting-a-patch](https://github.com/google/go-github/blob/master/CONTRIBUTING.md#submitting-a-patch).

github/pulls.go
joshuabezaleel: You're right, I thought it was `is_draft` from looking at this one.

joshuabezaleel:
Hi @dmitshur , I've made some changes. Looking forward to your review.
Also, I ran the commands that was written in the contributing guidelines and got this result. Is this the way they should work? 
image


Thank you very much!

gmlewis:
No, it looks like your generate didn't work since `github-accessors.go` was not updated with the new field.
It looks like maybe `GO111MODULE=on` might be set.  Could you please try `go generate ./...` from that same directory and see if that works?

joshuabezaleel:
Thank you very much @dmitshur and @gmlewis for the really kind responses! :slightly_smiling_face: 
Looking forward to more contributions in the future!

PR #966, opened 2018-08-05
Use PathEscape on search query instead of QueryEscape

gmlewis:
Fixes #963.

Issue #963, opened 2018-08-03
SearchService.Repositories doesn't work well


lizebang:
When I use SearchService.Repositories like this, there is a error.

```go
	client := github.NewClient(nil)
	rs, _, err := client.Search.Repositories(context.Background(), "language:golang+stars:>=200+pushed:>=2018-01-01",
		&github.SearchOptions{Sort: "stars", Order: "desc", ListOptions: github.ListOptions{Page: 2, PerPage: 100}})
	......
```

Error:

```shell
GET https://api.github.com/search/repositories?order=desc&page=2&per_page=100&q=language%3Agolang%2Bstars%3A%3E%3D200%2Bpushed%3A%3E%3D2018-01-01&sort=stars: 422 Validation Failed [{Resource:Search Field:q Code:invalid Message:None of the search qualifiers apply to this search type.}]
```

The right URL is `https://api.github.com/search/repositories?q=language:golang+pushed:%3E2018-01-01+stars:%3E=200&sort=stars&order=desc&page=2&per_page=100`, not `https://api.github.com/search/repositories?order=desc&page=2&per_page=100&q=language%3Agolang%2Bstars%3A%3E%3D200%2Bpushed%3A%3E%3D2018-01-01&sort=stars`.

gmlewis:
Ah, yes, you are quite right!  Thank you, @lizebang!

Looking at `search.go`, line 224, we set the query parameter like this: `params.Set("q", parameters.Query)` and then on line 228, we encode the parameters with `params.Encode`.

I tested this in the embedded examples playground here: https://golang.org/pkg/net/url/#Values
(result: https://play.golang.org/p/ye2rk_hq9zG ) and confirmed that `Values.Encode` does not first call [`QueryEscape`](https://golang.org/pkg/net/url/#QueryEscape).

It looks to me like we should call `QueryEscape` on line 224.

Let's open this up for discussion in case my analysis is incorrect since this would be a breaking change to the API.  I would like to hear if maybe existing users of this package simply call `QueryEscape` themselves before calling any of the search functions.

lizebang:
Thank you for your reply, @gmlewis 

`Values.Encode` calls [`QueryEscape`](https://golang.org/pkg/net/url/#QueryEscape), but `parameters.Query` should not be escaped. If `parameters.Query` is escaped, GitHub will think this is an invalid field and return `422 Validation Failed`.

Is this correct?

PR #1125, opened 2019-02-20
Use url.QueryEscape instead of url.PathEscape to escape query string

### LINKED ISSUE #1122: Question about query with date that includes "+"
Hello!

I tried to use `Search.Repositories` function.
I see it almost works well, but I tried to use query string `created:` that includes "+" doesn't work well.
Could you please give me some advice to use query in such situation.

Here's snippet I tried:

```go
// c is *github.Client
result, resp, err := c.Search.Repositories(
            context.Background(),                                                                                                       | 18             context.Background(),
            "user:pankona created:2019-01-18T17:35:34+09:00..2019-02-18T17:35:34+09:00 fork:true",                                      | 19             "user:pankona created:2019-01-18T17:35:34+09:00..2019-02-18T17:35:34+09:00 fork:true",
            &github.SearchOptions{                                                                                                      | 20             &github.SearchOptions{
                ListOptions: github.ListOptions{                                                                                        | 21                 ListOptions: github.ListOptions{
                    PerPage: 100,                                                                                                       | 22                     PerPage: 100,
                    Page:    page,                                                                                                      | 23                     Page:    page,
                }})  
```

According to result of `httpuitl.DumpRequestOut`, the escaped query seems as follows:
`GET /search/repositories?q=user:pankona+created:2019-01-18T17:35:34+09:00..2019-02-18T17:35:34+09:00+fork:true&page=1&per_page=100 HTTP/1.1`

I expects 1 results by this search but it returns 0.

(I think "+" included in query's date part is not escaped correctly. I guess it should be `%2B`. Is my assuming correct?)

pankona:
Fixes #1122.

- Quit replacing white space with "+"
- Use url.QueryEscape instead of url.PathEscape to escape query string

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
-	"strings"
 
 	qs "github.com/google/go-querystring/query"
 )
@@ -223,11 +222,10 @@ func (s *SearchService) search(ctx context.Context, searchType string, parameter
 	if err != nil {
 		return nil, err
 	}
-	q := strings.Replace(parameters.Query, " ", "+", -1)
 	if parameters.RepositoryID != nil {
 		params.Set("repository_id", strconv.FormatInt(*parameters.RepositoryID, 10))
 	}
-	query := "q=" + url.PathEscape(q)
+	query := "q=" + url.QueryEscape(parameters.Query)
 	if v := params.Encode(); v != "" {
 		query = query + "&" + v
 	}
--- a/github/search.go
+++ b/github/search.go
@@ -216,6 +216,9 @@ func (s *SearchService) Labels(ctx context.Context, repoID int64, query string, 
 
 // Helper function that executes search queries against different
 // GitHub search types (repositories, commits, code, issues, users, labels)
+//
+// If *searchParameters.Query includes multiple condition, it MUST NOT include "+" as condition separator.
+// For example, querying with "language:c++" and "leveldb", then *searchParameters.Query should be "language:c++ leveldb" but not "language:c+++leveldb".
 func (s *SearchService) search(ctx context.Context, searchType string, parameters *searchParameters, opt *SearchOptions, result interface{}) (*Response, error) {
 	params, err := qs.Values(opt)
 	if err != nil {
--- a/github/search.go
+++ b/github/search.go
@@ -217,8 +217,8 @@ func (s *SearchService) Labels(ctx context.Context, repoID int64, query string, 
 // Helper function that executes search queries against different
 // GitHub search types (repositories, commits, code, issues, users, labels)
 //
-// If *searchParameters.Query includes multiple condition, it MUST NOT include "+" as condition separator.
-// For example, querying with "language:c++" and "leveldb", then *searchParameters.Query should be "language:c++ leveldb" but not "language:c+++leveldb".
+// If searchParameters.Query includes multiple condition, it MUST NOT include "+" as condition separator.
+// For example, querying with "language:c++" and "leveldb", then searchParameters.Query should be "language:c++ leveldb" but not "language:c+++leveldb".
 func (s *SearchService) search(ctx context.Context, searchType string, parameters *searchParameters, opt *SearchOptions, result interface{}) (*Response, error) {
 	params, err := qs.Values(opt)
 	if err != nil {
--- a/github/search.go
+++ b/github/search.go
@@ -8,7 +8,6 @@ package github
 import (
 	"context"
 	"fmt"
-	"net/url"
 	"strconv"
 
 	qs "github.com/google/go-querystring/query"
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @pankona.

Can you please take a look at #966 and #963 for me?
I obviously didn't fix the original problem correctly since it is still an issue.
Yet the original code used this (see https://github.com/google/go-github/blame/fbd7fb9c7f3d27aa691e8421690e00be916f8d8d/github/search.go):

```go
import (
	qs "github.com/google/go-querystring/query"
)
...
	params, err := qs.Values(opt)
	if err != nil {
		return nil, err
	}
	params.Set("q", parameters.Query)
	if parameters.RepositoryID != nil {
		params.Set("repository_id", strconv.FormatInt(*parameters.RepositoryID, 10))
	}
	u := fmt.Sprintf("search/%s?%s", searchType, params.Encode())
```

and I'm wondering if the correct fix is to make the changes in the upstream `go-querystring/query` package instead?

Thoughts?

pankona:
@gmlewis 

I've checked #966 and #963 .
I've pushed a new commit 6bc0bca . I think go-querystring has no problem. Could you please check current patch is as you expected?

gmlewis:
Thank you, @pankona!  I'll have to take time and look at this tonight... gotta run... but in a nutshell, was #966 just broken? I'm curious how this current PR differs from the original, but don't have time right now to look... I just know there was a problem in #963 that needed fixing and hope that this addresses that issue as well.

pankona:
@gmlewis 

I've checked snippet on #963 (as follows)

```go
rs, _, err := client.Search.Repositories(context.Background(),
		"language:golang+stars:>=200+pushed:>=2018-01-01",
		&github.SearchOptions{Sort: "stars", Order: "desc",
				ListOptions: github.ListOptions{Page: 2, PerPage: 100}})
```

This query string includes "+" (instead of white space). I think this is NOT acceptable query string.
- `"language:golang+stars:>=200+pushed:>=2018-01-01"` doesn't work with GitHub's search input form
- [According to GitHub's search API documentation](https://developer.github.com/v3/search/#constructing-a-search-query), there're some examples that include "+" instead of white space, I think the white spaces are already escaped.
- I think the query string written above should be `"language:golang stars:>=200 pushed:>=2018-01-01"`

gmlewis:
> This query string includes "+" (instead of white space). I think this is NOT acceptable query string.
...although it worked when using `curl`... does your recommended change also work against the GitHub v3 API? If so, then I'm happy to incorporate your solution.

But are you saying here that the original code was correct?

pankona:
@gmlewis 

> But are you saying here that the original code was correct?

Yes. I think original code was correct.

In my understanding, escaping of query string is responsibility of go-github itself, but not of library user. Thus a query string that includes "+" instead of white space may return unexpected result.

> curl

- Here's curl command using "+" instead of white spaces (got OK)

```bash
curl -X GET "https://api.github.com/search/repositories?order=desc&page=2&per_page=100&q=language%3Agolang+stars%3A%3E%3D200+pushed%3A%3E%3D2018-01-01&sort=stars"
```
- Here's curl command using white spaces directly (got error)

```bash
curl -X GET "https://api.github.com/search/repositories?order=desc&page=2&per_page=100&q=language%3Agolang stars%3A%3E%3D200 pushed%3A%3E%3D2018-01-01&sort=stars"
```

In case of using curl, escaping of query string is responsibility of user. Thus former case goes okay.

gmlewis:
Thank you for the analysis, @pankona! It is greatly appreciated!

@gauntface - do you agree?

TL;DR: #966 was the wrong fix for #963 - the query in #963 was malformed. This PR goes back to the original implementation.

gauntface:
Ok, my understanding from everything is that this is the correct change.

Forcing users to **not** do this:

```
client.Search.Repositories(ctx, "language:golang+stars:>=200+pushed:>=2018-01-01", ...)
```

And instead do this:

```
client.Search.Repositories(ctx, "language:golang stars:>=200 pushed:>=2018-01-01", ...)
```

I like this approach just from the perspective of being close to the input you would provide to the Github.com search box.

Regarding this PR, @pankona could you add a comment to the `search()` method calling out the example input of " " vs "+" and highlight that we expect " ".

pankona:
@gauntface 
Thank you for checking!

> comment

okay, I will.

github/search.go
```diff
@@ -218,20 +216,19 @@ func (s *SearchService) Labels(ctx context.Context, repoID int64, query string,
 
 // Helper function that executes search queries against different
 // GitHub search types (repositories, commits, code, issues, users, labels)
+//
+// If *searchParameters.Query includes multiple condition, it MUST NOT include "+" as condition separator.
```
gmlewis: nit: please remove leading `*` as we typically do not put these for pointers in Godocs.

pankona: Thanks! fixed on 63ab8d9

pankona:
@gmlewis 

understood, I will!

pankona:
@gmlewis 

At b94a2b5, I introduced `c++` in unit tests to make escape result more clear. How about it?

gmlewis:
Perfect, @pankona !  Thank you!

Awaiting @gauntface LGTM before merging.

pankona:
Gently ping to @gauntface :)

pankona:
Thank you @gauntface for your confirmation!

@gmlewis 

If there's no concern for merging, could you please proceed?

gmlewis:
Thank you, @gauntface!
Merging.

Issue #1126, opened 2019-02-21
Support Webhooks without a Secret


jbrockopp:
I have been unable to find a way of using this library to handle webhooks that do not contain a [secret token](https://developer.github.com/webhooks/securing/#setting-your-secret-token).

Source code:

https://github.com/google/go-github/blob/00add49aab2728f7c246b27be612b002b22e7f94/github/messages.go#L142-L183

Currently, the function expects `secretKey []byte` as a parameter. I believe the simplest way to complete this would be to update the [relevant code](https://github.com/google/go-github/blob/master/github/messages.go#L178-L181) to look like:

```go
	if secretKey != nil {
		sig := r.Header.Get(signatureHeader)
		if err := ValidateSignature(sig, body, secretKey); err != nil {
			return nil, err
		}
	}

	return payload, nil
```

Playing off the provided example usage for the `ValidatePayload` function, it could be consumed like:

```go
func (s *GitHubEventMonitor) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	// Contains a webhook secret token
	payload, err := github.ValidatePayload(r, s.webhookSecretKey)
	if err != nil { ... }
	// Does not contain a webhook secret token
	payload, err := github.ValidatePayload(r, nil)
	if err != nil { ... }
	// Process payload...
}
```

Is there a reason why this functionality isn't offered today?

NOTE: I could see a reasoning that the library doesn't currently support this because of security related reasons. However, I'd like to propose we leave that up to the consumer of the library to determine the security standards for their project.

gmlewis:
Let's step back for one moment... can you please describe what your use case is?

If I'm not mistaken, GitHub requires that you set up a secret key when creating the webhook and your webhook should always be validated, otherwise any system could hit your webhook endpoint with garbage. OK, I see you make the "I want to handle my own security" argument... but I would still like to understand why you would want to remove this validation?

If we head down this potentially dangerous route, we have to make it extremely clear to all users of this package that disabling the validation check is highly discouraged.

willnorris:
I don't *think* this was a conscious decisions to always require a webhook secret.  In our initial use case, we always had one, and so maybe I just assumed folks would always want one present?  I don't remember.

But I'm personally fine with the suggested change, as it is quite minimal.  And to your point @gmlewis, we can add some clear documentation that this is probably a really bad idea unless you know what you're doing.  But there's probably no reason for us to actually prevent it.

gmlewis:
[...]

PR #1127, opened 2019-02-25
No webhook secret

jbrockopp:
Fixes https://github.com/google/go-github/issues/1126

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -130,6 +130,8 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 // The Content-Type header of the payload can be "application/json" or "application/x-www-form-urlencoded".
 // If the Content-Type is neither then an error is returned.
 // secretKey is the GitHub Webhook secret message.
+// If your webhook does not contain a secret message, you can pass nil. This
+// is intended for local development purposes only and all webhooks should contain a secret message.
 //
 // Example usage:
 //
@@ -175,10 +177,15 @@ func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err err
 		return nil, fmt.Errorf("Webhook request has unsupported Content-Type %q", ct)
 	}
 
-	sig := r.Header.Get(signatureHeader)
-	if err := ValidateSignature(sig, body, secretKey); err != nil {
-		return nil, err
+	// Only validate the signature if a secret message exists. This is intended for
+	// local development only and all webhooks should contain a secret message.
+	if secretKey != nil {
+		sig := r.Header.Get(signatureHeader)
+		if err := ValidateSignature(sig, body, secretKey); err != nil {
+			return nil, err
+		}
 	}
+
 	return payload, nil
 }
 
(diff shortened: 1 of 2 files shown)


jbrockopp:
I signed it!

github/messages.go
```diff
@@ -175,10 +177,15 @@ func ValidatePayload(r *http.Request, secretKey []byte) (payload []byte, err err
 		return nil, fmt.Errorf("Webhook request has unsupported Content-Type %q", ct)
 	}
 
-	sig := r.Header.Get(signatureHeader)
-	if err := ValidateSignature(sig, body, secretKey); err != nil {
-		return nil, err
+	// Only validate the signature if a secret message exists. This is intended for
+	// local development only and all webhooks should contain a secret message.
+	if secretKey != nil {
```
gmlewis: In Go, it is more idiomatic with slices to check against a length of zero which will work for both nil and a zero-length slice. Checking against `nil` is discouraged because it doesn't work for a zero-length slice.

So please change this to `if len(secretKey) > 0 {`.

github/messages_test.go
```diff
@@ -110,7 +110,7 @@ func TestValidatePayload_FormGet(t *testing.T) {
 
 	// check that if payload is invalid we get error
 	req.Header.Set(signatureHeader, "invalid signature")
-	if _, err = ValidatePayload(req, nil); err == nil {
+	if _, err = ValidatePayload(req, []byte{}); err == nil {
```
gmlewis: This and line 143 will now need to be changed... since it is checking for an invalid signature, please just go ahead and use something like `[]byte{0}` here and below.

github/messages.go
```diff
@@ -130,6 +130,8 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 // The Content-Type header of the payload can be "application/json" or "application/x-www-form-urlencoded".
 // If the Content-Type is neither then an error is returned.
 // secretKey is the GitHub Webhook secret message.
+// If your webhook does not contain a secret message, you can pass nil. This
```
gmlewis: How about `...pass nil or an empty slice.` ?

jbrockopp:
@gmlewis I have made changes per your feedback. Please let me know if anything else is required 👍

github/messages.go
```diff
@@ -130,6 +130,8 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 // The Content-Type header of the payload can be "application/json" or "application/x-www-form-urlencoded".
 // If the Content-Type is neither then an error is returned.
 // secretKey is the GitHub Webhook secret message.
+// If your webhook does not contain a secret message, you can pass nil or an empty slice.
+// This is intended for local development purposes only and all webhooks should contain a secret message.
```
kshitij10496: minor nit: I believe we should maintain semantic consistency between GitHub documentation and the library's documentation so that the user can easily navigate between both of them with ease. Thus, I request that we update this to read `"... all webhooks should ideally set up a secret token."`

Thoughts @gmlewis @jbrockopp

gmlewis: Sounds great!  Let's please go ahead and address this change in this PR.

jbrockopp: @kshitij10496 @gmlewis sure I can do that. I was just following the previously used verbiage in the existing code.

> // secretKey is the GitHub Webhook secret message.

jbrockopp:
@kshitij10496 @gmlewis I have made changes per your feedback. Please let me know if anything else is required 👍

gmlewis:
Thank you, @gauntface!
Merging.

PR #1130, opened 2019-03-07
Fix org webhook creation

chrisforrette:
Hello! This is fixing exactly what #1117 fixed but for org hooks too. The discussion on that PR implies that this only affects GitHub Enterprise but I've been experiencing issues with this in plain ol' GitHub. I suspect this has something to do with this deprecation timeline taking effect at the end of January: https://developer.github.com/v3/guides/replacing-github-services/#deprecation-timeline


chrisforrette:
Thanks y'all!

Issue #1131, opened 2019-03-10
Missing head_commit field in CheckSuiteEvent


binkkatal:
The Check suite event also contains `head_commit` in the `check_suite` object which is missing in the current `go-github` version.
```go
{
  "action": "requested",
  "check_suite": {
    "id": 74624585,
    "node_id": "MDEwOkNoZWNrU3VpdGU3NDYyNDU4NQ==",
    "head_branch": "feature/razorops",
    "head_sha": "64c9a8b532e040d6d2ccab1366b4ee20174f43dc",
    "status": "queued",
    "conclusion": null,
    "url": "https://api.github.com/repos/binkkatal/revel-websokets-chat/check-suites/74624585",
    "before": "9c2c43b2129727547099679e0795ef482aae0771",
    "after": "64c9a8b532e040d6d2ccab1366b4ee20174f43dc",
    "pull_requests": [

    ],
    "app": {  ...  },
    "created_at": "2019-03-06T08:14:10Z",
    "updated_at": "2019-03-06T08:14:10Z",
    "latest_check_runs_count": 0,
    "check_runs_url": "https://api.github.com/repos/binkkatal/revel-websokets-chat/check-suites/74624585/check-runs",
    "head_commit": { ... },
}
  "repository": { ... },
  "org":{ ... },
  "sender": { ... },
  "installation": { ... }
}

```

PR #1132, opened 2019-03-10
Use embeded CheckSuite and HeadCommit field for CheckSuiteEvent

binkkatal:
Connects https://github.com/google/go-github/issues/1131

The `check_suite` object from github's check_suite event also includes `head_commit`, which is not present in the `check_suite` object.

The `CheckSuiteEvent` also uses `CheckSuite` as one of its fields in the struct which does not has the head_commit, and therefore is not parsed during json parsing in the current `CheckSuiteEvent`, hence `head_commit` data is unavailable from the `CheckSuiteEvent` struct.

~To also parse the `head_commit`, `CheckSuiteEvent` should use an embeded field for `CheckSuite` with `HeadCommit`, so that all the details are received from the github api.~

To also parse the `head_commit`, 
add the missing field `HeadCommit` in the `CheckSuite` struct with a comment stating that the field is only populated by a webhook event.

```go
type CheckSuite struct {
	ID           *int64         `json:"id,omitempty"`
	NodeID       *string        `json:"node_id,omitempty"`
	HeadBranch   *string        `json:"head_branch,omitempty"`
	HeadSHA      *string        `json:"head_sha,omitempty"`
	URL          *string        `json:"url,omitempty"`
	BeforeSHA    *string        `json:"before,omitempty"`
	AfterSHA     *string        `json:"after,omitempty"`
	Status       *string        `json:"status,omitempty"`
	Conclusion   *string        `json:"conclusion,omitempty"`
	App          *App           `json:"app,omitempty"`
	Repository   *Repository    `json:"repository,omitempty"`
	PullRequests []*PullRequest `json:"pull_requests,omitempty"`

	HeadCommit *Commit `json:"head_commit,omitempty"`
}
```

~This would remove a function from `github-accessors.go`~

This would add a new accessor function for `head_commit` .
```go
// GetHeadCommit returns the HeadCommit field.
func (c *CheckSuite) GetHeadCommit() *Commit {
	if c == nil {
		return nil
	}
	return c.HeadCommit
}
```

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -37,7 +37,10 @@ type CheckRunEvent struct {
 //
 // GitHub API docs: https://developer.github.com/v3/activity/events/types/#checksuiteevent
 type CheckSuiteEvent struct {
-	CheckSuite *CheckSuite `json:"check_suite,omitempty"`
+	CheckSuite struct {
+		*CheckSuite
+		HeadCommit *Commit `json:"head_commit,omitempty"`
+	} `json:"check_suite,omitempty"`
 	// The action performed. Can be "completed", "requested" or "re-requested".
 	Action *string `json:"action,omitempty"`
 
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -37,10 +37,7 @@ type CheckRunEvent struct {
 //
 // GitHub API docs: https://developer.github.com/v3/activity/events/types/#checksuiteevent
 type CheckSuiteEvent struct {
-	CheckSuite struct {
-		*CheckSuite
-		HeadCommit *Commit `json:"head_commit,omitempty"`
-	} `json:"check_suite,omitempty"`
+	CheckSuite *CheckSuite `json:"check_suite,omitempty"`
 	// The action performed. Can be "completed", "requested" or "re-requested".
 	Action *string `json:"action,omitempty"`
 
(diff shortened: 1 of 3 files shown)


binkkatal:
Hey google/Go-Github,

I have signed the agreement.

Thanks and regards

Ranbir Singh

github/event_types.go
```diff
@@ -37,7 +37,10 @@ type CheckRunEvent struct {
 //
 // GitHub API docs: https://developer.github.com/v3/activity/events/types/#checksuiteevent
 type CheckSuiteEvent struct {
-	CheckSuite *CheckSuite `json:"check_suite,omitempty"`
+	CheckSuite struct {
+		*CheckSuite
+		HeadCommit *Commit `json:"head_commit,omitempty"`
```
gmlewis: Instead of doing it this way, let's please add the additional `HeadCommit` field to the existing `CheckSuite` struct.

We have done this in many other locations where a particular field is only populated in a webhook event.

When adding the field to the existing struct, please add a comment (similar to the ones that already exist for other fields like this) that describes how the field is currently only being populated in the case of receiving the struct from a webhook event. You should be able to find examples... but please let me know if you can't and I'll find some for you.

Thank you, @binkkatal!

Also, just FYI... please do not use force commits in PRs to this repo so that reviewers can easily see the changes since the last review. Thanks!

binkkatal:
@gmlewis , have updated the changes,

Please review.

Thanks

binkkatal:
@gauntface , @gmlewis , can i squash the commits now ?

PR #1141, opened 2019-04-01
Add comment to describe possible status values for CommitFile

nschhina:
It'd be great if we have this comment to describe the type of values CommitFile status might return. https://developer.github.com/v3/pulls/#list-pull-requests-files doesn't expand on CommitFile status much. Had to make a separate GET request to understand what values are supported. For example one might choose to compare commitFile.GetStatus() to "deleted" instead of "removed" as it appears much more natural.

EDIT: I can't sign the CLA, would have to go through a long process, can someone add this in a different PR?

### DIFF
--- a/github/repos_commits.go
+++ b/github/repos_commits.go
@@ -54,6 +54,8 @@ type CommitFile struct {
 	Additions        *int    `json:"additions,omitempty"`
 	Deletions        *int    `json:"deletions,omitempty"`
 	Changes          *int    `json:"changes,omitempty"`
+
+	// Status returns status of the changed file. Possible values include: "added", "modified", and "removed".
 	Status           *string `json:"status,omitempty"`
 	Patch            *string `json:"patch,omitempty"`
 	BlobURL          *string `json:"blob_url,omitempty"`


gmlewis:
Thanks, @nschhina.

I'm looking through the GitHub Developer API documentation, specifically:
https://developer.github.com/v3/repos/commits/

and am not finding where the possible values of `status` are documented.
Do you have a link that definitively states what values are possible?

nschhina:
Hi @gmlewis. That is exactly why I made this PR as the GitHub api documentation doesn't have any info on status whatsoever. I've tested out the API endpoint  /repos/:owner/:repo/pulls/:number/files on a few PRs and added, modified and removed are the only values which I observed. Also only 3 values makes sense as well, either the commit file was deleted, added or modified.

gmlewis:
OK, thanks. Let's do this.

I'd like to request that someone please contact support@github.com and ask them for a definitive answer on the possible values of commit `status` and preferably update their GitHub v3 API documentation. We can then confidently move forward with this change in a separate PR.

GitHub tech support has, in my experience, a top-tier team of developers who have been very responsive, friendly, and helpful.

Whoever volunteers to contact them, please update this PR with a comment stating that you are waiting to hear back from them, and then we can move forward.

Thank you!

gmlewis:
Closing abandoned PR with `cla:no`.

PR #1142, opened 2019-04-02
add SignKey gpg entity param to allow easier pgp signing of commits

anandkumarpatel:
This PR creates a new param for the `Commit` struct called `SignKey`.
`SignKey` is an `openpgp.Entity` type which is the standard way of reading and validating pgp keys.

When specified the key will be used to sign the commit. This makes it so the user does not have to sign the key externally and duplicate some of the logic the `CreateCommit` function does.

To maintain backward compatibility, if the `Verification.Signature` is defined `SignKey` will have no effect. 

This provides similar functionality that go-git has: https://github.com/src-d/go-git/commit/7b6c1266556f59ac436fada3fa6106d4a84f9b56

Note from the git docs: https://github.com/octokit/routes/blob/152c0b5d156844e17a1805d0509c9a0b7cfc848f/routes/api.github.com/git.json#L282
```
To pass a `signature` parameter, you need to first manually create a valid PGP signature, which can be complicated. You may find it easier to [use the command line]
```
So I wanted to make it less complicated 😄

### DIFF
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -6,9 +6,12 @@
 package github
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"time"
+
+	"golang.org/x/crypto/openpgp"
 )
 
 // SignatureVerification represents GPG signature verification.
@@ -37,6 +40,11 @@ type Commit struct {
 	// is only populated for requests that fetch GitHub data like
 	// Pulls.ListCommits, Repositories.ListCommits, etc.
 	CommentCount *int `json:"comment_count,omitempty"`
+
+	// SignKey denotes a key to sign the commit with. If not nil this key will
+	// be used to sign the commit. The private key must be present and already
+	// decrypted. Ignored if Verification.Signature is defined.
+	SignKey *openpgp.Entity
 }
 
 func (c Commit) String() string {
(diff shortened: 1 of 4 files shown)


gmlewis:
Thank you, @anandkumarpatel.

In the past, when "helper" functions have been suggested for this repo, we have typically requested that they be put in their own separate repo for a few reasons: it keeps the API surface area of this repo smaller, reduces the maintenance burden on this repo, and keeps its list of external dependencies smaller.

However, I can see how this can be useful, and I'm game to include it in this repo.

I'd like @willnorris and @gauntface to weigh in on their thoughts for including something like this first.

github/git_commits.go
```diff
@@ -37,6 +40,11 @@ type Commit struct {
 	// is only populated for requests that fetch GitHub data like
 	// Pulls.ListCommits, Repositories.ListCommits, etc.
 	CommentCount *int `json:"comment_count,omitempty"`
+
+	// SignKey denotes a key to sign the commit with. If not nil this key will
+	// be used to sign the commit. The private key must be present and already
+	// decrypted. Ignored if Verification.Signature is defined.
+	SignKey *openpgp.Entity
```
willnorris: nit: probably calling this `SigningKey` ?

Also, add a `json:"-"` tag to explicitly note that this does not part of the GitHub representation of this object.

gmlewis:
Closing abandoned PR.

anandkumarpatel:
@gmlewis Sorry, forgot about this one. 
I rebased and updated my branch based on @willnorris feedback. Is it possible to reopen? or should I make another PR?

github/git_commits.go
anandkumarpatel: Sounds good, Updated.

gmlewis:
Oh, sure... no problem!  I was just trying to do some Spring Cleaning in June.  :smile:

gmlewis:
Screenshot from 2019-06-14 09-40-42

Whups, it's not allowing me to reopen because of a force push. Can you please open another PR with these changes and we can run with that one?

anandkumarpatel:
No problem: Continued here: https://github.com/google/go-github/pull/1198

PR #1147, opened 2019-04-05
Added support for dismissed_review IssueEvent

### LINKED ISSUE #827: Support dismissed_review IssueEvent

vaibhavsingh97:
fixes #827

### DIFF
--- a/github/issues_events.go
+++ b/github/issues_events.go
@@ -69,14 +69,25 @@ type IssueEvent struct {
 	Issue     *Issue     `json:"issue,omitempty"`
 
 	// Only present on certain events; see above.
-	Assignee    *User        `json:"assignee,omitempty"`
-	Assigner    *User        `json:"assigner,omitempty"`
-	CommitID    *string      `json:"commit_id,omitempty"`
-	Milestone   *Milestone   `json:"milestone,omitempty"`
-	Label       *Label       `json:"label,omitempty"`
-	Rename      *Rename      `json:"rename,omitempty"`
-	LockReason  *string      `json:"lock_reason,omitempty"`
-	ProjectCard *ProjectCard `json:"project_card,omitempty"`
+	Assignee        *User            `json:"assignee,omitempty"`
+	Assigner        *User            `json:"assigner,omitempty"`
+	CommitID        *string          `json:"commit_id,omitempty"`
+	Milestone       *Milestone       `json:"milestone,omitempty"`
+	Label           *Label           `json:"label,omitempty"`
+	Rename          *Rename          `json:"rename,omitempty"`
+	LockReason      *string          `json:"lock_reason,omitempty"`
+	ProjectCard     *ProjectCard     `json:"project_card,omitempty"`
+	DismissedReview *DismissedReview `json:"dismissed_review,omitempty"`
+}
+
+// DismissedReview represents details for 'dismissed_review' events
+type DismissedReview struct {
+	// State represents the state of the dismissed review and
+	// possible state values are: commented, approved and changes_requested
+	State            *string `json:"state,omitempty"`
+	ReviewID         *int64  `json:"review_id,omitempty"`
+	DismissalMessage *string `json:"dismissal_message,omitempty"`
+	DismissalCommit  *string `json:"dismissal_commit_id,omitempty"`
 }
 
 // ListIssueEvents lists events for the specified issue.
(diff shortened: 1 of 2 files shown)


vaibhavsingh97:
@gmlewis Please review 😄

github/issues_events.go
```diff
@@ -69,14 +69,25 @@ type IssueEvent struct {
...
+	CommitID        *string          `json:"commit_id,omitempty"`
+	Milestone       *Milestone       `json:"milestone,omitempty"`
+	Label           *Label           `json:"label,omitempty"`
+	Rename          *Rename          `json:"rename,omitempty"`
+	LockReason      *string          `json:"lock_reason,omitempty"`
+	ProjectCard     *ProjectCard     `json:"project_card,omitempty"`
+	DismissedReview *DismissedReview `json:"dismissed_review,omitempty"`
+}
+
+// DismissedReview represents details for 'dismissed_review' events
```
gmlewis: Please terminate comments with periods... here and below.

github/issues_events.go
```diff
@@ -69,14 +69,25 @@ type IssueEvent struct {
...
+type DismissedReview struct {
+	// State represents the state of the dismissed review and
```
gmlewis: How about:

```go
// State represents the state of the dismissed review.
// Possible values are: "commented", "approved", and "changes_requested".
```

github/issues_events.go
```diff
@@ -69,14 +69,25 @@ type IssueEvent struct {
...
+	// possible state values are: commented, approved and changes_requested
+	State            *string `json:"state,omitempty"`
+	ReviewID         *int64  `json:"review_id,omitempty"`
+	DismissalMessage *string `json:"dismissal_message,omitempty"`
+	DismissalCommit  *string `json:"dismissal_commit_id,omitempty"`
```
gmlewis: s/DismissalCommit/DismissalCommitID/

github/issues_events.go
```diff
@@ -69,14 +69,25 @@ type IssueEvent struct {
...
-	ProjectCard *ProjectCard `json:"project_card,omitempty"`
+	Assignee        *User            `json:"assignee,omitempty"`
+	Assigner        *User            `json:"assigner,omitempty"`
+	CommitID        *string          `json:"commit_id,omitempty"`
+	Milestone       *Milestone       `json:"milestone,omitempty"`
+	Label           *Label           `json:"label,omitempty"`
+	Rename          *Rename          `json:"rename,omitempty"`
+	LockReason      *string          `json:"lock_reason,omitempty"`
+	ProjectCard     *ProjectCard     `json:"project_card,omitempty"`
+	DismissedReview *DismissedReview `json:"dismissed_review,omitempty"`
```
gmlewis: Please add documentation for this new event type in the comments above after line 65. Something like:

```go
//    review_dismissed
//      The review was dismissed and `DismissedReview` will be populated below.
```

github/issues_events.go
vaibhavsingh97: I had updated the PR 👍

github/issues_events.go
vaibhavsingh97: Sure, will keep in mind in future 😄

github/issues_events.go
vaibhavsingh97: Updated the PR

github/issues_events.go
vaibhavsingh97: Updated the PR

vaibhavsingh97:
@gmlewis I had updated the PR, please review 😄

github/issues_events.go
```diff
@@ -63,6 +63,8 @@ type IssueEvent struct {
 	//     head_ref_deleted, head_ref_restored
 	//       The pull request’s branch was deleted or restored.
 	//
+	//    review_dismissed
+	//      The review was dismissed and `DismissedReview` will be populated below.
```
gmlewis: Just to be consistent, let's please add back in the one more blank line comment here, like line 65 above.

vaibhavsingh97:
@gmlewis updated the PR 😄

gmlewis:
Thank you, @gauntface!
Merging.

Issue #1155, opened 2019-04-17
SetCheckSuitePreferences returns 422


annaliseirby:
**Go version**: 1.11.4
_Please note_: I substituted OWNER, REPO, and APP_ID for the actual string values below.

**Summary**: SetCheckSuitePreferences call is returning a 422 Invalid Request because of a bug in the json encoding of the [CheckSuitePreferenceOptions](https://github.com/google/go-github/blob/a8383c7edb4ac91a58ae2b931c0cc32d69c157fc/github/checks.go#L357) struct.

**Details**:
When I called SetCheckSuitePreferences, I got the following response:

```
2019/04/16 20:47:16 Failed to set check suite preferences : PATCH https://api.github.com/repos/OWNER/REPO/check-suites/preferences: 422 Invalid request.

For 'properties/auto_trigger_checks', {"auto_trigger_checks"=>[{"app_id"=>APP_ID, "setting"=>false}]} is not an array.
"app_id", "setting" weren't supplied. []
exit status 1
make: *** [server] Error 1
```

While debugging, I noticed that the [TestChecksService_SetCheckSuitePreferences](https://github.com/google/go-github/blob/a8383c7edb4ac91a58ae2b931c0cc32d69c157fc/github/checks_test.go#L392) method has testMethod() and testHeader(), but does not include testBody(). When this line is added to the test:
```
testBody(t, r, `{"auto_trigger_checks":[{"app_id":2,"setting":false}]}`+"\n")
```
... the test should pass if the SetCheckSuitePreferences method was working as intended. However, it fails with the following error:
```
--- FAIL: TestChecksService_SetCheckSuitePreferences (0.00s)
    github_test.go:134: request Body is {"auto_trigger_checks":{"auto_trigger_checks":[{"app_id":2,"setting":false}]}}
        , want {"auto_trigger_checks":[{"app_id":2,"setting":false}]}
```

**More Details**:
I also noticed that the encoded json request has two layers of auto_trigger_check fields, when it should only have one. In other words, according to the [Github Checks API docs](https://developer.github.com/v3/checks/suites/#set-preferences-for-check-suites-on-a-repository), it should look like this:
```
{"auto_trigger_checks":[{"app_id":2,"setting":false}]}
```
But instead, it looks like this:
```
{"auto_trigger_checks":{"auto_trigger_checks":[{"app_id":2,"setting":false}]}}
```

gmlewis:
[...]

PR #1156, opened 2019-04-17
Fix set check suite preferences options

ardakuyumcu:
Fixes https://github.com/google/go-github/issues/1155


ardakuyumcu:
Thank you for the swift reviews!

gmlewis:
Thank you, @gauntface!
Merging and then bumping versions.

PR #1169, opened 2019-05-11
Add RequestedTeam to PullRequestEvent

kremlinkev:
Adds `RequestedTeam` to the `PullRequestEvent` event type.

In the event that a Team is requested for a review instead of a User, it will replace the `requested_reviewer` in the JSON payload with `requested_team`.

This allows for reusing similar logic employed when consuming `RequestedReviewer` without having to iterate through `PullRequest.RequestedTeams` which could potentially cause duplication events without additional logic.

Tested with Github Enterprise 2.15.9.


kremlinkev:
Since this was a pretty trivial change, apologies in advance if I missed something basic.

PR #1171, opened 2019-05-16
Add GetTeamBySlug methods in teamsService

adrienzieba:
Add a method to call the following endpoint
https://developer.github.com/v3/teams/#get-team-by-name

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -121,10 +121,10 @@ func (s *TeamsService) GetTeam(ctx context.Context, team int64) (*Team, *Respons
 	return t, resp, nil
 }
 
-// GetTeamByName fetches a team by slug.
+// GetTeamBySlug fetches a team by slug.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/#get-team-by-name
-func (s *TeamsService) GetTeamByName(ctx context.Context, org, slug string) (*Team, *Response, error) {
+func (s *TeamsService) GetTeamBySlug(ctx context.Context, org, slug string) (*Team, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/teams/%v", org, slug)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
--- a/github/teams.go
+++ b/github/teams.go
@@ -124,7 +124,7 @@ func (s *TeamsService) GetTeam(ctx context.Context, team int64) (*Team, *Respons
 // GetTeamByName fetches a team by slug.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/#get-team-by-name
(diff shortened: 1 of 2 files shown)


adrienzieba:
I signed it!

gmlewis:
Hi @willnorris, when you get a chance, can you please check the status of the corporate CLA mentioned above?  Also, just FYI... I will be OOO for one week starting tomorrow without internet access, so feel free to moderate as you see fit. @gauntface - just FYI for you too. Thanks!

adrienzieba:
CLA should be good now

github/teams.go
```diff
@@ -121,6 +121,25 @@ func (s *TeamsService) GetTeam(ctx context.Context, team int64) (*Team, *Respons
 	return t, resp, nil
 }
 
+// GetTeamByName fetches a team by slug.
+//
+// GitHub API docs: https://developer.github.com/v3/teams/#get-team-by-name
+func (s *TeamsService) GetTeamByName(ctx context.Context, org string, slug string) (*Team, *Response, error) {
```
gmlewis: s/org string, slug string/org, slug string/

github/teams.go
```diff
@@ -121,6 +121,25 @@ func (s *TeamsService) GetTeam(ctx context.Context, team int64) (*Team, *Respons
 	return t, resp, nil
 }
 
+// GetTeamByName fetches a team by slug.
```
gmlewis: Since the parameter is called `slug` and your documentation also says `slug`, and the GitHub docs say "the team's slug", what do you think about calling this method `GetTeamBySlug` ?
I'm actually fine either way, but I had no clue what a slug was until I looked at the linked GitHub docs.

adrienzieba: It was just to use the same name used as title inside the github documentation.
I'm fine changing the name of the method to `GetTeamBySlug` if needed.

gmlewis: Right... that makes total sense. Let's see what @gauntface thinks.

adrienzieba:
Any update ?

gmlewis:
@gauntface - do you have an opinion about my [comment above](https://github.com/google/go-github/pull/1171#discussion_r285317230)?

github/teams.go
gauntface: Sorry for the delay on this - GetTeamBySlug please.

The docs say: `/orgs/:org/teams/:team_slug` so I would definitely use slug, since it's different from the name.

adrienzieba: @gauntface, @gmlewis I renamed the method. Thank you for your review.

gmlewis:
No problem, @gauntface! We greatly appreciate all your awesome reviews!
Merging.

PR #1178, opened 2019-05-26
Return new http.Client when nil passed to NewClient

### LINKED ISSUE #1173: Unsafe concurrent modification of http.DefaultClient
When the `RepositoriesService.DownloadReleaseAsset` [method](https://github.com/google/go-github/blob/10ee8e064fb3db61aa03ebf9c58eb88e6dac5ec5/github/repos_releases.go#L270) is called, the `http.Client` pointed to by `github.Client` is modified in order to [temporarily override](https://github.com/google/go-github/blob/10ee8e064fb3db61aa03ebf9c58eb88e6dac5ec5/github/repos_releases.go#L283-L288) `http.Client.CheckRedirect` for the span of a single request. A `sync.Mutex` is used on the `github.Client` struct to prevent multiple concurrent modifications by the same instance of `github.Client` - however, if multiple instances refer to the same `http.Client` and each call this method, it becomes unsafe. 

The intent may be that each `github.Client` receives its own `http.Client` to mitigate this type of issue, but the fact that `github.NewClient` specifically defaults to using `http.DefaultClient` when one is not provided suggests otherwise. This not only affects multiple concurrent `github.Client` instances, it will also (potentially) impact any other code in the program making use of `http.DefaultClient` with no indication.

This behavior presented itself in a case where multiple clients were created to download release assets from different sources. The underlying `http.Client` passed to `NewClient` was `nil` in the case that authentication was not required, which led to the case above with `http.DefaultClient` being shared by multiple github clients. The ultimate effect was that redirect URLs returned by different method calls were mismatched (i.e. the request to download repo `A` release asset `1` instead returned a URL for downloading repo `B`'s release asset `2`), leading to data corruption.

Additionally, I did notice [this method](https://github.com/google/go-github/blob/c40b4abe46b109280c2784d9a9bfb37ffba4b8aa/github/migrations_user.go#L148) which performs a similar modification of the underlying `http.Client` _without_ locking the client's mutex.

gmlewis:
Fixes #1173.

Reviewers: can you think of any negative impacts this change might have on users of this package?

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -244,12 +244,12 @@ func addOptions(s string, opt interface{}) (string, error) {
 }
 
 // NewClient returns a new GitHub API client. If a nil httpClient is
-// provided, http.DefaultClient will be used. To use API methods which require
+// provided, a new http.Client will be used. To use API methods which require
 // authentication, provide an http.Client that will perform the authentication
 // for you (such as that provided by the golang.org/x/oauth2 library).
 func NewClient(httpClient *http.Client) *Client {
 	if httpClient == nil {
-		httpClient = http.DefaultClient
+		httpClient = &http.Client{}
 	}
 	baseURL, _ := url.Parse(defaultBaseURL)
 	uploadURL, _ := url.Parse(uploadBaseURL)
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @gauntface!
Merging.

PR #1179, opened 2019-05-31
github.App: Use *Timestamp instead of *time.Time

aidansteele:
This is a similar issue to #965 - the `created_at` and `updated_at` fields in the `app` field of webhooks are formatted as Unix epoch integers, rather than RFC3339 for the CRUD APIs.


gmlewis:
Thank you, @vaibhavsingh97!
Merging.

Note to self: this is a breaking API change so we will bump the major version after merging.

PR #1180, opened 2019-05-31
Add missing permissions to InstallationPermissions

tarebyte:
This adds the remaining GitHub App installation permissions defined in https://developer.github.com/v3/apps/permissions/#metadata-permissions that are related to GitHub App installations.

These do not include permissions such as `emails` or `gpg_keys` as these are user permissions.

I think I found the relevant test(s) to update, but if there are more I can add please let me know.

✨ Thanks ✨


gmlewis:
Thank you, @tarebyte!

I'm looking around the GitHub API v3 Developer docs, and am having troubles finding these added fields. Where did you find these?
Are they really all strings?
If you have a documentation URL, could you please add it to a comment before the expanded struct?

tarebyte:
> I'm looking around the GitHub API v3 Developer docs, and am having troubles finding these added fields. Where did you find these?

The permissions are listed here: https://developer.github.com/v3/apps/permissions

> Are they really all strings?

Yup! The string is either `read`/`write`/`admin` and if it's not one of those then the field isn't part of the JSON response.

An example of this would be the `permissions` key in this response https://developer.github.com/v3/apps/#get-an-installation

> If you have a documentation URL, could you please add it to a comment before the expanded struct?

Happy to, is https://developer.github.com/v3/apps/permissions/ sufficient?

gmlewis:
Hmmm... I'm still confused. Let's take an example.
Where did you find `organization_pre_receive_hooks` ?

tarebyte:
> Where did you find the string `organization_pre_receive_hooks` ?

That's an Enterprise only permission, it's located in https://developer.github.com/enterprise/2.17/v3/apps/permissions/

gmlewis:
Ah!  Great, thanks.  Then can you please add the two URLs in the comment?
Something like:
```go
// Hook names taken from:
//   https://developer.github.com/v3/apps/permissions/
//   https://developer.github.com/enterprise/2.17/v3/apps/permissions/
``
Thank you!

**Comment by tarebyte** [comment]
> Ah! Great, thanks. Then can you please add the two URLs in the comment?

Done in https://github.com/google/go-github/pull/1180/commits/a8d3df256cd539b723943af1740a67447c601179

File: github/apps.go [hunk 1]
```diff
@@ -36,12 +36,35 @@ type InstallationToken struct {
 	ExpiresAt *time.Time `json:"expires_at,omitempty"`
 }
 
-// InstallationPermissions lists the permissions for metadata, contents, issues and single file for an installation.
+// InstallationPermissions lists the repository and organization permissions for an installation.
+//
+// Permission names taken from:
+//   https://developer.github.com/v3/apps/permissions/
+//   https://developer.github.com/enterprise/v3/apps/permissions/
```
tarebyte: This will redirect to the latest Enterprise version.

https://developer.github.com/enterprise/v3/apps/permissions/ -> https://developer.github.com/enterprise/2.17/v3/apps/permissions/

github/apps.go
```diff
@@ -36,12 +36,35 @@ type InstallationToken struct {
...
-	Contents   *string `json:"contents,omitempty"`
-	Issues     *string `json:"issues,omitempty"`
-	SingleFile *string `json:"single_file,omitempty"`
+	Administration              *string `json:"administration,omitempty"`
+	Checks                      *string `json:"checks,omitempty"`
+	Contents                    *string `json:"contents,omitempty"`
+	ContentReferences           *string `json:"content_references,omitempty"`
+	Deployments                 *string `json:"deployments,omitempty"`
+	Issues                      *string `json:"issues,omitempty"`
+	Metadata                    *string `json:"metadata,omitempty"`
```
itsdalmo: According to the following documentation:
https://developer.github.com/apps/building-github-apps/creating-github-apps-using-url-parameters/#github-app-permissions

It seems like this should be `RepositoryMetadata`, and more importantly, the JSON key should be `repository_metadata`?

tarebyte: @itsdalmo that seems to be an error in the docs, the key from the response is definitely `metadata`.

gmlewis: Thanks, @itsdalmo and @tarebyte!

Would one of you please volunteer to contact support@github.com and let them know about the documentation error?  They are typically very responsive, friendly, and helpful, and this would be appreciated feedback.   Thank you!

tarebyte: > Would one of you please volunteer to contact support@github.com and let them know about the documentation error?

I can file an internal issue, I work for GitHub 😃 

> They are typically very responsive, friendly, and helpful, and this would be appreciated feedback. Thank you!

💖

gmlewis: Oh, uh, Wow!  I'm humbled and honored that you are helping us out here, @tarebyte!!!

You and your team have been an absolute joy to work with!  I have a bunch of virtual friends at GitHub Tech Support.  Thanks for all you do!

So let's see... all I need is another LGTM and we are good to merge.

tarebyte: > Oh, uh, Wow! I'm humbled and honored that you are helping us out here, @tarebyte!!!

> You and your team have been an absolute joy to work with! I have a bunch of virtual friends at GitHub Tech Support. Thanks for all you do!

It's my pleasure! Thank you for working on such a ✨ library

gmlewis:
@itsdalmo - since you reviewed the code, can I please get an LGTM from you and we can merge this?

tarebyte:
@itsdalmo, those are all user permissions that aren’t granted to an app installation.

These are only requested and granted when a user goes through the authorization flow and the GitHub App gets a “user to server” token.

tarebyte:
Ooo actually I think `packages` might be a installation permission.

I’ll double check.

gmlewis:
It looks like we have two LGTMs but also some more commits coming in.

Just FYI, @tarebyte, we try not to use force pushes for PRs in this repo as it makes it more challenging for the reviewers to see what changed, but no big deal. Sometimes we have to use them to clear up the CLA checker and make it happy. :smile:

So I'm just checking in with you on the status... are we good to merge, or to you think there are other changes that need making first?

tarebyte:
Good to know thanks!

This is good to go from my perspective, the missing packages permission was all I needed to add.

gmlewis:
Thank you, @tarebyte and @itsdalmo!
Merging.

PR #1182, opened 2019-06-05
Add support for creating scoped installation tokens

tarebyte:
See https://developer.github.com/v3/apps/#create-a-new-installation-token

This creates a new function `CreateScopedInstallationToken`, which allows the caller to limit an installation's access to a subset of repositories and permissions.

These parameters are optional which is why I opted to created a new function for this.

✨  Feedback appreciated ✨

### DIFF
--- a/github/apps.go
+++ b/github/apps.go
@@ -32,8 +32,17 @@ type App struct {
 
 // InstallationToken represents an installation token.
 type InstallationToken struct {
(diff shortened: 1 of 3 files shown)


gmlewis:
friendly ping @gauntface

gauntface:
I would agree with @gmlewis that moving to a single struct would make this more consistent with the rest of the API. Otherwise, this change is LGTM from me

PR #1196, opened 2019-06-14
Add gen-stringify-test.go

gmlewis:
After looking at the coverage details here: https://codecov.io/gh/google/go-github/tree/master/github
and noticing that the `String` methods account for a non-trivial percentage of uncovered code, I decided to write some.

Then, after writing a couple tests by hand, I realized that this is another great opportunity for `go generate` to write this mundane code for us and put it in a test file that nobody needs to care or worry about (`github-stringify_test.go`).

Getting these trivial functions out of the way, I believe that the coverage details will now have more meaning and we can more clearly identify and focus our attention on methods that need better test coverage.


gmlewis:
friendly ping @gauntface

github/gen-stringify-test.go
```diff
@@ -0,0 +1,358 @@
...
+const (
+	ignoreFilePrefix1 = "gen-"
+	ignoreFilePrefix2 = "github-"
+	outputFileSuffix  = "-stringify_test.go"
+)
+
+var (
+	verbose = flag.Bool("v", false, "Print verbose log messages")
+
+	sourceTmpl = template.Must(template.New("source").Funcs(funcMap).Parse(source))
```
gauntface: Nit: could you move this to *after* the definition of funcMap.

gmlewis:
Thank you, @gauntface!
Merging.

PR #1198, opened 2019-06-14
add SignKey gpg entity param to allow easier pgp signing of commits

anandkumarpatel:
Reopening of https://github.com/google/go-github/pull/1142

This PR creates a new param for the `Commit` struct called `SignKey`.
`SignKey` is an `openpgp.Entity` type which is the standard way of reading and validating pgp keys.

When specified the key will be used to sign the commit. This makes it so the user does not have to sign the key externally and duplicate some of the logic the `CreateCommit` function does.

To maintain backward compatibility, if the `Verification.Signature` is defined `SignKey` will have no effect. 

This provides similar functionality that go-git has: https://github.com/src-d/go-git/commit/7b6c1266556f59ac436fada3fa6106d4a84f9b56

Note from the git docs: https://github.com/octokit/routes/blob/152c0b5d156844e17a1805d0509c9a0b7cfc848f/routes/api.github.com/git.json#L282
```
To pass a `signature` parameter, you need to first manually create a valid PGP signature, which can be complicated. You may find it easier to [use the command line]
```
So I wanted to make it less complicated 😄

### DIFF
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -150,8 +150,8 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
 	return c, resp, nil
 }
 
-func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, error) {
-	if SigningKey == nil || commit == nil {
+func createSignature(signingKey *openpgp.Entity, commit *createCommit) (string, error) {
+	if signingKey == nil || commit == nil {
 		return "", errors.New("createSignature: invalid parameters")
 	}
 
@@ -162,7 +162,7 @@ func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, 
 
 	writer := new(bytes.Buffer)
 	reader := bytes.NewReader([]byte(message))
-	if err := openpgp.ArmoredDetachSign(writer, SigningKey, reader, nil); err != nil {
+	if err := openpgp.ArmoredDetachSign(writer, signingKey, reader, nil); err != nil {
 		return "", err
 	}
 
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -157,7 +157,7 @@ func createSignature(signingKey *openpgp.Entity, commit *createCommit) (string, 
 
 	message, err := createSignatureMessage(commit)
 	if err != nil {
-		return "", fmt.Errorf("createSignature: createSignatureMessage error: %s", err.Error())
+		return "", err
 	}
 
 	writer := new(bytes.Buffer)
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -151,15 +151,14 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
 }
 
 func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, error) {
-	if commit.Author == nil {
-		return "", errors.New("createSignature: commit.Author=nil")
+	if SigningKey == nil || commit == nil {
+		return "", errors.New("createSignature: invalid parameters")
 	}
 
-	if commit.Message == nil {
-		return "", errors.New("createSignature: commit.Message=nil")
+	message, err := createSignatureMessage(commit)
+	if err != nil {
+		return "", fmt.Errorf("createSignature: createSignatureMessage error: %s", err.Error())
 	}
-
-	message := createSignatureMessage(commit)
 
 	writer := new(bytes.Buffer)
 	reader := bytes.NewReader([]byte(message))
@@ -170,7 +169,11 @@ func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, 
 	return writer.String(), nil
 }
 
-func createSignatureMessage(commit *createCommit) string {
+func createSignatureMessage(commit *createCommit) (string, error) {
+	if commit == nil || commit.Message == nil || commit.Author == nil {
+		return "", errors.New("createSignatureMessage: invalid parameters")
+	}
+
 	message := ""
 
 	if commit.Tree != nil {
@@ -190,5 +193,5 @@ func createSignatureMessage(commit *createCommit) string {
 	// There needs to be a double newline after committer
 	message += fmt.Sprintf("committer %s <%s> %d %s\n\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700"))
 	message += fmt.Sprintf("%s", *commit.Message)
-	return message
+	return message, nil
 }
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"golang.org/x/crypto/openpgp"
@@ -174,24 +175,26 @@ func createSignatureMessage(commit *createCommit) (string, error) {
 		return "", errors.New("createSignatureMessage: invalid parameters")
 	}
 
-	message := ""
+	var message []string
 
 	if commit.Tree != nil {
-		message = fmt.Sprintf("tree %s\n", *commit.Tree)
+		message = append(message, fmt.Sprintf("tree %s", *commit.Tree))
 	}
 
 	for _, parent := range commit.Parents {
-		message += fmt.Sprintf("parent %s\n", parent)
+		message = append(message, fmt.Sprintf("parent %s", parent))
 	}
 
-	message += fmt.Sprintf("author %s <%s> %d %s\n", commit.Author.GetName(), commit.Author.GetEmail(), commit.Author.GetDate().Unix(), commit.Author.GetDate().Format("-0700"))
+	message = append(message, fmt.Sprintf("author %s <%s> %d %s", commit.Author.GetName(), commit.Author.GetEmail(), commit.Author.GetDate().Unix(), commit.Author.GetDate().Format("-0700")))
+
 	committer := commit.Committer
 	if committer == nil {
 		committer = commit.Author
 	}
 
 	// There needs to be a double newline after committer
-	message += fmt.Sprintf("committer %s <%s> %d %s\n\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700"))
-	message += fmt.Sprintf("%s", *commit.Message)
-	return message, nil
+	message = append(message, fmt.Sprintf("committer %s <%s> %d %s\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700")))
+	message = append(message, fmt.Sprintf("%s", *commit.Message))
+
+	return strings.Join(message, "\n"), nil
 }
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -171,7 +171,7 @@ func createSignature(signingKey *openpgp.Entity, commit *createCommit) (string, 
 }
 
 func createSignatureMessage(commit *createCommit) (string, error) {
-	if commit == nil || commit.Message == nil || commit.Author == nil {
+	if commit == nil || commit.Message == nil || *commit.Message == "" || commit.Author == nil {
 		return "", errors.New("createSignatureMessage: invalid parameters")
 	}
 
@@ -194,7 +194,7 @@ func createSignatureMessage(commit *createCommit) (string, error) {
 
 	// There needs to be a double newline after committer
 	message = append(message, fmt.Sprintf("committer %s <%s> %d %s\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700")))
-	message = append(message, fmt.Sprintf("%s", *commit.Message))
+	message = append(message, *commit.Message)
 
 	return strings.Join(message, "\n"), nil
 }
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -154,6 +154,11 @@ func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, 
 	if commit.Author == nil {
 		return "", errors.New("createSignature: commit.Author=nil")
 	}
+
+	if commit.Message == nil {
+		return "", errors.New("createSignature: commit.Message=nil")
+	}
+
 	message := createSignatureMessage(commit)
 
 	writer := new(bytes.Buffer)
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -8,6 +8,7 @@ package github
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


github/git_commits_test.go
```diff
@@ -123,6 +127,172 @@ func TestGitService_CreateSignedCommit(t *testing.T) {
...
+	keyring, err := openpgp.ReadArmoredKeyRing(s)
+
+	date, _ := time.Parse("Mon Jan 02 15:04:05 2006 -0700", "Thu May 04 00:03:43 2017 +0200")
+	author := CommitAuthor{
+		Name:  String("go-github"),
+		Email: String("go-github@github.com"),
+		Date:  &date,
+	}
+	input := &Commit{
+		Message:    String("m"),
```
gmlewis: Can you please change `"m"` to `"Commit message."`, as it makes line 169 much easier to read below. Thanks.

github/git_commits_test.go
```diff
@@ -123,6 +127,172 @@ func TestGitService_CreateSignedCommit(t *testing.T) {
...
+		Tree:    String("t"),
+		Parents: []string{"p"},
+	}
+
+	_, err := createSignature(nil, a)
+
+	if err == nil {
+		t.Errorf("Expected error to be returned because no author was passed")
+	}
+	expectedString := "Commit Author is required to sign a commit"
```
gmlewis: Let's please not test against the error message value.... you can remove this section completely.
That way we can easily change it. :smile:

github/git_commits_test.go
```diff
@@ -123,6 +127,172 @@ func TestGitService_CreateSignedCommit(t *testing.T) {
...
+	if err == nil {
+		t.Errorf("Expected error to be returned due to invalid key")
+	}
+}
+
+func TestGitService_createSignatureMessage_withoutTree(t *testing.T) {
+	date, _ := time.Parse("Mon Jan 02 15:04:05 2006 -0700", "Thu May 04 00:03:43 2017 +0200")
+
+	msg := createSignatureMessage(&createCommit{
+		Message: String("m"),
```
gmlewis: Same here, please: `"Commit message."`.

github/git_commits_test.go
```diff
@@ -123,6 +127,172 @@ func TestGitService_CreateSignedCommit(t *testing.T) {
...
+	if msg != expected {
+		t.Errorf("Returned message incorrect. returned %s, want %s", msg, expected)
+	}
+}
+
+func TestGitService_createSignatureMessage_withoutCommitter(t *testing.T) {
+	date, _ := time.Parse("Mon Jan 02 15:04:05 2006 -0700", "Thu May 04 00:03:43 2017 +0200")
+
+	msg := createSignatureMessage(&createCommit{
+		Message: String("m"),
```
gmlewis: And here.

github/git_commits_test.go
```diff
@@ -131,3 +301,62 @@ func TestGitService_CreateCommit_invalidOwner(t *testing.T) {
 	_, _, err := client.Git.CreateCommit(context.Background(), "%", "%", &Commit{})
 	testURLParseError(t, err)
 }
+
+const testKey = `
```
gmlewis: `testGPGKey` please.

github/git_commits.go
```diff
@@ -133,3 +148,41 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+	}
+
+	return writer.String(), nil
+}
+
+func createSignatureMessage(commit *createCommit) string {
+	message := ""
+
+	if commit.Tree != nil {
+		message = fmt.Sprintf("tree %s\n", *commit.Tree)
```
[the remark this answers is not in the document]
anandkumarpatel: There are no getters defined for `createCommit` object

github/git_commits.go
```diff
@@ -133,3 +148,41 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
 
 	return c, resp, nil
 }
+
+func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, error) {
+	if commit.Author == nil {
+		return "", fmt.Errorf("Commit Author is required to sign a commit")
```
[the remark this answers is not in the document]
anandkumarpatel: Strange.. go vet didn't find a problem for me. Changing the error message anyway, I like the searchability aspect.

github/git_commits.go
```diff
@@ -133,3 +148,41 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+
+	if commit.Tree != nil {
+		message = fmt.Sprintf("tree %s\n", *commit.Tree)
+	}
+
+	for _, parent := range commit.Parents {
+		message += fmt.Sprintf("parent %s\n", parent)
+	}
+
+	message += fmt.Sprintf("author %s <%s> %d %s\n", *commit.Author.Name, *commit.Author.Email, commit.Author.Date.Unix(), commit.Author.Date.Format("-0700"))
```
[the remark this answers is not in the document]
anandkumarpatel: Same no getters for `createCommit` object. Getters are only defined for the normal `Commit` object not the one created in this file for some reason.

github/git_commits.go
```diff
@@ -133,3 +148,41 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+	commiter := commit.Committer
```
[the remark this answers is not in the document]
anandkumarpatel: nice catch!

anandkumarpatel:
@gmlewis fixed most of the comments you left. The only issue was with using getters on the `createCommit` object. they do not exist so I left those as is but used getters on the sub-objects that had them. 

I also updated all the short names in the test file and defined them more verbosely.

github/git_commits.go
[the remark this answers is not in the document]
gmlewis: Yes, whups, my bad.  Thanks, @anandkumarpatel.

github/git_commits.go
```diff
@@ -176,13 +176,14 @@ func createSignatureMessage(commit *createCommit) string {
...
 
-	message += fmt.Sprintf("author %s <%s> %d %s\n", *commit.Author.Name, *commit.Author.Email, commit.Author.Date.Unix(), commit.Author.Date.Format("-0700"))
-	commiter := commit.Committer
-	if commiter == nil {
-		commiter = commit.Author
+	message += fmt.Sprintf("author %s <%s> %d %s\n", commit.Author.GetName(), commit.Author.GetEmail(), commit.Author.GetDate().Unix(), commit.Author.GetDate().Format("-0700"))
+	committer := commit.Committer
+	if committer == nil {
+		committer = commit.Author
 	}
```
gmlewis: Ad this point, I still think we need a test if `committer==nil` since both `commit.Committer` and `commit.Author` could be `nil`.

github/git_commits.go
```diff
@@ -176,13 +176,14 @@ func createSignatureMessage(commit *createCommit) string {
...
+
 	// There needs to be a double newline after committer
-	message += fmt.Sprintf("committer %s <%s> %d %s\n\n", *commiter.Name, *commiter.Email, commiter.Date.Unix(), commiter.Date.Format("-0700"))
+	message += fmt.Sprintf("committer %s <%s> %d %s\n\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700"))
 	message += fmt.Sprintf("%s", *commit.Message)
```
gmlewis: We also need a test if `commit.Message` is `nil` if there is no getter for it.

gmlewis:
> I also updated all the short names in the test file and defined them more verbosely.

Just FYI... the other names were fine.  As you'll see in a lot of other tests, we use `o` for `org` and `r` for `repo` all over the place, and it would be nice to be consistent for these...

But is was just a bit challenging at a quick glance to figure out in the tests that:

```
...

m
```

was the commit message.  Thanks.

github/git_commits.go
anandkumarpatel: I check if `commit.Author` is nil at the top of the function since this function cannot work if that does not exist. Nothing modifies commit object in this function so not sure if we need to test for `nil` again?

https://github.com/google/go-github/pull/1198/files#diff-7da0b51e99c908bfec3fe2f646d4a0fdR154

gmlewis: Ah, I forgot about that.  Great, thanks!  You are correct... that is sufficient.  I'm trying to avoid crashes and that settles it. I'm going to add a tweak above, though.

github/git_commits.go
```diff
@@ -133,3 +149,41 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
 
 	return c, resp, nil
 }
+
+func createSignature(SigningKey *openpgp.Entity, commit *createCommit) (string, error) {
+	if commit.Author == nil {
```
gmlewis: Let's please make this:

```go
func createSignature(signingKey *openpgp.Entity, commit *createCommit) (string, error) {
  if signingKey == nil || commit == nil || commit.Author == nil {
    return "", errors.New("createSignature: invalid parameters")
  }
```

gmlewis: Note the lower-case `signingKey`... no need to capitalize that parameter.

github/git_commits.go
anandkumarpatel: added check to the caller function

anandkumarpatel:
@gmlewis ready for round 2 👊.

* added nil checks to both functions
* added respective tests for those nil checks
* renamed variable to start lower case

github/git_commits.go
```diff
@@ -133,3 +149,49 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
 }
+
+func createSignature(signingKey *openpgp.Entity, commit *createCommit) (string, error) {
+	if signingKey == nil || commit == nil {
+		return "", errors.New("createSignature: invalid parameters")
+	}
+
+	message, err := createSignatureMessage(commit)
+	if err != nil {
+		return "", fmt.Errorf("createSignature: createSignatureMessage error: %s", err.Error())
```
gmlewis: Let's please change this line to `return "", err` since the child error identifies itself.

github/git_commits.go
```diff
@@ -133,3 +149,49 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+
+	return writer.String(), nil
+}
+
+func createSignatureMessage(commit *createCommit) (string, error) {
+	if commit == nil || commit.Message == nil || commit.Author == nil {
+		return "", errors.New("createSignatureMessage: invalid parameters")
+	}
+
+	message := ""
```
gmlewis: I'm just going to throw this out for your consideration, and I'm fine either way.
When I'm building up potentially long strings that are a collection of lines, I often like to use a slice of strings and then join them at the end.  I think it is much easier this way to write and also to maintain... because eventually someone will come along and add another line but forget to add their `\n`... so it looks something like this:

```go
var lines []string
...
lines = append(lines, fmt.Sprintf("...", ...))  // no \n
...
return strings.Join(lines, "\n"), err
```

Totally your call.  I prefer it, but this way is fine too.

anandkumarpatel: I like this, does make it easier to modify!

anandkumarpatel:
@gmlewis Just pushed, lmk if there is something else.

github/git_commits.go
```diff
@@ -133,3 +150,51 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+
+	message = append(message, fmt.Sprintf("author %s <%s> %d %s", commit.Author.GetName(), commit.Author.GetEmail(), commit.Author.GetDate().Unix(), commit.Author.GetDate().Format("-0700")))
+
+	committer := commit.Committer
+	if committer == nil {
+		committer = commit.Author
+	}
+
+	// There needs to be a double newline after committer
+	message = append(message, fmt.Sprintf("committer %s <%s> %d %s\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700")))
```
gmlewis: On this line you can remove the `\n`.

github/git_commits.go
```diff
@@ -133,3 +150,51 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+	reader := bytes.NewReader([]byte(message))
+	if err := openpgp.ArmoredDetachSign(writer, signingKey, reader, nil); err != nil {
+		return "", err
+	}
+
+	return writer.String(), nil
+}
+
+func createSignatureMessage(commit *createCommit) (string, error) {
+	if commit == nil || commit.Message == nil || commit.Author == nil {
```
gmlewis: It just dawned on me that `commit.Message == nil` should be `commit.GetMessage() == ""` because we also care that there is no message, and this helper function does the extra work for us.

I apologize for not catching this one before.

github/git_commits.go
```diff
@@ -133,3 +150,51 @@ func (s *GitService) CreateCommit(ctx context.Context, owner string, repo string
...
+	message = append(message, fmt.Sprintf("author %s <%s> %d %s", commit.Author.GetName(), commit.Author.GetEmail(), commit.Author.GetDate().Unix(), commit.Author.GetDate().Format("-0700")))
+
+	committer := commit.Committer
+	if committer == nil {
+		committer = commit.Author
+	}
+
+	// There needs to be a double newline after committer
+	message = append(message, fmt.Sprintf("committer %s <%s> %d %s\n", committer.GetName(), committer.GetEmail(), committer.GetDate().Unix(), committer.GetDate().Format("-0700")))
+	message = append(message, fmt.Sprintf("%s", *commit.Message))
```
gmlewis: `message = append(message, commit.GetMessage())`

github/git_commits.go
anandkumarpatel: Yea I was debating what to do with this one. We need 2 new lines after the committer for this the be valid. It was between this or doing an empty string 
```
	message = append(message, fmt.Sprintf(""))
```
Though not sure if that is better or not.

github/git_commits.go
anandkumarpatel: Remember there are no getters for type `createCommit`

But will check for `nil` and empty string.

github/git_commits.go
anandkumarpatel: Again no getter, so will just use Message directly

anandkumarpatel:
@gmlewis Fixed the issues.
* checked for empty message
* added respective test
* simplified adding message (removed sprintf)

github/git_commits.go
gmlewis: Oh, sorry... yes, if you want a blank line definitely keep the `\n` here.  No, never use `fmt.Sprintf("")`... that would just be a `""` string literal... but in this case, please just keep the `\n`.

gmlewis:
friendly ping @gauntface

anandkumarpatel:
@gauntface I would assume if `commit.Verification` is set then the user wants to use that to sign instead of the `commit.SigningKey`. 
My assumption is that a default `SigningKey` is always passed and that `Verification` would only be passed in special cases which should override `SigningKey`. 

Are my assumptions invalid?

gmlewis:
Good discussion. Thanks, @gauntface and @anandkumarpatel !
I like the fact that the comment for `SigningKey` says `Ignored if Verification.Signature is defined.` so there shouldn't be much confusion in this regard.
Merging.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/google/go-github/v26
 require (
 	github.com/golang/protobuf v1.2.0 // indirect
 	github.com/google/go-querystring v1.0.0
+	github.com/jchavannes/go-pgp v0.0.0-20170517064339-cb4b0b0ac8cf
```
gmlewis: Why is this line needed if we are only importing `golang.org/x/crypto/openpgp`?
Can you please revert the changes to `go.mod` and `go.sum` if these changes aren't needed?

anandkumarpatel: Done! They were there with my first implementation (had nice sugar functions) but I decided to stick with `golang.org/x/crypto/openpgp` and stay away from the sugar.

gmlewis:
Thank you, @anandkumarpatel!
Merging.

PR #1207, opened 2019-06-24
Consistent errors for GetRef when a Ref doesn't exist

pilbot:


gmlewis:
This is fine with me. What do you think, @gauntface ?

If we do this, I believe it would fall under the category of a "minor" semver change... meaning that we would bump the middle number. It doesn't break any existing clients because it still returns an error, but the error response does change.

If @gauntface is fine with this change, then LGTM.

github/git_refs.go
```diff
@@ -69,6 +69,9 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
 	if _, ok := err.(*json.UnmarshalTypeError); ok {
 		// Multiple refs, means there wasn't an exact match.
 		return nil, resp, errors.New("no exact match found for this ref")
```
gauntface: Could we change the error message to "multiple matches found for this ref"?

github/git_refs.go
```diff
@@ -69,6 +69,9 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
...
+	} else if resp.StatusCode == 404 {
+		// No ref, there was no match for the ref
+		return nil, resp, errors.New("no exact match found for this ref")
```
gauntface: Could we change this to "no match found for this ref"?

github/git_refs.go
pilbot: DOne

github/git_refs.go
pilbot: Done

gmlewis:
Thank you, @gauntface and @pilbot !
Merging with a bump to the minor version since the error response changes.

PR #1215, opened 2019-07-04
Add Visibility to UserEmail

izumin5210:
Because `UserEmail.Visibility` field is missed.
https://developer.github.com/v3/users/emails/#list-email-addresses-for-a-user


izumin5210:
oh sorry... 🙈 i fixed it!

PR #1233, opened 2019-07-19
Use time.Until

muesli:
Instead of manually subtracting the timestamps, we can just use time.Until.


gmlewis:
Fine by me.  Thanks, @muesli.

gmlewis:
Thank you, @gauntface.

Merging.

Issue #1235, opened 2019-07-22
Support preview repository creation template API


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2019-07-16-repository-templates-api/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

waseem18:
I'd like to take this up @gmlewis

shanempope:
@waseem18 I started this as a learning exercise, feel free to use my changes or redo it. I did not want to claim it in case it was over my head. I've linked the PR above. But I appear to need more code coverage. I will look into it tomorrow if you haven't finished your own. Take care!

gmlewis:
Thank you, @waseem18, for your offer, and thanks also for your understanding about @shanempope putting together a PR.

Thanks, @shanempope, for your PR... I reviewed it and made some comments.  In the future, please place a note on the issue that you are working on it so others don't waste their time working on it too.  Thanks!

PR #1236, opened 2019-07-23
Add support for creating repos from template repos

shanempope:
Add API for `POST /repos/:template_owner/:template_repo/generate` endpoint which creates a repository from a template repository.
Developer Blog: https://developer.github.com/changes/2019-07-16-repository-templates-api/
API Reference: https://developer.github.com/v3/repos/#create-repository-using-a-repository-template

Issue: https://github.com/google/go-github/issues/1235


shanempope:
I signed it!

shanempope:
> Note: Creating and using repository templates is currently available for developers to preview. To access this new endpoint during the preview period, you must provide a custom media type in the Accept header:
> 
> `application/vnd.github.baptiste-preview+json`

And

> Warning: The API may change without advance notice during the preview period. Preview features are not supported for production use. If you experience any issues, contact GitHub Support.

github/repos.go
```diff
@@ -334,6 +338,36 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
 	return r, resp, nil
 }
 
+// TemplateRepoRequest represents a request to create a repo from a template
```
gmlewis: For the auto-generated godocs, please end the comment in a period.

github/repos.go
```diff
@@ -334,6 +338,36 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
...
+type TemplateRepoRequest struct {
+	// Name is required when creating a repo.
+	Name        *string `json:"name,omitempty"`
+	Owner       *string `json:"owner,omitempty"`
+	Description *string `json:"description,omitempty"`
+
+	Private *bool `json:"private,omitempty"`
+}
+
+// Generate a repository using a repository template
```
gmlewis: Hmmm... I see that GitHub named this endpoint `generate`, which seems odd to me.

@gauntface - what do you think about calling this endpoint `repositories.CreateFromTemplate` (or maybe just `FromTemplate`?) as that seems to more appropriate to me?

Either way, @shanempope, we will want to make this comment a complete sentence for the auto-generated godocs.  Hopefully something like:

`// CreateFromTemplate generates a repository from a template.`

github/repos.go
```diff
@@ -334,6 +338,36 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
...
+//
+// GitHub API docs: https://developer.github.com/v3/repos/#create-repository-using-a-repository-template
+func (s *RepositoriesService) Generate(ctx context.Context, templateOwner string, templateRepo string, templateRepoReq *TemplateRepoRequest) (*Repository, *Response, error) {
```
gmlewis: `templateOwner string, templateRepo string,` should be `templateOwner, templateRepo string,`

github/repos.go
```diff
@@ -334,6 +338,36 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
...
+	u := fmt.Sprintf("repos/%v/%v/generate", templateOwner, templateRepo)
+
+	req, err := s.client.NewRequest("POST", u, templateRepoReq)
+	if err != nil {
+		return nil, nil, err
+	}
+
```
gmlewis: Please add the required preview header here.

github/repos_test.go
```diff
@@ -333,6 +333,66 @@ func TestRepositoriesService_Create_org(t *testing.T) {
...
+
+	templateRepoReq := &TemplateRepoRequest{
+		Name: String("n"),
+	}
+
+	mux.HandleFunc("/repos/to/tr/generate", func(w http.ResponseWriter, r *http.Request) {
+		v := new(TemplateRepoRequest)
+		json.NewDecoder(r.Body).Decode(v)
+
+		testMethod(t, r, "POST")
```
gmlewis: Please test for the required preview header here.

github/repos.go
shanempope: Thank you!
The docs show the header required throughout the endpoints, in order to get the fields. (Create for example)  https://developer.github.com/v3/repos/#create
I will add add it for all the required endpoints.

shanempope:
I will add the custom headers to the other two endpoints later today.

shanempope:
Alright, I manually tested all of the endpoints. It's ready for a second review. Thanks @gmlewis

shanempope:
Thanks @gauntface and @gmlewis. I've squashed the commits

gmlewis:
Thank you, @shanempope, but we prefer that you don't use force-push in this repo, as it makes it more difficult for other contributors to see the change history in case they wish to make comments during the PR review period.  Additionally, we always "Squash and merge" in this repo, so it ends up being a single commit anyway, so the force-push is unnecessary. 

Thank you, @gauntface.

Merging.

PR #1238, opened 2019-07-23
Update apps.go

emoryruscus:
Add installation token parameters, allowing creation of permission-scoped github tokens.

https://developer.github.com/v3/apps/#create-a-new-installation-token

### DIFF
--- a/github/apps.go
+++ b/github/apps.go
@@ -32,8 +32,21 @@ type App struct {
 
 // InstallationToken represents an installation token.
 type InstallationToken struct {
-	Token     *string    `json:"token,omitempty"`
-	ExpiresAt *time.Time `json:"expires_at,omitempty"`
+	Token        *string                  `json:"token,omitempty"`
+	ExpiresAt    *time.Time               `json:"expires_at,omitempty"`
+	Permissions  *InstallationPermissions `json:"permissions,omitempty"`
+	Repositories []*Repository            `json:"repositories,emitempty"`
+}
+
+// InstallationTokenOptions allow restricting a token's access to specific repositories.
+type InstallationTokenOptions struct {
+	// The IDs of the repositories that the installation token can access.
+	// Providing repository IDs restricts the access of an installation token to specific repositories.
+	RepositoryIDs []int64 `json:"repository_ids,omitempty"`
+
+	// The permissions granted to the access token.
+	// The permissions object includes the permission names and their access type.
+	Permissions *InstallationPermissions `json:"permissions,omitempty"`
 }
 
 // InstallationPermissions lists the repository and organization permissions for an installation.
@@ -194,10 +207,10 @@ func (s *AppsService) ListUserInstallations(ctx context.Context, opt *ListOption
 // CreateInstallationToken creates a new installation token.
 //
 // GitHub API docs: https://developer.github.com/v3/apps/#create-a-new-installation-token
-func (s *AppsService) CreateInstallationToken(ctx context.Context, id int64) (*InstallationToken, *Response, error) {
+func (s *AppsService) CreateInstallationToken(ctx context.Context, id int64, opt *InstallationTokenOptions) (*InstallationToken, *Response, error) {
 	u := fmt.Sprintf("app/installations/%v/access_tokens", id)
 
-	req, err := s.client.NewRequest("POST", u, nil)
+	req, err := s.client.NewRequest("POST", u, opt)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/apps.go
+++ b/github/apps.go
@@ -35,7 +35,7 @@ type InstallationToken struct {
 	Token        *string                  `json:"token,omitempty"`
 	ExpiresAt    *time.Time               `json:"expires_at,omitempty"`
 	Permissions  *InstallationPermissions `json:"permissions,omitempty"`
-	Repositories []*Repository            `json:"repositories,emitempty"`
+	Repositories []*Repository            `json:"repositories,omitempty"`
 }
 
 // InstallationTokenOptions allow restricting a token's access to specific repositories.
--- a/github/apps_test.go
+++ b/github/apps_test.go
@@ -6,8 +6,12 @@
 package github
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"reflect"
 	"testing"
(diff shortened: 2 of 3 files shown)


github/apps.go
```diff
@@ -32,8 +36,21 @@ type App struct {
...
+	Token        *string                  `json:"token,omitempty"`
+	ExpiresAt    *time.Time               `json:"expires_at,omitempty"`
+	Permissions  *InstallationPermissions `json:"permissions,omitempty"`
+	Repositories *[]Repository            `json:"repositories,emitempty"`
+}
+
+// InstallationTokenParameters allow restricting a token's access to specific repositories.
+type InstallationTokenParameters struct {
+	// The ids of the repositories that the installation token can access.
+	// Providing repository ids restricts the access of an installation token to specific repositories.
```
gmlewis: Let's please change `ids` in the comment to `IDs` for the auto-generated godocs.

github/apps.go
```diff
@@ -194,10 +211,10 @@ func (s *AppsService) ListUserInstallations(ctx context.Context, opt *ListOption
 // CreateInstallationToken creates a new installation token.
 //
 // GitHub API docs: https://developer.github.com/v3/apps/#create-a-new-installation-token
-func (s *AppsService) CreateInstallationToken(ctx context.Context, id int64) (*InstallationToken, *Response, error) {
+func (s *AppsService) CreateInstallationToken(ctx context.Context, id int64, body ...interface{}) (*InstallationToken, *Response, error) {
```
gmlewis: Please let's not use `...interface{}` but keep this repo strictly typed.
There are many other examples where we pass in a pointer to an optional struct and we usually call it `opt`, as in:
`opt *InstallationTokenOptions`

github/apps.go
```diff
@@ -32,8 +36,21 @@ type App struct {
...
-	Token     *string    `json:"token,omitempty"`
-	ExpiresAt *time.Time `json:"expires_at,omitempty"`
+	Token        *string                  `json:"token,omitempty"`
+	ExpiresAt    *time.Time               `json:"expires_at,omitempty"`
+	Permissions  *InstallationPermissions `json:"permissions,omitempty"`
+	Repositories *[]Repository            `json:"repositories,emitempty"`
+}
+
+// InstallationTokenParameters allow restricting a token's access to specific repositories.
+type InstallationTokenParameters struct {
```
gmlewis: Let's please rename this to `InstallationTokenOptions` to be more consistent with the rest of the repo.

github/apps.go
```diff
@@ -32,8 +36,21 @@ type App struct {
 
 // InstallationToken represents an installation token.
 type InstallationToken struct {
-	Token     *string    `json:"token,omitempty"`
-	ExpiresAt *time.Time `json:"expires_at,omitempty"`
+	Token        *string                  `json:"token,omitempty"`
+	ExpiresAt    *time.Time               `json:"expires_at,omitempty"`
+	Permissions  *InstallationPermissions `json:"permissions,omitempty"`
+	Repositories *[]Repository            `json:"repositories,emitempty"`
```
gmlewis: Please change `*[]Repository` to `[]*Repository` instead.

github/apps.go
```diff
@@ -282,3 +299,32 @@ func (s *AppsService) getInstallation(ctx context.Context, url string) (*Install
 
 	return i, resp, nil
 }
+
+// GetReadWriter converts a body interface into an io.ReadWriter object.
+func GetReadWriter(body interface{}) (io.ReadWriter, error) {
```
gmlewis: I don't think this or the next function are necessary. Please remove them.

shanempope:
FYI This PR has related / same feature: https://github.com/google/go-github/pull/1182

gmlewis:
Ah, thank you for the pointer, @shanempope!

It appears that #1182 never got completed... so when this PR is merged we can close that other PR, unless that one gets finished first, then we can close this one. :smiley:

github/apps.go
emoryruscus: Done.

github/apps.go
emoryruscus: Done.

github/apps.go
emoryruscus: Done.

github/apps.go
emoryruscus: Done.

github/apps.go
emoryruscus: They are used in apps_test.go, and I would like to use them in the ghinstallation package in a subsequent pull request as well. It's useful for comparing results in testing, and will be useful in converting installation token options into a ReadWriter to pass to http.NewRequest in ghinstallation.

https://github.com/bradleyfalzon/ghinstallation

emoryruscus:
Friendly ping, @gmlewis do you have a moment to look over the small changes I made?

gmlewis:
Hi @emoryruscus, I was waiting to review again until after you fix the broken Travis CI build.

Also, I'm not convinced that we need the helper functions and do not wish to increase the maintenance burden of this repo by adding unnecessary functions... I haven't read your argument yet, just saw that you added one.  I don't have time now, but was going to give it more detailed attention once you fix the broken build.

emoryruscus:
Sorry about that! Fixed the broken Travis CI build. Also moved the helper functions into apps_test.go, which I think you will like more.

github/apps.go
```diff
@@ -32,8 +32,21 @@ type App struct {
 
 // InstallationToken represents an installation token.
 type InstallationToken struct {
-	Token     *string    `json:"token,omitempty"`
-	ExpiresAt *time.Time `json:"expires_at,omitempty"`
+	Token        *string                  `json:"token,omitempty"`
+	ExpiresAt    *time.Time               `json:"expires_at,omitempty"`
+	Permissions  *InstallationPermissions `json:"permissions,omitempty"`
+	Repositories []*Repository            `json:"repositories,emitempty"`
```
gmlewis: This should be `omitempty`, not `emitempty`.

github/apps_test.go
```diff
@@ -330,3 +382,31 @@ func TestAppsService_FindUserInstallation(t *testing.T) {
 		t.Errorf("Apps.FindUserInstallation returned %+v, want %+v", installation, want)
 	}
 }
+
+// GetReadWriter converts a body interface into an io.ReadWriter object.
```
gmlewis: Yes, I much prefer these two new helper functions be used for testing only... thank you!

In fact, since they are rather general purpose, I would be fine to move them to `github_test.go` if you want, or we could just leave them here for now and move them in a later PR if they become more generally used.

We'll see what @gauntface thinks about them.

github/apps.go
emoryruscus: Great catch, thanks! Done.

github/apps_test.go
emoryruscus: Welcome! Will leave them here for now if that is alright.

emoryruscus:
No problem! @gauntface is there anything else you would like to see in this change?

emoryruscus:
Sounds good, thanks!

gmlewis:
Thank you, @emoryruscus and @gauntface !

Merging.

Note to self: This is a breaking API change so the major version needs to be bumped after merging.

PR #1240, opened 2019-07-25
Update git_refs.go

dprotaso:
Fix a segfault when fetching refs from github returns an error (say due to connectivity)


dprotaso:
> I'm surprised this was not caught months ago

@gmlewis I was only playing with this part of the API - so there may be similar issues in other calls?

Also feel free to squash the commits on merge :)

gmlewis:
Thank you, @gauntface!
Merging then bumping minor version due to bug fix.

PR #1241, opened 2019-07-26
fix: add start and end column to Annotation

mcristina422:
According to https://developer.github.com/v3/checks/runs/#annotations-object `start_column` and `end_column` are part of the Annotation

Additionally I noticed `blob_href` is no longer listed on the docs. Should I remove it in this PR or raise another?  
https://github.com/google/go-github/pull/1241/files#diff-065d35fd95f6db43a7395a47966b9505R54


gmlewis:
Thank you, @gauntface!
Merging.

PR #1242, opened 2019-07-26
remove blob_href from check annotations

mcristina422:
BREAKING

As defined on https://developer.github.com/v3/checks/runs/#annotations-object `blob_href` is no longer a parameter.

Opening this as discussed in https://github.com/google/go-github/pull/1241#pullrequestreview-266956347

### DIFF
--- a/github/checks.go
+++ b/github/checks.go
@@ -51,7 +51,6 @@ type CheckRunOutput struct {
 // CheckRunAnnotation represents an annotation object for a CheckRun output.
 type CheckRunAnnotation struct {
 	Path            *string `json:"path,omitempty"`
-	BlobHRef        *string `json:"blob_href,omitempty"`
 	StartLine       *int    `json:"start_line,omitempty"`
 	EndLine         *int    `json:"end_line,omitempty"`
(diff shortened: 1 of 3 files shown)


gmlewis:
Thank you, @gauntface!
@willnorris - I believe this constitutes a breaking API change, therefore requiring a major semver version bump... please let me know if you disagree and then I won't bump the major version.

willnorris:
@gmlewis yeah, I think you're right.  But that just means the next time a new version is tagged it should bump the major version.  It doesn't necessarily mean we need to tag this right away.

gmlewis:
Oh, OK. Do you have a preference, @willnorris ?

I've been attempting to release&tag after each PR just in case we need to refer to a specific version (so that we don't have to use a commit or PR as our reference), but I can reduce the frequency if you prefer.

I'll go ahead and continue catching up with the outstanding PRs while I've been without internet, and then make a new release and tag, and that should hopefully reduce the noise.

gmlewis:
Ah!  I see you opened #1280 for discussion.  Excellent.  We can continue there.

Issue #1248, opened 2019-08-11
Shouldn't GetArchiveLink accept 301 redirections?


Soulou:
Currently GetArchiveLink is only accepting 302: https://github.com/google/go-github/blob/0830f2846d799e0c1b07b643f79c4a2af7e54295/github/repos_contents.go#L264

But sometimes, a 301 is returned: I got errors like `unexpected status code: 301 Moved Permanently` which I assume is related to a renaming of repository.

Shouldn't it be handled?

Thanks a lot

gmlewis:
Yes, that seems reasonable, to at the very least report it as a redirection and then let the client handle trying again with the new URL, or handling it in whatever manner it desires.  In other words, I don't think this package should do anything with the information, just report it as a 301.

The GitHub API Developer docs just say: `To follow redirects with curl, use the -L switch:`.

Anyway, I'm opening up this issue as an opportunity for a contributor to address.

Thanks for the report, @Soulou!

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

Soulou:
@gmlewis Thanks for your answer, I'll try to have a look if I've time :D

vaibhavsingh97:
can I take this up?

Soulou:
Of course if you want it :-)

vaibhavsingh97:
@Soulou PR sent: https://github.com/google/go-github/pull/1249

Soulou:
Great thanks

gmlewis:
So let's start over on this one.

Specifically, for the `GetArchiveLink` endpoint, we would like the ability to detect and follow redirects.

I'm thinking that having the option to follow the redirect might be nice, and have that as the default option. So having a `followRedirects=false` might be good, then the response could be looked at instead of performing the second roundtrip call.

Before anyone jumps on this issue to implement it, let's first find out what @gauntface, @Soulou, or others think about this idea.

Soulou:
Having an option is always good for the client, to analyse weird cases/bugs etc, so I agree with it.
To me, the default behaviour should be should be to follow all 301, until a 302 happens and at the moment, returning the `Location`

jcockbain96:
[...]

PR #1249, opened 2019-08-12
Fixed: GetArchiveLink now accepts 301 redirection

vaibhavsingh97:
fixes: #1248

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -261,7 +261,7 @@ func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo st
 		return nil, nil, err
 	}
 	resp.Body.Close()
-	if resp.StatusCode != http.StatusFound {
+	if !(resp.StatusCode == http.StatusFound || resp.StatusCode == http.StatusMovedPermanently) {
 		return nil, newResponse(resp), fmt.Errorf("unexpected status code: %s", resp.Status)
 	}
 	parsedURL, err := url.Parse(resp.Header.Get("Location"))
--- a/github/repos_contents_test.go
+++ b/github/repos_contents_test.go
@@ -379,7 +379,7 @@ func TestRepositoriesService_GetArchiveLink(t *testing.T) {
 	if err != nil {
 		t.Errorf("Repositories.GetArchiveLink returned error: %v", err)
 	}
-	if resp.StatusCode != http.StatusFound {
+	if !(resp.StatusCode == http.StatusFound || resp.StatusCode == http.StatusMovedPermanently) {
 		t.Errorf("Repositories.GetArchiveLink returned status: %d, want %d", resp.StatusCode, http.StatusFound)
 	}
 	want := "http://github.com/a"


vaibhavsingh97:
@gmlewis please review 😄

github/repos_contents.go
```diff
@@ -261,7 +261,7 @@ func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo st
 		return nil, nil, err
 	}
 	resp.Body.Close()
-	if resp.StatusCode != http.StatusFound {
+	if !(resp.StatusCode == http.StatusFound || resp.StatusCode == http.StatusMovedPermanently) {
```
gmlewis: How about: 

`if resp.StatusCode != http.StatusFound && resp.StatusCode != http.StatusMovedPermanently {`

here and in the test file?

Soulou:
I've actually another point : 301 redirection happens when a repository has been renamed, so isn't there one more rountrip to do to get the archive ?

`https://api.github.com/user/repo/archive?ref=master` --- 301 ---> `https://api.github.com/user/new-name/archive?ref=master` --- 302 ---> `https://codeload.github.com/user/new-name/archive?ref=master&token=<token>?

gmlewis:
Yes, your client will need to recognize and follow the link if it wants it.
As I mentioned before, we are not going to have this package automatically follow redirects... it will be up to the user of this package to detect the redirect and perform whatever action it wants to perform.

gmlewis:
@vaibhavsingh97 - can you please verify that the redirect URL will be available to the user of this package so it can take action, and add that URL value to the unit test as a demonstration of usage?

Soulou:
@gmlewis In the case the 301 redirect would target the API requiring authentication, it's not possible and the package has to follow the redirection until we got the `Location` with the authenticated token

gmlewis:
The reason we do it this way, @Soulou is so that the user of this package has full control over what they want to do.

I can imagine that some program may wish to move a bunch of repos, and then use this API to poll GitHub periodically until the redirect is properly returned, and then report back that the moves have successfully completed (without actually needing to follow the redirect).

gmlewis:
@Soulou - I'm not sure I understand what you are saying is not possible.
If this API returns the URL along with the 301 in the response, then your program can then get the new URL and then make another request with the new URL to get the repo info.

Soulou:
@gmlewis what I mean is the following:

Without any renaming:

```
└> curl --user ":$TOKEN" https://api.github.com/repos/user/repo/tarball/master -v -q 2>&1 | grep Location
< Location: https://codeload.github.com/repo/user/legacy.tar.gz/master?token=AACQR4JB4AR5CSOK5KLWXG
```

The API returns an URL which is directly usable by the client of the package `google/go-github`, it's authentified by a token valid for 5 minutes (according to GitHub documentation)

Once the app has been renamed, 301 is returned with the following

```
└> curl --user ":$TOKEN" https://api.github.com/repos/user/renamed/tarball/master -v -q 2>&1 | grep Location
< Location: https://api.github.com/repositories/<ID>/tarball/master
```

So the user of this package would have to check if:

* It is an URL which can be used to download the tarball itself
* It is still an URL targeting the API which does not contain all the information to make another call again: (indeed we don't have the new name here, and the name is required in this package if I'm not mistaken):

```
	client.RepositoriesService.GetArchiveLink(
		ctx, owner, repo, github.Tarball,
		&github.RepositoryContentGetOptions{Ref: ref},
	)
```

I think it would be really a lot of responsability for the user of this package, while if an additional roundtrip would be done here, it would be much easier and logical. `GetArchiveLink` would always return an archive link as defined in https://developer.github.com/v3/repos/contents/#get-archive-link and not a link to analyze.

gmlewis:
Thank you for the explanation, @Soulou!

That makes a lot of sense, and this PR will not be sufficient to fix #1248.

Since only two lines were slightly modified here, I'm going to close this PR and let's start over on this one.  I'll also copy your detailed description to the desired solution into the original issue.

PR #1251, opened 2019-08-13
Fix failure in closing PRs

### LINKED ISSUE #1250: closing a Pull Request fails if PullRequest.Base is non-nil
After our recent upgrade to GHE 2.17.5, calls to PullRequestService.Edit to close PRs
started to fail with an error:
```
Message:"Validation Failed",
Resource:"PullRequest", 
Field:"base", 
Code:"invalid", 
Message:"Cannot change base branch of closed pull request."
```

Even though https://developer.github.com/v3/pulls/#update-a-pull-request doesn't explicitly states so, it seems that now the GitHub API doesn't allow to have the `base` field set when `state` = `'closed'`.  The `base` is being set in https://github.com/google/go-github/blob/master/github/pulls.go#L345-L347 - so you would get a failure unless you specifically set `PullRequest.Base` to `nil`.  It seems it can easily happen if you pass the `PullRequest` returned by https://godoc.org/github.com/google/go-github/github#PullRequestsService.Create or https://godoc.org/github.com/google/go-github/github#PullRequestsService.Get.

I'll raise a PR to suggest _not_ setting `update.Base = pull.Base.Ref` if `PullRequest.State` is `closed` (that is, if `Edit` is being called to close the Pull Request).

The issue was replicated both with the latest `v27.0.4` as well as `v25.1.1` (the code snippet in `github/pulls.go` is ~3 years old); tested with `go1.12.7`.

dmitris:
Fix an error in PullRequestsService.Edit when closing PRs
due to 'base' field being present together with
State = 'closed'.  In this case the GitHub API v2.17.5
responds with an error:
'Cannot change base branch of closed pull request'.

Fixes #1250.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -342,7 +342,10 @@ func (s *PullRequestsService) Edit(ctx context.Context, owner string, repo strin
 		State:               pull.State,
 		MaintainerCanModify: pull.MaintainerCanModify,
 	}
-	if pull.Base != nil {
+	// avoid updating the base branch when closing the Pull Request
+	// - otherwise the GitHub API server returns a "Validation Failed" error:
+	// "Cannot change base branch of closed pull request".
+	if pull.Base != nil && pull.GetState() != "closed" {
 		update.Base = pull.Base.Ref
 	}
 


gmlewis:
Thank you, @gauntface!
Merging, but not creating new release yet.

PR #1256, opened 2019-08-20
Add missing `CreatedAt` field in users_keys class

simonkey007:


github/users_keys.go
```diff
@@ -8,15 +8,17 @@ package github
...
-	Key      *string `json:"key,omitempty"`
-	URL      *string `json:"url,omitempty"`
-	Title    *string `json:"title,omitempty"`
-	ReadOnly *bool   `json:"read_only,omitempty"`
+	ID        *int64     `json:"id,omitempty"`
+	Key       *string    `json:"key,omitempty"`
+	URL       *string    `json:"url,omitempty"`
+	Title     *string    `json:"title,omitempty"`
+	ReadOnly  *bool      `json:"read_only,omitempty"`
+	CreatedAt *time.Time `json:"created_at,omitempty"`
```
gmlewis: I believe we have decided to make new time values be of type `Timestamp`.
Could you please change this from `*time.Time` to `*Timestamp`?

simonkey007: @gmlewis I've changed type of CreatedAt to *Timestamp

PR #1258, opened 2019-08-20
Add sort, direction parameters to Repositories.ListByOrg

Xopherus:


github/repos.go
```diff
@@ -201,6 +201,14 @@ type RepositoryListByOrgOptions struct {
 	// forks, sources, member. Default is "all".
 	Type string `url:"type,omitempty"`
 
+	// How to sort the repository list. Can be one of created, updated, pushed,
+	// full_name. Default: full_name
```
gmlewis: It looks like the docs say that `created` is the default option. Could you please update your comment?

github/repos.go
```diff
@@ -201,6 +201,14 @@ type RepositoryListByOrgOptions struct {
...
+	Sort string `url:"sort,omitempty"`
```
gmlewis: It looks like all three of these parameters need to be `*string` instead of `string` since they are optional.

Note to self, this PR would mean a breaking API change and require a major semver version number bump.

Xopherus: Wouldn't `omitempty` handle optional arguments @gmlewis ? And for what it's worth it seems like the other option structs in this file do not use `*string`

gmlewis: Ah yes, sorry, you are right.  I'll strike my comment.  Thank you, @Xopherus!

gmlewis:
Thank you, @gauntface!
Merging.

Issue #1270, opened 2019-08-31
Invalid response expected for ListPlanAccountsForAccount


domsu:
Currently `ListPlanAccountsForAccount` from `MarketplaceService` expects `[]*MarketplacePlanAccount`.

But documentation https://developer.github.com/v3/apps/marketplace/#check-if-a-github-account-is-associated-with-any-marketplace-listing says that `GET /marketplace_listing/accounts/:account_id` returns `*MarketplacePlanAccount`.

This is causing my application to fail.

gmlewis:
Closing issue as obsolete.

If anyone is still having this issue with the latest version of the repo, please first contact support@github.com to clarify the documentation as mentioned in #1271 and then reopen this issue and we can proceed with a fix.

JarmoDi2:
Hi @gmlewis, would it be possible to reopen this issue or merge the PR?

GitHub seem to have updated their docs and there's no more paging support for this endpoint:
https://docs.github.com/en/rest/reference/apps#get-a-subscription-plan-for-an-account

I'm also experiencing the same error with go-github v33.

gmlewis:
Thank you @JarmoDi2 .

Specifically, this method: https://github.com/google/go-github/blob/master/github/apps_marketplace.go#L129-L132
needs to have the `opts *ListOptions` parameter removed and the `GitHub API Docs:` URL updated.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

sethaustinx:
@gmlewis hi, I'd like to volunteer for this issue. I've made a PR.

gmlewis:
I'm now a little bit confused... I thought that @JarmoDi2 had requested this issue to be re-opened and wrote some detailed information as to why... I can only assume that the user deleted the comment, and the details with it, which is unfortunate.

PR #1271, opened 2019-08-31
Fixed ListPlanAccountsForAccount

domsu:
Fix for https://github.com/google/go-github/issues/1270


gauntface:
I will defer to @gmlewis but from from glance this looks like it would be better off being a new method.

gmlewis:
Thank you, @gauntface. Looking at the GitHub docs, this situation is very confusing indeed.

The docs show a single object being returned, yet imply that paging is supported with the call, and I don't know why.

Before we review and continue with this PR, @domsu - can you please contact support@github.com and ask them why paging is supported in this endpoint if it returns only a single object?

If they decide to remove paging support, then I think the method rename is appropriate along with the removal of the `ListOptions`. If, however, we learn something else from them, we can respond appropriately.

Thank you, @domsu!

marco-m:
Hello @domsu, do you have time to follow-up @gmlewis request? What is the status of this PR? Thanks.

gmlewis:
This PR appears to have been abandoned. Closing.

PR #1272, opened 2019-09-04
Add support for BranchListOptions

shotat:
Add support for BranchListOptions for `RepositoriesService.ListBranches`.

ref. https://developer.github.com/v3/repos/branches/#list-branches

PR #1273, opened 2019-09-04
Bump Go version to 1.13

shotat:
Updated `go.mod` to use Go 1.13.

SEE: https://github.com/google/go-github/pull/1272


gmlewis:
Thank you, @shotat, but I also needed to update the Travis CI configuration, so I went ahead and took care of that in another PR.

PR #1276, opened 2019-09-06
correct the api endpoint for issue comment reactions

xoba:
documentation at https://developer.github.com/v3/reactions/#list-reactions-for-an-issue-comment
gives this pattern: "GET /repos/:owner/:repo/issues/comments/:comment_id/reactions"
which i have merely translated into this one-liner fix. now works for me in my app.

### DIFF
--- a/github/reactions.go
+++ b/github/reactions.go
@@ -48,7 +48,7 @@ func (r Reaction) String() string {
 //
 // GitHub API docs: https://developer.github.com/v3/reactions/#list-reactions-for-a-commit-comment
 func (s *ReactionsService) ListCommentReactions(ctx context.Context, owner, repo string, id int64, opt *ListOptions) ([]*Reaction, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/comments/%v/reactions", owner, repo, id)
+	u := fmt.Sprintf("repos/%v/%v/issues/comments/%v/reactions", owner, repo, id)
 	u, err := addOptions(u, opt)
 	if err != nil {
 		return nil, nil, err
--- a/github/reactions_test.go
+++ b/github/reactions_test.go
@@ -16,7 +16,7 @@ func TestReactionsService_ListCommentReactions(t *testing.T) {
 	client, mux, _, teardown := setup()
 	defer teardown()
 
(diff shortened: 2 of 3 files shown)


gmlewis:
Thank you, @xoba.

It looks like the Travis CI build is failing with this error message:

```
--- FAIL: TestReactionsService_ListCommentReactions (0.00s)
    reactions_test.go:29: ListCommentReactions returned error: GET http://127.0.0.1:41559/api-v3/repos/o/r/issues/comments/1/reactions: 404  []
    reactions_test.go:33: ListCommentReactions = [], want [github.Reaction{ID:1, User:github.User{Login:"l", ID:2}, Content:"+1"}]
```

If you could fix the corresponding unit test, that would be great.

Also, from the documentation, it looks like we are not supporting the optional query parameters for this endpoint, but that could be addressed in another PR.  I'll create an issue for it.

xoba:
ah, my apologies @gmlewis, i didn't see or run `go test github.com/google/go-github/github/...` before posting, and i see the error now too when i do. i'll look into it and let folks know!

gmlewis:
@xoba - I believe we have a problem here.

In your description above, you refer to the GitHub Developer API documentation here:
https://developer.github.com/v3/reactions/#list-reactions-for-an-issue-comment

But the code you are editing is for a different endpoint, specifically here:
https://developer.github.com/v3/reactions/#list-reactions-for-a-commit-comment

Instead of changing this method, I believe you should be using this method instead:
https://github.com/google/go-github/blob/master/github/reactions.go#L155-L180

Please verify and close this PR if I am correct.

xoba:
oops, you are right, @gmlewis! ... i guess let's just chalk it up to me learning how to use this sdk properly. works for me now with [ListIssueCommentReactions](https://github.com/google/go-github/blob/9ee92864bda904cef6d83b83dd878efee8797bd4/github/reactions.go#L155-L180) method as is. closing this PR because it is ill conceived as you pointed out, thanks again!

gmlewis:
OK, cool.  No problem, @xoba. Glad to hear the other method does what you want.

Whenever I'm searching for methods to call, I often visit the GitHub Developer docs first, then search for either the URL in the comments in this repo or search for the expected URL string (like `repos/%v/%v/issues/comments/%v/reactions` for example). One of those usually leads me to the right method. :smile:

PR #1278, opened 2019-09-06
Adding ListOptions to ListMigrations

### LINKED ISSUE #1277: Add ListOptions to ListMigrations
ListMigrations doesn't currently accept ListOptions meaning pagination is not currently possibly.

Would be help to have the API change to `func (s *MigrationService) ListMigrations(ctx context.Context, org string, opts *ListOptions) ([]*Migration, *Response, error)`

gauntface:
This PR adds ListOptions to the ListMigrations method.

Fixes #1277

### DIFF
--- a/github/migrations.go
+++ b/github/migrations.go
@@ -103,7 +103,7 @@ func (s *MigrationService) StartMigration(ctx context.Context, org string, repos
 
 // ListMigrations lists the most recent migrations.
 //
-// GitHub API docs: https://developer.github.com/v3/migration/migrations/#get-a-list-of-migrations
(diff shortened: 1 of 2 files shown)


github/migrations.go
```diff
@@ -104,8 +104,12 @@ func (s *MigrationService) StartMigration(ctx context.Context, org string, repos
 // ListMigrations lists the most recent migrations.
 //
 // GitHub API docs: https://developer.github.com/v3/migration/migrations/#get-a-list-of-migrations
```
gmlewis: LGTM, but while you are here, @gauntface, could you please update the outdated URL to this?
https://developer.github.com/v3/migrations/orgs/#get-a-list-of-organization-migrations

PR #1279, opened 2019-09-07
Support go1.13

gmlewis:
Add support for go1.13 and bump Travis CI tests from 1.11 to 1.12 and 1.13.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,8 @@ language: go
 
 go:
   - "1.x"
-  - "1.11.x"
+  - "1.12.x"
+  - "1.13.x"
   - master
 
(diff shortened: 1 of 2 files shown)


willnorris:
You actually don't need to add 1.13 explicitly... the "1.x" version will always test the latest.   So what I've been doing is test master plus the latest two version, with "1.x" always capturing the most recent version.  So this should really now just be:

```yaml
go:
 - "1.x"
 - "1.12.x"
 - "master"
```

I'd rather not go beyond these three, primarily to keep our build times down.

gmlewis:
Thank you, @willnorris!
I'll delete that line.

PR #1285, opened 2019-09-17
add search topics

lucmski:


github/search.go
```diff
@@ -71,6 +71,22 @@ func (s *SearchService) Repositories(ctx context.Context, query string, opt *Sea
 	return result, resp, err
 }
 
+// TopicsSearchResult represents the result of a topics search.
+type TopicsSearchResult struct {
+	Total             *int         `json:"total_count,omitempty"`
+	IncompleteResults *bool        `json:"incomplete_results,omitempty"`
+	Repositories      []Repository `json:"items,omitempty"`
```
gmlewis: I don't think these results are `Repositories`.

In fact, to match the other `Search` endpoints, you will most likely need to create a new struct called `TopicResult` and then fill in the fields and tags from the GitHub v3 API developer docs.

So this should be something like:

```go
Topics []*TopicResult `json:items,omitempty`
```

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/google/go-github/v28
 require (
 	github.com/golang/protobuf v1.2.0 // indirect
 	github.com/google/go-querystring v1.0.0
+	github.com/k0kubun/pp v3.0.1+incompatible // indirect
```
gmlewis: Please revert the changes to this file.

go.sum
```diff
@@ -2,6 +2,12 @@ github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/google/go-querystring v1.0.0 h1:Xkwi/a1rcvNg1PPYe5vI8GbeBY/jrVuDX5ASuANWTrk=
 github.com/google/go-querystring v1.0.0/go.mod h1:odCYkC5MyYFN7vkCjXpyrEuKhc/BUO6wN/zVPAxq5ck=
+github.com/k0kubun/pp v3.0.1+incompatible h1:3tqvf7QgUnZ5tXO6pNAZlrvHgl6DvifjDrd9g2S9Z40=
```
gmlewis: Please revert the changes to this file.

example/topics/main.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
```
gmlewis: Please change `2017` to `2019`. We place the current year in each file as it is created. Thanks.

github/search.go
```diff
@@ -71,6 +71,34 @@ func (s *SearchService) Repositories(ctx context.Context, query string, opt *Sea
...
+type TopicResult struct {
+	Name             *string  `json:"name,omitempty"`
+	DisplayName      *string  `json:"display_name,omitempty"`
+	ShortDescription *string  `json:"short_description,omitempty"`
+	Description      *string  `json:"description,omitempty"`
+	CreatedBy        *string  `json:"created_by,omitempty"`
+	CreatedAt        *string  `json:"created_at,omitempty"`
+	UpdatedAt        *string  `json:"updated_at,omitempty"`
+	Featured         *bool    `json:"featured,omitempty"`
+	Score            *float32 `json:"score,omitempty"`
```
gmlewis: Let's please use `float64` here for consistency with the other endpoints.

github/search.go
```diff
@@ -71,6 +71,34 @@ func (s *SearchService) Repositories(ctx context.Context, query string, opt *Sea
...
+
+type TopicResult struct {
+	Name             *string  `json:"name,omitempty"`
+	DisplayName      *string  `json:"display_name,omitempty"`
+	ShortDescription *string  `json:"short_description,omitempty"`
+	Description      *string  `json:"description,omitempty"`
+	CreatedBy        *string  `json:"created_by,omitempty"`
+	CreatedAt        *string  `json:"created_at,omitempty"`
+	UpdatedAt        *string  `json:"updated_at,omitempty"`
+	Featured         *bool    `json:"featured,omitempty"`
```
gmlewis: Please add the `curated` field:

```go
Curated *bool `json:"curated,omitempty"`
```

github/search.go
```diff
@@ -71,6 +71,34 @@ func (s *SearchService) Repositories(ctx context.Context, query string, opt *Sea
...
+	Topics            []*TopicResult `json:"items,omitempty"`
+}
+
+type TopicResult struct {
+	Name             *string  `json:"name,omitempty"`
+	DisplayName      *string  `json:"display_name,omitempty"`
+	ShortDescription *string  `json:"short_description,omitempty"`
+	Description      *string  `json:"description,omitempty"`
+	CreatedBy        *string  `json:"created_by,omitempty"`
+	CreatedAt        *string  `json:"created_at,omitempty"`
```
gmlewis: Please change `*string` to `*Timestamp` here and on the line below as well.

github/search.go
```diff
@@ -71,6 +71,34 @@ func (s *SearchService) Repositories(ctx context.Context, query string, opt *Sea
...
+	UpdatedAt        *string  `json:"updated_at,omitempty"`
+	Featured         *bool    `json:"featured,omitempty"`
+	Score            *float32 `json:"score,omitempty"`
+}
+
+// Topics searches repositories via various topics.
```
gmlewis: This sentence is a little bit awkward and the query parameter can contain more than just topics, it can also include qualifiers. So how about if we make this read a bit more closely to the GitHub API docs and say something like:

```go
// Topics finds topics via various criteria. Results are sorted by best match.
// Please see https://help.github.com/en/articles/searching-topics for more
// information about search qualifiers.
```

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/google/go-github/v28
 require (
 	github.com/golang/protobuf v1.2.0 // indirect
 	github.com/google/go-querystring v1.0.0
+	github.com/mattn/go-colorable v0.1.2 // indirect
```
gmlewis: Please revert the contents of this file.

go.sum
```diff
@@ -2,6 +2,10 @@ github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/google/go-querystring v1.0.0 h1:Xkwi/a1rcvNg1PPYe5vI8GbeBY/jrVuDX5ASuANWTrk=
 github.com/google/go-querystring v1.0.0/go.mod h1:odCYkC5MyYFN7vkCjXpyrEuKhc/BUO6wN/zVPAxq5ck=
+github.com/mattn/go-colorable v0.1.2 h1:/bC9yWikZXAL9uJdulbSfyVNIR3n3trXl+v8+1sx8mU=
```
gmlewis: Please revert the contents of this file.

lucmski: done

github/search.go
lucmski: done

github/search.go
lucmski: done

github/search.go
lucmski: done

github/search.go
lucmski: done

example/topics/main.go
lucmski: done

go.mod
lucmski: done

github/search.go
lucmski: # github.com/google/go-github/v28/github
../../github/github-accessors.go:11924:9: cannot use *t.CreatedAt (type Timestamp) as type string in return argument
../../github/github-accessors.go:11972:10: cannot use t.Score (type *float64) as type *float32 in return argument

gmlewis: You need to re-run `go generate ./...`.  Please see `CONTRIBUTING.md` for details.

gmlewis:
Thank you, @wesleimp!
Merging.

PR #1292, opened 2019-09-28
Remove needless HeadBranch from Create/Update CheckRun APIs

haya14busa:
I couldn't find the changelog of this change but the current document
[1] doesn't mension `head_branch` field anymore.

I also confirmed the Check API works without the field.

[1]: https://developer.github.com/v3/checks/runs/#create-a-check-run


gmlewis:
Thank you, @wesleimp !
Merging.

PR #1295, opened 2019-10-03
Add list and create projects for users

### LINKED ISSUE #1294: Support List user projects and Create a user project
## Background

GitHub Projects support user-based projects and there are APIs about `List user projects` and `Create a user project`.

Seems go-github does not support them and I want to implement the feature.

## Goal

* Add `List user projects` and `Create a user project`

## Questions

If there are no on-going PRs to implement the features, please assign this issue to me 🙇 I will push PR.

takanabe:
## Background

GitHub Projects support user-based projects and there are APIs about `List user projects` and `Create a user project`.

Seems go-github does not support them and I want to implement the feature.

## Goal

* Add `List user projects` and `Create a user project`

Fixes #1294.

### DIFF
--- a/github/users_projects.go
+++ b/github/users_projects.go
@@ -1,0 +1,60 @@
+// Copyright 2019 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// ListProjects lists the projects for an user.
+//
+// GitHub API docs: https://developer.github.com/v3/projects/#list-user-projects
+func (s *UsersService) ListProjects(ctx context.Context, user string, opt *ProjectListOptions) ([]*Project, *Response, error) {
+	u := fmt.Sprintf("users/%v/projects", user)
+	u, err := addOptions(u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeProjectsPreview)
+
+	var projects []*Project
+	resp, err := s.client.Do(ctx, req, &projects)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return projects, resp, nil
+}
+
+// CreateProject creates a GitHub Project for the specified user.
+//
+// GitHub API docs: https://developer.github.com/v3/projects/#create-a-user-project
+func (s *UsersService) CreateProject(ctx context.Context, user string, opt *ProjectOptions) (*Project, *Response, error) {
+	u := fmt.Sprintf("users/%v/projects", user)
+	req, err := s.client.NewRequest("POST", u, opt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeProjectsPreview)
+
+	project := &Project{}
+	resp, err := s.client.Do(ctx, req, project)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return project, resp, nil
+}
--- a/github/users_projects.go
+++ b/github/users_projects.go
@@ -10,14 +10,6 @@ import (
 	"fmt"
 )
 
-// CreateUserProjectOptions specifies the parameters to the UsersService.CreateProject method.
-type CreateUserProjectOptions struct {
-	// The name of the project. (Required.)
-	Name *string `json:"name"`
-	// The description of the project. (Optional.)
-	Body *string `json:"body,omitempty"`
-}
-
 // ListProjects lists the projects for the specified user.
 //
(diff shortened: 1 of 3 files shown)


github/users_projects.go
```diff
@@ -0,0 +1,60 @@
...
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// ListProjects lists the projects for an user.
```
gmlewis: nit: Could you please change "an user" to "the specified user"?

github/users_projects.go
```diff
@@ -0,0 +1,60 @@
...
+	var projects []*Project
+	resp, err := s.client.Do(ctx, req, &projects)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return projects, resp, nil
+}
+
+// CreateProject creates a GitHub Project for the specified user.
```
gmlewis: Please change "the specified user" to "the current user".

github/users_projects.go
```diff
@@ -0,0 +1,60 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/projects/#create-a-user-project
+func (s *UsersService) CreateProject(ctx context.Context, user string, opt *ProjectOptions) (*Project, *Response, error) {
+	u := fmt.Sprintf("users/%v/projects", user)
```
gmlewis: This is not the correct URL.  It should be:
```
u := "users/projects"
```

You can remove the `user string` argument to the method.

github/users_projects.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, resp, err
+	}
+
+	return projects, resp, nil
+}
+
+// CreateProject creates a GitHub Project for the specified user.
+//
+// GitHub API docs: https://developer.github.com/v3/projects/#create-a-user-project
+func (s *UsersService) CreateProject(ctx context.Context, user string, opt *ProjectOptions) (*Project, *Response, error) {
```
gmlewis: I think a new struct needs to be created here because the parameters are not the same for this endpoint.  Something like:

```
// CreateUserProjectOptions specifies the parameters to the UsersService.CreateProject method.
type CreateUserProjectOptions {
	// The name of the project. (Required.)
	Name string `json:"name"`
	// The description of the project. (Optional.)
	Body *string `json:"body,omitempty"`
}
```

github/users_projects.go
takanabe: updated: https://github.com/google/go-github/pull/1295/commits/2f8a9ac24bbf38d656756098254b88002dcd5447

github/users_projects.go
takanabe: updated: https://github.com/google/go-github/pull/1295/commits/2f8a9ac24bbf38d656756098254b88002dcd5447

github/users_projects.go
takanabe: Right 👍  https://github.com/google/go-github/pull/1295/commits/5f386136afec56ced92f37734555e31889c5c637

github/users_projects.go
takanabe: @gmlewis  

thank you for your feedback!

I found that `RepositoriesService` and `OrganizationsService` use  `ProjectOptions` (ref: https://github.com/google/go-github/blob/9a2656da257b4c8208d61ee971374b502a6be4b1/github/projects.go#L64-L87).

`ProjectOptions` includes optional fields which `UsersService.CreateProject` does not need it. But, I use `ProjectOptions` to make the code consistent.

gmlewis: Unfortunately, it is no longer acceptable to send extra parameters to GitHub endpoints (I don't have time right now to point you to the other relevant issues, but they are closed in this repo and can be searched for), so we have taken the stance in this repo that we will not allow the uninformed user of this client library to mistakenly send extra parameters.  Therefore, the cleanest option is to create a new options struct.  Please do so.

takanabe: updated: https://github.com/google/go-github/pull/1295/commits/02658a3d6da0288de76057533a934bffc02d9b39

github/users_projects.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+// CreateUserProjectOptions specifies the parameters to the UsersService.CreateProject method.
+type CreateUserProjectOptions struct {
+	// The name of the project. (Required.)
+	Name *string `json:"name"`
```
gmlewis: There is no need for the pointer here since this field is required, so please change this from `*string` to `string`.

Also, if you feel like it, it might be nice to move this entire struct to immediately before the `CreateProject` method below so it is more tightly coupled to the location where it is used... but that's not a huge big deal, so I'll leave it up to you.

takanabe: Updated https://github.com/google/go-github/pull/1295/commits/c06ddc162172e25ac117e8e01c5a55f5f86a0c92.

Should I rebase commits to make this clean?

gmlewis: No, we avoid force pushes and rebasing in this repo because the final step is always a "Squash and Merge", and it additionally makes code reviews easier because reviewers can see just what changed since the last review. Thanks.

gmlewis:
Thank you, @wesleimp!
Merging.

Issue #1266, opened 2019-08-29
Search.Code not returning results on encoded query string


edaena:
Repro:
1. Call API directly:
```
https://api.github.com/search/code?q=prometheus+repo:microsoft/fabrikate-definitions+language:yaml
```
Seen: list of results

2. Call API via the go-github client
```
client.Search.Code(ctx, "prometheus+language:yaml+repo:microsoft/fabrikate-definitions", nil)
```
This results in an encoded query string `https://api.github.com/search/code?q=prometheus%2Brepo%3Amicrosoft%2Ffabrikate-definitions%2Blanguage%3Ayaml`

Expected: The same list of results as step 1
Seen: Validation error
```
{
  "message": "Validation Failed",
  "errors": [
    {
      "message": "Must include at least one user, organization, or repository",
      "resource": "Search",
      "field": "q",
      "code": "invalid"
    }
  ],
  "documentation_url": "https://developer.github.com/v3/search/"
}
```

gmlewis:
Thank you, @edaena!

Here is the code responsible for encoding the query string:
https://github.com/google/go-github/blob/master/github/search.go#L217-L231

And here are the unit tests that go with this section of code:
https://github.com/google/go-github/blob/master/github/search_test.go

It looks like the `Query` parameter as in `&searchParameters{Query: query}` is not tested, which is unfortunate.

Possibly-related issues are: #963, #966, and #1122.

We should add unit tests for a variety of searches.

For anyone who wants to take up this issue, this will require a good amount of experimentation with the GitHub web interface by performing searches on the website and then performing the same search with this repo, and making sure that the query strings match (and therefore the search results match). This person should also carefully read the history of this issue and the linked issues above and make sure that all the queries from the past still work, including the one listed above by @edaena.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

jmidd:
Hi - could I help with this issue ? Could you assign me - thanks !

gmlewis:
I can't assign it to you, but consider this issue yours, and please read again carefully what I wrote above before writing your PR, as we would like to heavily test a number of search queries and make sure they match queries performed in the GitHub web UI.

Thank you, @jmidd!

[...]

PR #1300, opened 2019-10-05
Add document for query of search functions

iamquang95:
Addressed the document part of https://github.com/google/go-github/issues/1266

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -24,6 +24,10 @@ import (
 // will search for such issues, sorting by creation date in ascending order
 // (i.e., oldest first).
 //
+// If query includes multiple condition, it MUST NOT include "+" as condition separator.
+// For example, querying with "language:c++" and "leveldb", then query should be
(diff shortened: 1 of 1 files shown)


iamquang95:
@googlebot I signed it!

github/search.go
```diff
@@ -24,6 +24,10 @@ import (
 // will search for such issues, sorting by creation date in ascending order
 // (i.e., oldest first).
 //
+// If query includes multiple condition, it MUST NOT include "+" as condition separator.
```
gmlewis: nit: Could you please tweak this slightly so that it is easier to read?  How about:

```
// If query includes multiple conditions, it MUST NOT include "+" as the condition separator.
```

iamquang95: Sure, i will add more detail (using " " instead) to the documentation. Thanks for your review.

gmlewis:
Closes #1266.

Issue #55, opened 2013-09-05
add tests for resource JSON marshalling


willnorris:
see 5f1c20afed08516e098c13fd0a94a74be0d573c3 for more details

palash25:
Hi @willnorris I would like to take this up. I will start with `projects_test.go` will make a PR soon.

palash25:
This can be labeled a good-first-issue since beginners can work on writing test for different resources at the same time.

jamesgeorge007:
@palash25 @gmlewis I would like to work on what is left further. Also, I'm not familiar with the code-base.
It would be great if you could guide me on resolving this issue :+1:

gmlewis:
Hi @jamesgeorge007! Thank you for your interest in helping out.

First, you will want to read our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide.

Second, you will want to get familiar with the code by reading through it and trying it out.

Then, I suggest following the link to the original issue where it is explained what is needed in more detail.

Thanks again, and I look forward to seeing PRs from you.

palash25:
@jamesgeorge007 hi :wave: just take a look at my commit referenced here and you will get the idea how to write the JSON test for different resources. Once it is clear just pick a resource that hasn't been covered by the tests and write a unit test for it :smile:

vaibhavsingh97:
@jamesgeorge007 Are you working on this issue? I would like to work on it 😅

palash25:
@vaibhavsingh97 I think there are multiple resources that need the JSON marhsalling test so if you start with `pulls_comments_test.go` I don't think there will be any problem since other contributors can pick up the left out resources. :+1:

wesleimp:
@gmlewis we can close this one

gmlewis:
I believe this is a long-running issue that is gradually being worked on by volunteers to add tests for resource JSON marshaling... so let's please keep this one open.

wesleimp:
Sure, no problems

ajz01:
I was thinking of adding some more tests for this issue as a way to get familiar with the repo and start contributing. Would it be helpful to go down the list of public types from the godocs and start adding tests for any types that don't already have marshaling tests built?

gmlewis:
Sounds great, @ajz01 - thank you!  Please make sure to read out CONTRIBUTING.md guide to help you get started.  Also, this issue links a number of PRs that you can use as examples.

gitdotvaibhav:
[...]

PR #1303, opened 2019-10-08
Test for marshalling Organizations and make JSON comparison non sensitive to key order

sgarciac:
This pull request contains two items:

1. Modifies the JSON comparison utility for checking marshalling, making it non sensible to key order.This makes it easier to write tests for marshalling, since there is no need to worry about keeping the right order of the JSON keys in the wanted string.  It also removes a non working part of the test. *I made sure old tests using this utility would still break if given the wrong input*.
2. Adds a new test for marshalling organizations (issue #55 )

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -6,7 +6,6 @@
 package github
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


github/github_test.go
```diff
@@ -135,32 +134,26 @@ func testBody(t *testing.T, r *http.Request, want string) {
 	}
 }
 
-// Helper function to test that a value is marshalled to JSON as expected.
+// Test whether the marshalling of v produces a JSON that corresponds
```
gmlewis: Let's please spell "marshal" and "unmarshal" consistently throughout the PR with a single "l".
Please also terminate the comment sentence with a period.

sgarciac: done

sgarciac: (also, added the issue I'm helping with to the description of the PR and the commit message)

github/github_test.go
```diff
@@ -135,32 +134,26 @@ func testBody(t *testing.T, r *http.Request, want string) {
 	}
 }
 
-// Helper function to test that a value is marshalled to JSON as expected.
+// Test whether the marshaling of v produces a JSON that corresponds
```
gmlewis: Let's please say "produces JSON" instead of "produces a JSON".

github/github_test.go
```diff
@@ -135,32 +134,26 @@ func testBody(t *testing.T, r *http.Request, want string) {
...
-	j, err := json.Marshal(v)
-	if err != nil {
-		t.Errorf("Unable to marshal JSON for %v", v)
+	// Unmarshal the wanted JSON, to verify its correctness, and marshal it back
+	// to sort the keys.
+	u := reflect.New(reflect.TypeOf(v)).Interface()
+	if err := json.Unmarshal([]byte(want), &u); err != nil {
+		t.Errorf("Unable to unmarshal JSON for %v: %v", want, err)
 	}
+	w, _ := json.Marshal(u)
```
gmlewis: Please test the returned `err` here instead of ignoring it.

github/github_test.go
sgarciac: done!

github/github_test.go
sgarciac: done! (however, I wonder if it's ever possible for the marshal here to fail, if the immediate unmarshal had no errors)

gmlewis: Probably not... but we would like to promote good Go programming practices for all developers who take a look at this code, whether they are new to Go or not... and if developers start ignoring error codes when they think there is no possible way that it could error, that opens up a potential world of pain further down the line.

Let's say, for example, we left it as-is, but then someone in the future decided to change the code above this that generated the value of `u`, but left this line the same. Suddenly, it might be possible that `u` was malformed, but this line would not catch that problem.

In summary, it is always safest to check returned error values in Go to minimize future negative surprises.  Thanks.

github/github_test.go
```diff
@@ -135,32 +134,29 @@ func testBody(t *testing.T, r *http.Request, want string) {
...
+	// Unmarshal the wanted JSON, to verify its correctness, and marshal it back
+	// to sort the keys.
+	u := reflect.New(reflect.TypeOf(v)).Interface()
+	if err := json.Unmarshal([]byte(want), &u); err != nil {
+		t.Errorf("Unable to unmarshal JSON for %v: %v", want, err)
+	}
+	w, err := json.Marshal(u)
 	if err != nil {
-		t.Errorf("Unable to marshal JSON for %v", v)
+		t.Errorf("Unable to marshal JSON for %v", u)
```
gmlewis: nit: please change this from `%v` to `%#v` in order to better see the contents of `u`.

sgarciac: done!

wesleimp:
👌

PR #1306, opened 2019-10-10
Add support for repository dispatches

fleskesvor:
Add repository dispatch to trigger workflow on GitHub actions.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -1427,3 +1427,31 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string, 
 
 	return r, resp, nil
 }
+
+// DispatchRequest represents a request to trigger a repository_dispatch event.
+type DispatchRequest struct {
+	EventType string  `json:"event_type"`
+}
+
+// Dispatch triggers a repository_dispatch event in a GitHub Actions workflow.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/#create-a-repository-dispatch-event
+func (s *RepositoriesService) Dispatch(ctx context.Context, owner, repo string, dispatch DispatchRequest) (*Repository, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/dispatches", owner, repo)
+
+	req, err := s.client.NewRequest("POST", u, &dispatch)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches.
+	req.Header.Set("Accept", mediaTypeRepositoryDispatchPreview)
+
+	r := new(Repository)
+	resp, err := s.client.Do(ctx, req, r)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, nil
+}
--- a/github/repos.go
+++ b/github/repos.go
@@ -1428,18 +1428,19 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string, 
 	return r, resp, nil
 }
 
-// DispatchRequest represents a request to trigger a repository_dispatch event.
-type DispatchRequest struct {
-	EventType string  `json:"event_type"`
+// DispatchRequestOptions represents a request to trigger a repository_dispatch event.
+type DispatchRequestOptions struct {
+	// EventType is a custom webhook event name. (Required.)
+	EventType string `json:"event_type"`
 }
 
 // Dispatch triggers a repository_dispatch event in a GitHub Actions workflow.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/#create-a-repository-dispatch-event
-func (s *RepositoriesService) Dispatch(ctx context.Context, owner, repo string, dispatch DispatchRequest) (*Repository, *Response, error) {
+func (s *RepositoriesService) Dispatch(ctx context.Context, owner, repo string, opts DispatchRequestOptions) (*Repository, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/dispatches", owner, repo)
 
-	req, err := s.client.NewRequest("POST", u, &dispatch)
+	req, err := s.client.NewRequest("POST", u, &opts)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/repos.go
+++ b/github/repos.go
@@ -7,6 +7,7 @@ package github
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 )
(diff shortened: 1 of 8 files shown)


go.sum
```diff
@@ -1,5 +1,6 @@
 github.com/golang/protobuf v1.2.0 h1:P3YflyNX/ehuJFLhxviNdFxQPkGK5cDcApsge1SqnvM=
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/google/go-github v17.0.0+incompatible h1:N0LgJ1j65A7kfXrZnUDaYCs/Sf4rEjNlfyDHW9dolSY=
```
gmlewis: Please revert the changes to this file.

github/repos.go
```diff
@@ -1321,3 +1321,31 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string,
 
 	return r, resp, nil
 }
+
+// DispatchRequest represents a request to trigger a repository_dispatch event.
+type DispatchRequest struct {
```
gmlewis: Let's please call this `DispatchRequestOptions` for consistency with the repo.

github/repos.go
```diff
@@ -1321,3 +1321,31 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string,
...
+	EventType string  `json:"event_type"`
```
gmlewis: Let's please give this field a comment, something like:

```
// EventType is a custom webhook event name. (Required.)
```

github/repos.go
```diff
@@ -1321,3 +1321,31 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string,
...
+}
+
+// Dispatch triggers a repository_dispatch event in a GitHub Actions workflow.
+//
+// GitHub API docs: https://developer.github.com/v3/repos/#create-a-repository-dispatch-event
+func (s *RepositoriesService) Dispatch(ctx context.Context, owner, repo string, dispatch DispatchRequest) (*Repository, *Response, error) {
```
gmlewis: Let's please change the name of `dispatch` to `opts` for more consistency with the repo.

fleskesvor:
Thanks for the feedback, @gmlewis. I've adressed the requested changes and pushed a new commit. I've also read the contributors doc, and I ran `go generate` on the first commit yesterday and the new one today, but the build is still failing. There must be something I'm missing, since I'm pretty new to Go, and the command seems to have no effect at all.

gmlewis:
Thanks, @fleskesvor.  Just FYI... the `CONTRIBUTING.md` doc should also talk about the use of `gofmt`... you should never have to hunt down and find an extra space... always run your code through `gofmt` and it will never be an issue.

I personally make sure that every editor I use runs `goimports` on saving the file which takes care of all imports and formatting for me.  The Go tooling is awesome.

fleskesvor:
Thank you, @gmlewis. :slightly_smiling_face: `goimports` work in my IDE, but I'll try to find out why `gofmt` isn't working as it should. Yes, the tooling is great, and the language has a ton of nice features.

fleskesvor:
@gmlewis Yes, you are right. The endpoint now accepts an additional field, `client_payload`, that was added after I opened my PR. It makes the endpoint a lot more useful, since until now I've had to stuff separate data into the `event_type` field separated with a dash, and parse it with `IFS=- read` in my workflows.

yeya24:
Any progress for this PR?

abatilo:
Looking to follow up here. Anything I can do to help drive this to completion?

fleskesvor:
Sorry for taking so long. I've tried to address the missing web hook and recently added client payload to my PR now.

github/repos.go
gmlewis: Hi @fleskesvor, it looks like this comment was not addressed. Could you please fix this?

github/repos.go
gmlewis: Hi @fleskesvor, it looks like this comment was not addressed. Could you please fix this?

github/repos.go
```diff
@@ -1427,3 +1428,35 @@ func (s *RepositoriesService) Transfer(ctx context.Context, owner, repo string,
...
 	return r, resp, nil
 }
+
+// DispatchRequest represents a request to trigger a repository_dispatch event.
+type DispatchRequest struct {
+	// EventType is a custom webhook event name. (Required.)
+	EventType string `json:"event_type"`
+	// ClientPayload is a custom JSON payload with extra information about the webhook event.
+	// Defaults to an empty JSON object.
+	ClientPayload *json.RawMessage `json:"client_payload"`
```
gmlewis: Are you sure this should be a pointer? Also, should this have `omitempty`?

This one field should have a bunch of table-driven unit tests to make sure that the payload is being passed and parsed correctly. Can you please add tests showing that this field works?

github/repos.go
fleskesvor: Oh, right. I actually did that before but changed it to dispatch in my last commit, after looking at the function directly above. I will change it back to `opts` again.

github/repos.go
fleskesvor: Oh, right. I actually did that before, but changed it to `DispatchRequest` after looking at the functiondirectly above. I will change it back again.

github/repos.go
fleskesvor: 1 and 2: These were addressed previously, but I forgot about the reason since last time, and un-addressed them in my last commit. I will re-address them again.
3: The reason it's a pointer is because it's optional, and I thought that was the best way to handle that. It definitely should have `omitempty`. And yes, it absolutely should have some unit tests. Do you have an example of a test in the code base where this is done in a good, idiomatic way?

fleskesvor:
OK, I've re-addressed the naming issues, and I've also added some tests for scenarios with different client payloads. I don't have enough experience with Go to tell if they are good or terrible though, and couldn't find anything similar to base them on in the existing code base.

fleskesvor:
Yes, I've tested it on a GitHub Actions workflow and verified that
- I can filter the action on dispatches:
```
on:
  repository_dispatch:
    types: [name_of_action]
```
- action is set in the context variable `github.event.action`
- that client payload is set in context variables, eg. `github.event.client_payload.foo` or `github.event.client_payload.bar`
- That the full payload, action/event_type and client payload is set in the environment variable `$GITHUB_EVENT_PATH`

I wish I could share an example repo, but I can't for the life of me figure out a clean way to use my branch as a dependency in a project, other than removing `go.mod` and pushing to a separate, throwaway repo, or perhaps merging on master in my own fork.

gmlewis:
Sounds great!  When we get a second LGTM we can merge.

gmlewis:
Thank you, @wesleimp!
Merging.

Issue #1268, opened 2019-08-30
Deleting a file using TreeEntry/CreateTree/CreateCommit?


bmcstdio:
I know it is possible to delete a file using `DeleteFile`, but that doesn't seem to support signing the resulting commit. Hence, I'd like to be able to use `TreeEntry`, `CreateTree` and `CreateCommit` to delete a file, but I can't seem to be able to do that. Is there an example of deleting a file using these APIs I can have a look at?

gmlewis:
Looking at our source-of-truth, which is the GitHub Developer API v3 (that this repo is simply a client of):
https://developer.github.com/v3/git/trees/#create-a-tree
It appears that you could create a tree and commit it and the new tree will completely replace the contents of the previous tree.

I have not done this myself and don't have time to create an example, but it looks like this is possible.

If other contributors feel like cooking up an example for @bmcstdio, you are welcome to, and I'll leave this issue open.

PR #1307, opened 2019-10-11
Add delete file support to CreateTree

helmus:
This solves https://github.com/google/go-github/issues/1268 

Deleting a file trough the `CreateTree` operation right now is not possible trough this SDK because it is not possible to have the `SHA` field serialized to json `null` due to the `omitempty` struct flag.

I propose a separate type `TreeDeleteEntry` so the delete operation can be made explicit. 
It's not possible to remove the `omitempty` struct tag on the existing type because this will cause a conflict with the content field when making the api call.

I may add tests later but just wanted to point this out at least.

### DIFF
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -142,7 +142,7 @@ func (s *createTree) UnmarshalJSON(data []byte) error {
 	decoded := map[string]interface{}{}
 	decodeErr := json.Unmarshal(data, &decoded)
 	if decodeErr != nil {
-		return fmt.Errorf("couldn't decode to map %w", decodeErr)
+		return fmt.Errorf("couldn't decode to map %v", decodeErr)
 	}
 	var typeOk bool
 	if s.BaseTree, typeOk = decoded["base_tree"].(string); !typeOk {
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -114,7 +114,7 @@ type createTree struct {
 
 type JsonObject map[string]interface{}
 
-func (h JsonObject) GetString (path string) *string {
+func (h JsonObject) GetString(path string) *string {
 	value, hasPath := h[path]
 	if !hasPath {
 		return nil
@@ -126,7 +126,7 @@ func (h JsonObject) GetString (path string) *string {
 	return &valueStr
 }
 
-func (h JsonObject) GetInt (path string) *int {
+func (h JsonObject) GetInt(path string) *int {
 	value, hasPath := h[path]
 	if !hasPath {
 		return nil
@@ -174,10 +174,10 @@ func (s *createTree) UnmarshalJSON(data []byte) error {
 			})
 		} else {
 			s.Entries = append(s.Entries, TreeDeleteEntry{
-				SHA:     nil,
-				Path:    itemAsJson.GetString("path"),
-				Mode:    itemAsJson.GetString("mode"),
-				Type:    itemAsJson.GetString("type"),
+				SHA:  nil,
+				Path: itemAsJson.GetString("path"),
+				Mode: itemAsJson.GetString("mode"),
+				Type: itemAsJson.GetString("type"),
 			})
 		}
 	}
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -112,9 +112,9 @@ type createTree struct {
 	Entries  []ITreeEntry `json:"tree"`
 }
 
-type JsonObject map[string]interface{}
+type jsonObject map[string]interface{}
 
-func (h JsonObject) GetString(path string) *string {
+func (h jsonObject) GetString(path string) *string {
 	value, hasPath := h[path]
 	if !hasPath {
 		return nil
@@ -126,7 +126,7 @@ func (h JsonObject) GetString(path string) *string {
 	return &valueStr
 }
 
-func (h JsonObject) GetInt(path string) *int {
+func (h jsonObject) GetInt(path string) *int {
 	value, hasPath := h[path]
 	if !hasPath {
 		return nil
@@ -158,7 +158,7 @@ func (s *createTree) UnmarshalJSON(data []byte) error {
 		if !isMap {
 			return errors.New("tree needs to be array of map")
 		}
-		itemAsJson := JsonObject(itemAsMap)
+		itemAsJson := jsonObject(itemAsMap)
 		sha := itemAsJson.GetString("sha")
 		content := itemAsJson.GetString("content")
 
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -7,6 +7,8 @@ package github
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 )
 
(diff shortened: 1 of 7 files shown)


helmus:
@googlebot I signed it!

gmlewis:
I'm sorry, @helmus, but I'm not happy with this approach.

Any time strongly-typed structs are replaced by `[]interface{}`, some careful thought needs to go into why this is being done.

You mentioned in the description that

>  it is not possible to have the SHA field serialized to json `null` due to the omitempty struct flag.

In Go, we typically solve this by making the string a pointer, and then sending an explicit `String("")`.

However, I'm wondering if we should first document how #1268 can be solved using `curl` first and then figure out how that translates to this Go client repo without sacrificing the nice data structs that have been built up.

Maybe @gauntface has other thoughts or comments about this.

helmus:
@gmlewis  totally understand your point. I'm also not a fan of interface{} collections.

The correct json structure to delete a file would look like this.

```
{
    "path": "file.rb"
    "mode": "100644"
    "type": "blob"
    "sha":  null
}
```

However, if we where to pass in `string("")` for `sha` in the `TreeEntry` struct, we would get the following:

```
{
    "path": "file.rb"
    "mode": "100644"
    "type": "blob"
    "sha":  ""
}
```

And the github api will reject this as it requires an explicit null. 
One way to solve it might be by implementing a custom Marshaler to differentiate between nil pointer and an empty string. This is a fairly implicit solution, and would have to be well documented, as it deviates from standard go struct tag marshaling. 
Another approach could maybe be to define a common interface for the two different types that simply returns a base type. So you can be a little bit more specific for the collection type. [Similar to this approach
](https://stackoverflow.com/questions/26027350/go-interface-fields)

Happy to update the code to an approach that suits you more and add tests to it. 
The main issue is that deleting now is not possible. We can not simply remove the `omitempty` tag from the SHA field on the existing `TreeEntry` type, because that would conflict with the `Content` field.

gmlewis:
OK, I think I understand.  What about creating a separate `DeleteTree` method that passes the appropriate fields without the changing the current `CreateTree` functionality?

helmus:
@gmlewis  if we do that, then we'll lose the ability to delete and modify files in the same commit.

I pushed a new approach in https://github.com/google/go-github/pull/1307/commits/d8aa19de24df01c4ee12ad2b32c835a4561a2c35

Here I defined a new type `TreeEntryBase` which contains the common fields that define a `TreeDeleteEntry` and `TreeEntry`. 
Using this type we can define an interface `ITreeEntry` which allows us to maintain more reasonable type safety in the CreateTree method.

However this does break backwards compatibility. As go won't allow implicit type conversions between arrays. 

We could also leave `CreateTree` untouched, ( maybe deprecate it ) and define a new function `CreateITree`. But then deleting would only be possible trough `CreateITree`, which won't be that obvious.

gmlewis:
I can't review right now, but just a quick comment that this sounds like a good approach. Also, it is OK to break the API because we use semantic versioning in this repo (although we obviously wish to minimize breaking changes except for good reasons which I believe this is one).

gmlewis:
I haven't reviewed the PR yet, @helmus, but it looks like the unit tests are passing with this error:

```
=== RUN   TestGitService_CreateTree
--- FAIL: TestGitService_CreateTree (0.00s)
    git_trees_test.go:79: Git.CreateTree request body: &{BaseTree:b Entries:[<nil>]}, want &{BaseTree:b Entries:[github.TreeEntry{SHA:"7c258a9869f33c1e1e1f74fbb32f07c86cb5a75b", Path:"file.rb", Mode:"100644", Type:"blob"}]}
=== RUN   TestGitService_CreateTree_Content
--- FAIL: TestGitService_CreateTree_Content (0.00s)
    git_trees_test.go:141: Git.CreateTree request body: &{BaseTree:b Entries:[<nil>]}, want &{BaseTree:b Entries:[github.TreeEntry{Path:"content.md", Mode:"100644", Content:"file content"}]}
```

You should be able to duplicate the problem locally by running `go test ./...`.  See `CONTRIBUTING.md` for more details.

github/git_trees.go
```diff
@@ -110,6 +112,78 @@ type createTree struct {
...
+
+		if sha != nil || content != nil {
+			s.Entries = append(s.Entries, TreeEntry{
+				SHA:     sha,
+				Path:    itemAsJson.GetString("path"),
+				Mode:    itemAsJson.GetString("mode"),
+				Type:    itemAsJson.GetString("type"),
+				Size:    itemAsJson.GetInt("size"),
+				Content: content,
+				URL:     itemAsJson.GetString("url"),
```
helmus: This now breaks the contract with the struct tags on `TreeEntry` and `TreeDeleteEntry`, reflection would need to be used in order to get the correct struct tags. I think it's probably fine since A: `createTree` is a private type, B: `createTree` unmarshalling is only used in tests

gmlewis:
@helmus - this is looking way more complicated than I hoped the solution would be.

Can we please take a step back for a moment and give an example of how someone would currently delete `FileA` and modify `FileB` within the same commit using `curl` and the GitHub v3 Developer API?

helmus:
I think we can probably simplify the approach by implementing custom marshaling and unmarshaling logic for just the `TreeEntry` struct, which would deviate a little from standard go struct tag marshaling logic.
This would get rid of the type hierarchy and is a little bit more in line with the api's interface. 

The complication with this approach is caused because we lose the type information when serialization objects of type `TreeDeleteEntry` and `TreeEntry`, and the tests expect the json objects to be serialized back into in `ITreeEntry` array.

When serializing the json objects back to an array of `ITreeEntry` go can not decide if the type of the json object is actually a `TreeDeleteEntry` or `TreeEntry`, as there is no explicit type information ( it could unmarshal to either one ).
Hence I implemented the `UnmarshalJSON` function for createTree in order to differentiate between `TreeDeleteEntry` and `TreeEntry` during unmarshalling. 
We can determine if the json object is a `TreeDeleteEntry` if both the sha and content field are empty / missing. 

I created this repo to create this commit trough curl:
https://github.com/helmus/go-github-example/commit/caa51d1f3cc553efa6064960d0730c4abc09c8b3
Here are the curl examples, I hope this clarifies it a little more for you.

**Request files**
```
create_tree.json
----------------
{
  "base_tree": "ff8e70e80b8639df2483ba6c9faaa9b38094eb15",
  "tree": [
    {
      "path": "FileA",
      "mode": "100644",
      "type": "blob",
      "sha": null
    },
    {
      "path": "FileB",
      "mode": "100644",
      "type": "blob",
      "content": "updating this content !"
    }
  ]
}

commit.json
----------------
{
  "message": "Delete and update file",
  "author": {
    "name": "willem dhaeseleer",
    "email": "dhwillem@gmail.com",
    "date": "2019-01-09T16:13:30+12:00"
  },
  "parents": [
    "8db90c2b21894bef2375a86fdd1074e64ea7340c"
  ],
  "tree": "09d88eece5f37ada28978dfd814a7417da9dd62a"
}

```


**Request responses:**
```
curl -s -X POST -d @create_tree.json -H "Authorization: token $MYTOKEN" https://api.github.com/repos/helmus/go-github-example/git/trees

// Note that the api does not actually return the deleted files in tree, so there is no inherent need to ever deserialize to a `TreeDeleteEntry` struct. 
// However, we do need to deserialize to an ITreeEntry type, which requires specifying a concrete type.

{
  "sha": "09d88eece5f37ada28978dfd814a7417da9dd62a",
  "url": "https://api.github.com/repos/helmus/go-github-example/git/trees/09d88eece5f37ada28978dfd814a7417da9dd62a",
  "tree": [
    {
      "path": "FileB",
      "mode": "100644",
      "type": "blob",
      "sha": "de887b15e780940c2bc2d2334b80e465a1fadc96",
      "size": 23,
      "url": "https://api.github.com/repos/helmus/go-github-example/git/blobs/de887b15e780940c2bc2d2334b80e465a1fadc96"
    },
    {
      "path": "README.md",
      "mode": "100644",
      "type": "blob",
      "sha": "09a00deca4dee4d75b3173334670361c881d7160",
      "size": 19,
      "url": "https://api.github.com/repos/helmus/go-github-example/git/blobs/09a00deca4dee4d75b3173334670361c881d7160"
    }
  ],
  "truncated": false
}

curl -s -X POST -d @commit.json -H "Authorization: token $MYTOKEN" https://api.github.com/repos/helmus/go-github-example/git/commits

{
  "sha": "caa51d1f3cc553efa6064960d0730c4abc09c8b3",
  "node_id": "MDY6Q29tbWl0MjE0OTExODM4OmNhYTUxZDFmM2NjNTUzZWZhNjA2NDk2MGQwNzMwYzRhYmMwOWM4YjM=",
  "url": "https://api.github.com/repos/helmus/go-github-example/git/commits/caa51d1f3cc553efa6064960d0730c4abc09c8b3",
  "html_url": "https://github.com/helmus/go-github-example/commit/caa51d1f3cc553efa6064960d0730c4abc09c8b3",
  "author": {
    "name": "willem dhaeseleer",
    "email": "dhwillem@gmail.com",
    "date": "2019-01-09T04:13:30Z"
  },
  "committer": {
    "name": "willem dhaeseleer",
    "email": "dhwillem@gmail.com",
    "date": "2019-01-09T04:13:30Z"
  },
  "tree": {
    "sha": "09d88eece5f37ada28978dfd814a7417da9dd62a",
    "url": "https://api.github.com/repos/helmus/go-github-example/git/trees/09d88eece5f37ada28978dfd814a7417da9dd62a"
  },
  "message": "Delete and update file",
  "parents": [
    {
      "sha": "8db90c2b21894bef2375a86fdd1074e64ea7340c",
      "url": "https://api.github.com/repos/helmus/go-github-example/git/commits/8db90c2b21894bef2375a86fdd1074e64ea7340c",
      "html_url": "https://github.com/helmus/go-github-example/commit/8db90c2b21894bef2375a86fdd1074e64ea7340c"
    }
  ],
  "verification": {
    "verified": false,
    "reason": "unsigned",
    "signature": null,
    "payload": null
  }
}


```

gmlewis:
Thank you for the detailed example, @helmus!

Now, can you please explain to me again why we can't simply remove the `omitempty` on the `SHA`, like this:
https://play.golang.org/p/6WpjlsJf9ii
?

What does this break?

And if it breaks something, can we come up with a simple way to optionally include or remove the `SHA` field?

helmus:
On my phone right now, But if you do that, the sha field will always be
present in the request. You must have only the sha field or the content
field in your request for github to accept it.
I do think a custom json Marshall function will be a cleaner solution here.
We will just need to document it well, as there simply is no standard way
of doing this with struct tags that I know off.

gmlewis:
I'm wondering if we can piece together a struct using an array of `json.RawMessage` to build up the correct JSON?  I'm going to have to think some more about this... I think there must be a simple an clean way to do it.  Thank you, @helmus!

helmus:
I will say I did not anticipate the need for unmarshalling the createTree
struct in my initial approach, that's why this approach got more
complicated. Having custom marshalling on just the TreeEntry object will
probably make more sense now. I'll see if I can whip up an example in a
separate branch tomorrow.

helmus:
closing in favor https://github.com/google/go-github/pull/1310

PR #1313, opened 2019-10-15
add tests for scrape package

willnorris:


scrape/apps_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func Test_AppRestrictionsEnabled(t *testing.T) {
+	client, mux, cleanup := setup()
+	defer cleanup()
+
```
gauntface: Would be nice to make these table driven tests:

```
tests := []struct {
	description    string
	testFile string
	org string
	want bool
}{
  {
     description: "return true for enabled orgs",
     testFile: "access-restrictions-enabled.html",
     org: "e",
     want: true,
  },
  ....
}
for _, tt := range tests {
    t.Run(tt.description, func(t *testing.T) {
        client, mux, cleanup := setup()
        defer cleanup()
        mux.HandleFunc(fmt.Sprintf("/organizations/%v/settings/oauth_application_policy", tt.org), func(w http.ResponseWriter, r *http.Request) {
		copyTestFile(w, tt.file)
	})
        got, err := client.AppRestrictionsEnabled(tt.org)
	if err != nil {
		t.Errorf("AppRestrictionsEnabled() returned err: %v", err)
	}
	if want := true; got != want {
		t.Errorf("AppRestrictionsEnabled(); got %t, want %t", got, want)
	}
    })
}
```

scrape/forms_test.go
```diff
@@ -0,0 +1,73 @@
...
+			  <form action="a2" method="m2"><input name="n2" value="v2"></form>
+			</html>`,
+			[]Form{
+				{Action: "a1", Method: "m1", Values: url.Values{"n1": {"v1"}}},
+				{Action: "a2", Method: "m2", Values: url.Values{"n2": {"v2"}}},
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
gauntface: Please use:

```
for _, tt := range tests {
   t.Run(tt.description, func(t *testing.T){
       ....
   })
}
```

This makes the logs clear which test case fails (should it ever fail).

scrape/scrape_test.go
```diff
@@ -0,0 +1,33 @@
+package scrape
```
gauntface: It might be nice to pull this out into it's own package so that when using setup() or copyTestFile() it's clear where those functions are coming from.

scrape/apps_test.go
willnorris: done

scrape/forms_test.go
willnorris: done

scrape/scrape_test.go
willnorris: ack.  might do this in the future.

Issue #1318, opened 2019-10-23
Rate Limit handling may not work with GitHub Enterprise


tkhandel:
In the code at [github/github.go#L766](https://github.com/google/go-github/blob/master/github/github.go#L766), it parses the response body to check if the message starts with the words "API rate limit exceeded for ". 

Besides being brittle, it is also plain wrong for an enterprise version of GitHub. Reason - The error message which is returned seems to be a configurable option where an enterprise admin can tweak the exact message that is returned. I know this because at my company, the error message which is returned is different than this. Hence, the library is not able to correctly identify that the GitHub's rate limit has been reached. 

To rectify, remove the response body check. The HTTP status code and the corresponding rate-limiting headers ("X-RateLimit-Limit", "X-RateLimit-Remaining", "X-RateLimit-Reset") should be enough to confirm that it is indeed a rate-limiting error.

gmlewis:
Thanks, @tkhandel.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Please make sure that your unit tests are comprehensive for this change.

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

xitter:
@gmlewis can I work on this?

tkhandel:
@gmlewis Can you assign it to me?

gmlewis:
Thank you, @xitter, but let's let @tkhandel work on this since @tkhandel is the issue reporter.

This is actually my fault for opening it up to everyone before asking if the issue author wanted to work on it, so I apologize. Thanks for your understanding.

@tkhandel - it is yours.  Thanks.

tkhandel:
Thank you, @gmlewis. I will put up the PR soon.

PR #1321, opened 2019-10-25
Simplify rate limit error check

tkhandel:
PR for the enhancement mentioned in Issue #1318. Also added unit tests for both RateLimitError and AbuseRateLimitError.

Why remove the message check for RateLimitError?
The message can be different for enterprise versions of GitHub. Besides, the rate limit header and HTTP status codes are sufficient to identify a rate limit error.

Why not remove the URL check for AbuseRateLimitError?
HTTP response in this situation does not contain any unique header values that would be sufficient to confirm an abuse. Although it is not ideal to do a suffix match on documentation_url, there seems to be no better way.

Reference to GitHub API rate limiting: https://developer.github.com/v3/#rate-limiting

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -648,7 +648,7 @@ type TwoFactorAuthError ErrorResponse
 func (r *TwoFactorAuthError) Error() string { return (*ErrorResponse)(r).Error() }
 
 // RateLimitError occurs when GitHub returns 403 Forbidden response with a rate limit
-// remaining value of 0
+// remaining value of 0.
 type RateLimitError struct {
 	Rate     Rate           // Rate specifies last known rate limit for the client
 	Response *http.Response // HTTP response that caused this error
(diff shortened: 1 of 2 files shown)


tkhandel:
@googlebot I signed it!

github/github.go
```diff
@@ -648,7 +648,7 @@ type TwoFactorAuthError ErrorResponse
 func (r *TwoFactorAuthError) Error() string { return (*ErrorResponse)(r).Error() }
 
 // RateLimitError occurs when GitHub returns 403 Forbidden response with a rate limit
-// remaining value of 0, and error message starts with "API rate limit exceeded for ".
+// remaining value of 0
```
gmlewis: nit: please add the ending period ('.') back in for the auto-generated godocs.

tkhandel: Added the period.

mfridman:
Tested this out (coincidentally, obviously) and works as expected.

Looking forward to seeing this merged in so client can correctly surface a `*github.AbuseRateLimitError` error

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1335, opened 2019-11-19
Add App Manifest support.

### LINKED ISSUE #1334: FR: GitHub App Manifests
Tracking bug for adding support for GitHub App Manifests - https://developer.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest

Happy to take this on myself.

wlynch:
This adds support for creating GitHub Apps programmatically. See
https://developer.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest
for more details.

CreateApp was added to scrape, since it is effectively emulating the
POST the UI makes, and doesn't actually interact with api.github.com.

Fixes #1334.


github/apps_manifest.go
```diff
@@ -0,0 +1,54 @@
+package github
```
willnorris: needs standard license header (same as other files)

github/apps_manifest.go
```diff
@@ -0,0 +1,54 @@
...
+	WebhookSecret *string    `json:"webhook_secret,omitempty"`
+	PEM           *string    `json:"pem,omitempty"`
+}
+
+// AppManifestService contains methods for interacting with GitHub App
+// Manifests.
+// See
+// https://developer.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest
+// for more details.
+type AppManifestService service
```
willnorris: I feel like this probably belongs on the existing `AppsService`, rather than an entirely new one?

github/apps_manifest.go
```diff
@@ -0,0 +1,54 @@
...
+
+// GetConfig completes the App manifest handshake flow for the given code.
+// See
+// https://developer.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest/#3-you-exchange-the-temporary-code-to-retrieve-the-app-configuration
+func (s *AppManifestService) GetConfig(ctx context.Context, code string) (*AppConfig, *Response, error) {
```
willnorris: `GetConfig` sounds like it's just retrieving data, but what's happening here is actually the final step in the creation of the app.  So maybe `CreateAppFromManifest` or `CompleteAppManifest` or something like that?

github/apps_manifest_test.go
```diff
@@ -0,0 +1,48 @@
+package github
```
willnorris: add license header

scrape/apps.go
```diff
@@ -103,3 +107,40 @@ type OAuthApp struct {
 	State       OAuthAppReviewState
 	RequestedBy string
 }
+
+// Manifest represents a GitHub App manifest, used for preconfiguring
+// GitHub App configuration.
+type Manifest struct {
```
willnorris: `AppManifest` might be a little clearer

github/apps_manifest.go
wlynch: Done.

github/apps_manifest.go
wlynch: Done.

github/apps_manifest.go
wlynch: Done.

github/apps_manifest_test.go
wlynch: Done.

scrape/apps.go
wlynch: Done.

github/apps_manifest.go
wlynch: Okay actually done now. :wink:

PR #1336, opened 2019-11-20
Add option to follow 301 redirects for GetArchiveLink

jcockbain:
## Summary

Adds a `followRedirects` parameter, which when `true` will follow the URLs returned with `301 Moved Permanently` statuses (i.e if a repo has been renamed). Once a 302 status is found, that `Location` will be returned to the user.

## Testing 

Adds 2 unit tests for 301 status code return, for when `followRedirects` is `true` and `false`. 

Manually tested, invoking the function against repos both renamed and not renamed.

Fixes https://github.com/google/go-github/issues/1248

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -240,16 +240,27 @@ const (
 // or github.Zipball constant.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/contents/#get-archive-link
-func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo string, archiveformat archiveFormat, opt *RepositoryContentGetOptions) (*url.URL, *Response, error) {
+func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo string, archiveformat archiveFormat, dontFollowRedirects bool, opt *RepositoryContentGetOptions) (*url.URL, *Response, error) {
 	u := fmt.Sprintf("repos/%s/%s/%s", owner, repo, archiveformat)
 	if opt != nil && opt.Ref != "" {
 		u += fmt.Sprintf("/%s", opt.Ref)
 	}
+	resp, err := s.getArchiveLinkFromURL(ctx, dontFollowRedirects, u)
+	resp.Body.Close()
+	if resp.StatusCode != http.StatusFound {
+		return nil, newResponse(resp), fmt.Errorf("unexpected status code: %s", resp.Status)
+	}
+	parsedURL, err := url.Parse(resp.Header.Get("Location"))
+	return parsedURL, newResponse(resp), err
+}
+
+func (s *RepositoriesService) getArchiveLinkFromURL(ctx context.Context, dontFollowRedirects bool, u string) (*http.Response, error) {
+	var resp *http.Response
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
-	var resp *http.Response
+
 	// Use http.DefaultTransport if no custom Transport is configured
 	req = withContext(ctx, req)
 	if s.client.client.Transport == nil {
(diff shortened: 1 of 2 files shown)


jcockbain:
@googlebot I signed it!

github/repos_contents.go
```diff
@@ -240,16 +240,27 @@ const (
...
+
+func (s *RepositoriesService) getArchiveLinkFromURL(ctx context.Context, u string, followRedirects bool) (*http.Response, error) {
+	var resp *http.Response
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
-	var resp *http.Response
+
```
gmlewis: nit: can you please move line 258 down to after this blank line 263 to make the declaration of `resp` closer to its use?

github/repos_contents.go
```diff
@@ -240,16 +240,27 @@ const (
...
 //
 // GitHub API docs: https://developer.github.com/v3/repos/contents/#get-archive-link
-func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo string, archiveformat archiveFormat, opt *RepositoryContentGetOptions) (*url.URL, *Response, error) {
+func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo string, archiveformat archiveFormat, opt *RepositoryContentGetOptions, followRedirects bool) (*url.URL, *Response, error) {
 	u := fmt.Sprintf("repos/%s/%s/%s", owner, repo, archiveformat)
 	if opt != nil && opt.Ref != "" {
 		u += fmt.Sprintf("/%s", opt.Ref)
 	}
+	resp, err := s.getArchiveLinkFromURL(ctx, u, followRedirects)
+	resp.Body.Close()
```
gmlewis: It looks to me like the body is being closed twice - once on line 280, then a second time on line 249.

(Additionally, `err` needs to be checked first before `resp.Body.Close()` because `resp` might be `nil`...
however, I believe the correct solution is to actually delete line 249 entirely, since it is already being handled.)

github/repos_contents.go
```diff
@@ -258,12 +269,14 @@ func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo st
...
-		return nil, newResponse(resp), fmt.Errorf("unexpected status code: %s", resp.Status)
+
+	// If redirect response is returned, follow it
+	if followRedirects && resp.StatusCode == http.StatusMovedPermanently {
+		u = resp.Header.Get("Location")
+		resp, err = s.getArchiveLinkFromURL(ctx, u, false)
 	}
-	parsedURL, err := url.Parse(resp.Header.Get("Location"))
-	return parsedURL, newResponse(resp), err
+	resp.Body.Close()
```
gmlewis: I believe this `resp.Body.Close()` needs to be moved up to the line following 273.
Otherwise, if a `http.StatusMovedPermanently` is received, the `resp.Body` will get closed once on line 280 from the child, then once again on line 280 from the parent.

jcockbain:
Thanks @gmlewis for the quick review. I've committed addressing the issues.

jcockbain:
Thanks @gmlewis! 

Do you know when we might get a second look on this? I'm being asked as to whether we'll be able use the functionality in our next release.

gmlewis:
Hi @jcockbain - this repo is maintained by volunteers so I don't have an ETA on a second LGTM, and we need two reviewers on this PR before we can merge.

My best recommendation to you is to patch your local copy of this repo with this PR included if you need the functionality until we can get it merged.

Thanks for your patience and understanding.

jcockbain:
No problem @gmlewis - thanks for clarifying.

gmlewis:
Thank you, @wesleimp!
Merging.

PR #1338, opened 2019-11-21
Fix links to GitHub API docs

suhaibmujahid:


suhaibmujahid:
@googlebot I signed it!

PR #1340, opened 2019-11-26
Make IssueListCommentsOptions consistent with GitHub API

### LINKED ISSUE #1339: IssueListCommentsOptions is not consistent with GitHub API
## Problem
In the GitHub API document, an endpoint "list-comments-on-an-issue" takes only one parameter `since`.
https://developer.github.com/v3/issues/comments/#list-comments-on-an-issue

But `IssueListCommentsOptions` in this library takes 3 parameters `sort`, `direction` and `since`.
https://godoc.org/github.com/google/go-github/github#IssueListCommentsOptions

## Goal
Remove `sort` and `direction` from `IssueListCommentsOptions`.

tfujiwar:
Closes #1339 .

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -4588,6 +4588,30 @@ func (i *IssueEvent) GetURL() string {
 	return *i.URL
 }
 
+// GetDirection returns the Direction field if it's non-nil, zero value otherwise.
+func (i *IssueListCommentsOptions) GetDirection() string {
+	if i == nil || i.Direction == nil {
+		return ""
+	}
+	return *i.Direction
+}
+
+// GetSince returns the Since field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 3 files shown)


gmlewis:
As I commented in #1339:

No, this is not entirely correct.  This endpoint also serves this call:
https://developer.github.com/v3/issues/comments/#list-comments-in-a-repository

So when listing comments on an issue, the user should not populate `sort` or `direction`.

However, I see that there is indeed a bug. The `omitempty` is basically not useful as written because the two `string` options and the `time.Time` option should all be pointers.

The correct fix here is to make those three options pointers.

tfujiwar:
Thank you for pointing out the mistake. I've fixed the code as you mentioned.

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1343, opened 2019-11-26
Small enhancement to Client.Do

suhaibmujahid:
- Remove unnecessary `return` (line `565`)
- Add a missed `else if` statements (line `579`)

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -562,7 +562,6 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 			}
 
 			aerr.Raw = b
-			return response, aerr
 		}
 
 		return response, err
@@ -575,8 +574,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 			decErr := json.NewDecoder(resp.Body).Decode(v)
 			if decErr == io.EOF {
 				decErr = nil // ignore EOF errors caused by empty response body
-			}
-			if decErr != nil {
+			} else if decErr != nil {
 				err = decErr
 			}
 		}


github/github.go
```diff
@@ -575,8 +574,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 			decErr := json.NewDecoder(resp.Body).Decode(v)
 			if decErr == io.EOF {
 				decErr = nil // ignore EOF errors caused by empty response body
-			}
-			if decErr != nil {
+			} else if decErr != nil {
```
gmlewis: Actually, lines 574-579 could be simplified to this, which I think is clearer:

```go
if decErr := json.NewDecoder(resp.Body).Decode(v); decErr != io.EOF {
  err = decErr  // ignore EOF errors caused by empty response body
}
```

But I'm betting that the original code review wanted to make an explicit case for `io.EOF` and wanted to keep the logic super-easy to read...

So I'm leaning toward closing this entire PR since the first part is not correct, and this change is not needed.

github/github.go
```diff
@@ -562,7 +562,6 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 			}
 
 			aerr.Raw = b
-			return response, aerr
```
gmlewis: Sorry, this change is not correct. Line 565 returns `aerr` and line 568 returns `err`.

github/github.go
suhaibmujahid: I agree with you it's important to keep the logic super-easy to read. I meant with this change only removing the extra execution of `decErr != nil`. So if the value of  `decErr` is  `io.EOF`, there is no need to do again `decErr != nil` because for sure it will never be `nill`.

gmlewis: I understand, but my point was that sometimes we decide to waste a CPU cycle or two just for the sake of readability of the code, and I would have to go back to the original code review where these lines were discussed to find out if that was the case here.

github/github.go
suhaibmujahid: Thank you for your comment.

I think both of `aerr` and `err` are pointing to the same error struct. The `if` statement in the line `557` does assertion if the error type is `*AcceptedError`, if so, it copies the response body to the error without replacing it with another error. So, I think they are the same.

Do I miss something?

github/github.go
suhaibmujahid: I see. Thank you for the clarification, it makes sense.

github/github.go
gmlewis: Ah, I see what you are saying.  Yes, I think you are correct.
Let's see what @gauntface thinks about these changes.

gmlewis:
I don't see much benefit in this PR.  Closing.

Issue #1308, opened 2019-10-12
Discussion: github scraping package


willnorris:
One of the frustrations we've long had at Google is the amount of functionality that GitHub does **not** expose through either the REST or graphql API, for whatever reason.  Most recently, this has been around things like listing the OAuth and GitHub Apps that have been approved for a GitHub organization.  There are no official APIs for this (and lots of other) functionality, so you have to just do it manually through the GitHub UI, or screen scrape.

We've built a few single-purpose screen scraping tools like this using [GoogleChrome/puppeteer](https://github.com/GoogleChrome/puppeteer), but that's a bit more complicated (and heavyweight) to run in our production environment than a pure Go approach.

A long time ago, I had talked about having an experimental package in go-github that provided a clean API for some of this scraping functionality, but I never actually built it until now.

I now have a proof of concept that successfully authenticates to github.com, including performing two-factor-auth, and can then scrape some of the additional data we need.  The code is reasonably clean, but probably not particularly stable, given that it relies on specific class names and IDs of elements on the GitHub site.

**So my question is, where should this scraping package live?**

How do folks feel about having an experimental "scrape" package in go-github?  It would be forever unstable, and would probably need to be exempt from any versioning policy that we have for the library as a whole.  My main thought for keeping it in go-github (versus a standalone githubscrape repo, or whatever) is just visibility, but also because it would very specifically be designed to fill the gaps in the REST API, so it does actually track with the main go-github library itself pretty closely.

/cc @gauntface @gmlewis @dmitshur

gmlewis:
SGTM!

willnorris:
scraper added in 6b2a6456a7a093a3da8fc0f474afb4d1c71c1885.  We'll continue to build it out over time, but closing this issue.

dmitshur:
I think it's reasonable to include it in this repository, but it may make sense to keep it in a separate module. That way, its dependencies wouldn't affect the main module, and it can be versioned separately. The main module undergoes major version bumps quite often, and I understand those are not going to be useful to `scape` since it's effectively at v0.

willnorris:
@dmitshur that's a great idea (putting it into a separate module).  Since the main module is defined at the root of the repository, can you define a separate module in a sub-path?  I've never tried that.

willnorris:
Okay, it seems like it "just works"?  Though it looks like we might need to move the scraping example into the module as well so that goquery and friends don't get picked up as indirect dependencies in the parent module.  I'll send out a second PR.

[...]

PR #1344, opened 2019-12-04
Bump major version to v29

gmlewis:
Note to @willnorris - the only way I could figure out to get this to work (now with the inclusion of the scrape package) was to remove the `go.mod` and `go.sum` from the `scrape` directory.  (Because v29 of `go-github` does not exist until the PR is committed and the repo is tagged.)

I'm thinking this makes the most sense anyway since the top-level `go.mod` covers the whole repo.

If you find a better mechanism, though, I'm totally happy to close this PR and go with the better solution.


gmlewis:
Note to self: If this gets merged, then a release of this repo should be tagged with `v29.0.0` before other PRs get merged.

willnorris:
having a separate go.mod for the scrape package was very intentional (see https://github.com/google/go-github/issues/1308#issuecomment-542230233).  I likely won't be able to fully review this today, but do please wait to submit, as I'd like very much to keep the modules separate if at all possible.

/cc @dmitshur for comment and review as well

willnorris:
what if you just leave the scrape package as relying on v28, and don't update that one in this change?  Then, after this PR has landed and the new tag created, update the scrape package to use v29 in a separate change.  Would that work?

gmlewis:
Ah! OK. Sorry about that. I will fix it tonight and let you know when it is ready for review.
Thank you, Will!

gmlewis:
Hi @willnorris - I believe this preserves the scrape versioning.

(I was a bit surprised that the dependencies of scrape are now being brought up into the top-level `go.mod`... and that they weren't before when scrape was first brought in. All I did to update `go.mod` was to run `go test ./...` and then all the scrape dependencies showed up... but I believe this is correct.)

PTAL.

dmitshur:
I agree with Will’s comments, we must not remove the scrape/go.mod file, and it’s a good idea to keep this PR focused on just one module; updating the scrape module can be more safely done in later PRs.

Trying to make changes to more than one module (especially ones that are interconnected) in one PR is tricky and not worth doing until feeling very confident about the process.

Glenn, have you done `go mod tidy` in the latest version of this PR?

gmlewis:
Thank you, @dmitshur!

Now I've run 'go mod tidy'. :smile:

PTAL.

go.mod
```diff
@@ -1,10 +1,10 @@
-module github.com/google/go-github/v28
+module github.com/google/go-github/v29
 
 require (
 	github.com/golang/protobuf v1.3.2 // indirect
 	github.com/google/go-querystring v1.0.0
 	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
-	golang.org/x/net v0.0.0-20190311183353-d8887717615a // indirect
+	golang.org/x/net v0.0.0-20191014212845-da9a3fd4c582 // indirect
```
dmitshur: This change doesn’t look intentional.

I suspect it came about from the earlier version of this CL where the scrape/go.mod file was removed, and its requirements merged with that of the main go-github module. The scrape module happened to require a newer version of x/net:

https://github.com/google/go-github/blob/04a52c26bbdd106ed23331c4d61fde482ffb9413/scrape/go.mod#L10

And so it stayed around after `go mod tidy` because this module also needs x/net.

It’s not harmful to keep it since it’s just a newer x/net, but it’s also probably better to make changes intentionally rather than due to chance.

You can go back to original or stay with the newer, I’m okay with both.

gmlewis: Ah, I see.  Thank you, @dmitshur!
Reverted.

gmlewis:
We discussed this PR in #1365.  Should we merge master into this PR before merging or just move ahead as-is?

I would like to get approval from @willnorris on this PR since this is the first one containing `scrape`.

Thank you!

gmlewis:
Thank you, @willnorris !
Merging.

gmlewis:
Ugh... I broke the build.  Trying to figure out how to roll this back.

gmlewis:
#1367 should fix this release.

Issue #1260, opened 2019-08-22
panic occurred when a repository is empty by some go-github API


yunsoul:
* When a repository is empty(no code & no branch & no README.md), some go-github API made panic

* the go-github API
  * RepositoriesService.GetContens()
  * RepositoriesService.CreateFile()
  * IssuesService.CreateLabel()

* I think there would be more API can crash...

gmlewis:
Can you please provide the location of the crashes?  Most likely they are caused by a `nil` de-reference, correct?

gmlewis:
This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

joshuabezaleel:
Hi @gmlewis , if it's okay I would like to ask, what should be the expected behavior of this bug fix? (because of the `nil` dereference)?

gmlewis:
Yes, this package should never panic, but instead return errors that can be handled.

joshuabezaleel:
Is it okay to use `recover` and then print the `error` message? @gmlewis

gmlewis:
No, we want to prevent the panics in the first place.

gmlewis:
Now, having said that, it is another matter entirely if the user is passing in `nil` for a required parameter (which I don't think is the case here).
That would be user error and not a bug, and does not need fixing, as we attempt to carefully document the required parameters.

So in all honesty, this issue first needs investigation to determine if the bug report is user error or really one or more bugs in the `github` package.

We may end up closing this issue as user error and document it as such.

If, however, there really are possible cases where this package panics on valid responses from GitHub, then let's fix them.

geototti21:
Hey @gmlewis , I had a look on that and I couldn't reproduce `panic` with normal or empty repository for `RepositoriesService.GetContents` or any other API. 

They only way I managed to panic it is by passing `nil` context. As you mention this is a user error but maybe its better to return an error that `context` cannot be nil than panic? I created a small PR for that if you think its good to return an error than panic https://github.com/google/go-github/pull/1348 :smile:

gmlewis:
Thank you, @geototti21!  I like it.  :smile:

PR #1348, opened 2019-12-07
Add context nil check on client Doer

geototti21:
I am adding a nil check for the provided context and return an error instead of panic.
@gmlewis raised a valid point that this is a user error but probably is better to return and error than panic.

Related to #1260


gmlewis:
Thank you, @wesleimp!
Merging.

Issue #1352, opened 2019-12-22
How to create a review comment reply?


zulhfreelancer:
From GitHub docs page, looks like we can send replies to PR comment - https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply

I tried to find the right method inside this package but I didn't find anything. I tried `CreateCommentInReplyTo` but it was unsuccessful — from the [code](https://github.com/google/go-github/blob/master/github/pulls_comments.go#L144), looks like the URL is pointed to [creating PR comment](https://developer.github.com/v3/pulls/comments/#create-a-comment) (`/repos/:owner/:repo/pulls/:pull_number/comments`) and not [PR comment reply](https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply) (`/repos/:owner/:repo/pulls/:pull_number/comments/:comment_id/replies`).

Do we have that method somewhere inside this package?

gmlewis:
Thank you for the detailed issue post, @zulhfreelancer!
You might be surprised at how much time you just saved me by providing such excellent details.

I think you are absolutely right... the endpoint you listed above is not yet supported by this client library.

Before I open up this issue and offer it for anyone else to work on, would you like to tackle a PR for this one yourself?  Or are you happy to allow other volunteers to tackle it? :smile:

zulhfreelancer:
@gmlewis I just submitted PR for you - https://github.com/google/go-github/pull/1353

zulhfreelancer:
I've closed the PR. Need to do some cleanup when I'm free later and resubmit it. Thanks for your comments.

PR #1353, opened 2019-12-23
Add create review comment reply method

zulhfreelancer:
In https://github.com/google/go-github/issues/1352, it's confirmed that we haven't implement method to [create a review comment reply](https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply). Here is my PR for it.

### DIFF
--- a/github/pulls_comments.go
+++ b/github/pulls_comments.go
@@ -22,6 +22,10 @@ type PullRequestComment struct {
 	PullRequestReviewID *int64     `json:"pull_request_review_id,omitempty"`
 	Position            *int       `json:"position,omitempty"`
 	Line                *int       `json:"line,omitempty"`
+	StartLine           *int       `json:"start_line,omitempty"`
+	OriginalStartLine   *int       `json:"original_start_line,omitempty"`
+	StartSide           *string    `json:"start_side,omitempty"`
+	OriginalLine        *int       `json:"original_line,omitempty"`
 	Side                *string    `json:"side,omitempty"`
 	OriginalPosition    *int       `json:"original_position,omitempty"`
 	CommitID            *string    `json:"commit_id,omitempty"`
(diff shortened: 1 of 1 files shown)


zulhfreelancer:
@googlebot I signed it!

github/pulls_comments.go
```diff
@@ -156,6 +156,30 @@ func (s *PullRequestsService) CreateCommentInReplyTo(ctx context.Context, owner
 	return c, resp, nil
 }
 
+// CreateReviewCommentReply creates a reply to a review comment for a pull request.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply
+func (s *PullRequestsService) CreateReviewCommentReply(ctx context.Context, owner string, repo string, prNumber int, body string, commentID int64) (*PullRequestComment, *Response, error) {
+	comment := &struct {
+		Body string `json:"body,omitempty"`
```
gmlewis: You can remove the `,omitempty` here since `body` must be supplied (and it is not a pointer to a `string` anyway so the `omitempty` would have no effect).

github/pulls_comments.go
```diff
@@ -156,6 +156,30 @@ func (s *PullRequestsService) CreateCommentInReplyTo(ctx context.Context, owner
 	return c, resp, nil
 }
 
+// CreateReviewCommentReply creates a reply to a review comment for a pull request.
+//
+// GitHub API docs: https://developer.github.com/v3/pulls/comments/#create-a-review-comment-reply
+func (s *PullRequestsService) CreateReviewCommentReply(ctx context.Context, owner string, repo string, prNumber int, body string, commentID int64) (*PullRequestComment, *Response, error) {
```
gmlewis: s/owner string, repo string/owner, repo string/

zulhfreelancer: Sorry, I didn't get this.

github/pulls_comments.go
```diff
@@ -130,6 +136,27 @@ func (s *PullRequestsService) CreateComment(ctx context.Context, owner string, r
 	return c, resp, nil
 }
 
+// CreateCommentWithComfortFade creates a new comment on the specified pull request. It has `Accept: application/vnd.github.comfort-fade-preview+json` header.
```
gmlewis: Let's please not make two versions of the same endpoint (one with the custom header and one without).
In this repo, we decided to always be on the cutting edge and support all preview APIs as best we can so that we can give early feedback to the GitHub API support team and to give clients using this library access to the latest GitHub features.

So let's please add the custom media header to the original endpoint and remove this new function.

Also, please take a look at `github/github.go` for how we add custom media headers.

Finally, please write unit tests for all functions that are affected by this PR.  Thanks.

PR #1354, opened 2019-12-23
Add support for new fields in Branch Protection

### LINKED ISSUE #1345: Feature Request > New branch protection features support
Hi,

Please add support for the new branch protection features:
* `required_linear_history`
* `allow_force_pushes`
* `allow_deletions`

See: https://developer.github.com/v3/repos/branches/#update-branch-protection

Thanks!

pcantea:
Fixes #1345

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -728,6 +728,9 @@ type Protection struct {
 	RequiredPullRequestReviews *PullRequestReviewsEnforcement `json:"required_pull_request_reviews"`
 	EnforceAdmins              *AdminEnforcement              `json:"enforce_admins"`
 	Restrictions               *BranchRestrictions            `json:"restrictions"`
+	RequiredLinearHistory      bool                           `json:"required_linear_history"`
+	AllowForcePushes           bool                           `json:"allow_force_pushes"`
+	AllowDeletions             bool                           `json:"allow_deletions"`
 }
 
 // ProtectionRequest represents a request to create/edit a branch's protection.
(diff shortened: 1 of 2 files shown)


wesleimp:
Thank you, @pcantea 
👌

github/repos.go
```diff
@@ -728,6 +728,9 @@ type Protection struct {
 	RequiredPullRequestReviews *PullRequestReviewsEnforcement `json:"required_pull_request_reviews"`
 	EnforceAdmins              *AdminEnforcement              `json:"enforce_admins"`
 	Restrictions               *BranchRestrictions            `json:"restrictions"`
+	RequiredLinearHistory      bool                           `json:"required_linear_history"`
```
gmlewis: Each of these three `bool` fields need to be structures with a single `Enabled` field, based on the GitHub v3 API developer docs found here:
https://developer.github.com/v3/repos/branches/#get-branch-protection

Here is the example output from this endpoint:

```
...
  "required_linear_history": {
    "enabled": true
  },
  "allow_force_pushes": {
    "enabled": true
  },
  "allow_deletions": {
    "enabled": true
  }
```

Since each of these structs appear to only have a single `Enabled` field, I'm fine with making them all the same struct type... something like:
```
type EnabledType struct {
  Enabled bool `json:"enabled"`
}
```

and then these three fields would have struct pointers with `omitempty`, like this:

```
RequiredLinearHistory *EnabledType `json:"required_linear_history,omitempty"`
AllowForcePushes      *EnabledType `json:"allow_force_pushes,omitempty"`
AllowDeletions        *EnabledType `json:"allow_deletions,omitempty"`
```

That way, if GitHub provides the values, great... but if not, we know that they weren't returned.

github/repos.go
```diff
@@ -736,6 +739,9 @@ type ProtectionRequest struct {
 	RequiredPullRequestReviews *PullRequestReviewsEnforcementRequest `json:"required_pull_request_reviews"`
 	EnforceAdmins              bool                                  `json:"enforce_admins"`
 	Restrictions               *BranchRestrictionsRequest            `json:"restrictions"`
+	RequiredLinearHistory      bool                                  `json:"required_linear_history"`
```
gmlewis: Unfortunately, this won't work either.  Because these are optional fields, these three need to be type `*bool` and not `bool`.
They also need the `,omiteempty` flag on them

Issue #1267, opened 2019-08-30
Not possible to update a child team to be top-level


ashlineldridge:
Due to the `omitempty` tag on `NewTeam.ParentTeamID` it is not possible (in my understanding) to update a child team to be a top-level team. Because updating a team is a PATCH operation, when `parent_team_id` is omitted, GitHub does not update the team's parent.

Happy to submit a PR.

gmlewis:
For others listening in, I believe we are talking about this:
GitHub Developer v3 API docs: https://developer.github.com/v3/teams/#edit-team
and this:
https://github.com/gmlewis/go-github/blob/master/github/teams.go#L196-L216

So I'm not sure I understand the issue that you are raising.  `omitempty` actually gives us more flexibility... if you don't provide an `*int64`, then the parameter won't be passed. But if you do indeed set the parameter, it will be passed... so I don't see what the problem is.  Maybe if you give a more detailed example, I could possibly help.

But, having said that, I'm about to be out-of-town for 8 days with no internet... just FYI, I won't be responding for a while. Hopefully someone else can help you out.

@willnorris & @gauntface - just FYI I'm out-of-town for about a week, in case any urgent issues arise. Thanks.

ashlineldridge:
> Maybe if you give a more detailed example, I could possibly help.

Sure, what I'm seeing is that because [`ParentTeamID`](https://github.com/google/go-github/blob/master/github/teams.go#L149) uses an `omitempty` tag, the JSON field `parent_team_id` is not included in the `PATCH` operation if `ParentTeamID` is `nil`. As such, there is no way to tell GitHub to update a team to longer have a parent (i.e. to become a top-level team). To tell GitHub to update a team to be top-level you minimally need:

```json
{
  "name": "foo",
  "parent_team_id" null
}
```

But due to `omitempty` you get:

```json
{
  "name": "foo"
}
```

Which produces a no-op.

gmlewis:
Sorry for the delay, @ashlineldridge, and thank you for the detailed explanation and example!

Yes, I understand now. Since this particular method still needs that parameter to be optional, it sounds like the correct solution to this problem is to create a new method that specifically updates the team to be one with no parent.

Now the challenge becomes "What do we name the new method"?
I'm leaning toward `EditTeamRemoveParent`, but am option to suggestions.
This new method would be identical to `EditTeam` except that its `ParentTeamID` field would be set to `nil` and `omitempty` would be removed.

[...]

PR #1356, opened 2019-12-28
fix: Remove parent for a child team

yashLadha:
This PR conatins following changes:

* new method `EditTeamRemoveParent`

Fixes: #1267

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -7,7 +7,9 @@ package github
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"strings"
 	"time"
 )
@@ -187,9 +189,28 @@ func (s *TeamsService) CreateTeam(ctx context.Context, org string, team NewTeam)
 // EditTeam edits a team.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/#edit-team
-func (s *TeamsService) EditTeam(ctx context.Context, id int64, team NewTeam) (*Team, *Response, error) {
+func (s *TeamsService) EditTeam(ctx context.Context, id int64, team NewTeam, removeParent bool) (*Team, *Response, error) {
 	u := fmt.Sprintf("teams/%v", id)
-	req, err := s.client.NewRequest("PATCH", u, team)
+
+	buf, err := s.client.GetJSONEncodedData(team)
+	if err != nil {
+		return nil, nil, err
+	}
+	body, err := ioutil.ReadAll(buf)
+	if err != nil {
+		return nil, nil, err
+	}
+	var dumpTeam map[string]interface{}
+	err = json.Unmarshal(body, &dumpTeam)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if removeParent == true {
+		dumpTeam["parent_team_id"] = nil
+	}
+
+	req, err := s.client.NewRequest("PATCH", u, dumpTeam)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/github/teams.go
+++ b/github/teams.go
@@ -203,6 +203,52 @@ func (s *TeamsService) EditTeam(ctx context.Context, id int64, team NewTeam) (*T
 	return t, resp, nil
 }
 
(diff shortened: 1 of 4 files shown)


yashLadha:
@gmlewis Could you please review. The approach that I have followed is, instead of creating a new struct for a single case or using a generic interface. I have created the struct for this task itself and the method which can use this.

github/teams.go
```diff
@@ -203,6 +203,52 @@ func (s *TeamsService) EditTeam(ctx context.Context, id int64, team NewTeam) (*T
 	return t, resp, nil
 }
 
+// RemoveParentTeam represents a new team with no parent
+type RemoveParentTeam struct {
+	Name         string   `json:"name"` // Name of the team. (Required.)
+	Description  *string  `json:"description,omitempty"`
+	Maintainers  []string `json:"maintainers,omitempty"`
+	RepoNames    []string `json:"repo_names,omitempty"`
+	ParentTeamID *int64   `json:"parent_team_id,omitempty"`
```
gmlewis: I'm sorry, but I don't understand.

How is this struct different from the `NewTeam` struct on line 138 of teams.go?

Have you tested this out on GitHub and does it work?

Additionally, it seems like a lot of code duplication (especially as it is currently written).

It seems like an argument could be added (maybe `removeParent bool`) to the existing `EditTeam` call, and then the resulting JSON body could simply add in the field: `"parent_team_id": null`, right? 

Wouldn't that be much easier to maintain than copy/pasting all the surrounding code and tweaking one line?

github/teams.go
```diff
@@ -203,6 +203,52 @@ func (s *TeamsService) EditTeam(ctx context.Context, id int64, team NewTeam) (*T
 	return t, resp, nil
 }
 
+// RemoveParentTeam represents a new team with no parent
```
gmlewis: Comments should be complete sentences ended in a period for the auto-generated godocs.

github/teams.go
yashLadha: If we set the parent_team_id to `nil`, due to the `omitempty` it will be [removed](https://play.golang.org/p/Z8WhKO_Q_LK). Changing the buffer in parsing defies the logic for that method. Instead of converting changing the buffer, we can parse it to a different struct and pass on that to parse json in `EditTeam` function.

github/teams.go
yashLadha: Removing duplicate function

gmlewis:
Based on your reply, I'm not sure I communicated clearly in my code review.
I've written up #1357 to show what I meant. Feel free to comment.

yashLadha:
@gmlewis I have created a generic interface to add the key, due to this we don't have to create a new function and implement the functionality in the same. As type checking will be already done in `team` object so validations can be avoided in this generic interface for the sake of sending `parent_team_id`

gmlewis:
Hmm... I don't particularly like this approach, for the following reasons:

* It alters a highly-used section of code (`NewRequest`) which reduces the clarity of that method.
* We already have the `NewTeam` struct and it seems like a hack to first marshal it to JSON, then unmarshal it, override one of its fields, then marshal it again with a generic `map[string]interface{}`.
* It is less efficient than #1357.
* I think this solution is not as clear and is going to be more difficult to maintain.

I'm open to discussion.
Others are also welcome to comment, of course.
If you have an opinion, please comment.

yashLadha:
Yaa, I got it. I forgot to remove the `omitempty` in my initial commit. Else the PR is identical (first two commits). Also for alteration related to the `NewRequest` method, I think I just refactored to a function where the meaning is not altered and also clearly states the use of the block of code.  For your PR, aren't we basically doing the same thing which can be achieved by `map[string]interface{}` as the validation are same except for the `parent_team_id` and we can avoid the unnecessary steps for the internal validations present in unmarshalling.  Now the performance hit can only be occured with the considering two steps for marshalling. I hope i have communicated my thought. If any way we can go with your PR and can close this one.

gmlewis:
OK, thanks, @yashLadha.

PR #1357, opened 2019-12-29
Add removeParent to EditTeam

### LINKED ISSUE #1267: Not possible to update a child team to be top-level

gmlewis:
Fixes #1267.

I wrote this PR to communicate my thoughts about one way to fix #1267 since I'm not sure I clearly communicated this idea when reviewing #1356.

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -185,6 +185,33 @@ func (s *TeamsService) CreateTeam(ctx context.Context, org string, team NewTeam)
 	return t, resp, nil
 }
 
+// newTeamNoParent is the same as NewTeam but ensures that the
+// "parent_team_id" field will be null. It is for internal use
+// only and should not be exported.
+type newTeamNoParent struct {
+	Name         string   `json:"name"`
+	Description  *string  `json:"description,omitempty"`
+	Maintainers  []string `json:"maintainers,omitempty"`
+	RepoNames    []string `json:"repo_names,omitempty"`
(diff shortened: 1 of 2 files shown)


gmlewis:
Let's go ahead and have this PR reviewed as a possible fix for #1267.

github/teams.go
```diff
@@ -187,9 +188,32 @@ func (s *TeamsService) CreateTeam(ctx context.Context, org string, team NewTeam)
...
+			LDAPDN       *string  `json:"ldap_dn,omitempty"`
+		}{
+			Name:        team.Name,
+			Description: team.Description,
+			Maintainers: team.Maintainers,
+			RepoNames:   team.RepoNames,
+			Privacy:     team.Privacy,
+			LDAPDN:      team.LDAPDN,
+		}
+		req, err = s.client.NewRequest("PATCH", u, teamRemoveParent)
```
wesleimp: I'm not sure if it is the place for this code snippet. Maybe you can extract it to another function.

gmlewis: I've moved it.  PTAL.

gmlewis:
Thank you, @wesleimp !
Merging.

Issue #1331, opened 2019-11-09
Remove deprecated authentication API endpoints and add replacements


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

waseem18:
@gmlewis I'd like to take this up.

gmlewis:
OK, thanks @waseem18 - it is yours.

mubeenflora:
@waseem18  Are you working on it ? i just to know because i would like to do this

waseem18:
@mubeenflora Go ahead, I've started work on it but didn't get a chance to complete it.

mubeenflora:
thanks but it'll take time because i have started my new project

pavan541cs:
@mubeenflora Are you actively working on this ? Else I would love to work on this

yashLadha:
Is this issue still up for grabs? @gmlewis

gmlewis:
With no response from @mubeenflora in two days, I'm opening this issue back up.

pavan541cs:
@gmlewis I am picking this one and start working on this. Thank you for the update

gmlewis:
Thank you, @pavan541cs! It is yours.

pavan541cs:
@gmlewis #1360 Can you check this PR and tell me if this looks good so that I can make all the changes. I have done only for one endpoint

gmlewis:
In another PR, I think it helped (me, at least) to list the things that need to be done to complete an issue. So to complete this issue...

The following endpoints need to be removed:

- [ ] `GET /authorizations`
- [ ] `GET /authorizations/:authorization_id`
- [ ] `POST /authorizations`
- [ ] `PUT /authorizations/clients/:client_id`
- [ ] `PUT /authorizations/clients/:client_id/:fingerprint`
- [ ] `PATCH /authorizations/:authorization_id`
- [ ] `DELETE /authorizations/:authorization_id`
- [ ] `GET /applications/grants`
- [ ] `GET /applications/grants/:grant_id`
- [ ] `DELETE /applications/grants/:grant_id`
[...]

PR #1360, opened 2019-12-31
Changes to align with changes in git api for authoriztion endpoints

pavan541cs:
This PR is about #1331 . I have changed for one git api endpoint . If this looks good I can go ahead and make all the changes


github/authorizations.go
```diff
@@ -274,14 +274,18 @@ func (s *AuthorizationsService) Delete(ctx context.Context, id int64) (*Response
...
 
-	req, err := s.client.NewRequest("GET", u, nil)
+	req, err := s.client.NewRequest("POST", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	q := req.URL.Query()
+	q.Add("access_token", accessToken)
+	req.URL.RawQuery = q.Encode()
```
gmlewis: I don't think these 3 lines are correct.

If I'm reading this page correctly: https://developer.github.com/v3/apps/oauth_applications/#check-a-token
it is saying that the access token is being passed in the body of the request, not as a query parameter.

This means that instead of sending here, line 280 above would have to pass a struct, something like this:

```
u := fmt.Sprintf("applications/%v/token", clientID)
reqBody := &struct{
  AccessToken string `json:"access_token"`
}{ AccessToken: accessToken }
req, err := s.client.NewRequest("POST", u, reqBody)
```

pavan541cs:
@googlebot I signed it!

gmlewis:
Thanks, @pavan541cs - it looks like the unit tests are still failing because the calls to `Check` don't have the correct number of parameters.

github/authorizations.go
```diff
@@ -274,10 +274,14 @@ func (s *AuthorizationsService) Delete(ctx context.Context, id int64) (*Response
 // The returned Authorization.User field will be populated.
 //
 // GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#check-an-authorization
```
gmlewis: This URL should now be: https://developer.github.com/v3/apps/oauth_applications/#check-a-token

Also, the comment starting above at line 270 is no longer valid. Please update it.

github/authorizations.go
```diff
@@ -274,10 +274,14 @@ func (s *AuthorizationsService) Delete(ctx context.Context, id int64) (*Response
...
-func (s *AuthorizationsService) Check(ctx context.Context, clientID string, token string) (*Authorization, *Response, error) {
-	u := fmt.Sprintf("applications/%v/tokens/%v", clientID, token)
+func (s *AuthorizationsService) Check(ctx context.Context, clientID string, clientToken string, accessToken string) (*Authorization, *Response, error) {
```
gmlewis: s/clientID string, clientToken string, accessToken string/clientID, clientToken, accessToken string/

Please also add a comment about the difference between `clientToken` and `accessToken` and how to know which is for what purpose.

github/authorizations.go
```diff
@@ -302,10 +306,14 @@ func (s *AuthorizationsService) Check(ctx context.Context, clientID string, toke
 // The returned Authorization.User field will be populated.
 //
 // GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#reset-an-authorization
```
gmlewis: This URL should now be: https://developer.github.com/v3/apps/oauth_applications/#reset-a-token

Also, the comment starting above around line 302 is no longer valid. Please update it.

pavan541cs:
@gmlewis I changed the remaining endpoints also. Should I completely remove these endpoints 
`GET /applications/grants 
GET /applications/grants/:grant_id`. These are also getting deprecated and there is no replacement.

github/authorizations.go
```diff
@@ -302,10 +306,14 @@ func (s *AuthorizationsService) Check(ctx context.Context, clientID string, toke
...
-func (s *AuthorizationsService) Reset(ctx context.Context, clientID string, token string) (*Authorization, *Response, error) {
-	u := fmt.Sprintf("applications/%v/tokens/%v", clientID, token)
+func (s *AuthorizationsService) Reset(ctx context.Context, clientID string, token string, accessToken string) (*Authorization, *Response, error) {
```
gmlewis: s/clientID string, token string, accessToken string/clientID, token, accessToken string/

Please also add a comment about the difference between `token` and `accessToken` and how to know which is for what purpose.

gmlewis:
> @gmlewis I changed the remaining endpoints also. Should I completely remove these endpoints
> `GET /applications/grants GET /applications/grants/:grant_id`. These are also getting deprecated and there is no replacement.

Yes, please go ahead and completely remove the deprecated endpoints in this PR. We can then mark #1331 as completely fixed.  Thank you, @pavan541cs!

pavan541cs:
@gmlewis You can review now. I have removed deprecated things. Please have a look on the PR now.

github/authorizations.go
```diff
@@ -273,8 +142,8 @@ func (s *AuthorizationsService) Delete(ctx context.Context, id int64) (*Response
 //
 // The returned Authorization.User field will be populated.
 //
-// GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#check-an-authorization
-func (s *AuthorizationsService) Check(ctx context.Context, clientID string, clientToken string, accessToken string) (*Authorization, *Response, error) {
+// GitHub API docs: https://developer.github.com/v3/apps/oauth_applications/#check-a-token
+func (s *AuthorizationsService) Check(ctx context.Context, clientID string, accessToken string) (*Authorization, *Response, error) {
```
gmlewis: s/clientID string, accessToken string/clientID, accessToken string/

github/authorizations.go
```diff
@@ -305,8 +174,8 @@ func (s *AuthorizationsService) Check(ctx context.Context, clientID string, clie
 //
 // The returned Authorization.User field will be populated.
 //
-// GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#reset-an-authorization
-func (s *AuthorizationsService) Reset(ctx context.Context, clientID string, token string, accessToken string) (*Authorization, *Response, error) {
+// GitHub API docs: https://developer.github.com/v3/apps/oauth_applications/#delete-an-app-token
+func (s *AuthorizationsService) Reset(ctx context.Context, clientID string, accessToken string) (*Authorization, *Response, error) {
```
gmlewis: s/clientID string, accessToken string/clientID, accessToken string/

github/authorizations.go
```diff
@@ -334,7 +203,7 @@ func (s *AuthorizationsService) Reset(ctx context.Context, clientID string, toke
 // clientSecret. Invalid tokens will return a 404 Not Found.
 //
 // GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#revoke-an-authorization-for-an-application
-func (s *AuthorizationsService) Revoke(ctx context.Context, clientID string, token string, accessToken string) (*Response, error) {
+func (s *AuthorizationsService) Revoke(ctx context.Context, clientID string, accessToken string) (*Response, error) {
```
gmlewis: s/clientID string, accessToken string/clientID, accessToken string/

github/authorizations.go
```diff
@@ -349,59 +218,19 @@ func (s *AuthorizationsService) Revoke(ctx context.Context, clientID string, tok
...
-
 // DeleteGrant deletes an OAuth application grant. Deleting an application's
 // grant will also delete all OAuth tokens associated with the application for
 // the user.
 //
 // GitHub API docs: https://developer.github.com/v3/oauth_authorizations/#delete-a-grant
-func (s *AuthorizationsService) DeleteGrant(ctx context.Context, id int64) (*Response, error) {
-	u := fmt.Sprintf("applications/grants/%d", id)
-	req, err := s.client.NewRequest("DELETE", u, nil)
+func (s *AuthorizationsService) DeleteGrant(ctx context.Context, clientID string, accessToken string) (*Response, error) {
```
gmlewis: s/clientID string, accessToken string/clientID, accessToken string/

test/integration/authorizations_test.go
```diff
@@ -27,169 +27,60 @@ const envKeyClientSecret = "GITHUB_CLIENT_SECRET"
...
-	failIfNotStatusCode(t, resp, 200)
-
-	// Verify that the IDs are the same
-	if *createdAuth.ID != *secondAuth.ID {
-		t.Fatalf("The ID of the second returned auth should be the same as the first. Expected [%v], got [%v]", createdAuth.ID, secondAuth.ID)
-	}
-
 	// Verify the token
-	appAuth, resp, err := appAuthenticatedClient.Authorizations.Check(context.Background(), clientID, *createdAuth.Token, accessToken)
+	_, resp, err := appAuthenticatedClient.Authorizations.Check(context.Background(), clientID, accessToken)
```
gmlewis: Why are all the "Quick sanity checks" being removed from the integrations tests?

pavan541cs: `GetOrCreateForApp`. This function uses deprecated git api `PUT /authorizations/clients/:client_id/:fingerprint`. So had to remove everything which is dependent on it.

gmlewis: My point is that `Authorizations.Check` still exists even though it is using a new API... so this integration should now use the new returned values and perform an equivalent "Quick sanity check" like it was doing before. Isn't that possible?

gmlewis:
Closing stalled PR in favor of #1461.

Issue #1136, opened 2019-03-27
Incorrect error-parsing on failure of creating review


wataash:
Original error info from GitHub API (formatted for readability):

```json
$ curl -H "Authorization: token XXXXXXXX" -X POST -d '{"commit_id": "ffffffffff"}' https://api.github.com/repos/wataash/sandbox/pulls/1/reviews
{
  "message": "Unprocessable Entity",
  "errors": [
    "Variable commitOID of type GitObjectID was provided invalid value"
  ],
  "documentation_url": "https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review"
}
```

indicates the commit id `ffffffffff` doesn't exist.

Same one via go-github:

```go
ctx := context.Background()
ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: "XXXXXXXX"})
tc := oauth2.NewClient(ctx, ts)
client := github.NewClient(tc)
nonExistCommitID := "ffffffffff"
review := github.PullRequestReviewRequest{CommitID: &nonExistCommitID}
_, _, err := client.PullRequests.CreateReview(ctx, "wataash", "sandbox", 1, &review)
_, _ = fmt.Printf("%+v\n", err)
```

prints:

```
POST https://api.github.com/repos/wataash/sandbox/pulls/1/reviews: 422 Unprocessable Entity [{Resource: Field: Code: Message:}]
```

which lacks the error details but shows unrelated fields.

wataash:
Most of the non-200 response formats are seem to be undocumented, so I guess some other APIs have this problem.
Making response body available as `string` would be helpful for go-github users when they faced incorrectly parsed errors.
I think performance issue (as discussed in #1047) is negligible if the stringifying is done only on non-200.

gmlewis:
Thank you for the side-by-side comparison, @wataash, and the reference to #1047 where we discussed returning the body for *every* response (which we rejected).

I'm not against returning the full body response for non-200 responses.

Thoughts, @gauntface, @willnorris, or anyone else?

willnorris:
@wataash: some additional data can be fetched by casting the err to an [ErrorResponse](https://godoc.org/github.com/google/go-github/github#ErrorResponse).

``` go
_, _, err := client.PullRequests.CreateReview(ctx, "wataash", "sandbox", 1, &review)
if er, ok := err.(github.ErrorResponse); ok {
    _, _ = fmt.Printf("%#v\n", er)
}
```

[...]

Issue #540, opened 2017-02-02
Pull Request Reviews (Preview API) methods have unusual error format that is not parsed.


dmitshur:
[Pull Request Reviews (Preview API)](https://developer.github.com/changes/2016-12-14-reviews-api/) methods were tracked in #495 and implemented in #497. They currently seem to use an unusual format for errors, and go-github does not parse that format.

Originally discovered by @haya14busa during review of #497:

> As I posted request and error response example (https://github.com/google/go-github/pull/497#discussion_r92933362), error format for review API is not same as http://developer.github.com/v3/#client-errors
> 
> "errors" is list of string for review API.
> 
> ```json
> {
>   "message": "Validation Failed",
>   "errors": [
>     "Cannot return null for non-nullable field PullRequestReviewComment.position"
>   ],
>   "documentation_url": "https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments"
> }
> ```
> 
> Corresponding go-github struct (`ErrorResponse`)  is here. https://github.com/google/go-github/blob/fbc15caccf54600900355e302a0fafabf7e2d733/github/github.go#L463
> 
> Because `ErrorResponse` doesn't support review API error format, error information is dropped with unmarshalling and it's really difficult to debug error using go-github.
> (I tried to implement create review API in my personal project https://github.com/haya14busa/reviewdog/pull/63 using go-github, but it was really hard to debug error... )
> 
> I'll report this problem to GitHub support, but this review API error fromat could be intentional.
> If so, we have to handle this problem I guess.

See https://github.com/google/go-github/pull/497#issuecomment-267810912 for full context and details.

This issue is to track that problem (which will soon be merged into `master`) and resolve it afterwards. The unusual error format is not documented, and we're waiting on GitHub to potentially fix the error format for those endpoints, or update documentation.

dmitris:
@dmitshur - have you heard anything back from the GitHub support on the issue (since it is marked "wairing for GitHub")?

dmitris:
https://godoc.org/github.com/google/go-github/github#PullRequestsService.ListReviewComments contains a link to this issue:

> TODO: Follow up with GitHub support about an issue with this method's returned error format and remove this comment once it's fixed. Read more about it here - https://github.com/google/go-github/issues/540

dmitshur:
It was @haya14busa that emailed GitHub, not me. See https://github.com/google/go-github/pull/497#issuecomment-270400412.

[...]

PR #1364, opened 2020-01-03
Support unexpected error response format

haya14busa:
> "errors" is list of string for review API.
```
{
  "message": "Validation Failed",
  "errors": [
    "Cannot return null for non-nullable field PullRequestReviewComment.position"
  ],
  "documentation_url": "https://developer.github.com/v3/pulls/reviews/#get-a-single-reviews-comments"
}
```

See #540 for the detail of the problem.
Related: #1136

I assume GitHub won't change the error structure as it's already in production.
I hope they will document the error format, but anyway go-github needs to support it for the sane error handling,
Even if #1136 is implemented, it should be better to represent error as ErrorResponse struct, IMO.
so I implement it by re-using `message` field as a workaround.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -706,6 +706,10 @@ These are the possible validation error codes:
         some resources return this (e.g. github.User.CreateKey()), additional
         information is set in the Message field of the Error
 
+GitHub error responses structure are often undocumented and inconsistent.
+Sometimes error is just a simple string (Issue #540).
+In such cases, Message represents an error message as a workaround.
+
 GitHub API docs: https://developer.github.com/v3/#client-errors
 */
 type Error struct {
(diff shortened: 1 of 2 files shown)


github/github.go
```diff
@@ -706,6 +706,10 @@ These are the possible validation error codes:
         some resources return this (e.g. github.User.CreateKey()), additional
         information is set in the Message field of the Error
 
+GitHub error responses structure are often undocumented and inconsistent.
+Sometimes error is just a simple string (Issue #540).
+In such cases, Message represents an error message as as a workaround.
```
gmlewis: s/as as/as/

haya14busa: Thank you for the catch! Fixed.

gmlewis:
Thank you, @haya14busa and @wesleimp!
Merging.

PR #1366, opened 2020-01-08
Revert "Bump major version to v29"

gmlewis:
Reverts google/go-github#1344 due to breaking the build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ If you're interested in using the [GraphQL API v4][], the recommended library is
 ## Usage ##
 
 ```go
-import "github.com/google/go-github/v29/github"	// with go modules enabled (GO111MODULE=on or outside GOPATH)
+import "github.com/google/go-github/v28/github"	// with go modules enabled (GO111MODULE=on or outside GOPATH)
 import "github.com/google/go-github/github" // with go modules disabled
 ```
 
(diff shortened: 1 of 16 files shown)


gmlewis:
Merging to fix build.  I'll address v29 later tonight.  Sorry for all the noise.  :-(

dmitshur:
No problem @gmlewis, thank you for taking care of this.

I believe the issue in PR #1344 was that it did not update the [recently added](https://github.com/google/go-github/pull/1285) `example/topics/main.go` file, it still imports "github.com/google/go-github/v28/github" instead of "github.com/google/go-github/v29/github": https://github.com/google/go-github/blob/v29.0.0/example/topics/main.go#L15.

After we fix this, we'll need to pick a new version number, because v29.0.0 has already been published. Perhaps v29.0.1, since this will be fixing up a minor problem.

gmlewis:
Ah, thank you for the analysis, @dmitshur !
I'm going to try and remedy it now.
(I deleted the `v29.0.0` if that makes any difference.)

dmitshur:
> (I deleted the `v29.0.0` if that makes any difference.)

That version has already been [indexed](https://index.golang.org/index?since=2020-01-08T22:03:01.935587Z) and [added](https://sum.golang.org/lookup/github.com/google/go-github/v29@v29.0.0) to the [Go checksum database](https://sum.golang.org/), so we still won't be able to publish the same module version with different content without it causing a mismatch in the hash.

Issue #1365, opened 2020-01-08
New release?


haya14busa:
related: #1280 

It seems go-github didn't release new version since Sep 8, 2019 and there are [about 50 commits](https://github.com/google/go-github/compare/v28.1.1...master) since the last release.
Can you release new version? 

Thanks in advance!

gmlewis:
Yes, I would like to get approval from @willnorris for #1344 and merge that first, as that would be the first major release since the discussion in #1280.

Maybe we should merge master into #1344 before merging to make a single large release?  I'll comment on the PR.

haya14busa:
Thank you for the quick response! 

> Maybe we should merge master into #1344 before merging to make a single large release?

Were there any breaking changes? If so, go-github must bump major version first. If no, I think go-github should just bump minor tag.

p.s. @gmlewis Thank you for maintaining the project :heart:

gmlewis:
Yes, there have been many breaking changes.  #1344 is bumping the major version for the breaking changes up to that point... and if we decide to do another release immediately after that (without first merging master into #1344) we would need to bump the major version again with more breaking changes.

For a while there, I was trying to be strict about semver releases after every merge (major, minor, patch), but the volume just got to be too much.

Thank you for the kind words, @haya14busa!

fleskesvor:
A new version would be great, but if you can't wait for it to be released, you can get features from the head of master into a project that uses Go modules with `go get -v github.com/google/go-github/v28@master` (which I only recently learned).

gmlewis:
We just released `v29.0.0`.
Closing.

gmlewis:
Ugh, I broke the build... trying to figure out how to roll it back...

PR #1367, opened 2020-01-09
Bump major version to v29 (take 2)

gmlewis:
This is the second attempt to bump the major version to v29 after #1344 failed.
Release `v29.0.0` was broken so this will be released as `v29.0.1`.


gmlewis:
@dmitshur - if you could please check this when you get a chance, I would greatly appreciate it!
I will release this as `v29.0.1` if there are no objections.

gmlewis:
Fixes #1365.

gmlewis:
Thank you, @dmitshur!
Merging.

PR #1283, opened 2019-09-13
Added support for allowing apps to push to protected branches

shibasisp:
fixes #1275

PR #1370, opened 2020-01-09
Fix JSON unmarshal error of App.app.events

haya14busa:
I got following error.
json: cannot unmarshal string into Go struct field App.app.events of
type github.Event

It's regression from #1283 and event type should be actually string.

https://developer.github.com/v3/apps/#get-a-single-github-app
https://developer.github.com/v3/repos/branches/#list-apps-with-access-to-protected-branch


haya14busa:
@gmlewis Can you release new patch version (v29.0.2)? v29 is broken unless including this fix.

gmlewis:
Yes indeed.  Releasing 29.0.2 now.

haya14busa:
Thanks! :+1:

PR #1372, opened 2020-01-10
Implement file deletion for CreateTree

### LINKED ISSUE #1371: The Tree API does not allow for file deletion
In order to delete a file via the  [Github REST API v3](https://developer.github.com/v3/git/trees/#create-a-tree) a file's sha need to be null as per the documentation

> The SHA1 checksum ID of the object in the tree. Also called tree.sha. If the value is null then the file will be deleted. 

The library's implementation defines the `sha` field as a pointer to a string with  that will be ignored when json serialized:

```
type TreeEntry struct {
	SHA     *string 'json:"sha,omitempty"'
         ................
}
```

When submitting a `POST` request to the `repos/%v/%v/git/trees` API with no sha and no content validation fails with response `422 Must supply tree.sha or tree.content []`

Please advise how a file should be deleted using the library

gmlewis:
Fixes #1371.

### DIFF
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -44,6 +44,20 @@ func (t TreeEntry) String() string {
 	return Stringify(t)
 }
 
+// treeEntryWithFileDelete is used internally to delete a file whose
+// Content and SHA fields are empty. It does this by removing the "omitempty"
+// tag modifier on the SHA field which causes the GitHub API to receive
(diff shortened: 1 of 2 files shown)


gmlewis:
Awaiting code review before merging.

gmlewis:
Thank you, @wesleimp !
Merging.

Issue #1378, opened 2020-01-15
DownloadReleaseAsset should follow the redirect and download the resource


andersjanmyr:
Hi,

I think the `DownloadReleaseAsset` method should follow the redirect by default and to make it optional to return the redirectURL.

I read the discussion in https://github.com/google/go-github/pull/312, and I understand that a new `http.Client` needs to be used, but I don't understand why the api had to change.

It seems to me that most users of this method don't care where the file comes from and they will all have to write similar logic to perform the download after the redirectURL is returned.

What do you think? I would be happy to contribute the code.

gmlewis:
Hmmm... I think there was a good reason... ah, yes... I believe it was because our http.Client could not easily be duplicated.

When using this repo, clients can set up different http.Clients that have different features... timeouts, proxies, etc. There wasn't a clean way to duplicate the http.Client with identical features.

However, having said that, I know that we modified another endpoint recently to add the option to follow the redirect. Yes, here it is:

https://github.com/gmlewis/go-github/blob/master/github/repos_contents.go#L243

So if you would like to put together a PR that adds a `followRedirects bool` option, then I think that would be a nice enhancement, and yet it will not force it on everybody for the cases where a special http.Client must be used (and a general redirect might fail).

PR #1381, opened 2020-01-16
Allow DownloadReleaseAsset to follow redirects

andersjanmyr:
Adds an extra parameter to DownloadReleaseAsset `followRedirects`. This will tell the client to follow one level of redirects when downloading the asset.

Solves #1378

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -288,9 +288,9 @@ func (s *RepositoriesService) downloadReleaseAssetFromURL(ctx context.Context, u
 		loc = req.URL.String()
 		return errors.New("disable redirect")
 	}
-	defer func() { s.client.client.CheckRedirect = saveRedirect }()
 	req = withContext(ctx, req)
 	resp, err := s.client.client.Do(req)
+	s.client.client.CheckRedirect = saveRedirect
 	s.client.clientMu.Unlock()
 	if err != nil {
 		if !strings.Contains(err.Error(), "disable redirect") {
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -266,10 +266,13 @@ func (s *RepositoriesService) GetReleaseAsset(ctx context.Context, owner, repo s
 // If a redirect is returned, the redirect URL will be returned as a string instead
 // of the io.ReadCloser. Exactly one of rc and redirectURL will be zero.
 //
-// If followRedirects is true, a single redirect will be followed.
+// followRedirectsClient can be passed to download the asset from a redirected
+// location. Passing http.DefaultClient is recommended unless special circumstances
+// exists, but it's possible to pass any http.Client. If nil is passed the
+// redirectURL will be returned instead.
 //
 // GitHub API docs: https://developer.github.com/v3/repos/releases/#get-a-single-release-asset
-func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, repo string, id int64, followRedirects bool) (rc io.ReadCloser, redirectURL string, err error) {
+func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, repo string, id int64, followRedirectsClient *http.Client) (rc io.ReadCloser, redirectURL string, err error) {
 	u := fmt.Sprintf("repos/%s/%s/releases/assets/%d", owner, repo, id)
 
 	req, err := s.client.NewRequest("GET", u, nil)
@@ -295,8 +298,8 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
 		if !strings.Contains(err.Error(), "disable redirect") {
 			return nil, "", err
 		}
-		if followRedirects {
-			rc, err := s.downloadReleaseAssetFromURL(ctx, loc)
+		if followRedirectsClient != nil {
+			rc, err := s.downloadReleaseAssetFromURL(ctx, followRedirectsClient, loc)
 			return rc, "", err
 		}
 		return nil, loc, nil // Intentionally return no error with valid redirect URL.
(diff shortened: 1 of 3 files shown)


github/repos_releases.go
```diff
@@ -266,33 +266,40 @@ func (s *RepositoriesService) GetReleaseAsset(ctx context.Context, owner, repo s
...
 	saveRedirect := s.client.client.CheckRedirect
 	s.client.client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
 		loc = req.URL.String()
 		return errors.New("disable redirect")
 	}
 	defer func() { s.client.client.CheckRedirect = saveRedirect }()
-
 	req = withContext(ctx, req)
 	resp, err := s.client.client.Do(req)
+	s.client.clientMu.Unlock()
```
gmlewis: This mutex is no longer protecting `s.client.client.CheckRedirect` in line 291 such that this function is no longer re-entrant.

By this I mean that it is no longer safe to call this method from multiple threads, as the lock will be released here on line 294 in one thread, the next thread will get the lock and modify `s.client.client.CheckRedirect`, then the first thread will stomp all over `s.client.client.CheckRedirect` again in line 291 when it exits.

As it was written before, the first thread completely grabs the lock, and the defer on line 291 restores the protected variable, then the defer on line 280 is executed *after* that to release the lock, and finally the second thread can come in and do its thing (in the same manner).

I think we are going to have to come up with a different solution here, as I don't think we can use this one.

andersjanmyr: @gmlewis I completely missed the `defer`. 
I added a new commit to restore `s.client.client.CheckRedirect` directly before `clientMu.Unlock()`.
 I assume that the CheckRedirect has returned when `s.client.client.Do(req)` has returned.
Will this not work? Am I missing something.

gmlewis: Ah yes, very nice.  I believe this will work, @andersjanmyr.
Can you please try it out and verify it is working as desired?

github/repos_releases_test.go
```diff
@@ -286,6 +286,35 @@ func TestRepositoriesService_DownloadReleaseAsset_Redirect(t *testing.T) {
...
+		w.Header().Set("Content-Type", "application/octet-stream")
+		w.Header().Set("Content-Disposition", "attachment; filename=hello-world.txt")
+		fmt.Fprint(w, "Hello World")
+	})
+
+	reader, _, err := client.Repositories.DownloadReleaseAsset(context.Background(), "o", "r", 1, true)
+	content, err := ioutil.ReadAll(reader)
+	if err != nil {
+		t.Errorf("Repositories.DownloadReleaseAsset returned error: %v", err)
+	}
```
gmlewis: nit: since this is a client, could you please close `reader` here after reading it, just to document as a good example of how a client should behave?

github/repos_releases.go
andersjanmyr: @gmlewis It's not working! Please hold off on the merge. I'll come back with a working version soon.

gmlewis: No problem, @andersjanmyr!

Thank you very much for catching this before we check it in!  Your effort is greatly appreciated!

andersjanmyr:
@gmlewis It's working now. I had to use a new `http.Client`. I also added a new integration test to verify that it's working. It's getting an asset from one of my repos `andersjanmyr/goose`. I don't know if you want this external dependency, but `go-github` doesn't contain any assets to test with.

github/repos_releases.go
```diff
@@ -304,6 +310,25 @@ func (s *RepositoriesService) DownloadReleaseAsset(ctx context.Context, owner, r
...
 }
 
+func (s *RepositoriesService) downloadReleaseAssetFromURL(ctx context.Context, url string) (rc io.ReadCloser, err error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req = withContext(ctx, req)
+	req.Header.Set("Accept", "*/*")
+	httpClient := &http.Client{}
```
gmlewis: Thank you, @andersjanmyr - this is why we didn't do this originally... because this will not work for all users of this client library (for the reasons I explained earlier).

I like this solution... though I wondering if we could perform a simple change to make it work for everybody?

I'm thinking that... instead of passing `followRedirects bool`, what if we were to pass: `followRedirectsClient *http.Client`?

If `followRedirectsClient` is nil, no redirect following is attempted and the `redirectURL` string is returned as before.

If, however, `followRedirectsClient` is non-nil, it will call your new function and pass in the client!

Now, people will be able to build whatever wacky `http.Client` they need, pass it in, and let your code do its job.
If they don't need anything special, they could just pass in `http.DefaultClient`.

In fact, the documentation should state this particular case as an example for the majority of the users.
However, power users should be told that they can pass in any kind of `http.Client` that they want if they wish redirects to be followed.

Does that make sense?  What do you think about making this change?

andersjanmyr: Okay, it sounds reasonable. I'll make the change

andersjanmyr: Done.

github/repos_releases.go
```diff
@@ -266,10 +266,13 @@ func (s *RepositoriesService) GetReleaseAsset(ctx context.Context, owner, repo s
 // If a redirect is returned, the redirect URL will be returned as a string instead
 // of the io.ReadCloser. Exactly one of rc and redirectURL will be zero.
 //
-// If followRedirects is true, a single redirect will be followed.
+// followRedirectsClient can be passed to download the asset from a redirected
+// location. Passing http.DefaultClient is recommended unless special circumstances
+// exists, but it's possible to pass any http.Client. If nil is passed the
```
gmlewis: nit: exist (not exists).

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1394, opened 2020-01-23
Adding Node ID to contributor struct

gauntface:


willnorris:
see test failures... looks like you need to `go fmt`?

PR #1395, opened 2020-01-24
Use normal object initialization instead of new()

rsjethani:
There is no need for new() is this case and buf can be initialized in
a single statement as proposed.


rsjethani:
@gmlewis I have changed the title to match changes. My objective here is to simply avoid another usage of new() when there is a standard and idiomatic way to do the same thing. If this sounds good enough then I will change commit message also.

PR #1396, opened 2020-01-26
use pkg.go.dev for documentation

willnorris:
pkg.go.dev understands modules, whereas godoc.org does not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # go-github #
 
-[![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # go-github #
 
-[![GoDoc](https://godoc.org/github.com/google/go-github/github?status.svg)](https://godoc.org/github.com/google/go-github/github)
+[![GoDoc](https://img.shields.io/static/v1?label=godoc&message=reference&color=blue)](https://pkg.go.dev/github.com/google/go-github/v28/github)
```
gmlewis: Do you want this to point to `v29` instead of `v28`?

willnorris: ah yes, fixed.  I guess I was looking at an old commit and thought we were still on v28 :)

PR #1405, opened 2020-01-31
RepositoryContent.GetContent(): Check nil before decoding

carlmjohnson:


github/repos_contents.go
```diff
@@ -76,6 +76,9 @@ func (r *RepositoryContent) GetContent() (string, error) {
 
 	switch encoding {
 	case "base64":
+		if r.Content == nil {
+			return "", fmt.Errorf("malformed response: base64 encoding of null content")
```
gmlewis: s/fmt.Errorf/errors.New/

PR #1409, opened 2020-02-03
Support specification of parameters for enable pages API

### LINKED ISSUE #1408: Enable Pages Parameter specification
Allow specification of parameters when POST to enable pages API as per: https://developer.github.com/v3/repos/pages/#parameters

patrickmarabeas:
Support specification of parameters for enable pages API as per: https://developer.github.com/v3/repos/pages/#parameters

Fixes #1408.

### DIFF
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -46,7 +46,7 @@ type PagesBuild struct {
 // createPagesRequest is a subset of Pages and is used internally
 // by EnablePages to pass only the known fields for the endpoint.
 type createPagesRequest struct {
-	Source *PagesSource
+	Source *PagesSource `json:"source,omitempty"`
 }
 
 // EnablePages enables GitHub Pages for the named repo.
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -43,12 +43,23 @@ type PagesBuild struct {
 	UpdatedAt *Timestamp  `json:"updated_at,omitempty"`
 }
 
+// createPagesRequest is a subset of Pages and is used internally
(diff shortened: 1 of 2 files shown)


patrickmarabeas:
@googlebot I signed it!

github/repos_pages.go
```diff
@@ -43,12 +43,23 @@ type PagesBuild struct {
 	UpdatedAt *Timestamp  `json:"updated_at,omitempty"`
 }
 
+// createPagesRequest is a subset of Pages and is used internally
+// by EnablePages to pass only the known fields for the endpoint.
+type createPagesRequest struct {
+	Source *PagesSource
```
gmlewis: This needs to be:
```
Source *PagesSource `json:"source"`
```

and the unit tests should have caught that this was not passed properly to the backend.

Please add a section to the unit test where the JSON is decoded and a `reflect.DeepEqual` is called to test the parameters.  Here is an example: https://github.com/google/go-github/blob/master/github/users_projects_test.go#L44-L59

In fact, if you update the test first, you will see the failure I'm talking about, then you can fix line 49 in this file.

patrickmarabeas: I've also included the `omitempty`.

I wasn't able to get the unit test to error. Similarly, removing the json annotation from `createRepoRequest.name` field doesn't cause unit tests to fail.

gmlewis:
Hi @patrickmarabeas - in the future, please don't use force-push on PRs in this repo, as it makes it more difficult and time-consuming for reviewers to see what changed since their last code review.  Thanks.

gmlewis:
Thank you, @gauntface!
Merging.

PR #1413, opened 2020-02-08
Add the AuthorAssociation field to PullRequestReview

qaisjp:
#892 and #802, electric boogaloo.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -8708,6 +8708,14 @@ func (p *PullRequestMergeResult) GetSHA() string {
 	return *p.SHA
 }
 
+// GetAuthorAssociation returns the AuthorAssociation field if it's non-nil, zero value otherwise.
+func (p *PullRequestReview) GetAuthorAssociation() string {
+	if p == nil || p.AuthorAssociation == nil {
+		return ""
+	}
+	return *p.AuthorAssociation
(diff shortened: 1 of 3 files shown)


gmlewis:
Thank you, @qaisjp.

As in #892, could you please contact support@github.com and request that they update their examples where appropriate on this page: https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request
to include this field and report back their response?

qaisjp:
sent.

I suspect there will be more places where this is missing, a future PR author can use this canned message:

```
Some of the responses on https://developer.github.com/v3/pulls/reviews/ are missing the `author_association` key, specifically https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request.

The API is returning that info, but it's not documented. Is that a documentation mistake or should `author_association` not actually be in the response?
```

qaisjp:
> Hello -
> 
> Thanks for the report and sorry for the delay!
> 
> I believe this is a documentation issue - we'll go ahead and mention this to our documentation team and if I find out differently, we'll let you know.
> 
> Thanks again for reporting!

gmlewis:
Thank you for reporting back their response, @qaisjp!
It is greatly appreciated!

Issue #1422, opened 2020-02-11
Update teams_members.go


adrienzieba:
Hello,

Related to https://github.com/google/go-github/issues/1387, I noticed that all teams endpoints are affected

- https://developer.github.com/v3/teams/members/
- https://developer.github.com/v3/teams/discussion_comments/
- https://developer.github.com/v3/teams/discussions/

Is it possible to update other teams files?

gmlewis:
Thank you, @adrienzieba!

It seems that the [original blog post](https://developer.github.com/changes/2020-01-21-moving-the-team-api-endpoints/) that we were basing #1387 on didn't mention these changes.

Let's split this up into three issues to organize the work:

This issue will be for updating `teams_members.go`.

The work involved here it to update every endpoint listed here:
https://developer.github.com/v3/teams/members/
Note that #1400 can be used as a model because most endpoints have two ways to call them: `*ByID` and `*BySlug`. The same pattern should be followed for these updates.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

adrienzieba:
Hello,

I want to work on this one. Can you assign it to me please?

Thank you

gmlewis:
Thank you, @adrienzieba!
It is yours.

PR #1428, opened 2020-02-12
Update teams members endpoints+Remove deprecated ones

adrienzieba:
Fixes: https://github.com/google/go-github/issues/1422.

### DIFF
--- a/github/teams_members.go
+++ b/github/teams_members.go
@@ -20,12 +20,12 @@ type TeamListTeamMembersOptions struct {
 	ListOptions
 }
 
-// ListTeamMembers lists all of the users who are members of the specified
-// team.
+// ListTeamMembersByID lists all of the users who are members of a team, given a specified
+// organization ID, by ID.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/members/#list-team-members
-func (s *TeamsService) ListTeamMembers(ctx context.Context, team int64, opts *TeamListTeamMembersOptions) ([]*User, *Response, error) {
-	u := fmt.Sprintf("teams/%v/members", team)
+func (s *TeamsService) ListTeamMembersByID(ctx context.Context, orgID, teamID int64, opts *TeamListTeamMembersOptions) ([]*User, *Response, error) {
+	u := fmt.Sprintf("organizations/%v/team/%v/members", orgID, teamID)
 	u, err := addOptions(u, opts)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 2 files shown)


github/teams_members.go
```diff
@@ -45,29 +45,57 @@ func (s *TeamsService) ListTeamMembers(ctx context.Context, team int64, opts *Te
...
+
+	return members, resp, nil
+}
+
+// GetTeamMembershipByID returns the membership status for a user in a team, given a specified
+// organization ID, by ID.
 //
-// Deprecated: This API has been marked as deprecated in the Github API docs,
-// TeamsService.GetTeamMembership method should be used instead.
-func (s *TeamsService) IsTeamMember(ctx context.Context, team int64, user string) (bool, *Response, error) {
```
adrienzieba: I removed this method

github/teams_members.go
```diff
@@ -20,12 +20,12 @@ type TeamListTeamMembersOptions struct {
 	ListOptions
 }
 
-// ListTeamMembers lists all of the users who are members of the specified
-// team.
+// ListTeamMembersByID lists all of the users who are members of a team, given a specified
+// organization ID, by ID.
```
gmlewis: For all of these endpoints where it says "organization ID, by ID" could you please add in the word "team"
so that it reads "organization ID, by team ID"?  I think this would be clearer.

Likewise, where it says "organization name, by slug", could you say "organization name, by team slug"?

github/teams_members.go
```diff
@@ -96,26 +124,32 @@ type TeamAddTeamMembershipOptions struct {
 	Role string `json:"role,omitempty"`
 }
 
-// AddTeamMembership adds or invites a user to a team.
-//
-// In order to add a membership between a user and a team, the authenticated
```
gmlewis: Aren't all these comments (from old lines 101 through 114) still valid?
If so, can we please keep them?

github/teams_members.go
adrienzieba: sure i'll do the modification

github/teams_members.go
adrienzieba: On the github website it was a bit different. Instead of updating it, I removed it because, for other endpoints, the documentation is not imported as we have the link to the github documentation. 
I can put it back with the updated description if you want.

gmlewis: Oh, OK... gotcha.  Thanks, @adrienzieba... I'm fine either way, actually.
You are right that the other endpoints don't have it and we do link to the main docs, so that is probably sufficient.

adrienzieba:
Hello @gmlewis,
Any news about this PR?
Thanks

gmlewis:
Hi @adrienzieba - no news yet.  This repo is maintained by volunteers and we need to be patient for code reviews. Thank you for your understanding.

adrienzieba:
Thanks for the update. 
I understand. Waiting for the merge, I'm using my own fork :)

gmlewis:
Thank you, @wesleimp!
Merging.

227234 (commented):
@googlebot

github/teams_members.go
227234: > I removed this method

PR #1118, opened 2019-02-11
Added test case for commit ref with percent encoding

vaibhavsingh97:
Fixes https://github.com/google/go-github/issues/1101

Added test case for `Repositories.GetCommitSHA1`, when it was called with the branch name which includes the non-alphabet character.

Issue #1101, opened 2019-01-19
Add unit tests for URL escaping for ref parameters


gmlewis:
This issue is so that we don't lose track of @gauntface's [comment](https://github.com/google/go-github/pull/1099#pullrequestreview-193804131) about adding unit tests to #1099.

vaibhavsingh97:
@gmlewis I want to take this issue. Please assign to me

gmlewis:
I can't assign the issue to you, @vaibhavsingh97 but consider it yours.
Thank you!

vaibhavsingh97:
Thanks, will send PR for same 🤗

Avni-Sharma:
Hi @vaibhavsingh97 , are you working on this?

vaibhavsingh97:
@Avni-Sharma yes, sorry for the delay was a busy week, will send PR shortly

vaibhavsingh97:
@gmlewis I am stuck here for long time, please help me with this
I wrote the unit test but I don't know how to test it. I read how to contribute file aka `contributing.md`, but when I forked the repository and run `go build ./...` and `go test ./...` and it failed to find packages but when I clone `go-github` and run same commands then I am able to build and compile
![image](https://user-images.githubusercontent.com/8705386/52371896-b6a3a500-2a7c-11e9-8aa3-00ae804bd483.png)

gmlewis:
It looks like you are missing dependencies needed by this package. You should be able to run `go get -u github.com/google/go-github` and it will get all the needed dependencies. Alternatively, you can run `go get -u` on each package that is missing, but that  shouldn't be necessary.

PR #1433, opened 2020-02-17
do not encode / in ref names

### LINKED ISSUE #1432: Refs should not encode /
The commit #1099 changed the ref name to be URL Query encoded. This breaks ref names which contain a /:

A call to `curl -vH "Authorization: Bearer $TOKEN" https://$GHE_URL/repos/o/r/git/ref/heads/master` works, but `https://$GHE_URL/repos/o/r/git/ref/heads%2Fmaster` (like `GetRef()`, `GetRefs()` use it) fail with 404.

vetinari:
fixes #1432

Signed-off-by: Hanno Hecker <hanno@zalando.de>

### DIFF
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -58,7 +58,7 @@ type updateRefRequest struct {
 // GitHub API docs: https://developer.github.com/v3/git/refs/#get-a-reference
 func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref string) (*Reference, *Response, error) {
 	ref = strings.TrimPrefix(ref, "refs/")
-	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, url.QueryEscape(ref))
+	u := fmt.Sprintf("repos/%v/%v/git/refs/%v", owner, repo, refURLEscape(ref))
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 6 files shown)


github/git_refs.go
```diff
@@ -79,10 +79,12 @@ func (s *GitService) GetRef(ctx context.Context, owner string, repo string, ref
 	return r, resp, nil
 }
 
+// refURLEscape escapes every path segment of the given ref. Those must
+// not contain escaped "/" - as "%2F" - or github will not recognize it.
 func refURLEscape(ref string) string {
 	parts := strings.Split(ref, "/")
 	for i, s := range parts {
-		parts[i] = url.QueryEscape(s)
+		parts[i] = url.PathEscape(s)
```
gmlewis: Hmmm... with this change from `url.QueryEscape` to `url.PathEscape`, I'm now concerned that we are breaking what was fixed in #1099.

Can you please add a unit test that demonstrates the URL found in #1099?

Unfortunately, I did not request that unit tests be added in #1099 to demonstrate the before/after changes. :-(
But I'm trying to fix that now.

vetinari: There's one in #1118 (issue #1101). This still works, `GetCommitSHA1()` got the `refURLEscape()` also. I've added a test case for the original trailing `%` as given in #1099 where http.NewRequest was failing to parse the URL.

gmlewis:
Thank you, @wesleimp!
Merging.

Issue #1100, opened 2019-01-16
Pointer in Issue.Labels missing?


artem-sidorenko:
Almost everywhere pointers are used, but not [here](https://github.com/google/go-github/blob/e740a6811a5bc7a8fba64fe1594860d4f982b4c3/github/issues.go#L36). The type is `[]Label`. The similar struct for PRs [has](https://github.com/google/go-github/blob/e740a6811a5bc7a8fba64fe1594860d4f982b4c3/github/pulls.go#L33) also a type of `[]*Label`.

Is this intentional?

gmlewis:
Looking back at the history of this line, we find:
https://github.com/google/go-github/commit/b416fdaac841e7c2404b365cf6aa7b9f1d450d93
which is one of the early commits.

I'm betting that there wasn't a whole lot of precedent at that time.

Is this causing problems for you, or are you just curious?

artem-sidorenko:
@gmlewis thank you for having a look,

> Is this causing problems for you, or are you just curious?

Not really, as there are no generics (and I don't want to bypass compiler type checking via reflect), I have some duplicate code for handling of issues and PRs. So it's not a problem to have different types here. 

I was just wondering about that. Not sure if fixing this is a good idea: it would potentially break everything, what is using this part of code. Up to you, feel free to close this issue or let me know and I'll provide a fix.

gmlewis:
OK, thank you, @artem-sidorenko!

Since this has been like this for almost 6 years, I'm leaning toward not making a breaking change and leaving it alone. I'm betting that there is no real performance concern here either (one way or the other).

So I appreciate the report because it is good to point out inconsistencies and discuss them!

Let's leave this open for a couple days in case anyone wants to chime in if they have a strong opinion, and then we'll go ahead and close the issue otherwise.

wesleimp:
@gmlewis what about this issue, how can we proceed?

gmlewis:
Actually, in light of #180 and all the breaking API changes we've been making lately, I think I will go ahead and fix this one myself.

PR #1443, opened 2020-02-28
Add pointers to structs in slices

gmlewis:
Since we decided in #180 to standardize using pointers to structs when they appear in slices, and some slices have slipped through code reviews without having pointers, this PR cleans these up.

Fixes: #1100. 

This is a breaking API change.


joebowbeer:
@gmlewis Why was this breaking change in the API released in v29.0.3? I was expecting v29.0.3 to be interchangeable with v29.0.2

gmlewis:
I apologize... that was a slip on my part.  Someone asked for a new release to be created and I forgot that breaking changes were included.  I don't know how to fix it now, as it seems that once a release is made, it is archived in the mod tools.

To help this situation, I might be able to create a new v30.0.0 release now, but I'm afraid the damage has been done, and I'm sorry about that.

joebowbeer:
My mistake: I don't think this was released to v29.0.3. It is one of the 17 commits *since* v29.0.3

joebowbeer:
@gmlewis However, I am seeing the incompatible master src when I go get v29, which is causing inexplicable type errors to surface. A v30 release would be helpful, and would make it clear that master go.mod is v30 source.

gmlewis:
Alright, in that case I know for a fact that we have a large number of breaking changes between v29 and v30 and I believe one or two outstanding PRs will further break things. I was planning a v30 release would happen really soon in the hopes that the breaking changes would subside for a while (many APIs have been deprecated and changed lately from the v3 API)... so if we could hold off for just a few days, I think it would be the least disruptive for all users of the repo.

I'll try to prod along the outstanding PRs... but I can make no guarantees as to schedule, unfortunately.

gmlewis:
Taking a more detailed look, it appears that #1451 is the only one that is close that should get in before `v30.0.0`.  I need a second LGTM on it, hopefully from @dmitshur but he probably wouldn't mind if another LGTM came in from someone else, since it appears to me that his comments have already been addressed.

gmlewis:
@joebowbeer - here's the plan - one more change is needed in #1451, then we can get #1456 approved by two LGTMs and merged, then we can release `v30.0.0`. Hopefully this can all happen today.

Issue #1442, opened 2020-02-27
Support new "Delete Reactions" endpoints


gmlewis:
GitHub Developer API announcement:
https://developer.github.com/changes/2020-02-26-new-delete-reactions-endpoints/

These new endpoints need to be supported:

- [x] Delete a commit comment reaction- #1445
  * `DELETE /repos/:owner/:repo/comments/:comment_id/reactions/:reaction_id`
  * Docs: https://developer.github.com/v3/reactions/#delete-a-commit-comment-reaction
- [x] Delete an issue reaction
  * `DELETE /repos/:owner/:repo/issues/:issue_number/reactions/:reaction_id`
  * Docs: https://developer.github.com/v3/reactions/#delete-an-issue-reaction
- [x] Delete a reaction to a commit comment
  * `DELETE /repos/:owner/:repo/issues/comments/:comment_id/reactions/:reaction_id`
  * https://developer.github.com/v3/reactions/#delete-an-issue-comment-reaction
- [x] Delete a pull request comment reaction
  * `DELETE /repos/:owner/:repo/pulls/comments/:comment_id/reactions/:reaction_id`
  * https://developer.github.com/v3/reactions/#delete-a-pull-request-comment-reaction
- [x] Delete team discussion reaction
  * `DELETE /orgs/:org/teams/:team_slug/discussions/:discussion_number/reactions/:reaction_id`
  * Docs: https://developer.github.com/v3/reactions/#delete-team-discussion-reaction
- [x] Delete team discussion comment reaction
  * `DELETE /orgs/:org/teams/:team_slug/discussions/:discussion_number/comments/:comment_number/reactions/:reaction_id`
  * Docs: https://developer.github.com/v3/reactions/#delete-team-discussion-comment-reaction

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

wesleimp:
@gmlewis I could take it over

gmlewis:
Thank you, @wesleimp!  It is yours.

gmlewis:
I neglected to mention that the existing "delete reactions" endpoint also needs to be deleted (assuming we implemented it... I'll have to check later).

wesleimp:
[...]

PR #1445, opened 2020-02-28
Add support to delete commit comment reaction

wesleimp:
Delete a commit comment reaction
- `DELETE /repos/:owner/:repo/comments/:comment_id/reactions/:reaction_id`
- Docs: https://developer.github.com/v3/reactions/#delete-a-commit-comment-reaction

### DIFF
--- a/github/reactions.go
+++ b/github/reactions.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"net/http"
 )
 
 // ReactionsService provides access to the reactions-related functions in the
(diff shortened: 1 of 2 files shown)


github/reactions.go
```diff
@@ -109,6 +110,23 @@ func (s ReactionsService) CreateCommentReaction(ctx context.Context, owner, repo
 	return m, resp, nil
 }
 
+// DeleteCommentReaction delete the reaction for a commit comment.
```
gmlewis: s/delete/deletes/

github/reactions.go
```diff
@@ -109,6 +110,23 @@ func (s ReactionsService) CreateCommentReaction(ctx context.Context, owner, repo
...
+//
+// GitHub API docs: https://developer.github.com/v3/reactions/#delete-a-commit-comment-reaction
+func (s *ReactionsService) DeleteCommentReaction(ctx context.Context, owner, repo string, id, reactionID int64) (*Response, error) {
```
gmlewis: Let's please change `id` to `commentID` here.

wesleimp:
> Thank you, @wesleimp!
> 
> I think we should also add a new `DeleteCommentReactionByRepoID` method in this same PR which supports the comment in the docs that says:
> 
> > Note: You can also specify a repository by repository_id using the route DELETE /repositories/:repository_id/comments/:comment_id/reactions/:reaction_id.
> 
> You can either call out to a private method where these two public methods just generate the URL and pass it to the private method that does the rest (which I personally prefer), or just have to fully complete public methods. Your choice.

@gmlewis I think now the changes are ok

wesleimp:
@googlebot I signed it!

gmlewis:
@martinssipenko - if you feel like reviewing and approving, your help would be greatly appreciated!

gmlewis:
Thank you, @martinssipenko !
Merging.

gmlewis:
Relates to: #1442.

Issue #1399, opened 2020-01-28
Support GitHub Actions API


joshuabezaleel:
Announcement: https://developer.github.com/v3/actions/

gmlewis:
Usually there is a blog announcement, and apparently there wasn't for this.  Thanks, @joshuabezaleel.

This would be a great PR for any new contributor to this repo or any Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

gmlewis:
Assigned to @martinssipenko.

gmlewis:
In order to fully close this issue, we need PRs that will address all the following sections of the announcement:

- [ ] `actions_artifacts.go`
- [x] `actions_secrets.go` (#1402 - assigned to @martinsspenko)
- [ ] `actions_runners.go`
- [ ] `actions_workflows.go`
- [ ] `actions_workflow_jobs.go`
- [ ] `actions_workflow_runs.go`

Since this is a lot of work, let's open this up to multiple developers... please volunteer for one section in the list above, and piggy-back upon the excellent work that @martinsspenko has started in #1402. 

Thank you!

joshuabezaleel:
Haven't got a chance to look at it but I want to take a shot. Perhaps the `actions_workflows` one.

atorr:
May I contribute to the actions_artifacts section?

joshuabezaleel:
Since my incoming PR will be dependent to @martinssipenko 's [PR](https://github.com/google/go-github/pull/1402) that contains `ActionsService` on `actions.go`, should I wait for that particular PR to be merged or I can raise a PR with `ActionsService` written in my code as if it's already merged? 

cc: @gmlewis

martinssipenko:
@joshuabezaleel you can branch off of my branch.

gmlewis:
Hey volunteers, I am super-pleased with the way this Issue is coming together... starting off with the excellent groundwork from @martinssipenko followed by another great PR from @joshuabezaleel!

Let's keep up this excellent progress going by having more people volunteer to complete the unchecked and unassigned items [above](https://github.com/google/go-github/issues/1399#issuecomment-579554516).

[...]

PR #1449, opened 2020-03-01
Implement support for actions workflow runs

ayzu:
Relates to: #1399.

### DIFF
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -12,53 +12,37 @@ import (
 	"net/url"
 )
 
-// PersonWorkflowRun represents a person object as returned in actions workflow jobs endpoint response.
-type PersonWorkflowRun struct {
-	Name  *string `json:"name,omitempty"`
-	Email *string `json:"email,omitempty"`
-}
-
-// CommitWorkflowRun represents a commit object as returned in actions workflow jobs endpoint response.
-type CommitWorkflowRun struct {
-	ID        *string            `json:"id,omitempty"`
-	TreeID    *string            `json:"tree_id,omitempty"`
-	Message   *string            `json:"message,omitempty"`
-	Timestamp *Timestamp         `json:"timestamp,omitempty"`
-	Author    *PersonWorkflowRun `json:"author,omitempty"`
-	Committer *PersonWorkflowRun `json:"committer,omitempty"`
-}
-
 // WorkflowRun represents a repository action workflow run.
 type WorkflowRun struct {
-	ID             *int64            `json:"id,omitempty"`
-	NodeID         *string           `json:"node_id,omitempty"`
-	HeadBranch     *string           `json:"head_branch,omitempty"`
-	HeadSHA        *string           `json:"head_sha,omitempty"`
-	RunNumber      *int64            `json:"run_number,omitempty"`
-	Event          *string           `json:"event,omitempty"`
-	Status         *string           `json:"status,omitempty"`
-	Conclusion     *string           `json:"conclusion,omitempty"`
-	URL            *string           `json:"url,omitempty"`
-	HTMLURL        *string           `json:"html_url,omitempty"`
-	PullRequests   []*PullRequest    `json:"pull_requests,omitempty"`
-	CreatedAt      *Timestamp        `json:"created_at,omitempty"`
-	UpdatedAt      *Timestamp        `json:"updated_at,omitempty"`
-	JobsURL        *string           `json:"jobs_url,omitempty"`
-	LogsURL        *string           `json:"logs_url,omitempty"`
-	CheckSuiteURL  *string           `json:"check_suite_url,omitempty"`
-	ArtifactsURL   *string           `json:"artifacts_url,omitempty"`
-	CancelURL      *string           `json:"cancel_url,omitempty"`
-	RerunURL       *string           `json:"rerun_url,omitempty"`
-	HeadCommit     CommitWorkflowRun `json:"head_commit,omitempty"`
-	WorkflowURL    *string           `json:"workflow_url,omitempty"`
-	Repository     *Repository       `json:"repository,omitempty"`
-	HeadRepository *Repository       `json:"head_repository,omitempty"`
+	ID             *int64         `json:"id,omitempty"`
+	NodeID         *string        `json:"node_id,omitempty"`
+	HeadBranch     *string        `json:"head_branch,omitempty"`
+	HeadSHA        *string        `json:"head_sha,omitempty"`
+	RunNumber      *int           `json:"run_number,omitempty"`
+	Event          *string        `json:"event,omitempty"`
+	Status         *string        `json:"status,omitempty"`
+	Conclusion     *string        `json:"conclusion,omitempty"`
+	URL            *string        `json:"url,omitempty"`
+	HTMLURL        *string        `json:"html_url,omitempty"`
+	PullRequests   []*PullRequest `json:"pull_requests,omitempty"`
+	CreatedAt      *Timestamp     `json:"created_at,omitempty"`
+	UpdatedAt      *Timestamp     `json:"updated_at,omitempty"`
+	JobsURL        *string        `json:"jobs_url,omitempty"`
+	LogsURL        *string        `json:"logs_url,omitempty"`
+	CheckSuiteURL  *string        `json:"check_suite_url,omitempty"`
+	ArtifactsURL   *string        `json:"artifacts_url,omitempty"`
+	CancelURL      *string        `json:"cancel_url,omitempty"`
+	RerunURL       *string        `json:"rerun_url,omitempty"`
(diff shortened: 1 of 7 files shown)


gmlewis:
Thank you, @ayzu . It looks like you need to merge master into your PR because it is modifying files it shouldn't which might mean that you based your work on another PR that got changed during a code review. ... If I'm making any sense at all.

I can be more detailed if this doesn't make sense... But that should fix the broken tests.

ayzu:
> Thank you, @ayzu . It looks like you need to merge master into your PR because it is modifying files it shouldn't which might mean that you based your work on another PR that got changed during a code review. ... If I'm making any sense at all.
> 
> I can be more detailed if this doesn't make sense... But that should fix the broken tests.

Thank you, @gmlewis 

I didn't run `go-generate` :) By the way, what do you think of adding a Makefile like:

```
do:
	go generate ./...
	go test ./...
	go vet ./...
```

gmlewis:
> I didn't run go-generate :) By the way, what do you think of adding a Makefile like:

:smile: Funny you should mention it. I've had this bash script in my directory for a long time:

```bash
$ cat testit.sh 
#!/bin/bash -ex
GO111MODULE=on
go generate ./...
go test ./...
go vet ./...
```

I could probably get rid of that environment variable now.  :smile:

Adding a `Makefile` to Go projects just brings back memories of waiting for C++ compilations, so I'd rather not.  :smile:

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+	ArtifactsURL   *string           `json:"artifacts_url,omitempty"`
+	CancelURL      *string           `json:"cancel_url,omitempty"`
+	RerunURL       *string           `json:"rerun_url,omitempty"`
+	HeadCommit     CommitWorkflowRun `json:"head_commit,omitempty"`
+	WorkflowURL    *string           `json:"workflow_url,omitempty"`
+	Repository     *Repository       `json:"repository,omitempty"`
+	HeadRepository *Repository       `json:"head_repository,omitempty"`
+}
+
+// Runs represents a slice of repository action workflow run.
```
gmlewis: Let's please rename this to `WorkflowRuns`.

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+package github
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// PersonWorkflowRun represents a person object as returned in actions workflow jobs endpoint response.
```
gmlewis: Let's please remove this struct entirely. See below.

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+type PersonWorkflowRun struct {
+	Name  *string `json:"name,omitempty"`
+	Email *string `json:"email,omitempty"`
+}
+
+// CommitWorkflowRun represents a commit object as returned in actions workflow jobs endpoint response.
```
gmlewis: Let's please remove this struct entirely. See below.

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+	PullRequests   []*PullRequest    `json:"pull_requests,omitempty"`
+	CreatedAt      *Timestamp        `json:"created_at,omitempty"`
+	UpdatedAt      *Timestamp        `json:"updated_at,omitempty"`
+	JobsURL        *string           `json:"jobs_url,omitempty"`
+	LogsURL        *string           `json:"logs_url,omitempty"`
+	CheckSuiteURL  *string           `json:"check_suite_url,omitempty"`
+	ArtifactsURL   *string           `json:"artifacts_url,omitempty"`
+	CancelURL      *string           `json:"cancel_url,omitempty"`
+	RerunURL       *string           `json:"rerun_url,omitempty"`
+	HeadCommit     CommitWorkflowRun `json:"head_commit,omitempty"`
```
gmlewis: I see in this repo that we have a very similar struct that appears to serve the same purpose that I think might be good to reuse here: `PushEventCommit` in `event_types.go`, although some might argue with me that `PushEventCommit` is strictly to be used for events.

I would argue that the `HeadCommit` is how that struct is being used, and maybe it should be renamed to `HeadCommit` and then used in both places to represent the same thing.

Either way, whether `PushEventCommit` is renamed to `HeadCommit` in this PR or not, this instance of `HeadCommit` should be a pointer to a struct, not an embedded struct.

As for the rename of `PushEventCommit`, I'll leave it up to you and the second reviewer of this PR, as I could go either way.

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+	Committer *PersonWorkflowRun `json:"committer,omitempty"`
+}
+
+// WorkflowRun represents a repository action workflow run.
+type WorkflowRun struct {
+	ID             *int64            `json:"id,omitempty"`
+	NodeID         *string           `json:"node_id,omitempty"`
+	HeadBranch     *string           `json:"head_branch,omitempty"`
+	HeadSHA        *string           `json:"head_sha,omitempty"`
+	RunNumber      *int64            `json:"run_number,omitempty"`
```
gmlewis: I'm thinking `RunNumber` should be an `*int` here, not an `*int64`.

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+	HeadCommit     CommitWorkflowRun `json:"head_commit,omitempty"`
+	WorkflowURL    *string           `json:"workflow_url,omitempty"`
+	Repository     *Repository       `json:"repository,omitempty"`
+	HeadRepository *Repository       `json:"head_repository,omitempty"`
+}
+
+// Runs represents a slice of repository action workflow run.
+type Runs struct {
+	TotalCount *int           `json:"total_count,omitempty"`
+	Runs       []*WorkflowRun `json:"workflow_runs,omitempty"`
```
gmlewis: s/Runs/WorkflowRuns/
for consistency with the JSON parameter name.

github/actions_workflow_runs.go
```diff
@@ -0,0 +1,197 @@
...
+// GitHub API docs: https://developer.github.com/v3/actions/workflow_runs/#list-workflow-runs
+func (s *ActionsService) ListWorkflowRunsByID(ctx context.Context, owner, repo string, workflowID int64, opts *ListWorkflowRunsOptions) (*Runs, *Response, error) {
+	u := fmt.Sprintf("repos/%s/%s/actions/workflows/%v/runs", owner, repo, workflowID)
+	return s.listWorkflowRuns(ctx, u, opts)
+}
+
+// ListWorkflowRunsByFileName lists all workflow runs by workflow file name.
+//
+// GitHub API docs: https://developer.github.com/v3/actions/workflow_runs/#list-workflow-runs
+func (s *ActionsService) ListWorkflowRunsByFileName(ctx context.Context, owner, repo string, workflowFileName string, opts *ListWorkflowRunsOptions) (*Runs, *Response, error) {
```
gmlewis: `... owner, repo, workflowFileName string,...`

ayzu:
> This is looking great, @ayzu!
> Just a few tweaks, please, and then I think we'll be ready for a second review before merging.

Thank you for the review, @gmlewis 

I updated the code and renamed the structure to `HeadCommit`. It is great that we can reuse an existing type.

github/actions_workflow_runs_test.go
```diff
@@ -31,11 +31,11 @@ func TestActionsService_ListWorkflowRunsByID(t *testing.T) {
...
 	}
 
-	want := &Runs{
+	want := &WorkflowRuns{
 		TotalCount: Int(4),
-		Runs: []*WorkflowRun{
-			{ID: Int64(399444496), RunNumber: Int64(29679449), CreatedAt: &Timestamp{time.Date(2019, time.January, 02, 15, 04, 05, 0, time.UTC)}, UpdatedAt: &Timestamp{time.Date(2020, time.January, 02, 15, 04, 05, 0, time.UTC)}},
-			{ID: Int64(399444497), RunNumber: Int64(29679449), CreatedAt: &Timestamp{time.Date(2019, time.January, 02, 15, 04, 05, 0, time.UTC)}, UpdatedAt: &Timestamp{time.Date(2020, time.January, 02, 15, 04, 05, 0, time.UTC)}},
+		WorkflowRuns: []*WorkflowRun{
+			{ID: Int64(399444496), RunNumber: Int(29679449), CreatedAt: &Timestamp{time.Date(2019, time.January, 02, 15, 04, 05, 0, time.UTC)}, UpdatedAt: &Timestamp{time.Date(2020, time.January, 02, 15, 04, 05, 0, time.UTC)}},
```
gmlewis: nit: for the `RunNumber` values, let's kick this down to maybe `296` or another relatively small number so that we don't give the impression that these are IDs from a quick glance.

gmlewis:
Thank you, @wesleimp!
Merging.

Issue #1361, opened 2020-01-02
How to add custom domain to Github Page ?


proabiral:
I am trying to add custom domain to Github Repository by enabling github pages but I can not find right method for this in the documentation. 

[Here](https://developer.github.com/v3/repos/pages/#update-information-about-a-pages-site) is the API documentation for reference.

gmlewis:
Hi @proabiral, it looks like you found an endpoint that we haven't yet implemented.
Thank you for alerting us to this!

Would you like to submit a PR to fix this, or would you like me to open this up as an issue for one of our great volunteers to fix?

proabiral:
Hi @gmlewis ,
Thanks for such a quick response.

I would love to submit a PR to fix this but I have never worked on such a large project like this before. Is there any documentation that I can refer to that would help to better overview of the project ?

gmlewis:
Welcome to the project, @proabiral !
The best place to get started is to look at our CONTRIBUTING.md document at the top of our repo.

After that, feel free to ask questions. 😁

proabiral:
Hi @gmlewis , 

I have tried to implemented the feature on my fork [here](https://github.com/proabiral/go-github/commit/71bb479f4ede0bf0c5db2b21200b56031b8c9b4d).

Can you please provide feedback on this ?

gmlewis:
After looking at the PR and reading the [v3 API docs](https://developer.github.com/v3/repos/pages/#update-information-about-a-pages-site) again, it is not clear to me if the `cname` parameter is required, thoughts?

It seems like `source` is required, but if you aren't sure about `cname` as well, @proabiral, could you please contact support@github.com and point them to this API and ask them if `cname` (and `source` while you're asking) are required?   Also, could you ask them to please mark them appropriately as "Optional" or "Required"?  The, when you get a reply, could you please report back here?  (Lately they have been responding within a few hours... pretty impressive.)

That would help quite a bit... thank you!

gmlewis:
> Hi @gmlewis ,
> I have tried to implemented the feature on my fork here.
> Can you please provide feedback on this ?

Feedback provided, but as I commented in the commit, let's please create a PR first and then the code review history will be kept in this repo and others who watch this repo will be able to better participate as well, especially if I give feedback that they disagree with.  :smile:

proabiral:
Hi @gmlewis ,
Thanks for such a detailed feedback. :) Really appreciate it.

According to Github Support,
```
[...]

PR #1453, opened 2020-03-05
Implement support for update pages API

patrickmarabeas:
Implements support for updating Github Pages sites - allowing CName and
Source to be customized.

Fixes: #1361

### DIFF
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -100,8 +100,6 @@ func (s *RepositoriesService) UpdatePages(ctx context.Context, owner, repo strin
 		return nil, err
 	}
 
-	req.Header.Set("Accept", mediaTypeEnablePagesAPIPreview)
-
 	if *opts.CNAME == "" {
 		update := new(pagesUpdateNullCNAME)
 		resp, err := s.client.Do(ctx, req, update)
@@ -112,8 +110,7 @@ func (s *RepositoriesService) UpdatePages(ctx context.Context, owner, repo strin
 		return resp, nil
 	}
 
-	update := new(PagesUpdate)
-	resp, err := s.client.Do(ctx, req, update)
+	resp, err := s.client.Do(ctx, req, nil)
 	if err != nil {
 		return resp, err
 	}
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -78,18 +78,10 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
 // PagesUpdate sets up parameters needed to update a GitHub Pages site.
 type PagesUpdate struct {
 	// CNAME represents a custom domain for the repository.
-	// Setting CNAME as an empty string will remove the custom domain.
-	CNAME *string `json:"cname,omitempty"`
+	// Leaving CNAME empty will remove the custom domain.
+	CNAME *string `json:"cname"`
 	// Source must include the branch name, and may optionally specify the subdirectory "/docs".
 	// Possible values are: "gh-pages", "master", and "master /docs".
-	Source *string `json:"source,omitempty"`
-}
-
-// pagesUpdateNullCNAME is used internally by UpdatePages to pass a null
-// value for the CNAME field when an empty string has been provided in
-// order to remove the custom domain.
-type pagesUpdateNullCNAME struct {
-	CNAME  *string `json:"cname"`
 	Source *string `json:"source,omitempty"`
 }
 
@@ -102,16 +94,6 @@ func (s *RepositoriesService) UpdatePages(ctx context.Context, owner, repo strin
 	req, err := s.client.NewRequest("PUT", u, opts)
 	if err != nil {
 		return nil, err
-	}
-
-	if *opts.CNAME == "" {
-		update := new(pagesUpdateNullCNAME)
-		resp, err := s.client.Do(ctx, req, update)
-		if err != nil {
-			return resp, err
-		}
-
-		return resp, nil
 	}
 
 	resp, err := s.client.Do(ctx, req, nil)
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -75,6 +75,39 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
 	return enable, resp, nil
 }
 
+// UpdatePagesRequest sets up parameters needed to update a GitHub Pages site.
+type UpdatePagesRequest struct {
+	CNAME  *string `json:"cname,omitempty"`
+	Source *string `json:"source,omitempty"`
+}
+
+// UpdatePages updates GitHub Pages for the named repo.
+//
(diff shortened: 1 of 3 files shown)


github/repos_pages.go
```diff
@@ -75,6 +75,34 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
 	return enable, resp, nil
 }
 
+// PagesUpdate sets up parameters needed to update a GitHub Pages site.
+type PagesUpdate struct {
+	CNAME  *string `json:"cname,omitempty"`
+	Source *string `json:"source,omitempty"`
```
gmlewis: Can you please add a comment above `Source` like this:

```
// Source must include the branch name, and may optionally specify the subdirectory "/docs".
// Possible values are: "gh-pages", "master", and "master /docs".
```

github/repos_pages.go
```diff
@@ -75,6 +75,34 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
 	return enable, resp, nil
 }
 
+// PagesUpdate sets up parameters needed to update a GitHub Pages site.
+type PagesUpdate struct {
+	CNAME  *string `json:"cname,omitempty"`
```
gmlewis: Can you please add a comment above `CNAME` like this:

```
// CNAME represents a custom domain for the repository.
// Leaving CNAME empty will remove the custom domain.
```

github/repos_pages.go
```diff
@@ -75,6 +75,34 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
...
+// GitHub API docs: https://developer.github.com/v3/repos/pages/#update-information-about-a-pages-site
+func (s *RepositoriesService) UpdatePages(ctx context.Context, owner, repo string, opts *PagesUpdate) (*Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pages", owner, repo)
+
+	req, err := s.client.NewRequest("PUT", u, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Accept", mediaTypeEnablePagesAPIPreview)
```
gmlewis: If I'm reading the GitHub API docs correctly, this custom preview header is not needed for this endpoint, so let's please remove it.

github/repos_pages.go
```diff
@@ -75,6 +75,34 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
...
+
+	update := new(PagesUpdate)
```
gmlewis: No return value is desired, so please remove this line.

github/repos_pages.go
```diff
@@ -75,6 +75,34 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
...
+	resp, err := s.client.Do(ctx, req, update)
```
gmlewis: s/update/nil/

github/repos_pages_test.go
```diff
@@ -52,6 +52,35 @@ func TestRepositoriesService_EnablePages(t *testing.T) {
...
+		CNAME:  String("www.my-domain.com"),
+		Source: String("gh-pages"),
+	}
+
+	mux.HandleFunc("/repos/o/r/pages", func(w http.ResponseWriter, r *http.Request) {
+		v := new(PagesUpdate)
+		json.NewDecoder(r.Body).Decode(v)
+
+		testMethod(t, r, "PUT")
+		testHeader(t, r, "Accept", mediaTypeEnablePagesAPIPreview)
```
gmlewis: This line can be removed.

github/repos_pages.go
patrickmarabeas: Sorry, wasn't expecting such a prompt response, was still poking at it.

Thoughts on the secondary type I've since added so this can remain optional without removing the custom domain?

github/repos_pages.go
```diff
@@ -75,6 +75,53 @@ func (s *RepositoriesService) EnablePages(ctx context.Context, owner, repo strin
...
+// GitHub API docs: https://developer.github.com/v3/repos/pages/#update-information-about-a-pages-site
+func (s *RepositoriesService) UpdatePages(ctx context.Context, owner, repo string, opts *PagesUpdate) (*Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/pages", owner, repo)
+
+	req, err := s.client.NewRequest("PUT", u, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	if *opts.CNAME == "" {
```
gmlewis: Let's delete this block completely (lines 107-116) and remove the `pagesUpdateNullCNAME` and then simply remove the `,omitempty` on line 82.

That way, if `CNAME` is defined, it is used, but if `CNAME` is empty, `null` will be sent and the custom domain will be removed.

github/repos_pages_test.go
```diff
@@ -52,6 +52,34 @@ func TestRepositoriesService_EnablePages(t *testing.T) {
 	}
 }
 
+func TestRepositoriesService_UpdatePages(t *testing.T) {
```
gmlewis: You could add a second test proving that when `CNAME` is empty, `null` is truly passed.
An example like that is `TestGitService_CreateTree_Delete` in `git_trees_test.go`.

github/repos_pages.go
patrickmarabeas: Sure.

github/repos_pages_test.go
patrickmarabeas: Added test.

github/repos_pages_test.go
```diff
@@ -52,6 +52,61 @@ func TestRepositoriesService_EnablePages(t *testing.T) {
...
+		Source: String("gh-pages"),
+	}
+
+	mux.HandleFunc("/repos/o/r/pages", func(w http.ResponseWriter, r *http.Request) {
+		v := new(PagesUpdate)
+		json.NewDecoder(r.Body).Decode(v)
+
+		testMethod(t, r, "PUT")
+		want := &PagesUpdate{CNAME: nil, Source: String("gh-pages")}
+		if !reflect.DeepEqual(v, want) {
```
gmlewis: Unfortunately, this doesn't test that a `null` was actually sent via JSON.

Here's the example I was talking about:

```
		got, err := ioutil.ReadAll(r.Body)
		if err != nil {
			t.Fatalf("unable to read body: %v", err)
		}

		testMethod(t, r, "POST")

		want := []byte(`{"base_tree":"b","tree":[{"sha":null,"path":"content.md","mode":"100644"}]}` + "\n")
		if !bytes.Equal(got, want) {
			t.Errorf("Git.CreateTree request body: %s, want %s", got, want)
		}
```

patrickmarabeas: Ah, right.

gmlewis:
Thank you, @wesleimp! Merging.

Issue #1454, opened 2020-03-07
Support new "filter" parameter for list workflow jobs endpoint


nightlark:
GitHub Developer API announcement:
https://developer.github.com/changes/2020-03-06-filtering-jobs-for-a-workflow-run/

A new optional `filter` parameter was added to the endpoint that [lists jobs for a workflow run](https://developer.github.com/v3/actions/workflow_jobs/#list-jobs-for-a-workflow-run)

PR #1455, opened 2020-03-07
Add support for filter parameter to the list workflow jobs endpoint

nightlark:
This adds support for an optional `filter` parameter to the list workflow jobs endpoint that was recently announced. See #1454.

### DIFF
--- a/github/actions_workflow_jobs.go
+++ b/github/actions_workflow_jobs.go
@@ -46,10 +46,16 @@ type Jobs struct {
 	Jobs       []*WorkflowJob `json:"jobs,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


github/actions_workflow_jobs.go
```diff
@@ -46,10 +46,22 @@ type Jobs struct {
 	Jobs       []*WorkflowJob `json:"jobs,omitempty"`
 }
 
+// ListWorkflowJobsOptions specifies optional parameters to ListWorkflowJobs
```
gmlewis: Please add a period to the end of the docstring comment.

nightlark:
Would it be preferred to replace the existing test with one that uses the filter parameter, instead of making a new test?

Issue #1458, opened 2020-03-16
Missing `AuthorAssociation` field in issue struct


you06:
I found that there is already `author_association` field in issues API, but the [Issue struct](https://github.com/google/go-github/blob/6e0f6ebdef7d6db18d0eb92bb6f7aa9c0c7d4101/github/issues.go#L28-L63) is not updated. 

Also the example in document doesn't has this field yet https://developer.github.com/v3/issues/#get-an-issue. However it's available in API v4 https://developer.github.com/v4/object/issue/ .

vikkyomkar:
Let me look into it

gmlewis:
The way we typically handle this situation is to first contact support@github.com and ask them to please update their documentation for the API.  They have always been super helpful, friendly, and responsive, so please do this, @you06.

This does two things:
* it validates that indeed the field is being returned and we didn't make a mistake on our end
* improves the user experience for all GitHub v3 API users (regardless of programming language) by helping GitHub to improve their documentation.

As for #1460, I will comment on it directly.

Thank you!

you06:
I've sent an email to GitHub team about this. If there are any updates, I'll comment here.

PR #1460, opened 2020-03-16
Added AuthorAssociation field in issue struct

vikkyomkar:
This PR adds a new field "AuthorAssociation" in issue struct and removes custom media type for "get an issue" API as this API is fully functional now.

https://developer.github.com/v3/issues/#get-an-issue


vikkyomkar:
#1458

nightlark:
It looks like the test checking if `go generate github.com/google/go-github/...` needs to get run is failing.

github/issues.go
```diff
@@ -250,10 +251,6 @@ func (s *IssuesService) Get(ctx context.Context, owner string, repo string, numb
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept headers when APIs fully launch.
```
gmlewis: Why are these custom accept headers being removed?
That seems to be an unrelated issue to #1458 and should likely be handled in a separate PR.

Also, the URL in the comment for this method needs updating, if you wouldn't mind:
`https://developer.github.com/v3/issues/#get-an-issue`

But again, maybe a separate PR would be more appropriate and keep this PR focused on the missing field as its description states, so please revert this change.

PR #1462, opened 2020-03-17
Added AuthorAssociation field in issue struct

### LINKED ISSUE #1458: Missing `AuthorAssociation` field in issue struct

vikkyomkar:
This PR fixes
#1458


github/issues.go
```diff
@@ -26,32 +26,33 @@ type IssuesService service
...
+	State             *string           `json:"state,omitempty"`
+	Locked            *bool             `json:"locked,omitempty"`
+	Title             *string           `json:"title,omitempty"`
+	Body              *string           `json:"body,omitempty"`
+	User              *User             `json:"user,omitempty"`
+	Labels            []*Label          `json:"labels,omitempty"`
+	Assignee          *User             `json:"assignee,omitempty"`
+	Comments          *int              `json:"comments,omitempty"`
+	ClosedAt          *time.Time        `json:"closed_at,omitempty"`
+	AuthorAssociation *string           `json:"author_association,omitempty"`
```
gmlewis: Instead of splitting up three `*time.Time` fields, let's please move this new field up between lines 34 and 35 immediately before the first struct (`User`) and after the `Body *string`.
Thanks.

vikkyomkar: Updated as per the suggestion

gmlewis:
Can you please re-run `go generate ./...` to fix the tests that are failing?

Issue #1465, opened 2020-03-20
Removal of custom media type from Get an issue API


vikkyomkar:
https://developer.github.com/v3/issues/#get-an-issue

https://github.com/google/go-github/blob/master/github/issues.go#L254-L256

gmlewis:
Closing inactive issue for lack of activity/interest.  Feel free to reopen.

PR #1466, opened 2020-03-20
Removed  custom media type from get issue API function

vikkyomkar:
#1465

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -251,10 +251,6 @@ func (s *IssuesService) Get(ctx context.Context, owner string, repo string, numb
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeReactionsPreview, mediaTypeLockReasonPreview}
-	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
-
 	issue := new(Issue)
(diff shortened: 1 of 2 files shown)


github/issues.go
```diff
@@ -251,10 +251,6 @@ func (s *IssuesService) Get(ctx context.Context, owner string, repo string, numb
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeReactionsPreview, mediaTypeLockReasonPreview}
```
gmlewis: Sorry, this is not correct.  If you scroll down a bit further, you will see:
https://developer.github.com/v3/issues/#reactions-summary-4
which says that `application/vnd.github.squirrel-girl-preview` (aka `mediaTypeReactionsPreview`) is still needed.

Granted, `mediaTypeLockReasonPreview` can now be removed for `ListByOrg` and `Get`, but not for `Lock`, which is why I didn't close this PR.

If you want to update this PR to remove just one of the preview headers, feel free to. Obviously, you would not need a slice of strings anymore nor would you need the `strings.Join`. There are many other examples of how we add just a single custom accept header if you would like to take a look.

Thanks.

gmlewis:
Closing abandoned PR.

Issue #1472, opened 2020-03-25
func (*ActionsService) ListRunners always returns error


summerwind:
[func (*ActionsService) ListRunners](https://godoc.org/github.com/google/go-github/github#ActionsService.ListRunners) always returns error due to the GitHub API response format was changed.

**Previous response format**

```
[
  {
    "id": 23,
    "name": "MBP",
    "os": "macos",
    "status": "online"
  },
  {
    "id": 24,
    "name": "iMac",
    "os": "macos",
    "status": "offline"
  }
]
```

**New response format**

```
{
  "total_count": 2,
  "runners": [
    {
      "id": 23,
      "name": "MBP",
      "os": "macos",
      "status": "online"
    },
    {
      "id": 24,
      "name": "iMac",
      "os": "macos",
      "status": "offline"
    }
  ]
}
```

gmlewis:
Ah, yes!  Thank you, @summerwind!
I'll work on a fix.

summerwind:
Thanks for the quick fix!
Actions API support on go-github is really useful for me. Thank you!

PR #1473, opened 2020-03-25
Update ListRunners with new response type - breaking change

gmlewis:
Fixes: #1472.

Note to self: once this is merged, the next release to be built will need a semver major version bump, as this is a breaking change.

### DIFF
--- a/github/actions_runners.go
+++ b/github/actions_runners.go
@@ -71,10 +71,16 @@ type Runner struct {
 	Status *string `json:"status,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


gmlewis:
Awaiting two LGTMs before merging.

gmlewis:
Thank you, @wesleimp and @nightlark !
Merging.

Note that I will hold off on creating `v31.0.0` and making a release until someone asks for it.
Meanwhile, people can refer to this version by commit ref in their `go.mod` if they need this change immediately.

PR #1477, opened 2020-03-30
Create tool to update and validate GitHub v3 API URLs

### LINKED ISSUE #1476: Go generate should validate and fix all documentation URLs
Here's a wild and wacky idea. ~that is certainly non-trivial to get right and not for the faint-of-heart nor is it good for someone who wants to get quick brownie points for contributing to an open source project. :smile:~ Edit: I decided to take it on myself.

After merging #1475, it dawned on me that URLs occasionally change in the GitHub v3 API documentation, and it would be amazing and beautiful if `go generate ./...` could go through and not only validate that all the URLs are correct (including the [fragment identifier](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-the-fragment-identifier)), but also fix them when they are not.

A URL is deemed correct only when it truly matches the endpoint in question (which is the difficult part of solving this issue, especially when "fixing" a broken one).

For example, this repo consistently (for the most part) generates a URL for the corresponding API endpoint, then shortly afterward makes an HTTP request using a particular "method" such as `GET`, `PATCH`, `POST`, `PUT`, and `DELETE`. I envision this tool to find the URL path and associated method, and then use this information to find the correct full URL in the GitHub v3 API documentation.

In my experience from maintaining and enhancing the [Google Go API auto-generator](https://github.com/googleapis/google-api-go-client), I'm betting that this tool may require the use of a whitelist and/or blacklist for overriding exceptional cases that would otherwise be challenging to determine.

~I'm throwing out this issue for discussion or in case someone feels like they are up for a challenge and is as excited as I am about solving this issue. I have an idea how I would solve it, but will refrain for awhile and see if anyone else wants to take this on first. If nobody picks it up, I'll assign it to myself and work on it.~ Edit: I decided to take it on myself.

gmlewis:
Fixes #1476.

Note that this ~is a breaking API change because it~ found receivers that were being passed by value instead of by reference, and these have been fixed and pulled out into a separate PR #1487.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+*.sh
 *.test
 coverage.out
--- a/github/actions_workflow_jobs.go
+++ b/github/actions_workflow_jobs.go
@@ -84,7 +84,7 @@ func (s *ActionsService) ListWorkflowJobs(ctx context.Context, owner, repo strin
 
 // GetWorkflowJobByID gets a specific job in a workflow run by ID.
 //
-// GitHub API docs: https://developer.github.com/v3/actions/workflow_jobs/#list-jobs-for-a-workflow-run
+// GitHub API docs: https://developer.github.com/v3/actions/workflow_jobs/#get-a-workflow-job
 func (s *ActionsService) GetWorkflowJobByID(ctx context.Context, owner, repo string, jobID int64) (*WorkflowJob, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/jobs/%v", owner, repo, jobID)
 
--- a/github/activity_events.go
+++ b/github/activity_events.go
@@ -107,7 +107,7 @@ func (s *ActivityService) ListEventsForRepoNetwork(ctx context.Context, owner, r
 
 // ListEventsForOrganization lists public events for an organization.
 //
-// GitHub API docs: https://developer.github.com/v3/activity/events/#list-public-events-for-an-organization
+// GitHub API docs: https://developer.github.com/v3/activity/events/#list-public-organization-events
 func (s *ActivityService) ListEventsForOrganization(ctx context.Context, org string, opts *ListOptions) ([]*Event, *Response, error) {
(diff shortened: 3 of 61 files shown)


gmlewis:
Thank you, @wesleimp!

Since this is a non-trivial addition and affects future maintenance, I would like to also get a review from @willnorris to make sure he is on-board with this change before going ahead with it.

No rush, @willnorris - nothing is waiting on this. Thank you!

willnorris:
I haven't done a thorough review of the generation code, but it certainly looks to be pretty clever (and probably a lot of fun to write!).  But it's also rather complex and hefty (and potentially fragile), for maintaining what is a very minor aspect of our godocs.  In fact, I think I started include the GitHub doc URLs somewhat on a lark, as just an extra "nice to have".

I think the number of changes in the doc URLs certainly shows how out of date they are, so I'm totally supportive of whatever tools make it easier to keep those up to date.  I'm just worried about the ongoing maintenance of it.  What do you think of pulling it out of the main github folder, similar to what we did with the scrape package?  Maybe even make it its own go module, so it doesn't get downloaded as part of the main module when someone is just using go-github, but not actually contributing to it?

wdyt?

gmlewis:
> wdyt?

Great idea!  I like it.  Thanks, @willnorris !  I'll work on that and update this PR when I've got it moved.

gmlewis:
Tool moved to "update-urls" subdir with its own "go.mod" and "go.sum".  PTAL (low priority).
Thank you, Will!

gmlewis:
That makes sense, Will... thank you!

I'll pull those breaking changes out into a separate PR and not merge this one yet until that one is in.

gmlewis:
This PR is now on hold until #1487 has been merged, which is a prerequisite for tests to pass.

gmlewis:
This PR now contains only changes to comments and the addition of the `update-urls` tool.
Merging.

Issue #1483, opened 2020-04-08
Teams.CreateOrUpdateIDPGroupConnections and Teams.ListIDPGroupsForTeam generating requests resulting in 404s


anGie44:
github version: `v29.0.3`

While trying to create Idp Group connections for my team, I am receiving 404 errors when using the method `CreateOrUpdateIDPGroupConnections`. In the [logs](https://gist.github.com/anGie44/975a123fc361d4168e4f42588c7b54a8) attached, you can see the PATCH request is made with `PATCH /teams/test-team/team-sync/group-mappings` however according to the api docs (https://developer.github.com/v3/teams/team_sync/#create-or-update-idp-group-connections), the call is to be prefixed with `/orgs/:org` and from what I can see in `github.go` these requests do not handle prepending this prefix. As an alternative, I tried to directly hit the api endpoint in my browser using `https://api.github.com/orgs/test-org/teams/test-team/team-sync/group-mappings` and this correctly returned with an empty `groups` object as expected as my github team doesn't have any particular Identity Provider Groups set at the moment: 
```
{
  "groups": [

  ]
}
```
I expect the request generated when using `Teams.ListIDPGroupsForTeam` will result in a similar 404 as the method creates a `GET` request without prepending the `orgs/:org` prefix to the url

PR #1484, opened 2020-04-08
Update IdP team-sync list and create/update endpoints with ByID and BySlug

anGie44:
Addresses Issue #1483 
Changes include:
* Update method signatures of `#ListIDPGroupsForTeam` and `#CreateOrUpdateIDPGroupConnections`  to include the organization name identified as `org` and rename the `teamID` attribute to `slug` as it can be misleading/the api requires the slug for these particular calls
* Update respective tests with corrected arguments

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -765,10 +765,31 @@ func (s *TeamsService) ListIDPGroupsInOrganization(ctx context.Context, org stri
 	return groups, resp, nil
 }
 
-// ListIDPGroupsForTeam lists IDP groups connected to a team on GitHub.
+// ListIDPGroupsForTeamByID lists IDP groups connected to a team on GitHub
+// given organization and team IDs.
 //
 // GitHub API docs: https://developer.github.com/v3/teams/team_sync/#list-idp-groups-for-a-team
-func (s *TeamsService) ListIDPGroupsForTeam(ctx context.Context, org, slug string) (*IDPGroupList, *Response, error) {
+func (s *TeamsService) ListIDPGroupsForTeamByID(ctx context.Context, orgID, teamID int64) (*IDPGroupList, *Response, error) {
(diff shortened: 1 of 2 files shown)


anGie44:
@googlebot I signed it!

anGie44:
Hi @gmlewis, thank you for the feedback!  I'm totally onboard with supporting both endpoint versions. I'll submit a change shortly :)

gmlewis:
Thank you, @wesleimp!
Merging.

Note to self: this is a breaking API change which means that the next release must be `v31.0.0`.

Note to others: we only make new releases now on-demand which means that we will continue to address issues and update the repo without releasing until someone says "Hey, we want to use the feature in that latest PR for this huge project and we would like to please have a new release" or something to that effect.  :smile:

anGie44:
Hi @gmlewis thanks again for reviewing this PR! wanted to follow-up about having this change in a future release. in the terraform-provider-github project, I have a recent [PR](https://github.com/terraform-providers/terraform-provider-github/pull/400) to support team synchronization and this change would help make this possible :)

gmlewis:
OK, @anGie44, I have another breaking API change coming up in #1477 and it would be great to get that in at the same time so we don't have to create `v31.0.0` immediately followed with `v32.0.0` a few days later...

Let me pull in the big guns to make this decision...  hold on a sec...  :smile:

@willnorris - what do you think about getting #1477 merged before we bump the version to `v31.0.0`?  That one involves breaking API changes too because it found a bunch of receivers that were not pointers, but instead were being passed by value.

anGie44:
thank you, @gmlewis! I really appreciate the quick response :) On my end, it's not drastically time sensitive so I'm happy to wait for the other PRs to get merged for this next release.

PR #1487, opened 2020-04-09
Fix receivers and other breaking changes

gmlewis:
#1477 found a number of problems in the repo that would cause breaking API changes to fix them.

Instead of including them there, it [was decided](https://github.com/google/go-github/pull/1477#pullrequestreview-391172913) to pull the breaking API changes into a separate PR, which is this one.


gmlewis:
This is now the PR that would ideally be included before we bump the major version to `v31.0.0` along with #1484.

Awaiting LGTM before merging.

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1489, opened 2020-04-10
Prepare for release v31

gmlewis:
Due to breaking API changes in #1484 and #1487, we need to bump the major version for the next release. This PR prepares for that release.


gmlewis:
Thank you, @nightlark!
Merging and then releasing `v31.0.0`.

PR #1495, opened 2020-04-16
Update DeploymentStatus type

### LINKED ISSUE #1494: DeploymentStatus type does not contain all possible keys
I'm using `go-github` in conjunction with GitHub Actions + Deployments. 
During deployments we update the deployment status and provide a `log_url` value that points back to the current GitHub actions URL.
We then use `github.Client.Repositories.ListDeploymentStatuses` to retrieve the statuses to retrieve the log URL.

There are a few issues that I can see:
1. The `ListDeploymentStatuses` method doesn't contain preview headers. (though `GetDeploymentStatus` does contain the headers)
2. The `DeploymentStatus` type doesn't have all available fields so if you provide the headers it won't parse them anyway.
3. The response body seems to be closed before its returned to the requester so we can't manually parse the raw JSON response.

jrossiter:
Add preview keys to the DeploymentStatus type. 
Include preview headers in ListDeploymentStatuses request.

Fixes #1494.

### DIFF
--- a/github/repos_deployments.go
+++ b/github/repos_deployments.go
@@ -139,16 +139,16 @@ type DeploymentStatus struct {
 	State          *string    `json:"state,omitempty"`
 	Creator        *User      `json:"creator,omitempty"`
 	Description    *string    `json:"description,omitempty"`
-	TargetURL      *string    `json:"target_url,omitempty"`
+	Environment    *string    `json:"environment,omitempty"`
+	NodeID         *string    `json:"node_id,omitempty"`
 	CreatedAt      *Timestamp `json:"created_at,omitempty"`
 	UpdatedAt      *Timestamp `json:"updated_at,omitempty"`
+	TargetURL      *string    `json:"target_url,omitempty"`
 	DeploymentURL  *string    `json:"deployment_url,omitempty"`
 	RepositoryURL  *string    `json:"repository_url,omitempty"`
 	EnvironmentURL *string    `json:"environment_url,omitempty"`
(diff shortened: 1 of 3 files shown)


github/repos_deployments.go
```diff
@@ -134,16 +134,21 @@ func (s *RepositoriesService) CreateDeployment(ctx context.Context, owner, repo
...
+	TargetURL      *string    `json:"target_url,omitempty"`
+	CreatedAt      *Timestamp `json:"created_at,omitempty"`
+	UpdatedAt      *Timestamp `json:"updated_at,omitempty"`
+	DeploymentURL  *string    `json:"deployment_url,omitempty"`
+	RepositoryURL  *string    `json:"repository_url,omitempty"`
+	EnvironmentURL *string    `json:"environment_url,omitempty"`
+	LogURL         *string    `json:"log_url,omitempty"`
+	NodeID         *string    `json:"node_id,omitempty"`
+	URL            *string    `json:"url,omitempty"`
+	Environment    *string    `json:"environment,omitempty"`
```
gmlewis, line 151: Would you mind moving `Environment` up immediately after `Description`, followed by `NodeID`?
That way, all the URLs are together, and Creator, Description, Environment, and NodeID are at least in sorted order.

jrossiter: Done. I moved `TargetURL` down so it's grouped with the other URLs as well.

gmlewis:
Thank you, @wesleimp!
Merging.

PR #1498, opened 2020-04-20
Update ListRepositoryWorkflowRuns to support all query parameters

### LINKED ISSUE #1497: ListRepositoryWorkflowRuns doesn't support all query parameters
It appears that the ActionsService.ListRepositoryWorkflowRuns function doesn't currently support passing in all of the available query parameters for the actual API call, because it's currently using the generic `ListOptions` struct. The other two ListWorkflowRuns* functions are using a `ListWorkflowRunsOptions` struct that has all of the supported query parameters listed.

According to [the docs](https://developer.github.com/v3/actions/workflow_runs/#list-repository-workflow-runs) it appears all of the options in `ListWorkflowRunsOptions` are supported by the ListRepositoryWorkflowRuns endpoint. My thinking is that this was either missed in the original implementation, or that Github added support for the additional parameters recently - either way, is there a reason that the function can't be updated to use `ListWorkflowRunsOptions`?

acburdine:
closes #1497 
- update the method to use the more specific ListWorkflowRunsOptions struct

Let me know if I need to update any of the tests as well - I looked through the existing tests and didn't see any that were specifically testing the unique query params of workflow endpoints.

### DIFF
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -93,7 +93,7 @@ func (s *ActionsService) ListWorkflowRunsByFileName(ctx context.Context, owner, 
 // ListRepositoryWorkflowRuns lists all workflow runs for a repository.
 //
 // GitHub API docs: https://developer.github.com/v3/actions/workflow_runs/#list-repository-workflow-runs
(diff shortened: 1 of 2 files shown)


gmlewis:
Thank you, @wesleimp!
Merging.

PR #1506, opened 2020-04-24
Support self-hosted organization runners

### LINKED ISSUE #1505: Support organization runners API
Self-hosted runners on the organization level are now available: https://github.blog/changelog/2020-04-22-github-actions-organization-level-self-hosted-runners/

It would be great if this could be supported. I already have made an implementation somewhere else, but if you are willing to accept I could contribute the code to here as well

reiniertimmer:
Closes #1505 

This adds support for the organization-wide self-hosted runners API

The API for self-hosted runners on organization level is almost similar to the ones for repositories, only the the URL path is slightly different, and you need an organization owner token for the new functions to work.

Because of the similarities, this was basically a large copy & paste operation, so for each function there is now an `Organization` counterpart. This is not really DRY, but it makes it more explicit what API is being used.

(An alternative solution would be to have a switch inside each function to decide to use the organization or repository API based on the presence of the repo parameter. I have made that as well on a separate branch. The implementation is much smaller, but the organization functionality is a little more hidden there. If there's any votes for that, I could offer that implementation instead)

### DIFF
--- a/github/actions_runners.go
+++ b/github/actions_runners.go
@@ -20,7 +20,7 @@ type RunnerApplicationDownload struct {
 
 // ListRunnerApplicationDownloads lists self-hosted runner application binaries that can be downloaded and run.
 //
-// GitHub API docs: https://developer.github.com/v3/actions/self_hosted_runners/#list-downloads-for-the-self-hosted-runner-application
+// GitHub API docs: https://developer.github.com/v3/actions/self-hosted-runners/#list-runner-applications-for-a-repository
 func (s *ActionsService) ListRunnerApplicationDownloads(ctx context.Context, owner, repo string) ([]*RunnerApplicationDownload, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/runners/downloads", owner, repo)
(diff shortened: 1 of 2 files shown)


reiniertimmer:
> 📝 **Please visit https://cla.developers.google.com/ to sign.**

Done

reiniertimmer:
@googlebot I signed it!

reiniertimmer:
@gmlewis @wesleimp thanks for your quick responses 

However, if I understand the codecov/project checks, the coverage went down (0.15%). Seeing as this was mostly a copy-paste operation, I also adapted most of the unit testing code. So, quality wise it should be at the same level as the original code. But if it's necessary, I could always attempt to up the line coverage a bit (not a bad thing)

The codecov/patch check I don't fully understand. What should I do to resolve this?

gmlewis:
Thank you, @reiniertimmer , the `codecov/*` results are informational in nature and not blockers.
(When I click on "Details" I get 404, so we can safely proceed.)

gmlewis:
Thank you, @wesleimp!
Merging.

PR #1520, opened 2020-05-08
fix release tag url

smiller171:
While trying to use https://github.com/terraform-providers/terraform-provider-github I stumbled across a bug in this package in building the release tag url. This PR fixes that bug.


smiller171:
@googlebot I signed it!

smiller171:
ugh...after doing some more research it seems like some repos have `.../releases/tags/...` and others have `.../releases/tag/...` and I can't find any documentation on the difference.

gmlewis:
According to the official documentation: https://developer.github.com/v3/repos/releases/#get-a-release-by-tag-name
the correct string is `tags` which we currently have.

If you are finding a bug, please contact support@github.com and give them a reproducible test case so that they can investigate and fix the matter.

Then please report back here with their response and your findings.

Thank you.

smiller171:
@gmlewis Yeah, this looks like it's actually some form of GitHub bug, which is incredibly annoying.

gmlewis:
Closing PR based on feedback.

PR #1530, opened 2020-05-20
Enable BranchRestriction fields to be empty

michelangelomo:
## Description
I've encountered an issue trying to update BranchRestriction on a personal repository. Github doesn't allow `teams` and `users` fields in the request payload.

## How to reproduce
```go
p := github.ProtectionRequest{
	Restrictions: &github.BranchRestrictionsRequest{
		Apps: []string{"myapp"},
	},
}
client.Repositories.UpdateBranchProtection(ctx, org, repo, branch, p)
```
will return
```
For 'anyOf/1', {"users"=>nil, "teams"=>nil, "apps"=>["myapp"]} is not a null.
```
but if I set `Teams` and `Users` fields as `[]string{}` Github returns `422 Validation Failed [{Resource: Field: Code: Message:Only organization repositories can have users and team restrictions}]` as described above.

## Fix
Allow `Teams` and `Users` fields to be empty by using `omitempty`.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -864,9 +864,9 @@ type BranchRestrictions struct {
 // different from the response structure.
 type BranchRestrictionsRequest struct {
 	// The list of user logins with push access. (Required; use []string{} instead of nil for empty list.)
-	Users []string `json:"users"`
+	Users []string `json:"users,omitempty"`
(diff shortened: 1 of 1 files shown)


michelangelomo:
@googlebot I signed it!

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1536, opened 2020-05-23
Add ListCommit to the sdk

kespinola:
### Changes
- Add `ListCommit` function to `Git` scope

### Background
The commits endpoint returns list of all commits for a sha when the sha is passed as a query param instead of a path param.

### DIFF
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -49,6 +49,9 @@ type Commit struct {
 	SigningKey *openpgp.Entity `json:"-"`
 }
 
+// CommitList is a list of commits
+type CommitList []*Commit
+
 func (c Commit) String() string {
 	return Stringify(c)
 }
@@ -85,6 +88,29 @@ func (s *GitService) GetCommit(ctx context.Context, owner string, repo string, s
 	}
 
 	return c, resp, nil
+}
+
+// ListCommits fetches latest commits
+func (s *GitService) ListCommits(ctx context.Context, owner string, repo string, sha string) (*CommitList, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/git/commits", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	q := req.URL.Query()
+	q.Add("sha", sha)
+	req.URL.RawQuery = q.Encode()
+
+	commits := new(CommitList)
+	resp, err := s.client.Do(ctx, req, commits)
+
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return commits, resp, nil
 }
 
 // createCommit represents the body of a CreateCommit request.
--- a/github/git_commits_test.go
+++ b/github/git_commits_test.go
@@ -146,6 +146,37 @@ func TestGitService_GetCommit_invalidOwner(t *testing.T) {
 	testURLParseError(t, err)
 }
 
+func TestGitService_ListCommits(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/repos/o/r/git/commits", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+
+		_, ok := r.URL.Query()["sha"]
+
+		if !ok {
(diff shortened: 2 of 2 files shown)


github/git_commits.go
```diff
@@ -87,6 +90,29 @@ func (s *GitService) GetCommit(ctx context.Context, owner string, repo string, s
 	return c, resp, nil
 }
 
+// ListCommits fetches latest commits
+func (s *GitService) ListCommits(ctx context.Context, owner string, repo string, sha string) (*CommitList, *Response, error) {
```
gmlewis, line 94: A pointer to a slice is generally to be avoided, as it is already a reference type.

Let's please change this return type to:

```
([]*Commit, *Response, error)
```

And delete lines 52-54 above.

Additionally, it is not idiomatic Go to say `owner string, repo string, sha string` in the parameter list.
Instead, please change this to `owner, repo, sha string`.

kespinola:
@googlebot I signed it!

github/git_commits.go
```diff
@@ -87,6 +90,29 @@ func (s *GitService) GetCommit(ctx context.Context, owner string, repo string, s
 	return c, resp, nil
 }
 
+// ListCommits fetches latest commits
```
gmlewis, line 93: Please make this a complete sentence (ending with a period) for the sake of the auto-generated godocs.

Please also comment on the behavior of the `sha` parameter.  What happens if it is `""`?
Does it list commits _after_ the SHA?  Does it include the SHA in the list?
Inquiring minds want to know!  :smile:

Also, please send a request to GitHub developer tech support at support@github.com and request them to document this endpoint.  In the meantime, we can proceed with this PR, but when they respond, we will want to add a link to the official GitHub API docs like the other endpoints in this repo. It could either be in this PR or in a follow-on PR.  Usually they are incredibly fast in responding to requests like this.  Thank you!

github/git_commits.go
```diff
@@ -87,6 +90,29 @@ func (s *GitService) GetCommit(ctx context.Context, owner string, repo string, s
...
+func (s *GitService) ListCommits(ctx context.Context, owner string, repo string, sha string) (*CommitList, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/git/commits", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+
```
gmlewis, line 97: Please delete this line to closely bind the `if err != nil` test to the statement it is checking as per idiomatic Go.

github/git_commits.go
```diff
@@ -87,6 +90,29 @@ func (s *GitService) GetCommit(ctx context.Context, owner string, repo string, s
...
+		return nil, nil, err
+	}
+
+	q := req.URL.Query()
+	q.Add("sha", sha)
+	req.URL.RawQuery = q.Encode()
+
+	commits := new(CommitList)
+	resp, err := s.client.Do(ctx, req, commits)
+
```
gmlewis, line 108: Please delete this line to closely bind the `if err != nil` test to the statement it is checking as per idiomatic Go.

github/git_commits.go
```diff
@@ -87,6 +90,29 @@ func (s *GitService) GetCommit(ctx context.Context, owner string, repo string, s
...
+	u := fmt.Sprintf("repos/%v/%v/git/commits", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	q := req.URL.Query()
+	q.Add("sha", sha)
+	req.URL.RawQuery = q.Encode()
```
gmlewis, line 104: Should `sha` not be added if it is the empty string?

Why aren't you using `addOptions` like are used throughout the rest of this repo?  If you didn't know about them, please take a look and replace these 3 lines with a call to `addOptions` between lines 95 and 96 above.

kespinola:
I appreciate the review. On further exploration of the repo the endpoint is already accounted for in repo_commits. 
https://github.com/google/go-github/blob/master/github/repos_commits.go#L127

Closing PR. Thank you for the timely response.

gmlewis:
Whups!  I took a quick look, but didn't see it. (Apparently _**too**_ quick! :smile:)
Thank you for checking, @kespinola !

PR #1542, opened 2020-05-30
Return http.Response from DownloadContents

### LINKED ISSUE #1531: RepositoriesService.DownloadContents should handle response codes
To download the contents of a file the `DownloadContents` method is using a `http.Client` to request it and then returns only the body of the response.
```
			resp, err := s.client.client.Get(*contents.DownloadURL)
			if err != nil {
				return nil, err
			}
			return resp.Body, nil
```

The `http.Client` docs state that `A non-2xx response doesn't cause an error`. If the download request returns a failed response code then `resp.Body` most likely doesn't contain the content you wanted and there is no way to tell that something went wrong.

I'd suggest using `github.CheckResponse` on the response or perhaps a simpler 2XX check since `raw.githubusercontent.com` may have different behavior than the api.

mctofu:
This is the breaking change we discussed at: https://github.com/google/go-github/issues/1531#issuecomment-632666511

Fixes #1531

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -117,26 +117,30 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
 // specified file. This function will work with files of any size, as opposed
 // to GetContents which is limited to 1 Mb files. It is the caller's
 // responsibility to close the ReadCloser.
-func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (io.ReadCloser, error) {
+//
+// It is possible for the download to result in a failed response. Callers
+// should check the returned http.Response status code to verify the content
+// is from a successful response.
+func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (io.ReadCloser, *http.Response, error) {
 	dir := path.Dir(filepath)
 	filename := path.Base(filepath)
 	_, dirContents, _, err := s.GetContents(ctx, owner, repo, dir, opts)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	for _, contents := range dirContents {
 		if *contents.Name == filename {
 			if contents.DownloadURL == nil || *contents.DownloadURL == "" {
-				return nil, fmt.Errorf("No download link found for %s", filepath)
+				return nil, nil, fmt.Errorf("No download link found for %s", filepath)
 			}
 			resp, err := s.client.client.Get(*contents.DownloadURL)
 			if err != nil {
-				return nil, err
+				return nil, resp, err
 			}
-			return resp.Body, nil
+			return resp.Body, resp, nil
 		}
 	}
-	return nil, fmt.Errorf("No file named %s found in %s", filename, dir)
+	return nil, nil, fmt.Errorf("No file named %s found in %s", filename, dir)
 }
 
 // GetContents can return either the metadata and content of a single file
--- a/github/repos_contents_test.go
+++ b/github/repos_contents_test.go
@@ -118,9 +118,13 @@ func TestRepositoriesService_DownloadContents_Success(t *testing.T) {
 		fmt.Fprint(w, "foo")
 	})
 
-	r, err := client.Repositories.DownloadContents(context.Background(), "o", "r", "d/f", nil)
+	r, resp, err := client.Repositories.DownloadContents(context.Background(), "o", "r", "d/f", nil)
 	if err != nil {
 		t.Errorf("Repositories.DownloadContents returned error: %v", err)
+	}
+
+	if got, want := resp.StatusCode, http.StatusOK; got != want {
+		t.Errorf("Repositories.DownloadContents returned status code %v, want %v", got, want)
 	}
 
 	bytes, err := ioutil.ReadAll(r)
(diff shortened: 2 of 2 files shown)


github/repos_contents.go
```diff
@@ -117,26 +117,30 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
...
+// It is possible for the download to result in a failed response. Callers
+// should check the returned http.Response status code to verify the content
+// is from a successful response.
+func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (io.ReadCloser, *http.Response, error) {
 	dir := path.Dir(filepath)
 	filename := path.Base(filepath)
 	_, dirContents, _, err := s.GetContents(ctx, owner, repo, dir, opts)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
```
gmlewis, line 129: Since `s.GetContents` returns the response with its third return value, let's go ahead and return it here.

```
_, dirContents, resp, err := s.GetContents(ctx, owner, repo, dir, opts)
if err != nil {
  return nil, resp, err
```

github/repos_contents.go
```diff
@@ -117,26 +117,30 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
...
 	}
 	for _, contents := range dirContents {
 		if *contents.Name == filename {
 			if contents.DownloadURL == nil || *contents.DownloadURL == "" {
-				return nil, fmt.Errorf("No download link found for %s", filepath)
+				return nil, nil, fmt.Errorf("No download link found for %s", filepath)
```
gmlewis, line 134: Same here...
`return nil, resp, fmt.Errorf...`

github/repos_contents.go
```diff
@@ -117,26 +117,30 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
...
 			if err != nil {
-				return nil, err
+				return nil, resp, err
 			}
-			return resp.Body, nil
+			return resp.Body, resp, nil
 		}
 	}
-	return nil, fmt.Errorf("No file named %s found in %s", filename, dir)
+	return nil, nil, fmt.Errorf("No file named %s found in %s", filename, dir)
```
gmlewis, line 143: Same here.
`return nil, resp, fmt.Errorf...`

github/repos_contents.go
mctofu: The response from `s.GetContents` is of type `*github.Response` rather than `*http.Response`. Do you think there is any value in returning just the `http.Response` from the `github.Response`?

We could also consider adding a 4th value to the return to allow the caller to check the rate limits from `s.GetContents` regardless of other errors?

gmlewis: Ah!  Gotcha.  Hmmm... the vast majority of calls in this repo return `*github.Response`... is it a simple matter to do the same for this endpoint?  That would be ideal so that this endpoint would not be a glaring exception.
I'm thinking you can wrap the regular `*http.Response` just by calling `newResponse` (around line 438 in github.go) and returning the result.

mctofu: Sure, that looks doable. Would it make sense to return the `*github.Response` from the `s.GetContents` call with its `*http.Response` swapped out with the one from requesting the download url? I'm not very familiar with what callers are expecting from the `github.Response` but that would at least expose the rate limit information from `s.GetContents` as part of the return value.

This also makes me think it'd be useful to have a method like `RepositoriesService.GetDownloadURL(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (string, *github.Response, error)` that just returns the URL and leaves it up to the caller to make a request for the contents.

gmlewis: Hmmm... I would return the most recent response because the point of this PR is to return the most recent response code... so I don't think we want to return anything but the response from the most recent http call, if that makes sense.

github/repos_contents.go
```diff
@@ -118,29 +118,29 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
 // to GetContents which is limited to 1 Mb files. It is the caller's
 // responsibility to close the ReadCloser.
 //
-// It is possible for the download to result in a failed response. Callers
-// should check the returned http.Response status code to verify the content
-// is from a successful response.
-func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (io.ReadCloser, *http.Response, error) {
+// It is possible for the download to result in a failed response when the
+// returned error is nil. Callers should check the returned http.Response
```
gmlewis, line 122: `http.Response` => `Response`

github/repos_contents.go
mctofu: Ok, made the change. I worry a little that this looks consistent by returning `github.Response` but may not behave consistently. The call to get the download url contents isn't an api call so we aren't able to populate any additional fields on the `github.Response` like the remaining rate limits.

gmlewis: Right, but I think this is probably the best option, so let's try it out.  If it causes problems in practice, we can make another breaking API change later.  :smile:

But honestly, I like this best because the other calls that might fail could also be caused by a rate limit or some other issue that we are now exposing back to the caller.

gmlewis:
Ping @wesleimp - thank you!

PR #1545, opened 2020-05-30
Add support for Code Scanning API

nightlark:
Adds a Code Scanning service with functions for getting code scanning alerts.

Docs: https://developer.github.com/v3/code-scanning/

### DIFF
--- a/github/code-scanning.go
+++ b/github/code-scanning.go
@@ -8,7 +8,6 @@ package github
 import (
 	"context"
 	"fmt"
-	"time"
 )
 
 // CodeScanningService handles communication with the code scanning related
@@ -22,10 +21,10 @@ type Alert struct {
 	RuleSeverity    *string    `json:"rule_severity,omitempty"`
 	RuleDescription *string    `json:"rule_description,omitempty"`
 	Tool            *string    `json:"tool,omitempty"`
-	CreatedAt       *time.Time `json:"created_at,omitempty"`
+	CreatedAt       *Timestamp `json:"created_at,omitempty"`
 	Open            *bool      `json:"open,omitempty"`
 	ClosedBy        *string    `json:"closed_by,omitempty"`
-	ClosedAt        *time.Time `json:"closed_at,omitempty"`
+	ClosedAt        *Timestamp `json:"closed_at,omitempty"`
 	URL             *string    `json:"url,omitempty"`
 	HTMLURL         *string    `json:"html_url,omitempty"`
 }
--- a/github/code-scanning.go
+++ b/github/code-scanning.go
@@ -8,6 +8,8 @@ package github
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 )
 
 // CodeScanningService handles communication with the code scanning related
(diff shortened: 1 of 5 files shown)


github/code-scanning.go
```diff
@@ -0,0 +1,94 @@
...
+type CodeScanningService service
+
+type Alert struct {
+	RuleID          *string    `json:"rule_id,omitempty"`
+	RuleSeverity    *string    `json:"rule_severity,omitempty"`
+	RuleDescription *string    `json:"rule_description,omitempty"`
+	Tool            *string    `json:"tool,omitempty"`
+	CreatedAt       *time.Time `json:"created_at,omitempty"`
+	Open            *bool      `json:"open,omitempty"`
+	ClosedBy        *string    `json:"closed_by,omitempty"`
```
gmlewis, line 27: Might `ClosedBy` be a `*User` ?  Maybe you could ask support@github.com to find out?

github/code-scanning.go
```diff
@@ -0,0 +1,94 @@
...
+//
+// GitHub API docs: https://developer.github.com/v3/code-scanning/
+type CodeScanningService service
+
+type Alert struct {
+	RuleID          *string    `json:"rule_id,omitempty"`
+	RuleSeverity    *string    `json:"rule_severity,omitempty"`
+	RuleDescription *string    `json:"rule_description,omitempty"`
+	Tool            *string    `json:"tool,omitempty"`
+	CreatedAt       *time.Time `json:"created_at,omitempty"`
```
gmlewis, line 25: I believe we are trying to standardize on `*Timestamp` in this repo instead of `*time.Time`.
Here, and a few lines below for `ClosedAt`.

github/code-scanning.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return alerts, resp, nil
+}
+
+// GetAlert gets a single code scanning alert for a repository.
+//
+// You must use an access token with the security_events scope to use this endpoint.
+// GitHub Apps must have the security_events read permission to use this endpoint.
+//
+// The security alert_id is the number at the end of the security alert's URL.
```
gmlewis, line 76: Would it be possible to make this a little more accessible from `ListAlertsForRepo` so that users of the library don't have to parse it out themselves? 

I'm thinking we could add a helper method like:

```
func (a *Alert) ID() int64 {
...
}
```

that would then do the parsing for the user of this repo?  (If you agree, then this comment can be updated.)
Thoughts?

(Note that we want this to return `0` if `a==nil` or if parsing fails, and not panic. :smile: )
(Alternatively, we could change the signature to return `(int64, error)`, but I prefer the above since they could parse things themselves if they want.)

nightlark: That's a good idea. I added a helper function that checks Alert.HTMLURL first for something resembling a url with an ID at the end (split at last forward slash, if no forward slash then tries to split at the last trailing slash in Alert.URL), then converts whatever was found to an int64. Checking HTMLURL first was because the example given for getting the alert id used HTMLURL.

Do you think it's worth checking both HTMLURL and URL, or that there should be a more extensive check to see if the URL is valid before trying to convert whatever was after the last slash to an int64?

github/code-scanning.go
nightlark: If it's the same as `ClosedBy` for issues, yes. Will confirm with GitHub support.

github/code-scanning.go
```diff
@@ -22,14 +23,53 @@ type Alert struct {
...
+			// Both HTMLURL and URL were invalid and had no ID
+			return 0
+		}
+	}
+
+	// Return the alert ID as a 64-bit integer. Unable to convert or out of range returns 0.
+	if id, err := strconv.ParseInt(s, 10, 64); err == nil {
+		return id
+	}
+	return 0
```
gmlewis, line 70: Idiomatic Go would be to make the "happy path" the main path so that you can look at the end of the function to see the result that you want, and all other returns are typically error conditions. So that would mean that lines 67-70 would be:

```
id, err := strconv.ParseInt(s, 10, 64)
if err != nil {
  return 0
}

return id
```

github/code-scanning_test.go
```diff
@@ -14,6 +14,56 @@ import (
 	"time"
 )
 
+func TestActionsService_Alert_ID(t *testing.T) {
+	// Test: nil Alert ID == 0
+	a := (*Alert)(nil)
```
gmlewis, line 19: Idiomatic Go would be (since zero values are useful):

```
var a *Alert
```

github/code-scanning_test.go
```diff
@@ -14,6 +14,56 @@ import (
...
+	id := a.ID()
+	want := int64(0)
```
gmlewis, line 21: This is fine, but could also be:

```
var want int64 = 0
```

or even

```
var want int64
```

github/code-scanning.go
```diff
@@ -0,0 +1,134 @@
...
+	s := string("")
+	i := int(-1)
+
+	// Check if HTMLURL is valid
+	if a.HTMLURL != nil {
+		s = *a.HTMLURL
+	}
+
+	// Check HTMLURL for an ID to parse first, since it was used in the example
+	if i = strings.LastIndex(s, "/"); i != -1 {
```
gmlewis, line 50: Whups, I took a dinner break and forgot to comment that this would be:

```
if i := strings.LastIndex(s, "/"); i != -1 {
```

nightlark:
Should we wait until GitHub support responds with an answer for the `closed_by` field format, leave it as is for now, or switch it to `*User`?

gmlewis:
Hmmm... good question.  We already have a breaking change in the queue, unfortunately, immediately after the release of `v32.0.0`, so we could proceed with what we have here and then update it (if necessary) when we hear back from GitHub tech support... all before the release of `v33.0.0`.  I'm fine with that if you are OK with that.  But if you would prefer to wait, I'm totally fine with that too.

nightlark:
Updating when we hear back if needed would be okay. If I hear back before the next LGTM I'll update the PR.

nightlark:
Here's the response from GH support on the `closed_by` field:
```
I've had a look at our schema for that API response and, when the alert is closed, we'll return a user object for whoever closed it.

I don't have an example but it would be the same response as you see for user for this pull request:

https://developer.github.com/v3/pulls/#response-2
```

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1548, opened 2020-05-31
Add missing fields to user and org types

nightlark:
I noticed that there were a few fields in the `User` and `Organization` responses from GitHub that weren't in the types defined here.

https://developer.github.com/v3/users/
https://developer.github.com/v3/orgs/

For both: twitter_username
For orgs: is_verified, has_organization_projects, has_repository_projects


github/orgs.go
```diff
@@ -45,6 +46,9 @@ type Organization struct {
 	Type                        *string    `json:"type,omitempty"`
 	Plan                        *Plan      `json:"plan,omitempty"`
 	TwoFactorRequirementEnabled *bool      `json:"two_factor_requirement_enabled,omitempty"`
+	IsVerified                  *bool      `json:"is_verified,omitempty"`
+	HasOrgProjects              *bool      `json:"has_organization_projects,omitempty"`
+	HasRepoProjects             *bool      `json:"has_repository_projects,omitempty"`
```
gmlewis, line 51: In this repo, we have tried to keep a 1:1 transliteration of the JSON field names to the Go field names, (despite taking the liberty of shortening the API endpoint names)... so if you don't mind, could we please make these fields:

```
HasOrganizationProjects
HasRepositoryProjects
```

respectively?

nightlark: I expanded the abbreviations. The naming was following the shortening used for the `MemberCan*` permissions (which are already much longer even without shortening).

gmlewis: I'm not finding `MemberCan...` in this repo, but if they are not a 1:1 transliteration of the JSON field names, then I probably made a mistake in allowing them in.
Thanks for the changes.

nightlark: Ah, missed an s. This struct has several fields starting with `MembersCan` and `DefaultRepo` that are shortened.

PR #1556, opened 2020-06-15
Add missing workflow_id field from workflow run

patrickmarabeas:
Adds the missing `workflow_id` field from workflow runs endpoint.

See docs: https://developer.github.com/v3/actions/workflow-runs

### DIFF
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -22,6 +22,7 @@ type WorkflowRun struct {
 	Event          *string        `json:"event,omitempty"`
 	Status         *string        `json:"status,omitempty"`
 	Conclusion     *string        `json:"conclusion,omitempty"`
+	WorkflowID     *int           `json:"workflow_id,omitempty"`
 	URL            *string        `json:"url,omitempty"`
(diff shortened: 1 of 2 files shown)


github/actions_workflow_runs.go
```diff
@@ -22,6 +22,7 @@ type WorkflowRun struct {
 	Event          *string        `json:"event,omitempty"`
 	Status         *string        `json:"status,omitempty"`
 	Conclusion     *string        `json:"conclusion,omitempty"`
+	WorkflowID     *int           `json:"workflow_id,omitempty"`
```
gmlewis, line 25: Let's please change this to `*int64`.
Whenever we see GitHub use "ID", we've made them 64-bit ints just in case, so I think we should follow the pattern here as well.

Also, you'll see in line 100 below that `workflowID` is indeed already an `int64`:

```
func (s *ActionsService) ListWorkflowRunsByID(ctx context.Context, owner, repo string, workflowID int64, opts *ListWorkflowRunsOptions) (*WorkflowRuns, *Response, error) {
```

patrickmarabeas: Done!

Issue #1550, opened 2020-06-04
Support "package" webhook event


patrickmarabeas:
Add support for the `package` webhook event.

Docs: https://developer.github.com/webhooks/event-payloads/#package

gmlewis:
Thank you, @patrickmarabeas!

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

KevinMarioGerard:
Hi @gmlewis I would like to take this up. Will make a PR soon.

ravindranrahul:
Hi @gmlewis ,

If this is still relevant, I would like to give this a try.

gmlewis:
Thank you, @ravindranrahul!  It is yours.

jeet-parekh:
@ravindranrahul, any updates on this?

PR #1569, opened 2020-07-06
Handle package webhook

KrammerJake:
Addresses https://github.com/google/go-github/issues/1550

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -468,6 +468,22 @@ type OrgBlockEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// PackageEvent represents activity related to GitHub Packages.
+// The Webhook event name is "package".
+//
+// This event is triggered when a GitHub Package is published or updated.
+//
+// GitHub API docs: https://developer.github.com/webhooks/event-payloads/#package
+type PackageEvent struct {
+	// Action is the action that was performed.
+	// Can be "published" or "updated".
+	Action  *string       `json:"action,omitempty"`
+	Package *Package      `json:"package,omitempty"`
+	Repo    *Repository   `json:"repository,omitempty"`
+	Org     *Organization `json:"organization,omitempty"`
+	Sender  *User         `json:"sender,omitempty"`
+}
+
 // PageBuildEvent represents an attempted build of a GitHub Pages site, whether
 // successful or not.
 // The Webhook event name is "page_build".
--- a/github/packages.go
+++ b/github/packages.go
@@ -1,0 +1,101 @@
+// Copyright 2020 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+// Package represents a GitHub package.
+type Package struct {
+	ID             *int64           `json:"id,omitempty"`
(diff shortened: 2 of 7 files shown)


github/packages.go
```diff
@@ -0,0 +1,101 @@
...
+	Manifest            *string         `json:"manifest,omitempty"`
+	HTMLURL             *string         `json:"html_url,omitempty"`
+	TagName             *string         `json:"tag_name,omitempty"`
+	TargetCommitish     *string         `json:"target_commitish,omitempty"`
+	TargetOID           *string         `json:"target_oid,omitempty"`
+	Draft               *bool           `json:"draft,omitempty"`
+	Prerelease          *bool           `json:"prerelease,omitempty"`
+	CreatedAt           *Timestamp      `json:"created_at,omitempty"`
+	UpdatedAt           *Timestamp      `json:"updated_at,omitempty"`
+	PackageFiles        *[]PackageFile  `json:"package_files,omitempty"`
```
jeet-parekh, line 42: I think the type should be `[]*PackageFile`. Wait for somebody else to confirm it though.

gmlewis: Yes, deifinitely. Since a slice is already a reference type, we don't typically use pointers to slices.
Please change this to `[]*PackageFile`.

github/packages.go
```diff
@@ -0,0 +1,101 @@
...
+func (r PackageRelease) String() string {
+	return Stringify(r)
+}
+
+// PackageFile represents a GitHub package version release file.
+type PackageFile struct {
+	DownloadURL *string    `json:"download_url,omitempty"`
+	ID          *int64     `json:"id,omitempty"`
+	Name        *string    `json:"name,omitempty"`
+	Sha256      *string    `json:"sha256,omitempty"`
```
gmlewis, line 75: Since "SHA" is "Secure Hash Algorithm", let's please make these "SHA256" and "SHA1".

github/event_types.go
```diff
@@ -468,6 +468,22 @@ type OrgBlockEvent struct {
...
+// This event is triggered when a GitHub Package is published or updated.
+//
+// GitHub API docs: https://developer.github.com/webhooks/event-payloads/#package
+type PackageEvent struct {
+	// Action is the action that was performed.
+	// Can be "published" or "updated".
+	Action  *string       `json:"action,omitempty"`
+	Package *Package      `json:"package,omitempty"`
+	Repo    *Repository   `json:"repository,omitempty"`
+	Org     *Organization `json:"organization,omitempty"`
```
jeet-parekh, line 483: Use full names here? Repository and Organization instead of Repo and Org. For the sake of consistency.

KrammerJake: This file does have mixed naming conventions where both the full name and shorthand are used; however, the shorthand (`Repo` and `Org`) versions appear to be the more popular names in this file. I think we can leave these the way they are unless we're trying to standardize to the full name going forward.

gmlewis: Yes, we have not been totally consistent in this regard, so I didn't comment on these. I think that these are the two names we regularly allow to be shortened, as they are globally well known throughout the package and have very small chance of ever causing conflicts with new JSON tags that might show up in later releases.

I'm fine with leaving them as `Repo` and `Org`.

gmlewis:
Thank you, @wesleimp!
Merging.

PR #1571, opened 2020-07-09
github/repos: remove omitempty option in BranchRestrictionsRequest

### LINKED ISSUE #1546: Repositories.UpdateBranchProtection returning 422 Error when restrictions contain empty fields
With the release of `v32.0.0`, a `BranchRestrictionsRequest` [object](https://github.com/google/go-github/blob/407d73650d4bff607beaf8fb7a69eb9e9987af02/github/repos.go#L865) now omits empty fields in the JSON, however, when passing a `ProtectionRequest` with the following `BranchRestrictionsRequest` object to the `repositories.UpdateBranchProtection` method, the client returns the error below:
```
restrictions := new(github.BranchRestrictionsRequest)
restrictions.Users := []string{"angie44"}
restrictions.Teams := []string{}
restrictions.Apps := []string{}
```
```
Error: PUT https://api.github.com/repos/example-org/example-repo/branches/master/protection: 422 Invalid request.

        "teams" wasn't supplied.
        For 'anyOf/1', {"users"=>["anGie44"]} is not a null. []
```

Per the docs, empty fields in the `BranchRestrictionsRequest` object should be specified with the empty `[]string{}` instead of `nil`, but even so this results in the above error.

If however, the empty fields are set with a slice of size 1, i.e. `make([]string, 1)`, then the `PUT` request succeeds without error.

I wonder if there is another way we should be specifying the empty fields now that they are represented differently in the JSON request objects?

anGie44:
Closes #1546 

Notes:
* this reverts a previous change to the `BranchRestrictionsRequest` as it does not behave as expected such that restrictions configured for individual repositories still result in an error if users/teams fields are not set e.g.
```
"teams", "users" weren't supplied.
For 'anyOf/1', {"apps"=>["myApp"]} is not a null.
```
or if set to `nil` or `[]string{}` e.g.
```
422 Validation Failed [{Resource: Field: Code: Message:Only organization repositories can have users and team restrictions}]
```
* in addition, the change introduced a break in behavior for organization-level repositories as discussed in #1546
* a workaround didn't seem feasible as the validation issue seems to be upstream but any further suggestions are greatly appreciated 😃

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -864,9 +864,9 @@ type BranchRestrictions struct {
 // different from the response structure.
 type BranchRestrictionsRequest struct {
 	// The list of user logins with push access. (Required; use []string{} instead of nil for empty list.)
-	Users []string `json:"users,omitempty"`
+	Users []string `json:"users"`
 	// The list of team slugs with push access. (Required; use []string{} instead of nil for empty list.)
-	Teams []string `json:"teams,omitempty"`
+	Teams []string `json:"teams"`
 	// The list of app slugs with push access.
 	Apps []string `json:"apps,omitempty"`
 }


gmlewis:
Thank you, @andrew-waters!
Merging.

anGie44:
thank you @gmlewis and @andrew-waters!  would love to get this in the next release to help the Terraform Provider Github project benefit from the new features already in v32

gmlewis:
Sorry for the delay... hopefully I can get to this tomorrow.

gmlewis:
@anGie44 - this is now available in Release `v32.1.0`: https://github.com/google/go-github/releases/tag/v32.1.0
Thank you for your patience and understanding!

anGie44:
Not a problem, @gmlewis! thanks so much for your help and review 😃

PR #1576, opened 2020-07-16
Read response body for TCP connection reuse

### LINKED ISSUE #1575: Reuse Underlying TCP connection
Currently, the Github client only closes the response body which can be seen at https://github.com/google/go-github/blob/master/github/github.go#L557 . The resp.Body.Close() makes the connection available for reuse by others using the same http.Client. It doesn’t close the underlying HTTP or TCP connection.

According to the official Go docs:
https://golang.org/pkg/net/http/#Body
```
// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. It is the caller's responsibility to
// close Body. The default HTTP client's Transport may not
// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
// not read to completion and closed.
```

My proposal to solve the issue: 

- Read until Response is complete (i.e. ioutil.ReadAll(resp.Body) or similar)
- Call Body.Close()

itsksaurabh:
According to the official Go docs:
https://golang.org/pkg/net/http/#Body

```
// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. It is the caller's responsibility to
// close Body. The default HTTP client's Transport may not
// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
// not read to completion and closed.
```
To ensure http.Client connection reuse, I did the following things:

- Read until Response is complete (i.e. ioutil.ReadAll(resp.Body) or similar)
- Call Body.Close()

Closes #1575

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -554,7 +554,23 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 
 		return nil, err
 	}
-	defer resp.Body.Close()
+
+	defer func() {
+		// Ensure the response body is fully read and closed
+		// before we reconnect, so that we reuse the same TCP connection.
+		// Close the previous response's body. But read at least some of
+		// the body so if it's small the underlying TCP connection will be
+		// re-used. No need to check for errors: if it fails, the Transport
+		// won't reuse it anyway.
+		const maxBodySlurpSize = 2 << 10
+		if resp.ContentLength == -1 || resp.ContentLength <= maxBodySlurpSize {
+			io.CopyN(ioutil.Discard, resp.Body, maxBodySlurpSize)
+		}
+
+		if rerr := resp.Body.Close(); err == nil {
+			err = rerr
+		}
+	}()
 
 	response := newResponse(resp)
 
--- a/github/github.go
+++ b/github/github.go
@@ -567,9 +567,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 			io.CopyN(ioutil.Discard, resp.Body, maxBodySlurpSize)
 		}
 
-		if rerr := resp.Body.Close(); err == nil {
-			err = rerr
-		}
+		resp.Body.Close()
 	}()
 
 	response := newResponse(resp)


itsksaurabh:
> @googlebot I fixed it.

itsksaurabh:
@googlebot I fixed it.

itsksaurabh:
@googlebot I fixed it.

itsksaurabh:
@gmlewis It's my pleasure! :)
Thank you for your approval. 🙏

github/github.go
```diff
@@ -554,7 +554,23 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
+		// Close the previous response's body. But read at least some of
+		// the body so if it's small the underlying TCP connection will be
+		// re-used. No need to check for errors: if it fails, the Transport
+		// won't reuse it anyway.
+		const maxBodySlurpSize = 2 << 10
+		if resp.ContentLength == -1 || resp.ContentLength <= maxBodySlurpSize {
+			io.CopyN(ioutil.Discard, resp.Body, maxBodySlurpSize)
+		}
+
+		if rerr := resp.Body.Close(); err == nil {
```
philippfranke, line 570: I think, this if condition can be removed because `err` is not part of the named values. Therefore, the error is never returned if any occur.

itsksaurabh: @philippfranke Thank you for the review. I am glad that you liked the PR.
Yes, I also agree with your point. There is no need to check for error as it is not part of the named return values. Also, If it fails, the Transport won't reuse it anyway.

itsksaurabh: @philippfranke I just changed the code. Please let me know if it seems good to you. :)

itsksaurabh:
Thank you for the approval @philippfranke ♥️

gmlewis:
Thank you, @philippfranke !
Merging.

github/github.go
```diff
@@ -554,7 +554,21 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
+
+	defer func() {
+		// Ensure the response body is fully read and closed
+		// before we reconnect, so that we reuse the same TCP connection.
+		// Close the previous response's body. But read at least some of
+		// the body so if it's small the underlying TCP connection will be
+		// re-used. No need to check for errors: if it fails, the Transport
+		// won't reuse it anyway.
+		const maxBodySlurpSize = 2 << 10
+		if resp.ContentLength == -1 || resp.ContentLength <= maxBodySlurpSize {
```
gmlewis, line 566: I'm not sure why I didn't require unit tests on this PR, and that is my fault.

This PR causes a major breakage because it doesn't check if `resp` is `nil` and will therefore panic in those cases.

Please see #1601 for proposed fix.

itsksaurabh: @gmlewis I just saw the issue. I am not sure if that fix is actually required. 



The HTTP client returns the `nil` response whenever any error occurs.
Please check https://golang.org/src/net/http/client.go

In that case, we are already handling errors as you could see here at https://github.com/google/go-github/blob/master/github/github.go#L538

Moreover, The underlying Roundtripper() also checks such possibilities 

By simply checking for an error before doing anything with the response we can conclude what we can and cannot do next.
- if the error is not nil, we have an error. Return now or continue but avoid using the nil response
- if the error is nil, we don’t have an error. We can freely use the response as we originally intended.

Thanks!

bvaliev (commented):
`if resp.ContentLength == -1`
not needed here because it's int64 and it is always less than -1 if less than 2048

Checking **resp** for **nil** is unnecessary if the code is written according to Go standards because if **err != nil** is returned then **resp** must be **nil**

Generally you had to write `resp.ContentLength <= 2048` instead 565 and 566 strings

gmlewis:
Yes, @bvaliev - this PR has issues.  PRs are welcome and we'll review proposed changes.

itsksaurabh:
> `if resp.ContentLength == -1`
> not needed here because it's int64 and it is always less than -1 if less than 2048
> 
> Checking **resp** for **nil** is unnecessary if the code is written according to Go standards because if **err != nil** is returned then **resp** must be **nil**
> 
> Generally you had to write `resp.ContentLength <= 2048` instead 565 and 566 strings

@bvaliev I am not sure about everything you mentioned above. :)

```
    // ContentLength records the length of the associated content.
    // The value -1 indicates that the length is unknown.
    // Values >= 0 indicate that the given number of bytes may
    // be read from Body.
    //
    // For client requests, a value of 0 with a non-nil Body is
    // also treated as unknown.
    ContentLength int64
```
https://golang.org/pkg/net/http/

Do you have them somewhere specified in the docs? If so I would love to learn from it. Thanks!

itsksaurabh:
@gmlewis Thank you for your time on this PR. I would like to share with you that the latest official Go Http package does it internally now in the same way as you could see [here](https://github.com/golang/go/blob/master/src/net/http/client.go#L698). This shows the changes were needed indeed. So now, it is possible to clean up the code.

I have created a PR #1645 for it and now we could happily remove the code. :)

PR #1581, opened 2020-07-20
Remove unnecessary variable declaration.

MorrisLaw:
We don't need the `err` variable declaration prior to initialization. Since `err` is not needed as a return value at the end of the function, I think that relying on Go's built-in type inference makes sense here.

### DIFF
--- a/github/event.go
+++ b/github/event.go
@@ -129,8 +129,7 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 // Deprecated: Use ParsePayload instead, which returns an error
 // rather than panics if JSON unmarshaling raw payload fails.
 func (e *Event) Payload() (payload interface{}) {
-	var err error
-	payload, err = e.ParsePayload()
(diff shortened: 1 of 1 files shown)


MorrisLaw:
@gmlewis small refactor PR 😄

github/event.go
```diff
@@ -129,8 +129,7 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 // Deprecated: Use ParsePayload instead, which returns an error
 // rather than panics if JSON unmarshaling raw payload fails.
 func (e *Event) Payload() (payload interface{}) {
-	var err error
-	payload, err = e.ParsePayload()
+	payload, err := e.ParsePayload()
```
gmlewis, line 132: If you're going to make this change, then you should also change line 131 and remove the named return variable because you just declared a new `payload` and they are no longer the same thing.

gmlewis:
Closing abandoned PR.

PR #1582, opened 2020-07-21
Add RepositoriesService.CompareCommitsRaw method.

bradleymcallister97:
Follows the same high-level pattern as set out in PR #481 with PullRequestsService.GetRaw method.

### DIFF
--- a/github/repos_commits.go
+++ b/github/repos_commits.go
@@ -221,7 +221,6 @@ func (s *RepositoriesService) GetCommitSHA1(ctx context.Context, owner, repo, re
 }
 
 // CompareCommits compares a range of commits with each other.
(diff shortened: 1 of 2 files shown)


github/repos_commits.go
```diff
@@ -241,6 +240,34 @@ func (s *RepositoriesService) CompareCommits(ctx context.Context, owner, repo st
 	return comp, resp, nil
 }
 
+// CompareCommitsRaw compares a range of commits with each other in raw (diff or patch) format.
+//
```
gmlewis, line 244: Can you please add the following comments from the official docs?

```suggestion
//
// Both "base" and "head" must be branch names in "repo".
// To compare branches across other repositories in the same network as "repo",
// use the format "<USERNAME>:branch".
//
```

Issue #1583, opened 2020-07-22
Remove some fields of IssueListCommentsOptions


you06:
`Sort` and `Direction` not work since GitHub changed its API.

https://developer.github.com/v3/repos/commits/#list-commits

https://github.com/google/go-github/blob/e881974953e6ab6d1a6a1610e98ed6401a3aa1ba/github/issues_comments.go#L37-L48

gmlewis:
Thank you, @you06 !
The `Sort` and `Direction` fields need to be removed.  Although these apparently no longer are supported, this change is still a breaking API change, so I will mark it that way.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

juanri0s:
I can work on this! 😁

gmlewis:
Thank you, @juanri0s !  It is yours.

juanri0s:
@gmlewis I have removed the fields from the struct and the fields in the tests as well. Are we expecting this change in any other area other than `issues_comments.go`? Or do I need to go through the GH API docs and check myself?

gmlewis:
These fields were identified in the issue. You are welcome to look through the docs for other discrepancies.

tamalsaha:
@you06 @juanri0s @gmlewis @nightlark

Today I tried to update the client version and hit this issue. I am confused what is happening here. The OP points to the `list commits` api behavior changing. But the pr #1584 changed the `list comments` not `list commits` client. From what I can see the Github's api docs, these fields are still supported in the `list comments` api. https://docs.github.com/en/rest/reference/issues#list-issue-comments-for-a-repository .

I am confused what is happening here.

gmlewis:
> I am confused what is happening here

Wow... It looks like I messed this one up big time... I apologize, @tamalsaha .

It looks like I need to revert #1584 and go back and change the correct struct.

Thank you for the report!

gmlewis:
After further investigation, it turns out that this was an erroneous bug report.

Note that `ListCommits` does not have `Sort` or `Direction` fields and that there was no change to the GitHub API as was previously reported.

https://github.com/google/go-github/blob/master/github/repos_commits.go#L97-L116

Closing issue as incorrect and confusing.

PR #1584, opened 2020-07-23
Fix: Remove Sort & Direction Fields From IssueListCommentsOptions

juanri0s:
For: #1583 

- The `Sort` & `Direction` fields in` IssueListCommentsOptions` were removed from the github api so they no longer work.
- Fields have been removed from the struct and any corresponding tests. 

- [x] tests pass
- [x] code fmt

> These are tagged as breaking changes so not sure if anything else needs to be updated to reflect that

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -5084,28 +5084,12 @@ func (i *IssueEvent) GetURL() string {
 	return *i.URL
 }
 
-// GetDirection returns the Direction field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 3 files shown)


gmlewis:
Thank you, @nightlark !
Merging.

gmlewis:
Unfortunately, this PR removed fields that are still being used, as the original request #1583 reported the `ListCommits` method, but then showed the `ListComments` in error.

Reverting PR.

PR #1586, opened 2020-07-25
Add rocket and eyes as available emojis on the Reactions struct

breathingdust:
The [documentation](https://developer.github.com/v3/reactions/) lists two emojis which are not currently mapped over from the GitHub API response. This PR adds those emojis to the Reactions struct and aligns comments to reflect the new additions.


breathingdust:
@googlebot I signed it!

gmlewis:
Thank you, @nightlark!
Merging.

PR #1597, opened 2020-08-13
Add content_reference event

martin308:
Adds the content_reference web hook event along with the content reference type. 

`github/apps.go` seemed the best place for `ContentReference` to live but if there is a better place let me know.

GitHub API docs: https://developer.github.com/webhooks/event-payloads/#content_reference

### DIFF
--- a/github/apps.go
+++ b/github/apps.go
@@ -112,6 +112,13 @@ type Attachment struct {
 	Body  *string `json:"body,omitempty"`
 }
 
+// ContentReference represents a reference to a URL in an issue or pull request
+type ContentReference struct {
+	ID        *int64  `json:"id,omitempty"`
+	NodeID    *string `json:"node_id,omitempty"`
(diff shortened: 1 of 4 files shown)


martin308:
@googlebot I signed it!

martin308:
I don't really see any existing tests for this code path. Does that sound right?

martin308:
Found the ParseWebHook tests, added!

github/apps.go
```diff
@@ -112,6 +112,13 @@ type Attachment struct {
 	Body  *string `json:"body,omitempty"`
 }
 
+// ContentReference represents a reference to a URL in an issue or pull request
```
gmlewis, line 115:
```suggestion
// ContentReference represents a reference to a URL in an issue or pull request.
```

martin308:
@gmlewis great thanks! Found another issue implementing this content reference flow which is fixed in #1602

Issue #1552, opened 2020-06-05
No longer works with "api" GitHub Enterprise subdomain


oschwald:
As of #1165, `go-github` no longer works if the `api` GitHub Enterprise subdomain is used, which broke our use of it. The `/api/v3` suffix should only be added if the domain does not start with `api.`

```
$ curl -u <user>:<token> https://api.github.maxmind.com/api/v3/repos/maxmind/mm_website
{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/enterprise/2.20/v3"
}

$ curl -u <user>:<token> https://api.github.maxmind.com/repos/maxmind/mm_website
{
  <expected data>
}

$ curl -u <user>:<token> https://github.maxmind.com/api/v3/repos/maxmind/mm_website
{
  <expected data>
}
```

gmlewis:
Hmmm... I'm not sure I understand your example...  Why are you using `https://api.github.maxmind.com` when the third example is actually working properly (which this repo supports) - `https://github.maxmind.com` ?

oschwald:
The `api` subdomain is a feature of GitHub Enterprise and worked fine with `go-github` through version v28. In #1165, `go-github` started appending `/api/v3` to the end of Enterprise base URLs that didn't already have the suffix. This broke any existing users that were using the `api` subdomain.

We noticed this when upgrading from v28 to v32. The cause of the breakage was not immediately obvious. It was not listed as one of the breaking changes in the release notes and, from the PR, I don't think the intent was to break this particular use.

We did end up removing the `api` subdomain to work around this issue. However, the change in #1165 could be tweaked slightly to only add the suffix to base URLs that don't already have the suffix and are not using the `api` subdomain. This would prevent other users from running into this.

gmlewis:
OK, I apologize for the breakage, as I was unaware of the `api` subdomain.
Yes, please feel free to write a PR to fix this for other users, and thank you for alerting us to this breakage.

Or if you would like other volunteers to address this issue, that is fine too, and I can open it up for others to contribute.  It's your call.

chughpiyush:
I can work on this one.

chughpiyush:
Hi @gmlewis ,
Is this issue still something we would like to fix? If yes, I can have a PR ready quickly.

gmlewis:
Yes, please, @chughpiyush. Please make sure to fully unit test the various scenarios described above.

PR #1599, opened 2020-08-13
API subdomain fix for Github Enterprise and a few golint fixes

chughpiyush:
Hi,

This is a fix for issue #1552 

I added a condition that if the enterprise URL is something like `api.enterprise-url` or has an api-subdomain such as `something.api.enterprise-url` then there's no need to append `api/v3/` to the base endpoint URL. 

A few comment changes have also been done that were highlighted by the `golint` tool.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -313,7 +313,9 @@ func NewEnterpriseClient(baseURL, uploadURL string, httpClient *http.Client) (*C
 	if !strings.HasSuffix(baseEndpoint.Path, "/") {
 		baseEndpoint.Path += "/"
 	}
-	if !strings.HasSuffix(baseEndpoint.Path, "/api/v3/") {
+	if !strings.HasSuffix(baseEndpoint.Path, "/api/v3/") &&
+		!strings.HasPrefix(baseEndpoint.Host, "api.") &&
+		!strings.Contains(baseEndpoint.Host, ".api.") {
 		baseEndpoint.Path += "api/v3/"
 	}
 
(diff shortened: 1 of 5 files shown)


github/github.go
```diff
@@ -313,7 +313,9 @@ func NewEnterpriseClient(baseURL, uploadURL string, httpClient *http.Client) (*C
 	if !strings.HasSuffix(baseEndpoint.Path, "/") {
 		baseEndpoint.Path += "/"
 	}
-	if !strings.HasSuffix(baseEndpoint.Path, "/api/v3/") {
+	if !strings.HasSuffix(baseEndpoint.Path, "/api/v3/") &&
+		!strings.HasPrefix(baseEndpoint.Host, "api.") &&
```
gmlewis, line 317: It's never going to have this prefix because of the leading `https://`.

I'm not sure I'm happy yet with this PR because I'm not sure how many users of this repo we are helping or how many we are making more work for (to workaround the code we are adding).

github/github.go
```diff
@@ -324,7 +326,9 @@ func NewEnterpriseClient(baseURL, uploadURL string, httpClient *http.Client) (*C
 	if !strings.HasSuffix(uploadEndpoint.Path, "/") {
 		uploadEndpoint.Path += "/"
 	}
-	if !strings.HasSuffix(uploadEndpoint.Path, "/api/uploads/") {
+	if !strings.HasSuffix(uploadEndpoint.Path, "/api/uploads/") &&
+		!strings.HasPrefix(uploadEndpoint.Host, "api.") &&
```
gmlewis, line 330: Ditto. This line is not needed.

github/github.go
chughpiyush: Thank you for your response. 

`baseEndpoint.Host` won't contain `http` or `https`. It will just contain this highlighted part: 
 https://**api.domain.com**/query=abc
Example -> [https://play.golang.org/p/DaKtXOVQmKR](https://play.golang.org/p/DaKtXOVQmKR
)

The existing enterprise users won't have to do anything because they will be unaffected if they were not using the `api` subdomain in first place. We would be appending `api/v3/` to their URL, just as we do it today.

However, we would be helping new enterprise users (they won't have to do changes as had to be done by the issue reporter) and also helping existing enterprise users (who were not happy with this breaking change). 

For the enterprise users wich have URLs like:

_https://api.domain.com_ or _https://something.api.domain.com_ ,
we won't append suffix `api/v3/` 

For all other combinations of subdomains, we will continue to append suffix `/api/v3/` .

PR #1601, opened 2020-08-13
Checking for nil response before closing it.

kingbhanu:
Closing nil response will throw a panic so added a nil check before calling accessing resp object, defer function will be called even if there will be an error in response.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -555,7 +555,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	//Closing nil response will throw a panic. 
+	// Closing nil response will throw a panic.
 	if resp != nil {
 		defer func() {
 			// Ensure the response body is fully read and closed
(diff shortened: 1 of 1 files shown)


kingbhanu:
@googlebot I signed it!

github/github.go
```diff
@@ -555,20 +555,23 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
...
-		// Close the previous response's body. But read at least some of
-		// the body so if it's small the underlying TCP connection will be
-		// re-used. No need to check for errors: if it fails, the Transport
-		// won't reuse it anyway.
-		const maxBodySlurpSize = 2 << 10
-		if resp.ContentLength == -1 || resp.ContentLength <= maxBodySlurpSize {
-			io.CopyN(ioutil.Discard, resp.Body, maxBodySlurpSize)
-		}
+	// Closing nil response will throw a panic.
+	if resp != nil {
```
gmlewis, line 559: This is not the correct place to check for `resp` being nil.
It needs to happen within the defer and needs to be a wrapper around lines 568-572 below.

PR #1602, opened 2020-08-14
Update content type header for AppsService.CreateAttachment

martin308:
While implementing the content attachment flow I found that the header needed to use this API has been changed. Included the link to the changes blog post referencing this.


gmlewis:
Thank you, @ekrem95 !
Merging.

PR #1604, opened 2020-08-15
Clean URLs before fetching directory contents

### LINKED ISSUE #217: go-github incorrectly builds URLs when attempting fetching directory contents
If you call `(s *RepositoriesService) DownloadContents` on a file in the root directory of a repo on github enterprise it will get the dirname, `dir := path.Dir(filepath)`. It then passes this "dir" into `(s *RepositoriesService) GetContents`.

GetContents builds its urls using fmt.Sprintf:

```
u := fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, path)
```

When it does this you end up with a url that looks like:

```
/api/v3/repos/ids/sec_mon/contents//
```

Due to a bug in github (enterprise?) this will redirect to the User login page.

If you use a URL without the double trailing '/' then it works as expected.

This can easily be resolved by not building URLs using fmt.Sprintf and instead using path.Join:

```
path.Join("repos", owner, repo, "contents", path)
```

Github Enterprise version: Version 2.1.7

Bharat123rox:
Fixes #217 
Clean URLs before fetching directory contents, but attempt to make this more generic to the entire repository by cleaning the string in `NewRequest` itself, based on the discussion in #217 

Not sure of any potential downstream impacts this may have, or if a new test has to be introduced (I only see a test for [`testURLParseError`](https://github.com/google/go-github/blob/master/github/github_test.go#L122) )

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -18,6 +18,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"path"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


gmlewis:
Thank you, @Bharat123rox .

First, all the existing tests need to pass, but additionally a new test does need to be added to specifically address this comment: https://github.com/google/go-github/issues/217#issuecomment-548307957

gmlewis:
Closing abandoned PR.

Issue #1585, opened 2020-07-24
Create repo function doesn't wait until completion


kevinwongtyl:
In our github automation we create a github repo using `client.Repositories.Create(ctx, config.GITHUB_ORG, repo)` and call `ref, _, err := client.Git.GetRef(ctx, config.GITHUB_ORG, repoName, "refs/heads/master")` right after to prepare to create a nightly branch from master.  More than half the time we'll get an error because the client.Repositories.Create function returns without waiting for the actual github repo creation to fully finish.

Right now we are putting in a sleep to fix this issue in our automation but that is obviously not a great solution.

Here's the error we receive:
`2020/07/24 16:47:37 http: panic serving 172.25.0.1:41894: GET https://api.github.com/repos/<org>/test/git/refs/heads/master: 409 Git Repository is empty. []
reposervice_1  | goroutine 6 [running]:
reposervice_1  | net/http.(*conn).serve.func1(0xc0001c86e0)
reposervice_1  |        /usr/local/go/src/net/http/server.go:1772 +0x139
reposervice_1  | panic(0x943900, 0xc0002340f0)
reposervice_1  |        /usr/local/go/src/runtime/panic.go:973 +0x3e3
reposervice_1  | main.CreateBranch(0xc0000b6000, 0xaccba0, 0xc000028048, 0xc00009c015, 0xa, 0x9ffb7c, 0x6, 0xa00115, 0x7, 0xc000056aa8, ...)
reposervice_1  |        /go/src/app/service_github.go:140 +0x42a`

gmlewis:
Yes, that is correct based on my experience as well.  You basically need to set up a loop with a fixed number of attempts with exponential backoff (e.g. 1, 2, 4, 8, 16, 32 seconds) to query the repo until it is found. Presumably this is due to allowing GitHub to propagate the changes throughout their servers.

You will experience the same thing if you were to use the "curl" equivalent calls directly.

~Closing since this is known behavior on how GitHub operates.~

Feel free to contact support@github.com to see if they will wait to return the response until all changes have been fully propagated throughout their servers.

I'll leave this issue open for awhile in case someone would like to add comments to the `RepositoriesService.Create` endpoint to explain the situation... or if there is a response from GitHub tech support... but if there is no activity, then I'll close the issue.

chughpiyush:
@gmlewis I can give a PR to add relevant comment explaining this issue.

PR #1613, opened 2020-08-19
Added relevant comment clarifying the issue

chughpiyush:
Fixes: #1585 

Just added a comment clarifying the delay that takes place at Github's end and the method will return response without waiting for the creation process to complete.


github/repos.go
```diff
@@ -327,6 +327,11 @@ type createRepoRequest struct {
 // Note that only a subset of the repo fields are used and repo must
 // not be nil.
 //
+// Also note that this method will return the response without actually
+// waiting for Github to finish creating the repository and letting the
```
gmlewis, line 331:
```suggestion
// waiting for GitHub to finish creating the repository and letting the
```

chughpiyush: I updated this and took the liberty of updating the same in some other comments as well; hope that's fine.

PR #1615, opened 2020-08-21
Refactor check of http Response.ContentLength

linux019:
The check `resp.ContentLength <= maxBodySlurpSize` supersedes `resp.ContentLength == -1` as  ContentLength is signed int64
https://play.golang.org/p/Io9U59wxxkK


itsksaurabh:
@gmlewis @linux019 
The value -1 indicates that the length is unknown.

```
 // ContentLength records the length of the associated content.
    // The value -1 indicates that the length is unknown.
    // Values >= 0 indicate that the given number of bytes may
    // be read from Body.
    //
    // For client requests, a value of 0 with a non-nil Body is
    // also treated as unknown.
    ContentLength int64
```
Please check it here. https://golang.org/pkg/net/http/

gmlewis:
Closing this PR in favor of #1645.

PR #1635, opened 2020-10-02
switch to mediaTypeV3 for Checks Accept header

n1lesh:
switched Accept header's value to `mediaTypeV3` from `mediaTypeCheckRunsPreview`. Since **checks run** is out of preview, didn't see any value in keeping the old preview constant. (github.go)

Followed as per,

1. **Graduation of preview**: https://developer.github.com/changes/2020-10-01-graduate-antiope-preview/
2. **Docs**: https://docs.github.com/en/free-pro-team@latest/rest/reference/checks


n1lesh:
I signed it!

n1lesh:
@googlebot I fixed it.

PR #1618, opened 2020-08-24
Remove graduated GitHub v3 API preview headers

vikkyomkar:
This PR addresses below issue
#1616

PR #1636, opened 2020-10-02
Remove graduated GitHub v3 API preview header

n1lesh:
removed Accept header (with value mediaTypeCheckRunsPreview). Since checks run is out of preview, didn't see any value in keeping it.

Followed as per,

Graduation of preview: https://developer.github.com/changes/2020-10-01-graduate-antiope-preview/
Docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/checks

### DIFF
--- a/github/checks.go
+++ b/github/checks.go
@@ -104,7 +104,7 @@ func (s *ChecksService) GetCheckRun(ctx context.Context, owner, repo string, che
 		return nil, nil, err
 	}
 
-	req.Header.Set("Accept", mediaTypeCheckRunsPreview)
+	req.Header.Set("Accept", mediaTypeV3)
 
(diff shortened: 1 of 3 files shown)


gmlewis (changes requested):
Thank you, @n1lesh , but this is the wrong approach.
We need to literally remove it, not replace it.

Please see https://github.com/google/go-github/pull/1618/files as a recent example of how this is done.

gmlewis (approved):
Perfect!  Thank you, @n1lesh !
LGTM.

nightlark (approved):
LGTM

gmlewis:
Thank you, @nightlark !
Merging.

PR #1640, opened 2020-10-12
Add Workflow Dispatch Event by ID and FileName

### LINKED ISSUE #1639: Can't create workflow dispatch event by filename
Noticed that, [GitHub API docs: Create a workflow dispatch event](https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#create-a-workflow-dispatch-event), let's you create dispatch event by both workflowID and Filename.

However, [PR #1623 ](https://github.com/google/go-github/pull/1623) adds support for just workflowID, with **CreateWorkflowDispatchEvent**.

Can we split it into two _funcs_ for each type, as that's the general implementation in other APIs?

n1lesh:
Fixes #1639

### DIFF
--- a/github/actions_workflows.go
+++ b/github/actions_workflows.go
@@ -149,13 +149,26 @@ func (s *ActionsService) getWorkflowUsage(ctx context.Context, url string) (*Wor
 	return workflowUsage, resp, nil
 }
 
-// CreateWorkflowDispatchEvent manually triggers a GitHub Actions workflow run.
+// CreateWorkflowDispatchEventByID manually triggers a GitHub Actions workflow run.
 //
-// GitHub API docs: https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event
-func (s *ActionsService) CreateWorkflowDispatchEvent(ctx context.Context, owner, repo string, workflowID int64, event CreateWorkflowDispatchEventRequest) (*Response, error) {
+// GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#create-a-workflow-dispatch-event
+func (s *ActionsService) CreateWorkflowDispatchEventByID(ctx context.Context, owner, repo string, workflowID int64, event *CreateWorkflowDispatchEventRequest) (*Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/workflows/%v/dispatches", owner, repo, workflowID)
 
-	req, err := s.client.NewRequest("POST", u, &event)
+	return s.createWorkflowDispatchEvent(ctx, u, event)
+}
+
+// CreateWorkflowDispatchEventByFileName manually triggers a GitHub Actions workflow run.
+//
+// GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#create-a-workflow-dispatch-event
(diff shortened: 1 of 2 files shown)


github/actions_workflows.go
```diff
@@ -149,13 +149,26 @@ func (s *ActionsService) getWorkflowUsage(ctx context.Context, url string) (*Wor
 	return workflowUsage, resp, nil
 }
 
-// CreateWorkflowDispatchEvent manually triggers a GitHub Actions workflow run.
+// CreateWorkflowDispatchEventByID manually triggers a GitHub Actions workflow run.
 //
-// GitHub API docs: https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event
-func (s *ActionsService) CreateWorkflowDispatchEvent(ctx context.Context, owner, repo string, workflowID int64, event CreateWorkflowDispatchEventRequest) (*Response, error) {
+// GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#create-a-workflow-dispatch-event
```
gmlewis, line 154: @n1lesh - can you please explain to me what this extra `free-pro-team@latest` is and why you are adding these to the URL?

I have a utility in this repo that updates URLs, and it's not going to find these.

Thank you!

github/actions_workflows.go
```diff
@@ -149,13 +149,26 @@ func (s *ActionsService) getWorkflowUsage(ctx context.Context, url string) (*Wor
...
+func (s *ActionsService) CreateWorkflowDispatchEventByID(ctx context.Context, owner, repo string, workflowID int64, event *CreateWorkflowDispatchEventRequest) (*Response, error) {
```
gmlewis, line 155: We typically pass mandatory fields by value in this repo... so I'm thinking that these two exported functions should be left as they were and not use pointers.

However, I'm totally fine to let your `create...` helper function use a pointer since that stays internal to the package.

gmlewis (changes requested):
Thank you, @n1lesh !

One question and one comment, please.

github/actions_workflows.go
n1lesh: I actually took a little hint from `CreateUserImpersonation` from admin_users.go that expects pointer to `ImpersonateUserOptions`. I will update the two funcs to expect value though.

github/actions_workflows.go
n1lesh: For me, the docs open with `free-pro-team@latest` in the url, even when I open the old url. Maybe they changed it, or it's just for me?

gmlewis: Ah!  It does for me now too... it looks like I'll have to update all the URLs again at some point.  :joy:
Thanks, @n1lesh !

gmlewis (approved):
Thank you, @n1lesh !
LGTM.

gmlewis:
@erikkn - it would be nice to get all breaking changes into the next major release (v33.0.0) so that we don't have to bump the major version more often than necessary. If you feel like reviewing this PR, that would be greatly appreciated.

gmlewis:
friendly ping @wesleimp

gmlewis:
@nightlark - if you are available for a review, it would be greatly appreciated.

erikkn (approved):
LGTM

gmlewis:
Thank you, @erikkn !
Merging.

PR #1645, opened 2020-11-12
Clean up connection reuse code inside Do func as now Go HTTP packge does it internally

### LINKED ISSUE #1646: Clean up TCP connection reuse code inside Do func as now Go HTTP packge does it internally
The changes in the PR #1576 are not needed now as the latest official Go Http package does it internally. So Now, it is possible to remove the code from the PR.

https://github.com/golang/go/blob/master/src/net/http/client.go#L698

itsksaurabh:
The changes in the PR #1576 are not needed now as the latest official Go Http package does it internally. So Now, it is possible to remove the code from the PR.

https://github.com/golang/go/blob/master/src/net/http/client.go#L698

closes #1646

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -556,20 +556,7 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 		return nil, err
 	}
 
-	defer func() {
-		// Ensure the response body is fully read and closed
-		// before we reconnect, so that we reuse the same TCP connection.
-		// Close the previous response's body. But read at least some of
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @itsksaurabh !
LGTM.

itsksaurabh:
Thank you @gmlewis :)

nightlark (approved):
LGTM

gmlewis:
Thank you, @nightlark !
Merging.

PR #1649, opened 2020-11-15
Escape # in CompareCommits base and head refs

### LINKED ISSUE #1642: CompareCommits() doesn't encode given base/head refs?
Our GitHub integration does a number of automated steps, one being comparing a PR branch with the base, to determine if it can be cleanly merged. It uses `Repositories.CompareCommits()` passing `PullRequest.GetBase().GetRef()` and `PullRequest.GetHead().GetRef()` as base and head refs.

Now, someone on our team thought it would be a great idea to name a branch "feature/#nnnn-some-task", causing the `CompareCommits()` call to request the URL "https://api.github.com/repos/owner/repo/compare/develop...feature/#nnnn-some-task" which gave an unexpected 404. Replacing the "#" with "%23" fixed this.

As a work around, I'm now doing this replacement before calling `CompareCommits()` but I think this should probably be handled by the library instead, as it has a better chance of knowing which characters should be encoded when talking to this part of the GitHub API.

urohit011:
This change will fix #1642 by replacig `#` with `%23` in the base and head refs which are required for comparing commits.

Fixes #1642.

### DIFF
--- a/github/repos_commits.go
+++ b/github/repos_commits.go
@@ -9,7 +9,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"strings"
+	"net/url"
 	"time"
 )
 
@@ -223,12 +223,11 @@ func (s *RepositoriesService) GetCommitSHA1(ctx context.Context, owner, repo, re
 
 // CompareCommits compares a range of commits with each other.
 //
+// The url string u is escaped before making the request to api
+//
 // GitHub API docs: https://docs.github.com/en/rest/reference/repos/#compare-two-commits
 func (s *RepositoriesService) CompareCommits(ctx context.Context, owner, repo, base, head string) (*CommitsComparison, *Response, error) {
-	base = strings.ReplaceAll(base, "#", "%23")
-	head = strings.ReplaceAll(head, "#", "%23")
-
-	u := fmt.Sprintf("repos/%v/%v/compare/%v...%v", owner, repo, base, head)
+	u := url.QueryEscape(fmt.Sprintf("repos/%v/%v/compare/%v...%v", owner, repo, base, head))
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
--- a/github/repos_commits.go
+++ b/github/repos_commits.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strings"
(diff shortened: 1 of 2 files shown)


gmlewis (changes requested):
Please add a unit test that demonstrates this new behavior and also add a comment to the method that describes this behavior (between lines 225-226).
Thank you!

urohit011:
Hi @gmlewis , I have added the changes you requested. Please let me know if it's okay now. 
Thank you

github/repos_commits.go
```diff
@@ -222,9 +223,11 @@ func (s *RepositoriesService) GetCommitSHA1(ctx context.Context, owner, repo, re
 
 // CompareCommits compares a range of commits with each other.
 //
+// The url string u is escaped before making the request to api
```
gmlewis, line 226:
```suggestion
// The url query string is escaped before making the request.
```

gmlewis (changes requested):
Why don't I see `%23` anywhere in the updated unit test?
Shouldn't the `#b#` or `#h#` be escaped?

github/repos_commits.go
urohit011: Hi, acutally I store the escaped value for head and base in `h` and `b` and use the variables in the test instead of using literal values.

github/repos_commits_test.go
```diff
@@ -305,99 +306,118 @@ func TestRepositoriesService_CompareCommits(t *testing.T) {
...
+		h    string
+	}{
+		{name: "case 1", b: "b", h: "h"},
+		{name: "case 2", b: "#b#", h: "#h#"},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			b := url.QueryEscape(c.b)
+			h := url.QueryEscape(c.h)
```
urohit011, line 321: Hi, acutally I store the escaped value for head and base in `h` and `b` and use the variables in the test instead of using literal values.

gmlewis (approved):
Thank you, @urohit011 !
LGTM.

wesleimp (approved):
LGTM 👌🏼

gmlewis:
Thank you, @wesleimp !
Merging.

gmlewis:
This, unfortunately, causes problems that I should have caught. Reverting.

PR #1653, opened 2020-11-29
[New Method] Actions/Runner: List enabled repos in organization for Actions

erikkn:
This PR implements the `List selected repositories enabled for GitHub Actions in an organization` method. The endpoint for this feature is `/orgs/{org}/actions/permissions/repositories`. More details around this can be found here: https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#list-selected-repositories-enabled-for-github-actions-in-an-organization

### DIFF
--- a/github/actions_runners.go
+++ b/github/actions_runners.go
@@ -8,7 +8,6 @@ package github
 import (
 	"context"
 	"fmt"
-	"strings"
 )
 
 // RunnerApplicationDownload represents a binary for the self-hosted runner application that can be downloaded.
@@ -242,20 +241,20 @@ func (s *ActionsService) ListOrganizationRunners(ctx context.Context, owner stri
 // ListEnabledReposInOrg lists the selected repositories that are enabled for GitHub Actions in an organization.
 //
 // GitHub API docs: https://docs.github.com/en/rest/reference/actions#list-selected-repositories-enabled-for-github-actions-in-an-organization
-func (s *ActionsService) ListEnabledReposInOrg(ctx context.Context, owner string) (*ActionsEnabledOnOrgRepos, *Response, error) {
+func (s *ActionsService) ListEnabledReposInOrg(ctx context.Context, owner string, opts *ListOptions) (*ActionsEnabledOnOrgRepos, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/actions/permissions/repositories", owner)
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept headers when APIs fully launch.
-	acceptHeaders := []string{mediaTypeTopicsPreview}
-	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
-
 	repos := &ActionsEnabledOnOrgRepos{}
-	resp, err := s.client.Do(ctx, req, &repos)
+	resp, err := s.client.Do(ctx, req, repos)
 	if err != nil {
 		return nil, resp, err
 	}
--- a/github/actions_runners.go
+++ b/github/actions_runners.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"strings"
 )
 
 // RunnerApplicationDownload represents a binary for the self-hosted runner application that can be downloaded.
@@ -16,6 +17,12 @@ type RunnerApplicationDownload struct {
 	Architecture *string `json:"architecture,omitempty"`
 	DownloadURL  *string `json:"download_url,omitempty"`
 	Filename     *string `json:"filename,omitempty"`
+}
+
+// ActionsEnabledOnOrgRepos represents all the repsositories in an organization for which Actions is enabled.
+type ActionsEnabledOnOrgRepos struct {
+	TotalCount   int           `json:"total_count"`
+	Repositories []*Repository `json:"repositories"`
 }
 
 // ListRunnerApplicationDownloads lists self-hosted runner application binaries that can be downloaded and run.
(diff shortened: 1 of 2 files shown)


erikkn:
@googlebot I signed it!

github/actions_runners.go
```diff
@@ -232,6 +239,30 @@ func (s *ActionsService) ListOrganizationRunners(ctx context.Context, owner stri
 	return runners, resp, nil
 }
 
+// ListEnabledReposInOrg lists the selected repositories that are enabled for GitHub Actions in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/actions#list-selected-repositories-enabled-for-github-actions-in-an-organization
+func (s *ActionsService) ListEnabledReposInOrg(ctx context.Context, owner string) (*ActionsEnabledOnOrgRepos, *Response, error) {
```
gmlewis, line 245:
```suggestion
func (s *ActionsService) ListEnabledReposInOrg(ctx context.Context, owner string, opts *ListOptions) (*ActionsEnabledOnOrgRepos, *Response, error) {
```

github/actions_runners.go
```diff
@@ -232,6 +239,30 @@ func (s *ActionsService) ListOrganizationRunners(ctx context.Context, owner stri
...
+	u := fmt.Sprintf("orgs/%v/actions/permissions/repositories", owner)
+
```
gmlewis, line 247: Please add:

```
	u, err := addOptions(u, opts)
	if err != nil {
		return nil, nil, err
	}
```

before line 247.

github/actions_runners.go
```diff
@@ -232,6 +239,30 @@ func (s *ActionsService) ListOrganizationRunners(ctx context.Context, owner stri
...
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept headers when APIs fully launch.
+	acceptHeaders := []string{mediaTypeTopicsPreview}
+	req.Header.Set("Accept", strings.Join(acceptHeaders, ", "))
```
gmlewis, line 255: Please remove lines 253-256 as they are not needed according to the documentation. (Please correct me if I'm wrong... I'm not seeing any mention of them.)

github/actions_runners.go
```diff
@@ -232,6 +239,30 @@ func (s *ActionsService) ListOrganizationRunners(ctx context.Context, owner stri
...
+
+	repos := &ActionsEnabledOnOrgRepos{}
+	resp, err := s.client.Do(ctx, req, &repos)
```
gmlewis, line 258:
```suggestion
	resp, err := s.client.Do(ctx, req, repos)
```

gmlewis (changes requested):
Thank you, @erikkn - this is a good start.
Please add unit tests to cover this new method.  See our CONTRIBUTING.md guide for full details, including the requirement to run `go generate ...`.

erikkn:
Hi, @gmlewis thank you for your review, really appreciate it :)! Just added your requested changes. Will look into adding the necessary tests now

erikkn:
@gmlewis may I ask you for another review, please :)? Cheers

gmlewis (approved):
Thank you, @erikkn !
LGTM.

erikkn:
Hey, @wesleimp I know it is still early over there ;) and I obviously don't want to push, but do you have time to review this PR? Would be really great if we can get this in today :).

github/actions_runners.go
```diff
@@ -18,6 +18,12 @@ type RunnerApplicationDownload struct {
 	Filename     *string `json:"filename,omitempty"`
 }
 
+// ActionsEnabledOnOrgRepos represents all the repsositories in an organization for which Actions is enabled.
```
nightlark, line 21:
```suggestion
// ActionsEnabledOnOrgRepos represents all the repositories in an organization for which Actions is enabled.
```

github/actions_runners_test.go
```diff
@@ -225,6 +225,56 @@ func TestActionsService_ListOrganizationRunners(t *testing.T) {
...
+
+	want := &ActionsEnabledOnOrgRepos{TotalCount: int(2), Repositories: []*Repository{
+		{ID: Int64(2)},
+		{ID: Int64(3)},
+	}}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("Actions.ListEnabledReposInOrg returned %+v, want %+v", got, want)
+	}
+
+	// Test addOptions failure
```
nightlark, line 257: Is this testing an invalid owner, or bad list options? It looks like opt hasn't changed since the previous (valid) call that was tested. Is the error returned actually related to the arguments passed to the function, or is it an error about not finding the endpoint because the test only has an endpoint for `/orgs/o/actions/permissions/repositories` registered?

erikkn: Hey @nightlark :)! I am not sure if I understand your comment, could you give me some pointers?

nightlark: The comment about it testing addOptions doesn't quite feel like it fits what is happening (a newline in the middle of a url where the owner belongs).

I found ~4 other tests with the same thing so I guess that the tests you added were based on one of those. Not sure the comment is accurate in any of those cases, or is really needed in any of those cases (I think throwing a newline in the middle of a url by using \n as the owner or organization argument to any go-github function is likely to give an error, so in a way the test seems kinda pointless -- gmlewis maybe you have some insight I'm missing).

Given that the same sort of test case happens in a few other places, I don't think leaving another instance of it in this test does any harm.

nightlark (approved):
The two failure scenarios in the test look more generic to me than something that needs to be tested for each endpoint, but where/how to better test or handle those types of invalid inputs seems like a separate discussion.

PR LGTM.

erikkn:
Thank you @nightlark  :)! Is there any chance you can merge the PR for me as well?

github/actions_runners_test.go
gmlewis: Hi @nightlark - yes, this has been added simply to increase the code coverage reports.  Adding in the newline is basically an easy way to generate an error, so it exercises the error-handling path.

gmlewis:
Thank you for the review, @nightlark !
Merging.

erikkn:
@gmlewis thanks for merging! How do you usually deal with new releases :)? Or is there a README I forgot to check about this topic

gmlewis:
Hi @erikkn - releases lately have been on-demand.  I'm assuming you would like one, so I'll work on putting together a release. :smile:

erikkn:
Rockstar! Thank you soo much :100:

gmlewis:
@erikkn - https://github.com/google/go-github/releases/tag/v33.0.0 is now complete.
It is also now accessible here: https://pkg.go.dev/github.com/google/go-github/v33/github

PR #1658, opened 2020-12-02
Remove HeadSHA parameter from UpdateCheckRunOptions

### LINKED ISSUE #1652: Update Check Run HeadSHA
https://github.com/google/go-github/blob/1ff2714d8affd5ede5366bdb1a2d9472b669fe77/github/checks.go#L175-L186

Despite having an optional `HeadSHA` field, if you update this to a different commit hash it won't be updated.

We have some Check Runs that are PR-oriented and not commit-oriented and we were trying to use this field to rebind the Check Runs to the head commit of a PR on `synchronize`.

Is this a bug?

ishanupadhyay:
This PR is regarding [issue 1652.](https://github.com/google/go-github/issues/1652) head_sha is no longer supported as a parameter in case of [updating a check run](https://docs.github.com/en/free-pro-team@latest/rest/reference/checks#update-a-check-run). Therefore it is being removed from **type UpdateCheckRunOptions** in this PR. 

Fixes #1652.

### DIFF
--- a/github/checks.go
+++ b/github/checks.go
@@ -175,7 +175,6 @@ func (s *ChecksService) CreateCheckRun(ctx context.Context, owner, repo string, 
 // UpdateCheckRunOptions sets up parameters needed to update a CheckRun.
 type UpdateCheckRunOptions struct {
 	Name        string            `json:"name"`                   // The name of the check (e.g., "code-coverage"). (Required.)
-	HeadSHA     *string           `json:"head_sha,omitempty"`     // The SHA of the commit. (Optional.)
 	DetailsURL  *string           `json:"details_url,omitempty"`  // The URL of the integrator's site that has the full details of the check. (Optional.)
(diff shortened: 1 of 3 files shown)


ishanupadhyay:
@gmlewis can you please review this PR ?

gmlewis:
Note that this PR is a breaking API change, so I've added the label to keep track.

gmlewis (approved):
Thank you, @ishanupadhyay!
LGTM.

gmlewis:
@erikkn - it would be nice to get all breaking changes into the next major release (`v33.0.0`) so that we don't have to bump the major version more often than necessary.  If you feel like reviewing this PR, that would be greatly appreciated.

erikkn (approved):
We are aware of the breaking API change, so LGTM.

gmlewis:
Thank you, @erikkn ! 
Merging.

PR #1662, opened 2020-12-06
Add github-accessors_test.go

gmlewis:
While taking a look at code coverage, I noticed that the auto-generated `github-accessors.go` file had 0.3% code coverage.

This PR increases the `github-accessors.go` code coverage from 0.3% to 100%.


gmlewis:
Apparently, codecov.io ignores `github-accessors.go`, but this PR still benefits the local code coverage results, so I think it is still a good idea to include.

github/github-accessors_test.go
```diff
@@ -0,0 +1,18304 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
```
nightlark, line 1: Should the copyright year get updated based on when these files are (re)generated?

gmlewis: Good question, but the answer is no. Once a file is created, the copyright year remains unmodified.
(At least for this repo, that is the instruction given from the Google Open Source team.)

gmlewis: Oh!  I see your point now.  This is a new test file.

OK, I'm going to make an executive decision and keep the two sync'd to the same 2017 year.

Thank you, @nightlark .

PR #1663, opened 2020-12-06
Increase apps_installation.go code coverage

gmlewis:
This PR increases the code coverage of `apps_installation.go` from 76.2% to 95.2%.

### DIFF
--- a/github/apps_installation_test.go
+++ b/github/apps_installation_test.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"reflect"
 	"testing"
+	"time"
 )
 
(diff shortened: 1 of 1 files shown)


gmlewis:
Too bad that the codecov.io results don't match the results of running these commands:

```sh
#!/bin/bash -ex
go test -coverprofile=c.out ./...
go tool cover -html=c.out
```

gmlewis:
Here are the percentages as calculated by codecov.io:
https://codecov.io/gh/google/go-github/pull/1663/tree/github

nightlark (approved):
LGTM

PR #1668, opened 2020-12-06
Remove unused disableWorkflow method

gmlewis:
The `disableWorkflow` method was not being called, so this PR renames `enableWorkflow` and removes the unused method.


nightlark (approved):
LGTM

PR #1675, opened 2020-12-14
Add action, organization and installation objects to webhook payload

afdesk:


afdesk:
@googlebot I signed it!

PR #1683, opened 2020-12-23
Fix misspelling of timestamp as timstamp

Mukundan314:


gmlewis (approved):
Thank you, @Mukundan314 !
Merging.

Issue #592, opened 2017-03-14
FYI - Discontinue support for weak cryptographic standards


gmlewis:
GitHub API Announcement:
https://developer.github.com/changes/2017-03-13-discontinue-support-for-weak-cryptographic-standards/

Note that this should not affect this repo, as we currently support `sha1`, `sha256`, and `sha512` webhook messages in `messages.go`.

This issue is mainly for informational purposes only and can be closed after some short amount of time.

bradfitz:
What does that linked article have to do with the webhook messages we support?

The announcement was about TLS versions (Go is fine) and SSH ciphersuites (unrelated to the go-github package).

One of us is confused. Maybe both. :)

shurcooL:
I'll join in on the confused party.

> What does that linked article have to do with the webhook messages we support?

A _guess_ is that this was meant to be related to [`ValidatePayload`](https://godoc.org/github.com/google/go-github/github#ValidatePayload) func, which internally uses `messageMAC`, which currently does:

```Go
var hashFunc func() hash.Hash
switch sigParts[0] {
case sha1Prefix:
	hashFunc = sha1.New
case sha256Prefix:
	hashFunc = sha256.New
case sha512Prefix:
	hashFunc = sha512.New
default:
	return nil, nil, fmt.Errorf("unknown hash type prefix: %q", sigParts[0])
}
```

Source: https://github.com/google/go-github/blob/839082db564f53df5b8b386f217c7c217b5a624f/github/messages.go#L100-L110.

But I'll let @gmlewis confirm if my guess is right.

gmlewis:
Uh, sorry.  Yes, @shurcooL is right.
I presumed (possibly incorrectly) that GitHub would stop sending their webhook hashes using `sha1` and possibly switch to `sha256` although I didn't read that anywhere.

I was just trying to give an FYI update and let users of this package know that if/when GitHub makes that change, no work needs to be done on this side.

Does that make any sense?

reinoudk:
Hi, I've stumbled upon this issue today as I'm building a service that handles GitHub events. I use the `github.ValidatePayload` function and was surprised that my tests were failing. Turns out I was using the header `X-Hub-Signature-256`, whilst `github.ValidatePayload` has the constant `signatureHeader` defined as `X-Hub-Signature`. 

[...]

PR #1783, opened 2021-01-26
Add support for the X-Hub-Signature-256 header

reinoudk:
This relates to #592 and adds support for the `X-Hub-Signature-256` header. I've chosen to set the SHA256 header as the default and falling back to the SHA1 header when the SHA256 header is empty. The GitHub documentation [recommends](https://docs.github.com/en/developers/webhooks-and-events/securing-your-webhooks#validating-payloads-from-github) using the SHA256 header so I don't expect the fallback to be needed in a live environment. Still, this fallback prevents introducing breaking changes when used in tests.

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -137,19 +137,6 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 	return buf, hashFunc, nil
 }
 
-// payloadSignature tries to return the HMAC signature of the payload
-// It will first try to find the HMAC-SHA256 signature in the "X-Hub-Signature-256" header and falls back to
-// the HMAC-SHA1 signature in the "X-Hub-Signature" header.
-// Returns an empty string when both headers are not set / empty.
-func payloadSignature(r *http.Request) string {
-	var sig string
-	sig = r.Header.Get(sha256SignatureHeader)
-	if sig == "" {
-		sig = r.Header.Get(sha1SignatureHeader)
-	}
-	return sig
-}
-
 // ValidatePayload validates an incoming GitHub Webhook event request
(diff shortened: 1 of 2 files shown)


github/messages.go
```diff
@@ -190,7 +205,7 @@ func ValidatePayload(r *http.Request, secretToken []byte) (payload []byte, err e
 	// Only validate the signature if a secret token exists. This is intended for
 	// local development only and all webhooks should ideally set up a secret token.
 	if len(secretToken) > 0 {
-		sig := r.Header.Get(signatureHeader)
+		sig := payloadSignature(r)
```
willnorris, line 208: since payloadSignature is so short, it's only ever called from here, and as you note elsewhere we could probably drop support for SHA1 at some point anyway, maybe just check for the two headers inline here?

``` go
	sig = r.Header.Get(sha256SignatureHeader)
	if sig == "" {
		sig = r.Header.Get(sha1SignatureHeader)
	}
```

willnorris:
(and this looks great by the way... thanks for adding this!)

willnorris:
/cc @gauntface and @taquitos as FYI

if I remember correctly, there is some code in Google's CLA checker that constructs a fake webhook payload to manually trigger rescanning a CLA.  If go-github were to ever drop support for these SHA-1 signatures (which there's probably no need to do, but it's not inconceivable), you would need to make sure to move over to SHA-256 for signing those fake webhook payloads.

github/messages.go
reinoudk: Good point! Changed it to the inline version. Thanks for looking at this so quick :rocket:.

willnorris (approved):

PR #1786, opened 2021-01-27
fix api preview header, missing :

jurgenweber:
so then the api previews actually work. :)

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -30,8 +30,10 @@ const (
 	// sha256Prefix and sha512Prefix are provided for future compatibility.
 	sha256Prefix = "sha256"
 	sha512Prefix = "sha512"
-	// signatureHeader is the GitHub header key used to pass the HMAC hexdigest.
-	signatureHeader = "X-Hub-Signature"
+	// sha1SignatureHeader is the GitHub header key used to pass the HMAC-SHA1 hexdigest.
+	sha1SignatureHeader = "X-Hub-Signature"
+	// sha256SignatureHeader is the GitHub header key used to pass the HMAC-SHA256 hexdigest.
+	sha256SignatureHeader = "X-Hub-Signature-256"
 	// eventTypeHeader is the GitHub header key used to pass the event type.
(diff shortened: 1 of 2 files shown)


jurgenweber:
@googlebot I signed it!

jurgen-weber-deltatre:
considering all the failing CI, this doesn't seem to be the only place/thing I need to update but it is also making me second guess that this is the solution... Will await further input/thoughts.

gmlewis:
> so then the api previews actually work. :)

I'm sorry, I don't understand why this change is necessary.  This repo has been using custom preview headers for years now and they have been working well up to this point. Could you please explain what you saw that caused you to write this PR?

Also, you should be able to run "go test ./..." locally to make sure the unit tests pass.

jurgen-weber-deltatre:
> I'm sorry, I don't understand why this change is necessary. This repo has been using custom preview headers for years now and they have been working well up to this point. 

Yeha, I agree.. Seems weird to me.

> Could you please explain what you saw that caused you to write this PR?

https://github.com/google/go-github/pull/1786#issuecomment-768662735

It was a long story to get to this point, but I feel there is some more to it. I started with using the terraform-provider-github and in the end I can not make internal repos using the provider. I have subsequently ended up here. I started replicating the provider commands with curl commands to ID what is going on.

The beginings of my journey are here; https://github.com/integrations/terraform-provider-github/issues/662

If it isn't the headers, then I am not sure to why when using TF this fails.

GitHub support then pointed me to the concept of the API reviews and I noticed the string difference, using CURL the api call fails without a colon.

> Also, you should be able to run "go test ./..." locally to make sure the unit tests pass.

Ok

jurgen-weber-deltatre:
Just a follow up; I tried this locally (found some time this morning):

https://github.com/jurgenweber/go-github/blob/master/example/orgnewrepo/main.go

this works, so the problem is else wehre.

Thanks

jurgenweber:
I found it; https://github.com/integrations/terraform-provider-github/pull/680

PR #1789, opened 2021-02-02
Add support for removing enterprise Actions runners

zetaab:
adds support for

https://docs.github.com/en/rest/reference/enterprise-admin#delete-a-self-hosted-runner-from-an-enterprise

### DIFF
--- a/github/enterprise_actions_runners.go
+++ b/github/enterprise_actions_runners.go
@@ -54,10 +54,10 @@ func (s *EnterpriseService) ListRunners(ctx context.Context, enterprise string, 
 	return runners, resp, nil
 }
 
-// RemoveEnterpriseRunner forces the removal of a self-hosted runner from an enterprise using the runner id.
+// RemoveRunner forces the removal of a self-hosted runner from an enterprise using the runner id.
 //
 // GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/enterprise-admin/#delete-a-self-hosted-runner-from-an-enterprise
-func (s *ActionsService) RemoveEnterpriseRunner(ctx context.Context, enterprise string, runnerID int64) (*Response, error) {
+func (s *EnterpriseService) RemoveRunner(ctx context.Context, enterprise string, runnerID int64) (*Response, error) {
 	u := fmt.Sprintf("enterprises/%v/actions/runners/%v", enterprise, runnerID)
 
 	req, err := s.client.NewRequest("DELETE", u, nil)
--- a/github/enterprise_actions_runners.go
+++ b/github/enterprise_actions_runners.go
@@ -56,7 +56,7 @@ func (s *EnterpriseService) ListRunners(ctx context.Context, enterprise string, 
 
 // RemoveEnterpriseRunner forces the removal of a self-hosted runner from an enterprise using the runner id.
 //
-// GitHub API docs: https://docs.github.com/en/rest/reference/enterprise-admin#delete-a-self-hosted-runner-from-an-enterprise
+// GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/enterprise-admin/#delete-a-self-hosted-runner-from-an-enterprise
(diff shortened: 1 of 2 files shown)


zetaab:
@googlebot I signed it!

github/actions_runners.go
```diff
@@ -214,6 +214,50 @@ func (s *ActionsService) CreateOrganizationRegistrationToken(ctx context.Context
 	return registrationToken, resp, nil
 }
 
+// CreateEnterpriseRegistrationToken creates a token that can be used to add a self-hosted runner to an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/enterprise-admin#create-a-registration-token-for-an-enterprise
+func (s *ActionsService) CreateEnterpriseRegistrationToken(ctx context.Context, enterprise string) (*RegistrationToken, *Response, error) {
+	u := fmt.Sprintf("enterprises/%v/actions/runners/registration-token", enterprise)
+
+	req, err := s.client.NewRequest("POST", u, nil)
```
gmlewis, line 223: I'm confused.  It looks to me like we already support this endpoint:

https://github.com/google/go-github/blob/master/github/enterprise_actions_runners.go#L17

github/actions_runners.go
```diff
@@ -214,6 +214,50 @@ func (s *ActionsService) CreateOrganizationRegistrationToken(ctx context.Context
...
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/enterprise-admin#list-self-hosted-runners-for-an-enterprise
+func (s *ActionsService) ListEnterpriseRunners(ctx context.Context, enterprise string, opts *ListOptions) (*Runners, *Response, error) {
+	u := fmt.Sprintf("enterprises/%v/actions/runners", enterprise)
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
```
gmlewis, line 247: We support this one too:

https://github.com/google/go-github/blob/master/github/enterprise_actions_runners.go#L37

github/actions_runners.go
```diff
@@ -314,3 +358,17 @@ func (s *ActionsService) RemoveOrganizationRunner(ctx context.Context, owner str
...
 	return s.client.Do(ctx, req, nil)
 }
+
+// RemoveEnterpriseRunner forces the removal of a self-hosted runner from an enterprise using the runner id.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/enterprise-admin#delete-a-self-hosted-runner-from-an-enterprise
+func (s *ActionsService) RemoveEnterpriseRunner(ctx context.Context, enterprise string, runnerID int64) (*Response, error) {
+	u := fmt.Sprintf("enterprises/%v/actions/runners/%v", enterprise, runnerID)
+
+	req, err := s.client.NewRequest("DELETE", u, nil)
```
gmlewis, line 368: This one appears to be new, but it should be moved alongside the other two endpoints listed above.

gmlewis (changes requested):

github/actions_runners.go
zetaab: oh sorry I did not check that there are another file.. :)

github/actions_runners.go
zetaab: done

gmlewis (approved):
Thank you, @zetaab !
LGTM.

wesleimp (approved):
LGTM 👌🏼

zetaab:
@gmlewis now we could merge this?

gmlewis (approved):
Ah, thank you for the naming changes, @zetaab ... I'm glad you caught that, as I had missed it.
Thank you, @wesleimp for the code review.
LGTM.
Merging.

Oh, and in the future, @zetaab , we prefer that you don't force push to PRs in this repo because it typically makes the code reviews more challenging to see what was already reviewed... but in this case, it made no difference, fortunately... just an FYI for future contributions.  Thanks again!

PR #1794, opened 2021-02-04
Change ListRepos and ListUserRepos to return new ListRepositories struct with count

### LINKED ISSUE #1791: Return TotalCount in ListRepos in AppsService
`ListRepos` currently only returns the list of repositories from github's response but not the total_count ([Link](https://docs.github.com/en/rest/reference/apps#list-repositories-accessible-to-the-app-installation)). This total_count is different from just the length of the list of repositories returned if the number of repositories in the installation is greater than the per_page pagination limit set in the request. Is it possible to add it for ListRepos to return?

DarthBenro008:
- Fixes #1791
- Modified Tests in `apps_installations_test.go`

### DIFF
--- a/github/apps_installation.go
+++ b/github/apps_installation.go
@@ -10,10 +10,15 @@ import (
 	"fmt"
 )
 
+type ListRepositories struct {
+	TotalCount   *int          `json:"total_count,omitempty"`
+	Repositories []*Repository `json:"repositories"`
+}
+
 // ListRepos lists the repositories that are accessible to the authenticated installation.
 //
 // GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/apps/#list-repositories-accessible-to-the-app-installation
-func (s *AppsService) ListRepos(ctx context.Context, opts *ListOptions) ([]*Repository, *Response, error) {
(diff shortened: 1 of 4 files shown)


github/apps_installation.go
```diff
@@ -10,10 +10,15 @@ import (
 	"fmt"
 )
 
+type ListRepositories struct {
```
gmlewis, line 13: Can you please add a godoc-style comment here?  Maybe something like this?

```
// ListRespositories represents the response from the list repos endpoints.
```

gmlewis (changes requested):
Thank you, @DarthBenro008 !
Just one minor comment addition, please, then we should be ready for a second LGTM before merging.

github/apps_installation.go
DarthBenro008: Added the godoc-style comment!

gmlewis (approved):
Thank you, @DarthBenro008 !
LGTM.

wesleimp (approved):
LGTM 👌🏼

gmlewis:
Thank you, @wesleimp !
Merging.

Issue #1775, opened 2021-01-13
Add missing fields to MarketplacePlan and MarketplacePurchase structs


SudoVim:
According to the GitHub documentation: https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#list-accounts-for-a-plan the "number" field is missing from MarketplacePlan, and the "updated_at" field is missing from MarketplacePurchase. I've already made these changes in my fork, so I'd like to submit a PR for review. However, it might take me a little while to get approval for this license agreement from my employer.

gmlewis:
Thank you, @SudoVim .

I'll open this up to any volunteer contributor so that you don't have to worry about making the PR yourself.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started.

Thank you!

chughpiyush:
@gmlewis I can take up this one :)

gunadhya:
Hi @gmlewis, I'd like to work on this issue if it is available?

gmlewis:
Thanks, @gunadhya , but this issue is already assigned to @chughpiyush , so let's give an opportunity for a status update first.

chughpiyush:
@gmlewis I'll be giving a PR in a day or two.

alexal:
I haven't seen a PR yet, so I've decided to make a change myself. Please see PR #1798 and assign the issue to me. Thank you!

gmlewis:
> I haven't seen a PR yet, so I've decided to make a change myself. Please see PR #1798 and assign the issue to me. Thank you!

As described in #1798, this is not our standard way of proceeding (since the issue is already assigned), but in this case I'll allow it if you fix the PR problem, since it is a trivial change.

gmlewis:
Fixed by #1798.

PR #1798, opened 2021-02-11
Add missing fields to MarketplacePlan and MarketplacePurchase structs

alexal:
Hello,

As described in issue #1775, I've added missing fields to MarketplacePlan and MarketplacePurchase structs.

Thanks,
Alex

### DIFF
--- a/github/apps_marketplace.go
+++ b/github/apps_marketplace.go
@@ -30,6 +30,7 @@ type MarketplacePlan struct {
 	URL                 *string `json:"url,omitempty"`
 	AccountsURL         *string `json:"accounts_url,omitempty"`
 	ID                  *int64  `json:"id,omitempty"`
+	Number              *int    `json:"number,omitempty"`
 	Name                *string `json:"name,omitempty"`
 	Description         *string `json:"description,omitempty"`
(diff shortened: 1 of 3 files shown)


gmlewis (changes requested):
Thank you, @alexal , but this PR has a couple problems.
First, the issue was assigned to another person, and we try to avoid duplicating effort... however, I see in this case that it is a trivial change and no progress has been made by the other two parties, so if you can fix the other problem with this PR, we will move forward with it.

The second problem is that you need to run `go generate ./...` as described in our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) file.

Once you do this, the tests should pass and we can move forward with this PR.

Thanks.

gmlewis:
And in the future (also listed in our CONTRIBUTING.md file - search for # 7)... please do not use force push on PRs for this repo, as it makes the reviews much more difficult for the reviewers to keep track of changes.

Obviously, for this PR, it doesn't matter, but just please keep this in mind for future contributions.

Thank you.

gmlewis (approved):
Thank you, @alexal !
LGTM.
Merging.

alexal:
@gmlewis thank you. I've made all the required changes and all the tests were completed successfully.

PR #1799, opened 2021-02-12
Add JSON marshaling tests

renjithr23:
Moved JSON marshaling test for type Reviewers to appropriate test file. 

Add  JSON marshaling test for type ReviewersRequest

helps - #55


github/pulls_reviewers_test.go
```diff
@@ -13,6 +13,110 @@ import (
 	"testing"
 )
 
+func TestReviewersRequest_marshall(t *testing.T) {
```
gmlewis, line 16: While you are here, can you please change this from `marshall` to `marshal` ?

github/pulls_reviewers_test.go
```diff
@@ -13,6 +13,110 @@ import (
...
+		],
+		"team_reviewers" : [
+			"t"
+		]
+	}`
+
+	testJSONMarshal(t, u, want)
+}
+
+func TestReviewers_marshall(t *testing.T) {
```
gmlewis, line 38: Same here... can you please change this from `marshall` to `marshal` ?

gmlewis (changes requested):
Thank you, @renjithr23 !
Just a couple minor tweaks, please, while you are making this change, if you don't mind.

github/pulls_reviewers_test.go
renjithr23: Made the change

gmlewis (approved):
Thank you, @renjithr23 .
In the future, please don't use force push for PRs in this repo, as it makes it difficult for reviewers to see what changed since the last review.  In this case, it's not a big deal.
LGTM.
Merging.

renjithr23:
@gmlewis what is the ideal way for making changes to PRs ? Is it ok to add a new commit , even if it is for a small change to a PR ?

gmlewis:
Hi @renjithr23 - yes, it is totally fine in this repo to push as many commits as you want to a PR... this is because we "squash and merge" each PR so that it ends up being a single commit. You can read more about this in our CONTRIBUTING.md doc.

Issue #1802, opened 2021-02-16
InstallationPermissions model is missing Actions field


gsavit:
The `InstallationPermissions` struct lacks a property for [permissions on `Actions`](https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-actions)

Planning on forking and opening a PR with a fix, unless I'm missing something??

PR #1803, opened 2021-02-16
Add Actions field to InstallationPermissions model

gsavit:
Addresses https://github.com/google/go-github/issues/1802

### DIFF
--- a/github/apps_test.go
+++ b/github/apps_test.go
@@ -90,6 +90,7 @@ func TestAppsService_ListInstallations(t *testing.T) {
                                    "target_id":1,
                                    "target_type": "Organization",
                                    "permissions": {
+									   "actions": "read",
                                        "administration": "read",
                                        "checks": "read",
                                        "contents": "read",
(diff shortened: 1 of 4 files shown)


github/apps_test.go
```diff
@@ -90,6 +90,7 @@ func TestAppsService_ListInstallations(t *testing.T) {
                                    "target_id":1,
                                    "target_type": "Organization",
                                    "permissions": {
+									   "actions": "read",
```
gmlewis, line 93: Could you please make this line match the indentation of the following line (i.e. remove the tabs and replace with the same number of spaces in line 94)?

gmlewis (changes requested):
Thank you, @gsavit !
Just one minor tweak, please, and then I think we will be good to merge.

github/apps_test.go
gsavit: hmm yeh I noticed this, formatted multiple times and no change.. thought it might be weird gh rendering?

gsavit: nevermind, fixed, sorry!

gmlewis (approved):
Thank you, @gsavit !
LGTM.
Merging after tests complete.

PR #1605, opened 2020-08-15
Support Go 1.15 and 1.14 in workflows/tests.yml

gmlewis:
Since Go 1.15 has been released, and we want to test the two most recent major releases of Go on this repo, this PR bumps workflows/tests.yml to test the most recent (1.x) version of Go, and the next-most-recent (1.14.x) version.

PR #1807, opened 2021-02-21
Support Go 1.16 and 1.15 in workflows/tests.yml 

andygrunwald:
**What type of PR is this?**

improvement

**What this PR does / why we need it**:

go v1.16 is released.
go-github still tests against v1.14.

This PR raises the go version to v1.16 to test against.

**Special notes for your reviewer**:

None

**Does this PR introduce a user-facing change?**:

None

**Additional documentation e.g., usage docs, etc.**:

- [Go 1.16 is released)](https://blog.golang.org/go1.16)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.x, 1.14.x]
+        go-version: [1.x, 1.16.x]
         platform: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ jobs:
   test:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


gmlewis (changes requested):
Thank you, @andygrunwald , but this is not quite what we want.
Check out #1605 to see how we went from Go 1.14 to Go 1.15.
The `1.x` covers the most recent supported version of Go, and the second one should be changed to `1.15.x` to explicitly support the prior version of Go.
I know it is confusing... I had challenges with this too.

gmlewis:
> **What type of PR is this?**
> 
> bugfix

And technically, this is not correct.  This is not a bugfix... just for the record.

andygrunwald:
Thanks @gmlewis for the guidance. I pushed a fix in https://github.com/google/go-github/pull/1807/commits/5efeff8c70e43adf27186381c14e2a084a85f33f.

And indeed, I changed the type "bugifx" to "improvement".

gmlewis (approved):
LGTM.
Merging.

PR #1780, opened 2021-01-22
Repository.List: Add nebula-preview to receive visibility field

andygrunwald:
**What type of PR is this?**

feature

**What this PR does / why we need it**:

When calling `Repositories.List` (API Method: [List repositories for the authenticated user](https://docs.github.com/en/rest/reference/repos#list-repositories-for-the-authenticated-user)), the field `visibility` is always `nil`.

The `visibility` was introduced in a developer / API preview controlled by the media accept header. 
This PR enabled the API preview `nebula-preview` for `Repositories.List`, to retrieve the `visibility` of a repository.

**Special notes for your reviewer**:

None

**Does this PR introduce a user-facing change?**:

None

**Additional documentation e.g., usage docs, etc.**:

- [List repositories for the authenticated user @ GitHub REST API docs](https://docs.github.com/en/rest/reference/repos#list-repositories-for-the-authenticated-user): In the response, the `visibility` field is available.
- [API changes to support internal repository visibility (2019-12-03)](https://developer.github.com/changes/2019-12-03-internal-visibility-changes/#repository-visibility-fields)

PR #1808, opened 2021-02-21
Fix inline docs for InstallationEvent type

andygrunwald:
**What type of PR is this?**

documentation

**What this PR does / why we need it**:

The inline documentation for the `InstallationEvent` went out of date.
This PR fixes the link + possible values.

Furthermore, does the REST API send a `requester` key as well.
In my current use-case, this was `null`, hence I am not aware of the type (is not documented).
As a trade-off for now, I added a TODO. Once the type is known, this can be extended

**Special notes for your reviewer**:

This applies the same fix as in https://github.com/google/go-github/pull/1780

**Does this PR introduce a user-facing change?**:

None

**Additional documentation e.g., usage docs, etc.**:

- [Webhook events and payloads: installation](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#installation)

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -277,16 +277,18 @@ type TeamChange struct {
 	} `json:"repository,omitempty"`
 }
 
-// InstallationEvent is triggered when a GitHub App has been installed or uninstalled.
+// InstallationEvent is triggered when a GitHub App has been installed, uninstalled, suspend, unsuspended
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @andygrunwald !
LGTM.
Merging.

PR #1813, opened 2021-02-26
Add missing name field from workflow run

steveteuber:
Adds the missing name field from workflow runs endpoint.

See docs: https://docs.github.com/en/rest/reference/actions#get-a-workflow-run


gmlewis (approved):
Thank you, @steveteuber !
LGTM.
Merging.

Issue #1815, opened 2021-03-02
Not differentiating between blank string and null causes difficulties in merging pull requests


john-m-liu:
(Github API doc is at https://docs.github.com/en/rest/reference/pulls#merge-a-pull-request)

When making a request to merge a pull request, there is a difference between the following:
`curl -X PUT -H "Authorization: <token>"  https://api.github.com/repos/{owner}/{repo}/pulls/{pull_number}/merge -d '{"merge_method": "squash", "commit_title": "my commit", "commit_message": ""}'`
and
`curl -X PUT -H "Authorization: <token>"  https://api.github.com/repos/{owner}/{repo}/pulls/{pull_number}/merge -d '{"merge_method": "squash", "commit_title": "my commit"}'`
Specifically, the former will have no commit message, and the latter will have the default commit message (typically the concatenation of all commit messages in the pull request).

This differentiation isn't possible with the go-github library because the following struct has omitempty: https://github.com/google/go-github/blob/master/github/pulls.go#L435. The default commit message will always be used if you pass the blank string.

I believe this is a specific case of https://github.com/google/go-github/issues/19 but was not addressed

gmlewis:
One non-breaking-API-change solution would be to add a new endpoint, something like `MergeWithNoCommitMessage` or something like that.

Any other solution I can think of would involve a breaking API change, and additionally be rather ugly... like adding another parameter stating whether `CommitMessage` should be sent or not.

I'm leaning toward the former, but am not against the latter.

Thoughts?

john-m-liu:
> One non-breaking-API-change solution would be to add a new endpoint, something like `MergeWithNoCommitMessage` or something like that.
> 
> Any other solution I can think of would involve a breaking API change, and additionally be rather ugly... like adding another parameter stating whether `CommitMessage` should be sent or not.
> 
> I'm leaning toward the former, but am not against the latter.
> 
> Thoughts?

The latter feels a bit cleaner to me simply because it would avoid 2 public functions that do basically the same thing, but I don't really have a preference either way. I believe you could make the second option non-breaking by having the new parameter be in a struct field where false maintains the existing behavior

gmlewis:
> The latter feels a bit cleaner to me simply because it would avoid 2 public functions that do basically the same thing, but I don't really have a preference either way. I believe you could make the second option non-breaking by having the new parameter be in a struct field where false maintains the existing behavior

[...]

PR #1816, opened 2021-03-02
Support empty string as a commit message when merging a pull request

john-m-liu:
Fixes https://github.com/google/go-github/issues/1815

- changes the CommitMessage field in the request to merge a PR to a pointer
- adds a new boolean flag to the PullRequestOptions struct, which will make the commit message null if set and the given message is ""

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -458,8 +458,7 @@ func (s *PullRequestsService) Merge(ctx context.Context, owner string, repo stri
 		pullRequestBody.MergeMethod = options.MergeMethod
 		pullRequestBody.SHA = options.SHA
 		if options.DontDefaultIfBlank && commitMessage == "" {
-			emptyMessage := ""
-			pullRequestBody.CommitMessage = &emptyMessage
+			pullRequestBody.CommitMessage = &commitMessage
 		}
 	}
 	req, err := s.client.NewRequest("PUT", u, pullRequestBody)
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -430,13 +430,16 @@ type PullRequestOptions struct {
 
 	// The merge method to use. Possible values include: "merge", "squash", and "rebase" with the default being merge. (Optional.)
 	MergeMethod string
+
+	// If false, an empty string commit message will use the default commit message. If true, an empty string commit message will be used
+	DontDefaultIfBlank bool
 }
 
 type pullRequestMergeRequest struct {
-	CommitMessage string `json:"commit_message,omitempty"`
-	CommitTitle   string `json:"commit_title,omitempty"`
-	MergeMethod   string `json:"merge_method,omitempty"`
-	SHA           string `json:"sha,omitempty"`
+	CommitMessage *string `json:"commit_message,omitempty"`
+	CommitTitle   string  `json:"commit_title,omitempty"`
+	MergeMethod   string  `json:"merge_method,omitempty"`
+	SHA           string  `json:"sha,omitempty"`
 }
 
 // Merge a pull request.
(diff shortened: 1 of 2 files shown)


john-m-liu:
@googlebot I signed it!

github/pulls_test.go
```diff
@@ -805,3 +805,32 @@ func TestPullRequestsService_Merge_options(t *testing.T) {
...
+	madeRequest := false
+	expectedBody := ""
+	mux.HandleFunc("/repos/o/r/pulls/1/merge", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		testBody(t, r, expectedBody+"\n")
+		madeRequest = true
+	})
+
+	ctx := context.Background()
+	expectedBody = `{"commit_message":""}`
```
gmlewis, line 821: According to #1815, this will have no commit message, which is the case that I thought you added the `DontDefaultIfBlank` flag for.

github/pulls_test.go
```diff
@@ -805,3 +805,32 @@ func TestPullRequestsService_Merge_options(t *testing.T) {
...
+	_, _, _ = client.PullRequests.Merge(ctx, "o", "r", 1, "", nil)
+	if !madeRequest {
+		t.Error("TestPullRequestsService_Merge_Blank_Message #1 did not make request")
+	}
+
+	madeRequest = false
+	opts := PullRequestOptions{
+		DontDefaultIfBlank: true,
+	}
+	expectedBody = `{}`
```
gmlewis, line 831: According to #1815, this will use a default commit message, which I don't think is what you want when using the new `DontDefaultIfBlank` flag.

gmlewis (changes requested):
Thank you, @john-m-liu , but I don't think this is working the way you want it to.  See comments below.

To solve the problem, I believe you may need to use an additional helper struct that simply removes the `omitempty` JSON tag for the `CommitMessage` when using the new flag.

Note that you need to keep the original one and then switch over to the new one internally for the new special case with the flag and the empty commit message.

github/pulls.go
```diff
@@ -430,13 +430,16 @@ type PullRequestOptions struct {
 
 	// The merge method to use. Possible values include: "merge", "squash", and "rebase" with the default being merge. (Optional.)
 	MergeMethod string
+
+	// If false, an empty string commit message will use the default commit message. If true, an empty string commit message will be used
```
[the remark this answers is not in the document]
john-m-liu: fixed

github/pulls_test.go
john-m-liu: doh, I inverted the boolean 🤦 fixed this in the function body, which I felt was preferable to creating a duplicate struct

github/pulls_test.go
john-m-liu: fixed

github/pulls.go
```diff
@@ -449,13 +449,17 @@ type pullRequestMergeRequest struct {
...
+	if commitMessage != "" {
+		pullRequestBody.CommitMessage = &commitMessage
+	}
 	if options != nil {
 		pullRequestBody.CommitTitle = options.CommitTitle
 		pullRequestBody.MergeMethod = options.MergeMethod
 		pullRequestBody.SHA = options.SHA
 		if options.DontDefaultIfBlank && commitMessage == "" {
-			pullRequestBody.CommitMessage = nil
+			emptyMessage := ""
```
gmlewis, line 461: There's no need for a new variable here since we know at this point that `commitMessage == ""`.

john-m-liu: fixed

gmlewis (approved):
Thank you, @john-m-liu !
LGTM.

wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1818, opened 2021-03-04
Add reactions preview to search service

breathingdust:
Currently the search service does not include the reactions preview media type header so issue searches do not include the reaction counts this preview provides. The issue struct already includes this [attribute](https://github.com/google/go-github/blob/b06cfbb1abc090e2ff47cac2f0d2d1750f847293/github/issues.go#L52) so adding this preview header is enough to populate it.

This PR adds that preview so this structure is populated.

### DIFF
--- a/github/search.go
+++ b/github/search.go
@@ -284,6 +284,10 @@ func (s *SearchService) search(ctx context.Context, searchType string, parameter
 		// Accept header for search repositories based on topics preview endpoint
 		// TODO: remove custom Accept header when this API fully launches.
 		req.Header.Set("Accept", mediaTypeTopicsPreview)
+	case searchType == "issues":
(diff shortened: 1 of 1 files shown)


gmlewis (changes requested):
Thank you, @breathingdust !

Could you please add two new test methods in the `github/search_test.go` file to cover the new case?

* `TestSearchService_Issues`
* `TestSearchService_Issues_coverage`

gmlewis (approved):
Oh, never mind... I see we have these already.  Thanks.

LGTM.

wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1820, opened 2021-03-09
Audit Log REST API(Beta) Implementation

sridharavinash:
This PR implements the new AuditLog REST API Beta which was GA-ed recently. 

You can read more about it here: 
https://github.blog/changelog/2021-02-03-audit-log-git-events-and-rest-api-are-now-in-public-beta/

Most notably it exposes 2 new endpoints for Enterprises Accounts only 
* `orgs/<org>/audit-log`
* `enterprises/<enterprise>/audit-log` 

In this PR
* Extend `OrganizationsService` to implement `GetAuditLog` call
* Extend `EnterpriseService` to implement `GetAuditLog` call
* Tests for both services

I chose to put these into 2 separate files because they define 2 separate endpoints (I can be persuaded to have this live in one file if it makes it easier). 
API reference docs : https://docs.github.com/en/rest/reference/orgs#get-the-audit-log-for-an-organization


gmlewis:
Thank you, @sridharavinash - it looks like you might need to run `go generate ./...` as described in our CONTRIBUTING.md guide.

Hey, by the way... do you know what happened to this blog site: https://developer.github.com/changes/ ?  I haven't seen any changes on it since October 1, 2020, and your PR is the first I've seen of a different blog site URL.

sridharavinash:
>Thank you, @sridharavinash - it looks like you might need to run go generate ./... as described in our CONTRIBUTING.md guide.

@gmlewis 👋🏾 . I think this was `fmt` failure. Hopefully the last commit fixes that. 

>Hey, by the way... do you know what happened to this blog site: https://developer.github.com/changes/ ? I haven't seen any changes on it since October 1, 2020, and your PR is the first I've seen of a different blog site URL.

Good question. I'll find out more if that's being deprecated. Maybe in favor of OpenAPI descriptions 
 - https://github.com/github/rest-api-description 🤔 .

github/enterprise_audit_log.go
```diff
@@ -0,0 +1,35 @@
...
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// GetAuditLog gets the audit-log entries for an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/orgs#get-the-audit-log-for-an-organization
```
gmlewis, line 15: I think this URL should be different for the Enterprise version.

github/orgs_audit_log.go
```diff
@@ -0,0 +1,106 @@
...
+	Phrase  *string `json:"phrase,omitempty"`   // A search phrase. (Optional.)
+	Include *string `json:"include,omitempty"`  // Event type includes. Can be one of "web", "git", "all". Default: "web". (Optional.)
+	After   *string `json:"after,omitempty"`    // A cursor, as given in the Link header. If specified, the query only searches for events after this cursor. (Optional.)
+	Before  *string `json:"before,omitempty"`   // A cursor, as given in the Link header. If specified, the query only searches for events before this cursor. (Optional.)
+	Order   *string `json:"order,omitempty"`    // The order of audit log events. Can be one of "asc" or "desc". Default: "desc". (Optional.)
+	PerPage *int64  `json:"per_page,omitempty"` // Results per page (max 100).
+}
+
+// AuditEntry describes the fields that may be represented by various audit-log "action" entries.
+type AuditEntry struct {
```
gmlewis, line 24: I'm extremely confused at the moment.
When I look at the examples listed here:
https://docs.github.com/en/rest/reference/orgs#get-the-audit-log-for-an-organization

I see three audit entry examples, with just a tiny handful of fields.

Where is this huge list of fields coming from?

gmlewis (changes requested):
Thank you, @sridharavinash !
Please see comments below.

github/enterprise_audit_log.go
sridharavinash: Fixed, I had that as a placeholder till I found the real link.

github/orgs_audit_log.go
sridharavinash: We have an internal (not yet public) representation of the these exposed API fields for the 600+ events that the audit-log supports.  Some of these are defined here: https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/reviewing-the-audit-log-for-your-organization#audit-log-actions (not all these actions will be available in an org/enterprise setting).

sridharavinash: Thinking out aloud should we rather have these be more tightly coupled to the Action and then do a lookup based on `action` or maybe embed other Actions event fields?

so instead of 
```
type AuditEntry struct {
     <smattering of fields>
} 
```
do something more akin to 
```
type AuditEntry struct{
    DocumentID *string `json:"_document_id,omitempty"`
    Timestamp *int64 `json:"*timestamp,omitempty"`

    GitEventAuditEntry
    TeamEventAuditEntry
}

type GitEventAuditEntry struct {
   TransportProtcol *string `json:"transport_protocol,omitempty"`
    ....
}
```

github/org_audit_log_test.go
```diff
@@ -0,0 +1,93 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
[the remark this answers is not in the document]
sridharavinash: fixed in https://github.com/google/go-github/pull/1820/commits/1d91016ead0af1f0a964de0841fcaef68c64ffe3

github/orgs_audit_log.go
```diff
@@ -0,0 +1,106 @@
...
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// GetAuditLogOptions sets up optional parameters to query audit-log endpoint.
+type GetAuditLogOptions struct {
```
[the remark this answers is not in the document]
sridharavinash: @gmlewis Would you recommend that I move `After` and `Before` to `ListCursorOptions` 
IOW
```
type GetAuditLogOptions struct {
	Phrase  *string `json:"phrase,omitempty"`   // A search phrase. (Optional.)
	Include *string `json:"include,omitempty"`  // Event type includes. Can be one of "web", "git", "all". Default: "web". (Optional.)
	Order   *string `json:"order,omitempty"`    // The order of audit log events. Can be one of "asc" or "desc". Default: "desc". (Optional.)

         // Pagination Options
	ListCursorOptions
}
```

and update 
```
// ListCursorOptions specifies the optional parameters to various List methods that
// support cursor pagination.
type ListCursorOptions struct {
	// For paginated result sets, page of results to retrieve.
	Page string `url:"page,omitempty"`

	// For paginated result sets, the number of results to include per page.
        
        // A cursor, as given in the Link header. If specified, the query only searches for events after this cursor. (Optional.)
	After   string `url:"after,omitempty"`    

        // A cursor, as given in the Link header. If specified, the query only searches for events before this cursor. (Optional.)
	Before  string `url:"before,omitempty"`   

}
```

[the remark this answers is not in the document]
gmlewis: @sridharavinash - in the past, I think we decided that the `After` and `Before` URLs were not useful to users of this client library, so we have explicitly not added them.

So unless you have any objections, I think we can just drop them.

github/orgs_audit_log.go
gmlewis: Hmm... great question.  I'm fine with making one huge struct like we do for some of the other responses, but I also see the value of your idea...

I suppose we might want to consider how challenging one vs the other might be for maintenance of this repo as things change on the GitHub side.

How about you implement what makes sense to you, then we can see what it looks like and others can comment if they have an opinion as well.

github/orgs_audit_log.go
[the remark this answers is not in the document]
sridharavinash: Unfortunately for the AuditLog API we don't have a `page` param, we exclusively use `before` / `after` link `rel` headers to navigate pages. Using the API call with `PerPage` will probably be limited to just viewing one page without any forward/backwards navigation of results. 

I can drop it if that's preferred, it can always be added if someone needs it.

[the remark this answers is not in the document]
gmlewis: Ah!  OK, @sridharavinash - then in that case if it is going to be useful, go ahead and add it.

Now the question is whether or not it belongs in `ListCursorOptions`, and I'm guessing the answer is "yes", so go ahead and include it.

github/orgs_audit_log.go
sridharavinash: I went down this path of separating the fields into logical structs and decided against it only because I think discoverability of the fields might suffer and not all fields map 1:1 to an audit-action event. Instead I sorted the field list within here so that it's easily discoverable and gives the user a comprehensive look of what can be expected back.

github/orgs_audit_log.go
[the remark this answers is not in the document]
sridharavinash: added in https://github.com/google/go-github/pull/1820/commits/1814994b6d8b1ed35ae0ae09b6a5e34418c8f2f1

github/orgs_audit_log.go
```diff
@@ -69,16 +70,14 @@ type AuditEntry struct {
 	TargetLogin          *string    `json:"target_login,omitempty"`
 	TargetVersion        *string    `json:"target_version,omitempty"`
 	Team                 *string    `json:"team,omitempty"`
+	Timestamp            *int64     `json:"@timestamp,omitempty"`
+	TransportName        *string    `json:"transport_protocol_name,omitempty"`
```
gmlewis, line 74: To minimize confusion and maximize consistency, could we please change this from `TransportName` to `TransportProtocolName`?

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	"context"
+	"fmt"
+)
+
+// GetAuditLogOptions sets up optional parameters to query audit-log endpoint.
+type GetAuditLogOptions struct {
+	Phrase  *string `json:"phrase,omitempty"`  // A search phrase. (Optional.)
+	Include *string `json:"include,omitempty"` // Event type includes. Can be one of "web", "git", "all". Default: "web". (Optional.)
+	Order   *string `json:"order,omitempty"`   // The order of audit log events. Can be one of "asc" or "desc". Default: "desc". (Optional.)
+	ListCursorOptions
```
gmlewis, line 18: Could you please add a blank line above `ListCursorOptions`?

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	Config               *string    `json:"config,omitempty"`
+	ConfigWas            *string    `json:"config_was,omitempty"`
+	ContentType          *string    `json:"content_type,omitempty"`
+	CreatedAt            *int64     `json:"created_at,omitempty"`
+	DeployKeyFingerprint *string    `json:"deploy_key_fingerprint,omitempty"`
+	DocumentID           *string    `json:"_document_id,omitempty"`
+	Emoji                *string    `json:"emoji,omitempty"`
+	EnvironmentName      *string    `json:"environment_name,omitempty"`
+	Event                *string    `json:"event,omitempty"`
+	Events               *string    `json:"events,omitempty"`
```
gmlewis, line 42: Just FYI - I'm going to just trust you on the types since this is not officially documented yet.
As usual, we can always update these when they change.

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	Org                  *string    `json:"org,omitempty"`
+	PreviousVisibility   *string    `json:"previous_visibility,omitempty"`
+	ReadOnly             *string    `json:"read_only,omitempty"`
+	Repo                 *string    `json:"repo,omitempty"`
+	Repository           *string    `json:"repository,omitempty"`
+	RepositoryPublic     *bool      `json:"repository_public,omitempty"`
+	RunnerGroupID        *string    `json:"runner_group_id,omitempty"`
+	RunnerGroupName      *string    `json:"runner_group_name,omitempty"`
+	RunnerID             *string    `json:"runner_id,omitempty"`
+	RunnerLabels         *[]string  `json:"runner_labels,omitempty"`
```
gmlewis, line 65:
```suggestion
	RunnerLabels         []string  `json:"runner_labels,omitempty"`
```

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	RunnerName           *string    `json:"runner_name,omitempty"`
+	SecretsPassed        *[]string  `json:"secrets_passed,omitempty"`
```
gmlewis, line 67:
```suggestion
	SecretsPassed        []string  `json:"secrets_passed,omitempty"`
```

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	SourceVersion        *string    `json:"source_version,omitempty"`
+	StartedAt            *Timestamp `json:"started_at,omitempty"`
+	TargetLogin          *string    `json:"target_login,omitempty"`
+	TargetVersion        *string    `json:"target_version,omitempty"`
+	Team                 *string    `json:"team,omitempty"`
+	Timestamp            *int64     `json:"@timestamp,omitempty"`
+	TransportName        *string    `json:"transport_protocol_name,omitempty"`
+	TransportProtocol    *int64     `json:"transport_protocol,omitempty"`
```
gmlewis, line 75: `TransportProtocol` is really an `*int64` ?  That seems really odd to me.

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	RunnerGroupName      *string    `json:"runner_group_name,omitempty"`
+	RunnerID             *string    `json:"runner_id,omitempty"`
+	RunnerLabels         *[]string  `json:"runner_labels,omitempty"`
+	RunnerName           *string    `json:"runner_name,omitempty"`
+	SecretsPassed        *[]string  `json:"secrets_passed,omitempty"`
+	SourceVersion        *string    `json:"source_version,omitempty"`
+	StartedAt            *Timestamp `json:"started_at,omitempty"`
+	TargetLogin          *string    `json:"target_login,omitempty"`
+	TargetVersion        *string    `json:"target_version,omitempty"`
+	Team                 *string    `json:"team,omitempty"`
```
gmlewis, line 72: A `Team` doesn't point to a `*Team`?  It's really just a string?

github/orgs_audit_log.go
```diff
@@ -0,0 +1,105 @@
...
+	Timestamp            *int64     `json:"@timestamp,omitempty"`
+	TransportName        *string    `json:"transport_protocol_name,omitempty"`
+	TransportProtocol    *int64     `json:"transport_protocol,omitempty"`
+	TriggerID            *int64     `json:"trigger_id,omitempty"`
+	User                 *string    `json:"user,omitempty"`
```
gmlewis, line 77: A `User` doesn't point to a `*User` and is just a string?

gmlewis (changes requested):
Thank you, @sridharavinash !
I have a few questions, and a few minor tweaks, please.

github/orgs_audit_log.go
sridharavinash: good catch! I think this can be just `*int`

github/orgs_audit_log.go
sridharavinash: that's correct, it's just a string of the team name. Since the Audit Log is mostly immutable logs, we don't store full object details (in the event that the team / repo / user account has been deleted for example).

github/orgs_audit_log.go
sridharavinash: Same as https://github.com/google/go-github/pull/1820#discussion_r593277513

sridharavinash:
I'm going to update this PR with a little more detail and add documentation of the returned types where possible.

sridharavinash:
I made a few changes and addressed some of the feedback. Let me know how this is looking @gmlewis . 🙇🏾‍♂️

gmlewis (approved):
Thank you, @sridharavinash !
LGTM.

wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp .
Merging.

PR #1822, opened 2021-03-10
tidy up GitHub actions config

willnorris:
- only run with a single Go version on Windows, since we're just testing
  basic things like path separators
- only upload code coverage stats once, not for every build
  configuration
- use latest version of actions/cache


gmlewis (approved):
Thank you, @willnorris !
LGTM

PR #1828, opened 2021-03-25
Interfaces go:generate

migueleliasweb:


migueleliasweb:
@googlebot I signed it!

migueleliasweb:
I just found out that many of the `Services` like `RepositoriesService` are split in multiple files. Have said that, this might be a bit more challenging that I first thought as the tool that I used doesn't seem to support multiple files to generate the `interface`. =/

I will come up with something, don't worry. 🚀

gmlewis (changes requested):
One other requirement... please do not pull in any external dependencies for ease of maintainability of this repo.

Feel free to close this and assign the issue back to me, and I will implement it.

gmlewis:
Closing this PR as we will not be pulling in an external resource when we have an alternative implementation (#1832) that does not require any.

Issue #1280, opened 2019-09-07
Discussion: release cadence


willnorris:
gmlewis@: Looking at https://github.com/google/go-github/releases, it seems like you've been tagging a new release nearly every time you merge a pull request, is that right?  That's really not necessary, and we should probably slow it down a bit.  It creates a bunch of extra work (for you!) and noise to tag releases that end up getting newer releases tagged hours or even minutes later.  It also makes it much harder for users to stay up to date with the latest version, if that is their goal.

I totally appreciate you continuing to help us keep the library going!!  But I'd like to see if we can find a release cadence that's more on the order of every week or two, with exception made as needed for critical bug fixes and such.

What do you think?

/cc @gauntface as well if you have thoughts

gmlewis:
OK, that sounds great.  I will stop releasing on every PR and will work on creating one every week or two.  Thanks, @willnorris.

willnorris:
thank *you* :)

chapurlatn:
Hey guys, 
I also would expect much more release. But I fully understand @willnorris arguments.
May be there may be some way to publish two channels for release:
- a stable with monthly release.
- a separated beta one that may include a release for each merged issue (for example, starting for version v58.0.0, it may generate some v58.0.1-beta)?

PR #1830, opened 2021-03-25
Add information about release cadence to README.md

gmlewis:
With the recent requests for new releases, it would be nice to provide this FAQ in the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,10 @@ Preview functionality may take the form of entire methods or simply additional
 data returned from an otherwise non-preview method. Refer to the GitHub API
 documentation for details on preview functionality.
 
+Please note that as of #1280, all releases (major, minor, or patch) are provided on an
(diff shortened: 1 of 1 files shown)


willnorris:
are we just doing on-demand releases?  I think the point I was making in #1280 was just about having a more predictable release schedule, not abandoning scheduled releases at all.  That said, you're the one doing the bulk of the lifting here, so it's up to you if you want to try and do scheduled releases.

The most important part is that we set expectations appropriately, which this certainly does if this is the plan going forward.

gmlewis:
Well, it certainly is easier to perform on-demand releases from a maintenance perspective.

If you would prefer scheduled releases, what kind of schedule are you thinking?

willnorris:
4-6 weeks?

willnorris:
or even longer if there's not much new

gmlewis:
OK, sure, let's try that.  I'll close this PR.

PR #1831, opened 2021-03-25
Remove obsolete roadmap spreadsheet from README.md

gmlewis:
The roadmap spreadsheet has not been used lately and I believe it would improve the README.md to simply remove it to reduce confusion.


willnorris (approved):
totally agree, thanks!

PR #1832, opened 2021-03-25
Auto-generate interfaces for all services

### LINKED ISSUE #1800: Add Client interface/mock to allow Mocked testing
Right now, the Github client is very hard to use for unit tests for projects depending on this.

It would be a great idea to create an Interface for the client. Much like the K8S client mocked client, in the Go client project.

gmlewis:
Fixes #1800.


gmlewis:
@willnorris - assuming you are OK with this PR, would you prefer for me to include it into the imminent `v34.0.0` release or would you prefer I keep it out?

willnorris:
I think putting this into the main github package is a really bad idea.  I'm still not wild about adding it at all, but understand the arguments being made in #1800.  Would putting these into a separate test related package still serve the same purpose for clients to more easily test?

gmlewis:
OK, thank you, @willnorris - then I'll move ahead with `v34.0.0` as we discuss whether to do #1800 or not at all.

gmlewis:
> Would putting these into a separate test related package still serve the same purpose for clients to more easily test?

That's a great idea, but I don't know the answer.  Let's see what the responses are on #1800.

migueleliasweb:
> I think putting this into the main github package is a really bad idea. I'm still not wild about adding it at all, but understand the arguments being made in #1800. Would putting these into a separate test related package still serve the same purpose for clients to more easily test?

In order to allow users of this lib to properly use the interfaces to improve the current testing/mocking situation we might need to also replace all types in `github/github.go.Client` properties like I was trying to do on my PR (https://github.com/google/go-github/pull/1828). Without this the interfaces wouldn't yield much effect, I reckon.

Thoughts?

migueleliasweb:
Ps: @gmlewis beat me to it! Haha! Nice implementation! It's just a bit sad we can't/not supposed to use external dependencies here =(.

migueleliasweb:
It would also we great if we could add some testing examples in `examples/` to show some best practices (or intended ways of testing when using this lib).

gmlewis:
> In order to allow users of this lib to properly use the interfaces to improve the current testing/mocking situation we might need to also replace all types in github/github.go.Client properties like I was trying to do on my PR (#1828). Without this the interfaces wouldn't yield much effect, I reckon.

I'm not sure I understand this statement.

> It would also we great if we could add some testing examples in `examples/` to show some best practices (or intended ways of testing when using this lib).

Great idea, if we choose to pursue this.  I'll add a simple example that demonstrates the concept of embedding the interface.

gmlewis:
@migueleliasweb - I've gone ahead and added a simple example... but now that I've done that, I'm still not convinced that this is of any benefit whatsoever, as it is just as simple to mock only the one method that I need.

What am I missing?!?  What is so great about embedding an interface into a struct?

migueleliasweb:
> @migueleliasweb - I've gone ahead and added a simple example... but now that I've done that, I'm still not convinced that this is of any benefit whatsoever, as it is just as simple to mock only the one method that I need.
> 
> What am I missing?!? What is so great about embedding an interface into a struct?

Oh I see you example, yeah, that's not quite what I tried to do. I will create a gist with the test case I had to run. I'm stuck in a meeting rn, so this might take a lil bit. 😞

migueleliasweb:
I've created a PR to @gmlewis PR (PR-ception I guess?) just to provide a better example of the implementation I envisioned.

https://github.com/gmlewis/go-github/pull/1

github/github.go
```diff
@@ -154,32 +155,32 @@ type Client struct {
...
+	CodeScanning   CodeScanningServiceInterface
+	Enterprise     EnterpriseServiceInterface
+	Gists          GistsServiceInterface
+	Git            GitServiceInterface
+	Gitignores     GitignoresServiceInterface
+	Interactions   InteractionsServiceInterface
+	IssueImport    IssueImportServiceInterface
+	Issues         IssuesServiceInterface
+	Licenses       LicensesServiceInterface
 	Marketplace    *MarketplaceService
```
migueleliasweb, line 174: Just a reminder: There's a bit of work needed on https://github.com/gmlewis/go-github/blob/master/github/apps_marketplace_test.go#L62 before changing this to also use the interface. I tried to understand why that property is needed but I guess I don't have enough context on why.

gmlewis: > I tried to understand why that property is needed but I guess I don't have enough context on why.

Here is the context: https://github.com/google/go-github/pull/732#issuecomment-333726743

migueleliasweb: I will have a read, thanks!

migueleliasweb: After reading on that issue and also having a quick read on GitHub's Api, would there be any objections on adding `get ` and `set` methods for `s.Stubbed`? This would be a small change and we would still be able to use the interfaces as long as we regenerate `github-interfaces.go`. I guess the other option would be a switch type inside some of the methods from `MarketplaceService` to distinguish between the real implementation and the fake ones, but I'm not sure if that's preferable over the simple `get` and `set`.

migueleliasweb: This is PR with the getter and setter implementation: https://github.com/gmlewis/go-github/pull/2. Let me know your thoughts. :+1:

gmlewis: > This is PR with the getter and setter implementation: [gmlewis#2](https://github.com/gmlewis/go-github/pull/2). Let me know your thoughts.

That's probably the best way to proceed.

github/github_test.go
```diff
@@ -705,7 +704,7 @@ func TestDo_nilContext(t *testing.T) {
 	req, _ := client.NewRequest("GET", ".", nil)
 	_, err := client.Do(nil, req, nil)
 
-	if !reflect.DeepEqual(err, errors.New("context must be non-nil")) {
+	if err.Error() != "context must be non-nil" {
```
rspier, line 707: consider creating a 

```
var errNonNilContext = errors.New("context must be non-nil")
```
in github.go and then use `errors.Is()` here.

gmlewis: Taking a look at: https://blog.golang.org/go1.13-errors ...

If we do this, would we need to change the README.md such that this repo would then require Go 1.13 instead of Go 1.9?

Does anyone care about Go 1.9 compatibility anymore?

migueleliasweb: Since this repo supports go mods (which started on Go 1.11 if I'm not wrong) I think supporting Go 1.9 is...irrelevant?

gmlewis: OK, it's actually completely unrelated, so I think I'll just revert this change and open a new issue for it.

gmlewis: Here is the new issue: https://github.com/google/go-github/issues/1846

gmlewis:
Ah, right... @migueleliasweb - have you signed the CLAs?

github/github_test.go
gmlewis: Reverted this change.  Resolving conversation.

github/github.go
gmlewis: Stubbed methods incorporated.  Resolving conversation.

migueleliasweb:
> Ah, right... @migueleliasweb - have you signed the CLAs?

I reckon I had to sign when I created the other PR for this repo. Do I need to sign again for this one? 

This is the other PR i'm talking about, it says `cla:yes`: https://github.com/google/go-github/pull/1828

migueleliasweb:
> Ah, right... @migueleliasweb - have you signed the CLAs? I believe you have, but can you please officially respond that you are OK to include your changes into this PR?

It's all good from my end :+1: .

migueleliasweb:
@googlebot I consent.

gmlewis:
OK, I believe we now have a complete implementation with all issues resolved.

What would be really nice now is if a major user of this repo (Google? Twitter? Terraform? Disney? someone else?) could actually try out these changes and see if they are OK with the approach.

willnorris:
(Twitter is not a user of this package, so can't help there.  At some point, we may move to Terraform, in which case we'd be using it indirectly, but not currently.)

rspier:
I started an (internal) conversation about this, and have gotten some interesting feedback so far:

* Concern about this breaking *existing* tests if the interfaces are extended.
* Doing https://github.com/googleapis/google-api-go-client/blob/master/testing.md#testing-http-services-using-fakes instead of these interfaces.
* This quote: 
  > If I were going to make a change like this I think I would personally: keep returning structs, create a new test(service?) package where these interfaces are defined with clear documentation that they will break over time and examples of how to use them

That last one is particularly interesting.  I don't think we tried that exact approach.

I'll let you know if anything else comes along.


Regarding testing it, I'm swamped this week.  I'll try and poke at it time permitting.

migueleliasweb:
Those are some interesting points, @rspier .

> * Concern about this breaking _existing_ tests if the interfaces are extended.

Could you expand this idea a bit more? Would this be related to embedding interfaces to other interfaces?

> * Doing https://github.com/googleapis/google-api-go-client/blob/master/testing.md#testing-http-services-using-fakes instead of these interfaces.

As Glen mentioned before, it's not that the current package implemention was untestable but I guess the interfaces would just make it a bit more nice to write tests. With the interfaces a client of this package wouldn't have to deal with httptest.

> * This quote:
>   > If I were going to make a change like this I think I would personally: keep returning structs, create a new test(service?) package where these interfaces are defined with clear documentation that they will break over time and examples of how to use them
> 
> That last one is particularly interesting. I don't think we tried that exact approach.
> 
This is something we could think about doing. This is basically the way AWS Golang SDK does it as far as I can tell. They provide a "standard" implemention that is (at first glance) unrelated to the interfaces that are defined in a separate package. But in the end you still need to use the interfaces everytime you need to write any sort of mocking/testing in your code.
From my experience, I've found that Golang's interfaces concept doesn't play quite well when we access properties of a given struct directly ( like this pkg does in the main Client struct ). If we want to provide a seamless implemention of the Client that is also interfacable we might need to implement getter and setters for the properties where the internal "services" are defined, like in `Client.RepositoryService`.

rspier:
> Those are some interesting points, @rspier and thank you for sharing the conversation you've had in here.
> 
> > * Concern about this breaking _existing_ tests if the interfaces are extended.
> 
> Could you expand this idea a bit more? Would this be related to embedding interfaces to other interfaces?

Here's what was said:

> If $thing adds a new method it could start to break people as they might have written their own mock against this public interface and once a method gets added their mock would no longer satisfy the interface.

This is why local interfaces for only the used surface work well.  


> 
> > * Doing https://github.com/googleapis/google-api-go-client/blob/master/testing.md#testing-http-services-using-fakes instead of these interfaces.
> 
> As Glen mentioned before, it's not that the current package implemention was untestable but I guess the interfaces would just make it a bit more nice to write tests. With the interfaces a client of this package wouldn't have to deal with httptest.

I agree httptest feels heavy.  One approach might be to write some wrappers around httptest to simplify its use in our case.

> 
> > * This quote:
> >   > If I were going to make a change like this I think I would personally: keep returning structs, create a new test(service?) package where these interfaces are defined with clear documentation that they will break over time and examples of how to use them
> > 
> > That last one is particularly interesting. I don't think we tried that exact approach.
> 
> This is something we could think about doing. This is basically the way AWS Golang SDK does it as far as I can tell. They provide a "standard" implemention that is (at first glance) unaware of the interfaces that are defined in a separate package. But in the end you still need to use the interfaces everytime you need to write any sort of mocking/testing in your code.
> From my experience, I've found that Golang's interfaces concept doesn't play quite well when we access properties of a given struct directly ( like this pkg does in the main Client struct ). If we want to provide a seamless implemention of the Client that is also interfacable we might need to implement getter and setters for the properties where the internal "services" are defined, like in `Client.RepositoryService`.
> 
> Eg:
> 
> ```go
> // from README.md
> client := github.NewClient(nil)
> orgs, _, err := client.Organizations.List(context.Background(), "willnorris", nil)
> 
> // If we don't return `client.Organizations` as an interface (OrganizationsServiceInterface), I'm not sure if we would be able to 
> // reimplement it as a mock/fake implementation for unittests.
> ```


What we sometimes do is have our own local client definition:

``` go
type OrganizationsService interface {
	Get(ctx context.Context, org string) (*github.Organization, *github.Response, error)
}

type Client struct {
	Issues        IssuesService
	Organizations OrganizationsService
	PullRequests  PullRequestsService
	Repositories  RepositoriesService
}

// convertClient converts a (go-)github.Client to a Client struct.
func convertClient(c *github.Client) *Client {
	return &Client{
		Issues:        c.Issues,
		Organizations: c.Organizations,
		PullRequests:  c.PullRequests,
		Repositories:  c.Repositories,
	}
}
```	

Then in tests we just implement/stub/mock/fake our (local) client's needs.

> 
> Ps: I'm just giving ideas to the best of my abilities. Sorry if some of this make no sense 🙃.

I think it mostly makes sense.

gauntface:
Adding another 2 cents.

At Google we typically right code along the lines of this:

```go
type myClient struct{
  orgService ghOrgInterface
}

g, err := github.NewClient(...)

c := myClient{
  orgService: g.Orgnaizations,
}

type ghOrgInterface interface{
    ... Methods we use here
}
```
This allows us to stub out portions of the github client. We have a githubstubs package which contain stub clients and have properties to force certain behaviors. This works well for test coverage but can easily miss subtleties in the github package and we have to do the unusual pattern of breaking up the github client into it's services to simplify testing.

I'm not sure having the interfaces would be something we'd use as it's nice limiting the interface to what we use BUT if the interface is defined by a githubtest package it might be easier to use during upgrades.

More recently I've been experimenting with test packages that makes it easy to use packages like httptest and so far that's worked really well and is becoming a preference where possible:

```go
// In Test
ghc := githubtest.New(
    githubtest.WithOrgs(&github.Organization{}),
    githubtest.WithOrgMembers(&github.Member{}),
    githubtest.ForceListReposError(map[string]error{
        "failing-org": error.New("Injected error"),
    }),
)

c := &myClient{
  github: ghc,
}
err := c.DoSomething()
if err != nil {
  // Fail test...
}
```

This is a little verbose in some respects but it abstracts the underlying test infra / magic, i.e. httptest, while letting us use the concrete types in our code.

**tl;dr:** If you move forward with this I'd 100% move it to a separate test package to maintain a healthy split between library code vs test code. I agree with rspier that a wrapper for httptest of some kind would be really interesting and I would ❤️ to see what could be done there but I suspect it would be a larger undertaking that this current CL (and there may be situations where interfaces + stubs are more appropriate).

gmlewis:
@gauntface - I've tried to implement something like what you've described, but I'm hitting a wall.

When you get a chance, please take a look at:

- https://github.com/google/go-github/compare/master...gmlewis:add-mock#diff-b16432bf237ff51bf3a817452fb072abeeb962396e0c8dd6376d817e386cc08aR6-R34
  - the new githubtest package
- https://github.com/google/go-github/compare/master...gmlewis:add-mock#diff-6183bf978a62ebc21c6c7650a53c79f1100a617dcf008aa1cb012f2cee6e98baR18-R54
  - the example test program
- https://github.com/google/go-github/compare/master...gmlewis:add-mock#diff-7098be897e7bc7c6c2d69716ef56fe964fe1fbb8036b52d9ff4a5c59f9d09037R18-R43
  - the example main program that we wish to fake

and see if you can give me any guidance (or examples) of how you envision this to work.

Currently, I'm running up against errors like this:

```
example/embed-interface/main_test.go:48:25: cannot use &fakeOrgSvc{...} (type *fakeOrgSvc) as type *github.OrganizationsService in assignment
```

and can't seem to figure out how I can get the code to call the `fakeOrgSvc.List` method instead of the real one (without injecting interfaces into the main repo, that is).

migueleliasweb:
> ```go
> // In Test
> ghc := githubtest.New(
>     githubtest.WithOrgs(&github.Organization{}),
>     githubtest.WithOrgMembers(&github.Member{}),
>     githubtest.ForceListReposError(map[string]error{
>         "failing-org": error.New("Injected error"),
>     }),
> )
> 

For my usecase this would be a bit hard to use as we inject the whole github client to another struct like so:

```golang
type MySVC struct {
    gh github.Client
}
```

Far as I can tell, unless the whole client could be wrapped around an interface* this code wouldn't be able to create any kind of mocked implementation to help during the unittests as the return of `githubtest.New()` wouldn't be accepted as a `github.Client` type. Unless, the `githubtest.New()` returns a real `github.Client` that contains a custom `http.Roundtripper` injected to it and that makes things easier by marshalling and unmarshalling the correct structs given the request some something similar.

\* I think, currently it wouldn't be possible as inside the `github.Client` struct, all other services are exposed via public properties and properties can't be expressed/enforced in `interfaces`. Maybe if we reworked them as `getters` and `setters`, this would work... I guess?

gauntface:
@gmlewis I see where you were going with that. The missing piece is the httptest usage. That way you would use the normal organization service.

A basic example here (without generation): https://github.com/google/go-github/compare/master...gauntface:githubtest-example?expand=1

The most important part is:

```go
func New(t *testing.T, opts ...Option) (*github.Client, func()) {
	t.Helper()

	// Sort through inputs from test to dictate behavior of the
	o := &options{}
	for _, oo := range opts {
		err := oo(o)
		if err != nil {
			t.Fatalf("Failed to process githubtest.New() options: %v", err)
		}
	}

	ts := httptest.NewServer(http.HandlerFunc(fakeResponses(o)))
	gh := github.NewClient(ts.Client())
	tu := ts.URL
	if !strings.HasSuffix(tu, "/") {
		tu = tu + "/"
	}
	u, err := url.Parse(tu)
	if err != nil {
		t.Fatalf("Failed to parse test server URL: %v", err)
	}
	gh.BaseURL = u
	return gh, ts.Close
}

func fakeResponses(opts *options) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		switch r.URL.Path {
		case "/user/orgs":
			b, err := json.Marshal(opts.Organizations[""])
			if err != nil {
				fmt.Printf("Failed to marshal response for %v\n", r.URL)
				return
			}
			io.WriteString(w, string(b))
		default:
			fmt.Printf("Unhandled network request: %v\n", r.URL)
		}
	}
}
```

This does as @migueleliasweb commented above and returns an actual implementation of the GitHub client, it just points to a fake server.

There are plenty of things to worry about with this approach, my primary concern is whether this scales and what the right interface is for defining behaviors is.

For example, in the example I skipped over handling `/user/orgs` vs `/user/gauntface/orgs`, which a better solution might be to simplify the options to tracks URLs to response data, making the server logic fairly basic:

```go
func fakeResponses(opts *options) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
                for u, d := range opts.calls {
                  if u != r.URL.Path {
                      continue
                  }
                  io.WriteString(w, d[0])
                  // TODO: Increment a counter or remove first item from calls queue
                }
                fmt.Printf("Unhandled network request: %v\n", r.URL)
	}
}

type options struct {
  calls map[string][]string
} 

func WithOrganizations(user string, orgs []*github.Organization) Option {
  return func( o *options) {
    url = "/user/orgs"
    if u != "" {
      u := fmt.Sprintf("/user/%v/orgs")
    }
    o.calls[u] = append(o.calls[u], string(MustMarshal(orgs)))
  }
}
```

This may be a terrible idea. I've done this for a few projects where I've needed to fake out small portions of an API. Scaling this up to all of the GitHub project might result in headaches/hit unexpected limits.

gauntface:
Nope you're spot on and I agree the arguments are the thing that concerns me most in terms of the ballooning in API.

I will confess this approach assumes you want to hide as much of the complexity in the test package. What you're describing could be a simple middle ground to test if the general approach is desirable (i.e. give the githubtest package a function to call when requests occur and the caller determines the behavior)

migueleliasweb:
Hey @gmlewis , it's me again. Just like you said last time, I also let this idea sit for a lil bit and tried to tackle it with a fresher mind.

This time, my starting point was @gauntface implementation. I quite liked the idea of leaving the tests kinda detatched in a way and just mocking/faking the server responses.

Please have a look and let me know what you think. If you reckon this implementation might be promissing, I could enable the mock to some more methods.

https://github.com/google/go-github/pull/1980

gmlewis:
After a fair amount of discussion, I think it is clear that we don't want to go this route, and in fact, it may make sense to provide a separate, external repo for the purpose of mocking this client library as one possible way to go about unit testing.

See https://github.com/google/go-github/pull/1980#issuecomment-881619560 for further context.

Closing.

PR #1837, opened 2021-03-31
Update hook_id to int64

lindluni:
The hook_id field  of the Audit Log API is not a string. It is represented as a number, updated the HookID to int64 to match other fields of type number.

Signed-off-by: Brett Logan <lindluni@github.com>

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -701,9 +701,9 @@ func (a *AuditEntry) GetHeadSHA() string {
 }
 
 // GetHookID returns the HookID field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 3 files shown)


lindluni:
Working on the CLA

lindluni:
Pagination is also broken for the audit-log API, I've opened an issue internally as the current Pagination mechanism for this endpoint is completely different then all other endpoints. So while this change is good, it wont fix the actual audit log API. I'll open an issue here to track that problem as well.

gmlewis:
Thank you, @lindluni !
This is a small enough change that if you need this ASAP, we could close this PR and I could open a new one (until you get your CLA working).  It's your call.

PR #1839, opened 2021-03-31
Add in Workflow, WorkflowRun to WorkflowRunEvent

jacobfoard:
It looks like the docs got [updated recently](https://github.com/github/docs/commit/0d0d3c02ba21fc83e028a612cbae755c4a3862b6) to include these additional fields.


gmlewis:
Thank you, @jacobfoard !
This is also a tiny PR... let me know if you would like me to do something similar to how we handled #1837, although I'm happy to wait for the full process.

jacobfoard:
@gmlewis That'd be swell. I made the fork for the org anticipating the PR to sit for a bit 😅, so thanks for exceeding expectations and being quick on the reply!

PR #1850, opened 2021-04-06
Add Support For Millisecond Timestamps

### LINKED ISSUE #1849: Audit Log API Returns Non-Standard Timestamp
The GitHub Audit Log API return a Unix timestamp in milliseconds, which the current timestamp unmarshalJSON implementation in go-github does not take into account. An implementation needs to be added to account for the audit log having a millisecond-based timestamp.

lindluni:
The GitHub Audit Log API introduced timestamps with millisecond granularity. We add support for these timestamps by first parsing the timestamp into a Unix timestamp and then checking to see if the generated timestamp is in the future. If the timestamp is in the future we re-parse it as a Unix timestamp in Milliseconds.

Tests were added to ensure that when the timestamp is 1 millisecond off that the granularity of the timestamp is preserved by milliseconds correctly being parsed.

Closes #1849

Signed-off-by: Brett Logan <lindluni@github.com>

### DIFF
--- a/github/timestamp.go
+++ b/github/timestamp.go
@@ -29,6 +29,9 @@ func (t *Timestamp) UnmarshalJSON(data []byte) (err error) {
 	i, err := strconv.ParseInt(str, 10, 64)
 	if err == nil {
 		t.Time = time.Unix(i, 0)
+		if t.Time.Year() > 3000 {
+			t.Time = time.Unix(0, 0).Add(time.Duration(i) * time.Millisecond)
+		}
 	} else {
 		t.Time, err = time.Parse(`"`+time.RFC3339+`"`, str)
 	}
--- a/github/timestamp.go
+++ b/github/timestamp.go
@@ -30,7 +30,7 @@ func (t *Timestamp) UnmarshalJSON(data []byte) (err error) {
 	if err == nil {
 		t.Time = time.Unix(i, 0)
 		if t.Time.Year() > 3000 {
-			t.Time = time.Unix(0, 0).Add(time.Duration(i) * time.Millisecond)
+			t.Time = time.Unix(0, i*1e6)
 		}
 	} else {
 		t.Time, err = time.Parse(`"`+time.RFC3339+`"`, str)
--- a/github/enterprise_audit_log_test.go
+++ b/github/enterprise_audit_log_test.go
@@ -55,7 +55,7 @@ func TestEnterpriseService_GetAuditLog(t *testing.T) {
 	}
 	startedAt, _ := time.Parse(time.RFC3339, "2021-03-07T00:33:04.000Z")
 	completedAt, _ := time.Parse(time.RFC3339, "2021-03-07T00:35:08.000Z")
-	timestamp := time.Unix(0, 0).Add(time.Duration(1615077308538) * time.Millisecond)
+	timestamp := time.Unix(0, 1615077308538*1e6)
 
 	want := []*AuditEntry{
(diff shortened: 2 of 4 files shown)


lindluni:
Missed the enterprise_audit_log tests, fixing...

github/timestamp.go
```diff
@@ -29,6 +29,9 @@ func (t *Timestamp) UnmarshalJSON(data []byte) (err error) {
 	i, err := strconv.ParseInt(str, 10, 64)
 	if err == nil {
 		t.Time = time.Unix(i, 0)
+		if t.Time.After(time.Now()) {
```
gmlewis, line 32: My main concern with this implementation revolves around the following question:

Are we guaranteed that _all_ timestamps sent from GitHub will always be in the past?

I think the answer is "no"... I could imagine that a timestamp might represent when something expires in the future.  (There's also the matter of time skew between GitHub's clocks and the user's clocks that makes this fragile for "close-to-now" timestamps.)

I wrote a little experiment locally:

```
go build parsetime.go && parsetime in five minutes
"in five minutes" in milliseconds since epoch is:
1617717693956
Tuesday, 06-Apr-21 10:01:33 EDT
Tuesday, 06-Apr-21 14:01:33 UTC
Interpreting this as if it were seconds returns: 53233-06-07 06:05:56 -0400 EDT
```

and found out that interpreting milliseconds since epoch as seconds since epoch for a "now" timestamp returns the year 53233...

So I was thinking that a better approach would be to first parse the time as seconds, but then if the year is something greater than, say, 3000, go ahead and interpret it as milliseconds since epoch.

How does that sound to you?

gmlewis (changes requested):
Thank you, @lindluni !

github/timestamp.go
lindluni: It's amazing what you can't come up with at midnight. When I was writing this I was trying to think of scenarios where there might be a timestamp in the future, but couldn't think of one. I was also concerned about time "wobble" with timestamps generated from events occurring ridiculously close to the users clock and originally added a single day to the current date, but decided to back it out.

But I think just checking if the year is greater than 3,000 is elegantly simple enough, and hopefully software progresses enough in the next 1,000 years that this is obsolete by then 😂 

Updated to check for a `year > 3000`

gmlewis:
>  lindluni force-pushed the lindluni:timestamp branch from b00be51 to b3adf37 10 minutes ago

Just a friendly reminder - please don't force-push to PRs in this repo in the future, as it makes it more difficult for reviewers to see what changed since the last review.  Thanks.  :smile:

github/timestamp.go
```diff
@@ -29,6 +29,9 @@ func (t *Timestamp) UnmarshalJSON(data []byte) (err error) {
 	i, err := strconv.ParseInt(str, 10, 64)
 	if err == nil {
 		t.Time = time.Unix(i, 0)
+		if t.Time.Year() > 3000 {
+			t.Time = time.Unix(0, 0).Add(time.Duration(i) * time.Millisecond)
```
gmlewis, line 33:
```suggestion
			t.Time = time.Unix(0, i*1e6)
```

lindluni: Done

gmlewis (approved):
Thank you, @lindluni !
LGTM.

wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1853, opened 2021-04-09
Improve tokenauth example as debugging tool

gmlewis:
When assisting someone with GitHub auth token issues, I realized that this example could be more useful if it printed out the rate limit information.


wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp .
Merging.

Issue #1846, opened 2021-04-04
Follow recommended Go 1.13 error handling guidelines


gmlewis:
In a [recent code review](https://github.com/google/go-github/pull/1832#pullrequestreview-627503139), we felt that this repo should use the [recommended Go 1.13 error handling guidelines](https://blog.golang.org/go1.13-errors).

Specifically, we should never use `reflect.DeepEqual` for checking errors (even in tests) and would ideally use the "new" `error.Is` functionality provided in Go 1.13.

Additionally, we should update the README.md to reflect that it is no longer Go 1.9 that we require, but instead we need a minimum of Go 1.13.

mbalayil:
I would like to work on this issue.

gmlewis:
Thank you, @mbalayil ! It is yours.

mbalayil:
There are 6 places in `github/github_test.go` where `reflect.DeepEqual` is used for error comparisons. Switching it to `errors.Is` in `TestDo_nilContext` is straightforward.

```
@@ -705,7 +708,7 @@ func TestDo_nilContext(t *testing.T) {
        req, _ := client.NewRequest("GET", ".", nil)
        _, err := client.Do(nil, req, nil)

-       if !reflect.DeepEqual(err, errors.New("context must be non-nil")) {
+       if !errors.Is(err, errNonNilContext) {
                t.Errorf("Expected context must be non-nil error")
        }
 }
```
where `errNonNilContext`  is defined in `github/github.go`.

However for the rest of the 5 instances where `reflect.DeepEqual` is used, the errors being tested are not sentinel errors.
We can either implement `Is(error) bool` interface for `ErrorResponse`, `RateLimitError` and `AbuseRateLimitError` (possibly using github.com/google/go-cmp) and utilize `errors.Is` in the corresponding tests or use github.com/google/go-cmp directly to compare the expected and actual values as shown below.

```
@@ -1097,7 +1100,11 @@ func TestCheckResponse(t *testing.T) {
                        CreatedAt: &Timestamp{time.Date(2016, time.March, 17, 15, 39, 46, 0, time.UTC)},
                },
        }
-       if !reflect.DeepEqual(err, want) {
+       opts := []cmp.Option{
+               cmp.AllowUnexported(bytes.Buffer{}),
+               cmp.AllowUnexported(http.Request{}),
+       }
+       if !cmp.Equal(err, want, opts...) {
                t.Errorf("Error = %#v, want %#v", err, want)
        }
 }
```
[...]

PR #1854, opened 2021-04-13
Update error handling as per Go 1.13 guidelines

mbalayil:
Fixes [1846](https://github.com/google/go-github/issues/1846)

1. Updated `README.md` to reflect that it is no longer Go 1.9 that we require, but instead we need a minimum of Go 1.13.
2. Replaced `reflect.DeepEqual` with `errors.Is` in all the affected tests.
3. Implemented `Is(error) bool` interface for all the required errors (`ErrorResponse`, `RateLimitError`, `AcceptedError` and `AbuseRateLimitError`). All the individual fields are directly compared. However, for the `*http Response` field, only `StatusCode` is being compared. If the other fields of the `http.Response` struct need to be compared, please let me know.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -132,6 +132,8 @@ const (
 	mediaTypeContentAttachmentsPreview = "application/vnd.github.corsair-preview+json"
 )
 
+var errNonNilContext = errors.New("context must be non-nil")
+
 // A Client manages communication with the GitHub API.
 type Client struct {
 	clientMu sync.Mutex   // clientMu protects the client during calls that modify the CheckRedirect func.
@@ -530,7 +532,7 @@ func parseRate(r *http.Response) Rate {
 // canceled or times out, ctx.Err() will be returned.
 func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, error) {
 	if ctx == nil {
-		return nil, errors.New("context must be non-nil")
+		return nil, errNonNilContext
 	}
 	req = withContext(ctx, req)
 
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
 		r.Response.StatusCode, r.Message, r.Errors)
 }
 
+func (r *ErrorResponse) Is(target error) bool {
+	v, ok := target.(*ErrorResponse)
+	if !ok {
+		return false
+	}
+	if (r.Message != v.Message) || (r.DocumentationURL != v.DocumentationURL) {
+		return false
+	}
+	if r.Response != nil {
+		if r.Response.StatusCode != v.Response.StatusCode {
+			return false
+		}
+	}
+	if len(r.Errors) != len(v.Errors) {
+		return false
+	}
+	for idx := range r.Errors {
+		if r.Errors[idx] != v.Errors[idx] {
+			return false
+		}
+	}
+	if r.Block != nil {
+		if r.Block.Reason != v.Block.Reason || *(r.Block).CreatedAt !=
+			*(v.Block).CreatedAt {
+			return false
+		}
+	}
+	return true
+}
+
 // TwoFactorAuthError occurs when using HTTP Basic Authentication for a user
 // that has two-factor authentication enabled. The request can be reattempted
 // by providing a one-time password in the request.
@@ -702,6 +734,22 @@ func (r *RateLimitError) Error() string {
 		r.Response.StatusCode, r.Message, formatRateReset(time.Until(r.Rate.Reset.Time)))
 }
 
+func (r *RateLimitError) Is(target error) bool {
+	v, ok := target.(*RateLimitError)
+	if !ok {
+		return false
+	}
+	if r.Rate != v.Rate || r.Message != v.Message {
+		return false
+	}
+	if r.Response != nil {
+		if r.Response.StatusCode != v.Response.StatusCode {
+			return false
+		}
+	}
+	return true
+}
+
 // AcceptedError occurs when GitHub returns 202 Accepted response with an
 // empty body, which means a job was scheduled on the GitHub side to process
 // the information needed and cache it.
@@ -715,6 +763,14 @@ type AcceptedError struct {
 
 func (*AcceptedError) Error() string {
 	return "job scheduled on GitHub side; try again later"
+}
+
+func (ae *AcceptedError) Is(target error) bool {
+	v, ok := target.(*AcceptedError)
+	if !ok {
+		return false
+	}
+	return bytes.Compare(ae.Raw, v.Raw) == 0
 }
 
 // AbuseRateLimitError occurs when GitHub returns 403 Forbidden response with the
@@ -733,6 +789,22 @@ func (r *AbuseRateLimitError) Error() string {
 	return fmt.Sprintf("%v %v: %d %v",
 		r.Response.Request.Method, sanitizeURL(r.Response.Request.URL),
 		r.Response.StatusCode, r.Message)
+}
+
+func (r *AbuseRateLimitError) Is(target error) bool {
+	v, ok := target.(*AbuseRateLimitError)
+	if !ok {
+		return false
+	}
+	if r.Message != v.Message || r.RetryAfter != v.RetryAfter {
+		return false
+	}
+	if r.Response != nil {
+		if r.Response.StatusCode != v.Response.StatusCode {
+			return false
+		}
+	}
+	return true
 }
 
 // sanitizeURL redacts the client_secret parameter from the URL which may be
--- a/github/github.go
+++ b/github/github.go
@@ -655,6 +655,20 @@ func (c *Client) checkRateLimitBeforeDo(req *http.Request, rateLimitCategory rat
 	return nil
 }
 
+// compareHttpResponse returns whether two http.Response objects are equal or not.
+// Currently, only StatusCode is checked. This function is used when implementing the
+// Is(error) bool interface for the custom error types in this package.
+func compareHttpResponse(r1, r2 *http.Response) bool {
+	if r1 == nil && r2 == nil {
+		return true
+	}
+
+	if r1 != nil && r2 != nil {
+		return r1.StatusCode == r2.StatusCode
+	}
+	return false
+}
+
 /*
 An ErrorResponse reports one or more errors caused by an API request.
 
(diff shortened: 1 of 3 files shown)


github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
 		r.Response.StatusCode, r.Message, r.Errors)
 }
 
+func (r *ErrorResponse) Is(target error) bool {
+	v, ok := target.(*ErrorResponse)
+	if !ok {
+		return false
```
gmlewis, line 689: I'm not sure if you can see all the warnings I see on GitHub from Codecov, but please add a unit test that covers this line.  I'll highlight the others in case you can't see them.

github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
...
+	}
+	if (r.Message != v.Message) || (r.DocumentationURL != v.DocumentationURL) {
+		return false
```
gmlewis, line 692: Please add a unit test that covers this line.

github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
 		r.Response.StatusCode, r.Message, r.Errors)
 }
 
+func (r *ErrorResponse) Is(target error) bool {
```
gmlewis, line 686: Please add a comment for these new methods... something like:

```
// Is tests if the target error is of type ErrorResponse.
```

github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
...
+	v, ok := target.(*ErrorResponse)
+	if !ok {
+		return false
+	}
+	if (r.Message != v.Message) || (r.DocumentationURL != v.DocumentationURL) {
+		return false
+	}
+	if r.Response != nil {
+		if r.Response.StatusCode != v.Response.StatusCode {
+			return false
```
gmlewis, line 696: Please add a unit test that covers this line.

github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
...
+		}
+	}
+	if len(r.Errors) != len(v.Errors) {
+		return false
```
gmlewis, line 700: Please add a unit test that covers this line.

github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
...
+	}
+	for idx := range r.Errors {
+		if r.Errors[idx] != v.Errors[idx] {
+			return false
```
gmlewis, line 704: Please add a unit test that covers this line.

github/github.go
```diff
@@ -681,6 +683,36 @@ func (r *ErrorResponse) Error() string {
...
+		}
+	}
+	if r.Block != nil {
+		if r.Block.Reason != v.Block.Reason || *(r.Block).CreatedAt !=
+			*(v.Block).CreatedAt {
+			return false
```
gmlewis, line 710: Please add a unit test that covers this line.

github/github.go
```diff
@@ -702,6 +734,22 @@ func (r *RateLimitError) Error() string {
 		r.Response.StatusCode, r.Message, formatRateReset(time.Until(r.Rate.Reset.Time)))
 }
 
+func (r *RateLimitError) Is(target error) bool {
```
gmlewis, line 737: Same here, please.

github/github.go
```diff
@@ -702,6 +734,22 @@ func (r *RateLimitError) Error() string {
...
+	v, ok := target.(*RateLimitError)
+	if !ok {
+		return false
```
gmlewis, line 740: Please add a unit test that covers this line.

github/github.go
```diff
@@ -702,6 +734,22 @@ func (r *RateLimitError) Error() string {
...
+	}
+	if r.Rate != v.Rate || r.Message != v.Message {
+		return false
```
gmlewis, line 743: Please add a unit test that covers this line.

github/github.go
```diff
@@ -702,6 +734,22 @@ func (r *RateLimitError) Error() string {
...
+	}
+	if r.Response != nil {
+		if r.Response.StatusCode != v.Response.StatusCode {
+			return false
```
gmlewis, line 747: Please add a unit test that covers this line.

github/github.go
```diff
@@ -717,6 +765,14 @@ func (*AcceptedError) Error() string {
 	return "job scheduled on GitHub side; try again later"
 }
 
+func (ae *AcceptedError) Is(target error) bool {
```
gmlewis, line 768: Same here, please.

github/github.go
```diff
@@ -717,6 +765,14 @@ func (*AcceptedError) Error() string {
...
+	v, ok := target.(*AcceptedError)
+	if !ok {
+		return false
```
gmlewis, line 771: Please add a unit test that covers this line.

github/github.go
```diff
@@ -717,6 +765,14 @@ func (*AcceptedError) Error() string {
...
+	}
+	return bytes.Compare(ae.Raw, v.Raw) == 0
```
gmlewis, line 773: Please add a unit test that covers this line.

github/github.go
```diff
@@ -735,6 +791,22 @@ func (r *AbuseRateLimitError) Error() string {
 		r.Response.StatusCode, r.Message)
 }
 
+func (r *AbuseRateLimitError) Is(target error) bool {
```
gmlewis, line 794: Same here, please.

github/github.go
```diff
@@ -735,6 +791,22 @@ func (r *AbuseRateLimitError) Error() string {
...
+	v, ok := target.(*AbuseRateLimitError)
+	if !ok {
+		return false
```
gmlewis, line 797: Please add a unit test that covers this line.

github/github.go
```diff
@@ -735,6 +791,22 @@ func (r *AbuseRateLimitError) Error() string {
...
+	}
+	if r.Message != v.Message || r.RetryAfter != v.RetryAfter {
+		return false
```
gmlewis, line 800: Please add a unit test that covers this line.

github/github.go
```diff
@@ -735,6 +791,22 @@ func (r *AbuseRateLimitError) Error() string {
...
+	}
+	if r.Response != nil {
+		if r.Response.StatusCode != v.Response.StatusCode {
+			return false
```
gmlewis, line 804: Please add a unit test that covers this line.

gmlewis (changes requested):
Thank you, @mbalayil !
Please address the comments below.

mbalayil:
Have updated the PR with more tests to improve coverage. Please take a look.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: I have addressed all code coverage issues with additional tests.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

github/github.go
mbalayil: Done.

gmlewis (approved):
Thank you, @mbalayil !
LGTM.

wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp!

@willnorris - are you OK with this implementation?  I'll wait until we hear from you... no rush... I just want to make sure since you had some comments on the thread.

willnorris:
> @willnorris - are you OK with this implementation?

My biggest concern, just looking at this from a high level, is that this is a LOT of work and extra code just to be able to call `errors.Is` in a couple of test cases.  If having these errors be comparable in this way has real value outside of just these tests, then that may be fine, and it may be worth the effort.  But if it's **just** to have marginally cleaner tests (not calling `reflect.DeepEqual`), I find it harder to justify.

The whole point of the new go1.13 error handling is that the `Is` and `As` functions support wrapped errors.  Both of them will intelligently unwrap errors to try and find a match.  But we're not wrapping errors.  Our errors are *just values*, like any other value in Go.  So honestly, there's no advantage of using `errors.Is` versus the go-cmp package (in our tests, I mean).

I really appreciate all the effort @mbalayil put into this... really, I do.  But looking at what it ended up taking in the end, I'm not sure that this is the right solution for this problem.  (Unless there's some other use case that this is satisfying, outside of those two test cases.)

gmlewis:
Correct me if I'm wrong, @willnorris , but won't the availability of the `Is` methods (added by tihs PR) make it easier for the users of this package to wrap _their_ errors and still test if the base error is one of the types defined in this repo?

In other words, doesn't the work here make error handler easier for future users of this package (and also make the usage more inline with the [Go 1.13 Error Handling blog post](https://blog.golang.org/go1.13-errors))?

willnorris:
Yeah, that's a good point.  For the sentinel errors we have, I think that will be very useful (but also doesn't require a custom `Is` method).  Given how a couple of these errors are tied to very specific http requests/responses, I think the full equality check is going to be of very limited value, even for downstream users, but am happy to be proven wrong.

willnorris (approved):

gmlewis:
> I think the full equality check is going to be of very limited value, even for downstream users, but am happy to be proven wrong.

So do you think it would be more useful _without_ the equality checks?

Or should we run with this for awhile until we get feedback and then adjust accordingly?

willnorris:
Looking more at the examples I was concerned about, they seem to be more limited than I thought.  I was mostly focusing on the `CreatedAt` timestamp that exists in some errors.  I thought that was more common, but I now see it is only for HTTP 451 errors.  For those errors, an equality check will almost never match in the real world because you won't know the creation date ahead of time.

I guess I'd need to look at some more real-world errors to see how likely it would be for people to really be doing equality checks.  My guess is that most go-github users that want to wrap our errors and eventually test for error types will use `errors.As` (which we don't have to do anything for).

I know that doesn't directly answer your question, but I guess I'm discovering that my once somewhat-strongly-held view is softening considerably, so I can go either way on this now.

gmlewis:
OK, sounds good.  Thanks, Will! 
Let's run with this and see what kind of feedback we get.  :joy:
Merging.

PR #1858, opened 2021-04-26
Remove obsolete apps manifest preview custom header

chmouel:
Apps manifest are now out of preview and uses mediaTypeV3:

https://docs.github.com/en/rest/reference/apps#create-a-github-app-from-a-manifest

### DIFF
--- a/github/apps_manifest.go
+++ b/github/apps_manifest.go
@@ -10,10 +10,6 @@ import (
 	"fmt"
 )
 
-const (
-	mediaTypeAppManifestPreview = "application/vnd.github.fury-preview+json"
-)
-
 // AppConfig describes the configuration of a GitHub App.
 type AppConfig struct {
 	ID            *int64     `json:"id,omitempty"`
@@ -41,7 +37,6 @@ func (s *AppsService) CompleteAppManifest(ctx context.Context, code string) (*Ap
 	if err != nil {
 		return nil, nil, err
 	}
-	req.Header.Set("Accept", mediaTypeAppManifestPreview)
 
 	cfg := new(AppConfig)
 	resp, err := s.client.Do(ctx, req, cfg)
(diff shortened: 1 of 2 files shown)


github/apps_manifest.go
```diff
@@ -41,7 +37,7 @@ func (s *AppsService) CompleteAppManifest(ctx context.Context, code string) (*Ap
 	if err != nil {
 		return nil, nil, err
 	}
-	req.Header.Set("Accept", mediaTypeAppManifestPreview)
+	req.Header.Set("Accept", mediaTypeV3)
```
gmlewis, line 40: Please remove this line, as this header is added to all requests here:
https://github.com/google/go-github/blob/master/github/github.go#L383

github/apps_manifest_test.go
```diff
@@ -30,7 +30,7 @@ func TestGetConfig(t *testing.T) {
 
 	mux.HandleFunc("/app-manifests/code/conversions", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "POST")
-		testHeader(t, r, "Accept", mediaTypeAppManifestPreview)
+		testHeader(t, r, "Accept", mediaTypeV3)
```
gmlewis, line 33: Let's please also remove this line for conformity with the rest of the repo.

gmlewis (approved):
Thank you, @chmouel !
LGTM.

gmlewis:
For future PRs, please do not use force push in this repo. See CONTRIBUTING.md for details. Thanks.

wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp !
Merging.

bluekeyes:
@gmlewis thanks for checking, I think this is probably fine. The GitHub docs for the supported enterprise versions don't reference this header (not that it's a great guarantee the header is unused) and I've personally never used this endpoint with the instances I maintain.

Issue #1863, opened 2021-05-06
Teams is missing "permissions" field


derekjobst:
It appears the [`Repositories.ListTeams`](https://github.com/google/go-github/blob/master/github/repos.go#L691)([Docs](https://docs.github.com/en/rest/reference/repos#list-repository-teams)) API call returns a field `permissions` on each team which includes the more detailed set of permissions with triage/maintain.

A similar field is currently populated on **User** when calling [`Repositories.ListCollaborators`](https://github.com/google/go-github/blob/master/github/repos_collaborators.go#L45)([Docs].(https://docs.github.com/en/rest/reference/repos#list-repository-collaborators))

### Example
Tested on Github Enterprise v3.0.6
`https://<github-host>/api/v3/repos/test-org/test-repo/teams`
```json
[
  {
    "name": "Example Team",
    "id": 99,
    "node_id": "...",
    "slug": "example-team",
    "description": "",
    "privacy": "closed",
    "url": "...",
    "html_url": "...",
    "members_url": "...",
    "repositories_url": "...",
    "permission": "push",
    "permissions": {
      "pull": false,
      "triage": false,
      "push": false,
      "maintain": true,
      "admin": false
    },
    "parent": null
  },
]
```

### Docs
This endpoint appears to be the same in [Enterprise 3.0](https://docs.github.com/en/enterprise-server@3.0/rest/reference/repos#list-repository-teams) and [Github.com](https://docs.github.com/en/rest/reference/repos#list-repository-teams).

### Proposal
Add a Permissions field identical to what exists on [**User**](https://github.com/google/go-github/blob/master/github/users.go#L71) to the [**Team**](https://github.com/google/go-github/blob/master/github/teams.go#L24) type.

derekjobst:
Thanks for all the help @gmlewis!

PR #1864, opened 2021-05-06
Add permissions to teams

derekjobst:
This PR addresses https://github.com/google/go-github/issues/1863 by adding a Permissions field to the Team type.

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -32,6 +32,10 @@ type Team struct {
 	// Permission specifies the default permission for repositories owned by the team.
 	Permission *string `json:"permission,omitempty"`
 
+	// Permissions identifies the permissions that a team has on a given
+	// repository. This is only populated when calling Repositories.ListTeams.
+	Permissions *map[string]bool `json:"permissions,omitempty"`
+
 	// Privacy identifies the level of privacy this team should have.
 	// Possible values are:
 	//     secret - only visible to organization owners and members of this team
--- a/github/teams.go
+++ b/github/teams.go
@@ -34,7 +34,7 @@ type Team struct {
 
 	// Permissions identifies the permissions that a team has on a given
 	// repository. This is only populated when calling Repositories.ListTeams.
-	Permissions map[string]bool `json:"permissions,omitempty"`
+	Permissions *map[string]bool `json:"permissions,omitempty"`
 
 	// Privacy identifies the level of privacy this team should have.
 	// Possible values are:
--- a/github/teams.go
+++ b/github/teams.go
@@ -34,7 +34,7 @@ type Team struct {
 
 	// Permissions identifies the permissions that a team has on a given
 	// repository. This is only populated when calling Repositories.ListTeams.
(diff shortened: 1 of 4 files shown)


derekjobst:
@googlebot I signed it!

github/teams.go
```diff
@@ -32,6 +32,10 @@ type Team struct {
 	// Permission specifies the default permission for repositories owned by the team.
 	Permission *string `json:"permission,omitempty"`
 
+	// Permissions identifies the permissions that a team has on a given
+	// repository. This is only populated when calling Repositories.ListTeams.
+	Permissions *map[string]bool `json:"permissions,omitempty"`
```
gmlewis, line 37:
```suggestion
	Permissions map[string]bool `json:"permissions,omitempty"`
```

I see that there are two other places in this repo where we use `*map[string]bool`, but I believe that the pointer to a map is unnecessary since map is already a reference type and can be `nil` without the pointer.

Let's start here by making it just a map, and then clean up the other usage later.

gmlewis:
Also, please make sure to run `go generate ./...` as described in our `CONTRIBUTING.md` guide.

gmlewis:
You will want to commit/push (not force-push) the changes to the PR once you run `go generate ./...`.

github/teams.go
derekjobst: Ok, sounds good. Thanks @gmlewis!

gmlewis:
Hmmm... it looks like my `go generate ./...` code is breaking the tests... sorry about that @derekjobst - let me investigate.

derekjobst:
I got caught up on [Stringify](https://github.com/derekjobst/go-github/blob/5c30ef9f9382b5cfea673c8d837d8032bdd5e6b2/github/teams.go#L59) returning a string with the empty Permissions{} in it when Team is converted to a string.

This was easy to fix in [strings_test](https://github.com/derekjobst/go-github/blob/5c30ef9f9382b5cfea673c8d837d8032bdd5e6b2/github/strings_test.go#L126), but the auto-generated github-stringify tests for `TestTeam_String` refer to a parent team, which does not include the empty Permissions in the string.

It seems like the best solution might be to have the [stringifyValue](https://github.com/derekjobst/go-github/blob/5c30ef9f9382b5cfea673c8d837d8032bdd5e6b2/github/strings.go#L30) function in strings.go omit empty Permissions.

derekjobst:
I've reverted the pointer change to get tests passing — happy to change back if we'd like to change the stringify methods

gmlewis:
> I've reverted the pointer change to get tests passing — happy to change back if we'd like to change the stringify methods

OK, excellent.   I've gone ahead and created #1865 and #1866 to address this.

gmlewis (approved):
Thank you, @derekjobst !
LGTM.
Merging.

derekjobst:
@gmlewis do you know when we might be able to get this into a release?

gmlewis:
@derekjobst - here you go: https://github.com/google/go-github/releases/tag/v35.2.0

Issue #1865, opened 2021-05-06
Change all `*map...` to `map...` in JSON structs


gmlewis:
See #1866 for details.

This should _not_ be a breaking API change.

PR #1870, opened 2021-05-07
Change all *map to map in JSON structs

gmlewis:
Fixes: #1865.

Note that this should _not_ be a breaking API change if I'm not mistaken.


wesleimp (approved):
💙

gmlewis:
Thank you, @wesleimp !
Merging.

PR #1872, opened 2021-05-13
ListPullsForCommit

emmjohnson:


gmlewis (commented):
Thank you, @emmjohnson , but I believe this endpoint is already implemented here:
https://github.com/google/go-github/blob/master/github/pulls.go#L159-L185

gmlewis:
Closing as already implemented.

Issue #1851, opened 2021-04-08
switch tests to use github.com/google/go-cmp package


willnorris:
#1846 calls out not using `reflect.DeepEqual` for comparing error values, but in reality we shouldn't be using it all. https://pkg.go.dev/github.com/google/go-cmp/cmp is designed to be a much safer alternative, so we should use that for tests.

willnorris:
and just to be clear, we should use go-cmp for all of our non-error values (of which there appear to be 600+ comparisons?!).  Switching to go1.13's `errors.Is` is the right change for errors.

photongupta:
@willnorris Can I start working on this?

willnorris:
have at it!  (With so many instances, you might want to do a global find and replace, or use `gofmt -r`)

sagar23sj:
Hi @photongupta. Are You Still working on this issue.

PR #1875, opened 2021-05-26
Update comparisons in test files from reflect to cmp

sagar23sj:
Replaced all reflect.DeepEqual comparisons in test files with cmp.Equal.

Fixes: #1851

### DIFF
--- a/github/actions_artifacts_test.go
+++ b/github/actions_artifacts_test.go
@@ -10,8 +10,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"reflect"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 )
 
 func TestActionsService_ListArtifacts(t *testing.T) {
@@ -37,7 +38,7 @@ func TestActionsService_ListArtifacts(t *testing.T) {
 	}
 
 	want := &ArtifactList{TotalCount: Int64(1), Artifacts: []*Artifact{{ID: Int64(1)}}}
-	if !reflect.DeepEqual(artifacts, want) {
+	if !cmp.Equal(artifacts, want) {
 		t.Errorf("Actions.ListArtifacts returned %+v, want %+v", artifacts, want)
 	}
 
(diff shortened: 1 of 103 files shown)


sagar23sj:
Can someone help me understand, why checks are failing?
I ran all 3 commands go generate, go test, and go vet but still these are somehow failing in checks.

willnorris:
Looks like there is still a reference to the reflect package, but it's not imported any longer:

```
Error: github/orgs_members_test.go:835:6: undefined: reflect
```

sagar23sj:
> Looks like there is still a reference to the reflect package, but it's not imported any longer:
> 
> ```
> Error: github/orgs_members_test.go:835:6: undefined: reflect
> ```
> 
> https://github.com/google/go-github/pull/1875/checks?check_run_id=2677192197#step:8:9

Thanks for replying, but I checked there is no "reflect" imported for that particular file in the branch I created PR with.

willnorris:
It looks like you need to rebase onto the upstream master branch, as you're missing some commits that apparently are getting picked up by the CI jobs (notably d39b94da583ae6dbe8f3039a95f7334fd47b253f).  

I don't entirely understand why it's not simply running the CI jobs against your branch, which should pass, but would then have problems come merge time.  Maybe it's automatically trying to merge your changes with master, resulting in these errors?

sagar23sj:
> It looks like you need to rebase onto the upstream master branch, as you're missing some commits that apparently are getting picked up by the CI jobs (notably [d39b94d](https://github.com/google/go-github/commit/d39b94da583ae6dbe8f3039a95f7334fd47b253f)).
> 
> I don't entirely understand why it's not simply running the CI jobs against your branch, which should pass, but would then have problems come merge time. Maybe it's automatically trying to merge your changes with master, resulting in these errors?

Thanks @willnorris , I didn't realize my branch was not up to date.
Pushed again after rebasing, I hope checks pass now.

willnorris:
okay, so now the error about the reflect reference in `orgs_members_test.go` is correct :)  You'll need to switch that one to use the cmp package as well.

sagar23sj:
> okay, so now the error about the reflect reference in `orgs_members_test.go` is correct :) You'll need to switch that one to use the cmp package as well.

Thanks for helping @willnorris . I have fixed the issue and seems like checks also passed.

gmlewis (approved):
Thank you, @sagar23sj and @willnorris !
LGTM.

willnorris (approved):

willnorris:
@gmlewis oops, looks like I missed catching your issue rename before merging :)

gmlewis:
:joy: No problem, Will!

PR #1876, opened 2021-05-26
compare decoded go struct for request body

willnorris:
The use of `cmp.AllowUnexported` stuck out to me as code smell when reviewing #1875, but since the weird comparison was already there, I decided to leave it alone in that PR.

This was a very weird test method, since it doesn't follow the pattern of the rest of go-github for comparing POST request bodies, and as a result makes things way more complicated than they need to be.

This eliminates the `GetReadWriter` and `GetReadCloser` funcs.  There was [discussion in #1238](https://github.com/google/go-github/pull/1238/files#r313185975) about these potentially being general purpose funcs, but they never ended up being used beyond this one method, nor should they ever be necessary I don't think.  Even though these were exported funcs, because they were only defined in the test package, this shouldn't be a breaking change.

### DIFF
--- a/github/apps_test.go
+++ b/github/apps_test.go
@@ -6,12 +6,9 @@
 package github
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"net/http"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @willnorris !
LGTM.

Issue #1895, opened 2021-06-15
Redirect on GetBranch


asraa:
It's common for master branch to be renamed to main branch.

I noticed that when trying to get a renamed master branch, I normally get a 301:
```
$ curl -i  -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/asraa/asraa.github.io/branches/master
HTTP/2 301 
server: GitHub.com
date: Tue, 15 Jun 2021 15:31:51 GMT
content-type: application/json; charset=utf-8
content-length: 190
x-oauth-scopes: public_repo, read:org
x-accepted-oauth-scopes: 
location: https://api.github.com/repositories/asraa/asraa.github.io/branches/main
x-github-media-type: github.v3; format=json
x-ratelimit-limit: 5000
x-ratelimit-remaining: 4628
x-ratelimit-reset: 1623773879
x-ratelimit-used: 372
x-ratelimit-resource: core
access-control-expose-headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset
access-control-allow-origin: *
strict-transport-security: max-age=31536000; includeSubdomains; preload
x-frame-options: deny
x-content-type-options: nosniff
x-xss-protection: 0
referrer-policy: origin-when-cross-origin, strict-origin-when-cross-origin
content-security-policy: default-src 'none'
vary: Accept-Encoding, Accept, X-Requested-With
x-github-request-id: AEFA:589F:209F980:435C7FE:60C8C7E7

{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/asraa/asraa.github.io/branches/main",
  "documentation_url": "https://docs.github.com/v3/#http-redirects"
}

```

But using go-github, I get a 404 response. Is there any way to handle the 301 response and include the redirect URL in the response body when calling `GetBranch`?

asraa:
I'm confused why it's a 404 and not at least returning the correct status plus the JSON decoding error... 

am I missing something?

gmlewis:
Thank you, @asraa !

We could do something similar to what we did for:

* [`DownloadArtifact`](https://github.com/google/go-github/blob/master/github/actions_artifacts.go#L107-L150)
[...]

PR #1901, opened 2021-06-18
Modify GetBranch to handle redirects

n1lesh:
Adds a new followRedirects param for handling redirects on the **GetBranch** API. If redirected, the `response.Request.URL` will contain the redirected URL.

As per: #1895.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"strings"
 )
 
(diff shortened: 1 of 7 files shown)


github/repos.go
```diff
@@ -933,20 +934,48 @@ func (s *RepositoriesService) ListBranches(ctx context.Context, owner string, re
...
+		resp, err = s.client.client.Transport.RoundTrip(req)
+	}
+	if err != nil {
+		return nil, err
 	}
 
-	return b, resp, nil
+	// If redirect response is returned, follow it
+	if followRedirects && resp.StatusCode == http.StatusMovedPermanently {
+		u = resp.Header.Get("Location")
```
gmlewis, line 975: We need the following immediately before line 975 in order to close the existing response body:

```go
resp.Body.Close()
```

gmlewis (changes requested):
Please add the missing line and also add a test that runs through the code with a redirect.

See https://github.com/google/go-github/blob/master/github/repos_contents_test.go#L668-L693 for inspiration.

github/repos.go
n1lesh: I actually gave it a lot of thought but couldn't decide whether or not to add it just before variable re-assignment.

github/repos.go
```diff
@@ -933,20 +934,49 @@ func (s *RepositoriesService) ListBranches(ctx context.Context, owner string, re
...
-	req, err := s.client.NewRequest("GET", u, nil)
+
+	resp, err := s.getBranchFromURL(ctx, u, followRedirects)
 	if err != nil {
 		return nil, nil, err
 	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, newResponse(resp), fmt.Errorf("unexpected status code: %s", resp.Status)
```
gmlewis, line 947: Is it possible for you to write a unit test that covers line 947?

github/repos.go
```diff
@@ -933,20 +934,49 @@ func (s *RepositoriesService) ListBranches(ctx context.Context, owner string, re
...
+		return nil, err
+	}
+
+	var resp *http.Response
+	// Use http.DefaultTransport if no custom Transport is configured
+	req = withContext(ctx, req)
+	if s.client.client.Transport == nil {
+		resp, err = http.DefaultTransport.RoundTrip(req)
+	} else {
+		resp, err = s.client.client.Transport.RoundTrip(req)
```
gmlewis, line 967: Is it possible for you to write a unit test that covers line 967?

github/repos.go
```diff
@@ -933,20 +934,49 @@ func (s *RepositoriesService) ListBranches(ctx context.Context, owner string, re
...
+	}
+	if err != nil {
+		return nil, err
```
gmlewis, line 970: Is it possible for you to write a unit test that covers line 970?

gmlewis (changes requested):

github/repos.go
n1lesh: Transport.RoundTrip is not used much, and where ever it is used (as mentioned by you [here](https://github.com/google/go-github/issues/1895#issuecomment-861631842)), its not covered by a test.

Would it make sense to create a mock for round tripper and add tests for all 4?

gmlewis: Possibly... if you could investigate, that would be great.  Thanks.
(I don't want to pull in additional dependencies, though... so please avoid doing that.)

n1lesh: We shouldn't need any additional dependencies. However, would it be okay to put the transport mock under `github_test.go` as multiple tests will be using it?

gmlewis: Sure, that sounds good to me.

gmlewis (approved):
Very nice, @n1lesh !  Thank you!
LGTM.

Parker77 (approved):
LGTM.

Issue #1899, opened 2021-06-17
Client.RateLimits shouldn't return RateLimitError


sa-spag:
👋 Hello,

I noticed that `Client.RateLimits` can return a `RateLimitError`, as per:
https://github.com/google/go-github/blob/75644ea48541a9c99ad73bf86e50bae667b10a20/github/github.go#L648-L652

According to [GitHub REST API documentation](https://docs.github.com/en/rest/reference/rate-limit), rate limit status can be checked "at any time using the Rate Limit API", i.e. even when the remaining number of allowed requests is 0, and the `GET /rate_limit` endpoint never responds with a HTTP 403. I verified myself and I can confirm the documentation is accurate. Therefore, I think `Client.RateLimits` should never return `RateLimitError` errors and return the actual rate limit status, by performing a HTTP request.

I have 2 solutions in mind:
* `Client.checkRateLimitBeforeDo` is aware of some allowlist of endpoints, including `GET /rate_limits` (and perhaps others I don't know yet about), for which it should always return `nil`;
* a higher layer passes a boolean which indicates that the request should not be subject to the pre-request rate limit check,  possibly exposed to the user so one could hit GitHub API regardless of the internal rate limit status.

Happy to discuss any solution before I (or someone else) patch(es) the issue. Thanks!

gmlewis:
May I please ask what the actual problem is that you are trying to solve?

In other words, why is it a problem that `Client.RateLimits()`  returns a `RateLimitError` (which, by the way, should have all the information you need to avoid causing the next rate limit error... specifically the `Reset` timestamp)?

sa-spag:
> May I please ask what the actual problem is that you are trying to solve?

`RateLimitError` only exposes the rate limit status for the request's category, i.e. `core` for `GET /rate_limits`, but there are a bunch of other "resources" for which rate limit is independent and for which I would like to request an up-to-date status.

gmlewis:
I suppose a third option is to modify `Client.RateLimits()` to always perform the call regardless of saved history by bypassing the existing check.

sa-spag:
> I suppose a third option is to modify `Client.RateLimits()` to always perform the call regardless of saved history by bypassing the existing check.

Thanks for your feedback. Exactly, and I think we have at least 2 ways of implementing such bypass mechanism (the 2 options I previously listed). Do not hesitate to provide any guidance to take a decision.

gmlewis:
[...]

PR #1907, opened 2021-06-22
Enable `Client.RateLimits` to bypass the rate limit check

sa-spag:
Fixes https://github.com/google/go-github/issues/1899.

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -201,6 +201,9 @@ func testNewRequestAndDoFailure(t *testing.T, methodName string, client *Client,
 	client.BaseURL.Path = "/api-v3/"
 	client.rateLimits[0].Reset.Time = time.Now().Add(10 * time.Minute)
 	resp, err = f()
+	if bypass := resp.Request.Context().Value(bypassRateLimitCheck); bypass != nil {
+		return
+	}
 	if want := http.StatusForbidden; resp == nil || resp.Response.StatusCode != want {
 		if resp != nil {
 			t.Errorf("rate.Reset.Time > now %v resp = %#v, want StatusCode=%v", methodName, resp.Response, want)
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -1711,6 +1711,52 @@ func TestRateLimits_coverage(t *testing.T) {
 	})
 }
 
+func TestRateLimits_overQuota(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	client.rateLimits[coreCategory] = Rate{
+		Limit:     1,
+		Remaining: 0,
+		Reset:     Timestamp{time.Now().Add(time.Hour).Local()},
+	}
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Very nice, @sa-spag !
I really like this solution... I was concerned that there would be much more disruption, but this looks very nice and clean... thank you!
LGTM.

gmlewis:
Whups, sorry, it looks like the unit tests are not passing:

```
    github_test.go:1708: rate.Reset.Time > now RateLimits resp = &http.Response{Status:"404 Not Found", StatusCode:404, Proto:"HTTP/1.1", ProtoMajor:1, ProtoMinor:1, Header:http.Header{"Content-Length":[]string{"19"}, "Content-Type":[]string{"text/plain; charset=utf-8"}, "Date":[]string{"Tue, 22 Jun 2021 17:21:16 GMT"}, "X-Content-Type-Options":[]string{"nosniff"}}, Body:io.nopCloser{Reader:(*bytes.Buffer)(0xc0003c1860)}, ContentLength:19, TransferEncoding:[]string(nil), Close:false, Uncompressed:false, Trailer:http.Header(nil), Request:(*http.Request)(0xc0002f5e00), TLS:(*tls.ConnectionState)(nil)}, want StatusCode=403
--- FAIL: TestRateLimits_coverage (0.00s)
```

gmlewis (changes requested):
Changing approval until unit tests pass.

sa-spag:
> Very nice, @sa-spag !
> I really like this solution... I was concerned that there would be much more disruption, but this looks very nice and clean... thank you!

Credits to @ViBiOh for the `context.WithValue` trick I had no clue about.

> Whups, sorry, it looks like the unit tests are not passing...

Whups indeed, my bad. Got a little too confident, it is now fixed.

github/github_test.go
```diff
@@ -201,6 +201,9 @@ func testNewRequestAndDoFailure(t *testing.T, methodName string, client *Client,
 	client.BaseURL.Path = "/api-v3/"
 	client.rateLimits[0].Reset.Time = time.Now().Add(10 * time.Minute)
 	resp, err = f()
+	if bypass := resp.Request.Context().Value(bypassRateLimitCheck); bypass != nil {
+		return
```
gmlewis, line 205: I'm concerned that we are not checking the value of `err` here... might there be a case where we get an error that will be  ignored here?

sa-spag: This helper explicitly sets the client up to trigger errors, and at this stage a rate limit error. As is, the test would fail if we check for `err != nil` for `methodName == "RateLimits"` as the request would result in a HTTP 404, since no handler has been created. We can set one up on `TestRateLimits_coverage` but I think it makes the test less straightforward, and the nominal case is already covered by other functions.

gmlewis (approved):
LGTM.

gmlewis:
Friendly ping @wesleimp

Parker77 (approved):
LGTM.

PR #1934, opened 2021-07-06
Add support for listing and getting repository/organization webhook deliveries

### LINKED ISSUE #1933: Add support for Webhook Deliveries API
Hey! Thanks a lot for maintaining this awesome project.

FYI, GitHub has recently released the GA version of ["Webhook Deliveries API"](https://github.blog/changelog/2021-06-30-webhook-deliveries-api/).

I'm going to submit a PR to add the initial support for that, containing two API methods that I'm gonna need in my own project.

It would be great if you could review and include it in this awesome project, if possible. Thanks!

mumoshu:
Resolves #1933

To make this PR minimal as possible for ease of reviewing, I intentionally included only four APIs related to the webhook deliveries API:

- Listing repository hook deliveries
- Getting a specific repository hook delivery
- Listing organization hook deliveries
- GEtting a specific organization hook delivery

That being said, the following methods aren't covered in this PR. If you like, I'd gladly submit subsequent PRs to add all of them.

- Trigger redelivery of a hook delivery
- ~~List/get/redelivery for organizational webhooks~~

Note that changes on the following files are completely generated by running `go generate` as documented in your guide, so you won't need to read them that carefully.

- `github/github-accessors_test.go`
- `github/github-accessors.go`

FWIW, you can see how these additional methods can be used to retrieve historical webhook deliveries here:

https://github.com/actions-runner-controller/actions-runner-controller/blob/c3cfab81c85d33c65f7b6f6443da0d2f57fd72db/pkg/githubwebhookdeliveryforwarder/githubwebhookdelivery.go#L96-L139

The part that propagates `Cursor` from the response to the ListCursorOptions for the next paginated request is beneficial to understanding how the cursor can be used for pagination.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -212,6 +212,9 @@ type ListCursorOptions struct {
 
 	// A cursor, as given in the Link header. If specified, the query only searches for events before this cursor.
 	Before string `url:"before,omitempty"`
+
+	// A cursor, as given in the Link header. If specified, the query continues the search using this cursor.
+	Cursor string `url:"cursor,omitempty"`
 }
 
 // UploadOptions specifies the parameters to methods that support uploads.
@@ -445,6 +448,11 @@ type Response struct {
 	// calling the endpoint again.
 	NextPageToken string
 
+	// For APIs that support cursor pagination, such as RepositoryService.ListRepositoryHookDeliveries,
+	// the following field will be populated to point to the next page.
+	// Set ListCursorOptions.Cursor to this value when calling the endpoint again.
+	Cursor string
+
 	// Explicitly specify the Rate type so Rate's String() receiver doesn't
 	// propagate to Response.
 	Rate Rate
@@ -481,25 +489,33 @@ func (r *Response) populatePageValues() {
 			if err != nil {
 				continue
 			}
-			page := url.Query().Get("page")
-			if page == "" {
+
+			if page := url.Query().Get("page"); page != "" {
+				for _, segment := range segments[1:] {
+					switch strings.TrimSpace(segment) {
+					case `rel="next"`:
+						if r.NextPage, err = strconv.Atoi(page); err != nil {
+							r.NextPageToken = page
+						}
+					case `rel="prev"`:
+						r.PrevPage, _ = strconv.Atoi(page)
+					case `rel="first"`:
+						r.FirstPage, _ = strconv.Atoi(page)
+					case `rel="last"`:
+						r.LastPage, _ = strconv.Atoi(page)
+					}
+				}
+
 				continue
 			}
 
-			for _, segment := range segments[1:] {
-				switch strings.TrimSpace(segment) {
-				case `rel="next"`:
-					if r.NextPage, err = strconv.Atoi(page); err != nil {
-						r.NextPageToken = page
+			if cursor := url.Query().Get("cursor"); cursor != "" {
+				for _, segment := range segments[1:] {
+					switch strings.TrimSpace(segment) {
+					case `rel="next"`:
+						r.Cursor = cursor
 					}
-				case `rel="prev"`:
-					r.PrevPage, _ = strconv.Atoi(page)
-				case `rel="first"`:
-					r.FirstPage, _ = strconv.Atoi(page)
-				case `rel="last"`:
-					r.LastPage, _ = strconv.Atoi(page)
 				}
-
 			}
 		}
 	}
--- a/github/github.go
+++ b/github/github.go
@@ -490,31 +490,34 @@ func (r *Response) populatePageValues() {
 				continue
 			}
 
-			if page := url.Query().Get("page"); page != "" {
-				for _, segment := range segments[1:] {
-					switch strings.TrimSpace(segment) {
-					case `rel="next"`:
-						if r.NextPage, err = strconv.Atoi(page); err != nil {
-							r.NextPageToken = page
-						}
-					case `rel="prev"`:
-						r.PrevPage, _ = strconv.Atoi(page)
-					case `rel="first"`:
-						r.FirstPage, _ = strconv.Atoi(page)
-					case `rel="last"`:
-						r.LastPage, _ = strconv.Atoi(page)
-					}
-				}
+			q := url.Query()
 
-				continue
-			}
-
-			if cursor := url.Query().Get("cursor"); cursor != "" {
+			if cursor := q.Get("cursor"); cursor != "" {
 				for _, segment := range segments[1:] {
 					switch strings.TrimSpace(segment) {
 					case `rel="next"`:
 						r.Cursor = cursor
 					}
+				}
+			}
+
+			page := q.Get("page")
+			if page == "" {
+				continue
+			}
+
+			for _, segment := range segments[1:] {
+				switch strings.TrimSpace(segment) {
+				case `rel="next"`:
+					if r.NextPage, err = strconv.Atoi(page); err != nil {
+						r.NextPageToken = page
+					}
+				case `rel="prev"`:
+					r.PrevPage, _ = strconv.Atoi(page)
+				case `rel="first"`:
+					r.FirstPage, _ = strconv.Atoi(page)
+				case `rel="last"`:
+					r.LastPage, _ = strconv.Atoi(page)
 				}
 			}
 		}
--- a/github/github.go
+++ b/github/github.go
@@ -499,6 +499,8 @@ func (r *Response) populatePageValues() {
 						r.Cursor = cursor
 					}
 				}
+
+				continue
 			}
 
 			page := q.Get("page")
--- a/github/orgs_hooks_deliveries.go
+++ b/github/orgs_hooks_deliveries.go
@@ -1,0 +1,48 @@
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// ListHookDeliveries lists webhook deliveries for a webhook configured in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/orgs#list-deliveries-for-an-organization-webhook
+func (s *OrganizationsService) ListHookDeliveries(ctx context.Context, org string, id int64, opts *ListCursorOptions) ([]*HookDelivery, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/hooks/%v/deliveries", org, id)
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	deliveries := []*HookDelivery{}
+	resp, err := s.client.Do(ctx, req, &deliveries)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return deliveries, resp, nil
+}
+
+// GetHookDelivery returns a delivery for a webhook configured in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/orgs#get-a-webhook-delivery-for-an-organization-webhook
+func (s *OrganizationsService) GetHookDelivery(ctx context.Context, owner string, hookID, deliveryID int64) (*HookDelivery, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/hooks/%v/deliveries/%v", owner, hookID, deliveryID)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	h := new(HookDelivery)
+	resp, err := s.client.Do(ctx, req, h)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return h, resp, nil
+}
--- a/github/orgs_hooks_deliveries.go
+++ b/github/orgs_hooks_deliveries.go
@@ -43,6 +43,7 @@ func (s *OrganizationsService) GetHookDelivery(ctx context.Context, owner string
 	if err != nil {
 		return nil, nil, err
 	}
+
 	h := new(HookDelivery)
 	resp, err := s.client.Do(ctx, req, h)
 	if err != nil {
--- a/github/orgs_hooks_deliveries.go
+++ b/github/orgs_hooks_deliveries.go
@@ -1,3 +1,8 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package github
 
 import (
--- a/github/repos_hooks_deliveries.go
+++ b/github/repos_hooks_deliveries.go
@@ -39,6 +39,8 @@ func (d HookDelivery) String() string {
 	return Stringify(d)
 }
 
+// HookRequest is a part of HookDelivery that contains
+// the HTTP headers and the JSON payload of the webhook request.
 type HookRequest struct {
 	Headers    map[string]string `json:"headers,omitempty"`
 	RawPayload *json.RawMessage  `json:"payload,omitempty"`
@@ -48,6 +50,8 @@ func (r HookRequest) String() string {
 	return Stringify(r)
 }
 
+// HookResponse is a part of HookDelivery that contains
+// the HTTP headers and the response body served by the webhook endpoint.
 type HookResponse struct {
 	Headers    map[string]string `json:"headers,omitempty"`
 	RawPayload *json.RawMessage  `json:"payload,omitempty"`
--- a/github/repos_hooks_deliveries.go
+++ b/github/repos_hooks_deliveries.go
@@ -29,9 +29,9 @@ type HookDelivery struct {
 	InstallationID *string    `json:"installation_id"`
 	RepositoryID   *int64     `json:"repository_id"`
 
-	// Request is populated by GetHookDelivery
+	// Request is populated by GetHookDelivery.
 	Request *HookRequest `json:"request,omitempty"`
-	// Response is populated by GetHookDelivery
+	// Response is populated by GetHookDelivery.
 	Response *HookResponse `json:"response,omitempty"`
 }
 
--- a/github/repos_hooks_deliveries.go
+++ b/github/repos_hooks_deliveries.go
@@ -61,7 +61,7 @@ func (r HookResponse) String() string {
 //
 // GitHub API docs: https://docs.github.com/en/rest/reference/repos#list-deliveries-for-a-repository-webhook
 func (s *RepositoriesService) ListHookDeliveries(ctx context.Context, owner, repo string, id int64, opts *ListCursorOptions) ([]*HookDelivery, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/hooks/%d/deliveries", owner, repo, id)
+	u := fmt.Sprintf("repos/%v/%v/hooks/%v/deliveries", owner, repo, id)
 	u, err := addOptions(u, opts)
 	if err != nil {
 		return nil, nil, err
@@ -85,7 +85,7 @@ func (s *RepositoriesService) ListHookDeliveries(ctx context.Context, owner, rep
 //
 // GitHub API docs: https://docs.github.com/en/rest/reference/repos#get-a-delivery-for-a-repository-webhook
 func (s *RepositoriesService) GetHookDelivery(ctx context.Context, owner, repo string, hookID, deliveryID int64) (*HookDelivery, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/hooks/%d/deliveries/%d", owner, repo, hookID, deliveryID)
+	u := fmt.Sprintf("repos/%v/%v/hooks/%v/deliveries/%v", owner, repo, hookID, deliveryID)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
 		return nil, nil, err
--- a/github/repos_hooks_deliveries.go
+++ b/github/repos_hooks_deliveries.go
@@ -40,7 +40,7 @@ func (d HookDelivery) String() string {
 }
 
 type HookRequest struct {
-	Header     map[string]string `json:"headers,omitempty"`
+	Headers    map[string]string `json:"headers,omitempty"`
 	RawPayload *json.RawMessage  `json:"payload,omitempty"`
 }
 
@@ -49,7 +49,7 @@ func (r HookRequest) String() string {
 }
 
 type HookResponse struct {
-	Header     map[string]string `json:"headers,omitempty"`
+	Headers    map[string]string `json:"headers,omitempty"`
 	RawPayload *json.RawMessage  `json:"payload,omitempty"`
 }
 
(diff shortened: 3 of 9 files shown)


github/github.go
```diff
@@ -212,6 +212,9 @@ type ListCursorOptions struct {
 
 	// A cursor, as given in the Link header. If specified, the query only searches for events before this cursor.
 	Before string `url:"before,omitempty"`
+
+	// A cursor, as given in the Link header. If specified, the query continues the search using this cursor.
+	Cursor string `url:"cursor,omitempty"`
```
mumoshu, line 217: I believe this is the first GitHub API that uses `Cursor` for pagination. And I've added it to `ListCursorOptions` rather than `ListOptions` as... the name seemed to state that it wants all the possible list options for GitHub APIs that support pagination.

Please tell me if you'd like a brand new struct like `ListHookDeliveriesOptions`. Honestly, I'd personally prefer that as the Webhook Deliveries API as of today doesn't support other fields defined in ListCursorOptions. But go-github's current design looked more inclined to have a fewer number of common structs, rather than having a lot of specific structs. That's why I've chosen to change ListCursorOptions rather than having a new struct.

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+	h := new(HookDelivery)
+	resp, err := s.client.Do(ctx, req, h)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return h, resp, nil
+}
+
+func (d *HookDelivery) ParseRequestPayload() (interface{}, error) {
```
mumoshu, line 103: This is a utility function to ease parsing event payloads, and it is derived from the equivalent function for Events API https://github.com/google/go-github/blob/05e95d3c51abdf832f9430209ccb46b494ac7be8/github/event.go#L29-L31

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func (r HookResponse) String() string {
+	return Stringify(r)
+}
+
+// ListHookDeliveries lists webhook deliveries for a webhook configured in a repository.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/repos#list-deliveries-for-a-repository-webhook
+func (s *RepositoriesService) ListHookDeliveries(ctx context.Context, owner, repo string, id int64, opts *ListCursorOptions) ([]*HookDelivery, *Response, error) {
```
mumoshu, line 64: I"ve chosen to use the term `Hook` rather than `Webhook` across all the changes following  the fact that you seems to prefer the former according to the fact that there are `GetHook` for [Get repository webhooks API](https://docs.github.com/en/rest/reference/repos#get-a-repository-webhook) and `ListHooks` for [List repository webhooks API](https://docs.github.com/en/rest/reference/repos#list-repository-webhooks) and so on.

If you prefer the term `Webhook` or another one for some reason, please let me know!

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+	Response *HookResponse `json:"response,omitempty"`
+}
+
+func (d HookDelivery) String() string {
+	return Stringify(d)
+}
+
+type HookRequest struct {
+	Header     map[string]string `json:"headers,omitempty"`
+	RawPayload *json.RawMessage  `json:"payload,omitempty"`
```
mumoshu, line 45: I intentionally added `Raw-` prefix to the field name following the fact that the very similar `payload` field in the event object uses `RawPayload` https://github.com/google/go-github/blob/7a077bbd7313615d4f924f12b065913ee94e1d58/github/event.go#L13-L23

github/github.go
```diff
@@ -481,25 +489,33 @@ func (r *Response) populatePageValues() {
 			if err != nil {
 				continue
 			}
-			page := url.Query().Get("page")
-			if page == "" {
+
```
gmlewis, line 492: Hmmm... I'm not sure I'm happy with this section (especially regarding increasing the already-deep indentation levels).

Since `"cursor"` is the less-common and smaller case, let's please handle it first, and let's revert all the changes made to the handling of the `"page"` section, except maybe to pull out the shared `query` first... something like this:

```go
q := url.Query()
if cursor := q.Get("cursor"); cursor != "" {
...
}

page := q.Get("page")
if page == "" {
  continue
}
...
```

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+
+// HookDelivery represents the data that is received from GitHub's Webhook Delivery API
+//
+// GitHub API docs:
+// - https://docs.github.com/en/rest/reference/repos#list-deliveries-for-a-repository-webhook
+// - https://docs.github.com/en/rest/reference/repos#get-a-delivery-for-a-repository-webhook
+type HookDelivery struct {
+	ID             *int64     `json:"id"`
+	GUID           *string    `json:"guid"`
+	DeliveredAt    *time.Time `json:"delivered_at"`
```
gmlewis, line 23:
```suggestion
	DeliveredAt    *Timestamp `json:"delivered_at"`
```

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+	Redelivery     *bool      `json:"redelivery"`
+	Duration       *float64   `json:"duration"`
```
gmlewis, line 25: Shouldn't this be `*int64` ?  Where is this documented?
Neither of the URLs above have a `duration` field on their pages.

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+	Header     map[string]string `json:"headers,omitempty"`
+	RawPayload *json.RawMessage  `json:"payload,omitempty"`
+}
+
+func (r HookRequest) String() string {
+	return Stringify(r)
+}
+
+type HookResponse struct {
+	Header     map[string]string `json:"headers,omitempty"`
```
gmlewis, line 53:
```suggestion
	Headers     map[string]string `json:"headers,omitempty"`
```

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (r HookResponse) String() string {
+	return Stringify(r)
+}
+
+// ListHookDeliveries lists webhook deliveries for a webhook configured in a repository.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/repos#list-deliveries-for-a-repository-webhook
+func (s *RepositoriesService) ListHookDeliveries(ctx context.Context, owner, repo string, id int64, opts *ListCursorOptions) ([]*HookDelivery, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/hooks/%d/deliveries", owner, repo, id)
```
gmlewis, line 65:
```suggestion
	u := fmt.Sprintf("repos/%v/%v/hooks/%v/deliveries", owner, repo, id)
```

github/repos_hooks_deliveries.go
gmlewis: Why can't you reuse https://github.com/google/go-github/blob/master/github/event.go#L29-L132 here?

gmlewis (changes requested):
Thank you, @mumoshu.
I'm thinking that this can be improved by reusing more of the existing event code or refactoring it so that it is easier to reuse.

Also, the drop in code coverage is a bit concerning, but this might get fixed by reducing the amount of new code and reusing more of the existing code.

github/repos_hooks_deliveries.go
mumoshu: @gmlewis Hey! Interesting idea. How would you like to do that? That snippet in event.go switch on `Type`, where each value can be e.g. `CheckRunEvent`. On the other hand, this one switches on `event` field, where each value is e.g. `check_run`. They're very similar but different.

github/repos_hooks_deliveries.go
mumoshu: It isn't documented but looks like the below example. I've grabbed this by calling the list deliveries API on my own repo:

```
  {
    "id": 12076761161,
    "guid": "5a73eedc-d706-11eb-9a56-eb84e3dea363",
    "delivered_at": "2021-06-27T05:13:05Z",
    "redelivery": false,
    "duration": 0.29,
    "status": "Invalid HTTP Response: 530",
    "status_code": 530,
    "event": "push",
    "action": null,
    "installation_id": null,
    "repository_id": 366236975
  },
```

github/repos_hooks_deliveries.go
mumoshu: Fine! But out of curiosity, why do you prefer using `%v` for an integer value?

github/repos_hooks_deliveries.go
gmlewis: Awesome. Thank you. Please ignore this comment of mine.

github/github.go
mumoshu: Thanks! Addressed in f6d688a.

github/repos_hooks_deliveries.go
mumoshu: Addressed in https://github.com/google/go-github/pull/1934/commits/ff5a47a9e3d968697afa6cf425a5220f774f9c9c !

github/repos_hooks_deliveries.go
mumoshu: Fixed in https://github.com/google/go-github/pull/1934/commits/dffbb1cd1b1975ba93f77fb2d7331dffae56d198

github/repos_hooks_deliveries.go
mumoshu: Fixed in 5f0e1e4 anyway

ohpintu90:
Ok

github/repos_hooks_deliveries.go
mumoshu: Also note that this is almost fully covered by `TestHookDelivery_ParsePayload`, if that helps.

mumoshu:
@gmlewis Hey! I've addressed all the comments. Would you mind reviewing once again?

Also, please feel free to squash on merge, or just tell me to do so, if necessary. Thanks!

mumoshu:
FYI: I've added 431c77c for improving test coverage

github/repos_hooks_deliveries.go
gmlewis: I prefer `%v` for the following reasons:

* more consistent with the rest of the repo,
* idiomatic Go,
* and most importantly, it makes it much easier to search for an existing endpoint when you don't have to figure out the types of the segments of the URLs... you can more easily transform a URL from the documentation into a search string into the code base.

mumoshu: Gotcha! That totally makes sense. Thanks a lot for clarifying.

Regarding consistency with the rest of the repo, JFYI, I found a bunch of `%d` occurrences in the existing sources like `repos_hook.go`. Probably you'd want to fix it in another circumstance.

mumoshu:
I've added 23922a8 for more code coverage on `cursor` handling.

All the new codes should have been covered. I have no immediate plan to add more commits now so it should be a good timing for you to review again 😄

mumoshu:
Apparently, the last commit 23922a8 seems not included in https://app.codecov.io/gh/google/go-github/branch/webhook-deliveries-api yet. But theoretically speaking the code coverage shouldn't get decreased now, as I've already added enough tests to cover all the new code paths.

github/github.go
gmlewis: I think adding it to `ListCursorOptions` is fine.  Thanks.

gmlewis:
> Also, please feel free to squash on merge, or just tell me to do so, if necessary. Thanks!

We always squash and merge in this repo, so don't worry about the number of commits.  In fact, we prefer that author's don't force-push so that it is easier for reviewers to see what changes have been made since the last review.

github/github.go
```diff
@@ -481,7 +489,19 @@ func (r *Response) populatePageValues() {
...
+			q := url.Query()
+
+			if cursor := q.Get("cursor"); cursor != "" {
+				for _, segment := range segments[1:] {
+					switch strings.TrimSpace(segment) {
+					case `rel="next"`:
+						r.Cursor = cursor
+					}
+				}
+			}
```
gmlewis, line 502: Are `"cursor"` and `"page"` mutually exclusive?  (I think the answer is "yes" but I could be wrong.)

If so, do you want to add a `continue` within this `if` block just to make that clear?
I'm actually fine either way... your call.

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,204 @@
...
+	Redelivery     *bool      `json:"redelivery"`
+	Duration       *float64   `json:"duration"`
+	Status         *string    `json:"status"`
+	StatusCode     *int       `json:"status_code"`
+	Event          *string    `json:"event"`
+	Action         *string    `json:"action"`
+	InstallationID *string    `json:"installation_id"`
+	RepositoryID   *int64     `json:"repository_id"`
+
+	// Request is populated by GetHookDelivery
```
gmlewis, line 32: nit:

```suggestion
	// Request is populated by GetHookDelivery.
```

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,204 @@
...
+	Request *HookRequest `json:"request,omitempty"`
+	// Response is populated by GetHookDelivery
```
gmlewis, line 34: nit:

```suggestion
	// Response is populated by GetHookDelivery.
```

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,204 @@
...
+	Response *HookResponse `json:"response,omitempty"`
+}
+
+func (d HookDelivery) String() string {
+	return Stringify(d)
+}
+
+type HookRequest struct {
```
gmlewis, line 42: Can you please add a comment here describing `// HookRequest ...` ?

github/repos_hooks_deliveries.go
```diff
@@ -0,0 +1,204 @@
...
+	Headers    map[string]string `json:"headers,omitempty"`
+	RawPayload *json.RawMessage  `json:"payload,omitempty"`
+}
+
+func (r HookRequest) String() string {
+	return Stringify(r)
+}
+
+type HookResponse struct {
```
gmlewis, line 51: Can you please add a comment here describing `// HookResponse ...` ?

github/repos_hooks_deliveries.go
gmlewis: What I'm thinking here is to make a helper function that both `ParsePayload` and `ParseRequestPayload` both call.

The helper method could take an anonymous function to perform the required mapping... one of them would use the `eventTypeMethod` map found in [messages.go](https://github.com/google/go-github/blob/master/github/messages.go#L45).

Let me know if you don't think this is possible and I'll download your PR and see if I can get it to work.

gmlewis (changes requested):

github/github.go
mumoshu: Thanks! Yes, I believe so, although it isn't documented anywhere. Added `continue` in 72c3380

github/repos_hooks_deliveries.go
mumoshu: Addressed in 5b86aff 👍

github/repos_hooks_deliveries.go
mumoshu: Addressed in 5b86aff 👍

github/repos_hooks_deliveries.go
mumoshu: Added in 194eff6 👍

github/repos_hooks_deliveries.go
mumoshu: Added in 194eff6 👍

mumoshu:
> What I'm thinking here is to make a helper function that both ParsePayload and ParseRequestPayload both call.

Sounds good!

> The helper method could take an anonymous function to perform the required mapping... one of them would use the eventTypeMethod map found in messages.go.

And would you mind if I also created another, reverse map of `eventTypeMethod`, named like `typeEventMapping`, and use that from `ParsePayload`?

mumoshu:
Apparently, another method named `ParseWebhook` that calls `ParsePayload` seems to already use `eventTypeMapping` on its own 🤔 Probably adding second use of `eventTypeMapping` in an anonymous function defined in `ParseWebhook` doesn't make much sense?

Perhaps we'd better revise `ParseWebHook` as well? I'm pretty lost on how I should go.

gmlewis:
> I'm pretty lost on how I should go.

OK. I'll take a look and get back to you.

gmlewis:
> OK. I'll take a look and get back to you.

I think I've got a simple solution for you.  It also caught 3 errors in your test table, which I think justifies the reuse of the existing code.  I'll add review comments so you can see what I did.

Let me know what you think.

github/repos_hooks_deliveries.go
gmlewis: Here's my proposed replacement for this method:

```go
// ParseRequestPayload parses the request payload. For recognized event types,
func (d *HookDelivery) ParseRequestPayload() (interface{}, error) {
	eType, ok := eventTypeMapping[*d.Event]
	if !ok {
		return nil, fmt.Errorf("unsupported event type %q", *d.Event)
	}

	e := &Event{Type: &eType, RawPayload: d.Request.RawPayload}
	return e.ParsePayload()
}
```

github/repos_hooks_deliveries_test.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+var hookDeliveryPayloadTypeToStruct = map[string]interface{}{
+	"check_run":                      &CheckRunEvent{},
+	"check_suite":                    &CheckSuiteEvent{},
+	"commit_comment":                 &CommitCommentEvent{},
+	"content_reference":              &ContentReferenceEvent{},
+	"create":                         &CreateEvent{},
+	"delete":                         &DeleteEvent{},
+	"deploy_ket":                     &DeployKeyEvent{},
```
gmlewis, line 117:
```suggestion
	"deploy_key":                     &DeployKeyEvent{},
```

github/repos_hooks_deliveries_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"fork":                           &ForkEvent{},
+	"github_app_authorization":       &GitHubAppAuthorizationEvent{},
+	"gollum":                         &GollumEvent{},
+	"installation":                   &InstallationEvent{},
+	"installation_repositories":      &InstallationRepositoriesEvent{},
+	"issue_comment":                  &IssueCommentEvent{},
+	"issues":                         &IssuesEvent{},
+	"label":                          &LabelEvent{},
+	"marketplace_purchase":           &MarketplacePurchaseEvent{},
+	"member_event":                   &MemberEvent{},
```
gmlewis, line 129:
```suggestion
	"member":                   &MemberEvent{},
```

github/repos_hooks_deliveries_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"membership_event":               &MembershipEvent{},
```
gmlewis, line 130:
```suggestion
	"membership":               &MembershipEvent{},
```

github/repos_hooks_deliveries_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	d := &HookDelivery{
+		Event: String("some_invalid_event"),
+		Request: &HookRequest{
+			RawPayload: &p,
+		},
+	}
+
+	_, err := d.ParseRequestPayload()
+	if err == nil || err.Error() != "unexpected end of JSON input" {
```
gmlewis, line 201:
```suggestion
	if err == nil || err.Error() != `unsupported event type "some_invalid_event"` {
```

gmlewis (changes requested):
Hi @mumoshu - here are my suggestions for reusing existing code, along with a few error corrections. Please let me know if they make sense to you.

Please remember to run `gofmt` after accepting my suggestions, as they will need proper formating.

mumoshu:
@gmlewis Thank you so much for taking your time and the suggestions! All looked awesome and I've included your suggestions in the last commit, 1381c0f.

Regarding the change on `ParseRequestPayload`, I initially had some preconceived idea that it won't be acceptable to create a new `Event` object just for calling ParsePayload. Apparently, it was the best impl though. Good job!

gmlewis (approved):
Thank you, @mumoshu !
LGTM.

mumoshu:
@gmlewis Hey! Would you mind if I added another commit for adding similar support for organization variants of these APIs?

https://github.com/mumoshu/go-github/commit/0a3739c94ba1eb5fe9f040f465b1e5387eb802e1.

It already has full test coverage, and I've manually verified it to work as part of https://github.com/actions-runner-controller/actions-runner-controller/pull/682, too.

mumoshu:
@gmlewis Hey! Would you mind if I added another commit for adding similar support for organization variants of these APIs?

https://github.com/mumoshu/go-github/commit/0a3739c94ba1eb5fe9f040f465b1e5387eb802e1.

It already has full test coverage, and I've manually verified it to work as part of https://github.com/actions-runner-controller/actions-runner-controller/pull/682, too.

@Parker77 Hi! If you're reading this and going to put the second LGTM, please go ahead!

I'm just trying to reduce the time required on both ends to make the set of changes related to #1933. If you merged this, I'll just submit another PR for https://github.com/mumoshu/go-github/commit/0a3739c94ba1eb5fe9f040f465b1e5387eb802e1.

gmlewis:
> @gmlewis Hey! Would you mind if I added another commit for adding similar support for organization variants of these APIs?

Hi @mumoshu - sure, you can add more commits if you wish... please just push and don't use "force push" so that it will be easier to review more recent additions. Thank you.

mumoshu:
@gmlewis Thanks for confirming! I've added 0a3739c

github/orgs_hooks_deliveries.go
```diff
@@ -0,0 +1,48 @@
+package github
```
gmlewis, line 1: Please add the copyright header to this new file (with year 2021 and a blank line before the `package` line).

github/orgs_hooks_deliveries.go
```diff
@@ -0,0 +1,48 @@
...
+// GetHookDelivery returns a delivery for a webhook configured in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/orgs#get-a-webhook-delivery-for-an-organization-webhook
+func (s *OrganizationsService) GetHookDelivery(ctx context.Context, owner string, hookID, deliveryID int64) (*HookDelivery, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/hooks/%v/deliveries/%v", owner, hookID, deliveryID)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	h := new(HookDelivery)
```
gmlewis, line 41: Please add a blank line above this line for consistency with this repo and idiomatic Go (after an `if err` block).

gmlewis (changes requested):
Thank you, @mumoshu !
Just a couple minor tweaks, please.

github/orgs_hooks_deliveries.go
mumoshu: Thanks! Addressed in 123bafa

github/orgs_hooks_deliveries.go
mumoshu: Addressed in fc8d20ffc 👍

gmlewis (approved):
Thank you, @mumoshu !
LGTM.

Parker77 (approved):
LGTM.

PR #1963, opened 2021-07-10
Add test cases for JSON resource marshaling

sagar23sj:
Resource Covered :

* TeamChange
* Page

Helps with issue: #55

### DIFF
--- a/github/event_types_test.go
+++ b/github/event_types_test.go
@@ -146,3 +146,27 @@ func TestProjectColumnChange_Marshal_NameChange(t *testing.T) {
 
 	testJSONMarshal(t, u, want)
 }
+
+func TestPage_Marshal(t *testing.T) {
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @sagar23sj !
LGTM.

By the way, is there any reason why you don't just combine all the changes for a single file into one PR?

sagar23sj:
I generally do it that way only @gmlewis , but there are a lot of JSON resources in event_types.go file and it was taking time good amount of time to even test few of them. So i decided to  test for 4-5 resources in a single PR, so that i don't get blocked for a long time since i find time to do this in between my regular work schedule.

gmlewis:
OK, that's fine. It seems like GitHub was having merge issues before on the multiple additions to the single file. We'll see how it goes this time. 😁

PR #1975, opened 2021-07-12
Add test cases for JSON resource marshaling

sagar23sj:
Resources Covered :
- InteractionRestriction

Helps with issue: #55

### DIFF
--- a/github/interactions_test.go
+++ b/github/interactions_test.go
@@ -1,0 +1,26 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import "testing"
+
+func TestInteractionRestriction_Marshal(t *testing.T) {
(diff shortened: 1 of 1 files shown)


github/interactions_test.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2018 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2021 The go-github AUTHORS. All rights reserved.
```

gmlewis (changes requested):
Thank you, @sagar23sj !
One minor tweak to fix, please, and then it LGTM.
Whenever a new file is created, we use the year it was created as its copyright year.

github/interactions_test.go
sagar23sj: Sure @gmlewis. Will make the change.
Thank you for your instantaneous feedback.

gmlewis (approved):
Thank you, @sagar23sj !
LGTM.

Parker77 (approved):
LGTM.

PR #1980, opened 2021-07-12
Add draft for mock http client for unittesting

migueleliasweb:


gmlewis:
Thanks, @migueleliasweb .
Before doing a full code review, let's please get all the tests passing.

It is currently stuck on:

```
Run go vet ./...
# github.com/google/go-github/v37/mock
vet: mock/client.go:62:20: NopCloser not declared by package io
Error: Process completed with exit code 2.
```

migueleliasweb:
> Thanks, @migueleliasweb .
> Before doing a full code review, let's please get all the tests passing.
> 
> It is currently stuck on:
> 
> ```
> Run go vet ./...
> # github.com/google/go-github/v37/mock
> vet: mock/client.go:62:20: NopCloser not declared by package io
> Error: Process completed with exit code 2.
> ```

This is due to the differences in golang 1.16. I changed to use the `ioutil` package instead.

gmlewis:
Now the error is:

```
--- FAIL: TestMockClient (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0

goroutine 6 [running]:
testing.tRunner.func1.2(0x8ce040, 0xc00001e570)
	/opt/hostedtoolcache/go/1.16.5/x64/src/testing/testing.go:1143 +0x49f
testing.tRunner.func1(0xc000001800)
	/opt/hostedtoolcache/go/1.16.5/x64/src/testing/testing.go:1146 +0x695
panic(0x8ce040, 0xc00001e570)
	/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:971 +0x499
github.com/google/go-github/v37/mock.(*MockRoundTripper).RoundTrip(0xc000010290, 0xc000146600, 0x0, 0x0, 0x0)
	/home/runner/work/go-github/go-github/mock/client.go:52 +0x5c5
net/http.send(0xc000146600, 0x971340, 0xc000010290, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc0000102b0, 0xc00016b508)
	/opt/hostedtoolcache/go/1.16.5/x64/src/net/http/client.go:251 +0x6db
net/http.(*Client).send(0xc00006d5f0, 0xc000146600, 0x0, 0x0, 0x0, 0xc0000102b0, 0x0, 0x1, 0x730000004a82e2)
	/opt/hostedtoolcache/go/1.16.5/x64/src/net/http/client.go:175 +0x1d6
net/http.(*Client).do(0xc00006d5f0, 0xc000146600, 0x0, 0x0, 0x0)
	/opt/hostedtoolcache/go/1.16.5/x64/src/net/http/client.go:717 +0x2cc
```

migueleliasweb:
> Now the error is:
> 
> ```
> --- FAIL: TestMockClient (0.00s)
> panic: runtime error: index out of range [0] with length 0 [recovered]
> 	panic: runtime error: index out of range [0] with length 0
> 
> goroutine 6 [running]:
> testing.tRunner.func1.2(0x8ce040, 0xc00001e570)
> 	/opt/hostedtoolcache/go/1.16.5/x64/src/testing/testing.go:1143 +0x49f
> testing.tRunner.func1(0xc000001800)
> 	/opt/hostedtoolcache/go/1.16.5/x64/src/testing/testing.go:1146 +0x695
> panic(0x8ce040, 0xc00001e570)
> 	/opt/hostedtoolcache/go/1.16.5/x64/src/runtime/panic.go:971 +0x499
> github.com/google/go-github/v37/mock.(*MockRoundTripper).RoundTrip(0xc000010290, 0xc000146600, 0x0, 0x0, 0x0)
> 	/home/runner/work/go-github/go-github/mock/client.go:52 +0x5c5
> net/http.send(0xc000146600, 0x971340, 0xc000010290, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc0000102b0, 0xc00016b508)
> 	/opt/hostedtoolcache/go/1.16.5/x64/src/net/http/client.go:251 +0x6db
> net/http.(*Client).send(0xc00006d5f0, 0xc000146600, 0x0, 0x0, 0x0, 0xc0000102b0, 0x0, 0x1, 0x730000004a82e2)
> 	/opt/hostedtoolcache/go/1.16.5/x64/src/net/http/client.go:175 +0x1d6
> net/http.(*Client).do(0xc00006d5f0, 0xc000146600, 0x0, 0x0, 0x0)
> 	/opt/hostedtoolcache/go/1.16.5/x64/src/net/http/client.go:717 +0x2cc
> ```

For some reason this seems to be a flakey test :thinking: It passes most times but eventually fails. I'm troubleshooting....

migueleliasweb:
The tests should pass now :crossed_fingers: . Ping @gmlewis :bell: .

gmlewis:
OK, thanks, @migueleliasweb .

I have some big concerns about this approach...  especially from a maintenance perspective.

* to support the full library, each endpoint will need to be (manually most likely) added and the resulting size is going to be huge
* this is driven with regexps which are challenging to always get right (and not interfere with each other)

Before doing a full code review (where I would suggest that headers be added and `Marshal` should have a single `l`, for example), I would like to hear what @gauntface and @willnorris think about this approach.

migueleliasweb:
> OK, thanks, @migueleliasweb .
> 
> I have some big concerns about this approach... especially from a maintenance perspective.
> 
> * to support the full library, each endpoint will need to be (manually most likely) added and the resulting size is going to be huge
> * this is driven with regexps which are challenging to always get right (and not interfere with each other)
> 
> Before doing a full code review (where I would suggest that headers be added and `Marshal` should have a single `l`, for example), I would like to hear what @gauntface and @willnorris think about this approach.

Damn it, got Marshal wrong. I always spell it the wrong way. Second language problems, sry.

Yeah,I'm afraid you're not wrong about the regexps.

I was trying to come up with another way to maybe generate the code to match the requests but I ended up trying this one first. I thought of two other ways:

- Try to use `runtime.CallersFrames` inside the custom `RoundTripper` to determine the last call made in the `go-github` code thus finding the caller function in the SDK. With this approach, the matcher code could be something much more clear like a string `UsersService.Get` instead of the somewhat ugly regexps.
- Add a header to the request and match the request to that inside the `RoundTripper` to make it easier to distinguish between calls

migueleliasweb:
I just wrote a poc with the runtime frames. The API looks much cleaner without the `regexps` and I reckon it would be doable to auto generate all endpoints in this case. Have a look: https://github.com/migueleliasweb/go-github/compare/mock-transport...migueleliasweb:mock-transport-pc-frames?expand=1

gmlewis:
> I just wrote a poc with the runtime frames. The API looks much cleaner without the `regexps` and I reckon it would be doable to auto generate all endpoints in this case.
> 
> Have a look: https://github.com/migueleliasweb/go-github/blob/mock-transport-pc-frames/mock/client.go

Ah, wow... that is quite intriguing, @migueleliasweb !

I'm guessing that the GitHub API call would be closer to the "recent" end of the call stack as opposed to the "oldest" end of the call stack...  so it might make the most sense (from a performance perspective) to search for the call from the end to the beginning... actually maybe you are already doing that... if so, great.

I'm not too concerned about the performance implications of the endpoint search... I think this would be much easier to support than the other solution... especially if the critical list of methods that users would use are auto-generated.

Overall, I like this approach a lot...  I think you are on to something big here!

willnorris:
(just some random thoughts...) 

The use of caller frames to match the intercepted request is certainly an interesting approach, and one I probably wouldn't have considered.  But I think it's a bit more rigid in tying the test to the method that made the http request, rather than what the request itself is.  

This is very close to being a really interesting, generic, standalone approach to intercepting and mocking outbound http requests.  If that were the goal (to make this reusable beyond just go-github), then I wonder if the original URL matching approach might be more flexible.  I also wonder if some of the regex logic could be made safer (or at least more maintainable) by using some of the path parsing and matching logic in something like gorillatoolkit (or similar framework)?  That could allow writing paths to match using a friendlier syntax like `/users/{user}`

migueleliasweb:
> (just some random thoughts...)
> 
> The use of caller frames to match the intercepted request is certainly an interesting approach, and one I probably wouldn't have considered. But I think it's a bit more rigid in tying the test to the method that made the http request, rather than what the request itself is.
> 

The problem as @gmlewis already pointed out, there's not (as far as I know) way to determine which call was made in the current state of the codebase. Using the regexps could end up being a bit tedious to handcraft all the required expressions and ensure they don't collide to each other. The workaround was the `runtime.Frames` poc I wrote.

> I also wonder if some of the regex logic could be made safer (or at least more maintainable) by using some of the path parsing and matching logic in something like gorillatoolkit (or similar framework)? That could allow writing paths to match using a friendlier syntax like `/users/{user}`

Using something like the gorillamux would definitely make the request matching much more friendly but I thought this repo in general was trying to not bring more remote dependencies. And also, once again, maintaining all the paths, although it would be much easier with the gorilla mux or something similar, would still mean all those paths would have to be handcrafted and maintained for the future. 

Thanks for the feedbackm @willnorris ! :+1:

migueleliasweb:
I will try to add the code generation to the `runtime.Frame` poc PR for some of the service methods to see how it would look.

migueleliasweb:
Closing in favor of https://github.com/google/go-github/pull/2008

migueleliasweb:
> Even if this were not fully generic, and is instead specific to go-github, this is only intended for use by users of go-github. It's not used by the library itself. So I think this should really be its own repo, which we can optionally link to from go-github's README as one approach to testing applications that use the library. (/cc @gmlewis)

It is true that this code is more like a helper other than something integral to this repo so I see no big problem of moving this logic to a new repo. Something like `go-github-testing` or similar.

Where this new repo would reside? If inside `google`, I have no access, so the repo would have to be created by someone with proper access to so I could create a PR and we follow up the process in there. If not, I could create it under my user (or any of your users as well).

Also, if we're decided to move this PR to a new repo, @willnorris idea of using like `gorilla` to have a much more friendly URL matching would become doable. So I reckon I could close https://github.com/google/go-github/pull/2008 and improve this one.

gmlewis:
> Where would this new repo reside?

I see no reason why it couldn't be under your own user name, but we'll see what @willnorris says.

willnorris:
yeah, I was imagining just under your own account, at least to start out.  I don't think it will ever make sense to put it under @google, since no one from Google is actively involved with this at this point.

migueleliasweb:
Hi @willnorris and @gmlewis I think I have something that is somewhat presentable. Let me know if, other than the missing CI, there's something else I'm missing.

https://github.com/migueleliasweb/go-github-mock

ps: don't know where to have this conversation, if you guys prefer to create issues in the repo itself, feel free

gmlewis:
Sounds great, @migueleliasweb !  Thanks for doing that!

Feel free to create a PR to add a reference to your new repo in this repo's README.md so we can point people to it.

migueleliasweb:
Will do, @gmlewis . I just wanted a couple more hours to ensure the tests are being run in the CI workflow before opening the floodgates :joy: .

PR #1998, opened 2021-07-14
Add allows_public_repositories to actions_runner_group

### LINKED ISSUE #1996: We do not set allows_public_repositories for actions runner groups
We currently cannot set the `allows_public_repositories` variable wich is an entity of a runner group and controls weither public repos can use that group. If it is set to false, public repos, even if they are specifically in the selected repos list, will not be able to use the runners from this group.

The docs on this are:
https://docs.github.com/en/rest/reference/actions#create-a-self-hosted-runner-group-for-an-organization

However there is a documentation bug since we tried it out and one can actually pass this value to api in the request body and not only gets it as response. It will be set properly. This is not how it is documented in GitHub API Docs. Thus we raised an issue to change docs on this since we guess it is intended to be used in request body and just not properly documented.
https://github.com/github/docs/issues/8209

However we still miss the implementation in go-github. 
So currently all Runner Groups created with this lib are  just for private repo since this flag is defaulted to false.

FloThinksPi:
Adds allows_public_repositories to request bodies of the
CreateRunnerGroup as well as UpdateRunnerGroup methods for github
actions.

This allows us to set the allows_public_repositories flag even if it is
not yet properly documented see https://github.com/github/docs/issues/8209
That flag controls if public repos can use runners of this group.
If it is disabled(default) even public repos that are added to the
selected_repositories list cannot use runners of this group.

With this change we can now set this properly and thus use runner groups
created by go-github also for public repos.

Fixes #1996

### DIFF
--- a/github/actions_runner_groups.go
+++ b/github/actions_runner_groups.go
@@ -36,12 +36,15 @@ type CreateRunnerGroupRequest struct {
 	SelectedRepositoryIDs []int64 `json:"selected_repository_ids,omitempty"`
 	// Runners represent a list of runner IDs to add to the runner group.
 	Runners []int64 `json:"runners,omitempty"`
+	// If set to True, public repos can use this runner group
+	AllowsPublicRepositories *bool `json:"allows_public_repositories,omitempty"`
 }
 
 // UpdateRunnerGroupRequest represents a request to update a Runner group for an organization.
 type UpdateRunnerGroupRequest struct {
-	Name       *string `json:"name,omitempty"`
-	Visibility *string `json:"visibility,omitempty"`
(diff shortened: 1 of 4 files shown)


FloThinksPi:
I mirrored the style from the existing structs to do pass by reference so having a bool pointer. 
Is this the wanted way/interface for users ?

gmlewis (changes requested):
Thank you, @FloThinksPi !
Please run `go generate ./...` and push (not force-push) the changes to this PR, as described in our CONTRIBUTING.md document.

gmlewis (approved):
Thank you, @FloThinksPi !
LGTM.
Merging.

Issue #1859, opened 2021-04-27
Modules: Sub-Dependency is using v33, main project v35 - Type incompatibility + Module replacement


andygrunwald:
Hey everyone,
thanks for maintaining this library, keeping everything in shape, and continuously release new versions.
I am facing the challenge of using the new versions:

### Sub-Dependency is using v33, main project v35 - Type incompatibility

Imagine Application A that is using
- this library in v35
- another library (e.g., https://github.com/palantir/go-githubapp) which is using v33

Independent of the breaking changes in versions (lets keep them out for now), when using both libraries, errors like
```go
applicationA/file.go:129:89: cannot use listReposOpts (type *"github.com/google/go-github/v35/github".ListOptions) as type *"github.com/google/go-github/v33/github".ListOptions in argument to installationClient.Apps.ListRepos
```

What is the preferred way to solve this?

I can think of multiple solutions:

**1) Updating dependency in sub-library**
If the sub-library is from a different vendor/user, you don't have control over it and might lead to forking.

**2) Cast the type `v33/github".ListOptions` to `v35/github".ListOptions`**
This might work if the two types are the same. If this type is part of a breaking change, this might not work that simple and the code would be bloated with transferring the one type into another.

**3) Module replacement**
Here I am not sure. Is there a way on how Application A can force a module replacement for all other sublibraries?

This might not be an issue with the library itself, but rather to the usage.
However, I would like to get your thoughts on this and if you see a solution to this challenge.
Happy to provide further information.

gmlewis:
My knee-jerk reaction is to fork https://github.com/palantir/go-githubapp, update it to `v35`, and submit a PR (or use the forked version).

I'll leave this open as others might have better responses since my knee is not always correct.  :smile:

andygrunwald:
Thanks, @gmlewis for the quick response. That is,  indeed, what I did anyway, see https://github.com/palantir/go-githubapp/pull/81
However, I suppose this is not the only case or I am not the only one facing this.

bluekeyes:
My understanding is that this is a known consequence of how Go modules work (specifically, the fact that the major version appears in the import path and the import path is what determines if two modules are the same) and @gmlewis's suggestion to fork the library is the expected solution.

I believe the `replace` directive won't work in this case because Go considers `v33` and `v35` to be completely separate modules and therefore not eligible for replacement with each other.

----

[...]

PR #2015, opened 2021-07-20
Expose Client.Client

gmlewis:
Fixes: #1859.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -137,7 +137,7 @@ var errNonNilContext = errors.New("context must be non-nil")
 // A Client manages communication with the GitHub API.
 type Client struct {
 	clientMu sync.Mutex   // clientMu protects the client during calls that modify the CheckRedirect func.
-	client   *http.Client // HTTP client used to communicate with the API.
+	Client   *http.Client // HTTP client used to communicate with the API.
 
 	// Base URL for API requests. Defaults to the public GitHub API, but can be
 	// set to a domain endpoint to use with GitHub Enterprise. BaseURL should
@@ -273,7 +273,7 @@ func NewClient(httpClient *http.Client) *Client {
 	baseURL, _ := url.Parse(defaultBaseURL)
 	uploadURL, _ := url.Parse(uploadBaseURL)
 
-	c := &Client{client: httpClient, BaseURL: baseURL, UserAgent: userAgent, UploadURL: uploadURL}
+	c := &Client{Client: httpClient, BaseURL: baseURL, UserAgent: userAgent, UploadURL: uploadURL}
 	c.common.client = c
 	c.Actions = (*ActionsService)(&c.common)
 	c.Activity = (*ActivityService)(&c.common)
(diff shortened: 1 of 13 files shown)


github/github.go
```diff
@@ -137,7 +137,7 @@ var errNonNilContext = errors.New("context must be non-nil")
 // A Client manages communication with the GitHub API.
 type Client struct {
 	clientMu sync.Mutex   // clientMu protects the client during calls that modify the CheckRedirect func.
```
bluekeyes, line 139: Will exposing `client` directly lead to problems with this mutex? I think two different `github.Client` objects would each use separate locks to protect the same `*http.Client` object, leading to race conditions.

Maybe instead there could be a new `Client()` or `HTTPClient()` function that returns a shallow copy of the internal `client` field? That would enable reuse of the transport and timeout while allowing each instance to safely update the redirect function.

gmlewis: Hmmm... that's an interesting question...

The idea was to prevent multiple calls that check for redirects to avoid swapping out the client at the same time.

But maybe a read-only `Client()` method as you recommend would be a better solution anyway.

jeffmendoza:
Tested this with my real-world code that could use it (below). Works great, I'm not using any concurrency.
```go
// Check performs the policy check for Binary Artifacts based on the
// configuration stored in the org/repo, implementing policydef.Policy.Check()
func (b Binary) Check(ctx context.Context, c *github.Client, owner,
	repo string) (*policydef.Result, error) {
	oc, rc := getConfig(ctx, c, owner, repo)
	enabled := config.IsEnabled(oc.OptConfig, rc.OptConfig, repo)
	log.Info().
		Str("org", owner).
		Str("repo", repo).
		Str("area", polName).
		Bool("enabled", enabled).
		Msg("Check repo enabled")

	gh32c := gh32.NewClient(c.Client)
	repoClient := githubrepo.CreateGithubRepoClient(ctx, gh32c)
	if err := repoClient.InitRepo(owner, repo); err != nil {
		return nil, err
	}
	l := logger{}
	cr := &checker.CheckRequest{
		Ctx:         ctx,
		Client:      gh32c,
		RepoClient:  repoClient,
		HTTPClient:  nil,
		Owner:       owner,
		Repo:        repo,
		GraphClient: nil,
		Logf:        l.Logf,
	}
	// TODO, likely this should be a "scorecard" policy that runs multiple checks
	// here, and uses config to enable/disable checks.
	res := checks.BinaryArtifacts(cr)
	if res.Error != nil {
		return nil, res.Error
	}
```

gmlewis:
Thank you, @jeffmendoza - hey, how about if I make a read-only version of the Client as described by @bluekeyes [above](https://github.com/google/go-github/pull/2015#discussion_r673418705) ?

jeffmendoza:
I don't see any reason why that wouldn't work as well.

gmlewis:
Closing due to preferred solution in #2016.

PR #2019, opened 2021-07-21
Set up Dependabot for gomod and github-actions dependencies 

mmorel-35:

PR #2020, opened 2021-07-21
Set up workflow for golangci

mmorel-35:
This PR doesn't fix the identified problem raised by the analysis

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,18 @@
+on: [push, pull_request]
+name: linter
+
+jobs:
+  lint:
+    strategy:
+      matrix:
+        go-version: [1.x]
+        platform: [ubuntu-latest]
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v2
+      with:
+        version: v1.41
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,4 +16,4 @@ jobs:
       uses: golangci/golangci-lint-action@v2
       with:
         version: v1.41
-        args: --verbose
+        args: --verbose --issues-exit-code=0
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,3 +16,4 @@ jobs:
       uses: golangci/golangci-lint-action@v2
       with:
         version: v1.41
+        args: --verbose
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,4 +16,4 @@ jobs:
       uses: golangci/golangci-lint-action@5c56cd6c9dc07901af25baab6f2b0d9f3b7c3018 #v2.5.2
       with:
         version: v1.41
-        args: --verbose --issues-exit-code=0
+        args: --verbose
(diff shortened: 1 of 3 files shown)


gmlewis:
What prompted you to write this PR?

mmorel-35:
Hi! I saw it defined in https://github.com/google/go-querystring, I thought the project might be interested

gmlewis:
@willnorris - I defer to your judgment on this one and #2019.

.golangci.yml
```diff
@@ -0,0 +1,13 @@
+linters:
+  enable:
+    - dogsled
```
obalunenko, line 3: suggest to add ercheck and govet to the list
https://golangci-lint.run/usage/linters/

mmorel-35: Those two are active by default, see https://golangci-lint.run/usage/linters/#enabled-by-default-linters. I haven't deactivated them

.github/workflows/linter.yml
```diff
@@ -0,0 +1,19 @@
...
+      matrix:
+        go-version: [1.x]
+        platform: [ubuntu-latest]
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v2
```
willnorris, line 16:
```suggestion
      uses: golangci/golangci-lint-action@5c56cd6c9dc07901af25baab6f2b0d9f3b7c3018 #v2.5.2
```

willnorris: Google requires using exact SHAs for third-party actions.

willnorris (changes requested):

.github/workflows/linter.yml
mmorel-35: Done :)!

willnorris (approved):

willnorris:
so [that's a lot of errors](https://github.com/google/go-github/runs/3364021858?check_suite_focus=true#step:3:67).  I see that you have golangci-lint returning a 0 status, so this doesn't break builds.  What's the plan then... a handful of follow-up PRs that address these individual errors, and then we eventually remove the `--issues-exit-code=0`?  The other approach I've taken in the past is to just disable the failing linters, and then in the commits that fix the errors, also re-enable the linter.

We might also end up wanting to disable some of these, or at least for certain paths.  For example, the dogsled errors in test files are probably not worth fixing.  Same for the duplicate code detection.  Our tests do duplicate some code and that's fine.  Trying to refactor to remove the duplication is likely to make the tests less readable (at least in some cases).

mmorel-35:
The --issues-exit-code=0 is to display the actual triggered warnings/errors without making the job failing indeed.
Then it's entirely up to you to decide what to do, if you want to remove a l'inter or just some rules indeed

willnorris:
What about something like:

``` yaml
linters:
  # TODO: fix errors so that all of the linters below pass.
  # The linters that are commented out, as well as those explicitly disabled,
  # are currently failing.  We should fix those failures or define exclusion
  # rules, and then enable those linters.
  enable:
    - dogsled
    - dupl
    - gofmt
    - goimports
  # - gosec
    - misspell
    - nakedret
  # - stylecheck
  # - unconvert
  # - unparam
  # - whitespace
  disable:
    - errcheck
    - gosimple
    - staticcheck
    - ineffassign
    - unused
issues: 
  exclude-rules:
  - linters:
    - dogsled
    text:  "declaration has 3 blank identifiers"
    path: _test\.go
  - linters:
    - dupl
    path: _test\.go
```

then, as folks want to fix these, they can enable a specific linter locally, run golangci-lint to see what's failing, and fix those errors.  That doesn't give us visibility in the CI logs of what is failing, but I think that's okay for now, especially since these won't take very long to fix.

PR #2032, opened 2021-08-06
Deprecate WebHookPayload

### LINKED ISSUE #303: Deprecate WebHookPayload
The `PushEvent` type was originally added in June 2013 in beaff13913b3dd17651e12f622473b1fb99756a1.  I then added the `WebHookPayload` type a month later in 023e8db8cbba83d90a62e2e546a96d3dcecbbce4.  I honestly am not sure why I did that, since they are effectively duplicates.  Maybe it wasn't clear at that point that we should be using the same types for webhook payloads and activity events?  Or maybe I just wasn't thinking.

In any event, #302 reminded me that we've updated `PushEvent` to include a variant of `Repository` that has a string Organization filed (see d722750ab19bf6318a2c8dfad8d4aa034779aa8f), but the same update was never made to `WebHookPayload`.  Instead of fixing that, I think we should probably just deprecate `WebHookPayload` and point people to `PushEvent` instead.

gmlewis:
Fixes #303.


gmlewis:
> ```diff
> - Coverage   97.87%   97.87%   -0.01%     
> ```

This looks like a Codecov bug!  Please ignore.

github/repos_hooks.go
```diff
@@ -11,64 +11,27 @@ import (
...
 )
 
+// Deprecated: Please use PushEvent instead.
 // WebHookPayload represents the data that is received from GitHub when a push
 // event hook is triggered. The format of these payloads pre-date most of the
 // GitHub v3 API, so there are lots of minor incompatibilities with the types
 // defined in the rest of the API. Therefore, several types are duplicated
 // here to account for these differences.
 //
 // GitHub API docs: https://help.github.com/articles/post-receive-hooks
```
dmitshur, line 21: It's not viable to place the "// Deprecated: [...]" comment at the top of a symbol's documentation, since that breaks the convention that an exported identifier named "Thing" should have documentation that begins with "Thing [...]". See https://go.dev/wiki/Deprecated. (It can be at the very bottom or in the middle, just not at the top.)

This applies here and in other places where the deprecation comment is added.

github/repos_hooks.go
```diff
@@ -11,64 +11,27 @@ import (
...
-	Email    *string `json:"email,omitempty"`
-	Name     *string `json:"name,omitempty"`
-	Username *string `json:"username,omitempty"`
-}
-
-func (w WebHookAuthor) String() string {
-	return Stringify(w)
-}
+//
+// NOTE Breaking API change: the `Username` field is now called `Login`.
```
dmitshur, line 33: Does this API change warrant incrementing the major version component? If so, should that happen in this PR, or in another one (before the next release tag is pushed)?

dmitshur (commented):
In general this looks good, see two inline comments. One of them is a question.

github/repos_hooks.go
gmlewis: Yes, we typically label the PR as a "Breaking API Change" and then about every 30-60 days we create a new release and bump the major version if there are any of these changes made.

github/repos_hooks.go
gmlewis: Ah, yes... thank you, Dmitri!

github/repos_hooks.go
dmitshur: I see, thanks.

github/repos_hooks.go
dmitshur: Thanks, marking resolved.

dmitshur (approved):
Thank you.

gmlewis:
Thank you, @dmitshur !
Merging.

PR #2034, opened 2021-08-09
Use x crypto instead of sodium

andersjanmyr:
The example that is using libsodium is hard to cross compile.
Here's an alternative example using `google.com/x/crypto` instead.

Perhaps it should be added as an alternative file instead of change to the current one.


gmlewis:
> Perhaps it should be added as an alternative file instead of change to the current one.

Yes, please add a new example and distinguish between the two with pros/cons in a comment if you wish.

The `libsodium` example was a special request to see how it might be done, so I'm totally fine with adding a new example.
Thank you, @andersjanmyr !

andersjanmyr:
@gmlewis I pushed it as a new directory, but I didn't change the name of the old directory.
Perhaps I should change the name of it to `newreposecretwithlibsodium`?

example/newreposecretwithxcrypto/main.go
```diff
@@ -0,0 +1,176 @@
+// Copyright 2020 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2021 The go-github AUTHORS. All rights reserved.
```

example/newreposecretwithxcrypto/main.go
```diff
@@ -0,0 +1,176 @@
...
+	client := github.NewClient(tc)
+	return ctx, client, nil
+}
+
+// addRepoSecret will add a secret to a GitHub repo for use in GitHub Actions.
+//
+// Finally, the secretName and secretValue will determine the name of the secret added and it's corresponding value.
+//
+// The actual transmission of the secret value to GitHub using the api requires that the secret value is encrypted
+// using the public key of the target repo. This encryption must be done using sodium.
```
gmlewis, line 118: > This encryption must be done using sodium.

Is this still true?

example/newreposecretwithxcrypto/main.go
```diff
@@ -0,0 +1,176 @@
...
+//
+// First, the public key of the repo is retrieved. The public key comes base64
+// encoded, so it must be decoded prior to use in sodiumlib.
```
gmlewis, line 121: > so it must be decoded prior to use in sodiumlib.

How about this?

example/newreposecretwithxcrypto/main.go
```diff
@@ -0,0 +1,176 @@
...
+//
+// Second, the secret value is converted into a slice of bytes.
+//
+// Third, the secret is encrypted with sodium.CryptoBoxSeal using the repo's decoded public key.
```
gmlewis, line 125: > Third, the secret is encrypted with sodium.CryptoBoxSeal using the repo's decoded public key.

...and this?

gmlewis (changes requested):
> Perhaps I should change the name of it to `newreposecretwithlibsodium`?

Sure, that sounds like a good idea.

example/newreposecretwithxcrypto/main.go
andersjanmyr: You are right, I'll go through the comments and clean it up. Sorry about that.
\

example/newreposecretwithxcrypto/go.mod
```diff
@@ -0,0 +1,11 @@
+module newreposecret
```
gmlewis, line 1:
```suggestion
module newreposecretwithxcrypto
```

gmlewis (approved):
Thank you, @andersjanmyr !
LGTM.

If you want to rename the older libsodium example dir in this PR, that's fine with me.
Also, if you want to clean up any of the comment in that older example, that would be awesome too...  (I see that you cleaned up some of the wording here... and maybe some of those cleanups also apply to the older example, but if not, that's OK too... thanks!)

andersjanmyr:
Fixed the comments and modules.

example/newreposecretwithlibsodium/main.go
```diff
@@ -1,13 +1,13 @@
-// Copyright 2020 The go-github AUTHORS. All rights reserved.
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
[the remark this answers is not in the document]
andersjanmyr: 🤦 fixed.

gmlewis (approved):
Thank you, @andersjanmyr !
LGTM.
Merging.

PR #2036, opened 2021-08-11
Allow empty selected_repository_ids for SetSelectedReposForOrgSecret

### LINKED ISSUE #2035: API error when passing empty array as selected_repository_ids in SetSelectedReposForOrgSecret
In [SetSelectedReposForOrgSecret](https://github.com/google/go-github/blob/master/github/actions_secrets.go#L301), when trying to set selected repositories for a organisation secret to empty, it returns an API error:
> 422 Invalid request.
> "selected_repository_ids" wasn't supplied. []

I assume this is because `omitempty` is set for `selected_repository_ids` in the request data. Considering that there should be a valid use case to clear up all selected repositories for an organisation secret, `omitempty` will need to be removed.

xun-guo-anzx:
Fix #2035 - Allow empty selected_repository_ids for SetSelectedReposForOrgSecret

### DIFF
--- a/github/actions_secrets.go
+++ b/github/actions_secrets.go
@@ -302,7 +302,7 @@ func (s *ActionsService) SetSelectedReposForOrgSecret(ctx context.Context, org, 
 	u := fmt.Sprintf("orgs/%v/actions/secrets/%v/repositories", org, name)
 
 	type repoIDs struct {
-		SelectedIDs SelectedRepoIDs `json:"selected_repository_ids,omitempty"`
+		SelectedIDs SelectedRepoIDs `json:"selected_repository_ids"`
 	}
 
 	req, err := s.client.NewRequest("PUT", u, repoIDs{SelectedIDs: ids})


gmlewis:
The GitHub v3 API docs for this endpoint state:

> You can only provide a list of repository ids when the `visibility` is set to `selected`. 

Am I reading this correctly that this endpoint should simply not be called when `visibility` is set to some other value?

I ask because I'm trying to figure out if there is any case when you would want to call this endpoint but _NOT_ send a value for `selected_repository_ids` ?

xun-guo-anzx:
Yes - this endpoint should simply not be called when `visibility` is not `selected`, otherwise the following error will be returned:
> "errors": "You cannot update selected repositories for a secret when the visibility is not set to 'selected'",

My use case for calling this endpoint without sending a value for `selected_repository_ids` is to set selected repository list to empty. However like I mentioned above I'll send over an empty array instead of `nil` or something else.

gmlewis (approved):
Thank you, @xun-guo-anzx !
LGTM.
Merging.

xun-guo-anzx:
Thanks @gmlewis , may I request a release for this change?

gmlewis:
Sure, I'll get you a release.

gmlewis:
Here you go, @xun-guo-anzx - https://github.com/google/go-github/releases/tag/v38.1.0

PR #2042, opened 2021-08-16
Update Cursor comment to reference an existing method

jbowes:
This comment probably became out of date over time, as things were moved and renamed. Update it to the current naming.


gmlewis (approved):
Thank you, @jbowes !
LGTM.
Merging.

PR #2045, opened 2021-08-18
Upgrade crypto and net dependencies

yahavi:
The currently used `golang.org/x/crypto` and `golang.org/x/net` libraries are old and vulnerable.
Let's upgrade them to the latest.


gmlewis (approved):
Thank you, @yahavi !
LGTM.
Merging.

PR #2054, opened 2021-08-19
chore(ci) : cache go-build folders

mmorel-35:
Signed-off-by: Matthieu MOREL <mmorel-35@users.noreply.github.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,10 +21,12 @@ jobs:
           # is very little in the library that is platform specific
           - go-version: 1.x
             platform: windows-latest
+            go-build: ~/AppData/Local/go-build
 
           # only update test coverage stats with the most recent go version on linux
           - go-version: 1.x
             platform: ubuntu-latest
+            go-build: ~/.cache/go-build
             update-coverage: true
     runs-on: ${{ matrix.platform }}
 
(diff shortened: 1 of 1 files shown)


willnorris:
Here's an interesting cross platform approach that uses the output of `got env GOCACHE` and `go env GOMODCACHE`: https://markphelps.me/2019/11/speed-up-your-go-builds-with-actions-cache/

what do you think?  he has it split into two separate cache steps, but I'm not sure that' necessary, since they're both using the hash of `**/go.sum` for the cache key.

.github/workflows/tests.yml
```diff
@@ -43,12 +38,16 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
     - uses: actions/checkout@v2
-
+    
+    # Get values for cache paths to be used in later steps 
+    - id: go-cache
+      run: echo "::set-output name=go-build::$(go env GOCACHE)"
```
willnorris, line 44: might as well use this approach for `go env GOMODCACHE` too?  I mean, it's not likely to change, but you never know and the consistency is nice.

mmorel-35: I agree

willnorris:
sweet, thanks!  I'll let these latest round of tests finish running then merge.

PR #2064, opened 2021-08-29
chore(linter) : activation of more linters

mmorel-35:


example/tokenauth/main.go
```diff
@@ -14,14 +14,15 @@ import (
 	"log"
 	"syscall"
 
+
```
mmorel-35, line 17:
```suggestion
```

mmorel-35:
Hi @willnorris , can you take a look when you find the time :) ?

willnorris:
this is way bigger than I'm going to have bandwidth to review in one go.  I'm also not convinced we event *want* to enable all of these linters, as there are quite a few that are highly subjective.  I'm not sure if that includes any of the ones you've enabled here, I'm just saying that in general.  Can you break this down to just enable one linter at a time so that we can see the impact of enabling that individual linter and discuss them individually?

Also, please don't change the casing in the workflow names "linter" vs "Linter".  I'm pretty sure that will break our readme badges, and I just prefer the lowercase naming anyway.

Issue #2088, opened 2021-09-09
Team does not provide HTMLURL


kplimack:
The teams package provides `URL` which points at the api url, but not `HTMLURL` which is provided via api

when getting a team it will supply

```json
   "html_url": "https://github.com/orgs/github/teams/justice-league"
```

kplimack:
IMO, better off providing it vs having people parse/mutate it on their own.

gmlewis:
Fixed by #2244.
Closing.

PR #2089, opened 2021-09-09
teams.htmlurl

kplimack:
Addresses https://github.com/google/go-github/issues/2088

* https://docs.github.com/en/rest/reference/teams#get-a-team-by-name


gmlewis:
Thanks, @kplimack.

Please take a look at our `CONTRIBUTING.md` guide and take care of the CLA and the `go generate ./...` so that the tests pass.

Issue #2095, opened 2021-09-14
Support for Apps Webhook configuration endpoints


rpelliard:
Hello,

I saw that there is support for configuring [Repository webhooks](https://docs.github.com/en/rest/reference/repos#webhooks) and [Organization webhooks](https://docs.github.com/en/rest/reference/orgs#webhooks), but it looks like [AppsService](https://pkg.go.dev/github.com/google/go-github/v34/github#AppsService) has no support for the endpoints allowing to get / update the webhook configuration of an app: https://docs.github.com/en/rest/reference/apps#webhooks.

Can you confirm that this is not supported ? Is that something that you would consider adding support for ? If yes, I'd be happy to work on a PR for this !

gmlewis:
> Can you confirm that this is not supported ? Is that something that you would consider adding support for ? If yes, I'd be happy to work on a PR for this !

I confirm your assertion. :smile:
Yes, I think that would be perfectly reasonable to add to this repo since it it part of the GitHub v3 API.
One interesting challenge for your PR will be that we currently do not support sending a JWT, so you will need to add a nice and convenient way of doing so.

Thank you, @rpelliard !

rpelliard:
@gmlewis Thanks for confirming :)

About the JWT, would it be reasonable to leave the responsibility of authentication to the underlying transport, like what is used for requests as installations / as users ? e.g. using https://github.com/bradleyfalzon/ghinstallation/blob/master/appsTransport.go#L46
I might be wrong, but it looks like for example, [this existing method](https://pkg.go.dev/github.com/google/go-github/v34/github#AppsService.GetInstallation) queries an endpoint that requires authentication as an app using a JWT

gmlewis:
> About the JWT, would it be reasonable to leave the responsibility of authentication to the underlying transport, like what is used for requests as installations / as users ? e.g. using https://github.com/bradleyfalzon/ghinstallation/blob/master/appsTransport.go#L46
> I might be wrong, but it looks like for example, [this existing method](https://pkg.go.dev/github.com/google/go-github/v34/github#AppsService.GetInstallation) queries an endpoint that requires authentication as an app using a JWT

Ah, yes! Perfect. Thank you... Let's do that.
Please make sure to add a comment in your new methods showing how to authenticate.

PR #2096, opened 2021-09-15
Add support for apps webhook config endpoints

rpelliard:
This adds support for two Apps API endpoints (see #2095)
- [Get a webhook configuration for an app](https://docs.github.com/en/rest/reference/apps#get-a-webhook-configuration-for-an-app)
- [Update a webhook configuration for an app](https://docs.github.com/en/rest/reference/apps#update-a-webhook-configuration-for-an-app)

Both endpoints require authentication as an app (with a JWT), which can be dealt with by the underlying transport.

A few notes:
- I reused / extended the `HookConfig` struct which was defined in `orgs_audit_logs.go`, since it represents the same object.

- The `InsecureSSL` field of the struct is a `*string`. However, GitHub only returns `"0"` or `"1"` (insecure SSL disabled / enabled), so I think it'd be more user-friendly to use a `*bool`, and write some logic so that boolean values are correctly (un)marshaled from `"0"`/`"1"`. I'd be happy to write the code to deal with this, but that would be a breaking change so I wanted to check with you first.

- This is probably out of scope of this PR, but the [`Hook` struct](https://github.com/google/go-github/blob/master/github/repos_hooks.go#L87) used for repository / organization webhooks endpoints is using a `map[string]interface{}` for the `Config` field, but I think it could be using the `HookConfig` struct as well since it represents the same object. Happy to make that change in a separate PR, but this would be a breaking change as well.

### DIFF
--- a/github/apps_hooks.go
+++ b/github/apps_hooks.go
@@ -1,0 +1,43 @@
+package github
+
+import (
+	"context"
+)
+
+// GetHookConfig returns the webhook configuration for a GitHub App.
+// The underlying transport must be authenticated as an app.
+//
+// GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#get-a-webhook-configuration-for-an-app
+func (s *AppsService) GetHookConfig(ctx context.Context) (*HookConfig, *Response, error) {
+	req, err := s.client.NewRequest("GET", "app/hook/config", nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	config := new(HookConfig)
+	resp, err := s.client.Do(ctx, req, &config)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return config, resp, nil
+}
+
+// UpdateHookConfig updates the webhook configuration for a GitHub App.
+// The underlying transport must be authenticated as an app.
+//
+// GitHub API docs: https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#update-a-webhook-configuration-for-an-app
+func (s *AppsService) UpdateHookConfig(ctx context.Context, config *HookConfig) (*HookConfig, *Response, error) {
+	req, err := s.client.NewRequest("PATCH", "app/hook/config", config)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	c := new(HookConfig)
+	resp, err := s.client.Do(ctx, req, c)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return c, resp, nil
+}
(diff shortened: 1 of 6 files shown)


github/orgs_audit_log.go
```diff
@@ -24,6 +24,9 @@ type HookConfig struct {
 	ContentType *string `json:"content_type,omitempty"`
 	InsecureSSL *string `json:"insecure_ssl,omitempty"`
 	URL         *string `json:"url,omitempty"`
+
+	// Secret is returned obfuscated by GitHub, but it can be set for outgoing requests
```
gmlewis, line 28:
```suggestion
	// Secret is returned obfuscated by GitHub, but it can be set for outgoing requests.
```

gmlewis (changes requested):
Thank you, @rpelliard !
One tiny nit to fix, please, but otherwise LGTM.

gmlewis (approved):
LGTM.

rpelliard:
Hello @gmlewis @wesleimp, do you know when we'd be able to merge this ?

gmlewis:
> Hello @gmlewis @wesleimp, do you know when we'd be able to merge this ?

Maybe @Parker77 has time to take a look at this one.

github/apps_hooks.go
```diff
@@ -0,0 +1,43 @@
+package github
```
gmlewis, line 1:
```suggestion
// Copyright 2021 The go-github AUTHORS. All rights reserved.
//
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package github
```

No need for action... I'll add this required header.

github/apps_hooks_test.go
```diff
@@ -0,0 +1,98 @@
+package github
```
gmlewis, line 1:
```suggestion
// Copyright 2021 The go-github AUTHORS. All rights reserved.
//
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package github
```

No need for action... I'll add this required header.

github/doc.go
```diff
@@ -73,6 +73,10 @@ BasicAuthTransport.
 
 GitHub Apps authentication can be provided by the
 https://github.com/bradleyfalzon/ghinstallation package.
+It supports both authentication as an installation, using an installation access token
```
Parker77, line 76:
```suggestion
It supports both authentication as an installation, using an installation access token,
```

Parker77 (approved):
One minor nit, otherwise LGTM.

Parker77:
@googlebot I fixed it.

gmlewis:
> One minor nit, otherwise LGTM.

Thank you, @Parker77 !
Merging.

PR #2100, opened 2021-09-19
Add pagination fields to code scanning

ganeshkumarsv:


github/code-scanning.go
```diff
@@ -115,6 +115,12 @@ type AlertListOptions struct {
 
 	// Return code scanning alerts for a specific branch reference. The ref must be formatted as heads/<branch name>.
 	Ref string `url:"ref,omitempty"`
+
+	// Returns the results based on Page number. Page number is set to 1 by default
+	Page int `url:"page,omitempty"`
+
+	// Results per page. Maximum is 100 results page
+	PerPage int `url:"per_page,omitempty"`
```
gmlewis, line 123:
```suggestion
	ListOptions
```

ganeshkumarsv: @gmlewis I'm not understanding properly why we are removing the 2 entries and replacing it with `ListOptions`. Can you please explain?

gmlewis: > @gmlewis I'm not understanding properly why we are removing the 2 entries and replacing it with `ListOptions`. Can you please explain?

In Go, this is called "type embedding" or "struct embedding".  Since `ListOptions` provides all we need here, we aren't duplicating the code every place we need these fields.  Take a look through our repo, and this is how we embed these list options in other options.

ganeshkumarsv: Thanks for the detailed explanation! @gmlewis

gmlewis (approved):
Thank you, @ganeshkumarsv !
LGTM.
Merging.

ganeshkumarsv:
@gmlewis Can we have a new release version?

gmlewis:
@ganeshkumarsv - https://github.com/google/go-github/releases/tag/v39.1.0 now contains this change.

ganeshkumarsv:
Thank you! 🙇‍♂️

PR #2103, opened 2021-09-23
Add support for workflow_job event's `labels` field

mumoshu:
This adds support for new `workflow_job` webhook events' `labels` field announced at https://github.blog/changelog/2021-09-20-github-actions-ephemeral-self-hosted-runners-new-webhooks-for-auto-scaling/.

I've already using go-github with this patch for weeks while [participating in the alpha test](https://github.com/actions-runner-controller/actions-runner-controller/issues/697) of the new webhook event. So I'm pretty confident this works. But I'd do anything I can do if necessary!

### DIFF
--- a/github/actions_workflow_jobs.go
+++ b/github/actions_workflow_jobs.go
@@ -38,7 +38,7 @@ type WorkflowJob struct {
 	Name        *string     `json:"name,omitempty"`
 	Steps       []*TaskStep `json:"steps,omitempty"`
 	CheckRunURL *string     `json:"check_run_url,omitempty"`
-	// Labels is runner labels from the `runs-on:` key from a GitHub Actions workflow
+	// Labels represents runner labels from the `runs-on:` key from a GitHub Actions workflow.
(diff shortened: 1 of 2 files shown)


github/actions_workflow_jobs.go
```diff
@@ -38,6 +38,8 @@ type WorkflowJob struct {
 	Name        *string     `json:"name,omitempty"`
 	Steps       []*TaskStep `json:"steps,omitempty"`
 	CheckRunURL *string     `json:"check_run_url,omitempty"`
+	// Labels is runner labels from the `runs-on:` key from a GitHub Actions workflow
```
gmlewis, line 41:
```suggestion
	// Labels represents runner labels from the `runs-on:` key from a GitHub Actions workflow.
```

(This is a minor nit tweak, so I'll go ahead and commit it... nothing necessary to do on your end.)

gmlewis (approved):
Thank you, @mumoshu !
LGTM.
Merging.

PR #2104, opened 2021-09-26
Some fix linters issues

alexander-melentyev:


gmlewis (approved):
Thank you, @alexander-melentyev !
LGTM.
Merging.

PR #2108, opened 2021-09-28
Bump CI version of Go to test versions 1.17.x and 1.16.x

### LINKED ISSUE #2106: CI uses an older version of Go
Hi, why does `go.mod` already use Go **1.16**, but `tests.yml` still uses **1.15.x**? Should we update the CI config?

gmlewis:
Fixes #2106.


Parker77 (approved):
LGTM!

PR #2109, opened 2021-09-28
Update GitHub Pages API to add new response parameters

### LINKED ISSUE #2107: GitHub Pages Response Out of Date
The GitHub Pages API response appears to be a bit out of date. There are new additional fields for GitHub Pages that aren't included in the struct.  These include information about the visibility of the page, certificate issuance info and whether or not HTTPS is enforced.  

The additional properties required are visible in the docs [here](https://docs.github.com/en/rest/reference/repos#get-a-github-pages-site) but specifically are as follows:

```json
  "public": true,
  "https_certificate": {
    "state": "approved",
    "description": "Certificate is approved",
    "domains": [
      "developer.github.com"
    ],
    "expires_at": "2021-05-22"
  },
  "https_enforced": true
```

I'm planning on working on this over the next few days and hope to have a PR soon.

rreichel3:
This PR updates the GitHub Pages API structs to support the latest documented response from GitHub.  

This fixes #2107 

Its based upon the documentation on GitHub's public docs [here](https://docs.github.com/en/rest/reference/repos#get-a-github-pages-site) and enables support for viewing a GitHub Pages' visibility and HTTPS configuration.

### DIFF
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -51,7 +51,8 @@ type PagesHTTPSCertificate struct {
 	State       *string  `json:"state,omitempty"`
 	Description *string  `json:"description,omitempty"`
 	Domains     []string `json:"domains,omitempty"`
-	ExpiresAt   *string  `json:"expires_at,omitempty"`
+	// GitHub's API doesn't return a standard Timestamp, rather it returns a YYYY-MM-DD string
+	ExpiresAt *string `json:"expires_at,omitempty"`
 }
 
 // createPagesRequest is a subset of Pages and is used internally
--- a/github/repos_pages.go
+++ b/github/repos_pages.go
@@ -12,12 +12,15 @@ import (
 
 // Pages represents a GitHub Pages site configuration.
 type Pages struct {
-	URL       *string      `json:"url,omitempty"`
-	Status    *string      `json:"status,omitempty"`
-	CNAME     *string      `json:"cname,omitempty"`
-	Custom404 *bool        `json:"custom_404,omitempty"`
-	HTMLURL   *string      `json:"html_url,omitempty"`
-	Source    *PagesSource `json:"source,omitempty"`
+	URL              *string                `json:"url,omitempty"`
+	Status           *string                `json:"status,omitempty"`
+	CNAME            *string                `json:"cname,omitempty"`
+	Custom404        *bool                  `json:"custom_404,omitempty"`
+	HTMLURL          *string                `json:"html_url,omitempty"`
+	Source           *PagesSource           `json:"source,omitempty"`
+	Public           *bool                  `json:"public,omitempty"`
+	HttpsCertificate *PagesHttpsCertificate `json:"https_certificate,omitempty"`
+	HttpsEnforced    *bool                  `json:"https_enforced,omitempty"`
 }
 
 // PagesSource represents a GitHub page's source.
@@ -41,6 +44,14 @@ type PagesBuild struct {
 	Duration  *int        `json:"duration,omitempty"`
 	CreatedAt *Timestamp  `json:"created_at,omitempty"`
 	UpdatedAt *Timestamp  `json:"updated_at,omitempty"`
+}
+
+// PagesBuild represents the HTTPS Certificate information for a GitHub Pages site.
+type PagesHttpsCertificate struct {
+	State       *string   `json:"state,omitempty"`
+	Description *string   `json:"description,omitempty"`
+	Domains     *[]*string `json:"domains,omitempty"`
+	ExpiresAt   *string   `json:"expires_at,omitempty"`
 }
 
 // createPagesRequest is a subset of Pages and is used internally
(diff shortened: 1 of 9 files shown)


github/repos_pages.go
```diff
@@ -12,12 +12,15 @@ import (
...
-	HTMLURL   *string      `json:"html_url,omitempty"`
-	Source    *PagesSource `json:"source,omitempty"`
+	URL              *string                `json:"url,omitempty"`
+	Status           *string                `json:"status,omitempty"`
+	CNAME            *string                `json:"cname,omitempty"`
+	Custom404        *bool                  `json:"custom_404,omitempty"`
+	HTMLURL          *string                `json:"html_url,omitempty"`
+	Source           *PagesSource           `json:"source,omitempty"`
+	Public           *bool                  `json:"public,omitempty"`
+	HttpsCertificate *PagesHttpsCertificate `json:"https_certificate,omitempty"`
```
gmlewis, line 22:
```suggestion
	HTTPSCertificate *PagesHTTPSCertificate `json:"https_certificate,omitempty"`
```

github/repos_pages.go
```diff
@@ -12,12 +12,15 @@ import (
...
+	HttpsEnforced    *bool                  `json:"https_enforced,omitempty"`
```
gmlewis, line 23:
```suggestion
	HTTPSEnforced    *bool                  `json:"https_enforced,omitempty"`
```

github/repos_pages.go
```diff
@@ -43,6 +46,14 @@ type PagesBuild struct {
 	UpdatedAt *Timestamp  `json:"updated_at,omitempty"`
 }
 
+// PagesBuild represents the HTTPS Certificate information for a GitHub Pages site.
```
gmlewis, line 49:
```suggestion
// PagesHTTPSCertificate represents the HTTPS Certificate information for a GitHub Pages site.
```

github/repos_pages.go
```diff
@@ -43,6 +46,14 @@ type PagesBuild struct {
...
+type PagesHttpsCertificate struct {
```
gmlewis, line 50:
```suggestion
type PagesHTTPSCertificate struct {
```

github/repos_pages.go
```diff
@@ -43,6 +46,14 @@ type PagesBuild struct {
...
+	State       *string    `json:"state,omitempty"`
+	Description *string    `json:"description,omitempty"`
+	Domains     *[]*string `json:"domains,omitempty"`
```
gmlewis, line 53:
```suggestion
	Domains     []string `json:"domains,omitempty"`
```

github/repos_pages.go
```diff
@@ -43,6 +46,14 @@ type PagesBuild struct {
...
+	ExpiresAt   *string    `json:"expires_at,omitempty"`
```
gmlewis, line 54:
```suggestion
	ExpiresAt   *Timestamp    `json:"expires_at,omitempty"`
```

gmlewis (changes requested):
Thank you, @rreichel3!
Just a few changes, please, and then we should be ready for a second LGTM before merging.

github/repos_pages.go
rreichel3: This doesn't seem to work properly - The GitHub API is only returning a `YYYY-MM-DD` date and the `Timestamp` deserializer expects a full timestamp.  I looked at other places this library uses `Timestamp` and all of them appear to have the full date. Its a bit of a quirk with this API it seems. 

If we need it to be a timestamp we'll likely need to modify the Timestamp deserializer (which can be done, but seems a bit heavy-handed).  Happy to do whatever you think is best @gmlewis

rreichel3: Thanks! Dropped a comment on the Struct

github/repos_pages.go
```diff
@@ -43,6 +46,15 @@ type PagesBuild struct {
 	UpdatedAt *Timestamp  `json:"updated_at,omitempty"`
 }
 
+// PagesHTTPSCertificate represents the HTTPS Certificate information for a GitHub Pages site.
+type PagesHTTPSCertificate struct {
+	State       *string  `json:"state,omitempty"`
+	Description *string  `json:"description,omitempty"`
+	Domains     []string `json:"domains,omitempty"`
+	// GitHub's API doesn't return a standard Timestamp, rather it returns a YYYY-MM-DD string
```
gmlewis, line 54:
```suggestion
	// GitHub's API doesn't return a standard Timestamp, rather it returns a YYYY-MM-DD string.
```

gmlewis (approved):
Thank you, @rreichel3 !
LGTM.

rreichel3:
👋 @wesleimp Just wanted to bump this!

cpanato (approved):
lgtm
thanks for this addition

gmlewis:
Thank you, @cpanato !
Merging.

Issue #2110, opened 2021-09-28
ProjectOptions has wrong field & Project struct miss Organization related fields


DariuszPorowski:
According to official GitHub API docs for [Update a project](https://docs.github.com/en/rest/reference/projects#update-a-project), one of the body fields is **private**, but current implementation in the go-github is **public**. [projects.go#L86](https://github.com/google/go-github/blob/eefa002529986d0f766419e9e03b6e6371e57bd7/github/projects.go#L86)

Moreover, the current struct that returns the Project is incomplete. During [Get a project](https://docs.github.com/en/rest/reference/projects#get-a-project) call on the Organization Project, the response also returns two fields: **organization_permission** and **private** (related to the issue described above). Example response:
`
{
    "owner_url": "https://api.github.com/orgs/ORG",
    "url": "https://api.github.com/projects/1234",
    "html_url": "https://github.com/orgs/ORG/projects/19",
    "columns_url": "https://api.github.com/projects/1234/columns",
    "id": 1234,
    "node_id": "PRO_kwDOBXDcOc4AzHJH",
    "name": "test",
    "body": "",
    "number": 19,
    "state": "open",
    "creator": {
        "login": "USER",
        "id": 5678,
        "node_id": "MDQ6VXNlcjM0MzE4MTM=",
        "avatar_url": "",
        "gravatar_id": "",
        "url": "https://api.github.com/users/USER",
        "html_url": "https://github.com/USER",
        "followers_url": "https://api.github.com/users/USER/followers",
        "following_url": "https://api.github.com/users/USER/following{/other_user}",
        "gists_url": "https://api.github.com/users/USER/gists{/gist_id}",
        "starred_url": "https://api.github.com/users/USER/starred{/owner}{/repo}",
        "subscriptions_url": "https://api.github.com/users/USER/subscriptions",
        "organizations_url": "https://api.github.com/users/USER/orgs",
        "repos_url": "https://api.github.com/users/USER/repos",
        "events_url": "https://api.github.com/users/USER/events{/privacy}",
        "received_events_url": "https://api.github.com/users/USER/received_events",
        "type": "User",
        "site_admin": false
    },
    "created_at": "2021-09-28T16:52:01Z",
    "updated_at": "2021-09-28T17:10:19Z",
    "organization_permission": "write",
    "private": false
}
`
I'll try to fix it and do PR.

PR #2111, opened 2021-09-28
ProjectOptions (public to private) & Project struct (org related fields) fixes 

DariuszPorowski:
This fixes https://github.com/google/go-github/issues/2110

- fixed ProjectOptions: Public -> Private (according to API doc: https://docs.github.com/en/rest/reference/projects#update-a-project)
- added **OrganizationPermission** and **Private** fields to Project struct, which presents during response on Organization Project.

I've done UATs as well, and all API calls looked good:
- Update Org Project with new private field
- Get Org Project (OrganizationPermission and Private presents as expected)
- Get Repo Project (OrganizationPermission and Private not presents as expected)
- Get User Project (OrganizationPermission and Private not presents as expected)

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -10236,12 +10236,28 @@ func (p *Project) GetNumber() int {
 	return *p.Number
 }
 
+// GetOrganizationPermission returns the OrganizationPermission field if it's non-nil, zero value otherwise.
+func (p *Project) GetOrganizationPermission() string {
+	if p == nil || p.OrganizationPermission == nil {
+		return ""
+	}
+	return *p.OrganizationPermission
+}
+
 // GetOwnerURL returns the OwnerURL field if it's non-nil, zero value otherwise.
 func (p *Project) GetOwnerURL() string {
 	if p == nil || p.OwnerURL == nil {
 		return ""
 	}
 	return *p.OwnerURL
+}
+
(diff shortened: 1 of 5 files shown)


DariuszPorowski:
@googlebot I fixed it.

gmlewis (approved):
Thank you, @DariuszPorowski !
LGTM.

This is technically a Breaking API change... however, I'm tempted to declare this one as non-breaking simply because what we are changing was already broken... so in theory, any attempts to use the prior version would have failed.

Thoughts?  Is it OK to _NOT_ declare this PR as a breaking APi change (and therefore not have to bump the major version number of this repo at the next release due to this PR alone)?

@willnorris - what is your opinion?

willnorris:
Even though this was never working as intended, it's a breaking change in the Go API in that if anyone *was* calling it, their code would no longer compile now.  And since it's been in the wild [for 3 years](https://github.com/google/go-github/commit/d8bee7566740e39c1c0e2c3990e1e5f348458852), I feel like it would be more responsible to treat it as a breaking change.

DariuszPorowski:
@gmlewis I agree with @willnorris
Even if it is a change to a valid GitHub API, it changes the entire usage of Go API.

cpanato (approved):
lgtm
we need to document that in the release notes

gmlewis:
Thank you, @cpanato - merging.

PR #2121, opened 2021-10-08
Bump scrape apps to google-go v39

chmouel:
While it's possible, this cause confusion when we want to use the "scrape" library with githubv39 and having to import v32.

Bumping it to latest,


gmlewis:
> > At one point, @willnorris asked me to keep the versions separate, so I defer to him on this one.
> 
> I think I probably just meant to keep the the scrape package in its own go module that is versioned independently. Bumping the version of go-github/github like this should always be fine.

Thank you, @willnorris !
LGTM.
Merging.

PR #2127, opened 2021-10-18
runnerID and runnerGroupID are int64, not string

sfunkhouser:
When running this before the change, results that included runnerID or runnerGroupID could not be parsed. This PR changes the type for both from `string` -> `int64`

### DIFF
--- a/github/orgs_audit_log_test.go
+++ b/github/orgs_audit_log_test.go
@@ -268,9 +268,9 @@ func TestAuditEntry_Marshal(t *testing.T) {
 		"repo": "r",
 		"repository": "repo",
 		"repository_public": false,
-		"runner_group_id": "rgid",
+		"runner_group_id": 1,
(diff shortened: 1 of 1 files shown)


sfunkhouser:
> Thank you, @sfunkhouser - can you find any official GitHub docs that show these values as integers? I'm curious because I thought the original implementation came from an example where they were strings... but now I'm having troubles finding the examples.
> 
> Also, if you could please fix the CLA and the unit tests, we can move forward with this PR.

Looking for the docs, I've only found the runner docs, which reference them as integers (https://docs.github.com/en/enterprise-server@2.22/rest/reference/enterprise-admin#add-a-self-hosted-runner-to-a-group-for-an-enterprise) and when I ran through the audit logs I got an unmarshal error but let me see if I can find some more concrete documentation on this. I'll follow-up tomorrow.

Issue #1848, opened 2021-04-05
Audit Log API Returns Non-Standard Link Header


lindluni:
For tracking purposes, I've requested information on how they intend to handle this:

While working with the beta release of the `google/go-github` SDK I noted that pagination was not working for this API. After digging into the SDK I discovered the reason for this was that the audit-log API does not return traditional page parameters in the link array of the header. For example this query:

```
curl -v -H "Authorization: token $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" https://api.github.com/orgs/lindluni-enterprise/audit-log > /dev/null
```

Returns this link header:

```
link: <https://api.github.com/organizations/80307827/audit-log?after=MS42MTUzMzYzNzg4MzNlKzEyfFlBdkVhclBZR21reGtQMG5uRng4bVE%3D&before=>; rel="next"
```

Which unlike the traditional link headers such as this:

```
link: <https://api.github.com/organizations/<id trunc>/repos?page=2>; rel="next", <https://api.github.com/organizations/<id trunc>/repos?page=124>; rel="last"
```

The audit log API does not return a page, it returns the traditional graphql `After` and `Before` fields. This is despite the fact the API documentation states that the API accepts a page parameter. If I do in fact include a page parameter, such as this query:

```
 curl -H "Authorization: token $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" "https://api.github.com/orgs/<org trunc>/audit-log?page=1"
```

The resultant link header does include the page property:

```
https://api.github.com/organizations/<org trunc>/audit-log?page=1&after=MS42MTcxNjQ3MzUxZSsxMnxNR1h4NFpmTUh6aEFyZGFQVWlWUzhn&before=
```
However no matter what page I pass in it returns the same dataset, and the page property returned is whatever I pass in.

gmlewis:
Relates to: #1825.

gmlewis:
As I mentioned in #1825, this appears to be a GitHub bug and GitHub tech support needs to be contacted and asked why this is happening.  Please report back here once you find out the answer, but in general, this repo does not track GitHub API bugs.

lindluni:
I'll track the information in #1825

Yiyiyimu:
@lindluni FYI, I've submitted a ticket to Github Support on this problem, and I was told the bug is confirmed but there would be no ETA to fix it.

PR #2154, opened 2021-10-27
Parse Before/After Links into Response

dghubble:
* Some Github APIs use `before`/`after` for pagination, rather than `page` or `cursor` pagination. Parse these response values into the github `Response`
* For example, the Organizations GetAuditLog (Enterprise Cloud) https://docs.github.com/en/rest/reference/orgs#get-the-audit-log-for-an-organization

With this change, pagination can be performed

```go
auditParams := &github.GetAuditLogOptions{}
for {
  entries, resp, err := client.Organizations.GetAuditLog(ctx, "myorg", auditParams)
  if err != nil {
    return err
  }

  if resp.After == "" {
    break
  }
  auditParams.After = resp.After
}
```

Rel: https://github.com/google/go-github/issues/1848

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -462,7 +462,7 @@ type Response struct {
 	// Set ListCursorOptions.Cursor to this value when calling the endpoint again.
 	Cursor string
 
-	// For APIs that support before/after pagniation, such as OrganizationsService.AuditLog.
+	// For APIs that support before/after pagination, such as OrganizationsService.AuditLog.
 	Before string
 	After  string
 
--- a/github/github.go
+++ b/github/github.go
@@ -462,6 +462,10 @@ type Response struct {
 	// Set ListCursorOptions.Cursor to this value when calling the endpoint again.
 	Cursor string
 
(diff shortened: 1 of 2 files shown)


dghubble:
Maybe another reviewer has time / interest?

gmlewis:
Perhaps @Parker77 has time?

github/github.go
```diff
@@ -462,6 +462,10 @@ type Response struct {
 	// Set ListCursorOptions.Cursor to this value when calling the endpoint again.
 	Cursor string
 
+	// For APIs that support before/after pagniation, such as OrganizationsService.AuditLog.
```
Parker77, line 465:
```suggestion
	// For APIs that support before/after pagination, such as OrganizationsService.AuditLog.
```

Parker77 (approved):
One minor change, otherwise LGTM!

gmlewis:
Thank you, @Parker77 !
Merging after tests pass.

PR #2125, opened 2021-10-15
Remove REST API preview headers that have been promoted out of preview

andygrunwald:
## What type of PR is this?

cleanup

## What this PR does / why we need it:

On October 14, 2021, GitHub has promoted 23 preview headers. See [October 14, 2021: REST API preview promotions @ GitHub Changelog blog](https://github.blog/changelog/2021-10-14-rest-api-preview-promotions/).

This PR will remove 25 (!) preview headers from the codebase, but we keep the two headers that are still in preview by GitHub (mentioned in the blog post):

* `application/vnd.github.corsair-preview+json`
* `application/vnd.github.scarlet-witch-preview+json`

Removed headers:

* `application/vnd.github.v3.star+json`
* `application/vnd.github.wyandotte-preview+json`
* `application/vnd.github.ant-man-preview+json`
* `application/vnd.github.flash-preview+json`
* `application/vnd.github.squirrel-girl-preview`
* `application/vnd.github.mockingbird-preview+json`
* `application/vnd.github.inertia-preview+json`
* `application/vnd.github.cloak-preview+json`
* `application/vnd.github.giant-sentry-fist-preview+json`
* `application/vnd.github.mercy-preview+json`
* `application/vnd.github.luke-cage-preview+json`
* `application/vnd.github.eye-scream-preview`
* `application/vnd.github.zzzax-preview+json`
* `application/vnd.github.starfox-preview+json`
* `application/vnd.github.sombra-preview+json`
* `application/vnd.github.switcheroo-preview+json`
* `application/vnd.github.dorian-preview+json`
* `application/vnd.github.london-preview+json`
* `application/vnd.github.lydian-preview+json`
* `application/vnd.github.groot-preview+json`
* `application/vnd.github.surtur-preview+json`
* `application/vnd.github.baptiste-preview+json`
* `application/vnd.github.comfort-fade-preview+json`
* `application/vnd.github.doctor-strange-preview+json`
* `application/vnd.github.nebula-preview+json`

Why there is a difference of 2 (this PR removes 25 previews, the blog post is mentioning 23)?

Sadly there is no complete list provided in the blogpost of previews that have been removed.
My assumption is that this codebase contained preview headers that have been graduated before.

### Size of the Pull Request

This Pull Request is very big and (IMO) hard to review.
Every commit is atomic and removes only one preview.
If you wish to receive 25 Pull Requests (one PR for each preview header), please let me know.

### Other obsolete code

By removing the preview headers, I found the following code to be also obsolete:

[...]

PR #2188, opened 2021-10-31
Comment to keep API preview headers around

andygrunwald:
## What type of PR is this?

documentation

## What this PR does / why we need it:

API preview headers (media types) are used to enable new features in GitHub API.
These media types will be added to the API request as headers and used to enable particular features on GitHub API that are still in preview.
After some time, specific media types will be promoted (to a "stable" state).
From then on, the preview headers are not required anymore to activate the additional feature on GitHub.com's API. However, this API header might still be needed for users to run a GitHub Enterprise Server on-premise.
It's not uncommon for GitHub Enterprise Server customers to run older versions which would probably rely on the preview headers for some time.
While the header promotion is going out for GitHub.com, it may be some time before it even arrives in GitHub Enterprise Server.
We keep those preview headers around to avoid breaking older GitHub Enterprise Server versions. Additionally, an non-functional (preview) header doesn't create any side effects on GitHub Cloud version.

See https://github.com/google/go-github/pull/2125 for full context.

## Special notes for your reviewer:

This Pull Request was triggered out of a discussion from https://github.com/google/go-github/pull/2125

### Explicit comment

Personally, I prefer and see value in extensive comments that tell the story and the full context next to the code (like in this PR).
I can understand when this interrupts the code reading flow for others.
I do think that the added advantage (by the additional context) is bigger.

Please let me know what you think about it and if any change is required.

## Additional documentation, usage docs, etc.:

- https://github.com/google/go-github/pull/2125

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -52,7 +52,7 @@ const (
 	// and used to enable particular features on GitHub API that are still in preview.
 	// After some time, specific media types will be promoted (to a "stable" state).
 	// From then on, the preview headers are not required anymore to activate the additional
-	// feature on GitHub.com's API. However, this API header might still be needed for users to run a GitHub Enterprise Server
+	// feature on GitHub.com's API. However, this API header might still be needed for users
 	// to run a GitHub Enterprise Server on-premise.
 	// It's not uncommon for GitHub Enterprise Server customers to run older versions which
 	// would probably rely on the preview headers for some time.
--- a/github/github.go
+++ b/github/github.go
@@ -48,6 +48,21 @@ const (
 	mediaTypeIssueImportAPI    = "application/vnd.github.golden-comet-preview+json"
 
 	// Media Type values to access preview APIs
+	// These media types will be added to the API request as headers
+	// and used to enable particular features on GitHub API that are still in preview.
+	// After some time, specific media types will be promoted (to a "stable" state).
+	// From then on, the preview headers are not required anymore to activate the additional
+	// feature on GitHub.com's API. However, this API header might still be needed for users to run a GitHub Enterprise Server
+	// on-premise.
+	// It's not uncommon for GitHub Enterprise Server customers to run older versions which
+	// would probably rely on the preview headers for some time.
+	// While the header promotion is going out for GitHub.com, it may be some time before it
+	// even arrives in GitHub Enterprise Server.
+	// We keep those preview headers around to avoid breaking older GitHub Enterprise Server
+	// versions. Additionally, an non-functional (preview) header doesn't create any side effects
+	// on GitHub Cloud version.
+	//
+	// See https://github.com/google/go-github/pull/2125 for full context.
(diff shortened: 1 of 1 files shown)


github/github.go
```diff
@@ -48,6 +48,21 @@ const (
...
+	// After some time, specific media types will be promoted (to a "stable" state).
+	// From then on, the preview headers are not required anymore to activate the additional
+	// feature on GitHub.com's API. However, this API header might still be needed for users to run a GitHub Enterprise Server
+	// on-premise.
+	// It's not uncommon for GitHub Enterprise Server customers to run older versions which
+	// would probably rely on the preview headers for some time.
+	// While the header promotion is going out for GitHub.com, it may be some time before it
+	// even arrives in GitHub Enterprise Server.
+	// We keep those preview headers around to avoid breaking older GitHub Enterprise Server
+	// versions. Additionally, an non-functional (preview) header doesn't create any side effects
```
gmlewis, line 62:
```suggestion
	// versions. Additionally, non-functional (preview) headers don't create any side effects
```

github/github.go
```diff
@@ -48,6 +48,21 @@ const (
 	mediaTypeIssueImportAPI    = "application/vnd.github.golden-comet-preview+json"
 
 	// Media Type values to access preview APIs
+	// These media types will be added to the API request as headers
+	// and used to enable particular features on GitHub API that are still in preview.
+	// After some time, specific media types will be promoted (to a "stable" state).
+	// From then on, the preview headers are not required anymore to activate the additional
+	// feature on GitHub.com's API. However, this API header might still be needed for users to run a GitHub Enterprise Server
+	// on-premise.
```
gmlewis, line 56:
```suggestion
	// to run a GitHub Enterprise Server on-premise.
```

github/github.go
```diff
@@ -48,6 +48,21 @@ const (
 	mediaTypeIssueImportAPI    = "application/vnd.github.golden-comet-preview+json"
 
 	// Media Type values to access preview APIs
+	// These media types will be added to the API request as headers
+	// and used to enable particular features on GitHub API that are still in preview.
+	// After some time, specific media types will be promoted (to a "stable" state).
+	// From then on, the preview headers are not required anymore to activate the additional
+	// feature on GitHub.com's API. However, this API header might still be needed for users to run a GitHub Enterprise Server
```
gmlewis, line 55:
```suggestion
	// feature on GitHub.com's API. However, this API header might still be needed for users
```

gmlewis (changes requested):
Thank you, @andygrunwald !
Just a couple tiny nits, then LGTM and can be merged.

andygrunwald:
Thanks for the quick review @gmlewis.
I added your changes.

gmlewis (approved):
Thank you, @andygrunwald !
LGTM.
Merging.

PR #2206, opened 2021-11-18
Add missing attribute job_runs in WorkflowRunBill

marcelocarlos:
### What type of PR is this?

Update `WorkflowRunBill` to include a missing attribute

### What this PR does / why we need it

Brings the implementation inline with the expected API response ([docs here](https://docs.github.com/en/rest/reference/actions#get-workflow-run-usage)).

`WorkflowRunBill` current implementation is missing the JobRuns (`job_runs`) attribute. This is important when calculating the actual cost and time spent in a given workflow run as the `total_ms` doesn't include the rounding ups or multipliers that GitHub applied to each job run.


gmlewis (approved):
Thank you, @marcelocarlos !
LGTM.
Merging.

PR #2210, opened 2021-11-26
Add CodeScanningService.ListAnalysesForRepo and CodeScanningService.GetAnalysis

yogur:
Add ListAnalysesForRepo and GetAnalysis to CodeScanningService, including unit tests.

### DIFF
--- a/github/code-scanning.go
+++ b/github/code-scanning.go
@@ -130,10 +130,10 @@ type AnalysesListOptions struct {
 	ListOptions
 }
 
-// Analysis represents an individual GitHub Code Scanning Analysis on a single repository.
+// ScanningAnalysis represents an individual GitHub Code Scanning ScanningAnalysis on a single repository.
 //
 // GitHub API docs: https://docs.github.com/en/rest/reference/code-scanning#list-code-scanning-analyses-for-a-repository
-type Analysis struct {
+type ScanningAnalysis struct {
 	ID           *int64     `json:"id,omitempty"`
 	Ref          *string    `json:"ref,omitempty"`
 	CommitSHA    *string    `json:"commit_sha,omitempty"`
@@ -255,7 +255,7 @@ func (s *CodeScanningService) UploadSarif(ctx context.Context, owner, repo strin
 // GitHub Apps must have the security_events read permission to use this endpoint.
 //
 // GitHub API docs: https://docs.github.com/en/rest/reference/code-scanning#list-code-scanning-analyses-for-a-repository
-func (s *CodeScanningService) ListAnalysesForRepo(ctx context.Context, owner, repo string, opts *AnalysesListOptions) ([]*Analysis, *Response, error) {
+func (s *CodeScanningService) ListAnalysesForRepo(ctx context.Context, owner, repo string, opts *AnalysesListOptions) ([]*ScanningAnalysis, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/code-scanning/analyses", owner, repo)
 	u, err := addOptions(u, opts)
 	if err != nil {
@@ -267,7 +267,7 @@ func (s *CodeScanningService) ListAnalysesForRepo(ctx context.Context, owner, re
 		return nil, nil, err
 	}
 
-	var analyses []*Analysis
+	var analyses []*ScanningAnalysis
 	resp, err := s.client.Do(ctx, req, &analyses)
 	if err != nil {
 		return nil, resp, err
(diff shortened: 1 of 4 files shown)


yogur:
I just submitted a CLA.

yogur:
I noticed that most of the unit tests are prefixed with "TestActionsService", shouldn't these unit tests be prefixed with "TestCodeScanningService" instead?

gmlewis:
> I noticed that most of the unit tests are prefixed with "TestActionsService", shouldn't these unit tests be prefixed with "TestCodeScanningService" instead?

Good catch!  Yes indeed, these should be changed.  Would you mind addressing these changes in this PR?
Thank you, @yogur !

@googlebot rescan.

github/code-scanning.go
```diff
@@ -119,6 +119,38 @@ type AlertListOptions struct {
 	ListOptions
 }
 
+// AnalysesListOptions specifies optional parameters to the CodeScanningService.ListAnalysesForRepo method.
+type AnalysesListOptions struct {
+	// Return code scanning analyses belonging to the same SARIF upload.
+	SarifID string `url:"sarif_id,omitempty"`
```
gmlewis, line 125:
```suggestion
	SarifID *string `url:"sarif_id,omitempty"`
```

github/code-scanning.go
```diff
@@ -119,6 +119,38 @@ type AlertListOptions struct {
...
+
+	// Return code scanning analyses for a specific branch reference. The ref can be formatted as refs/heads/<branch name> or simply <branch name>.
+	Ref string `url:"ref,omitempty"`
```
gmlewis, line 128:
```suggestion
	Ref *string `url:"ref,omitempty"`
```

github/code-scanning.go
```diff
@@ -119,6 +119,38 @@ type AlertListOptions struct {
...
+
+	ListOptions
+}
+
+// Analysis represents an individual GitHub Code Scanning Analysis on a single repository.
+//
+// GitHub API docs: https://docs.github.com/en/rest/reference/code-scanning#list-code-scanning-analyses-for-a-repository
+type Analysis struct {
```
gmlewis, line 136: Can this be merged with the existing `SarifAnalysis` struct?
If not, can we give it a more specific name?

github/code-scanning.go
```diff
@@ -119,6 +119,38 @@ type AlertListOptions struct {
...
+	ID           *int64     `json:"id,omitempty"`
+	Ref          *string    `json:"ref,omitempty"`
+	CommitSha    *string    `json:"commit_sha,omitempty"`
```
gmlewis, line 139:
```suggestion
	CommitSHA    *string    `json:"commit_sha,omitempty"`
```

github/code-scanning.go
```diff
@@ -215,3 +247,56 @@ func (s *CodeScanningService) UploadSarif(ctx context.Context, owner, repo strin
 
 	return sarifID, resp, nil
 }
+
+// ListAnalysesForRepo lists code scanning analyses for a repository.
+//
+// Lists the details of all code scanning analyses for a repository, starting with the most recent.
+// You must use an access token with the security_events scope to use this endpoint. GitHub Apps must have the security_events
+// read permission to use this endpoint.
```
gmlewis, line 255:
```suggestion
// GitHub Apps must have the security_events read permission to use this endpoint.
```

gmlewis (changes requested):
Thank you, @yogur !
I have a couple questions about this. See below.

yogur:
> > I noticed that most of the unit tests are prefixed with "TestActionsService", shouldn't these unit tests be prefixed with "TestCodeScanningService" instead?
> 
> Good catch! Yes indeed, these should be changed. Would you mind addressing these changes in this PR? Thank you, @yogur !
> 
> @googlebot rescan.

@gmlewis No problem! I committed the name change.

github/code-scanning.go
yogur: It can't because `SarifAnalysis` is used for POST body params while `Analysis` is used to unmarshal json response of listing code scanning analysis instances. 

I thought about renaming to to `ScanningAnalysis` or `CodeScanningAnalysis` but I felt that was redundant with the service name. `CodeScanningService`. What are your thoughts on this?

gmlewis: The problem is that this name is global throughout the `package github` namespace. So even though it is _used_ within the `CodeScanningService`, it is really `github.Analysis` which is pretty general, so if you don't mind switching to `ScanningAnalysis`, I think that would be preferable.  Thank you!

yogur: Oh yes you're right. I updated it. Thanks for the review!

gmlewis (approved):
Thank you, @yogur !
LGTM.

(Please note that ALL other contributors to this repo are welcome to provide the second PR review/comment/approval that we need for merging and that we are not waiting for any particular reviewer unless otherwise noted.)

cpanato (approved):
LGTM
thanks!

gmlewis:
Thank you, @cpanato !
Merging.

Issue #2112, opened 2021-09-30
func populatePageValues can't handle API's returning 'since'


brianlangdon:
API's such as the Organization ListAll can return many pages of Organizations. The pagination value is called since and holds the id of the last Org returned. A simple change to populatePageValues appears, so far, to be able to support this. Not sure if this a fault or a improvement! I'm new to golang so please forgive.

Old code in populatePageValues:

			page := q.Get("page")
			if page == "" {
				continue
			}

proposed change to populatePageValues

			page := q.Get("page")
			since := q.Get("since")
			if page == "" && since == "" {
				continue
			}

			if since != "" {
				page = since
			}

gmlewis:
I see that orgs `ListAll` uses a `since` parameter to control getting the next group of results here:
https://docs.github.com/en/rest/reference/orgs#list-organizations
but the link then points here:
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#link-header
which makes no mention of receiving `since` in the header.

Additionally, the only pagination documentation I could find is here:
https://docs.github.com/en/enterprise-server@3.0/rest/guides/traversing-with-pagination
which also does not include `since`.

Can you please show an example header response that includes the `since` value?
I believe you, but would like to have it documented in this issue, please.

brianlangdon:
Here is a slightly formatted and sanitised copy of our header - we are using GHE not github.com
net/http.Header ["Content-Type": ["application/json; charset=utf-8"], 
"Cache-Control": ["private, max-age=60, s-maxage=60"], 
"Vary": ["Accept, Authorization, Cookie, X-GitHub-OTP"], 
"X-Accepted-Oauth-Scopes": [""], 
"Strict-Transport-Security": ["max-age=31536000; includeSubdomains"], 
"Referrer-Policy": ["origin-when-cross-origin, strict-origin-when-cross-origin"], 
"X-Oauth-Scopes": ["admin:gpg_key, admin:org, admin:org_hook, admin:pre_receive_hook, admin:public_key, admin:repo_hook, delete_repo, gist, notifications, repo, site_admin, user, write:discussion"], 
"X-Github-Media-Type": ["github.v3; format=json"], 
"Link": ["<https://ourserver/api/v3/organizations?since=296>; rel=\"next\", <https://ourserver/api/v3/organizations{?since}>; rel=\"first\""],
[...]

PR #2218, opened 2021-12-05
Fix since+page pagination bug

gmlewis:
Fixes: #2112.


gmlewis:
@lonre - Could you please review this PR and approve if you agree with the fix?

lonre:
@gmlewis Thanks for your help.

I agree with the fix.

gmlewis:
Thank you, @lonre !
Merging.

PR #2235, opened 2021-12-17
Change type of HookDelivery.InstallationID to Int64.

### LINKED ISSUE #2234: Incorrect type in HookDelivery
[Here](https://github.com/google/go-github/blob/1d9884e7ac8ca3c6d3d469af28af227fc773a246/github/repos_hooks_deliveries.go#L29) the `HookDelivery` defines the `InstallationID` as type `string`, but from the [documentation](https://docs.github.com/en/rest/reference/webhooks#list-deliveries-for-a-repository-webhook) it seems like a number. When testing I receive the error string 

```
json: cannot unmarshal number into Go struct field HookDelivery.installation_id of type string
```

Is this a possible bug?

pierce-m:
Resolves #2234

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -5381,9 +5381,9 @@ func (h *HookDelivery) GetID() int64 {
 }
 
 // GetInstallationID returns the InstallationID field if it's non-nil, zero value otherwise.
-func (h *HookDelivery) GetInstallationID() string {
+func (h *HookDelivery) GetInstallationID() int64 {
 	if h == nil || h.InstallationID == nil {
-		return ""
+		return 0
 	}
 	return *h.InstallationID
 }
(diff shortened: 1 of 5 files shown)


gmlewis (commented):
I'm wondering why these fields are pointers but don't have `omitempty`?

gmlewis (changes requested):
I've looked through the code and can't find an explanation for why the fields in the `HookDelivery` do not have an `omitempty` even though they are all references.

So while we are breaking the API, please go ahead and add `omitempty` to all the fields in lines 20-30.  Thank you!

pierce-m:
@gmlewis yeah it looks like it was added on the `Request` and `Response` fields 🤔, maybe because this struct isn't used as input (as far as I can tell)? Anyways, added! 👍

gmlewis (approved):
Thank you, @pierce-m !
LGTM.

Awaiting second LGTM from any other repo contributor before merging.

pierce-m:
Is there anyone else who can provide review? Doesn't seem like @wesleimp is very active any more.

Parker77 (approved):
LGTM.

gmlewis:
Thank you, @Parker77 !
Merging.

PR #2264, opened 2022-01-21
Add support for additional PagesUpdate parameters

### LINKED ISSUE #2261: Support additional GitHub pages update parameters
The https://pkg.go.dev/github.com/google/go-github/v41/github#RepositoriesService.UpdatePages API only currently supports updating the CNAME and source properties: https://github.com/google/go-github/blob/178169fc04bc4a05ca48f98709cdd02ab5b556e3/github/repos_pages.go#L90-L98

There are additional parameters which would be useful for consumers (i.e. terraform provider) to be able to update (see https://docs.github.com/en/rest/reference/pages#update-information-about-a-github-pages-site for reference):
* https_enforced
* public

quinqu:
Closes #2261 

This change introduces 2 additional parameters provided by the GitHub API to the PagesUpdate type, `Public` and `HTTPSEnforced`.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -10484,6 +10484,22 @@ func (p *PagesUpdate) GetCNAME() string {
 	return *p.CNAME
 }
 
+// GetHTTPSEnforced returns the HTTPSEnforced field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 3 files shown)


gmlewis:
Thank you, @quinqu !
Please fix the "repository" typo and sign the CLA and then we can move forward with this PR.

Issue #2260, opened 2022-01-21
Use https://staticcheck.io/


gauntface:
I just updated the internal (Google) version of go-github and one of the existing tools flagged some issues that seem reasonable to flag on GitHub.

The tool is staticcheck: https://staticcheck.io/

They look like they have a CLI and GitHub Actions.

Example of the errors it found, on [actions_artifacts.go Line 121](https://github.com/google/go-github/blob/master/github/actions_artifacts.go#L121) there is an unused error value

```go
parsedURL, err := url.Parse(resp.Header.Get("Location"))
return parsedURL, newResponse(resp), nil
```

@gmlewis and @willnorris curious what you think?

gmlewis:
> @gmlewis and @willnorris curious what you think?

Cool find, @gauntface !  Can you create a PR to incorporate it into the GitHub action workflow in this repo?

gmlewis:
Curiously, it looks like `staticcheck` might not work yet with Go 1.18 beta:

```
$ go version
go version devel go1.18-8ff254e30b Thu Dec 9 19:01:08 2021 +0000 linux/amd64

$ staticcheck ./...
-: cannot import "internal/cpu" (unknown iexport format version 2), export data is newer version - update tool (compile)
-: cannot import "internal/goarch" (unknown iexport format version 2), export data is newer version - update tool (compile)
```

gmlewis:
This looks pretty trivial: https://staticcheck.io/docs/running-staticcheck/github-actions/
I can put together a PR after I clean up the violations if there are no objections.

willnorris:
We [already run](https://github.com/google/go-github/blob/master/.github/workflows/linter.yml) golangci-lint.  We can just enable staticcheck [there](https://github.com/google/go-github/blob/master/.golangci.yml) (https://golangci-lint.run/usage/linters/)

willnorris:
apparently we explicitly disable staticcheck in https://github.com/google/go-github/blob/master/.golangci.yml, probably because it was failing at the time.  What I will typically do with golangci-lint is check in the initial config with failing checks disabled, then do follow-up commits that fix the failures and re-enable the check.  We just never did that follow-up here I guess.

gauntface:
I'll take a stab at cleaning and enabling it later today if no one is working on it.

gmlewis:
Fixed by #2903.
Closing.

PR #2267, opened 2022-01-22
Addressing issues raised by staticcheck

gauntface:
I've disabled the checks for switch from go's openpgp implementation to a community fork because it seems like it's not a drop in replacement and needs some extra work.

```
github/git_commits.go:16:2: SA1019: package golang.org/x/crypto/openpgp is deprecated: this package is unmaintained except for security fixes. New applications should consider a more focused, modern alternative to OpenPGP for their specific task. If you are required to interoperate with OpenPGP systems and need a maintained package, consider a community fork. See https://golang.org/issue/44226. (staticcheck)
        "golang.org/x/crypto/openpgp"
```

Original issue: #2260

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,10 +18,10 @@ linters:
   disable:
     - errcheck
     - gosimple
-    - staticcheck
     - ineffassign
+    - staticcheck
     - unused
-issues: 
+issues:
   exclude-rules:
   - linters:
     - dogsled
@@ -30,3 +30,15 @@ issues:
   - linters:
     - dupl
     path: _test\.go
+  - linters:
+    - staticcheck
+    path: github/github_test\.go
+    text: 'SA1012'
+  - linters:
+    - staticcheck
+    path: _test\.go
+    text: 'SA1019'
+  - linters:
+    - staticcheck
+    path: github/git_commits\.go
+    text: 'SA1019'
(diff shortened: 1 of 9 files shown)


github/issue_import.go
```diff
@@ -89,7 +89,7 @@ func (s *IssueImportService) Create(ctx context.Context, owner, repo string, iss
 		if ok {
 			decErr := json.Unmarshal(aerr.Raw, i)
 			if decErr != nil {
-				err = decErr
+				return nil, resp, decErr
```
gmlewis, line 92: I'm concerned about this change in behavior... it looks to me like what we had before was most likely broken (because `i` should not have been returned), but I'm thinking that the original `AcceptedError` should be returned here instead of the attempt to unmarshal the JSON.

Because if we return the `json.Unmarshal` error, the client will have no way of knowing that this is actually an `AcceptedError` meaning that GitHub accepted the call but decided to not yet return a value.

Thoughts?

gauntface: yeah it's unusual here because the err was never used in this can, so we could simply drop the decErr which will be the current behavior

gmlewis: Except that if `decErr` is `nil` then we do want to actually return the value of `i`, so we cannot drop it entirely. I'm just saying we should return `aerr` here instead.

.golangci.yml
```diff
@@ -30,3 +30,15 @@ issues:
   - linters:
     - dupl
     path: _test\.go
+  - linters:
```
willnorris, line 33: it's probably worth adding at least a one-line comment explaining what each of these rules are that are being excluded

Issue #2274, opened 2022-01-27
[Feature Request] Add checks field to RequiredStatusChecks


k24dizzle:
According to the [Github API](https://docs.github.com/en/rest/reference/branches#update-branch-protection), there is a new `checks` field that is supposed to replace the `contexts` field when setting required status checks.
<img width="798" alt="Screen Shot 2022-01-27 at 12 16 01 PM" src="https://user-images.githubusercontent.com/15153943/151436727-3034f3bd-4f1d-453f-90f4-c5c9c430e2b2.png">

Note: if `contexts` is required, not sure how that will interact with the `checks` field.

Add this field [here](https://github.com/google/go-github/blob/295020540d58efd38acedd29f5cdd4e2852b5c00/github/repos.go#L876).
<img width="1097" alt="Screen Shot 2022-01-27 at 12 14 26 PM" src="https://user-images.githubusercontent.com/15153943/151436525-9ca9d387-8919-46c1-af46-ca8f57a31ad3.png">

gmlewis:
Thank you, @k24dizzle !  Would you like to make a PR, or shall I open this up to other contributors to this repo?

k24dizzle:
Hi @gmlewis, new to contributing to this repo but I can give it a shot

gmlewis:
Excellent, @k24dizzle !
Please make sure to check out our CONTRIBUTING.md file for how to get started.

k24dizzle:
I did some more research, with some CURLs and realized that in `required_status_checks` you must supply either `contexts` or `checks`, but you cannot supply both at the same time:

```
curl   -X PUT   -H "Accept: application/vnd.github.v3+json" \
  -H "Authorization: token REDACTED" \
    https://api.github.com/repos/ORG/REPO_NAME/branches/BRANCH_NAME/protection  \
  -d '{"enforce_admins": null, "required_pull_request_reviews": null,
     "required_status_checks":{"strict":true,"contexts":["contexts"],"checks":[{"context":"contexts"}]}, 
     "restrictions": null}'
{
  "message": "Invalid request.\n\nNo subschema in \"anyOf\" matched.\nMore than one subschema in \"oneOf\" matched.\nNot all subschemas of \"allOf\" matched.\nFor 'anyOf/1', {\"strict\"=>true, \"contexts\"=>[\"contexts\"], \"checks\"=>[{\"context\"=>\"contexts\"}]} is not a null.",
[...]

PR #2276, opened 2022-01-28
Add checks field to RequiredStatusChecks

k24dizzle:
Attempt to solve #2274.

From my own experimentation, I discovered that there's two ways that this `Checks` field can be used to update the required status checks for a protected branch:
- [update branch protection](https://docs.github.com/en/rest/reference/branches#update-branch-protection) -> via the `ProtectionRequest` -> `RequiredStatusChecks` -> `RequiredStatusCheck` struct, in this API call we must provide either one of `Checks` or `Contexts` (the other must be `nil`) in `RequiredStatusChecks` to successfully update a branch protection. If you provide both or neither, you will get an `Invalid Request` error.
- [update status check protection](https://docs.github.com/en/rest/reference/branches#update-status-check-protection) -> via the `RequiredStatusChecksRequest` -> `RequiredStatusCheck` struct, in this API call `Checks` and `Contexts` are both optional fields, if you provide both of them only `Checks` will be used to update the status checks.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -877,14 +877,33 @@ type RequiredStatusChecks struct {
 	// Require branches to be up to date before merging. (Required.)
 	Strict bool `json:"strict"`
 	// The list of status checks to require in order to merge into this
-	// branch. (Required; use []string{} instead of nil for empty list.)
-	Contexts []string `json:"contexts"`
+	// branch. (Deprecated. Note: only one of Contexts/Checks can be populated,
+	// but at least one must be populated).
+	Contexts []string `json:"contexts,omitempty"`
+	// The list of status checks to require in order to merge into this
+	// branch.
+	Checks []*RequiredStatusCheck `json:"checks,omitempty"`
 }
 
 // RequiredStatusChecksRequest represents a request to edit a protected branch's status checks.
 type RequiredStatusChecksRequest struct {
-	Strict   *bool    `json:"strict,omitempty"`
-	Contexts []string `json:"contexts,omitempty"`
+	Strict *bool `json:"strict,omitempty"`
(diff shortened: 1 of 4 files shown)


github/repos.go
```diff
@@ -877,14 +877,33 @@ type RequiredStatusChecks struct {
...
+	Checks []*RequiredStatusCheck `json:"checks,omitempty"`
 }
 
 // RequiredStatusChecksRequest represents a request to edit a protected branch's status checks.
 type RequiredStatusChecksRequest struct {
-	Strict   *bool    `json:"strict,omitempty"`
-	Contexts []string `json:"contexts,omitempty"`
+	Strict *bool `json:"strict,omitempty"`
+	// Note: if both Contexts and Checks are populated,
+	// the Github API will only use Checks.
```
gmlewis, line 892:
```suggestion
	// the GitHub API will only use Checks.
```

github/repos.go
```diff
@@ -877,14 +877,33 @@ type RequiredStatusChecks struct {
...
+	Contexts []string              `json:"contexts,omitempty"`
+	Checks   []RequiredStatusCheck `json:"checks,omitempty"`
```
gmlewis, line 894:
```suggestion
	Checks   []*RequiredStatusCheck `json:"checks,omitempty"`
```

gmlewis (changes requested):
Thank you, @k24dizzle !
Just a couple tweaks, please.

gmlewis:
It looks like the linter has a couple issues as well.

gmlewis (approved):
Thank you, @k24dizzle !
LGTM.
Merging.

Issue #2292, opened 2022-02-15
not able to activate advanced security


dustincrogers:
Forgive my naivety as an amateur developer, especially in go.

I am trying to use `Repositories.Edit` to update the `SecurityAndAnalysis.AdvancedSecurity.Status` to `enabled` and even though I get 200 status, I receive an empty response. The field does not update as expected. 

The response body appears to not have the field available, which is probably the issue. 

I am definitely able to see the status of the field, using `SecurityAndAnalysis.GetAdvancedSecurity()`

Does anyone know what I may be doing wrong or if this is maybe a bug somewhere?
Thanks in advance!

gmlewis:
Two options: 

* try the curl-only version and see if you get the same results.
* post an example snippet of your request here so we can see what you are attempting

dustincrogers:
Thanks for taking a look and for the advice! This is working fine to enable/disable github advanced security with curl, so I am not getting the same results. Probably just something with my code then. 

Here is my snippet, please excuse the fact that its messy still:
```
// ActivateSecurityAndAnalysis activates the security and analysis object for the given repo
func (r *Repo) ActivateSecurityAndAnalysis(client *ghtools.Client) error {
	repo, _, err := client.API.Repositories.Get(context.Background(), r.Owner, r.Name)
	if err != nil {
		return errors.Wrap(err, "could not get repository definition")
	}

	if repo.GetSecurityAndAnalysis() == nil {
		repo.SecurityAndAnalysis = &github.SecurityAndAnalysis{}
	}

	if repo.SecurityAndAnalysis.GetAdvancedSecurity() == nil {
		repo.SecurityAndAnalysis.AdvancedSecurity = &github.AdvancedSecurity{}
	}

	// now we know that repo.SecurityAndAnalysis.AdvancedSecurity is initialized
	// we are finally positive there are no nil pointers along the chain
	//set it to enabled
	enabled := "enabled"
	fmt.Printf("Before repo.SecurityAndAnalysis.AdvancedSecurity.Status: %v\n", *repo.SecurityAndAnalysis.AdvancedSecurity.Status)
	repo.SecurityAndAnalysis.AdvancedSecurity.Status = &enabled
	fmt.Printf("After repo.SecurityAndAnalysis.AdvancedSecurity.Status: %v\n", *repo.SecurityAndAnalysis.AdvancedSecurity.Status)
	//repo.SecurityAndAnalysis.SecretScanning.Status = &enabled

[...]

PR #2295, opened 2022-02-18
Add ability to dump GitHub API requests as `curl` commands using context.WithValue

gmlewis:
This is a feature that I have wanted for many years, and #2292 finally caused me to write it.
By adding a context value like this:

```go
ctx = context.WithValue(ctx, github.DebugRequest, "curl")
```

this client will now dump the equivalent `curl` command for every call that is being attempted, which is useful for debugging.

This is an alternative implementation to #2302 in order to see which is better.

### DIFF
--- a/example/topics/main.go
+++ b/example/topics/main.go
@@ -6,6 +6,9 @@
 // The simple command demonstrates the functionality that
 // prompts the user for a GitHub topic and lists all the entities
 // that are related to the specified topic or subject.
+//
+// It also demonstrates how to debug a GitHub v3 API call
+// by dumping its `curl` equivalent.
 package main
 
 import (
@@ -18,7 +21,9 @@ import (
 // Fetch and lists all the public topics associated with the specified GitHub topic
 func fetchTopics(topic string) (*github.TopicsSearchResult, error) {
 	client := github.NewClient(nil)
-	topics, _, err := client.Search.Topics(context.Background(), topic, nil)
+	ctx := context.Background()
+	ctx = context.WithValue(ctx, github.DebugRequest, "curl") // comment out this line to stop debug info.
+	topics, _, err := client.Search.Topics(ctx, topic, nil)
 	return topics, err
 }
 
--- a/github/github.go
+++ b/github/github.go
@@ -1324,9 +1324,7 @@ func dumpRequestAsCurl(req *http.Request) (string, error) {
 		if err != nil {
 			return "", err
 		}
-
 		lines = append(lines, fmt.Sprintf("-d '%s'", buf))
-
 		req.Body = ioutil.NopCloser(bytes.NewBuffer(buf))
 	}
 
--- a/github/github.go
+++ b/github/github.go
@@ -1299,6 +1299,10 @@ func formatRateReset(d time.Duration) string {
 	return fmt.Sprintf("[rate reset in %v]", timeString)
 }
 
+func escapeSingleQuote(s string) string {
+	return strings.ReplaceAll(s, "'", `\'`)
+}
+
 // dumpRequestAsCurl dumps an outbound request as a curl command to a string
 // for debugging purposes. It redacts any "Authorization" string in the
 // header or client secret in the URL in order to prevent logging secrets.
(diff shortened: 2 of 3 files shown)


gmlewis:
@willnorris and/or @gauntface - what do you think about adding this to aid in debugging and user support?

willnorris:
This is definitely super cool, but does it belong in go-github?  Is there anything github specific about it?  Would it make more sense for this to be a standalone `http.RoundTripper` or `http.Client` that you feed into your `github.Client` much in the same way you do to add oauth support?  That would make it usable for any http client.  In fact, I wonder if something like this already exists.

(I even vaguely remember doing something like this at Google for debugging, though I think it might have been an http.Handler for debugging server-side requests.)

gmlewis:
@willnorris - would you prefer to see an implementation using a `RoundTripper`?
I could make a separate PR and we could compare the two for ease-of-use/simplicity/cleanness-of-code if you are interested.

willnorris:
I think seeing a `RoundTripper` implementation would certainly be helpful (even if it's just partial).  I'm also curious what @gauntface thinks.

gmlewis:
Now that I've implemented both versions and thought about this further, I'm even more convinced that this solution belongs in this repo and that this version (as opposed to the Transport/RoundTripper version) is the better one.  Here's why:

* By placing the solution in this repo, we can carefully redact passwords in two placed that might be leaked by other implementations (in the URL query and in the `Authorization` header)
* This version is a one-liner that can be easily commented out (or guarded by an `if` at runtime) with no other modifications to the code and is therefore the lowest-friction version to use
* This version allows us to pass a string describing what kind of debugging we would like to be performed (`"curl"` in this case), thereby easily allowing future expansion to other kinds of debugging helpers

Thoughts?

raynigon:
Hi @gmlewis ,
I understand the points and fully agree.

In addition I think an implementation in this repo would allow easily understandable curl statements whereas a generic implementation could become cluttered with unecessary parameters. These unecessary parameters could be safely removed in a local implementation.

willnorris:
I'm still not convinced this belongs in go-github, at least certainly not for the initial implementation.  I think this is better suited as a custom http.RoundTripper (or possibly as an [httptrace.ClientTrace](https://pkg.go.dev/net/http/httptrace#ClientTrace)?!  I just discovered this package recently) in a standalone package.

To some of your specific points:

> By placing the solution in this repo, we can carefully redact passwords in two placed that might be leaked by other implementations (in the URL query and in the Authorization header)

If you author the standalone library, you have exactly the same amount of control of how carefully it redacts information.  That's also outside the concerns of go-github.  We wouldn't have a dependency on that library, it's just something that we may recommend as a useful debugging tool, no different than any other number of useful debugging tools.

This same point applies to @raynigon's comment about concern of an external implementation becoming cluttered with unnecessary parameters.  It's both a) no different if the library is github.com/gmlewis/httpdebug and b) not the concern of go-github.  In fact, I very much don't **want** it to be the concern of go-github.

> This version is a one-liner that can be easily commented out (or guarded by an if at runtime) with no other modifications to the code and is therefore the lowest-friction version to use

Certainly, I can't argue that this is easier.  But I also don't think the alternative of importing a special debug package is **that** much harder, and I don't think how seldom this is likely to be used warrants its inclusion.

> This version allows us to pass a string describing what kind of debugging we would like to be performed ("curl" in this case), thereby easily allowing future expansion to other kinds of debugging helpers

And using a custom `http.RoundTripper` provides infinite flexibility as well, since you can simply wrap something different.  This is a **very** common pattern on the server side of the http package with nested http.Handlers.


In summary, I've always tried very hard to keep go-github laser-focused on implementing the GitHub REST API and nothing more.  In fact, moving go-querystring out into a separate package proved to be a great choice, since it's now [used by thousands](https://pkg.go.dev/github.com/google/go-querystring/query?tab=importedby) of other similar client libraries.  I think the same applies here, and better follows Go idioms.  We're talking about generic functionality that isn't specific to GitHub, and is only even needed occasionally for debugging.

And if we did decide it made sense to move it into go-github, that's always something that could be done in the future after the standalone package had time to mature.  But once we put it in, we can't easily remove it, and so I think we should be appropriately cautious about what we add.

All of that said, I recognize that this isn't really my call.  I'm not the primary maintainer of this library anymore, and I'm not even using it on a regular basis at this point (though I certainly hope to be doing so again soon! 😄).  But that's my opinion for what it's worth.

gmlewis:
> But that's my opinion for what it's worth.

Thank you, @willnorris , for the thoughtful feedback.

I highly value your opinion from years of working with you, and also obviously because you are the original author of this repo. 😁

I'm going to close this PR and its counterpart and consider implementing this in an external repo... maybe even with both implementations... or maybe even splurge and make two separate repos. 😁

emilva:
That was quick, thanks!

PR #2301, opened 2022-02-22
Revert "Remove graduated GitHub v3 API preview header"

### LINKED ISSUE #2300: Preview API Headers removed too soon from check API.
Having updated from v32 to v42 the checks API is now failing on our Github Enterprise instance. It looks like #1636 removed the headers. #2188 suggests we now know this shouldn't be done. Can we get those headers back?

brackendawson:
Reverts google/go-github#1636

Fixes #2300


gmlewis (approved):
Thank you, @brackendawson !
LGTM.
Merging after tests pass.

PR #2302, opened 2022-02-24
Add ability to dump GitHub API requests as curl commands using RoundTripper

gmlewis:
This is a feature that I have wanted for many years, and #2292 finally caused me to write it.
By adding a RoundTripper like this (when using a custom `Transport`):

```go
	tp := &github.DebugCurlTransport{Transport: tc.Transport}
	client := github.NewClient(tp.Client())
```

or like this (when using a `DefaultTransport`):

```go
	tp := &github.DebugCurlTransport{}
	client := github.NewClient(tp.Client())
```

this client will now dump the equivalent `curl` command for every call that is being attempted, which is useful for debugging.

This is an alternative implementation to #2295 in order to see which is better.

### DIFF
--- a/example/tokenauth/main.go
+++ b/example/tokenauth/main.go
@@ -6,6 +6,9 @@
 // The tokenauth command demonstrates using the oauth2.StaticTokenSource.
 // You can test out a GitHub Personal Access Token using this simple example.
 // You can generate them here: https://github.com/settings/tokens
+//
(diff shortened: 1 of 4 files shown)


gmlewis:
@willnorris and @gauntface - here is the RoundTripper version of the curl debug helper for comparison.  See description above.

gmlewis:
Closing. See discussion in #2295.

PR #2310, opened 2022-03-02
Move examples dir into its own Go module

wlynch:
We (in particular shoutout to @asvoboda for noticing this in https://github.com/bradleyfalzon/ghinstallation/pull/63)
 noticed that go-github has an example that references ghinstallation,
which causes a circular module dependency. While this is technically
fine (https://go.dev/ref/mod#minimal-version-selection should resolve
it), there's probably not a strong reason to pull in ghinstallation and
other example dependencies not needed by the core github package unless
it is strictly needed.

This just breaks the example folder into its own module so that users
can import the core go-github package with the minimal set of
dependencies. This should not affect any functionality.
I tried to keep versions roughly the same for compatibility, though
replaced the go-github dependency with the parent to try and keep
examples up-to-date with HEAD (this could possibly be replaced with
https://go.dev/ref/mod#workspaces later).


example/go.mod
```diff
@@ -0,0 +1,24 @@
+module github.com/google/go-github/example/v43
```
gmlewis, line 1: I'm thinking that it might be better to mimic the other sub-modules and remove the version number here:

```bash
$ grep module */go.mod
scrape/go.mod:module github.com/google/go-github/scrape
update-urls/go.mod:module github.com/google/go-github/update-urls
```

```suggestion
module github.com/google/go-github/example
```

gmlewis (changes requested):

example/go.mod
wlynch: Done!

gmlewis (approved):
Thank you, @wlynch !
(In the future, please don't force-push PRs for this repo, as we always squash-and-merge at the end anyway... see CONTRIBUTING.md for more details.)
LGTM.

Awaiting second LGTM from any other contributor to this repo.

frioux:
Looking forward to this one getting merged.  The fact that the current release expresses a dependency on an older version of itself... seems not great 😅

gmlewis:
> Looking forward to this one getting merged. The fact that the current release expresses a dependency on an older version of itself... seems not great 😅

If you give me a second LGTM/Approval, @frioux , I can go ahead and merge it. 😁

asvoboda:
👍 LGTM.

gmlewis:
Thank you, @asvoboda !
Merging.

PR #2326, opened 2022-03-30
Add support for the pull request review thread event

### LINKED ISSUE #2324: Webhook event type `pull_request_review_thread` missing
It seems like GitHub is sending the webhook event type `pull_request_review_thread` when someone resolves a comment in a PR.

Webhook event schema can be seen here: https://github.com/octokit/webhooks/blob/d93eebf4c48dc3c36f9c7ceffab356fa444bf9fc/payload-schemas/api.github.com/pull_request_review_thread/resolved.schema.json

The event doesn't seem to be documented in [Webhook events and payload](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request_review_thread) though.

taisph:
Fixes #2324

### DIFF
--- a/github/pulls_threads.go
+++ b/github/pulls_threads.go
@@ -1,0 +1,17 @@
+// Copyright 2022 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+// PullRequestThread represents a thread of comments on a pull request.
+type PullRequestThread struct {
+	ID       *int64                `json:"id,omitempty"`
+	NodeID   *string               `json:"node_id,omitempty"`
+	Comments []*PullRequestComment `json:"comments,omitempty"`
+}
+
+func (p PullRequestThread) String() string {
+	return Stringify(p)
+}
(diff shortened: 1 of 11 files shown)


github/pulls_threads.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+// PullRequestThread represents a ...
```
gmlewis, line 8: Please continue. :smile:

gmlewis (changes requested):

taisph:
I added the link https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#pull_request_review_thread which presumably will be the correct link once they update the docs by looking at the existing structure. I don't know what to expect with regards to when they'll update it as I managed to find [a misnamed issue](https://github.com/github/docs/issues/12295) about the missing documentation that was created back in November 2021.

 Should this PR be kept back until the docs are updated?

gmlewis:
> Should this PR be kept back until the docs are updated?

I don't think that's necessary.  Many endpoints have been added in the based just based on observing the response received, so I think this is fine. It can always be updated once the official docs are released. Thank you, @taisph !

gmlewis (approved):
Thank you, @taisph !
LGTM.

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

PR #2333, opened 2022-04-11
Add create repository with an initial commit with empty README

yuseferi:


gmlewis (changes requested):
Thank you, @yuseferi .
Instead of adding a new example that is almost identical to the `newrepo` example but with a new flag, please just edit the other example to add your new flag.

yuseferi:
> Thank you, @yuseferi . Instead of adding a new example that is almost identical to the `newrepo` example but with a new flag, please just edit the other example to add your new flag.

sure. just updated.

gmlewis (changes requested):
I think you just need to run `gofmt` on the modified file, and we should be ready to merge.

yuseferi:
> I think you just need to run `gofmt` on the modified file, and we should be ready to merge.

Done

gmlewis (approved):
Thank you, @yuseferi !
LGTM.
Merging.

PR #2334, opened 2022-04-13
Restrict permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,8 +1,14 @@
 on: [push, pull_request]
 name: linter
 
+permissions:
+  contents: read
+
 jobs:
   lint:
+    permissions:
+      contents: read  # for actions/checkout to fetch code
+      pull-requests: read  # for golangci/golangci-lint-action to fetch pull requests
     strategy:
       matrix:
         go-version: [1.x]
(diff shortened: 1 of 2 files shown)


naveensrinivasan:
> In all the links you provided, I didn't see any need or example for adding `actions: write`.
> Can you please convince me that it is needed?



This has the https://github.com/styfle/cancel-workflow-action permissions https://github.com/styfle/cancel-workflow-action#advanced-token-permissions HTH. 

> What would fail if we have `actions: read` ?

I don't know. Probably asking the owner of the action would help.

gmlewis (approved):
Thank you, @naveensrinivasan - LGTM.
Merging.

PR #2348, opened 2022-04-28
commit with obvious gofmt issue

willnorris:


willnorris:
@gmlewis well this one failed: https://github.com/google/go-github/runs/6220310903?check_suite_focus=true

weird.

willnorris:
This looks to be a problem with golangci-lint, or how we're using it.  Running it manually at commit 51df4647e1a6f8565afb511fba405b883d36e4c3 (the last in #2333 before the gofmt fix) passes when it shouldn't.  I'll keep digging.

willnorris:
Ah, I think I found it.... it's because the example directory is its own go module.  I don't think golangci-lint will descent into other modules.  cd'ing into the examples directory and then running golangci-lint there correctly identifies the issue.  So we'll need to change our actions workflow to manually run on examples and scrape and any other separate modules we have.  Gotta run for now, but I can take a look at that later.

PR #2350, opened 2022-04-29
run golangci-lint on all modules

### LINKED ISSUE #2349: Fix golangci-lint workflow to run in submodules
See discussion in #2348

willnorris:
Fixes #2349


willnorris:
I don't love that this now expands out to four linter runs for every PR, but it's the cleanest approach I think and they run super fast.

gmlewis (approved):
Thank you, @willnorris !
LGTM.

PR #2365, opened 2022-05-15
Add missing Sender field in RepositoryVulnerabilityAlertEvent

ldez:
Reference: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#repository_vulnerability_alert


gmlewis (approved):
Thank you, @ldez !
LGTM.
Merging.

Issue #2360, opened 2022-05-05
Documentation links have changed


Jille:
The documentation links in the Godoc seem to have broken.

For example "https://docs.github.com/en/free-pro-team@latest/rest/reference/apps/#delete-an-app-token" should now be /rest/**apps**/apps/.

gmlewis:
> The documentation links in the Godoc seem to have broken.
> 
> For example "https://docs.github.com/en/free-pro-team@latest/rest/reference/apps/#delete-an-app-token" should now be /rest/**apps**/apps/.

Yes, the URLs on the GitHub documentation website change quite frequently. PRs are welcome to fix the links.
Thanks.

PR #2369, opened 2022-05-22
update-urls: update sidebar child articles links #2360

ANaphade:


gmlewis (approved):
Thank you, @ANaphade .
LGTM.
Merging.

PR #2370, opened 2022-05-23
fix: correct endpoints for custom roles

joshuahancox:
This function was created based on the docs here: 
https://docs.github.com/en/rest/orgs/custom-roles#list-custom-repository-roles-in-an-organization

The docs incorrectly give the endpoint `https://api.github.com/organizations/ORGANIZATION_ID/custom_roles`.

When using this function, I get a 404. 

The correct endpoint for the custom_roles endpoint is `https://api.github.com/orgs/ORGANIZATION_ID/custom_roles` - and this is consistent with other github endpoints under orgs. I have raised a support ticket with github, but I would expect them to change the documentation, not the api endpoint.

Also changed the file names for consistency.

### DIFF
--- a/github/orgs_custom_roles.go
+++ b/github/orgs_custom_roles.go
@@ -44,4 +44,3 @@ func (s *OrganizationsService) ListCustomRepoRoles(ctx context.Context, org stri
 
 	return customRepoRoles, resp, nil
 }
-diff --git a/github/org_custom_roles.go b/github/org_custom_roles.go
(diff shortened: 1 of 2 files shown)


joshuahancox:
Hey both @gmlewis @tamboliasir1!

Could you take a look at this change for me when you get a second please?

gmlewis (approved):
Thank you, @joshuahancox !
LGTM.
Merging.

PR #2379, opened 2022-06-07
Include RoleName field in Repository struct

joshuahancox:


gmlewis (approved):
Thank you, @joshuahancox !
LGTM.
Merging.

PR #2384, opened 2022-06-15
Add missing common event fields.

wlynch:
See
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties
for the list of common fields.

I may have missed some - but gives better coverage for fields we were
missing before.

Signed-off-by: Billy Lynch <billy@chainguard.dev>


gmlewis (approved):
Thank you, @wlynch !
LGTM.
Merging.

PR #2397, opened 2022-06-26
Add hooks support for WebSub (formerly PubSubHubbub) protocol

### LINKED ISSUE #1265: Support PubSubHubbub API for creating hooks
Are there any plans to support [the PubSubHubbub API for creating webhooks](https://developer.github.com/v3/repos/hooks/#pubsubhubbub)?

I could try to create a PR for implementing this (I'm hoping to use it for implementing https://github.com/terraform-providers/terraform-provider-github/issues/275) if there is any chance of it being accepted.

YuyaAbo:
resolves #1265

_Note: PubSubHubbub seems to have been renamed WebSub when it was adopted by the W3C. So we use the name WebSub in our code.
ref https://github.com/pubsubhubbub/PubSubHubbub_

Added support for API to subscribe and unsubscribe to repository events via WebSub protocol.

GitHub API Docs: https://docs.github.com/en/rest/webhooks#pubsubhubbub
WebSub Docs: https://www.w3.org/TR/websub/

WebSub requires Content-Type to be application/x-www-form-urlencoded, so a new factory function has been added.

> This request MUST have a Content-Type header of application/x-www-form-urlencoded (described in Section 4.10.22.6 [[HTML5](https://www.w3.org/TR/websub/#bib-HTML5)])
> https://www.w3.org/TR/websub/#subscriber-sends-subscription-request

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -424,6 +424,33 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
 	return req, nil
 }
 
+// NewFormRequest creates an API request. A relative URL can be provided in urlStr,
+// in which case it is resolved relative to the BaseURL of the Client.
+// Relative URLs should always be specified without a preceding slash.
+// Body is sent with Content-Type: application/x-www-form-urlencoded.
+func (c *Client) NewFormRequest(urlStr string, body io.Reader) (*http.Request, error) {
+	if !strings.HasSuffix(c.BaseURL.Path, "/") {
+		return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL)
+	}
+
+	u, err := c.BaseURL.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, u.String(), body)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Accept", mediaTypeV3)
+	if c.UserAgent != "" {
+		req.Header.Set("User-Agent", c.UserAgent)
+	}
+	return req, nil
+}
+
 // NewUploadRequest creates an upload request. A relative URL can be provided in
 // urlStr, in which case it is resolved relative to the UploadURL of the Client.
 // Relative URLs should always be specified without a preceding slash.
--- a/github/repos_hooks.go
+++ b/github/repos_hooks.go
@@ -8,6 +8,9 @@ package github
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
 	"time"
 )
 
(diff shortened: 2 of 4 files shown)


github/repos_hooks.go
```diff
@@ -197,3 +200,55 @@ func (s *RepositoriesService) TestHook(ctx context.Context, owner, repo string,
...
+func (s *RepositoriesService) Subscribe(ctx context.Context, owner, repo, event, callback string, secret []byte) (*Response, error) {
+	req, err := s.createWebSubRequest("subscribe", owner, repo, event, callback, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// Unsubscribe lets servers unregister to don't receive updates when a topic is updated.
```
gmlewis, line 216:
```suggestion
// Unsubscribe lets servers unregister to no longer receive updates when a topic is updated.
```

github/repos_hooks.go
```diff
@@ -197,3 +200,55 @@ func (s *RepositoriesService) TestHook(ctx context.Context, owner, repo string,
...
+func (s *RepositoriesService) Unsubscribe(ctx context.Context, owner, repo, event, callback string, secret []byte) (*Response, error) {
+	req, err := s.createWebSubRequest("unsubscribe", owner, repo, event, callback, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.client.Do(ctx, req, nil)
+}
+
+// createWebSubRequest returns subscribe/unsubscribe request that implements
```
gmlewis, line 228:
```suggestion
// createWebSubRequest returns a subscribe/unsubscribe request that implements
```

gmlewis (changes requested):
Thank you, @YuyaAbo !
Just a couple tiny tweaks, please, then we should be ready for a second approval before merging.

gmlewis (approved):
Thank you, @YuyaAbo !
LGTM.

github/github.go
```diff
@@ -424,6 +424,33 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}) (*http.Requ
...
+
+	u, err := c.BaseURL.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, u.String(), body)
+	if err != nil {
+		return nil, err
+	}
```
raynigon, line 444: It would be nice if this case got tested.

raynigon (approved):
Can be merged, but it would be nice if the error case test gets added first.

github/github.go
YuyaAbo: There seems to be no case where this `http.NewRequest` returns an error.

There are two cases in which `http.NewRequest` returns an error: when the method is invalid or when the url fails to parse. `NewFormRequest` is set to POST, so the former error does not occur. Also, since url parsing is done in `c.BaseURL.Parse` before this process, if url parsing fails, an error is returned before this process.

Even the already existing NewUploadRequest does not seem to cover this case. Is there a policy for such cases?

gmlewis: Could you pass in an `iotest.ErrReader` as the body to `NewRequest` to make it fail?
https://pkg.go.dev/testing/iotest#ErrReader

YuyaAbo: thanks. It didn't work. Just generating a `http.Request` will not cause it to be read by the `iotest.ErrReader`.

gmlewis: OK, thanks for trying.  I'm OK with that.
Merging.

Issue #2400, opened 2022-06-28
Dependabot Org Secrets creation returns 422 error


lneves75:
Hi all. This surfaced when creating dependabot organisation secrets using the terraform provider.

### Versions

```
Terraform v0.14.5
+ provider registry.terraform.io/hashicorp/aws v4.20.1
+ provider registry.terraform.io/hashicorp/github v4.26.1
```

Terraform github provider
- ["github.com/google/go-github/v45/github"](https://github.com/integrations/terraform-provider-github/blob/ab903b6addf72b006bd1dda250623b8c9f3b508f/go.mod#L8)
- [go version 1.16](https://github.com/integrations/terraform-provider-github/blob/ab903b6addf72b006bd1dda250623b8c9f3b508f/go.mod#L3)

### Expected Behaviour 

The API is consistent with the docs and returns a type mismatch

### Actual Behaviour

`Error: PUT https://api.github.com/orgs/myorg/dependabot/secrets/BAR: 422 Invalid request.`

### Debug Output

[Output of request and error response call to the `/orgs/{org}/dependabot/secrets/{secret_name}` endpoint](https://gist.github.com/lneves75/d4154e9504cc3f4ee2694ea18cb3508c)

[Output of request and success response call to the `/orgs/{org}/actions/secrets/{secret_name}` endpoint (for comparison)](https://gist.github.com/lneves75/d4a65d4603f3436c2827509178b779c8)

### Other inputs / findings

There's an inconsistency between the behaviour of both the [actions secrets](https://docs.github.com/en/rest/actions/secrets#create-or-update-an-organization-secret--parameters) and [dependabot secrets](https://docs.github.com/en/rest/dependabot/secrets#create-or-update-an-organization-secret--parameters) endpoints, despite the fact that the docs mention the same type, an array of strings, for the `selected_repository_ids` attribute.

Maybe the actions secret endpoint converts the array of integers into an array of strings or there's no hard type validation in place.

Making a PUT API call to `/orgs/myorg/dependabot/secrets/BAR` with the body bellow, I'm able to successfully add a dependabot secret for a couple of repos in the org.
```json
{
 "key_id": "<redacted>",
 "encrypted_value": "<redacted>",
 "visibility": "selected",
 "selected_repository_ids": [
  "123",
  "456"
 ]
}
```

gmlewis:
Thank you for the bug report, @lneves75 , apologies for the problem.

Would you like to submit a PR for this issue or would you like someone else to work on it?

PR #2401, opened 2022-06-28
Change DependabotSecretsSelectedRepoIDs to []string

gmlewis:
Fixes: #2400.

### DIFF
--- a/github/actions_secrets.go
+++ b/github/actions_secrets.go
@@ -186,8 +186,8 @@ func (s *ActionsService) GetEnvSecret(ctx context.Context, repoID int, env, secr
 	return s.getSecret(ctx, url)
 }
 
-// SelectedRepoIDs are the repository IDs that have access to the secret.
-type SelectedRepoIDs []string
+// SelectedRepoIDs are the repository IDs that have access to the action secrets.
+type SelectedRepoIDs []int64
 
 // EncryptedSecret represents a secret that is encrypted using a public key.
 //
--- a/github/actions_secrets.go
+++ b/github/actions_secrets.go
@@ -187,7 +187,7 @@ func (s *ActionsService) GetEnvSecret(ctx context.Context, repoID int, env, secr
 }
 
 // SelectedRepoIDs are the repository IDs that have access to the secret.
-type SelectedRepoIDs []int64
+type SelectedRepoIDs []string
 
 // EncryptedSecret represents a secret that is encrypted using a public key.
 //
--- a/github/actions_secrets.go
+++ b/github/actions_secrets.go
@@ -186,7 +186,7 @@ func (s *ActionsService) GetEnvSecret(ctx context.Context, repoID int, env, secr
 	return s.getSecret(ctx, url)
 }
 
-// SelectedRepoIDs are the repository IDs that have access to the action secrets.
+// SelectedRepoIDs are the repository IDs that have access to the actions secrets.
 type SelectedRepoIDs []int64
 
 // EncryptedSecret represents a secret that is encrypted using a public key.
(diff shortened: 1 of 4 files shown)


gmlewis:
@lneves75 - could you please review and approve if you agree to this change when you have a chance?

gmlewis:
Awaiting LGTM+Approval from two contributors to this repo before merging.

lneves75:
@gmlewis apparently the problem lies on github's API implementation for both endpoints.
If we call the endpoint `/orgs/myorg/actions/secrets/BAR` with a set of strings we get a 422
```json
{
	"message": "Invalid request.\n\nFor 'items', \"123\" is not an integer.\nFor 'items', \"456\" is not an integer.",
	"documentation_url": "https://docs.github.com/rest/reference/actions#create-or-update-an-organization-secret"
}
```

So I'm afraid that we need to have two distinct types 😞 
```go
type ActionSecretsSelectedRepoIDs []int64
type DependabotSecretsSelectedRepoIDs []string
```  
WDYT ?

timrogers:
> OK, that is really weird. Can you please contact GitHub tech support and verify this correct?

👋🏻 Hey there! I'm the PM for the API at GitHub, and I happened to just spot this thread.

I can confirm that the current behaviour of the API is that:

* the Actions "[Create or update an organization secret](https://docs.github.com/en/rest/actions/secrets#create-or-update-an-organization-secret)" API only has documented support for an array of integers, but it actually accepts strings or integers. (Until recently, it said it was an array of strings.)
* the Dependabot "[Create or update an organization secret](https://docs.github.com/en/rest/dependabot/secrets#create-or-update-an-organization-secret)" API only supports an array of strings

This is obviously a bit weird at the moment and is something we need to clean up - but in a way that avoids breaking changes for users!

gmlewis:
> 👋🏻 Hey there! I'm the PM for the API at GitHub, and I happened to just spot this thread.

Hi @timrogers - it's great to have you here!  Thank you to you and your team for an amazing product... it is greatly appreciated!

> I can confirm that the current behaviour of the API is that:
> 
> * the Actions "[Create or update an organization secret](https://docs.github.com/en/rest/actions/secrets#create-or-update-an-organization-secret)" API only has documented support for an array of integers (although in practice it actually accepts strings or integers). (Until recently, the docs said it should be an array of strings.)
> * the Dependabot "[Create or update an organization secret](https://docs.github.com/en/rest/dependabot/secrets#create-or-update-an-organization-secret)" API only supports an array of strings

If I'm reading this correctly, then making them both use an array of strings should work.
However, when we tried this here: https://github.com/google/go-github/pull/2401/commits/e3028fbc563b35173008debb2bcabd9eddf63c83
it failed... see comment from @lneves75 above: https://github.com/google/go-github/pull/2401#issuecomment-1168648355

So then we tried splitting it into two separate implementations, as you see this PR now.

> Sorry for the pain! This is obviously a bit weird at the moment and is something we need to clean up - but in a way that minimises breaking changes for users.

No problem... we would just like to get it working, so if you have recommendations on how to proceed, that would be great.
We don't mind updating the repo when/if it changes... we are updating this repo all the time.  :smile:

timrogers:
@gmlewis So sorry - you’re right, I showed too much confidence and I actually misread the internal PR 😅: the Actions API really does only accept integers, whereas before it claimed incorrectly in the docs to accept strings.

Looking at the inputs, integers seem like the right data type, so I will raise an issue with the Dependabot team to look at making this more consistent in a non-breaking way.  

Very glad I came by this repo to discover this and see all of the great work you’re doing to make the API accessible to Go developers ❤️

timrogers:
In other words, it looks like you’re doing exactly what you need to do in this PR!

gmlewis:
> Very glad I came by this repo to discover this and see all of the great work you’re doing to make the API accessible to Go developers

Thank you so much, @timrogers !

We really appreciate your input, and we always love to hear from GitHub team members and tech support whenever we can.  :smile:

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

PR #2403, opened 2022-07-03
Include the version of `go-github` in `User-Agent` headers sent to the GitHub API

timrogers:
👋🏻 Hi there! I'm Tim, and I'm a Product Manager at GitHub. To allow us to better understand how people are using our API, it's super helpful to know what versions of SDKs like `go-github` they are using. 

At the moment, this package sends the `User-Agent` "go-github" in requests, but it doesn't say what version is being used.

This updates the `User-Agent` header to include that information, whilst still maintaining the ability for people to choose their own user agent by setting the `UserAgent` field on `Client`.

It also exports a `Version` constant, so people can use that information - for example to construct custom `User-Agent` headers.

This will mean that someone has to update the `Version` constant with the new version when it is incremented.

This is the first ever Go I've written, so I won't be upset if you tell me that I've done this in a terrible way 😉

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,3 +127,5 @@ this][modified-comment].
 [git-aliases]: https://github.com/willnorris/dotfiles/blob/d640d010c23b1116bdb3d4dc12088ed26120d87d/git/.gitconfig#L13-L15
 [rebase-comment]: https://github.com/google/go-github/pull/277#issuecomment-183035491
 [modified-comment]: https://github.com/google/go-github/pull/280#issuecomment-184859046
+
+**When creating a release, don't forget to update the `Version` constant in `github.go`.** This is used to send the version in the `User-Agent` header to identify clients to the GitHub API.
--- a/github/github.go
+++ b/github/github.go
@@ -30,9 +30,9 @@ import (
 const (
 	packageVersion = "45.2.0"
 
-	defaultBaseURL = "https://api.github.com/"
-	uploadBaseURL  = "https://uploads.github.com/"
-	userAgent      = "go-github" + "/" + packageVersion
+	defaultBaseURL   = "https://api.github.com/"
+	uploadBaseURL    = "https://uploads.github.com/"
+	defaultUserAgent = "go-github" + "/" + packageVersion
 
 	headerRateLimit     = "X-RateLimit-Limit"
 	headerRateRemaining = "X-RateLimit-Remaining"
@@ -305,7 +305,7 @@ func NewClient(httpClient *http.Client) *Client {
 	baseURL, _ := url.Parse(defaultBaseURL)
 	uploadURL, _ := url.Parse(uploadBaseURL)
 
-	c := &Client{client: httpClient, BaseURL: baseURL, PackageVersion: packageVersion, UserAgent: userAgent, UploadURL: uploadURL}
+	c := &Client{client: httpClient, BaseURL: baseURL, PackageVersion: packageVersion, UserAgent: defaultUserAgent, UploadURL: uploadURL}
 	c.common.client = c
 	c.Actions = (*ActionsService)(&c.common)
 	c.Activity = (*ActivityService)(&c.common)
--- a/github/github.go
+++ b/github/github.go
@@ -168,8 +168,6 @@ type Client struct {
 	// User agent used when communicating with the GitHub API.
 	UserAgent string
 
-	PackageVersion string
-
 	rateMu     sync.Mutex
 	rateLimits [categories]Rate // Rate limits for the client as determined by the most recent API calls.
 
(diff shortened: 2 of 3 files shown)


github/github.go
```diff
@@ -28,9 +28,11 @@ import (
 )
 
 const (
+	packageVersion = "45.2.0"
```
timrogers, line 31: I couldn't find this defined anywhere else in the package code, so I added it here. I'm happy to move it elsewhere, or access it in some other way if there are good options out there.

github/github.go
```diff
@@ -28,9 +28,11 @@ import (
...
+
 	defaultBaseURL = "https://api.github.com/"
 	uploadBaseURL  = "https://uploads.github.com/"
-	userAgent      = "go-github"
+	userAgent      = "go-github" + "/" + packageVersion
```
timrogers, line 35: What do you think of renaming this to `defaultUserAgent`, which is more representative of what it is?

github/github.go
```diff
@@ -167,6 +169,8 @@ type Client struct {
 	// User agent used when communicating with the GitHub API.
 	UserAgent string
 
+	PackageVersion string
```
timrogers, line 172: I thought that exposing this on the `Client` would help with (a) people constructing customer `User-Agent` headers and (b) me testing the default `User-Agent`!

github/github_test.go
```diff
@@ -507,10 +507,16 @@ func TestNewRequest(t *testing.T) {
...
 
+	userAgent := req.Header.Get("User-Agent")
+
 	// test that default user-agent is attached to the request
-	if got, want := req.Header.Get("User-Agent"), c.UserAgent; got != want {
+	if got, want := userAgent, c.UserAgent; got != want {
 		t.Errorf("NewRequest() User-Agent is %v, want %v", got, want)
 	}
+
+	if !strings.Contains(userAgent, c.PackageVersion) {
```
timrogers, line 517: Is there somewhere else I should add a test to help confirm that this is being added to _all_ requests? Or do you think this is sufficient?

github/github.go
```diff
@@ -301,7 +305,7 @@ func NewClient(httpClient *http.Client) *Client {
 	baseURL, _ := url.Parse(defaultBaseURL)
 	uploadURL, _ := url.Parse(uploadBaseURL)
 
-	c := &Client{client: httpClient, BaseURL: baseURL, UserAgent: userAgent, UploadURL: uploadURL}
+	c := &Client{client: httpClient, BaseURL: baseURL, PackageVersion: packageVersion, UserAgent: userAgent, UploadURL: uploadURL}
```
gmlewis, line 308: So this would no longer be needed since the `PackageVersion` constant is now exported and available.

github/github_test.go
gmlewis: I think this is sufficient, thanks.

gmlewis (changes requested):
Great idea, @timrogers !
Just a couple minor tweaks, please, then we should be ready for a second LGTM+Approval before merging.

github/github.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 const (
-	defaultBaseURL = "https://api.github.com/"
-	uploadBaseURL  = "https://uploads.github.com/"
-	userAgent      = "go-github"
+	CurrentVersion   = "45.2.0"
```
timrogers, line 31: Any ideas of where I can document this to make sure it gets updated on release? I guess the "Maintainer's Guide" in `CONTRIBUTING.md` could be a good place.

timrogers (commented):
Thanks for the feedback! I've addressed those comments and I think it's looking better now. Plus I learned about exported constants, which is great 🧠 ⬆️

github/github.go
gmlewis: How about we just change this to "Version" since that is pretty common?

gmlewis: Oh, and sure... feel free to add a comment to CONTRIBUTING.md, although I don't think we have any written-up instructions for maintainers yet.  Maybe I should do that sometime.  :smile:

gmlewis: Feel free to add a blank line after `Version` to uncouple it from the following unexported consts.

timrogers: Done! 🏁

gmlewis (approved):
Thank you, @timrogers !
LGTM.

Awaiting second LGTM+Approval from any other contributor to this report before merging.

CONTRIBUTING.md
```diff
@@ -127,3 +127,5 @@ this][modified-comment].
 [git-aliases]: https://github.com/willnorris/dotfiles/blob/d640d010c23b1116bdb3d4dc12088ed26120d87d/git/.gitconfig#L13-L15
 [rebase-comment]: https://github.com/google/go-github/pull/277#issuecomment-183035491
 [modified-comment]: https://github.com/google/go-github/pull/280#issuecomment-184859046
+
+**When creating a release, don't forget to update the `Version` constant in `github.go`.** This is used to send the version in the `User-Agent` header to identify clients to the GitHub API.
```
raynigon, line 131: This could be automated. Since the version constant is currently only used for the user agent, an error should not be a big problem, but if somebody depends on the constant this could lead to problems along the way.
Unfortunatly the release process in this project is not automated yet. If it gets automated in the future we need to include this constant in there.

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

Issue #2405, opened 2022-07-05
List CODEOWNERS errors


cbrgm:
Hi there,

I was wondering whether this endpoint is available in the google/go-github client.
https://docs.github.com/en/rest/repos/repos#list-codeowners-errors

If it's not the case, I'd be happy to contribute!

gmlewis:
> If it's not the case, I'd be happy to contribute!

No, it doesn't appear to be.  Thank you, @cbrgm !  It's all yours.

cbrgm:
@gmlewis It took me a while, but I finally found some time preparing a PR. I'm looking forward to a review sometime 😄 Happy weekend!

PR #2408, opened 2022-07-08
feat: add GetCodeownersErrors to RepositoriesService (#2405)

cbrgm:
This PR adds a new function `GetCodeownersErrors` to the `RepositoriesService` struct.
It adds support for the endpoint https://docs.github.com/en/rest/repos/repos#list-codeowners-errors as described in #2405

### DIFF
--- a/github/repos_codeowners.go
+++ b/github/repos_codeowners.go
@@ -1,0 +1,41 @@
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// CodeownersErrors represents a list of syntax errors detected in the CODEOWNERS file.
+type CodeownersErrors struct {
+	Errors []*CodeownersError `json:"errors"`
+}
+
+// CodeownersError represents a syntax error detected in the CODEOWNERS file.
+type CodeownersError struct {
+	Line       *int    `json:"line"`
+	Column     *int    `json:"column"`
+	Kind       *string `json:"kind"`
+	Source     *string `json:"source"`
+	Suggestion *string `json:"suggestion"`
+	Message    *string `json:"message"`
+	Path       *string `json:"path"`
+}
+
+// GetCodeownersErrors lists any syntax errors that are detected in the CODEOWNERS file.
+//
+// GitHub API docs: https://docs.github.com/en/rest/repos/repos#list-codeowners-errors
+func (s *RepositoriesService) GetCodeownersErrors(ctx context.Context, owner, repo string) (*CodeownersErrors, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/codeowners/errors", owner, repo)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	codeownersErrors := &CodeownersErrors{}
+	resp, err := s.client.Do(ctx, req, codeownersErrors)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


github/repos_codeowners.go
```diff
@@ -0,0 +1,41 @@
+package github
```
gmlewis, line 1: Please add the standard copyright header with the year 2022 here before the `package` line with a blank line separating the two.

github/repos_codeowners.go
```diff
@@ -0,0 +1,41 @@
...
+	Errors []*CodeownersError `json:"errors"`
+}
+
+// CodeownersError represents a syntax error detected in the CODEOWNERS file.
+type CodeownersError struct {
+	Line       *int    `json:"line"`
+	Column     *int    `json:"column"`
+	Kind       *string `json:"kind"`
+	Source     *string `json:"source"`
+	Suggestion *string `json:"suggestion"`
```
gmlewis, line 19:
```suggestion
	Suggestion *string `json:"suggestion,omitempty"`
```

github/repos_codeowners.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+)
+
+// CodeownersErrors represents a list of syntax errors detected in the CODEOWNERS file.
+type CodeownersErrors struct {
+	Errors []*CodeownersError `json:"errors"`
+}
+
+// CodeownersError represents a syntax error detected in the CODEOWNERS file.
+type CodeownersError struct {
```
gmlewis, line 14: Looking at the official docs for this data, it appears that all the fields will be returned except the `suggestion` field, so I'm thinking that we might want to break from the surrounding practice in this repo and make the rest of these fields values instead of pointers.

github/repos_codeowners_test.go
```diff
@@ -0,0 +1,105 @@
+package github
```
gmlewis, line 1: Same here, please.

gmlewis (changes requested):
This is looking great, @cbrgm - thank you!
Just a few minor tweaks, please, before we have a second reviewer approval.

github/repos_codeowners.go
cbrgm: Done! Thanks, forgot about that.

github/repos_codeowners_test.go
cbrgm: 👍

github/repos_codeowners.go
cbrgm: Changed that. Do you think
```go
// CodeownersErrors represents a list of syntax errors detected in the CODEOWNERS file.
type CodeownersErrors struct {
	Errors []*CodeownersError `json:"errors"`
}
```
should also be a slice of values instead of a slice of pointers?

gmlewis: No, I think that is fine because when we have slices of structs, we typically use pointers to reduce the chance of errors when iterating over the items... so let's leave this one.  Thanks!

gmlewis (approved):
Thank you, @cbrgm !
LGTM.

Awaiting LGTM+Approval from any other contributor to this repo before merging.

spikeb4 (approved):

gmlewis:
Thank you, @spikeb4 !
Merging.

Issue #2421, opened 2022-07-27
Support actions/runs/RUN_ID/pending_deployments endpoint


DavidGamba:
Add support for https://docs.github.com/en/rest/actions/workflow-runs#review-pending-deployments-for-a-workflow-run

PR in progress.

PR #2422, opened 2022-07-27
Add 'pending_deployments' endpoint support (#2421)

DavidGamba:


github/actions_workflow_runs.go
```diff
@@ -94,6 +94,13 @@ type WorkflowRunAttemptOptions struct {
 	ExcludePullRequests *bool `url:"exclude_pull_requests,omitempty"`
 }
 
+// PendingDeploymentsRequest specificies body parameters to PendingDeployments
+type PendingDeploymentsRequest struct {
+	EnvironmentIDs *[]int  `json:"environment_ids"`
```
DavidGamba, line 99: I am passing integers here directly, I wonder if this needs to be wired to some other type.

DavidGamba:
CLA signed.

github/actions_workflow_runs.go
gmlewis: Yes, IDs are typically `int64`, so please change this to `[]int64` and remove the leading asterisk.

DavidGamba: Thanks! fixed.

github/actions_workflow_runs.go
```diff
@@ -94,6 +94,13 @@ type WorkflowRunAttemptOptions struct {
 	ExcludePullRequests *bool `url:"exclude_pull_requests,omitempty"`
 }
 
+// PendingDeploymentsRequest specificies body parameters to PendingDeployments
+type PendingDeploymentsRequest struct {
+	EnvironmentIDs []int64 `json:"environment_ids"`
+	State          *string `json:"state"`
+	Comment        *string `json:"comment"`
```
gmlewis, line 101: Since these two last fields are required, let's please change them to `string` (instead of `*string`).

github/actions_workflow_runs.go
```diff
@@ -94,6 +94,13 @@ type WorkflowRunAttemptOptions struct {
 	ExcludePullRequests *bool `url:"exclude_pull_requests,omitempty"`
 }
 
+// PendingDeploymentsRequest specificies body parameters to PendingDeployments
```
gmlewis, line 97:
```suggestion
// PendingDeploymentsRequest specifies body parameters to PendingDeployments.
```

github/actions_workflow_runs.go
```diff
@@ -94,6 +94,13 @@ type WorkflowRunAttemptOptions struct {
...
+type PendingDeploymentsRequest struct {
+	EnvironmentIDs []int64 `json:"environment_ids"`
+	State          *string `json:"state"`
```
gmlewis, line 100: Let's please add a comment to this field:

```suggestion
	// State can be one of: "approved", "rejected".
	State          string `json:"state"`
```

gmlewis (changes requested):
Just a few minor tweaks, please @DavidGamba.

DavidGamba:
@gmlewis PTAL

gmlewis (approved):
Thank you, @DavidGamba !
LGTM.

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

PR #2430, opened 2022-08-02
Remove omitempty from Content field of RepositoryContentFileOptions

### LINKED ISSUE #2427: Unable to create empty file
When attempting to commit an empty file to a repository using `ghClient.Repositories.CreateFile` and passing in `[]byte{}` under the `Content` field of a `github.RepositoryContentFileOptions` the GitHub API will return a `422` error stating that `content` was not supplied. This is most likely due to the line:

https://github.com/google/go-github/blob/7a7a7f0943f3fcffc4b8f30feaaed7333b035a99/github/repos_contents.go#L54

Which adds `omitempty` to the `Content` field of `github.RepositoryContentFileOptions`, even though it is a necessary field of the request.

Ashvin-Ranjan:
fixes #2427

This removes `omitempty` from `RepositoryContentFileOptions.Content` so that an empty `content` field can be passed in to generate empty files. Before, the `content` field was omitted, which caused the GitHub API to not accept the request as the field is required, even if it is empty.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -51,7 +51,7 @@ type RepositoryContentResponse struct {
 // RepositoryContentFileOptions specifies optional parameters for CreateFile, UpdateFile, and DeleteFile.
 type RepositoryContentFileOptions struct {
 	Message   *string       `json:"message,omitempty"`
-	Content   []byte        `json:"content,omitempty"` // unencoded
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @Ashvin-Ranjan !
LGTM.
Merging.

PR #2440, opened 2022-08-20
Add parameters to configure merge commit messages to repo

garrettheel:
These new parameters have been added to the [repository api](https://docs.github.com/en/rest/repos/repos#update-a-repository) to allow more granular control of merge commit messages:

 * `squash_merge_commit_title`
 * `squash_merge_commit_message`
 * `merge_commit_title`
 * `merge_commit_message`


gmlewis (approved):
Thank you, @garrettheel !
LGTM.
Merging.

gmlewis:
Looking at the linked docs, these 4 new fields have a small number of allowed values...
I think it might be beneficial to include them as comments, such as:

```go
// SquashMergeCommitTitle can be one of: "PR_TITLE", "COMMIT_OR_PR_TITLE"
```

Thoughts?

garrettheel:
> Looking at the linked docs, these 4 new fields have a small number of allowed values... I think it might be beneficial to include them as comments

@gmlewis yep, that makes sense! I wasn't sure how much we wanted to re-document the API here, but I think this would help quickly figure out valid inputs. Pushed up another commit with comments, let me know if that's the right spot for them

PR #2445, opened 2022-08-24
Remove GenerateReleaseNotes from RepositoriesService.EditRelease

### LINKED ISSUE #2444: GenerateReleaseNotes will have no effect in RepositoriesService.EditRelease
According to the [GitHub v3 API reference](https://docs.github.com/en/rest/releases/releases#update-a-release), `generate_release_notes` is not a recognised body parameter of the endpoint to update a release. In this library, the value of 
`RepositoryRelease.GenerateReleaseNotes` [is passed to the PATCH request to the GitHub API](https://github.com/google/go-github/blob/243bda850b1f78957ba0d50fb62ecdb4fbe977d5/github/repos_releases.go#L223-L232) regardless, where it will have no effect.

From what I can tell, `RepositoryRelease.GenerateReleaseNotes` will only have an effect when calling `RepositoriesService.CreateRelease`, whereas [the comments imply](https://github.com/google/go-github/blob/243bda850b1f78957ba0d50fb62ecdb4fbe977d5/github/repos_releases.go#L31) that it has an effect everywhere.

I recognise this field can't have an effect until there is a change in the GitHub API; I think what I would prefer in the meantime would be an update to the code/documentation in this library to make it clear that `RepositoriesService.UpdateRelease` will not cause release note generation, as this was rather confusing to me when trying to consume this function in a downstream library.

coro:
This PR closes #2444 

According to the [GitHub v3 API reference](https://docs.github.com/en/rest/releases/releases#update-a-release),
`generate_release_notes` is not a recognised body parameter of the endpoint to update a release. If provided to
the API, it will have no effect.

This commit removes the setting of this parameter when updating a release, to avoid confusion in the effect of
setting this parameter.

### DIFF
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -26,7 +26,9 @@ type RepositoryRelease struct {
 	Draft                  *bool   `json:"draft,omitempty"`
 	Prerelease             *bool   `json:"prerelease,omitempty"`
 	DiscussionCategoryName *string `json:"discussion_category_name,omitempty"`
-	GenerateReleaseNotes   *bool   `json:"generate_release_notes,omitempty"`
+
+	// The following fields are not used in EditRelease:
+	GenerateReleaseNotes *bool `json:"generate_release_notes,omitempty"`
 
 	// The following fields are not used in CreateRelease or EditRelease:
 	ID          *int64          `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @coro !
LGTM.

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

PR #2450, opened 2022-09-01
Add alphanumeric param to AutolinkOptions

### LINKED ISSUE #2449: Support Autolinks with alphanumeric or numeric IDs
GitHub updated their [autolink feature](https://github.blog/changelog/2022-08-31-autolinks-with-alphanumeric-or-numeric-ids/) to support either alphanumeric or numeric IDs.

This is a request to add support for the new `is_alphanumeric: boolean` argument, which indicates whether this autolink reference matches alphanumeric characters. If false, this autolink reference is a legacy autolink that only matches numeric characters.

Example Payload:
```json
{
  "id": 1,
  "key_prefix": "TICKET-",
  "url_template": "https://example.com/TICKET?query=<num>",
  "is_alphanumeric": true
}
```

---

I'm interested in working on adding this feature. I'm a complete `go` novice, so I'm thankful 99% of the work for calling this API is already done 😅 

Thanks!

douglascayers:
### Summary

Resolves #2449

### Changes

* Adds `IsAlphanumeric` param to `AutolinkOptions` and `Autolink` structs of `github/repos_autolinks.go`
* Adds unit and integration tests for creating and getting autolinks for a repo
* Updates the `createRandomTestRepository` function in `test/integration/github_test.go` to look up the owner to create the repo under if a name isn't specified explicitly. Checks `GITHUB_OWNER` env var else falls back to user the oauth token belongs to.
* Removes redundant env var check from `init` function of `test/integration/github_test.go` because those checks already occur in the `test/integration/authorizations_test.go` file as needed. This made it easier for me to run integrations tests without setting env vars that weren't needed.

### Tests

* `github/repos_autolinks_test.go`
* `test/integration/repos_test.go`

#### Created an alphanumeric autolink

image

#### Created a numeric autolink

image

### DIFF
--- a/github/repos_autolinks.go
+++ b/github/repos_autolinks.go
@@ -12,15 +12,17 @@ import (
 
 // AutolinkOptions specifies parameters for RepositoriesService.AddAutolink method.
 type AutolinkOptions struct {
-	KeyPrefix   *string `json:"key_prefix,omitempty"`
-	URLTemplate *string `json:"url_template,omitempty"`
+	KeyPrefix      *string `json:"key_prefix,omitempty"`
+	URLTemplate    *string `json:"url_template,omitempty"`
+	IsAlphanumeric *bool   `json:"is_alphanumeric,omitempty"`
 }
 
 // Autolink represents autolinks to external resources like JIRA issues and Zendesk tickets.
 type Autolink struct {
-	ID          *int64  `json:"id,omitempty"`
-	KeyPrefix   *string `json:"key_prefix,omitempty"`
-	URLTemplate *string `json:"url_template,omitempty"`
+	ID             *int64  `json:"id,omitempty"`
+	KeyPrefix      *string `json:"key_prefix,omitempty"`
+	URLTemplate    *string `json:"url_template,omitempty"`
+	IsAlphanumeric *bool   `json:"is_alphanumeric,omitempty"`
 }
 
 // ListAutolinks returns a list of autolinks configured for the given repository.
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -1374,6 +1374,14 @@ func (a *Autolink) GetID() int64 {
 	return *a.ID
 }
 
+// GetIsAlphanumeric returns the IsAlphanumeric field if it's non-nil, zero value otherwise.
+func (a *Autolink) GetIsAlphanumeric() bool {
+	if a == nil || a.IsAlphanumeric == nil {
+		return false
+	}
+	return *a.IsAlphanumeric
+}
+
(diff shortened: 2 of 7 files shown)


gmlewis (approved):
Thank you very much, @douglascayers !
Looks great to me.

douglascayers:
Hi @alexal, as a contributor to the autolinks feature, I'd appreciate your review of my proposed changes. Thanks!

github/repos_autolinks.go
```diff
@@ -12,15 +12,17 @@ import (
...
 
 // Autolink represents autolinks to external resources like JIRA issues and Zendesk tickets.
 type Autolink struct {
-	ID          *int64  `json:"id,omitempty"`
-	KeyPrefix   *string `json:"key_prefix,omitempty"`
-	URLTemplate *string `json:"url_template,omitempty"`
+	ID             *int64  `json:"id,omitempty"`
+	KeyPrefix      *string `json:"key_prefix,omitempty"`
+	URLTemplate    *string `json:"url_template,omitempty"`
+	IsAlphanumeric *bool   `json:"is_alphanumeric,omitempty"`
```
douglascayers, line 25: @gmlewis curious, what do these \`json:"<fieldName>,omitempty"\` mean? Does golang parse them or is it more like a comment for devs to know the intent of the struct property? 🤔

gmlewis: > @gmlewis curious, what do these `json:"<fieldName>,omitempty"` mean? Does golang parse them or is it more like a comment for devs to know the intent of the struct property? thinking

Just so I don't lead you astray, according to the [official docs](https://pkg.go.dev/encoding/json#Marshal):

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

douglascayers: Thanks, the piece that connected it for me in that doc was:

> The encoding of each struct field can be customized by the format string stored under the "json" key in the struct field's tag. The format string gives the name of the field, possibly followed by a comma-separated list of options. The name may be empty in order to specify options without overriding the default field name.

I then googled "struct field tag" and came across https://www.digitalocean.com/community/tutorials/how-to-use-struct-tags-in-go which said:

> Struct tags are small pieces of metadata attached to fields of a struct that provide instructions to other Go code that works with the struct.

I now know that those _comment_ looking things on struct fields are **struct tags** and packages may parse them to influence behavior.

Thanks!

raynigon (approved):

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

PR #2452, opened 2022-09-06
InstallationPermissions: organization perms

thepwagner:
Adds `organization_custom_roles` and `organization_packages` as available permissions on the `InstallPermissions` struct, per https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app

I've functionally tested the `organization_custom_roles` bit, but I'm not sure how to verify `organization_packages`. It is documented online and in other libraries like octokit: https://github.com/search?q=org%3Aoctokit+organization_packages&type=code , but I can't add it to a test GitHub.com app. It may only be available on GitHub Enterprise? 🤷


gmlewis (approved):
Thank you, @thepwagner !
LGTM.
Merging.

PR #2454, opened 2022-09-08
Escape special characters in ref name for update ref

### LINKED ISSUE #2453: Update ref doesn't escape special characters in ref name
The `UpdateRef` function doesn't substitute special characters for their escaped versions within ref names before making a patch request. This consistently leads to invalid paths if the ref has a special character (ex: #).

Planning to fix.

shravan1k:
Closes #2453 

Most of the functions that deal with refs make sure to escape the special characters within the ref's name. The `UpdateRef` function doesn't seem to call the function that does this substitution of special characters with their encodings. This PR solves this problem and should now successfully patch remote refs if the `GetRef` function says the corresponding ref exists.

### DIFF
--- a/github/git_refs.go
+++ b/github/git_refs.go
@@ -142,7 +142,7 @@ func (s *GitService) CreateRef(ctx context.Context, owner string, repo string, r
 // GitHub API docs: https://docs.github.com/en/rest/git/refs#update-a-reference
 func (s *GitService) UpdateRef(ctx context.Context, owner string, repo string, ref *Reference, force bool) (*Reference, *Response, error) {
 	refPath := strings.TrimPrefix(*ref.Ref, "refs/")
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @shravan1k !
Please add a unit test that demonstrates this change in functionality.

gmlewis (changes requested):
Whups - checked the wrong box.

Thank you, @shravan1k !
Please add a unit test that demonstrates this change in functionality.

gmlewis (approved):
Beautiful, @shravan1k !  Thank you!
LGTM.

marcelosousa (approved):
LGTM 👍

gmlewis:
Thank you, @marcelosousa !
Merging.

PR #2460, opened 2022-09-13
Fix unhandled error in actions_artifacts.go

mo1ein:


gmlewis (changes requested):
Thank you, @mo1ein !
Could you please add a unit test that exercises this new code path?

gmlewis:
It also looks like you need to run `gofmt` on the edited file to fix the formatting.  Please see our CONTRIBUTING.md guide for more information.

mo1ein:
> It also looks like you need to run `gofmt` on the edited file to fix the formatting. Please see our CONTRIBUTING.md guide for more information.

Always after writing code, I run `gofmt` but for this file, I can't see 8 spaces after `gofmt`. That's the reason of my second commit.

gmlewis:
> > It also looks like you need to run `gofmt` on the edited file to fix the formatting. Please see our CONTRIBUTING.md guide for more information.
> 
> Always after writing code, I run `gofmt` but for this file, I can't see 8 spaces after `gofmt`. That's the reason of my second commit.

`gofmt` uses tab characters and not spaces. Maybe that explains the difference?

mo1ein:
> > > It also looks like you need to run `gofmt` on the edited file to fix the formatting. Please see our CONTRIBUTING.md guide for more information.
> > 
> > 
> > Always after writing code, I run `gofmt` but for this file, I can't see 8 spaces after `gofmt`. That's the reason of my second commit.
> 
> `gofmt` uses tab characters and not spaces. Maybe that explains the difference?
OMG! yeah. Thank's for your clarifying.

gmlewis:
> yeah. Thank's for your clarifying.

Cool!

Could you please add a unit test that exercises this new code path?

mo1ein:
> > yeah. Thank's for your clarifying.
> 
> Cool!
> 
> Could you please add a unit test that exercises this new code path?

I can't figure out how to fix `Added lines #L125 - L126 were not covered by tests`. Can you help?

gmlewis:
> I can't figure out how to fix `Added lines #L125 - L126 were not covered by tests`. Can you help?

I believe that if you give a "Location" value that `url.Parse` will return an error for, then that should test the line.  Maybe something like "https://\n/bogus\nurl" or something like that?

mo1ein:
> > I can't figure out how to fix `Added lines #L125 - L126 were not covered by tests`. Can you help?
> 
> I believe that if you give a "Location" value that `url.Parse` will return an error for, then that should test the line. Maybe something like "https://\n/bogus\nurl" or something like that?

I added the test but it has a bug. I think the test idea is correct. Just need to produce `URL parse error`. I believe your review is helpful to fix bugs.

gmlewis:
> I added the test but it has a bug. I think the test idea is correct. Just need to produce `URL parse error`. I believe your review is helpful to fix bugs.

Apparently, according to the source code for `url.Parse`, if you put control characters in your URL, this function should return an error:

https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/net/url/url.go;l=503-505;drc=d36466fe2aaf51c4b38c3a2ea3164cb3a56b8059

mo1ein:
- [x] Fixed. I think everything is OK. :-)

github/actions_artifacts.go
```diff
@@ -121,6 +121,10 @@ func (s *ActionsService) DownloadArtifact(ctx context.Context, owner, repo strin
 	}
 
 	parsedURL, err := url.Parse(resp.Header.Get("Location"))
+	if err != nil {
+		return nil, nil, err
```
gmlewis, line 125: Please return a copy of the previous response here, as in line 120 above.

```suggestion
		return nil, newResponse(resp), err
```

gmlewis (changes requested):

github/actions_artifacts.go
mo1ein: - [x] Done.

gmlewis (approved):
Thank you, @mo1ein !
LGTM.
Merging.

PR #2461, opened 2022-09-14
Fix variables name

mo1ein:


gmlewis:
Thank you, @mo1ein - but I honestly see no value in reducing the length of these variable names any further.
I know idiomatic Go uses short variable names, but this is especially true for iterators, and less true for important values.
The names are already short and descriptive and make it easy to see at a quick glance what they are for.
Reducing their length further makes the code harder to read in my opinion, which is against the Go philosophy.

Closing.

mo1ein:
> Thank you, @mo1ein - but I honestly see no value in reducing the length of these variable names any further. I know idiomatic Go uses short variable names, but this is especially true for iterators, and less true for important values. The names are already short and descriptive and make it easy to see at a quick glance what they are for. Reducing their length further makes the code harder to read in my opinion, which is against the Go philosophy.
> 
> Closing.

That's true but when I read the code, I saw somewhere is used short and somewhere is not. 
For example watch this line: https://github.com/google/go-github/blob/master/github/repos.go#L444
And this line: https://github.com/google/go-github/blob/master/github/repos.go#L505

gmlewis:
Yes, you are right that this repo is not consistent in its use of variable names, and that is my fault.

However, for the case of `r := new(Respository)` vs `repository := new(Repository)`, I would probably argue that neither are actually ideal.  Maybe a happy medium of `repo := &Repository{}` would be my preference.

But having said that, I'm not sure that a PR solely dedicated to changing variable names is of great value, just for the sake of renaming. If other value is being added, then we can possibly discuss cleaning up variable names.

As it stands, I don't think the names are actually poorly chosen... they are mostly personal preferences of the individual contributors and I have not been very strict about codifying any variable naming guidelines, and honestly am not sure it is worth the effort, as I think the code is easily readable and understandable as-is.

PR #2468, opened 2022-09-17
Always send 'checks' field when creating branch protection

### LINKED ISSUE #2467: Error when creating branch protection
Since we updated our code to use the latest version of this library, we can't create a branch protection that requires branches to be up to date before merging but don't require any status checks:

```go
func (r *RepoSettings) protectionRequestOptions() *github.ProtectionRequest {
	return &github.ProtectionRequest{
		EnforceAdmins: r.EnforceAdmins,
			DismissStaleReviews:          r.DismissStaleReviews,
			RequiredApprovingReviewCount: r.RequiredApprovingReviewCount,
			RequireCodeOwnerReviews:      r.RequireCodeOwnerReviews,
		},
		RequiredStatusChecks: &github.RequiredStatusChecks{
			Strict: true,
		},
	}
}
```

The above results in an error:

```
PUT https://api.github.com/repos/****/****/branches/main/protection: 422 Invalid request.

No subschema in "oneOf" matched.
For 'anyOf/1', {"strict"=>true} is not a null. []
```

Such rule is possible to be created from the UI:
image

This is also affecting the Github Terraform provider: https://github.com/integrations/terraform-provider-github/issues/1147

luisdavim:
Fixes #2467

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -905,7 +905,7 @@ type RequiredStatusChecks struct {
 	Contexts []string `json:"contexts,omitempty"`
 	// The list of status checks to require in order to merge into this
 	// branch.
-	Checks []*RequiredStatusCheck `json:"checks,omitempty"`
+	Checks []*RequiredStatusCheck `json:"checks"`
 }
 
 // RequiredStatusChecksRequest represents a request to edit a protected branch's status checks.
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -1543,6 +1543,130 @@ func TestRepositoriesService_UpdateBranchProtection_Checks(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


gmlewis (changes requested):
Thank you, @luisdavim .
Could you please add a unit test demonstrating that this field is indeed sent when empty as a means of documenting that we explicitly wish to send it in all cases?

luisdavim:
@gmlewis I've added a new test case for this scenario, where we set strict branch protection checks with an empty set of checks but, whilst looking at the tests, I found some inconsistency between the inputs and the mocks of the existing tests so I also fixed those in the same commit, I hope that's ok.

gmlewis (approved):
Thank you, @luisdavim !
LGTM.

raynigon (approved):
LGTM

gmlewis:
Thank you, @raynigon !
Merging.

PR #2469, opened 2022-09-17
Add explicit error return

mo1ein:


github/repos_contents.go
```diff
@@ -317,5 +317,8 @@ func (s *RepositoriesService) GetArchiveLink(ctx context.Context, owner, repo st
 	}
 
 	parsedURL, err := url.Parse(resp.Header.Get("Location"))
-	return parsedURL, newResponse(resp), err
+	if err != nil {
+		return nil, newResponse(resp), err
+	}
+	return parsedURL, newResponse(resp), nil
```
gmlewis, line 323: I know this is another thing that is inconsistent across this repo, but while I'm thinking about it, let's go ahead and add a blank line between the error handling and the final return, please.

```suggestion
	}

	return parsedURL, newResponse(resp), nil
```

gmlewis:
Note that I'm going to change the title of this PR because it is technically not an "unhandled error"... it is simply an error that is not explicitly tested for.

github/repos_contents.go
mo1ein: - [x] Done.

gmlewis (approved):
Thank you, @mo1ein !
LGTM.
Merging.

Issue #2338, opened 2022-04-20
How are you supposed to access the output of ListSCIMProvisionedIdentities?


Osipion:
Hi all. Thanks so much for your work on this. QQ: `ListSCIMProvisionedIdentities` appears to not return any entities and also closes the underlying HTTP response body reader. What's the expected way to read the list of identities returned by the API?

gmlewis:
Hi @Osipion , and thank you for the report!

This is totally my fault... I let #2062 go through code review without returning a populated struct. I apologize.

This needs to be implemented... would you like to submit a PR, or shall I open this up to any other contributor to this repo?

Osipion:
Thanks for getting back to me. I may have a chance to look into submitting a PR this weekend but don't wait up for me

gmlewis:
No rush, @Osipion - it is yours. Thank you!

PR #2474, opened 2022-09-21
Return proper responses for SCIM provisioned identities

tenyo:
This PR adds return responses to `ListSCIMProvisionedIdentities` and `GetSCIMProvisioningInfoForUser`, as before it was not possible to get the output of those calls.
Fixes https://github.com/google/go-github/issues/2338

### DIFF
--- a/github/scim.go
+++ b/github/scim.go
@@ -112,10 +112,12 @@ func (s *SCIMService) ProvisionAndInviteSCIMUser(ctx context.Context, org string
 	if err != nil {
 		return nil, err
 	}
+
 	req, err := s.client.NewRequest("POST", u, nil)
 	if err != nil {
 		return nil, err
 	}
+
 	return s.client.Do(ctx, req, nil)
 }
 
@@ -147,10 +149,12 @@ func (s *SCIMService) UpdateProvisionedOrgMembership(ctx context.Context, org, s
 	if err != nil {
 		return nil, err
 	}
+
 	req, err := s.client.NewRequest("PUT", u, nil)
 	if err != nil {
 		return nil, err
 	}
+
 	return s.client.Do(ctx, req, nil)
 }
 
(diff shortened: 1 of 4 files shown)


github/scim.go
```diff
@@ -62,17 +82,22 @@ type ListSCIMProvisionedIdentitiesOptions struct {
...
-		return nil, err
+		return nil, nil, err
 	}
-	return s.client.Do(ctx, req, nil)
+	identities := new(SCIMProvisionedIdentities)
+	resp, err := s.client.Do(ctx, req, identities)
+	if err != nil {
+		return nil, resp, err
+	}
+	return identities, resp, nil
```
gmlewis, line 100: Let's please clean this up a bit while we're here and add a blank line before lines 100, 95, and 91.

gmlewis (changes requested):
Thank you, @tenyo !
Just a couple minor nits, please, and then we should be ready for a second LGTM+Approval from any other contributor to this repo.

github/scim.go
tenyo: Sure, I can do that - I actually removed the blank lines on purpose to match the formatting in the other functions.

tenyo: OK, I've fixed those, let me know if you want me to add similar blank lines in the rest of the funcs in this file, so they all match.

gmlewis: > OK, I've fixed those, let me know if you want me to add similar blank lines in the rest of the funcs in this file, so they all match.

You can if you would like to... Thank you!

tenyo: All done, much better now ☺️  
You can probably add a linter rule in `.golangci.yml` to enforce that but there might be a bunch of files to fix 😅

wesleimp (approved):

gmlewis:
Thank you, @wesleimp !
Merging.

PR #2482, opened 2022-09-28
Add support for User SSH signing keys

neersighted:
Introduce support for the new [SSH signing key endpoints](https://docs.github.com/en/rest/users/ssh-signing-keys) introduced into the REST API with the [introduction of SSH commit verification](https://github.blog/changelog/2022-08-23-ssh-commit-verification-now-supported/).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,6 +59,7 @@ Beyang Liu <beyang.liu@gmail.com>
 Billy Keyes <bluekeyes@gmail.com>
 Billy Lynch <wlynch92@gmail.com>
 Björn Häuser <b.haeuser@rebuy.de>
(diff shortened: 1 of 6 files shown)


gmlewis (approved):
Thank you, @neersighted !
LGTM.

gmlewis:
@neersighted - please use regular push of commits to this PR and not force push, as it makes it really challenging for the reviewer to see what changes each time.

We always use squash and merge in this repo anyway so it doesn't matter how many commits it takes. Thank you.

neersighted:
Fair -- I'm usually used to a "clean history" workflow -- in this case `go generate` was missed.

valbeat (approved):

gmlewis:
Thank you, @valbeat !
Merging.

PR #2504, opened 2022-10-18
Add DiscussionCommentEvent

qascade:
Hi, I have created Disscussion Comment event as per the Github API docs. Please review and suggest necessary changes. Thanks.

### DIFF
--- a/github/event.go
+++ b/github/event.go
@@ -54,6 +54,8 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 		payload = &DeploymentStatusEvent{}
 	case "DiscussionEvent":
 		payload = &DiscussionEvent{}
+	case "DiscussionCommentEvent":
+		payload = &DiscussionCommentEvent{}
 	case "ForkEvent":
 		payload = &ForkEvent{}
 	case "GitHubAppAuthorizationEvent":
--- a/github/event.go
+++ b/github/event.go
@@ -52,10 +52,10 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 		payload = &DeploymentEvent{}
 	case "DeploymentStatusEvent":
 		payload = &DeploymentStatusEvent{}
-	case "DiscussionEvent":
-		payload = &DiscussionEvent{}
 	case "DiscussionCommentEvent":
 		payload = &DiscussionCommentEvent{}
+	case "DiscussionEvent":
+		payload = &DiscussionEvent{}
 	case "ForkEvent":
 		payload = &ForkEvent{}
 	case "GitHubAppAuthorizationEvent":
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -205,6 +205,16 @@ type DiscussionEvent struct {
 	Installation *Installation `json:"installation,omitempty"`
 }
 
+// DiscussionCommentEvent represents a discussion comment event.
+type DiscussionCommentEvent struct {
+	Action     *string            `json:"action,omitempty"`
+	Comment    *DiscussionComment `json:"comment,omitempty"`
+	Discussion *Discussion        `json:"discussion,omitempty"`
+	Repo       *Repository        `json:"repository,omitempty"`
+	Org        *Organization      `json:"organization,omitempty"`
+	Sender     *User              `json:"sender,omitempty"`
+}
+
 // Discussion represents a discussion in a GitHub DiscussionEvent.
 type Discussion struct {
 	RepositoryURL      *string             `json:"repository_url,omitempty"`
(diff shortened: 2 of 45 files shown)


github/event.go
```diff
@@ -54,6 +54,8 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 		payload = &DeploymentStatusEvent{}
 	case "DiscussionEvent":
 		payload = &DiscussionEvent{}
+	case "DiscussionCommentEvent":
```
gmlewis, line 57: Since "DiscussionCommentEvent" is lexicographically less than "DiscussionEvent", let's please move these two lines before line 55 above.

github/event_types.go
```diff
@@ -228,6 +228,16 @@ type Discussion struct {
 	Body               *string             `json:"body,omitempty"`
 }
 
+// DiscussionCommentEvent represents discussion comment event
```
gmlewis, line 231:
```suggestion
// DiscussionCommentEvent represents a discussion comment event.
```

github/event_types.go
```diff
@@ -228,6 +228,16 @@ type Discussion struct {
...
+type DiscussionCommentEvent struct {
+	Action     *string            `json:"action,omitempty"`
+	Comment    *DiscussionComment `json:"comment,omitempty"`
+	Discussion *Discussion        `json:"discussion,omitempty"`
+	Repo       *Repository        `json:"repository,omitempty"`
+	Org        *Organization      `json:"organization,omitempty"`
+	Sender     *User              `json:"user,omitempty"`
```
gmlewis, line 238:
```suggestion
	User     *User              `json:"user,omitempty"`
```

gmlewis (changes requested):
Thank you, @qascade !
Please run `go generate ./...` and push (not force-push) the results to this PR as described in CONTRIBUTING.md.

github/event_types.go
qascade: Screenshot 2022-10-19 at 1 08 45 AM

The Field is mentioned as `Sender` in the Github API docs. Should I change `Sender` -> `User` ??

gmlewis: Yes, the JSON makes must match.

qascade:
@gmlewis I done the requested changes, do let me know if further changes are required. Thanks.

github/event_types.go
```diff
@@ -228,6 +228,16 @@ type Discussion struct {
...
 }
 
+// DiscussionCommentEvent represents a discussion comment event.
+type DiscussionCommentEvent struct {
+	Action     *string            `json:"action,omitempty"`
+	Comment    *DiscussionComment `json:"comment,omitempty"`
+	Discussion *Discussion        `json:"discussion,omitempty"`
+	Repo       *Repository        `json:"repository,omitempty"`
+	Org        *Organization      `json:"organization,omitempty"`
+	User       *User              `json:"user,omitempty"`
```
gmlewis, line 238:
```suggestion
	Sender       *User              `json:"sender,omitempty"`
```

The field names must match the official documentation, obviously. I do not have time for a full review right now. Please verify for yourself that your solution is correct before I perform a full review.

github/event_types.go
```diff
@@ -228,6 +228,16 @@ type Discussion struct {
...
+// DiscussionCommentEvent represents a discussion comment event.
+type DiscussionCommentEvent struct {
+	Action     *string            `json:"action,omitempty"`
+	Comment    *DiscussionComment `json:"comment,omitempty"`
+	Discussion *Discussion        `json:"discussion,omitempty"`
+	Repo       *Repository        `json:"repository,omitempty"`
+	Org        *Organization      `json:"organization,omitempty"`
+	Sender     *User              `json:"sender,omitempty"`
+}
+
```
gmlewis, line 240: Could you please move this new struct before `DiscussionEvent` on line 192?

gmlewis (changes requested):
Thank you, @qascade !
Just one minor change, please, then we will be ready for a second LGTM+Approval before merging.

github/event.go
```diff
@@ -52,6 +52,8 @@ func (e *Event) ParsePayload() (payload interface{}, err error) {
 		payload = &DeploymentEvent{}
 	case "DeploymentStatusEvent":
 		payload = &DeploymentStatusEvent{}
+	case "DiscussionCommentEvent":
+		payload = &DiscussionCommentEvent{}
```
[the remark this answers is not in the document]
qascade: The test for the switch cases are auto generated by `go generate ./...`. I am not able to understand what do you mean by this? Do you want me to add a small test for each case explicitly so, if someone changes struct name by mistake gets informed for the same??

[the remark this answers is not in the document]
raynigon: Please correct me if I am wrong, but I did not find any tests for the switch/case, neither manual nor generated.

> Do you want me to add a small test for each case explicitly so, if someone changes struct name by mistake gets informed for the same??
It would be nice if you could add this. Then anybody can add more cases in the future.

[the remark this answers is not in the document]
qascade: I saw all a file called `git-accessors_test.go` that was auto-generated, it contained tests for all the payload attributes.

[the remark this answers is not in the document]
qascade: >It would be nice if you could add this. Then anybody can add more cases in the future.
Will try to do this. I think it's better to add these tests because auto-generated tests will automatically changed as per the changes done by another contributor, which is not the behaviour we want.

gmlewis (changes requested):
Please note that `github/messages.go` also needs modification around line 59 by adding the new event type to the `eventTypeMapping` map.

PR #2508, opened 2022-10-19
chore: capitalize H on event type and delivery id header keys

marcelosousa:
This pull request introduces the following changes:

- Refactor `Github` to `GitHub` in the message header keys for `EventTypeHeader` and `DeliveryIDHeader`. I found this to be useful because when we use AWS Gateway to process the payload of a GitHub app, we don't receive a HTTP request (see https://github.com/aws/aws-lambda-go/blob/main/events/apigw.go#L6). By using the current constants we will not be able to correctly access those header values as GitHub sends them with a capitalized H.
- Added unit tests to verify that the http header get is case insensitive and doesn't introduce regressions for users that use `WebHookType()` and `DeliveryID()`.

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -37,9 +37,9 @@ const (
 	// SHA256SignatureHeader is the GitHub header key used to pass the HMAC-SHA256 hexdigest.
 	SHA256SignatureHeader = "X-Hub-Signature-256"
 	// EventTypeHeader is the GitHub header key used to pass the event type.
-	EventTypeHeader = "X-Github-Event"
+	EventTypeHeader = "X-GitHub-Event"
 	// DeliveryIDHeader is the GitHub header key used to pass the unique ID for the webhook event.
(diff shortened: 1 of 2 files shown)


marcelosousa:
FYI: I've signed the CLA.

gmlewis:
Thank you, @marcelosousa , but according to: https://www.rfc-editor.org/rfc/rfc4485.txt

> Although header field
   names are case insensitive, a single common capitalization SHOULD be
   used throughout the document.  It is RECOMMENDED that each English
   word present in the header field name have its first letter
   capitalized. 

So this is not a change we wish to make.
Closing.

gmlewis:
> Thanks @gmlewis. I just want to point out that as is, these headers will not be directly usable for other requests such as the one provided by AWS Gateway. This is because the headers there are a simple map string to string and GitHub sends those headers as 'GitHub' and not 'Github'.

That doesn't make sense to me because these headers are case insensitive in my experience and I believe also in the spec.

PR #2518, opened 2022-10-25
Enterprise actions permissions endpoints

algorythmic:
Add `/enterprises/` versions of the `/orgs/` endpoints which have already implemented from https://docs.github.com/en/rest/actions/permissions

Since `orgs_actions_permissions.go` and `orgs_actions_allowed.go` already existed, just created `enterprise_*` versions of those.

The existing `/orgs/{org}/actions/permissions/repositories` implementation was in `actions_runners.go`, which doesn't seem to fit well. Rather than adding the enterprise version there as well, moved it out to `orgs_actions_permissions.go`, hope that's OK.


github/enterprise_actions_allowed.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2022 The go-github AUTHORS. All rights reserved.
```

github/enterprise_actions_allowed_test.go
```diff
@@ -0,0 +1,93 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2022 The go-github AUTHORS. All rights reserved.
```

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2022 The go-github AUTHORS. All rights reserved.
```

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, resp, err
+	}
+
+	return p, resp, nil
+}
+
+// ListEnabledOrgsInEnterprise lists the selected organizations that are enabled for GitHub Actions in an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#list-selected-organizations-enabled-for-github-actions-in-an-enterprise
+func (s *ActionsService) ListEnabledOrgsInEnterprise(ctx context.Context, owner string, opts *ListOptions) (*ActionsEnabledOnEnterpriseOrgs, *Response, error) {
```
gmlewis, line 74: Shouldn't this be `EnterpriseService` here instead of `ActionsService`?

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, resp, err
+	}
+
+	return orgs, resp, nil
+}
+
+// SetEnabledOrgsInEnterprise replaces the list of selected organizations that are enabled for GitHub Actions in an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#set-selected-organizations-enabled-for-github-actions-in-an-enterprise
+func (s *ActionsService) SetEnabledOrgsInEnterprise(ctx context.Context, owner string, organizationIDs []int64) (*Response, error) {
```
gmlewis, line 98: Shouldn't this be `EnterpriseService` here instead of `ActionsService`?

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// AddEnabledOrgInEnterprise adds an organization to the list of selected organizations that are enabled for GitHub Actions in an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#enable-a-selected-organization-for-github-actions-in-an-enterprise
+func (s *ActionsService) AddEnabledOrgInEnterprise(ctx context.Context, owner string, organizationID int64) (*Response, error) {
```
gmlewis, line 119: Shouldn't this be `EnterpriseService` here instead of `ActionsService`?

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// RemoveEnabledOrgInEnterprise removes an organization from the list of selected organizations that are enabled for GitHub Actions in an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#disable-a-selected-organization-for-github-actions-in-an-enterprise
+func (s *ActionsService) RemoveEnabledOrgInEnterprise(ctx context.Context, owner string, organizationID int64) (*Response, error) {
```
gmlewis, line 138: Shouldn't this be `EnterpriseService` here instead of `ActionsService`?

github/enterprise_actions_permissions_test.go
```diff
@@ -0,0 +1,221 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2022 The go-github AUTHORS. All rights reserved.
```

gmlewis (changes requested):
Thank you, @algorythmic !

github/enterprise_actions_permissions.go
algorythmic: My intention was for the new `ActionsService.ListEnabledOrgsInEnterprise` to mirror the existing [ActionsService.ListEnabledReposInOrg](https://pkg.go.dev/github.com/google/go-github/v48/github#ActionsService.ListEnabledReposInOrg), etc. since they implement the similar `/enterprises/{enterprise}/actions/permissions/organizations` and `/orgs/{org}/actions/permissions/repositories` endpoints respectively.

It _would_ make more sense to have `ListEnabledOrgsInEnterprise` be an `EnterpriseService` method but should `ListEnabledReposInOrg` and friends be moved to `OrganizationsService` then to match?

gmlewis: We used to determine which service an endpoint belonged to using the organization imposed by the official GitHub v3 API documentation.

Now I'm not sure that the decision is as clear-cut with the documentation site re-organization.

I'm wondering if the `/enterprises/...` endpoints should be in the `EnterpriseService`, the `/orgs/...` in the OrganizationsService, etc... but then am hesitant to do a complete shuffle of methods, as that is extremely disruptive to existing users.

Does anyone have thoughts or ideas on how to address this with the least amount of disruption?

raynigon: > I'm wondering if the /enterprises/... endpoints should be in the EnterpriseService, the /orgs/... in the OrganizationsService, etc... but then am hesitant to do a complete shuffle of methods, as that is extremely disruptive to existing users.

In this situation i think its a bit difficult, but if documentation is added, existing users should be able to easily migrate to the new methods. If the code is oriented on the GitHub API itself, it should make it easier for new users to archieve their goal with this library, because they know the structure from playing around with the GitHub API in curl. 
Therefore I think all Services should be named by their API Endpoint whenever possible.

gmlewis: OK, I think my prior recommendations were incorrect.  Following the documentation links, the official documentation places these endpoints under the "Actions" services, so let's please keep them there.

It looks like we now have a situation of duplicated code in two different files, so let's please resolve that problem, @algorythmic,  and then we can continue with this PR.

RickleAndMortimer:
Is this PR still being worked on? I'd be down to help finish and close it out

ajschmidt8:
My team could also benefit from these changes being merged.

@RickleAndMortimer, since @algorythmic is unresponsive, will you be opening a new PR to continue this work?

RickleAndMortimer:
For sure, I will set up a new PR for this

gmlewis:
Closing PR as abandoned.

PR #2535, opened 2022-10-26
Add test for resource JSON marshaling

Harikesh00:
Added test JSON with more data for `RepositoriesSearchResult`

### DIFF
--- a/github/search_test.go
+++ b/github/search_test.go
@@ -654,6 +654,134 @@ func TestRepositoriesSearchResult_Marshal(t *testing.T) {
 	}`
 
 	testJSONMarshal(t, u, want)
+
+	u2 := &RepositoriesSearchResult{
+		Total:             Int(9),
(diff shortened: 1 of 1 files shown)


Harikesh00:
plz add the hacktoberfest accepted label
Thank you

gmlewis (commented):
Thank you, @Harikesh00 , but upon looking at this PR, I don't think there is really much value added here because you are essentially only testing the expansion of the `Repository` struct which I believe is well-tested elsewhere.

Please focus your attention on structs that are not well-tested in future PRs... thank you very much, and your contributions are greatly appreciated!

Also, in future PRs, only two `testJSONMarshal` calls per struct are desired... an initial empty one and a filled-in one. Thank you.

Closing PR.

Harikesh00:
ok, sure
I'll take care of these things

Issue #2574, opened 2022-11-23
Add new Branch Protection "Require Last Push Approval"


fproulx-boostsecurity:
New Branch Protection features added in October are not yet supported https://github.blog/changelog/2022-10-20-new-branch-protections-last-pusher-and-locked-branch/

https://docs.github.com/en/rest/branches/branch-protection#get-branch-protection
`require_last_push_approval`

gmlewis:
Thank you, @fproulx-boostsecurity !

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

XaurDesu:
Hey there, is this issue still available? I'm currently learning how to contribute to open source, and this was marked as a good place to start.

gmlewis:
> Hey there, is this issue still available? I'm currently learning how to contribute to open source, and this was marked as a good place to start.

Cool! Yes, this issue is available.
Thank you, @XaurDesu , this issue is yours.

XaurDesu:
Where should this change be located, exactly? I am currently reviewing the code structure of this project, yet I am not really sure where should we add this change. commit cb24cabfacd4bff06727d5b3acea48ba56e5f690 is concerned with this functionality too and references `github/github-accessors.go`, `github/github-accessors_test.go` and `github/repos.go`, yet i'm not sure if either of those is the target for this change, sorry if this is an amateurish question, then again, i'm pretty new.

dtdao:
@XaurDesu Dang you beat me getting to this ticket. 😄 
I was also kinda poking around as well. And i think https://github.com/google/go-github/commit/cb24cabfacd4bff06727d5b3acea48ba56e5f690 added `require_last_push_approval` but there still no implementations for `lock_branch` and `allow_fork_synching`.

So...i would think you need to update the `Protection` struct with those 2 new fields.
and following the patterns with the other fields in that struct.  Add 2 new get functions `GetLockBranch` and `GetAllowForkSynching` .  But that just kinda a guess, from a guy trying to learn as well. 😅 

cc: @gmlewis

gmlewis:
[...]

PR #2577, opened 2022-11-26
Add LockBranch and AllowForkSyncing to repos.go

XaurDesu:
Fixes https://github.com/google/go-github/issues/2574 , and it is mostly a direct, if barebones, implementation of the plan ideated there. Adding references to `LockBranch` and `AllowForkSyncing`. References to `RequireLastPushApproval` were added in cb24cabfacd4bff06727d5b3acea48ba56e5f690 . There might be some changes left to do.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -841,9 +841,9 @@ type Protection struct {
 	AllowDeletions                 *AllowDeletions                 `json:"allow_deletions"`
 	RequiredConversationResolution *RequiredConversationResolution `json:"required_conversation_resolution"`
 	// LockBranch represents if the branch is marked as read-only. If this is true, users will not be able to push to the branch.
-	LockBranch                     *bool                     `json:"lock_branch,omitempty"`
+	LockBranch *bool `json:"lock_branch,omitempty"`
 	// AllowForkSyncing represents whether users can pull changes from upstream when the branch is locked.
-	AllowForkSyncing               *bool               `json:"allow_fork_syncing,omitempty"`
+	AllowForkSyncing *bool `json:"allow_fork_syncing,omitempty"`
 }
 
 // BranchProtectionRule represents the rule applied to a repositories branch.
@@ -898,19 +898,9 @@ type AdminEnforcedChanges struct {
 	From *bool `json:"from,omitempty"`
 }
 
-// LockBranch represents if the branch is marked as read-only. If this is true, users will not be able to push to the branch.
-type LockBranch struct {
-	From *bool `json:"lock_branch"`
-}
-
 // AllowDeletionsEnforcementLevelChanges represents the changes made to the AllowDeletionsEnforcementLevel policy.
 type AllowDeletionsEnforcementLevelChanges struct {
 	From *string `json:"from,omitempty"`
-}
-
-// AllowForkSyncing represents whether users can pull changes from upstream when the branch is locked.
-type AllowForkSyncing struct {
-	From *bool `json:"allow_fork_syncing"`
 }
 
 // AuthorizedActorNames represents who are authorized to edit the branch protection rules.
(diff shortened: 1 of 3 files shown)


github/repos.go
```diff
@@ -894,11 +896,21 @@ type AdminEnforcedChanges struct {
 	From *bool `json:"from,omitempty"`
 }
 
+// LockBranch represents if the branch is marked as read-only. If this is true, users will not be able to push to the branch.
+type LockBranch struct {
+	From *bool `json:"lock_branch"`
+}
+
```
gmlewis, line 903: Please delete these lines.

github/repos.go
```diff
@@ -840,6 +840,8 @@ type Protection struct {
 	AllowForcePushes               *AllowForcePushes               `json:"allow_force_pushes"`
 	AllowDeletions                 *AllowDeletions                 `json:"allow_deletions"`
 	RequiredConversationResolution *RequiredConversationResolution `json:"required_conversation_resolution"`
+	LockBranch                     *LockBranch                     `json:"lock_branch"`
+	AllowForkSyncing               *AllowForkSyncing               `json:"allow_fork_syncing"`
```
gmlewis, line 844:
```suggestion
	// LockBranch represents if the branch is marked as read-only. If this is true, users will not be able to push to the branch.
	LockBranch                     *bool                     `json:"lock_branch,omitempty"`
	// AllowForkSyncing represents whether users can pull changes from upstream when the branch is locked.
	AllowForkSyncing               *bool               `json:"allow_fork_syncing,omitempty"`
```

github/repos.go
```diff
@@ -894,11 +896,21 @@ type AdminEnforcedChanges struct {
...
 // AllowDeletionsEnforcementLevelChanges represents the changes made to the AllowDeletionsEnforcementLevel policy.
 type AllowDeletionsEnforcementLevelChanges struct {
 	From *string `json:"from,omitempty"`
 }
 
+// AllowForkSyncing represents whether users can pull changes from upstream when the branch is locked.
+type AllowForkSyncing struct {
+	From *bool `json:"allow_fork_syncing"`
+}
+
```
gmlewis, line 913: Please delete these lines.

gmlewis (changes requested):
After making the requested changes below, please run `gofmt` on `repos.go` and then also run `go generate ./...` at the top of the repo and push (not force-push) any changes that were made, as described in the file CONTRIBUTING.md.  
Thanks.

github/repos.go
XaurDesu: Just commited a deletion for both struct

XaurDesu:
I can't seem to replicate the failing test in my local environment, already ran `gofmt` and `go generate`, but it seems to have no effect on tests, i think i might have a configuration problem on my end but i can't seem to figure it out.

dtdao:
From checking the workflow file. The tests didnt run.  Its failing at the
https://github.com/google/go-github/blob/621c4ba7155bee18761f308a86fa71fddf3476b4/.github/workflows/tests.yml#L64

specifically at `git diff --exit-code; ` portion of it.

gmlewis:
Yes, you are correct, @dtdao - thank you for the detailed explanation. 

This information should also be covered in our CONTRIBUTING.md guide at the top of this repo... if you feel that it can be made clearer, we welcome PRs for it as well.

XaurDesu:
I agree that `CONTRIBUTING.md` could (and probably should) be clearer about some things, but as it should be evident by now, I'm pretty new to this kind of workflow, and i would rather solve a few more issues to get accustomed to the codebase before messing with any documentation we have, especially kind that regards new contributors. Maybe @dtdao could check that out.

github/repos.go
XaurDesu: done.

gmlewis (approved):
Thank you, @XaurDesu !
LGTM.
Merging.

PR #2583, opened 2022-11-29
Add new field for issue state reason

ganeshkumarsv:
Adds new field `state_reason` for GitHub Issue struct.
reference on the field: https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -28,6 +28,7 @@ type Issue struct {
 	ID                *int64            `json:"id,omitempty"`
 	Number            *int              `json:"number,omitempty"`
 	State             *string           `json:"state,omitempty"`
+	// StateReason can be one of: "completed", "not_planned", "reopened".
 	StateReason       *string           `json:"state_reason,omitempty"`
 	Locked            *bool             `json:"locked,omitempty"`
 	Title             *string           `json:"title,omitempty"`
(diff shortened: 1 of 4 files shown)


github/issues.go
```diff
@@ -28,6 +28,7 @@ type Issue struct {
 	ID                *int64            `json:"id,omitempty"`
 	Number            *int              `json:"number,omitempty"`
 	State             *string           `json:"state,omitempty"`
+	StateReason       *string           `json:"state_reason,omitempty"`
```
gmlewis, line 31: Since the options are few, can we please add a comment to this field?
(You will probably also have to reformat this file if you take this suggestion directly as-is.)

```suggestion
	// StateReason can be one of: "completed", "not_planned", "reopened".
	StateReason       *string           `json:"state_reason,omitempty"`
```

gmlewis (changes requested):
Thank you, @ganeshkumarsv !
I think a comment here might be useful, if you don't mind.

ganeshkumarsv:
Thanks for the suggestion @gmlewis! I committed your suggestion

ganeshkumarsv:
@gmlewis once the PR gets approved and merged, can we have a new release?

gmlewis:
> @gmlewis once the PR gets approved and merged, can we have a new release?

I would really like to get #2581 in before releasing `v48.2.0`.  Hopefully that can happen today.

gmlewis (approved):
Thank you, @ganeshkumarsv !
LGTM.
Merging.

Issue #2590, opened 2022-12-05
ci(tests): use 'concurrency' instead of 'styfle/cancel-workflow-action'


gurza:
Use `concurrency` instead of `styfle/cancel-workflow-action`

Concurrency docs: https://docs.github.com/en/actions/using-jobs/using-concurrency
PR to archive `styfle/cancel-workflow-action`: https://github.com/styfle/cancel-workflow-action

PR #2591, opened 2022-12-05
ci(tests): use 'concurrency' instead of 'styfle/cancel-workflow-action'

gurza:
Solve #2590


.github/workflows/tests.yml
```diff
@@ -15,9 +19,6 @@ permissions:
 
 jobs:
   test:
-    permissions:
-      actions: write  # for styfle/cancel-workflow-action to cancel/stop running workflows
-      contents: read  # for actions/checkout to fetch code
```
gmlewis, line 20: I thought that this `contents: read` permission was orthogonal to the `actions:` permission and ensured that bad actors cannot write to the repo within an action.

If this is the case, I think we should keep this permission and not delete it.

Thoughts?

gurza: @gmlewis thank you for your check.

In `tests.yml` there is high level `permissions` section which is set to read access.

```
permissions:
  contents: read
```

I have deleted `jobs/test/permissions` subsection I think it enhances read access. As you can see in the comment `actions: write` is used for `styfle/cancel-workflow-action`.

Or am I missing something?

gmlewis: Ah, so that `contents: read` was redundant!  OK, thank you, @gurza !

gmlewis (approved):
Thank you, @gurza !
LGTM.

Awaiting signed CLA before merging.

PR #2597, opened 2022-12-13
Change actions billing structs to maps

### LINKED ISSUE #2595: Add new machine types to response of GetWorkflowRunUsageByID
Hi! As we're all probably aware, Github has added a bumch of different machine sizes to the actions runners. The timings endpoint returns these under keys like `UBUNTU_4_CORE`, etc. But the present version of this package only supports the 'vanilla' `WINDOWS`, `MACOS` and `UBUNTU` runners. It would be helpful to add the others if possible. Would obviously be a breaking change - not sure what the process is for handling that but happy to throw a PR together if someone can point me to all that.

tragiclifestories:
Fixes #2595.

The structs hardcode the three machine types that Github Actions supported until recently. However, we now have a lot more different types of machines, with the potential for more to be added at any time. Using maps instead of structs will allow us to transparently support more machine classes in the future.

I've done this pretty naively instead of creating safe access methods or anything like that, and kept the map values as pointers for consistency with other maps in the codebase. Can't think of any tests that would be necessary (beyond the generated ones) but open to suggestions.

### DIFF
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -70,6 +70,7 @@ type WorkflowRunUsage struct {
 }
 
 // WorkflowRunBillMap represents different runner environments available for a workflow run.
+// Its key is the name of its environment, e.g. "UBUNTU", "MACOS", "WINDOWS", etc.
 type WorkflowRunBillMap map[string]*WorkflowRunBill
 
 // WorkflowRunBill specifies billable time for a specific environment in a workflow run.
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -65,16 +65,12 @@ type ListWorkflowRunsOptions struct {
 
 // WorkflowRunUsage represents a usage of a specific workflow run.
 type WorkflowRunUsage struct {
-	Billable      *WorkflowRunEnvironment `json:"billable,omitempty"`
-	RunDurationMS *int64                  `json:"run_duration_ms,omitempty"`
+	Billable      *WorkflowRunBillMap `json:"billable,omitempty"`
+	RunDurationMS *int64              `json:"run_duration_ms,omitempty"`
 }
 
-// WorkflowRunEnvironment represents different runner environments available for a workflow run.
-type WorkflowRunEnvironment struct {
-	Ubuntu  *WorkflowRunBill `json:"UBUNTU,omitempty"`
-	MacOS   *WorkflowRunBill `json:"MACOS,omitempty"`
-	Windows *WorkflowRunBill `json:"WINDOWS,omitempty"`
-}
+// WorkflowRunBillMap represents different runner environments available for a workflow run.
+type WorkflowRunBillMap map[string]*WorkflowRunBill
 
 // WorkflowRunBill specifies billable time for a specific environment in a workflow run.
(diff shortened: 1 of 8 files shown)


github/actions_workflow_runs.go
```diff
@@ -65,16 +65,12 @@ type ListWorkflowRunsOptions struct {
...
 }
 
-// WorkflowRunEnvironment represents different runner environments available for a workflow run.
-type WorkflowRunEnvironment struct {
-	Ubuntu  *WorkflowRunBill `json:"UBUNTU,omitempty"`
-	MacOS   *WorkflowRunBill `json:"MACOS,omitempty"`
-	Windows *WorkflowRunBill `json:"WINDOWS,omitempty"`
-}
+// WorkflowRunBillMap represents different runner environments available for a workflow run.
+type WorkflowRunBillMap map[string]*WorkflowRunBill
```
gmlewis, line 73:
```suggestion
// Its key is the name of its environment, e.g. "UBUNTU", "MACOS", "WINDOWS", etc.
type WorkflowRunBillMap map[string]*WorkflowRunBill
```

github/actions_workflow_runs_test.go
```diff
@@ -37,8 +37,8 @@ func TestActionsService_ListWorkflowRunsByID(t *testing.T) {
 	want := &WorkflowRuns{
 		TotalCount: Int(4),
 		WorkflowRuns: []*WorkflowRun{
-			{ID: Int64(399444496), RunNumber: Int(296), CreatedAt: &Timestamp{time.Date(2019, time.January, 02, 15, 04, 05, 0, time.UTC)}, UpdatedAt: &Timestamp{time.Date(2020, time.January, 02, 15, 04, 05, 0, time.UTC)}},
-			{ID: Int64(399444497), RunNumber: Int(296), CreatedAt: &Timestamp{time.Date(2019, time.January, 02, 15, 04, 05, 0, time.UTC)}, UpdatedAt: &Timestamp{time.Date(2020, time.January, 02, 15, 04, 05, 0, time.UTC)}},
+			{ID: Int64(399444496), RunNumber: Int(296), CreatedAt: &Timestamp{time.Date(2019, time.January, 0o2, 15, 0o4, 0o5, 0, time.UTC)}, UpdatedAt: &Timestamp{time.Date(2020, time.January, 0o2, 15, 0o4, 0o5, 0, time.UTC)}},
```
gmlewis, line 40: There's no reason for these numbers to be octal.
Please revert all number changes.
If your editor did this automatically, feel free to drop the leading zeros (which are also unnecessary) if that will prevent it from making the changes.

gmlewis (changes requested):
Thank you, @tragiclifestories !

github/actions_workflow_runs_test.go
tragiclifestories: I can probably just do some git surgery here. I think the leading zeros must have misled gofumpt which runs on save ...

tragiclifestories: @gmlewis fixed now.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

PR #2601, opened 2022-12-16
Include response in `ErrorResponse`s that are parsed from the response body

stoneman:
When an error in the response body is successfully parsed, the errorResponse no longer has a `Response` property which leads to panics when calling `errorResponse.Error()`. This PR fixes that by setting the `Response` property after the unmarshal attempt rather than before.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -1070,11 +1070,12 @@ func CheckResponse(r *http.Response) error {
 		return nil
 	}
 
-	errorResponse := &ErrorResponse{Response: r}
+	errorResponse := &ErrorResponse{}
 	data, err := io.ReadAll(r.Body)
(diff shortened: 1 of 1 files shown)


gmlewis:
Thank you, @stoneman .
This seems really odd to me that this was not caught in a unit test.
Could you please add a unit test that demonstrates that this problem is fixed (and one which fails without the fix, of course)?

stoneman:
It turned out that the application I was working on was using `go-github-mock` which was [known to cause this issue](https://github.com/migueleliasweb/go-github-mock/issues/6) so I [submitted a fix](https://github.com/migueleliasweb/go-github-mock/pull/41) there.

PR #2606, opened 2022-12-23
Fix JSON protection unmarshal error

Xorima:
## Description of change

Currently when trying to load in protection an error is thrown due to a data structure mismatch resulting in errors:
```bash
json: cannot unmarshal object into Go struct field Protection.lock_branch of type bool 
json: cannot unmarshal object into Go struct field Protection.allow_fork_syncing of type bool
```
This is due to them using an enabled field which the current structure does not understand. This change introduces the relevant structure and has been tested locally.

## API Payload:

```json
{
    "url": "https://api.github.com/repos/sous-chefs/sc_vscode/branches/main/protection",
   ...
    "required_conversation_resolution": {
        "enabled": false
    },
    "lock_branch": {
        "enabled": false
    },
    "allow_fork_syncing": {
        "enabled": false
    }
}
```

As can be seen on the payload above they are under the `enabled` key and are not bool fields directly in their own right. 

Signed-off-by: Jason Field <jason@avon-lea.co.uk>

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -470,6 +470,14 @@ func (a *AllowDeletionsEnforcementLevelChanges) GetFrom() string {
 	return *a.From
 }
 
+// GetEnabled returns the Enabled field if it's non-nil, zero value otherwise.
+func (a *AllowForkSyncing) GetEnabled() bool {
+	if a == nil || a.Enabled == nil {
+		return false
+	}
+	return *a.Enabled
+}
+
 // GetRef returns the Ref field if it's non-nil, zero value otherwise.
 func (a *AnalysesListOptions) GetRef() string {
 	if a == nil || a.Ref == nil {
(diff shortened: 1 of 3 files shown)


gmlewis (approved):
Thank you, @Xorima !
LGTM.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

Xorima:
Thanks @gmlewis, can we get a release cut in the next week or two so I don't have to continue to use my fork?

gmlewis:
@Xorima - might you have time to review https://github.com/google/go-github/pull/2611 (so that we can incorporate it into the next release, as I think that will be an important one to include)?

Xorima:
Sure I'll try to take a look in the morning @gmlewis

Issue #2602, opened 2022-12-20
Repository environment creation fails on private Teams repos


AnitaErnszt:
This is related to the discussion on the https://github.com/integrations/terraform-provider-github/issues/1262

In June GitHub made the environments and environment protection rules [generally available](https://github.blog/changelog/2021-06-22-github-actions-environments-environment-protection-rules-and-environment-secrets-are-generally-available/), however, the `wait_timer` and  `reviewers` remains limited to enterprise.

Many developers would like to make use of this great new feature, but the bug in the Terraform provider currently proves to be a blocker. 
The provider uses the `go-github` Go library under the hood and makes the following call:

```
_, _, err := client.Repositories.CreateUpdateEnvironment(ctx, owner, repoName, escapedEnvName, &updateData)
```

where the `&updateData` is a `github.CreateUpdateEnvironment` struct.

Looking at the [code](https://github.com/google/go-github/blob/master/github/repos_environments.go) it seems that instead of sending nil value for `wait_timer` it defaults to 0. 

```
// CreateUpdateEnvironment represents the fields required for the create/update operation
// following the Create/Update release example.
// See https://github.com/google/go-github/issues/992 for more information.
// Removed omitempty here as the API expects null values for reviewers and deployment_branch_policy to clear them.
type CreateUpdateEnvironment struct {
	WaitTimer              *int            `json:"wait_timer"`
	Reviewers              []*EnvReviewers `json:"reviewers"`
	DeploymentBranchPolicy *BranchPolicy   `json:"deployment_branch_policy"`
}
```

This is causing the issue for us, as the API doesn't expect this parameter and returns 422.

I understand why the `omitempty` was removed, so I'm looking at what alternative solutions are possible to ensure environments fully work regardless of the payment plan.

gmlewis:
Please read #2467 and see if that answers your questions.

AnitaErnszt:
@gmlewis It seems unrelated to me.

I'm very new to `go-github` and go in general, so I could be wrong.

My core issue is that the API is expecting different objects from Enterprise and Teams customers. 
Above code is correct for Enterprise.  Adding `omitempty` would make it work for Teams, but would introduce bug for enterprise customers.

AnitaErnszt:
@gmlewis do you mean creating separate endpoints based on pricing plan (one for teams/pro where `wait_timer` is not sent, and one for enterprise where `wait_timer` defaults to 0 if not set)?

[...]

PR #2611, opened 2022-12-29
Add support for GitHub Environments for Pro/Teams pricing plans

AnitaErnszt:
The PR fixes the issue reported on https://github.com/google/go-github/issues/2602

In June GitHub made the environments and environment protection rules [generally available](https://github.blog/changelog/2021-06-22-github-actions-environments-environment-protection-rules-and-environment-secrets-are-generally-available/), however, the wait_timer and reviewers remain limited to enterprise.

Code was written to send `wait_timer` and `reviewers` with default values if no value is present, so existing settings can be cleared. 
Due to this, the code was returning 422 when called to create environments for Pro/Team private repos, as these parameters are only supported for enterprise or public repos.

PR intends to handle the error with a retry when the following conditions are met:
- error code is 422
- wait_timer is the default value (0)
- reviewers is the default value ([])

When retrying the call, a new object is sent, which only contains the parameters that are accepted for Pro/Teams private repos.

### DIFF
--- a/github/repos_environments.go
+++ b/github/repos_environments.go
@@ -189,7 +189,6 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
 
 	e := new(Environment)
 	resp, err := s.client.Do(ctx, req, e)
-
 	if err != nil {
 		// The API returns 422 when the pricing plan doesn't support all the fields sent.
 		// This path will be executed for Pro/Teams private repos.
@@ -199,28 +198,23 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
 		// and return an error if they did.
 		if resp != nil && resp.StatusCode == http.StatusUnprocessableEntity && environment != nil && len(environment.Reviewers) == 0 && environment.GetWaitTimer() == 0 {
 			return s.createNewEnvNoEnterprise(ctx, u, environment)
-
-		} else {
-			return nil, resp, err
 		}
-
+		return nil, resp, err
 	}
 	return e, resp, nil
 }
 
-// Creating an internal function for cases where the original call returned 422
+// createNewEnvNoEnterprise is an internal function for cases where the original call returned 422.
 // Currently only the `deployment_branch_policy` paramter is supported for Pro/Team private repos
 func (s *RepositoriesService) createNewEnvNoEnterprise(ctx context.Context, u string, environment *CreateUpdateEnvironment) (*Environment, *Response, error) {
 	req, err := s.client.NewRequest("PUT", u, &createUpdateEnvironmentNoEnterprise{
 		DeploymentBranchPolicy: environment.DeploymentBranchPolicy,
 	})
-
 	if err != nil {
 		return nil, nil, err
 	}
 
 	e := new(Environment)
-
 	resp, err := s.client.Do(ctx, req, e)
 	if err != nil {
 		return nil, resp, err
--- a/github/repos_environments.go
+++ b/github/repos_environments.go
@@ -189,15 +189,14 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
 	e := new(Environment)
 	resp, err := s.client.Do(ctx, req, e)
 
-	fmt.Printf("%#v", environment.Reviewers)
-	fmt.Printf("runs")
-
 	if err != nil {
 		// The API returns 422 when the pricing plan doesn't support all the fields sent.
 		// This path will be executed for Pro/Teams private repos.
 		// For public repos, regardless of the pricing plan, all fields supported.
 		// For Free plan private repos the returned error code is 404.
-		if resp.StatusCode == 422 {
+		// We are checking that the user didn't try to send a value for unsupported fields,
+		// and return an error if they did.
+		if resp.StatusCode == 422 && len(environment.Reviewers) == 0 && *environment.WaitTimer == 0 {
 
 			req, err = s.client.NewRequest("PUT", u, &CreateUpdateEnvironmentWithoutEnterprise{
 				DeploymentBranchPolicy: environment.DeploymentBranchPolicy,
--- a/github/repos_environments.go
+++ b/github/repos_environments.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 )
 
 // Environment represents a single environment in a repository.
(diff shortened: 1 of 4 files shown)


AnitaErnszt:
@gmlewis Please kindly check if this implementation makes sense. I have tested the code locally for:
- pro/teams private repos -> successfully creates the environment
- free plan public repos -> successfully creates the environment
- free plan private repos -> 404, which is expected

private enterprise repos should have the same 'route' as the free plan public repos.

Unfortunately, I don't know how to extend the testing function to test the different 'paths'

github/repos_environments.go
```diff
@@ -168,6 +168,13 @@ type CreateUpdateEnvironment struct {
 	DeploymentBranchPolicy *BranchPolicy   `json:"deployment_branch_policy"`
 }
 
+// CreateUpdateEnvironmentWithoutEnterprise represents the fields accepted for Pro/Teams private repos.
+// Ref: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment
+// See https://github.com/google/go-github/issues/2602 for more information.
+type CreateUpdateEnvironmentWithoutEnterprise struct {
```
gmlewis, line 174: This should not be exported as it is only used internally.

```suggestion
type createUpdateEnvironmentWithoutEnterprise struct {
```

github/repos_environments.go
```diff
@@ -181,8 +188,32 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
...
+
 	if err != nil {
-		return nil, resp, err
+		// The API returns 422 when the pricing plan doesn't support all the fields sent.
+		// This path will be executed for Pro/Teams private repos.
+		// For public repos, regardless of the pricing plan, all fields supported.
+		// For Free plan private repos the returned error code is 404.
+		// We are checking that the user didn't try to send a value for unsupported fields,
+		// and return an error if they did.
+		if resp.StatusCode == 422 && len(environment.Reviewers) == 0 && *environment.WaitTimer == 0 {
```
gmlewis, line 199:
```suggestion
		if resp != nil &&  resp.StatusCode == http.StatusUnprocessableEntity && environment != nil && len(environment.Reviewers) == 0 && environment.GetWaitTimer() == 0 {
```

gmlewis (changes requested):
Thank you, @AnitaErnszt !

We definitely need to add in tests for the special cases, and it should be relatively simple.
~In a new test, instead of returning a 200 OK, your could would initially return a 422, and then your test would look like all the others and return new responses.~

Alternatively, and maybe this would be the best approach anyway...

Instead of making a huge indented block like we have now, you could make a new method that is _NOT_ exported... something like `createNewEnvNoEnterprise` and then simply call it:

```go
if resp != nil &&  resp.StatusCode == http.StatusUnprocessableEntity && environment != nil && len(environment.Reviewers) == 0 && environment.GetWaitTimer() == 0 {
  return s.createNewEnvNoEnterprise(u, environment)
}
```

Then that way, you can have two new test functions... one just to make sure the 422 triggers the `if` and the other test function can fully test out `createNewEnvNoEnterprise` (like we already do for every other existing endpoint method).  Does that make sense?

github/repos_environments.go
```diff
@@ -180,6 +188,39 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
 	}
 
 	e := new(Environment)
+	resp, err := s.client.Do(ctx, req, e)
+
```
gmlewis, line 192: Please delete this blank line, as we want to keep the error processing tightly coupled to the line that generates the error.

github/repos_environments.go
```diff
@@ -180,6 +188,39 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
...
+		} else {
+			return nil, resp, err
+		}
+
+	}
+	return e, resp, nil
+}
+
+// Creating an internal function for cases where the original call returned 422
+// Currently only the `deployment_branch_policy` paramter is supported for Pro/Team private repos
```
gmlewis, line 212:
```suggestion
// createNewEnvNoEnterprise is an internal function for cases where the original call returned 422.
// Currently only the `deployment_branch_policy` parameter is supported for Pro/Team private repos.
```

github/repos_environments.go
```diff
@@ -180,6 +188,39 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
...
+func (s *RepositoriesService) createNewEnvNoEnterprise(ctx context.Context, u string, environment *CreateUpdateEnvironment) (*Environment, *Response, error) {
+	req, err := s.client.NewRequest("PUT", u, &createUpdateEnvironmentNoEnterprise{
+		DeploymentBranchPolicy: environment.DeploymentBranchPolicy,
+	})
+
```
gmlewis, line 217: Please delete this blank line.

github/repos_environments.go
```diff
@@ -180,6 +188,39 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	e := new(Environment)
+
```
gmlewis, line 223: Please delete this blank line.

github/repos_environments.go
```diff
@@ -180,6 +188,39 @@ func (s *RepositoriesService) CreateUpdateEnvironment(ctx context.Context, owner
...
+		// For Free plan private repos the returned error code is 404.
+		// We are checking that the user didn't try to send a value for unsupported fields,
+		// and return an error if they did.
+		if resp != nil && resp.StatusCode == http.StatusUnprocessableEntity && environment != nil && len(environment.Reviewers) == 0 && environment.GetWaitTimer() == 0 {
+			return s.createNewEnvNoEnterprise(ctx, u, environment)
+
+		} else {
+			return nil, resp, err
+		}
+
```
gmlewis, line 206:
```suggestion
		}
		return nil, resp, err
	}
```

github/repos_environments.go
```diff
@@ -168,6 +169,13 @@ type CreateUpdateEnvironment struct {
 	DeploymentBranchPolicy *BranchPolicy   `json:"deployment_branch_policy"`
 }
 
+// CreateUpdateEnvironmentWithoutEnterprise represents the fields accepted for Pro/Teams private repos.
```
gmlewis, line 172:
```suggestion
// createUpdateEnvironmentNoEnterprise represents the fields accepted for Pro/Teams private repos.
```

gmlewis (changes requested):
I know this is in "Draft" state, but wanted to provide some feedback.
Obviously, we need two more test functions to test the added "if" condition and the new method.
Thank you!

AnitaErnszt:
@gmlewis I have added the tests. 
I wasn't sure how to decide what API response to return (422 or 200), so I used a counter for now. I'm open to suggestions

AnitaErnszt:
On another note the contributor guide is recommending to run `golint` however, this project has been [deprecated](https://github.com/golang/lint)

Xorima (approved):
LGTM

gmlewis:
Thank you, @Xorima !
Merging.

Issue #2628, opened 2023-01-16
PullRequestReviewsEnforcementRequest struct should also include the RequireLastPushApproval field


liaodaniel:
Currently, the only way to enable the `RequireLastPushApproval` feature is to use [UpdatePullRequestReviewEnforcement](https://pkg.go.dev/github.com/google/go-github/v49/github#RepositoriesService.UpdatePullRequestReviewEnforcement).

According to the [docs](https://docs.github.com/en/rest/branches/branch-protection#update-branch-protection), we should be able to do it via [UpdateBranchProtection](https://pkg.go.dev/github.com/google/go-github/v49/github#RepositoriesService.UpdateBranchProtection). This would require updating the [PullRequestReviewsEnforcementRequest](https://pkg.go.dev/github.com/google/go-github/v49/github#PullRequestReviewsEnforcementRequest) struct.

gmlewis:
Thank you, @liaodaniel !
Do you want to put together a PR for this issue, or shall I open it up to other contributors to this repo?

liaodaniel:
@gmlewis, happy to submit the PR

gmlewis:
> @gmlewis, happy to submit the PR

Thank you, @liaodaniel !
It's yours.

PR #2629, opened 2023-01-16
Add RequireLastPushApproval field to UpdateBranchProtection

liaodaniel:
Fixes https://github.com/google/go-github/issues/2628

Signed-off-by: Daniel Liao <10663736+liaodaniel@users.noreply.github.com>

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -1064,6 +1064,8 @@ type PullRequestReviewsEnforcementRequest struct {
 	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
 	// Valid values are 1-6.
 	RequiredApprovingReviewCount int `json:"required_approving_review_count"`
+	// RequireLastPushApproval specifies whether the last pusher to a pull request branch can approve it.
+	RequireLastPushApproval bool `json:"require_last_push_approval"`
 }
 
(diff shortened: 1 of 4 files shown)


github/repos.go
```diff
@@ -1064,6 +1064,8 @@ type PullRequestReviewsEnforcementRequest struct {
 	// RequiredApprovingReviewCount specifies the number of approvals required before the pull request can be merged.
 	// Valid values are 1-6.
 	RequiredApprovingReviewCount int `json:"required_approving_review_count"`
+	// RequireLastPushApproval specifies whether the last pusher to a pull request branch can approve it.
+	RequireLastPushApproval bool `json:"require_last_push_approval"`
```
gmlewis, line 1068:
```suggestion
	RequireLastPushApproval *bool `json:"require_last_push_approval,omitempty"`
```

gmlewis (changes requested):
Thank you, @liaodaniel!
This field is optional, so we need to treat it a bit differently.

github/repos.go
liaodaniel: Updated with the suggestion

gmlewis (approved):
Thank you, @liaodaniel !
LGTM.
Merging.

PR #2635, opened 2023-01-19
Add secondary rate limit handling (prevent remote requests) and fix primary rate limit categories

### LINKED ISSUE #2633: Rate limit does not cover all categories & secondary rate limit not preventing requests

gal-legit:
* Add check-before-do for secondary rate limit
* Add optional hook for secondary limits
* Add missing categories for primary rate limits

Notes:
- As far as I could see, there is no documented API for runner registration - so I didn't handle this case.

fixes #2633 

Edit:
Following the conversation in this PR, the actual changes are:
* Add check-before-do for secondary rate limit
* Add missing categories for primary rate limits
* Add examples for using [go-github-ratelimit](https://github.com/gofri/go-github-ratelimit)

### DIFF
--- a/example/ratelimit/main.go
+++ b/example/ratelimit/main.go
@@ -1,4 +1,4 @@
-// Copyright 2017 The go-github AUTHORS. All rights reserved.
+// Copyright 2023 The go-github AUTHORS. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
--- a/example/ratelimit/main.go
+++ b/example/ratelimit/main.go
@@ -1,0 +1,39 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The ratelimit command demonstrates using the github_ratelimit.SecondaryRateLimitWaiter.
+// By using the waiter, the client automatically sleeps and retry requests
+// when it hits secondary rate limits.
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/gofri/go-github-ratelimit/github_ratelimit"
+	"github.com/google/go-github/v49/github"
+)
+
+func main() {
+	rateLimiter, err := github_ratelimit.NewRateLimitWaiterClient(nil)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+		return
+	}
+
+	client := github.NewClient(rateLimiter)
+
+	// arbitrary usage of the client
+	username := "some-username"
+	organizations, _, err := client.Organizations.List(context.Background(), username, nil)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+		return
+	}
+
+	for i, organization := range organizations {
+		fmt.Printf("%v. %v\n", i+1, organization.GetLogin())
+	}
+}
--- a/example/ratelimit/main.go
+++ b/example/ratelimit/main.go
@@ -13,10 +13,14 @@ import (
 	"fmt"
 
 	"github.com/gofri/go-github-ratelimit/github_ratelimit"
-	"github.com/google/go-github/v49/github"
+	"github.com/google/go-github/v50/github"
 )
 
 func main() {
+	var username string
+	fmt.Print("Enter GitHub username: ")
+	fmt.Scanf("%s", &username)
+
 	rateLimiter, err := github_ratelimit.NewRateLimitWaiterClient(nil)
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
@@ -26,7 +30,6 @@ func main() {
 	client := github.NewClient(rateLimiter)
 
 	// arbitrary usage of the client
-	username := "some-username"
 	organizations, _, err := client.Organizations.List(context.Background(), username, nil)
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
--- a/github/github.go
+++ b/github/github.go
@@ -171,8 +171,9 @@ type Client struct {
 	// User agent used when communicating with the GitHub API.
 	UserAgent string
 
-	rateMu     sync.Mutex
-	rateLimits [categories]Rate // Rate limits for the client as determined by the most recent API calls.
+	rateMu                  sync.Mutex
+	rateLimits              [categories]Rate // Rate limits for the client as determined by the most recent API calls.
+	secondaryRateLimitReset time.Time        // Secondary rate limit reset for the client as determined by the most recent API calls.
 
 	common service // Reuse a single struct instead of allocating one for each service on the heap.
 
@@ -712,6 +713,12 @@ func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, erro
 				Rate:     err.Rate,
 			}, err
 		}
+		// If we've hit a secondary rate limit, don't make further requests before Retry After.
+		if err := c.checkSecondaryRateLimitBeforeDo(ctx, req); err != nil {
+			return &Response{
+				Response: err.Response,
+			}, err
+		}
 	}
 
 	resp, err := c.client.Do(req)
@@ -762,6 +769,14 @@ func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, erro
 
 			aerr.Raw = b
 			err = aerr
+		}
+
+		// Update the secondary rate limit if we hit it.
+		rerr, ok := err.(*AbuseRateLimitError)
+		if ok && rerr.RetryAfter != nil {
+			c.rateMu.Lock()
+			c.secondaryRateLimitReset = time.Now().Add(*rerr.RetryAfter)
+			c.rateMu.Unlock()
 		}
 	}
 	return response, err
@@ -821,6 +836,35 @@ func (c *Client) checkRateLimitBeforeDo(req *http.Request, rateLimitCategory rat
 			Rate:     rate,
 			Response: resp,
 			Message:  fmt.Sprintf("API rate limit of %v still exceeded until %v, not making remote request.", rate.Limit, rate.Reset.Time),
+		}
+	}
+
+	return nil
+}
+
+// checkSecondaryRateLimitBeforeDo does not make any network calls, but uses existing knowledge from
+// current client state in order to quickly check if *AbuseRateLimitError can be immediately returned
+// from Client.Do, and if so, returns it so that Client.Do can skip making a network API call unnecessarily.
+// Otherwise it returns nil, and Client.Do should proceed normally.
+func (c *Client) checkSecondaryRateLimitBeforeDo(ctx context.Context, req *http.Request) *AbuseRateLimitError {
+	c.rateMu.Lock()
+	secondary := c.secondaryRateLimitReset
+	c.rateMu.Unlock()
+	if !secondary.IsZero() && time.Now().Before(secondary) {
+		// Create a fake response.
+		resp := &http.Response{
+			Status:     http.StatusText(http.StatusForbidden),
+			StatusCode: http.StatusForbidden,
+			Request:    req,
+			Header:     make(http.Header),
+			Body:       io.NopCloser(strings.NewReader("")),
+		}
+
+		retryAfter := time.Until(secondary)
+		return &AbuseRateLimitError{
+			Response:   resp,
+			Message:    fmt.Sprintf("API secondary rate limit exceeded until %v, not making remote request.", secondary),
+			RetryAfter: &retryAfter,
 		}
 	}
 
--- a/github/github.go
+++ b/github/github.go
@@ -1242,7 +1242,7 @@ const (
 )
 
 // category returns the rate limit category of the endpoint, determined by HTTP method and Request.URL.Path.
-func category(method string, path string) rateLimitCategory {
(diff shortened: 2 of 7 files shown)


gal-legit:
I have just realized that the hook must get the client as a parameter (and it's reflected in the tests, where I use the wrong client for the retry). I'll push a fix for that

gmlewis:
Hmmm... I'm not sure I'm happy about these changes, and I've got a heavy workload right now. Please give me a week to mull this over.

github/github.go
```diff
@@ -1192,12 +1267,35 @@ const (
 )
 
 // category returns the rate limit category of the endpoint, determined by Request.URL.Path.
-func category(path string) rateLimitCategory {
+func category(method string, path string) rateLimitCategory {
```
gmlewis, line 1270: Please update comment on line 1269: "determined by HTTP method and Request.URL.Path."

Also, please remove the extra "string":

```go
func category(method, path string) rateLimitCategory {
```

github/github.go
```diff
@@ -1192,12 +1267,35 @@ const (
...
+
+	// https://docs.github.com/en/rest/code-scanning#upload-an-analysis-as-sarif-data
+	case strings.HasSuffix(path, "/code-scanning/sarifs"):
+		return codeScanningUploadCategory
+
+	// https://docs.github.com/en/enterprise-cloud@latest/rest/scim
+	case strings.HasPrefix(path, "/scim/"):
+		return scimCategory
+
+		// missing a case for actionsRunnerRegistrationCategory: no API found
```
gmlewis, line 1298: This indentation seems off - did you run "gofmt" on this file?

github/github.go
```diff
@@ -301,18 +303,31 @@ func addOptions(s string, opts interface{}) (string, error) {
 	return u.String(), nil
 }
 
+type clientOption func(*Client)
```
gmlewis, line 306:
```suggestion
type ClientOption func(*Client)
```

github/github.go
```diff
@@ -301,18 +303,31 @@ func addOptions(s string, opts interface{}) (string, error) {
...
+type secondaryRateLimitHook func(ctx context.Context, client *Client, req *http.Request, orgError *AbuseRateLimitError) (*Response, error)
```
gmlewis, line 307:
```suggestion
type SecondaryRateLimitHook func(ctx context.Context, client *Client, req *http.Request, orgError *AbuseRateLimitError) (*Response, error)
```

github/github.go
```diff
@@ -301,18 +303,31 @@ func addOptions(s string, opts interface{}) (string, error) {
...
+
+func WithSecondaryRateLimitHook(hook secondaryRateLimitHook) clientOption {
```
gmlewis, line 309: Please add GoDoc-style comments (complete sentences ending in punctuation) to the 3 new exported items on lines 306, 307, and 309.

gmlewis (changes requested):
Thank you, @gal-legit .
I think this will be beneficial, but was initially unhappy with the added complexity in BareDo.
Let's move forward with the PR, and I apologize for the delay.

Please see if you can handle the currently-uncovered code path with a unit test if possible, and thank you for the very comprehensive tests you have already added!

gal-legit:
hey @gal-legit, thanks for the feedback!

I feel you - this PR adds complexity and new behavior to the client, and I don't feel too comfortable with adding this "mess" to the very elegant and clean style of the client.

I think that we can actually benefit from the delay.
During this week, I tackled this issue for the project I'm working on.
It started as a meantime solution, but I now think that taking this approach is probably better,
and it allows us to drop the hook and let the users handle it (somewhat like caching/token).

Check out secondaryRateLimitWaiter [here](https://github.com/Legit-Labs/legitify/blob/0c8137910869332ad32151026aa1834e6b56728a/internal/clients/github/transport/secnodary_rate_limit.go).
If you like the idea of dropping it, I can remove all hook-related stuff from this PR.
Also, if you think that this reference is worth recommending (like the cache/etc.), I can move that secondary rate limit transport to a different repo so that everyone can use it as a plugin.
Obviously, I'll also adjust it to fit general availability (remove the injecter and the log, add support for max-sleep, etc.).

Let me know what you think, and I'll continue from there.
Thanks!

gmlewis:
> Let me know what you think, and I'll continue from there.

Ah, genius!  I think handling it in an external-to-this-repo RoundTripper is a thing of beauty!
Let's proceed that way, and then this PR could instead point to the external RoundTripper in its examples and/or comments.

Thank you, @gal-legit !

gal-legit:
@gmlewis cool! I'll do the separation and update the PR with all the fixes and reference when it's ready.
I should keep the basic handling of the secondary rate limit in the client though. right?
i.e., saving it in the response and handling it in `checkSecondaryRateLimitBeforeDo()` to prevent remote requests, but with an implementation that looks just like `checkRateLimitBeforeDo()` so it isn't messy.

gmlewis:
Hmmm... would that tie this "go-github" repo to the external RoundTripper repo?
Is there a way to add the functionality without adding an external dependency?

gal-legit:
1. I meant in the context of the PR -- adding the handling to improve the client, regardless of the RoundTripper.
2. The current implementation of the RoundTripper uses go-github because the hosting project already does (so I didn't mind). The standalone version would parse the header and remove the dependency on go-github. So it shouldn't be tied anyway.

gmlewis:
OK, cool... I think I understand, but it will become more clear as I see the PR update.
Thanks again, @gal-legit !  I appreciate your patience and willingness to make this improvement!

gal-legit:
I have the new repo ready (https://github.com/gofri/go-github-ratelimit),
gonna have the changes to the PR ready soon

gmlewis:
> I have the new repo ready (https://github.com/gofri/go-github-ratelimit), gonna have the changes to the PR ready soon

Woohoo!  I gave it its very first GitHub star!  :smile:

github/github.go
gal-legit: yep, this is using gofmt.
I guess that it associates the comment with the case so it requires extra indentation.
I rearranged the comment to be part of the "default" case, so it makes more sense and there's no weird indentation.

gal-legit:
@gmlewis
I pushed the new version.
I resolved the comments that are related to the hooks because it's now omitted. Also, I responded to (and applied) the fixes from the other comments.

example/ratelimit/main.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2017 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
```

github/github.go
gmlewis: It looks like you still need to remove the extra "string". Please see my last comment above.

gmlewis (changes requested):
Thank you, @gal-legit !

gal-legit:
@gmlewis oops, done

gmlewis:
We keep getting linter errors.

```
level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: inspect: failed to load package : could not load export data: no export data for \"github.com/gofri/go-github-ratelimit/github_ratelimit\"\n\n"
```

It looks like you need to make a versioned release (0.0.1 maybe?) of your new external package, then update this PR to explicitly pull in that version in the `examples/go.mod` file to clear up this problem.

gmlewis:
It looks like there is a problem on Windows... I think it may have to do with how time comparisons are performed in your test.

```
    github_test.go:1461: abuseRateLimitErr RetryAfter = 2m3s, want 2m3s
--- FAIL: TestDo_rateLimit_abuseRateLimitError_retryAfter (0.00s)
```

Please check out [time.Equal](https://pkg.go.dev/time#Time.Equal) and see if you can fix this.

gal-legit:
@gmlewis 
My bad. It was actually a bug in the test:
I meant to test the delta between the expected and actual retry-after is less than a second,
but, I forgot to multiply by time.Second and the comparison was the other way around.
So it actually tested that the delta is bigger than a nanosecond. I guess that the timer has a lower resolution on Win so it caught it.
p.s. it's comparing time.Duration, so AFAIK it should be fine as long as I compare it to time.Second.

gmlewis:
> My bad. It was actually a bug in the test:

I find it bizarre that only the Windows version of the test caught the issue, and not the Linux build!?!?!?

OK, thanks.  I have to step out but hopefully will get back to this PR later tonight.

gofri:
LOL, I found it a bit embarrassing for me, being a Linux kernel engineer.
I guess that sometimes having a worse implementation might be good for testing 🤷

gmlewis:
> being a Linux kernel engineer.

Cool! I'd love to ask you about your favorite distros, but this is probably not the right forum. 😂

github/github_test.go
```diff
@@ -1136,6 +1141,36 @@ func TestDo_rateLimit(t *testing.T) {
 	}
 }
 
+func TestDo_rateLimitCategory(t *testing.T) {
+	if c := category(http.MethodGet, "/"); c != coreCategory {
```
gmlewis, line 1145: In reason that this is not a table-driven test?

I'm OK leaving it as-is if you want... since it would probably be the same number of lines, but I often find that table-driven tests are easier to read because you can just look at the data all in one place.

But no biggie... I'm fine either way.

github/github_test.go
```diff
@@ -1407,6 +1442,27 @@ func TestDo_rateLimit_abuseRateLimitError_retryAfter(t *testing.T) {
 	if got, want := *abuseRateLimitErr.RetryAfter, 123*time.Second; got != want {
 		t.Errorf("abuseRateLimitErr RetryAfter = %v, want %v", got, want)
 	}
+
+	// expect prevention of a following request
+	_, err = client.Do(ctx, req, nil)
+
+	if err == nil {
```
gmlewis, line 1449:
```suggestion
	if _, err := client.Do(ctx, req, nil); err == nil {
```

gmlewis (changes requested):
Thank you, @gal-legit !
One minor tweak please, plus one question, otherwise LGTM.

After the tweak, we should be ready for an LGTM+Approval from any other contributor to the repo before merging.

github/github_test.go
gal-legit: tbh I just tried to follow TestRateLimits & TestRateLimits_overQuota,
but I fixed it to be table-driven (and fixed those two while at it)

github/github_test.go
gal-legit: I thought it'd be a bit misleading to put it inside the if statement because we use err later (casting to *AbuseRateLimitError), but I fixed it anyway

github/github_test.go
```diff
@@ -1136,6 +1141,67 @@ func TestDo_rateLimit(t *testing.T) {
...
 }
 
+func TestDo_rateLimitCategory(t *testing.T) {
+	tests := []struct {
+		method   string
+		url      string
+		category rateLimitCategory
+	}{
+		{
+			http.MethodGet,
```
gmlewis, line 1151: This is a tiny nit, and if you don't want to do this, I'm fine...
but in this repo we prefer to label the fields for clarity-at-a-quick-glance (e.g. `method: http.MethodGet,`, etc.).
I'm going to go ahead and approve, but if you want to update, great.

gmlewis (approved):
Thank you, @gal-legit !

github/github_test.go
gal-legit: sure, np, I'll do that. 
p.s. Ubuntu for desktop, RHEL/fedora-based for server 🙃

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

Issue #2640, opened 2023-01-24
Possible missed json tag in github.ErrorResponse


migueleliasweb:
Hi all,

I've been struck by an weird interaction between Github's API and `go-github` on https://github.com/migueleliasweb/go-github-mock/issues/6 for a while now.

The issue seems to be that mocking errors from the API can be trickier than I first thought.

Recently, someone contributed with a pull request that mitigated the problem in https://github.com/migueleliasweb/go-github-mock/pull/41 but I think the problem goes a bit deeper than that. Although the mitigation PR fixes the issue, I think the root cause should be fixed in this repo.

The root cause seems to be that `github.ErrorResponse` is missing a `json tag` for ommiting the `Response` attr when it's not set. 

Here's why I think this is it:

Here's a not" found error" returned from the real Github's API, **notice there's no `Response` attribute set.**

```
{"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/teams#list-team-members"}
```

When the SDK receives this response, it first checks the status code and then goes and creates an error response that contains the mashalled information in the body.

See: https://github.com/google/go-github/blob/master/github/github.go#L1076
```golang
func CheckResponse(r *http.Response) error {
	if r.StatusCode == http.StatusAccepted {
		return &AcceptedError{}
	}
	if c := r.StatusCode; 200 <= c && c <= 299 {
		return nil
	}

	errorResponse := &ErrorResponse{Response: r}
	data, err := io.ReadAll(r.Body)
	if err == nil && data != nil {
		json.Unmarshal(data, errorResponse) // <- this is what causes the issue down the line
	}
        // [...] ommited for clarity
}
```

The problem here is that IF and ONLY IF the `r.Body` contains a `nil` `Response` atribute, the following function call will throw a **panic: runtime error: invalid memory address or nil pointer dereference**.

See: https://github.com/google/go-github/blob/master/github/github.go#L863

```golang
func (r *ErrorResponse) Error() string {
	return fmt.Sprintf("%v %v: %d %v %+v",
		r.Response.Request.Method, sanitizeURL(r.Response.Request.URL),
		r.Response.StatusCode, r.Message, r.Errors)
}
```

[...]

PR #2641, opened 2023-01-24
Add JSON tag for ErrorResponse

migueleliasweb:
Fixes: #2640.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -847,7 +847,7 @@ An ErrorResponse reports one or more errors caused by an API request.
 GitHub API docs: https://docs.github.com/en/rest/#client-errors
 */
 type ErrorResponse struct {
-	Response *http.Response // HTTP response that caused this error
+	Response *http.Response `json:"-"`       // HTTP response that caused this error
 	Message  string         `json:"message"` // error message
 	Errors   []Error        `json:"errors"`  // more detail on individual errors
 	// Block is only populated on certain types of errors such as code 451.


gmlewis:
Why is this even needed?
What problem is it solving?

github/github.go
```diff
@@ -841,9 +841,9 @@ An ErrorResponse reports one or more errors caused by an API request.
 GitHub API docs: https://docs.github.com/en/rest/#client-errors
 */
 type ErrorResponse struct {
-	Response *http.Response // HTTP response that caused this error
-	Message  string         `json:"message"` // error message
-	Errors   []Error        `json:"errors"`  // more detail on individual errors
+	Response *http.Response `json:"omitempty"` // HTTP response that caused this error
```
gmlewis, line 844: I've never seen a bare "omitempty" before, so can we please name it as "response"?

```suggestion
	Response *http.Response `json:"response,omitempty"` // HTTP response that caused this error
```

migueleliasweb: I'm fine adding the name of the field. But I wonder in this case if it might needs to be `Response` to maintain compatibility with the current code.

gmlewis: You are saying it should be "Response" with a capital-R ?
When you say "compatibility with the current code", do you mean the "go-github" repo?
If so, I'm not sure why we would need to capitalize the R, and I would think the lower-case version should be fine.

Alternatively, another approach might be to simply use `json:"-"` and always leave it out... but I'll let you decide if that is the best option for compatibility with your mocking package.

migueleliasweb: Hi Glenn, I actually liked the idea of going with `json:"-"`. Let me test a few things and I will come back to you.

Thanks for the swift responses!

migueleliasweb: Hi @gmlewis , it seems the `json:"-"`  should work just fine. I will update the PR soon.

migueleliasweb: done :)

gmlewis (approved):
Thank you, @migueleliasweb !
LGTM.

migueleliasweb:
It's always a pleasure, @gmlewis  :smiley:  . Is there someone I could ping here to get this going?

gmlewis:
@Parker77 - do you have time for an LGTM+Approval, please?

Parker77 (approved):
LGTM!

gmlewis:
Thank you, @Parker77!
Merging.

migueleliasweb:
Btw, merging this fixed the default error formatting issue on `go-github-mock` 😃 . Thanks all!

PR #2644, opened 2023-01-25
Use NewTokenClient in tests and examples

icholy:


gmlewis:
@Parker77 - do you have time for another?

gmlewis:
Thank you, @Parker77!
Merging.

Omobarew0:
Nice work more grace to your elbow

Issue #2645, opened 2023-01-25
[DevExperience]: Inconsistency in time related field types


zombieleet:
I am currently working with `go-github` sdk, and I noticed some inconsistencies in the types of time fields. Some fields are `*time.Time` while others are `*Timestamp`.

For example, this difference can be spotted in the [PullRequest](https://pkg.go.dev/github.com/google/go-github/v49/github#PullRequest) and [Projects](https://pkg.go.dev/github.com/google/go-github/v49/github#Project) struct. It is also littered around the codebase. In my opinion, let's stick to one type for time fields which is `*Timestamp`.

if this is seen as a valid issue, I will open a PR to fix it. 

Thanks.

gmlewis:
Thank you, @zombieleet .
It seems like there was a discussion about this before, but I'm not finding it.

Consistency would be nice, and I can't think of any downsides other than a boat-load of breaking API changes.
But that's why we version this, and people can hold back until they feel like upgrading.

So I have no objections to a PR.

gmlewis:
Fixed by: https://github.com/google/go-github/pull/2646
Closing.

PR #2646, opened 2023-01-26
Change all fields from time.Time to github.Timestamp

zombieleet:
This PR introduces breaking API changes as described here #2645 
It also fix the related test for the breaking API changes


.golangci.yml
```diff
@@ -21,7 +21,9 @@ linters:
     - staticcheck
     - ineffassign
     - unused
-issues: 
+issues:
+  exclude:
+    - composites
```
zombieleet, line 26: the examples project was having a type error, it was fixed by using `github.Timestamp`, which caused another error in `go vet` relating to unkeyed fields. excluding `composites` fixes the error

gmlewis: Curious what we were missing by excluding composites, I ran across these issues:

* https://github.com/golangci/golangci-lint/issues/446
* https://github.com/golang/go/issues/43864

in case anyone in the future finds this and wonders about it.

gmlewis:
@willnorris - this is a significant breaking change affecting a large number of fields.

Personally, I'm fine with it, but before we make such a sweeping change, I'd like to make sure you are fine with this too.

willnorris:
In general, I don't really love moving away from standard Go types where it's not actually necessary.  It makes it harder to interop with other libraries and existing code.  In hindsight, I kinda wonder if we could have made Timestamp a time.Time rather than a struct containing a Time.  (`type Timestamp time.Time`) But whatever, that ship has long sailed.

It looks like `github.Timestamp` was introduced in c6d364f065a1d24973bfb9b94542b6e4ab29a725 for timestamps in the repo struct.  It looks like we're already using `github.Timestamp` quite a bit in the library, though I'm curious how much of that is actually needing to parse unix timestamps versus just doing what other parts of the library are doing.  At the end of the day, I guess it probably doesn't matter too much.

Given that timestamps are almost exclusively read (I don't know of any GitHub API off the top of my head that involves providing a timestamp), and `github.Timestamp` includes all of the same methods as `time.Time`, I think this is probably fine.  I haven't reviewed all of the changes in this PR, but the general approach sounds okay.

ganeshkumarsv (approved):
LGTM!

gmlewis:
Thank you, @ganeshkumarsv !
Merging.

gmlewis:
Fixes: #2645.

PR #2650, opened 2023-01-30
Fix #2649: Fix parse token expiration

### LINKED ISSUE #2649: Fail to parse time expirations
One token i have is getting this expiration from `/rate_limit` : 

```
% export TOKEN=github_pat_*********;curl -v -H "Accept: application/vnd.github.v3+json" -H "Authorization: Bearer $TOKEN" https://api.github.com/rate_limit 2>&1|grep -n expiration
97:< github-authentication-token-expiration: 2023-04-26 23:23:18 +0200
```

the code that parse the token expiration here: 

https://github.com/google/go-github/blob/a67a565/github/github.go#L677
 
is not working for me since : 

```
2006-01-02 03:04:05 MST
```

would not parse that value, 

on another token I get the expiration token as UTC:

```
github-authentication-token-expiration: 2023-01-31 23:00:00 UTC
```
the parsing would fail as well since it's not using a AM/PM thing. A parsing like that instead in this case would work : 

```
"2006-01-02 15:04:05 MST"
```

I wonder if it's a misconfigured Github server and if we should handle all those use cases. All of those happen on public github not on public enteprise.

jerolimov:
Solves #2649 issue without breaking API changes.

1. Fixed date parsing and allow 24-hour format :smile:
2. Add support for tokens that include the timezone offset instead of the timezone, we found both in our logs :man_shrugging:

If you prefer I could also make it optional (`TokenExpiration *Timestamp`) so that it's not required to check for "is timestamp equals 0001-01-01)

Or by adding a new `type TokenExpiration time.Time` with a `func IsValid` for a token that has no expiration header?!

But both would change the public API and for my first PR here I focused on fixing just the bug without an API change.


github/github.go
```diff
@@ -571,7 +571,8 @@ type Response struct {
 	// propagate to Response.
 	Rate Rate
 
-	// token's expiration date
+	// token's expiration date. Timestamp is 0001-01-01 when token don't expire.
```
gmlewis, line 574:
```suggestion
	// token's expiration date. Timestamp is 0001-01-01 when token doesn't expire.
```

github/github.go
```diff
@@ -672,14 +673,19 @@ func parseRate(r *http.Response) Rate {
...
+// Returns 0001-01-01 if the header is not defined or could not be parsed.
 func parseTokenExpiration(r *http.Response) Timestamp {
-	var exp Timestamp
 	if v := r.Header.Get(headerTokenExpiration); v != "" {
-		if t, err := time.Parse("2006-01-02 03:04:05 MST", v); err == nil {
-			exp = Timestamp{t.Local()}
+		if t, err := time.Parse("2006-01-02 15:04:05 MST", v); err == nil {
+			return Timestamp{t.Local()}
+		}
+		// Some tokens includes the timezone offset instead of the timezone
```
gmlewis, line 682:
```suggestion
		// Some tokens include the timezone offset instead of the timezone.
```

github/github_test.go
```diff
@@ -2860,6 +2860,16 @@ func TestParseTokenExpiration(t *testing.T) {
 			header: "2021-09-03 02:34:04 UTC",
 			want:   Timestamp{time.Date(2021, time.September, 3, 2, 34, 4, 0, time.UTC)},
 		},
+		{
+			header: "2021-09-03 14:34:04 UTC",
+			want:   Timestamp{time.Date(2021, time.September, 3, 14, 34, 4, 0, time.UTC)},
+		},
+		// Some tokens includes the timezone offset instead of the timezone
```
gmlewis, line 2867:
```suggestion
		// Some tokens include the timezone offset instead of the timezone.
```

github/github_test.go
```diff
@@ -2870,8 +2880,8 @@ func TestParseTokenExpiration(t *testing.T) {
 
 		res.Header.Set(headerTokenExpiration, tt.header)
 		exp := parseTokenExpiration(res)
-		if !exp.Equal(tt.want) {
-			t.Errorf("parseTokenExpiration returned %#v, want %#v", exp, tt.want)
+		if !exp.Equal(Timestamp(tt.want)) {
+			t.Errorf("parseTokenExpiration of %#v\nreturned %#v\n    want %#v", tt.header, exp, tt.want)
```
gmlewis, line 2884:
```suggestion
			t.Errorf("parseTokenExpiration of %q\nreturned %#v\n    want %#v", tt.header, exp, tt.want)
```

gmlewis (changes requested):
Thank you, @jerolimov !
Just a few minor tweaks, please, then we should be ready for a second LGTM+Approval before merging.

chmouel:
nice one, thanks

jerolimov:
@gmlewis thanks for the feedback, updated the PR. I took everything. :+1:

gmlewis (approved):
Thank you, @jerolimov !
LGTM.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

Issue #2679, opened 2023-02-26
SCIM routes pagination


exageraldo:
I'm facing a issue when I try to paginate resources from the [SCIM provisioned identities list](https://docs.github.com/en/enterprise-cloud@latest/rest/scim?apiVersion=2022-11-28#list-scim-provisioned-identities) (`client.SCIM.ListSCIMProvisionedIdentities`) and when I try to set the number of items per page.

When I added the [httpdebug](https://github.com/gmlewis/go-httpdebug/) module, I noticed that the query strings are being sent with the first letter capitalized.

```
2023/02/25 20:37:23 curl -X GET \
  https://api.github.com/scim/v2/organizations/<ORG>/Users?Count=100&Filter=&StartIndex=1 \
  -H 'Accept: application/vnd.github.v3+json' \
  -H 'Authorization: <REDACTED>' \
  -H 'User-Agent: go-github/v50.0.0' \
  -H 'X-Github-Api-Version: 2022-11-28'
```

In the documentation the parameters start with lower case letters.

```
Query parameters
--
startIndex integer
count integer
filter string
```

And if the parameters are not sent exactly as in the documentation, the value is ignored.

This makes it impossible to paginate this resource because even if the `StartIndex` value is changed and sent, the value is not identified by the Github API.

exageraldo:
When trying to do the request in a "manual" way (`client.NewRequest` + `client.Do`), the pagination worked perfectly.

```go
func (gc *GithubClient) ManualListSamlProviderUsers(organizationName string) (allSamlUsers []*github.SCIMUserAttributes, err error) {
	startIndex := 1

	req, err := gc.client.NewRequest(
		"GET",
		fmt.Sprintf("scim/v2/organizations/%s/Users", organizationName),
		nil,
	)

	q := req.URL.Query()
	q.Add("count", fmt.Sprint(gc.perPage))
	q.Add("startIndex", fmt.Sprint(startIndex))

	for {
		req.URL.RawQuery = q.Encode()

		scimList := &github.SCIMProvisionedIdentities{}
		_, err = gc.client.Do(
			context.Background(),
			req,
			scimList,
		)

		if err != nil {
			fmt.Printf("error: %v", err)
			return nil, err
		}

		for _, user := range scimList.Resources {
			allSamlUsers = append(allSamlUsers, user)
		}

		if len(scimList.Resources) < gc.perPage {
			break
		}
		startIndex += gc.perPage
[...]

PR #2680, opened 2023-02-27
Fix SCIM query params

exageraldo:
Fixes: #2679

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -169,8 +169,22 @@ func testJSONMarshal(t *testing.T, v interface{}, want string) {
 	}
 }
 
-func testAddOptions(t *testing.T, url string, v interface{}, want string) {
+// Test whether the v fields have the url tag and the parsing of v
+// produces query parameters that corresponds to the want string.
+func testAddURLOptions(t *testing.T, url string, v interface{}, want string) {
 	t.Helper()
+
+	vt := reflect.Indirect(reflect.ValueOf(v)).Type()
+	for i := 0; i < vt.NumField(); i++ {
+		field := vt.Field(i)
+		if alias, ok := field.Tag.Lookup("url"); ok {
+			if alias == "" {
+				t.Errorf("The field %+v has a blank url tag", field)
+			}
+		} else {
+			t.Errorf("The field %+v has no url tag specified", field)
+		}
+	}
 
 	got, err := addOptions(url, v)
 	if err != nil {
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -169,6 +169,19 @@ func testJSONMarshal(t *testing.T, v interface{}, want string) {
 	}
 }
 
+func testAddOptions(t *testing.T, url string, v interface{}, want string) {
+	t.Helper()
+
+	u, err := addOptions(url, v)
+	if err != nil {
+		t.Errorf("Unable to add %#v as query parameters", v)
+	}
+
+	if u != want {
+		t.Errorf("addOptions(%q, %#v) returned %s, want %s", url, v, u, want)
+	}
+}
+
 // Test how bad options are handled. Method f under test should
 // return an error.
 func testBadOptions(t *testing.T, methodName string, f func() error) {
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -172,13 +172,15 @@ func testJSONMarshal(t *testing.T, v interface{}, want string) {
 func testAddOptions(t *testing.T, url string, v interface{}, want string) {
 	t.Helper()
 
-	u, err := addOptions(url, v)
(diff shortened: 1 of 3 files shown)


github/github_test.go
```diff
@@ -169,6 +169,19 @@ func testJSONMarshal(t *testing.T, v interface{}, want string) {
 	}
 }
 
+func testAddOptions(t *testing.T, url string, v interface{}, want string) {
+	t.Helper()
+
+	u, err := addOptions(url, v)
```
gmlewis, line 175:
```suggestion
	got, err := addOptions(url, v)
```

github/github_test.go
```diff
@@ -169,6 +169,19 @@ func testJSONMarshal(t *testing.T, v interface{}, want string) {
...
+	if err != nil {
+		t.Errorf("Unable to add %#v as query parameters", v)
+	}
+
+	if u != want {
+		t.Errorf("addOptions(%q, %#v) returned %s, want %s", url, v, u, want)
```
gmlewis, line 181:
```suggestion
		t.Errorf("addOptions(%q, %#v) returned %v, want %v", url, v, got, want)
```

github/scim_test.go
```diff
@@ -428,22 +429,31 @@ func TestUpdateAttributeForSCIMUserOptions_Marshal(t *testing.T) {
 	testJSONMarshal(t, u, want)
 }
 
-func TestListSCIMProvisionedIdentitiesOptions_Marshal(t *testing.T) {
-	testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{}`)
+func TestListSCIMProvisionedIdentitiesOptions_addOptions(t *testing.T) {
+	url := "some/path"
```
gmlewis, line 433: Can we please keep this line:

```go
testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{}`)
```

for consistency?  We always test to see how an empty struct gets mashaled.

gmlewis (changes requested):

github/scim_test.go
exageraldo: Done! ✅

exageraldo: Since the `ListSCIMProvisionedIdentitiesOptions` struct has no json tag (url only), the string/json fields should be the same as the structure fields, and no field should be omitted if empty.

I think the expected value should be something like this:

```go
testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{
	"StartIndex": null,
	"Count": null,
	"Filter": null
}`)
```
I just created a new issue so we can talk more about this subject. #2699

github/github_test.go
```diff
@@ -169,6 +169,19 @@ func testJSONMarshal(t *testing.T, v interface{}, want string) {
 	}
 }
 
+func testAddOptions(t *testing.T, url string, v interface{}, want string) {
```
exageraldo, line 172: Since the structures passed in must have "url" tags in the fields (more in [addOptions](https://github.com/google/go-github/blob/master/github/github.go#L286)), is it valid to test if the fields have this tag?

Something like this:

```go
st := reflect.TypeOf(v)
for i := 0; i < st.NumField(); i++ {
	field := st.Field(i)
	if alias, ok := field.Tag.Lookup("url"); ok {
		if alias == "" {
			// t.Errorf("...")
		}
	} else {
		// t.Errorf("...")
	}
}
```

gmlewis: Sure.  Please add a comment before line 172 that describes this requirement so that people will not start adding `testAddOptions` to every test.

In fact, let's make the function name more descriptive.  How about `testAddURLOptions` ?

gmlewis: Or maybe `testAdURLTags` ?

github/scim_test.go
```diff
@@ -428,22 +429,37 @@ func TestUpdateAttributeForSCIMUserOptions_Marshal(t *testing.T) {
 	testJSONMarshal(t, u, want)
 }
 
-func TestListSCIMProvisionedIdentitiesOptions_Marshal(t *testing.T) {
-	testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{}`)
+func TestListSCIMProvisionedIdentitiesOptions_addOptions(t *testing.T) {
+	testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{
```
gmlewis, line 433: OK, now I understand why you removed this before, as it really is no longer relevant.
Feel free to delete this if you wish, either in this PR or in your upcoming one.

gmlewis (approved):
Thank you, @exageraldo !
LGTM.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

Issue #2363, opened 2022-05-11
Does this package support configuring a proxy?


songleo:
I want to use this package to au-create the issue in my repo. but my env is private, and we provide a proxy to access the internet. so I am not sure if I can configure a proxy with this package to call my github repo. thanks.

gmlewis:
I believe this is possible, but don't have the details. Hopefully others can comment and provide you with what you need.
In the meantime, if you figure it out, feel free to report back to this issue with the solution to benefit others.  Thanks.

songleo:
@gmlewis thanks, for now, I have no idea to do it. if I figure it out, I will post it here. thanks.

extravio:
You need to configure the proxy in a new http.client   and pass it as a parameter to your "github client"

```go

package main

import (
	"context"
    "fmt"
    "log"
    "github.com/google/go-github/v45/github"
	"net/http"
	"net/url"
)

func main() {
	//creating the proxyURL
	proxyStr := "<your-proxy-url>"
	proxyURL, err := url.Parse(proxyStr)
	if err != nil {
		log.Println(err)
	}

	transport := &http.Transport{
		Proxy: http.ProxyURL(proxyURL),
	}

	//adding the Transport object to the http Client
	httpclient := &http.Client{
		Transport: transport,
	}

	client := github.NewClient(httpclient)

	// list public repositories for org "github"
	opt := &github.RepositoryListByOrgOptions{Type: "public"}
	repos, _, err := client.Repositories.ListByOrg(context.Background(), "github", opt)
	if err != nil {
        log.Fatal(err)
    }

    // If no error was returned, print the returned message
    // to the console.
    fmt.Println(repos)
}

```

gmlewis:
Thank you, @extravio !

Closing due to solution provided above: https://github.com/google/go-github/issues/2363#issuecomment-1138153077

PR #2686, opened 2023-03-03
Support HTTP Proxy from environment

hangscer8:
Add the support of http prxoy to `NewClientWithEnvProxy`.

Relates to: #2363.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -348,6 +348,11 @@ func NewClient(httpClient *http.Client) *Client {
 	return c
 }
 
+// NewClientWithEnvProxy enhances NewClient with the HttpProxy env.
+func NewClientWithEnvProxy() *Client {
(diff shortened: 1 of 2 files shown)


gmlewis:
@hangscer8 - I'm concerned that this change is going to affect current users of this repo who are not expecting this behavior to be the default.

Could we instead add a new `NewClientWithEnvProxy` or something like that?

hangscer8:
> @hangscer8 - I'm concerned that this change is going to affect current users of this repo who are not expecting this behavior to be the default.
> 
> Could we instead add a new `NewClientWithEnvProxy` or something like that?

OK, I will fix that. @gmlewis

gmlewis (approved):
Thank you, @hangscer8 !
LGTM.

gmlewis (approved):

valbeat (approved):
LGTM!

gmlewis:
> LGTM!

Thank you, @valbeat !
Merging.

Issue #2687, opened 2023-03-03
List workflow runs for a repository by SHA


alexellis:
Hi, I want to trigger a workflow for a given SHA, then query the ID of the run, to eventually collect the logs and status.

This is available in the API now, however the latest version of go-github doesn't have it available in the struct yet.
 
https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository

```go
// ListWorkflowRunsOptions specifies optional parameters to ListWorkflowRuns.
type ListWorkflowRunsOptions struct {
	Actor   string `url:"actor,omitempty"`
	Branch  string `url:"branch,omitempty"`
	Event   string `url:"event,omitempty"`
	Status  string `url:"status,omitempty"`
	Created string `url:"created,omitempty"`
	ListOptions
}
```

I don't know how to contribute to this library, or whether external contributions are welcome/harder for the maintainers than adding this themselves.

What would be a good next step to get this in?

gmlewis:
> I don't know how to contribute to this library, or whether external contributions are welcome/harder for the maintainers than adding this themselves.

We welcome and encourage people to contribute to this repo.  Please start by reading CONTRIBUTING.md, then go ahead and submit your PR.

Thank you, @alexellis !  We look forward to your contributions!

PR #2703, opened 2023-03-08
Add head_sha for ListWorkflowRunsOptions

alexellis:
## Description

Add head_sha for ListWorkflowRunsOptions (fixes: #2687)

## Why is this needed?

GitHub added this option to the following API:

/repos/{owner}/{repo}/actions/runs

Only returns workflow runs that are associated with the specified head_sha.

See also:

https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository

### DIFF
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -60,6 +60,7 @@ type ListWorkflowRunsOptions struct {
 	Event   string `url:"event,omitempty"`
 	Status  string `url:"status,omitempty"`
 	Created string `url:"created,omitempty"`
+	HeadSHA string `url:"head_sha,omitempty"`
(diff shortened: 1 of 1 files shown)


alexellis:
@gmlewis I've signed the CLA, and read the contributing guide, ran the unit tests through.

Let me know if you need anything else.

gmlewis (approved):
Thank you, @alexellis !
LGTM.
Merging.

alexellis:
Thanks for accepting the contribution @gmlewis 

I used to write interactions with GitHub's API [manually](https://github.com/alexellis/derek/blob/master/main.go#L70) before discovering this client.

Issue #2699, opened 2023-03-06
about testJSONMarshal's behavior


exageraldo:
When I was investigating issue #2679, I faced the following situation:

After changing the tags in the [`ListSCIMProvisionedIdentitiesOptions`](https://github.com/google/go-github/pull/2680/files#diff-c2dda2100fd0c700da11a5d96582391b4923658db7e717627d248ae1c029c5fbL71) structure, from `json` to `url`, I was expecting the [`TestListSCIMProvisionedIdentitiesOptions_Marshal`](https://github.com/google/go-github/pull/2680/files#diff-2723dd506a61e0f7abe849217ded7c4aa5dfef10455f14ed10a423c65f4dac49L431) test to fail, but it didn't.

```go
func TestListSCIMProvisionedIdentitiesOptions_Marshal(t *testing.T) {
	testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{}`)

	u := &ListSCIMProvisionedIdentitiesOptions{
		StartIndex: Int(1),
		Count:      Int(10),
		Filter:     String("test"),
	}

	want := `{
		"startIndex": 1,
		"count": 10,
	 	"filter": "test"
	}`

	testJSONMarshal(t, u, want)
}
```

It made sense that this test was passing before the change, because the structure was in accordance with the defined tags. However, after the change the tests kept passing.

```go
// BEFORE
type ListSCIMProvisionedIdentitiesOptions struct {
	StartIndex *int    `json:"startIndex,omitempty"`
	Count      *int    `json:"count,omitempty"` 
	Filter     *string `json:"filter,omitempty"`
}

// AFTER
type ListSCIMProvisionedIdentitiesOptions struct {
	StartIndex *int    `url:"startIndex,omitempty"`
	Count      *int    `url:"count,omitempty"` 
	Filter     *string `url:"filter,omitempty"`
}
```

If we add a few more test cases, we can see that regardless of how the letters are (upper or lower case), the tests always pass (both using the `json` and `url` tags).

I created a temporary function with the test cases that should fail, the `TestListSCIMProvisionedIdentitiesOptions_Marshal_MustFailed` function, but all are passing so far.

```go
func TestListSCIMProvisionedIdentitiesOptions_Marshal(t *testing.T) {
	// Case one: Empty struct
	testJSONMarshal(t, &ListSCIMProvisionedIdentitiesOptions{}, `{}`)

	// Case two: Happy case. The fields are as described in the structure tags
[...]

PR #2708, opened 2023-03-15
Change testJSONMarshal

exageraldo:
Fixes: #2699


gmlewis:
Thanks, @exageraldo ! Please remove "Draft" status when you would like me to review this PR.

github/actions_artifacts_test.go
```diff
@@ -453,25 +453,19 @@ func TestArtifact_Marshal(t *testing.T) {
...
+		`"name":"n",`,
+		`"size_in_bytes":1,`,
+		`"url":"u",`,
+		`"archive_download_url":"a",`,
+		`"expired":false,`,
+		`"created_at":`+referenceTimeStr+`,`,
+		`"updated_at":`+referenceTimeStr+`,`,
+		`"expires_at":`+referenceTimeStr+`,`,
+		`"workflow_run":{"id":1,"repository_id":1,"head_repository_id":1,"head_branch":"b","head_sha":"s"}}`,
+	)
```
exageraldo, line 468: @gmlewis  I was thinking of structuring the expected strings this way. I think it will be easier (or at least less difficult) to read, understand and maintain these values.

What do you think about this?

gmlewis: Why not just keep all of the line breaks but remove all but the first and last backticks?

That, to me, seems like a lot less noisy clutter.

exageraldo: It sounds great!

We need to remove tabs and line breaks. It goes something like this:

```go
s = strings.Replace(s, "\t", "", -1)
s = strings.Replace(s, "\n", "", -1)
```

Do you think we can put it inside the `testJSONMarshal` function, so that all values passed are already "corrected"?

gmlewis: Hmm... that's an interesting idea... How about trying it out and seeing if it can handle all the cases we have?

exageraldo: I tried it and all the tests kept passing! 🟢 
I will now apply it to all the other test cases.

github/activity_notifications.go
```diff
@@ -97,15 +97,15 @@ func (s *ActivityService) ListRepositoryNotifications(ctx context.Context, owner
 }
 
 type markReadOptions struct {
-	LastReadAt Timestamp `json:"last_read_at,omitempty"`
+	LastReadAt *Timestamp `json:"last_read_at,omitempty"`
```
exageraldo, line 100: @gmlewis  I found this case where the documentation says that the value can be omitted. I think this should be a pointer (as well as other cases). And also the "read" field is missing. (Github docs: https://docs.github.com/en/rest/activity#mark-notifications-as-read)

In going through the tests I found some interesting cases. Do you think it is better if I open an issue for each one, to talk more about each one?

github/github.go
```diff
@@ -981,8 +981,8 @@ func (r *TwoFactorAuthError) Error() string { return (*ErrorResponse)(r).Error()
 // RateLimitError occurs when GitHub returns 403 Forbidden response with a rate limit
 // remaining value of 0.
 type RateLimitError struct {
-	Rate     Rate           // Rate specifies last known rate limit for the client
-	Response *http.Response // HTTP response that caused this error
+	Rate     Rate           `json:"rate"`    // Rate specifies last known rate limit for the client
+	Response *http.Response `json:"-"`       // HTTP response that caused this error
 	Message  string         `json:"message"` // error message
 }
```
exageraldo, line 987: @gmlewis  Is it really needed to create "marshal" test for structures like this and the `AbuseRateLimitError`, or `Rate`?

As far as I understand it, it is not a structure that represents anything we send or receive from GitHub.

gmlewis: Probably not then... let's just please be super-careful to try and imagine ways that modifying things like this might affect end-users of this client library.  If it breaks things, we need to think long and carefully about whether it is worth the change since we have thousands of users of this client library in the wild.

exageraldo: I left the structures as they were and removed the two tests `TestRateLimitError_Marshal` and `TestRateLimitError_Marshal`, as I believe they do not provide much value.

What do you think about this idea?

exageraldo:
Some tests are no longer `_Marshal` but `_addOtions`, because the structures are responsible for mounting query strings, not a json.

```
TestGetAuditLogOptions_addOptions
TestTrafficBreakdownOptions_addOptions
TestSearchOptions_addOptions
TestTeamListTeamMembersOptions_addOptions
TestListExternalGroupsOptions_addOptions
TestUserListOptions_addOptions
TestHovercardOptions_addOptions
```

Since we do a tag validation in the `testAddURLOptions` function, I had to make some small changes to the following structures for the tests to pass.

```
GetAuditLogOptions ~> ListCursorOptions `url:",omitempty"`
SearchOptions ~> ListOptions `url:",omitempty"`
TeamListTeamMembersOptions ~> ListOptions `url:",omitempty"`
ListExternalGroupsOptions ~> ListOptions `url:",omitempty"`
UserListOptions ~> ListOptions `url:",omitempty"`
```

Theoretically it is just an explicit way of leaving the structure as it already was (did that make any sense?)

exageraldo:
In some cases, I had to use `json.Marshal` to mount part of the expected string.

[github/event_types_test.go](https://github.com/exageraldo/go-github/blob/c45dba4f725ee7fb392147543c13fbe140ddf579/github/event_types_test.go#L14872)

```go
part, _ := json.Marshal(">= 2.0.0, < 2.0.2")
```
[github/teams_discussions_test.go](https://github.com/exageraldo/go-github/blob/c45dba4f725ee7fb392147543c13fbe140ddf579/github/teams_discussions_test.go#L588)

```go
bodyHTML, _ := json.Marshal(`<p>test</p>`)
```

In both cases, the idea is just to convert "`>`" into "`\u003e`" and "`<`" into "`\u003c`".

Do you think we can keep it that way, or can we add two other `strings.Replace` inside the `testJSONMarshal` function?

```go
want = strings.Replace(want, ">", "\u003e", -1)
want = strings.Replace(want, "<", "\u003c", -1)
```

gmlewis:
Yeah, those cases look a bit odd. 
Go ahead with what you think is best, and then when I take the time to review the whole PR, I might come up with other ideas. 😄

gmlewis:
We'll give this PR a couple weeks to get a reply and have the conflicts resolved, then if we haven't got any updates, it will be closed as abandoned.

exageraldo:
I'll fix these things by this weekend 😄

gmlewis:
Awesome - @exageraldo - thank you for the update!
No rush, actually - I just need to clean out old, abandoned PRs periodically.

exageraldo:
hey, sorry for the delay. ~i had some unforeseen circumstances here, but~ I'm already finalizing it! by monday I'll be pushing up the changes.

gmlewis:
Why are you changing GitHub workflows?

exageraldo:
I didn't change anything in the workflows. I ended up opting for a "merge" instead of a "rebase" because some changes were disappearing and I didn't know how to solve it. If you look in the "Files Changed" tab, you'll see that all the changes are related to tests only.

gmlewis:
OK, I must have been looking only at a merge commit... looking now and it is looking much better, thanks.
Performing code review...

github/orgs_audit_log.go
```diff
@@ -16,7 +16,7 @@ type GetAuditLogOptions struct {
 	Include *string `url:"include,omitempty"` // Event type includes. Can be one of "web", "git", "all". Default: "web". (Optional.)
 	Order   *string `url:"order,omitempty"`   // The order of audit log events. Can be one of "asc" or "desc". Default: "desc". (Optional.)
 
-	ListCursorOptions
+	ListCursorOptions `url:",omitempty"`
```
gmlewis, line 19: This seems odd - why is this necessary for an embedded struct?

github/search.go
```diff
@@ -54,7 +54,7 @@ type SearchOptions struct {
 	// Whether to retrieve text match metadata with a query
 	TextMatch bool `url:"-"`
 
-	ListOptions
+	ListOptions `url:",omitempty"`
```
gmlewis, line 57: This also seems odd - why is this needed?

github/users.go
```diff
@@ -196,7 +196,7 @@ type UserListOptions struct {
 	// Note: Pagination is powered exclusively by the Since parameter,
 	// ListOptions.Page has no effect.
 	// ListOptions.PerPage controls an undocumented GitHub API parameter.
-	ListOptions
+	ListOptions `url:",omitempty"`
```
gmlewis, line 199: ditto

gmlewis (commented):
Wow, thank you, @exageraldo !
A have a few questions...

github/github_test.go
```diff
@@ -2705,53 +2733,20 @@ func TestErrorBlock_Marshal(t *testing.T) {
...
-	}
-
-	want := `{
-		"message": "msg"
+		"reason":"reason",
+		"created_at":` + referenceTimeStr + `
 	}`
 
 	testJSONMarshal(t, u, want)
 }
```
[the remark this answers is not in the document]
exageraldo: wow. It wasn't supposed to happen.
I think it was when I sync the branch. I'll add it again!

WillAbides:
We could simplify this by changing assertJSONMarshal to this:

```go

// Test whether the marshaling of v produces JSON that corresponds
// to the want string.
func testJSONMarshal(t *testing.T, v interface{}, want string) {
	t.Helper()
	got, err := json.Marshal(v)
	if err != nil {
	}
	got = normalizeJSON(t, got)
	wantBytes := normalizeJSON(t, []byte(want))
	diff := cmp.Diff(string(wantBytes), string(got))
	if diff != "" {
		t.Errorf("json.Marshal returned:\n%s\nwant:\n%s\ndiff:\n%v", string(got), string(wantBytes), diff)
	}
}

// normalizeJSON normalizes the JSON in b by unmarshaling and marshaling it
// again.
func normalizeJSON(t *testing.T, b []byte) []byte {
	t.Helper()
	var v interface{}
	err := json.Unmarshal(b, &v)
	if err != nil {
		t.Errorf("Unable to unmarshal JSON for %v: %v", string(b), err)
	}
	w, err := json.MarshalIndent(v, "", "  ")
	if err != nil {
	}
	return w
}

```

That way differences like casing are failures but it isn't required to have the same whitespace or order of fields. This is similar to what testify does with `assert.JSONEq()`.

gmlewis:

exageraldo:
I'll organize the branch to push up something like @WillAbides suggested!

exageraldo:
@WillAbides 's solution worked very well! Thanks a lot!
I "messed up" this branch a lot and ended up creating another one here https://github.com/google/go-github/pull/3519
Now I'm working on fixing all the tests that broke.

Maybe we can close this PR.

gmlewis:
> @WillAbides 's solution worked very well! Thanks a lot! I "messed up" this branch a lot and ended up creating another one here #3519 Now I'm working on fixing all the tests that broke.
> 
> Maybe we can close this PR.

SGTM.
Closing as obsolete.

PR #2728, opened 2023-03-30
Add support for deleting an org

### LINKED ISSUE #2726: Add support for deleting an organization.
A new REST endpoint was added to delete an organization at `/orgs/{org}'`. The documentation doesn't seem to have gone live quite yet but I'll update the issue as soon as I see it. In the meantime the endpoint can be evoked like:

`curl -X DELETE https://api.github.com/orgs/:org`

Now with [documentation](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#delete-an-organization)!

tjcorr:
Fixes: #2726.

### DIFF
--- a/github/orgs.go
+++ b/github/orgs.go
@@ -262,6 +262,19 @@ func (s *OrganizationsService) Edit(ctx context.Context, name string, org *Organ
 	return o, resp, nil
 }
 
+// Delete an organization by name.
+//
+// GitHub API docs: https://docs.github.com/en/rest/orgs/orgs#delete-an-organization
+func (s *OrganizationsService) Delete(ctx context.Context, org string) (*Response, error) {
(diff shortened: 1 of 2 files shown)


gmlewis (changes requested):
~Since we use only a single client for all APIs in this repo and therefore we don't have the stutter problem, could we please clarify the method name by calling it `DeleteOrg`?~

tjcorr:
> Since we use only a single client for all APIs in this repo and therefore we don't have the stutter problem, could we please clarify the method name by calling it `DeleteOrg`?

I was just matching the format I saw in other areas like: https://github.com/google/go-github/blob/master/github/repos.go#L590-L601

gmlewis (approved):
Thank you, @tjcorr !
LGTM.

valbeat (approved):
LGTM!

PR #2732, opened 2023-03-31
verify payload signature if present

### LINKED ISSUE #2731: ValidatePayload should not accept nil or empty secretToken slice
https://github.com/google/go-github/pull/1127 added support for calling `ValidatePayload` with a `nil` or empty `secretToken` slice. The justification for this change is explained in the [current docs](https://pkg.go.dev/github.com/google/go-github/v45@v45.2.0/github#ValidatePayload):

> If your webhook does not contain a secret token, you can pass nil or an empty slice. This is intended for local development purposes only and all webhooks should ideally set up a secret token.

I think this leaves `ValidatePayload` with dangerous default behaviour. Because now every caller of this function has to check that they have a non-`nil` and non-empty `secretToken` slice if they want to validate a payload. If they don't, it's possible to accidentally call the function with a `nil` or empty slice (bad configuration, etc) and never know about it. This is not a safe default. Better would be to have the function `panic` or return an error in case `secretToken` is `nil` or empty - because such a value can never be used to validate a payload, which is obviously what the caller intended to do in calling the function.

The same applies to `ValidatePayloadFromBody()`. 

I haven't looked at other functions/methods.

Returning to the goal of #1126, I would instead have pushed back and suggested that developers who are in development mode change their calling code to simply not call `ValidatePayload` when they detect they are in such a mode.

willnorris:
Verify the payload signature if the request has a signature present in HTTP headers, or if a non-empty secretToken is passed to the ValidatePayload method, indicating that a signature is expected.

This modifies the behavior added in #1127, but not the spirit of what was requested in #1126, which is to support webhooks that don't have configured secrets.

Specifically, this no longer allows signature checking to be skipped entirely, even for webhooks with a configured secret, simply by passing an empty secretToken to ValidatePayload.

This also cleans up some documentation and unused fields in the accompanying test.

Fixes #2731 

/cc @myitcv

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -148,13 +148,13 @@ func messageMAC(signature string) ([]byte, func() hash.Hash, error) {
 	return buf, hashFunc, nil
 }
 
-// ValidatePayload validates an incoming GitHub Webhook event request body
+// ValidatePayloadFromBody validates an incoming GitHub Webhook event request body
 // and returns the (JSON) payload.
 // The Content-Type header of the payload can be "application/json" or "application/x-www-form-urlencoded".
 // If the Content-Type is neither then an error is returned.
 // secretToken is the GitHub Webhook secret token.
-// If your webhook does not contain a secret token, you can pass nil or an empty slice.
-// This is intended for local development purposes only and all webhooks should ideally set up a secret token.
+// If your webhook does not contain a secret token, you can pass an empty secretToken.
+// Webhooks without a secret token are not secure and should be avoided.
 //
 // Example usage:
 //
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @willnorris !
LGTM.

willnorris:

I don't think a second approval should be necessarily when the change is coming from another maintainer... they are basically the second :) At least, that's how I recall internal Google systems handling it.  (But I also don't recall when that review policy got added to the process and what its purpose was.)

gmlewis:
Oh, sorry. Just habit. Thank you, @willnorris !
Merging.

Issue #2736, opened 2023-04-05
Failure unmarshalling organization with large PrivateRepos number in plan


shaines1:
Hello,

I am attempting to use the [integrations/github](https://registry.terraform.io/providers/integrations/github) Terraform provider with GitHub Enterprise Server. Executing `terraform plan` and `terraform apply` with infrastructure definitions that require interacting with organizations hosted in the GHES instance result in the following go-github error:

```
│ Error: json: cannot unmarshal number 999999999999 into Go struct field Plan.plan.private_repos of type int
│
│   with data.github_organization.architecture,
│   on main.tf line 14, in data "github_organization" "myorg":
│   14: data "github_organization" "myorg" {
```

On inspection of the organization JSON representation, the plan.private_repos has a value of 999999999999. My assumption is that the unmarshalling error is due to the [PrivateRepos variable in go-github only being declared as an int](https://github.com/google/go-github/blob/master/github/orgs.go#L124). 

Here's an abbreviated snippet of the organization JSON output:

```json
{
  "login": "myorg",
  "id": 1,
  "plan": {
    "name": "enterprise",
    "space": 1,
    "private_repos": 999999999999,
    "filled_seats": 1,
    "seats": 0
  }
}
```

Would it be possible to modify this data type? Please let me know if there's anything else that would be helpful to provide!

### Environment

* GitHub Enterprise Server: 3.8.0
* Terraform: 1.4.2
* integrations/github: 5.20
* go-github: v50.2.0

gmlewis:
Something seems fishy here...  have you looked at the JSON output directly from the server (using `curl` for example)?
Is it really sending that number?!?  Doesn't that seem like a GitHub problem?  Have you contacted GitHub tech support?

shaines1:
> Something seems fishy here...

I agree

> have you looked at the JSON output directly from the server (using `curl` for example)? Is it really sending that number?!?

Yes. Using an PAT as an org owner to `https://<base_url>/api/v3/orgs/<org_name>` with Postman returns the same result. It doesn't _appear_ to be a permission issue, as the remainder of the response looks correct

![image](https://user-images.githubusercontent.com/16926752/230096407-0e83993d-9e95-4676-992f-0333d9bdd404.png)

> Doesn't that seem like a GitHub problem? Have you contacted GitHub tech support?

Yes, my colleague has. I'll update here with their response when we receive it

MISO-mgriffin:
[...]

PR #2738, opened 2023-04-05
Change PrivateRepos from int to int64

gmlewis:
Fixes: #2736.


shaines1:
Thank you @gmlewis 

LGTM!

Issue #2741, opened 2023-04-07
Add Topics to PushEventRepository


ecrupper:
I would like to leverage the `topics` field found in the `repository` object of several webhook event payloads. I noticed that for the `push` event, which has a [special repo struct](https://github.com/google/go-github/blob/c96ba417f7b022bee92bc04d831e89fd9fc70edc/github/event_types.go#L1028-L1065), the `Topics` field was missing. 

I plan on implementing this enhancement the same way the standard [repository](https://github.com/google/go-github/blob/c96ba417f7b022bee92bc04d831e89fd9fc70edc/github/repos.go#L78) struct does it, unless there are any objections. I am curious why it is of type `[]string` and not `*[]string` though, if anyone would be willing to clarify that for me? It doesn't get an accessor generated as a result.

PR #2742, opened 2023-04-07
Add Topics to PushEventRepository

ecrupper:
Closes https://github.com/google/go-github/issues/2741

Adding the `Topics` field, which is included in push event payloads under `repository`


gmlewis (approved):
Thank you, @ecrupper !
LGTM.
Merging.

Issue #2749, opened 2023-04-12
Add `Owner` to `EditChange` struct for repository transfer webhook event payloads


ecrupper:
When a repository has a transfer of ownership, it will send a [repository event webhook](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=transferred#repository). The structure ends up looking like this:

```json
{
  "action": "transferred",
  "changes": {
    "owner": {
      "from": {
        "user": {
          "login": "octocat",
          ... etc
``` 
OR
```json
{
  "action": "transferred",
  "changes": {
    "owner": {
      "from": {
        "organization": {
          "login": "github",
          ... etc
```

I'd like to add the missing `Owner` field to the [`EditChange`](https://github.com/google/go-github/blob/683c2d7d7d40e2018285d356d0a45abaa5a9c286/github/event_types.go#L330-L335) struct. I was thinking it would look something like this:

```go
type EditChange struct {
	Title *EditTitle `json:"title,omitempty"`
	Body  *EditBody  `json:"body,omitempty"`
	Base  *EditBase  `json:"base,omitempty"`
	Repo  *EditRepo  `json:"repository,omitempty"`
	Owner *EditOwner `json:"owner,omitempty"`
}

type EditOwner struct {
	OwnerInfo *OwnerInfo `json:"from,omitempty"`
}

type OwnerInfo struct {
	Actor *User `json:"user,omitempty"`
	Org   *User `json:"organization,omitempty"`
}
```

Unfortunately, due to the structure of the payload, `OwnerInfo` will have to account for those two different options, even though they both refer to [owners](https://github.com/google/go-github/blob/683c2d7d7d40e2018285d356d0a45abaa5a9c286/github/repos.go#L31).

gmlewis:
Thank you, @ecrupper - that sounds reasonable.

PR #2750, opened 2023-04-12
Add `Owner` to `EditChange` struct

ecrupper:
Closes https://github.com/google/go-github/issues/2749

Let me know if you have a better idea for struct names.

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -368,7 +368,7 @@ type EditOwner struct {
 
 // OwnerInfo represents the account info of the owner of the repo (could be User or Organization but both are User structs).
 type OwnerInfo struct {
-	Actor *User `json:"user,omitempty"`
+	User *User `json:"user,omitempty"`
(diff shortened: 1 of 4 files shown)


github/event_types.go
```diff
@@ -360,6 +361,17 @@ type EditRepo struct {
...
 }
 
+// EditOwner represents a change of repository ownership.
+type EditOwner struct {
+	OwnerInfo *OwnerInfo `json:"from,omitempty"`
+}
+
+// OwnerInfo represents the account info of the owner of the repo (could be User or Organization but both are User structs).
+type OwnerInfo struct {
+	Actor *User `json:"user,omitempty"`
```
gmlewis, line 371: I think the following would be less confusing or at least more consistent.

```suggestion
	User *User `json:"user,omitempty"`
```

gmlewis (approved):
Thank you, @ecrupper !
LGTM.
Merging after tests pass.

Issue #2756, opened 2023-04-14
Org Required workflow create and update api's should return the workflow object


avivek:
The api for required workflow, for both create and update , returns the created workflow object in the response.
Currently the code for these do not return the create object back in an object instead only return the JSON version of these objects.
It would be good to return a typed object as this would enable us to get the details of the created workflow such as name, id etc.

gmlewis:
Is #2755 related?

avivek:
Hi @gmlewis ,

This is for  CreateRequiredWorkflow and UpdateRequiredWorkflow methods in the actions_required_workflows file.
I dont think this is related to 2755.

avivek:
Hi @gmlewis ,

I have raised PR#2759 for this.
Pls take a look.
Thanks.

PR #2758, opened 2023-04-14
create and update api's should return the workflow object

avivek:
Updated create and update org required workflow methods to return the required workflow object returned by the github api.
Merging this Fixes https://github.com/google/go-github/issues/2756


avivek:
Sorry for confusion, Replaced with https://github.com/google/go-github/pull/2759

Issue #2425, opened 2022-07-29
PushEvent GetCommits() Method


midnightconman:
I would like to use an interface for this event in a local implementation, but unfortunately a getter for this type doesn't exist. Can we create a getter for `PushEvent.GetCommits`?

```
type PushEvent struct {
...
Commits      []*HeadCommit
...
}

func (p *PushEvent) GetCommits() []*HeadCommit {
    return p.Commits
}
```

gmlewis:
In general, getters are only provided for anything where a `nil` might be a problem (which is not the case for slices of anything), and all getters are auto-generated.

So if you want to add one for this, then I suppose we have a few options:

1. add getters for all fields, but that seems wasteful
2. modify the auto-generator to add a list of "special" getters that we want to always add (e.g. `PushEvent.Commits`)
3. manually add a getter to the main source code and completely ignore the auto-generator

I'm kinda leaning toward option 2.

Thoughts?

midnightconman:
2 sounds good to me 😸 Let me take a look at the auto-generator code. Thank you!

olibaa:
I would like to propose some changes and request for a review. 
- https://github.com/google/go-github/pull/2776

I would appreciate it if you could check it out!

PR #1867, opened 2021-05-07
Support map type in go generate

gmlewis:
Fixes: #1866.

PR #2776, opened 2023-04-30
Support array type in go generate with whitelist

olibaa:
Fixes: https://github.com/google/go-github/issues/2425

This PR is similar to the following PR:
- https://github.com/google/go-github/pull/1867

### DIFF
--- a/github/gen-accessors.go
+++ b/github/gen-accessors.go
@@ -51,6 +51,11 @@ var (
 	skipStructs = map[string]bool{
 		"Client": true,
 	}
+
+	// whitelistSliceGetters lists "struct.field" to add getter method
+	whitelistSliceGetters = map[string]bool{
+		"PushEvent.Commits": true,
+	}
 )
 
 func logf(fmt string, args ...interface{}) {
@@ -137,6 +142,13 @@ func (t *templateData) processAST(f *ast.File) error {
 					case *ast.MapType:
 						t.addMapType(x, ts.Name.String(), fieldName.String(), false)
 						continue
+					case *ast.ArrayType:
+						if key := fmt.Sprintf("%v.%v", ts.Name, fieldName); whitelistSliceGetters[key] {
+							logf("Method %v is whitelist; adding getter method.", key)
+							t.addArrayType(x, ts.Name.String(), fieldName.String(), false)
+							continue
+						}
+						continue
 					}
 
 					logf("Skipping field type %T, fieldName=%v", field.Type, fieldName)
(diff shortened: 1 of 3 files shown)


github/gen-accessors.go
```diff
@@ -356,8 +369,8 @@ func ({{.ReceiverVar}} *{{.ReceiverType}}) Get{{.FieldName}}() *{{.FieldType}} {
   }
   return {{.ReceiverVar}}.{{.FieldName}}
 }
-{{else if .MapType}}
-// Get{{.FieldName}} returns the {{.FieldName}} map if it's non-nil, an empty map otherwise.
+{{else if or .MapType .ArrayType }}
+// Get{{.FieldName}} returns the {{.FieldName}} {{if .MapType}}map{{else if .ArrayType }}array{{end}} if it's non-nil, an empty map otherwise.
```
gmlewis, line 373:
```suggestion
// Get{{.FieldName}} returns the {{.FieldName}} {{if .MapType}}map{{else if .ArrayType }}slice{{end}} if it's non-nil, {{if .MapType}}an empty map{{else if .ArrayType }}nil{{end}} otherwise.
```

gmlewis (changes requested):
While I appreciate what you are trying to do, @olibaa, this is not how I think it should be done.

Please read very carefully what I wrote [here](https://github.com/google/go-github/issues/2425#issuecomment-1200035340). 
I do not want to add getters for every possible slice.
They are simply unnecessary and add a huge amount of bloat for zero benefit.

What I recommend here is that you add a whitelist of fields that your code should process.
So, for example, we have only a single field currently that we want to add to the whitelist: `PushEvent.Commits`.

Then, on lines 140-142 of github/gen-accessors.go, you would check this whitelist and only call `t.addArrayType` if the current field matches an item in the whitelist (which itself could simply be a `var whitelistSliceGetters = map[string]bool{ ...: true}` to make lookup super-easy).

That way, when we look at the diffs of github/github-accessors.go and github/github-accessors_test.go, we should only see one new field in this PR... the one for `PushEvent.Commits`.

How does that sound to you?

olibaa:
@gmlewis
Sounds good!
I will fix the code.

github/gen-accessors.go
```diff
@@ -137,6 +137,9 @@ func (t *templateData) processAST(f *ast.File) error {
 					case *ast.MapType:
 						t.addMapType(x, ts.Name.String(), fieldName.String(), false)
 						continue
+					case *ast.ArrayType:
```
olibaa, line 140: @gmlewis 
I add my whitelist here

Is my understanding correct?

gmlewis: Yes, you check your whitelist here. If it matches, then you go ahead with the operation, otherwise you ignore it.

olibaa: thanks for answering !

olibaa: @gmlewis 
Fixed!
I would like a review 🙇‍♂️

gmlewis (approved):
Beautiful, @olibaa!  Well done!
LGTM.

valbeat (approved):

gmlewis:
Thank you, @valbeat !
Merging.

Issue #2789, opened 2023-05-31
Support for rulesets API


liaodaniel:
GitHub has recently released rule sets at both the organization and repository level.

References:
* https://docs.github.com/en/rest/orgs/rules?apiVersion=2022-11-28
* https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28

Would be nice to have support for this feature.

gmlewis:
Would you like to submit a PR or would you rather that we open this issue up to another contributor to this repo?

liaodaniel:
Happy to take a look at this one.

gmlewis:
> Happy to take a look at this one.

Great! Thank you, @liaodaniel.
It's yours.

gmlewis:
> Happy to take a look at this one.

Great! Thank you, @liaodaniel.
It's yours.

PR #2795, opened 2023-06-02
Implement rulesets

liaodaniel:
This PR includes both the organization and repository rulesets API.

Resolves https://github.com/google/go-github/issues/2789

Signed-off-by: Daniel Liao [10663736+liaodaniel@users.noreply.github.com](mailto:10663736+liaodaniel@users.noreply.github.com)


liaodaniel:
Linter appears to be failing due to https://github.com/golangci/golangci-lint/issues/3862. Guess I'll wait until they make a fix for it

github/orgs_rules.go
```diff
@@ -0,0 +1,105 @@
...
+// GitHub API docs: https://docs.github.com/en/rest/orgs/rules#get-an-organization-repository-ruleset
+func (s *OrganizationsService) GetOrganizationRuleset(ctx context.Context, org string, rulesetID int64) (*Ruleset, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/rulesets/%v", org, rulesetID)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var rulesets *Ruleset
```
gmlewis, line 64:
```suggestion
	var ruleset *Ruleset
```

github/orgs_rules.go
```diff
@@ -0,0 +1,105 @@
...
+// GitHub API docs: https://docs.github.com/en/rest/orgs/rules#update-an-organization-repository-ruleset
+func (s *OrganizationsService) UpdateOrganizationRuleset(ctx context.Context, org string, rulesetID int64, rs *Ruleset) (*Ruleset, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/rulesets/%v", org, rulesetID)
+
+	req, err := s.client.NewRequest("PUT", u, rs)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var rulesets *Ruleset
```
gmlewis, line 84:
```suggestion
	var ruleset *Ruleset
```

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+// BypassActor represents the bypass actors from a ruleset.
+type BypassActor struct {
+	ActorID int64 `json:"actor_id,omitempty"`
```
gmlewis, line 16:
```suggestion
	ActorID *int64 `json:"actor_id,omitempty"`
```

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+	// Possible values for ActorType are: Team, Integration
+	ActorType string `json:"actor_type,omitempty"`
```
gmlewis, line 18:
```suggestion
	ActorType *string `json:"actor_type,omitempty"`
```

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+
+// RulesetLinks represents the "_links" object in a Ruleset.
+type RulesetLinks struct {
+	Self *RulesetLink `json:"self,omitempty"`
+}
+
+// RulesetRefConditionParameters represents the conditions object for ref_names.
+type RulesetRefConditionParameters struct {
+	Include []string `json:"include"`
+	Exclude []string `json:"exclude"`
```
gmlewis, line 34:
```suggestion
	Include []string `json:"include,omitempty"`
	Exclude []string `json:"exclude,omitempty"`
```

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+// RulesetRepositoryConditionParameters represents the conditions object for repository_names.
+type RulesetRepositoryConditionParameters struct {
+	Include   []string `json:"include"`
+	Exclude   []string `json:"exclude"`
```
gmlewis, line 40: It is challenging to tell if fields need to be `omitempty` from the docs, but it seems like `include` and `exclude` would be a good candidate for optionally including one or the other.

```suggestion
	Include   []string `json:"include,omitempty"`
	Exclude   []string `json:"exclude,omitempty"`
```

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+// RequiredStatusChecksRuleParameters represents the required_status_checks rule parameters.
+type RequiredStatusChecksRuleParameters struct {
+	RequiredStatusChecks             []RuleRequiredStatusChecks `json:"required_status_checks"`
+	StrictRequiredStatusChecksPolicy bool                       `json:"strict_required_status_checks_policy"`
+}
+
+// RepositoryRule represents a GitHub Rule.
+type RepositoryRule struct {
+	Type       string      `json:"type"`
+	Parameters interface{} `json:"parameters,omitempty"`
```
gmlewis, line 94: Instead of `interface{}`, I'm wondering if this would be better to have a `*json.RawMessage` here like we do in other places in this repo?

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+
+	r.Type = RepositoryRule.Type
+
+	switch RepositoryRule.Type {
+	case "creation", "deletion", "required_linear_history", "required_signatures", "non_fast_forward":
+		r.Parameters = nil
+	case "update":
+		RepositoryRule.Parameters = &UpdateAllowsFetchAndMergeRuleParameters{}
+		if err := json.Unmarshal(data, &RepositoryRule); err != nil {
+			return err
```
gmlewis, line 114: For these coverage warnings, it would be nice if you could add some tests that placed bogus JSON in the data such that these `json.Unmarshal` lines flagged by codecov would fail.  Something like "{]" usually does the trick.

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+// NewRequiredSignaturesRule creates a rule a to require commits pushed to matching branches to have verified signatures.
+func NewRequiredSignaturesRule() (rule RepositoryRule) {
+	return RepositoryRule{
+		Type: "required_signatures",
+	}
+}
+
+// NewPullRequestRule creates a rule to require all commits be made to a non-target branch and submitted via a pull request before they can be merged.
+func NewPullRequestRule(params *PullRequestRuleParameters) (
+	rule RepositoryRule) {
```
gmlewis, line 196: There's no need to break this line here... please add the return value here to the end of line 195.

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+	// Possible values for Target are branch, tag
+	Target *string `json:"target,omitempty"`
+	// Possible values for SourceType are: Repository, Organization
+	SourceType *string `json:"source_type,omitempty"`
+	Source     string  `json:"source"`
+	// Possible values for Enforcement are: disabled, active, evaluate
+	Enforcement string `json:"enforcement"`
+	// Possible values for BypassMode are: none, repository, organization
+	BypassMode   *string            `json:"bypass_mode,omitempty"`
+	BypassActors *[]BypassActor     `json:"bypass_actors,omitempty"`
```
gmlewis, line 271: Is `*[]BypassActor` necessary here?  Can't it be `[]*BypassActor` instead, which is preferred?

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+	NodeID       *string            `json:"node_id,omitempty"`
+	Links        *RulesetLinks      `json:"_links,omitempty"`
+	Conditions   *RulesetConditions `json:"conditions,omitempty"`
+	Rules        *[]RepositoryRule  `json:"rules,omitempty"`
```
gmlewis, line 275: Is `*[]RepositoryRule` necessary here?  Can't it be `[]*RepositoryRule` instead, which is preferred?

github/repos_rules.go
```diff
@@ -0,0 +1,392 @@
...
+	resp, err := s.client.Do(ctx, req, &rules)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return rules, resp, nil
+}
+
+// GetAllRulesets gets all the rules that apply to the specified repository.
+// If includesParents is true, rulesets configured at the organisation level that apply to the repository will be returned.
```
gmlewis, line 299:
```suggestion
// If includesParents is true, rulesets configured at the organization level that apply to the repository will be returned.
```

gmlewis (changes requested):
Thank you, @liaodaniel !
This looks fantastic!  I have just a few change requests, and feel free to disagree on any of them. As I commented, it is challenging to tell if we have the right combination of `omitempty`, but this looks good to me if I haven't made comments otherwise.

github/repos_rules.go
liaodaniel: Can I get some guidance on how this might be done? I took a look at some of the other examples and they seem to suggest to allow users to pass in `*json.RawMessage` types as the parameters? I am on the fence with this one as I feel like you lose out on all the typed constructs when you switch over to this it in this particular case?

```
# Example of creating a pull request rule
func NewPullRequestRule(params *json.RawMessage) (rule *RepositoryRule) {
	return &RepositoryRule{
		Type:       "pull_request",
		Parameters: params,
	}
}

pullRequestParams := json.RawMessage(`{
	"dismiss_stale_reviews_on_push": true,
	"require_code_owner_review": true,
	"require_last_push_approval": true,
	"required_approving_review_count": 1,
	"required_review_thread_resolution": true
}`)

rule := NewPullRequestRule(&pullRequestParams),
```

gmlewis: Sorry, I'm OOO today and will take a look tomorrow.

gmlewis: In general, you don't have to give up strong typing with `json.RawMessage`... you can always marshal other strongly-typed structs into the raw message.  It is just another way to handle variadic sub-messages within a JSON blob.
So I'm not sure you lose any functionality with `json.RawMessage` over the use of `any`... I think it is just a cleaner way of implementing the interface... but maybe I'm not understanding your question.

liaodaniel: Ok, thank you. Think I may have interpreted what you meant incorrectly previously.

Was this what you had in mind? See https://github.com/google/go-github/pull/2795/commits/5618f6f4a68cd1cea4f355cda1dca67367ac50d3

gmlewis (approved):
Thank you, @liaodaniel !
LGTM.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

Issue #2816, opened 2023-06-22
BUG: Regression in version 0.53 in `Dependabot.CreateOrUpdateOrgSecret`


frankywahl:
Hello,

I believe there was a regression in https://github.com/google/go-github/pull/2794. 

[Documentation](https://docs.github.com/en/rest/dependabot/secrets?apiVersion=2022-11-28#create-or-update-an-organization-secret) of organization secrets expects a string but we're now sending integers.

This in turn causes a failure when calling the API. 

I think the issue lies in that "setting" repository ids is done via integers on the API ([REF](https://docs.github.com/en/rest/dependabot/secrets?apiVersion=2022-11-28#set-selected-repositories-for-an-organization-secret)), while creating or updating a secrets expects repository ids to be a string ([REF](https://docs.github.com/en/rest/dependabot/secrets?apiVersion=2022-11-28#create-or-update-an-organization-secret)). So they cannot have the same underlying type in the client as long as the API expected different types

<details>
<summary> Sample Go Code </summary> 

```go
package main

import (
	"context"
	"encoding/base64"
	"fmt"
	"log"
	"os"

	"github.com/google/go-github/v53/github"
	"golang.org/x/crypto/nacl/box"
	"golang.org/x/oauth2"
)

func main() {
	ctx := context.Background()
	if err := run(ctx); err != nil {
		log.Fatal(err)
	}
}

func run(ctx context.Context) error {
	ts := oauth2.StaticTokenSource(
		&oauth2.Token{AccessToken: os.Getenv("GITHUB_API_TOKEN")},
	)
	tc := oauth2.NewClient(ctx, ts)

	client := github.NewClient(tc)
	org := os.Getenv("GITHUB_OWNER")

	pubKey, _, err := client.Dependabot.GetOrgPublicKey(ctx, org)
	if err != nil {
		return err
	}

	enc, err := encryptPlaintext("SECRET", *pubKey.Key)
	if err != nil {
		return err
	}

	_, err = client.Dependabot.CreateOrUpdateOrgSecret(ctx, org, &github.DependabotEncryptedSecret{
		Name:                  "EXAMPLE",
		KeyID:                 *pubKey.KeyID,
		EncryptedValue:        base64.StdEncoding.EncodeToString(enc),
		SelectedRepositoryIDs: []int64{123456789},
		Visibility:            "selected",
	})

	return err
}

func encryptPlaintext(plaintext, publicKeyB64 string) ([]byte, error) {
[...]

PR #2817, opened 2023-06-23
Fix `CreateOrUpdateOrgSecret` regression introduced in v53

frankywahl:
`CreateOrUpdateOrgSecret`'s API uses `string` for selected repository IDS while other APIs use `int`. This fixes the issue locally by implementing its own type.

fixes https://github.com/google/go-github/issues/2816

### DIFF
--- a/github/dependabot_secrets.go
+++ b/github/dependabot_secrets.go
@@ -146,7 +146,7 @@ func (s *DependabotService) CreateOrUpdateRepoSecret(ctx context.Context, owner,
 func (s *DependabotService) CreateOrUpdateOrgSecret(ctx context.Context, org string, eSecret *DependabotEncryptedSecret) (*Response, error) {
 	repoIDs := make([]string, len(eSecret.SelectedRepositoryIDs))
 	for i, secret := range eSecret.SelectedRepositoryIDs {
-		repoIDs[i] = fmt.Sprintf("%d", secret)
+		repoIDs[i] = fmt.Sprintf("%v", secret)
 	}
 	params := struct {
 		*DependabotEncryptedSecret
--- a/github/dependabot_secrets.go
+++ b/github/dependabot_secrets.go
@@ -144,8 +144,25 @@ func (s *DependabotService) CreateOrUpdateRepoSecret(ctx context.Context, owner,
 //
 // GitHub API docs: https://docs.github.com/en/rest/dependabot/secrets#create-or-update-an-organization-secret
 func (s *DependabotService) CreateOrUpdateOrgSecret(ctx context.Context, org string, eSecret *DependabotEncryptedSecret) (*Response, error) {
+	repoIDs := make([]string, len(eSecret.SelectedRepositoryIDs))
+	for i, secret := range eSecret.SelectedRepositoryIDs {
+		repoIDs[i] = fmt.Sprintf("%d", secret)
+	}
+	params := struct {
+		*DependabotEncryptedSecret
+		SelectedRepositoryIDs []string `json:"selected_repository_ids,omitempty"`
+	}{
+		DependabotEncryptedSecret: eSecret,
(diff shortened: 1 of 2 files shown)


github/dependabot_secrets.go
```diff
@@ -144,8 +145,29 @@ func (s *DependabotService) CreateOrUpdateRepoSecret(ctx context.Context, owner,
...
+	}
+
+	putSecret := func(ctx context.Context, url string, eSecret interface{}) (*Response, error) {
+		req, err := s.client.NewRequest("PUT", url, eSecret)
+		if err != nil {
+			return nil, err
+		}
+
+		return s.client.Do(ctx, req, nil)
+	}
```
frankywahl, line 167: Created this inline for now as it's the only place it's used. Happy to pull it out as an unexported function to match the `putSecret` package function (maybe `putOrgSecret`). That would require though defining a package type for `params` (for used as an input argument). I thought it was an overkill for now, so left it as such.

gmlewis: There's no need for the inline function here.  Let's just please delete lines 160 and 167, and move lines 161-166 down to replace line 170 below.

github/dependabot_secrets.go
```diff
@@ -144,8 +145,29 @@ func (s *DependabotService) CreateOrUpdateRepoSecret(ctx context.Context, owner,
...
+
 	url := fmt.Sprintf("orgs/%v/dependabot/secrets/%v", org, eSecret.Name)
-	return s.putSecret(ctx, url, eSecret)
+	return putSecret(ctx, url, params)
```
gmlewis, line 170:
```suggestion
	req, err := s.client.NewRequest("PUT", url, params)
	if err != nil {
		return nil, err
	}

	return s.client.Do(ctx, req, nil)
```

github/dependabot_secrets.go
```diff
@@ -144,8 +145,29 @@ func (s *DependabotService) CreateOrUpdateRepoSecret(ctx context.Context, owner,
 //
 // GitHub API docs: https://docs.github.com/en/rest/dependabot/secrets#create-or-update-an-organization-secret
 func (s *DependabotService) CreateOrUpdateOrgSecret(ctx context.Context, org string, eSecret *DependabotEncryptedSecret) (*Response, error) {
+	repoIDs := make([]string, len(eSecret.SelectedRepositoryIDs))
+	for i, secret := range eSecret.SelectedRepositoryIDs {
+		repoIDs[i] = strconv.Itoa(int(secret))
```
gmlewis, line 150:
```suggestion
		repoIDs[i] = fmt.Sprintf("%v", secret)
```

gmlewis (changes requested):
Thank you, @frankywahl!
Just a few tweaks, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/dependabot_secrets.go
frankywahl: Sounds good - was trying to keep the symmetry with what was previously there. Removed!

github/dependabot_secrets.go
gmlewis: No, let's please change it back to `%v` because it is idiomatic Go to use `%v` unless there is a specific reason otherwise, such as if we need leading zeros, as in `%05d`.
Although you may be able to find other examples in this repo where this is a `%d`, we have attempted to remain consistent and use `%v` unless there is a specific need.  Thanks.

gmlewis:
Also, in the future, please don't use force-push in PRs to this repo, as it makes reviewing more challenging to see what has changed since the last review.
See CONTRIBUTING.md for more details.  Thanks.

gmlewis (approved):
Thank you, @frankywahl !
LGTM.

valbeat (approved):

gmlewis:
Thank you, @valbeat !
Merging.

PR #2818, opened 2023-06-23
Add Runner to generate-jitconfig method

prksu:
Provide full response for generate-jitconfig method instead of `encoded_jit_config` only. According to the [doc](https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#create-configuration-for-a-just-in-time-runner-for-a-repository) there is a `runner` object in the API response.

why do we need it? since the generate-jitconfig API is just not generating jitconfig but also creating new self-hosted runner to the github so there is no way we can get the newly created runner id other than from this response.

cc @mkulke @gmlewis

### DIFF
--- a/github/actions_runners.go
+++ b/github/actions_runners.go
@@ -58,6 +58,7 @@ type GenerateJITConfigRequest struct {
 
 // JITRunnerConfig represents encoded JIT configuration that can be used to bootstrap a self-hosted runner.
 type JITRunnerConfig struct {
+	Runner           *Runner `json:"runner,omitempty"`
 	EncodedJITConfig *string `json:"encoded_jit_config,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


github/actions_runners.go
```diff
@@ -58,6 +58,7 @@ type GenerateJITConfigRequest struct {
 
 // JITRunnerConfig represents encoded JIT configuration that can be used to bootstrap a self-hosted runner.
 type JITRunnerConfig struct {
+	Runner           Runner  `json:"runner"`
```
gmlewis, line 61:
```suggestion
	Runner           *Runner  `json:"runner,omitempty"`
```

gmlewis:
Please run `go generate ./...` and push (not force push) the changes.
See CONTRIBUTING.md for details.

prksu:
Hi @gmlewis apologize, I thought it was not reviewed yet. I have run `go generate ./...` as you suggest and push the changes.

gmlewis (approved):
No problem, @prksu !
LGTM.
Merging after tests pass.

PR #2819, opened 2023-06-26
Remove unnecessary use of fmt.Sprintf

testwill:
unnecessary use of fmt.Sprint
unnecessary use of fmt.Sprintf


gmlewis (approved):
Thank you, @testwill !
LGTM.
Merging after tests pass.

PR #2823, opened 2023-06-28
Add Repository struct to SecretScanningAlert

emilva:
Add support for getting Repository when [listing secret scanning alerts for enterprise](https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#list-secret-scanning-alerts-for-an-enterprise)

There is no easy way of getting the org/repo when listing secret scanning alerts for enterprise in the current struct. The API returns the Repository field, and with this change it will be included in the SecretScanningAlert struct

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -19510,6 +19510,14 @@ func (s *SecretScanningAlert) GetNumber() int {
 	return *s.Number
 }
 
+// GetRepository returns the Repository field.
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @emilva !
LGTM.
Merging.

emilva:
Thank you for the quick merge 🥇

PR #2826, opened 2023-07-04
Add support for `personal_access_token_request` webhook event type

joaopenteado:
This PR adds support for the [`personal_access_token_request`](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#personal_access_token_request) webhook event.

> This event occurs when there is activity relating to a request for a fine-grained personal access token to access resources that belong to a resource owner that requires approval for token access.

PR #2827, opened 2023-07-05
Add support for personal access tokens request review API

joaopenteado:
Following up with my previous PR #2826, I am also proposing the following changes, which add support for [programmatically approving or denying](https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#review-a-request-to-access-organization-resources-with-a-fine-grained-personal-access-token) fine-grained personal access token requests.

I thought about including the rest of this API, but I don't have an immediate need for it and the PR might get a little too big. I was also a bit unsure on the function signature, so let me know your thoughts.

### DIFF
--- a/github/orgs_personal_access_tokens.go
+++ b/github/orgs_personal_access_tokens.go
@@ -22,7 +22,7 @@ type ReviewPersonalAccessTokenRequestOptions struct {
 // `action` can be one of `approve` or `deny`.
 //
 // GitHub API docs: https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#review-a-request-to-access-organization-resources-with-a-fine-grained-personal-access-token
-func (s *OrganizationsService) ReviewPersonalAccessTokenRequest(ctx context.Context, org, requestID string, opts ReviewPersonalAccessTokenRequestOptions) (*Response, error) {
+func (s *OrganizationsService) ReviewPersonalAccessTokenRequest(ctx context.Context, org string, requestID int64, opts ReviewPersonalAccessTokenRequestOptions) (*Response, error) {
 	u := fmt.Sprintf("orgs/%v/personal-access-token-requests/%v", org, requestID)
 
 	req, err := s.client.NewRequest(http.MethodPost, u, &opts)
--- a/github/orgs_personal_access_tokens.go
+++ b/github/orgs_personal_access_tokens.go
@@ -1,3 +1,8 @@
+// Copyright 2023 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package github
 
 import (
@@ -6,7 +11,7 @@ import (
 	"net/http"
 )
 
-// Approves or denies a pending request to access organization resources via a fine-grained personal access token.
+// ReviewPersonalAccessTokenRequest approves or denies a pending request to access organization resources via a fine-grained personal access token.
 // Only GitHub Apps can call this API, using the `organization_personal_access_token_requests: write` permission.
 // `action` can be one of `approve` or `deny`.
 //
--- a/github/orgs_personal_access_tokens.go
+++ b/github/orgs_personal_access_tokens.go
@@ -11,22 +11,21 @@ import (
 	"net/http"
 )
 
+// ReviewPersonalAccessTokenRequestOptions specifies the parameters to the ReviewPersonalAccessTokenRequest method.
+type ReviewPersonalAccessTokenRequestOptions struct {
+	Action string  `json:"action"`
+	Reason *string `json:"reason,omitempty"`
+}
+
 // ReviewPersonalAccessTokenRequest approves or denies a pending request to access organization resources via a fine-grained personal access token.
 // Only GitHub Apps can call this API, using the `organization_personal_access_token_requests: write` permission.
 // `action` can be one of `approve` or `deny`.
 //
 // GitHub API docs: https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#review-a-request-to-access-organization-resources-with-a-fine-grained-personal-access-token
-func (s *OrganizationsService) ReviewPersonalAccessTokenRequest(ctx context.Context, org, requestID, action, reason string) (*Response, error) {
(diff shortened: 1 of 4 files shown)


github/orgs_personal_access_tokens.go
```diff
@@ -0,0 +1,30 @@
+package github
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
//
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package github
```

github/orgs_personal_access_tokens_test.go
```diff
@@ -0,0 +1,57 @@
+package github
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
//
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package github
```

github/orgs_personal_access_tokens.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+// Approves or denies a pending request to access organization resources via a fine-grained personal access token.
```
gmlewis, line 9:
```suggestion
// ReviewPersonalAccessTokenRequest approves or denies a pending request to access organization resources via a fine-grained personal access token.
```

github/orgs_personal_access_tokens.go
```diff
@@ -0,0 +1,30 @@
...
+// Only GitHub Apps can call this API, using the `organization_personal_access_token_requests: write` permission.
+// `action` can be one of `approve` or `deny`.
+//
+// GitHub API docs: https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#review-a-request-to-access-organization-resources-with-a-fine-grained-personal-access-token
+func (s *OrganizationsService) ReviewPersonalAccessTokenRequest(ctx context.Context, org, requestID, action, reason string) (*Response, error) {
+	u := fmt.Sprintf("orgs/%v/personal-access-token-requests/%v", org, requestID)
+	body := struct {
```
gmlewis, line 16: In this repo, we typically make structs like this be public options that are passed into the method, and one of the reasons for this is that GitHub has a history of adding new options to API endpoints. If we didn't do this, then we would be continually breaking the API of this client library by being forced to update the signature of this method to keep adding new fields.

So let's please follow the same pattern for this new endpoint as well.

gmlewis (changes requested):
Thank you, @joaopenteado !

joaopenteado:
@gmlewis thanks again for the quick feedback! I'll probably commit the requested changes by tomorrow.

> one of the reasons for this is that GitHub has a history of adding new options to API endpoints. If we didn't do this, then we would be continually breaking the API of this client library by being forced to update the signature of this method to keep adding new fields.

This makes perfect sense. I should've thought of that, thanks for the heads up!

Just two questions though, as I am still not very familiar with this package's approach to API design. I saw a lot of other methods using pointers to the `Options` struct containing the body payload. This is good in the sense that it could potentially avoid a copy of a big struct, but it also allows for passing in a `nil` pointer which could then likely be dereferenced later on and cause a `panic` or a runtime error in case a payload is required. I take it this is an accepted risk in this package's design?

Additionally, the `Options` structs themselves will often (_but strangely not always_) be composed of only reference types. I imagine this is done intentionally in order to distinguish between Go's default initialized values and the presence or absence of values in the option itself. However, this also could introduce a runtime panic if a required parameter is forgotten, is this also by design then?

In order words, despite `Action` being mandatory for this specify endpoint, I should design my `Options` struct like this:
```go
type ReviewPersonalAccessTokenRequestOptions struct {
	Reason *string `json:"reason,omitempty"`
}
```
And not like this:
```go
type ReviewPersonalAccessTokenRequestOptions struct {
	Action string `json:"action"`
	Reason *string `json:"reason,omitempty"` // or string
}
```

Is my understanding of the above points, correct?

gmlewis:
> ... I saw a lot of other methods using pointers to the Options struct containing the body payload. ... I take it this is an accepted risk in this package's design?

So you will probably see inconsistencies in this regard within this client library, which I will take the blame for.
Ideally, if a method requires body parameters to be provided, then we would simply pass an `opts` parameter that is a struct passed by value (and not a struct pointer), and what I would recommend in this case.

> Additionally, the Options structs themselves will often (but strangely not always) be composed of only reference types. ...  is this also by design then?

Again, ideally, if a passed parameter is mandatory, we would not use a pointer, and we would not use `omitempty` in the JSON tag.  If a passed parameter is optional, then we use a pointer and include `omitempty`. There are also a few fiddly cases in the API where we must send a `null` or other weirdness where we take other measures, but I don't think this particular endpoint has any fiddly cases.

So it appears to me that your last suggestion would be preferred since `action` is mandatory and `reason` is not:

```go
type ReviewPersonalAccessTokenRequestOptions struct {
	Action string `json:"action"`
	Reason *string `json:"reason,omitempty"`
}
```

gmlewis:
In case the last comment wasn't clear, I would think this method would have this in it:
`opts ReviewPersonalAccessTokenRequestOptions) ...`

(in other words, _not_ a pointer to a struct, but instead using pass-by-value.)

joaopenteado:
@gmlewis Thank you for clarifying!
I've pushed some changes, let me know if you'd like me to change anything else.

gmlewis:
> Do you think I should keep it as a `string` just in case GitHub changes their mind about this ID later, or switch it to a `int64`?

Since the official docs say that it is an "integer", then let's please make it an `int64` in the hopes that that will reduce developer confusion. Thanks.

gmlewis:
Oh, and you can leave all the "%v" in the `fmt.Sprintf` as that is idiomatic Go and commonly used throughout this repo.  We only change this "%v" to other things when formatting must be non-standard (e.g. `"%05d"`).

joaopenteado:
Done!

gmlewis (approved):
Thank you, @joaopenteado !
LGTM.

liaodaniel (approved):
LGTM

gmlewis:
Thank you, @liaodaniel !
Merging.

PR #2831, opened 2023-07-15
Fix repo rules bypass settings

o-sama:
This PR reconciles library with GitHub API for the following regarding [repo rulesets](https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#create-a-repository-ruleset):
- `BypassMode` should be part of `BypassActor`.
- `ActorType` possible values documentation updated to match API docs.
- `BypassMode possible values documentation updated to match API docs.
- `Ruleset` ID field changed to `Int64` pointer and added `omitempty` to allow for nil, ruleset ID is unknown prior to creation and is not required when sending a POST request to create ruleset.

### DIFF
--- a/github/repos_rules.go
+++ b/github/repos_rules.go
@@ -14,8 +14,10 @@ import (
 // BypassActor represents the bypass actors from a ruleset.
 type BypassActor struct {
 	ActorID *int64 `json:"actor_id,omitempty"`
-	// Possible values for ActorType are: Team, Integration
+	// Possible values for ActorType are: RepositoryRole, Team, Integration, OrganizationAdmin
 	ActorType *string `json:"actor_type,omitempty"`
+	// Possible values for BypassMode are: always, pull_request
+	BypassMode *string `json:"bypass_mode,omitempty"`
 }
 
 // RulesetLink represents a single link object from GitHub ruleset request _links.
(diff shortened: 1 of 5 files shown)


github/repos_rules.go
```diff
@@ -14,8 +14,10 @@ import (
 // BypassActor represents the bypass actors from a ruleset.
 type BypassActor struct {
 	ActorID *int64 `json:"actor_id,omitempty"`
-	// Possible values for ActorType are: Team, Integration
+	// Possible values for ActorType are: RepositoryRole, Team, Integration, OrganizationAdmin
 	ActorType *string `json:"actor_type,omitempty"`
+	// Possible values for BypassMode are: always, pull_request
```
gmlewis, line 19: It seems odd to me that the original author noted that the possible values were completely different from these possible values. Are you sure that you found all the endpoints that the original author was covering?

o-sama: I believe so. Per my understanding of repo rules, you can create them on the org level and the repo level which should leave us with only the two endpoints. Both of these endpoints have the updated values. I may be missing something though so I'll take another look!

o-sama: I don’t see other references to the `Ruleset` struct in the codebase except for org rules which has the same possible values for the `BypassMode` field. I didn’t find anything else in the API docs either.

gmlewis: OK, I've gone through the docs as well, and I think you are correct.  Thank you, @o-sama !

gmlewis (approved):
Thank you @o-sama !
LGTM.

@liaodaniel - could you please look at these breaking API changes since you are the original author of this section of code and make sure you are happy with them?

liaodaniel:
> Thank you @o-sama ! LGTM.
> 
> @liaodaniel - could you please look at these breaking API changes since you are the original author of this section of code and make sure you are happy with them?

Yep, this looks good from my end. Part of me felt that I probably should have waited a bit until the feature was out of beta before introducing it but the feature seemed too compelling to wait 😄

liaodaniel (approved):
LGTM

gmlewis:
Thank you, @liaodaniel !
Merging.

PR #2834, opened 2023-07-20
Add secret type display to secret scanning alert

salamachinas:
added support for getting secret type display name when listing secret scanning alerts


gmlewis (approved):
Thank you, @salamachinas !
LGTM.
Merging.

PR #2837, opened 2023-07-21
Use a sentinel error when blocking paths for `RepositoriesServices.GetContents`

k0ral:
Related to https://github.com/google/go-github/pull/2805#issuecomment-1638119252 .

This allows users to catch this specific error using `errors.Is`.


gmlewis (approved):
Thank you, @k0ral !
LGTM.
Merging.

PR #2842, opened 2023-07-25
Add GetAutomatedSecurityFixes to report status

### LINKED ISSUE #2080: Feature: Check if automated security fixes are enabled for a repository
In https://github.com/google/go-github/pull/1188 the endpoint was added to enable or disable automated security fixes.

However, there seems to be no way to check the current status of this setting. This is required to implement https://github.com/integrations/terraform-provider-github/issues/272.

I also noticed on the [REST api documentation page](https://docs.github.com/en/rest/reference/repos) there doesn't seem to be an API endpoint to check this (there is one for [checking vulnerability alerts](https://docs.github.com/en/rest/reference/repos#check-if-vulnerability-alerts-are-enabled-for-a-repository)).

Is it possible to implement this feature? If not, where do I create an issue for making such an endpoint available?

grahamhar:
This is an attempt to fix #2080. It is also required by upstream projects such as [terraform-provider-github](https://github.com/integrations/terraform-provider-github), see [issue 1301](https://github.com/integrations/terraform-provider-github/issues/1301)

```
go test github.com/google/go-github/...
ok      github.com/google/go-github/v53/github  1.165s
?       github.com/google/go-github/v53/test/fields     [no test files]
?       github.com/google/go-github/v53/test/integration        [no test files]

```

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -687,6 +687,28 @@ func (s *RepositoriesService) DisableVulnerabilityAlerts(ctx context.Context, ow
 	return s.client.Do(ctx, req, nil)
 }
 
+// GetAutomatedSecurityFixes checks if the automated security fixes for a repository are enabled.
+//
+// GitHub API docs: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#check-if-automated-security-fixes-are-enabled-for-a-repository
+func (s *RepositoriesService) GetAutomatedSecurityFixes(ctx context.Context, owner, repository string) (*AutomatedSecurityFixes, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/automated-security-fixes", owner, repository)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
+	req.Header.Set("Accept", mediaTypeRequiredAutomatedSecurityFixesPreview)
+
+	p := new(AutomatedSecurityFixes)
+	resp, err := s.client.Do(ctx, req, p)
+	if err != nil {
+		return nil, resp, err
+	}
+	return p, resp, nil
+}
+
 // EnableAutomatedSecurityFixes enables the automated security fixes for a repository.
 //
 // GitHub API docs: https://docs.github.com/en/rest/repos/repos#enable-automated-security-fixes
@@ -1210,6 +1232,12 @@ type SignaturesProtectedBranch struct {
 	URL *string `json:"url,omitempty"`
 	// Commits pushed to matching branches must have verified signatures.
 	Enabled *bool `json:"enabled,omitempty"`
+}
+
+// AutomatedSecurityFixes represents their status.
+type AutomatedSecurityFixes struct {
+	Enabled *bool `json:"enabled"`
+	Paused  *bool `json:"paused"`
 }
 
 // ListBranches lists branches for the specified repository.
--- a/github/repos.go
+++ b/github/repos.go
@@ -698,9 +698,6 @@ func (s *RepositoriesService) GetAutomatedSecurityFixes(ctx context.Context, own
 		return nil, nil, err
 	}
 
-	// TODO: remove custom Accept header when this API fully launches
-	req.Header.Set("Accept", mediaTypeRequiredAutomatedSecurityFixesPreview)
-
 	p := new(AutomatedSecurityFixes)
 	resp, err := s.client.Do(ctx, req, p)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


github/repos.go
```diff
@@ -1212,6 +1234,12 @@ type SignaturesProtectedBranch struct {
 	Enabled *bool `json:"enabled,omitempty"`
 }
 
+// AutomatedSecurityFixes represents their status.
+type AutomatedSecurityFixes struct {
+	Enabled *bool `json:"enabled"`
+	Paused  *bool `json:"paused"`
```
gmlewis, line 1240:
```suggestion
	Enabled *bool `json:"enabled,omitempty"`
	Paused  *bool `json:"paused,omitempty"`
```

github/repos_test.go
```diff
@@ -664,6 +664,31 @@ func TestRepositoriesService_EnableAutomatedSecurityFixes(t *testing.T) {
...
+	}
+
+	want := &AutomatedSecurityFixes{
+		Enabled: Bool(true),
+		Paused:  Bool(false),
+	}
+	if !cmp.Equal(fixes, want) {
+		t.Errorf("Repositories.GetAutomatedSecurityFixes returned #{fixes}, want #{want}")
+	}
+}
```
gmlewis, line 690:
```suggestion

	const methodName = "GetAutomatedSecurityFixes"
	testBadOptions(t, methodName, func() (err error) {
		_, _, err = client.Repositories.GetAutomatedSecurityFixes(ctx, "\n", "\n")
		return err
	})
	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
		got, resp, err := client.Repositories.GetAutomatedSecurityFixes(ctx, "o", "r")
		if got != nil {
			t.Errorf("testNewRequestAndDoFailure %v = %#v, want nil", methodName, got)
		}
		return resp, err
	})
}
```

github/repos.go
grahamhar: @gmlewis could you explain this to me so I understand it's purpose in any future PR I do?

gmlewis: Please let me know if https://github.com/google/go-github/pull/2827#issuecomment-1621990607 answers your question.

grahamhar: Thanks for the link @gmlewis it makes it very clear. However, having read that I think my original form was possibly correct. Looking at the response schema in the API documentation it states that both fields are required in the response, so if I understood the link correctly these should not have omitempty set?

gmlewis (approved):
Thank you, @grahamhar !
LGTM.

github/repos.go
gmlewis: Ah, then I apologize.  If both are required, then you are right - not only does `omitempty` not need to be added, but these fields also do not need to be pointers.  They can simply be `bool`.

grahamhar: let me update the PR

gmlewis: Historically, we have erred on the side of using pointers since sometimes GitHub responses have included things in some API calls and omitted them in others.  But if these will always be present, then there is no need to make them pointers.  

However, I'm fine if you just want to drop the `omitempty` and revert to your original version.

grahamhar: Pushed the changes to remove the omitempty.

gmlewis (approved):

github/repos.go
```diff
@@ -687,6 +687,28 @@ func (s *RepositoriesService) DisableVulnerabilityAlerts(ctx context.Context, ow
...
+// GitHub API docs: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#check-if-automated-security-fixes-are-enabled-for-a-repository
+func (s *RepositoriesService) GetAutomatedSecurityFixes(ctx context.Context, owner, repository string) (*AutomatedSecurityFixes, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/automated-security-fixes", owner, repository)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// TODO: remove custom Accept header when this API fully launches
```
vandanrohatgi, line 701: Looks like API docs now recommend using the common `application/vnd.github+json` header. Maybe we can remove this custom header?

grahamhar: Happy to do that as part of this PR.

grahamhar: I've made the changes I think are needed please let me know if I need to do anything else.

vandanrohatgi (approved):
LGTM!

github/repos.go
vandanrohatgi: Thank you @grahamhar !

gmlewis:
Thank you, @vandanrohatgi !
Merging.

PR #2843, opened 2023-07-26
Add `old_name` field to `AuditEntry`

lmartinking:
This field is used when a repository is renamed.

Fixes: #2844.

PR #2849, opened 2023-07-28
Fix `old_name` field in `AuditEntry`

lmartinking:
Much to my chagrin, the audit log view in the site manager does not follow the shape of the API response.

I previously raised a PR #2843 which was merged yesterday. Unfortunately, it does not work with the payload returned from the API.

My sincere apologies about this.

### DIFF
--- a/github/orgs_audit_log.go
+++ b/github/orgs_audit_log.go
@@ -125,7 +125,7 @@ type AuditEntry struct {
 	AuditEntryData `json:"data,omitempty"`
 }
 
-// Some audit entries have additional information stuffed into a `data` field.
+// AuditEntryData represents additional information stuffed into a `data` field.
 type AuditEntryData struct {
 	OldName *string `json:"old_name,omitempty"` // The previous name of the repository, for a name change
 }
--- a/github/orgs_audit_log.go
+++ b/github/orgs_audit_log.go
@@ -78,7 +78,6 @@ type AuditEntry struct {
 	Message                *string                 `json:"message,omitempty"`
 	Name                   *string                 `json:"name,omitempty"`
 	OAuthApplicationID     *int64                  `json:"oauth_application_id,omitempty"`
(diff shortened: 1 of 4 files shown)


github/orgs_audit_log.go
```diff
@@ -122,6 +121,13 @@ type AuditEntry struct {
 	Visibility             *string                 `json:"visibility,omitempty"` // The repository visibility, for example `public` or `private`.
 	WorkflowID             *int64                  `json:"workflow_id,omitempty"`
 	WorkflowRunID          *int64                  `json:"workflow_run_id,omitempty"`
+
+	AuditEntryData `json:"data,omitempty"`
```
gmlewis, line 125:
```suggestion
	Data  *AuditEntryData `json:"data,omitempty"`
```

github/orgs_audit_log.go
```diff
@@ -122,6 +121,13 @@ type AuditEntry struct {
...
+}
+
+// Some audit entries have additional information stuffed into a `data` field.
```
gmlewis, line 128:
```suggestion
// AuditEntryData represents additional information stuffed into a `data` field.
```

gmlewis (changes requested):
Thank you, @lmartinking !
Now worries.  I'm glad you caught it before I cut the new release.  😄

lmartinking:
@gmlewis I have one more field to add, which I will do in a other PR

lmartinking:
@gmlewis do you want the structs to mirror the shape of the API or would you prefer it flattened? Which was why I had it as a composite member.

gmlewis:
> @gmlewis do you want the structs to mirror the shape of the API or would you prefer it flattened? Which was why I had it as a composite member.

We typically do _not_ use composite members in this repo.

gmlewis (approved):
Thank you, @lmartinking !
LGTM.
Merging.

PR #2850, opened 2023-07-28
Add `OldLogin` field to `AuditEntryData`

lmartinking:
When renaming a Github organisation, the audit log entry (with action `org.rename`) will contain `data.old_login` which provides the prior name of the organisation.

This PR builds upon prior work which added support for the `data` field within an Audit Log entry.


gmlewis (approved):
Thank you, @lmartinking !
LGTM.
Merging.

PR #2854, opened 2023-08-02
Check for nil pointer in update rule parameters

o-sama:
Since discussing this with the GitHub folks, I’ve been told that the `update_allows_fetch_and_merge` parameter for the update rule only applies to forked repos. **Note: I was told this is still bugged for forked repos when sending a request through the API, but will be fixed in the near future.**

Since it only applies to forked repos, the API returns a ruleset which contains:
```
{
   "type": “update"
}
```

This doesn't have a parameters field, so when we try to unmarshal the JSON we run into a nil pointer dereference error. No functionality is changed here except to check for a nil pointer before we try to unmarshal the JSON, and if it isn’t included then we just set the parameters to nil.

I didn’t think this needed a change in testing because it doesn’t change any functionality, but I’ll be happy to go and add something if you think otherwise.

@gmlewis I’d really appreciate if this can be part of the upcoming release (which I know is short notice) since I’d love for it to be included in https://github.com/integrations/terraform-provider-github/pull/1808. No worries if it’s too late though, I can just keep it out for now. Thanks!!

### DIFF
--- a/github/repos_rules.go
+++ b/github/repos_rules.go
@@ -118,7 +118,7 @@ func (r *RepositoryRule) UnmarshalJSON(data []byte) error {
 	case "creation", "deletion", "required_linear_history", "required_signatures", "non_fast_forward":
 		r.Parameters = nil
 	case "update":
-		if RepositoryRule.Parameters != nil {
+		if RepositoryRule.Parameters == nil {
 			r.Parameters = nil
 			return nil
 		} else {
@@ -190,13 +190,18 @@ func NewCreationRule() (rule *RepositoryRule) {
 
 // NewUpdateRule creates a rule to only allow users with bypass permission to update matching refs.
 func NewUpdateRule(params *UpdateAllowsFetchAndMergeRuleParameters) (rule *RepositoryRule) {
-	bytes, _ := json.Marshal(params)
+	if params != nil {
+		bytes, _ := json.Marshal(params)
 
-	rawParams := json.RawMessage(bytes)
+		rawParams := json.RawMessage(bytes)
 
+		return &RepositoryRule{
+			Type:       "update",
+			Parameters: &rawParams,
+		}
+	}
 	return &RepositoryRule{
-		Type:       "update",
-		Parameters: &rawParams,
+		Type: "update",
 	}
 }
 
--- a/github/repos_rules.go
+++ b/github/repos_rules.go
@@ -121,17 +121,17 @@ func (r *RepositoryRule) UnmarshalJSON(data []byte) error {
 		if RepositoryRule.Parameters == nil {
 			r.Parameters = nil
 			return nil
-		} else {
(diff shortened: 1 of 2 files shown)


github/repos_rules.go
```diff
@@ -119,14 +119,18 @@ func (r *RepositoryRule) UnmarshalJSON(data []byte) error {
 		r.Parameters = nil
 	case "update":
 		params := UpdateAllowsFetchAndMergeRuleParameters{}
-		if err := json.Unmarshal(*RepositoryRule.Parameters, &params); err != nil {
-			return err
-		}
+		if RepositoryRule.Parameters != nil {
```
gmlewis, line 122: Let's please change this to make it easier to read:

```go
if RepositoryRule.Parameters == nil {
  r.Parameters = nil
  return nil
}
```

gmlewis (changes requested):
Thank you, @o-sama!
Please add a new unit test to cover this case.
Yes, we can hold off on the next release until this PR is merged.

github/repos_rules.go
gmlewis: Right, I understand, but I want to highlight the "happy path" and provide the early return which is idiomatic Go.
Please switch to my suggestion, and feel free to put a blank line after the new block if you think it reads better.

o-sama: Done, also added a bit in the update rule constructor to check for empty params since this specific rule is a bit different than others 🙂

github/repos_rules.go
```diff
@@ -118,15 +118,20 @@ func (r *RepositoryRule) UnmarshalJSON(data []byte) error {
 	case "creation", "deletion", "required_linear_history", "required_signatures", "non_fast_forward":
 		r.Parameters = nil
 	case "update":
-		params := UpdateAllowsFetchAndMergeRuleParameters{}
-		if err := json.Unmarshal(*RepositoryRule.Parameters, &params); err != nil {
-			return err
+		if RepositoryRule.Parameters == nil {
+			r.Parameters = nil
+			return nil
+		} else {
```
gmlewis, line 124: One of the points was to remove the "else" and un-indent the following section, please:

```suggestion
		}
```

o-sama: You’re totally right, I think I had a brain freeze. Don’t know why I kept an else with a return in there.

gmlewis (approved):
Thank you, @o-sama !
LGTM.

o-sama:
Thanks @gmlewis! sorry about the sloppiness earlier with the requested change 🙏

liaodaniel (approved):
LGTM

gmlewis:
> Thanks @gmlewis! sorry about the sloppiness earlier with the requested change 🙏

No problem! Sorry for the delayed responses... Work sometimes gets in the way. 😂

gmlewis:
Thank you, @liaodaniel !
Merging.

PR #2858, opened 2023-08-07
Fix dropped test error

alrs:
This picks up a dropped `err` variable in tests and clarifies the error message.


gmlewis (approved):
Thank you, @alrs !
LGTM.
Merging.

PR #2865, opened 2023-08-10
Add WebhookTypes and EventForType methods

### LINKED ISSUE #2863: Support enumerating the known event types
It looks like `go-github` has [an internal list of all the GitHub event types ](https://github.com/google/go-github/blob/899235e0a9d750d6fecf9048a676046d50f9d4a3/github/messages.go#L46).  I'd like to be able to iterate over the list in my own code where I route events based on event type.  I'm thinking of exposing the following methods:

```golang
// WebhookTypes returns a list of all the known GitHub event type strings supported by go-github.
func WebhookTypes() []string {
  // ...
}

// EventForType returns an empty struct matching the specified GitHub event type.
func EventForType(messageType string) any {

}
```

The second method would extract a lot of `Event.ParsePayload` https://github.com/google/go-github/blob/master/github/event.go#L30, so it wouldn't be much additional code.

evankanderson:
Fixes #2863


github/messages.go
```diff
@@ -307,3 +315,144 @@ func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
...
+		types = append(types, t)
+	}
+	return types
+}
+
+// EventForType returns an empty struct matching the specified GitHub event type.
+// If messageType does not match any known event types, it returns nil.
+func EventForType(messageType string) interface{} {
+	switch messageType {
+	case "branch_protection_rule":
```
gmlewis, line 333: This thing is such a pain to remember for new PR contributors.

What if we were to update `eventTypeMapping` to include both fields we want to look up?

For example:

```go
type eventTypeTuple struct {
  eventName string
  emptyStruct interface{}
}

var (
  eventTypeMapping = map[string]eventTypeTuple{
    "branch_protection_rule": {"BranchProtectionRuleEvent", &BranchProtectionRuleEvent{}},
    "check_run":              {"CheckRunEvent", &CheckRunEvent{}},
    ...
  }
)
```

Then we could change this `EventForType` to also use the same map.
Finally, every time a new event is added, only one line needs updating in the `eventTypeMapping`.

Thoughts?

gmlewis (changes requested):

gmlewis (changes requested):
Could you please also add tests for requesting a `nil` event, an unknown event, and any other place where we pass in garbage to make sure it behaves properly in all cases?

github/messages.go
evankanderson: I thought about this -- we need to make a _copy_ of the object, which means using reflection given that the declared type is `interface{}`.  If the performance cost seems acceptable, that sounds great to me.

github/event.go
```diff
@@ -27,132 +27,11 @@ func (e Event) String() string {
...
-	case "WorkflowJobEvent":
-		payload = &WorkflowJobEvent{}
-	case "WorkflowRunEvent":
-		payload = &WorkflowRunEvent{}
-	}
-	err = json.Unmarshal(*e.RawPayload, &payload)
+func (e *Event) ParsePayload() (interface{}, error) {
+	// It would be nice if e.Type were the snake_case name of the event,
+	// but the existing interface uses the struct name instead.
+	payload := EventForType(typeToMessageMapping[*e.Type])
```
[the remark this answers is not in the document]
evankanderson: Sure.  I got thrown off because the other methods were in `github-accessors.go`, so I missed that they existed.

github/event.go
```diff
@@ -27,132 +27,11 @@ func (e Event) String() string {
...
-		payload = &WorkflowJobEvent{}
-	case "WorkflowRunEvent":
-		payload = &WorkflowRunEvent{}
-	}
-	err = json.Unmarshal(*e.RawPayload, &payload)
+func (e *Event) ParsePayload() (interface{}, error) {
+	// It would be nice if e.Type were the snake_case name of the event,
+	// but the existing interface uses the struct name instead.
+	payload := EventForType(typeToMessageMapping[e.GetType()])
+	err := json.Unmarshal(*e.RawPayload, &payload)
```
gmlewis, line 34: Can this use an accessory?

evankanderson: Yeah... it turns out this is also subtly wrong if payload is `nil`, I think.  I'll have a fix shortly, along with tests for `ParsePayload` on an `Event{}` (zero-valued struct).

github/messages.go
```diff
@@ -43,71 +44,83 @@ const (
...
-		"status":                         "StatusEvent",
-		"team":                           "TeamEvent",
-		"team_add":                       "TeamAddEvent",
-		"user":                           "UserEvent",
-		"watch":                          "WatchEvent",
-		"workflow_dispatch":              "WorkflowDispatchEvent",
-		"workflow_job":                   "WorkflowJobEvent",
-		"workflow_run":                   "WorkflowRunEvent",
+	eventTypeMapping = map[string]interface{}{
+		"branch_protection_rule":         &BranchProtectionRuleEvent{},
```
gmlewis, line 48: Hey - here's a thought to avoid reusing the same pointer over and over again... what if these lines all returned a struct by _VALUE_ instead of by _reference_?

```suggestion
		"branch_protection_rule":         BranchProtectionRuleEvent{},
```

And then each call to the function would return a brand new pointer.

Thoughts?

evankanderson: Let me make sure we don't end up aliasing the returned values.

evankanderson: Ah, I could do that, but would it be any better than the current `reflect.New` usage?

github/event_test.go
```diff
@@ -30,6 +30,11 @@ func TestPayload_NoPanic(t *testing.T) {
 	e.Payload()
 }
 
+func TestEmptyEvent_NoPanic(t *testing.T) {
+	e := &Event{}
+	e.ParsePayload()
+}
```
gmlewis, line 36: Also, please:

```suggestion
  var nilEvent *Event
  nilEvent.ParsePayload()
}
```

github/messages.go
```diff
@@ -307,3 +320,25 @@ func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
...
+// EventForType returns an empty struct matching the specified GitHub event type.
+// If messageType does not match any known event types, it returns nil.
+func EventForType(messageType string) interface{} {
+	prototype := eventTypeMapping[messageType]
+	if prototype != nil {
+		// return a _copy_ of the pointed-to-object.  Unfortunately, for this
+		// we need to use reflection.
+		return reflect.New(reflect.TypeOf(prototype).Elem()).Interface()
+	}
+	return prototype
```
gmlewis, line 343: Let's please return the happy path at the end, which is idiomatic Go style:

```suggestion
	if prototype == nil {
		return nil
	}
	// return a _copy_ of the pointed-to-object.  Unfortunately, for this
	// we need to use reflection.
	return reflect.New(reflect.TypeOf(prototype).Elem()).Interface()
```

github/messages.go
gmlewis: Oh, I gotcha.  No, what you have looks good, thanks!

github/event_test.go
evankanderson: Done.

evankanderson: (I also checked that both of these returned error, which seems correct.)

github/messages.go
evankanderson: Done.  I also referenced this issue so people could find [the comment above](https://github.com/google/go-github/pull/2865#discussion_r1290433599).

github/messages.go
evankanderson: Oh, one more thought over the weekend: except for "GitHubAppAuthorizationEvent", all of these follow a regular snake_case --> TitleCase conversion.  I _could_ build this map with a type list and reflection, plus a special case for "GitHub" --> "github", but that seemed _too_ clever, so I didn't.

gmlewis: Yes, we prefer simplicity over cleverness, so thank you for keeping it simpler... it is greatly appreciated.  😄

github/messages.go
```diff
@@ -308,3 +321,27 @@ func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
...
+// If messageType does not match any known event types, it returns nil.
+func EventForType(messageType string) interface{} {
+	prototype := eventTypeMapping[messageType]
+	if prototype == nil {
+		return nil
+	}
+	// return a _copy_ of the pointed-to-object.  Unfortunately, for this we
+	// need to use reflection.  If we store the actual objects in the map,
+	// we still need to use reflection to convert from `any` to the actual
+	// type, so this was deemed the lesser of two weevils. (#2865)
```
gmlewis, line 345: Although I got a chuckle out of `weevils`, I'm thinking that we might want to be a bit more friendly to our friends where English is their Nth language (for N > 1), so let's please revert this to "evils".  Thank you. 😄

gmlewis (changes requested):
Thank you, @evankanderson !
A few more tweaks, please, then I think we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/messages.go
```diff
@@ -308,3 +321,27 @@ func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
 	}
 	return event.ParsePayload()
 }
+
+// WebhookTypes returns a list of all the known GitHub event type strings
```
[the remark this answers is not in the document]
evankanderson: I really wish that go would include https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set, but in the meantime, I can sort the list.

github/messages.go
evankanderson: Yeah, I had to try.  That's what code review is for!

github/event.go
```diff
@@ -28,134 +28,15 @@ func (e Event) String() string {
...
+	// but the existing interface uses the struct name instead.
+	typed := EventForType(typeToMessageMapping[e.GetType()])
+	if typed != nil {
+		payload = typed
 	}
-	err = json.Unmarshal(*e.RawPayload, &payload)
-	return payload, err
+	err = json.Unmarshal(e.GetRawPayload(), &payload)
+
+	return
```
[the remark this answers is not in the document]
evankanderson: Ah -- this subtly changes the semantics to always return `nil` (rather than an empty struct) if the `json.Unmarshal` fails.  I'm okay with that.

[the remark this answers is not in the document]
evankanderson: And if I'm doing that, there's no point in naming the return parameters, so I won't.

evankanderson:
Ready for another look / another reviewer, I think.

github/event.go
```diff
@@ -27,139 +27,16 @@ func (e Event) String() string {
...
-	case "WorkflowDispatchEvent":
-		payload = &WorkflowDispatchEvent{}
-	case "WorkflowJobEvent":
-		payload = &WorkflowJobEvent{}
-	case "WorkflowRunEvent":
-		payload = &WorkflowRunEvent{}
+func (e *Event) ParsePayload() (interface{}, error) {
+	// It would be nice if e.Type were the snake_case name of the event,
+	// but the existing interface uses the struct name instead.
+	var payload interface{} = EventForType(typeToMessageMapping[e.GetType()])
```
gmlewis, line 33:
```suggestion
	payload := EventForType(typeToMessageMapping[e.GetType()])
```

evankanderson: Done.  Earlier, I'd had some trouble with the compiler getting the types correct, but I think that was with the map-of-structs rather than map-of-pointers-to-structs solution.

evankanderson:
Do you want me to flatten my commits?

gmlewis:
> Do you want me to flatten my commits?

No need to flatten any commits, as we always "squash and merge" in this repo.  Thank you, though.

gmlewis (approved):
@evankanderson - this looks great to me!
Thank you for your patience and your willingness to experiment with this one.
I think this looks super clean and will help with reducing not only the maintenance burden of this repo, but also make it _much_ easier for contributors to add new event types as GitHub keeps cranking them out.  :grin:

LGTM.

gmlewis:
@googlebot - why have you been silent on this PR? What happened to your automated CLA checks?!?

gmlewis:
Ah, is @googlebot dead and been replaced by @google-cla instead?

gmlewis:
@google-ospo-team - can you please investigate why this PR was never checked by the @google-cla bot?
I can't move forward on this PR until it verifies the CLA is signed by all parties.
Thank you!

gmlewis:
CLA has been cleared.  Thank you, @google-ospo-team !

Now awaiting second LGTM+Approval from any other contributor to this repo before merging.

gmlewis:
@Parker77 - do you have time for a code review, please?

github/messages.go
```diff
@@ -43,74 +45,86 @@ const (
...
+		"status":                         &StatusEvent{},
+		"team":                           &TeamEvent{},
+		"team_add":                       &TeamAddEvent{},
+		"user":                           &UserEvent{},
+		"watch":                          &WatchEvent{},
+		"workflow_dispatch":              &WorkflowDispatchEvent{},
+		"workflow_job":                   &WorkflowJobEvent{},
+		"workflow_run":                   &WorkflowRunEvent{},
 	}
+	// forward mapping of event types to the sting names of the structs
```
Parker77, line 114:
```suggestion
	// forward mapping of event types to the string names of the structs
```

Parker77 (approved):
LGTM!

gmlewis:
Thank you, @Parker77 !
Merging.

PR #2870, opened 2023-08-14
Defer closing body before checking HTTP status code

nikpivkin:
This is a quote from the [documentation](https://pkg.go.dev/net/http#example-Get) of the `net/http package`:
> When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it.

In the current implementation, if the error code is 404, the response body will not close.


gmlewis (approved):
Thank you, @nikpivkin !
LGTM.

vandanrohatgi (approved):
Makes sense. LGTM!

gmlewis:
Thank you, @vandanrohatgi !  ❤️ 
Merging.

PR #2871, opened 2023-08-14
Fix SecretScanning API by switching arguments from url to json

jentfoo:
These structs were incorrectly defined with `url` parameters, instead these need to be encoded into `json` for the request.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -20046,6 +20046,30 @@ func (s *SecretScanningAlertEvent) GetSender() *User {
 	return s.Sender
 }
 
+// GetResolution returns the Resolution field if it's non-nil, zero value otherwise.
+func (s *SecretScanningAlertListOptions) GetResolution() string {
+	if s == nil || s.Resolution == nil {
+		return ""
+	}
+	return *s.Resolution
+}
+
+// GetSecretType returns the SecretType field if it's non-nil, zero value otherwise.
+func (s *SecretScanningAlertListOptions) GetSecretType() string {
+	if s == nil || s.SecretType == nil {
+		return ""
+	}
+	return *s.SecretType
+}
+
+// GetState returns the State field if it's non-nil, zero value otherwise.
+func (s *SecretScanningAlertListOptions) GetState() string {
+	if s == nil || s.State == nil {
+		return ""
+	}
+	return *s.State
+}
+
 // GetDetails returns the Details field.
 func (s *SecretScanningAlertLocation) GetDetails() *SecretScanningAlertLocationDetails {
 	if s == nil {
--- a/github/secret_scanning.go
+++ b/github/secret_scanning.go
@@ -53,14 +53,14 @@ type SecretScanningAlertLocationDetails struct {
 // SecretScanningAlertListOptions specifies optional parameters to the SecretScanningService.ListAlertsForEnterprise method.
 type SecretScanningAlertListOptions struct {
 	// State of the secret scanning alerts to list. Set to open or resolved to only list secret scanning alerts in a specific state.
-	State string `json:"state,omitempty"`
+	State *string `json:"state,omitempty"`
 
 	// A comma-separated list of secret types to return. By default all secret types are returned.
-	SecretType string `json:"secret_type,omitempty"`
+	SecretType *string `json:"secret_type,omitempty"`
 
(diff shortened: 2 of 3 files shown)


github/secret_scanning.go
```diff
@@ -53,14 +53,14 @@ type SecretScanningAlertLocationDetails struct {
 // SecretScanningAlertListOptions specifies optional parameters to the SecretScanningService.ListAlertsForEnterprise method.
 type SecretScanningAlertListOptions struct {
 	// State of the secret scanning alerts to list. Set to open or resolved to only list secret scanning alerts in a specific state.
-	State string `url:"state,omitempty"`
+	State string `json:"state,omitempty"`
```
gmlewis, line 56:
```suggestion
	State *string `json:"state,omitempty"`
```

github/secret_scanning.go
```diff
@@ -53,14 +53,14 @@ type SecretScanningAlertLocationDetails struct {
...
 
 	// A comma-separated list of secret types to return. By default all secret types are returned.
-	SecretType string `url:"secret_type,omitempty"`
+	SecretType string `json:"secret_type,omitempty"`
```
gmlewis, line 59:
```suggestion
	SecretType *string `json:"secret_type,omitempty"`
```

gmlewis (changes requested):
Thank you, @jentfoo !
Now that they are using `json` tags, we need to change the optional strings to pointers.

gmlewis:
Please run `go generate ./...` at the top of the repo and push (not force-push) the changes to this PR.
See CONTRIBUTING.md for details.

gmlewis:
It looks like you will also need to update the unit tests by changing some of the strings to `String(...)` so that their pointers are taken.

Please run "go test ./..." locally to make sure all tests pass before pushing your next commit.  Thanks.

gmlewis:
@jentfoo - we'll leave this PR open for a week or two, but if we haven't heard back from you, we'll close it as abandoned.

jentfoo:
@gmlewis This is a bug that should be addressed, however I don't have time to complete the requested test changes

gmlewis:
OK, thank you for the update, @jentfoo - I'll take a look when I get a chance.

gmlewis:
Hmmm... I'm looking at:
https://docs.github.com/en/enterprise-server@3.5/rest/secret-scanning/secret-scanning#list-secret-scanning-alerts-for-an-enterprise
and it shows that the parameters in question are indeed URL query parameters and not sent via JSON (which this PR is changing them to).

So let's back up a moment.

When you say "This is a bug that should be addressed"... could you please elaborate on the actual bug you are seeing?

What endpoint(s) are you calling?
What error messages do you get?
Do you have a simple test case that demonstrates the problem?

Unless I can better understand what the actual problem is, I'm going to have to close this PR as not having enough information to fix.

jentfoo:
> What endpoint(s) are you calling?

This PR was opened a while ago and I ultimately ended up switching to GraphQL for most of my needs.  So I don't fully recall why I was suggesting updating `SecretScanningAlertListOptions`.  It looks like based off your linked docs that recommendation may have been in error.

However `SecretScanningAlertUpdateOptions` is appears to me incorrect, here are the relevant docs: https://docs.github.com/en/enterprise-server@3.5/rest/secret-scanning/secret-scanning#update-a-secret-scanning-alert

I recall an API error when the current `github.Client.SecretScanning.UpdateAlert` is attempted with any configuration in the `SecretScanningAlertUpdateOptions` struct.

I apologize for not being able to help further right now, but I will try to answer questions as I am able.

gmlewis:
Ah, yes!  I see the problem now!  Thank you, @jentfoo !
I'll see what I can do here.

github/github-accessors.go
```diff
@@ -20046,6 +20046,30 @@ func (s *SecretScanningAlertEvent) GetSender() *User {
...
+}
+
+// GetState returns the State field if it's non-nil, zero value otherwise.
+func (s *SecretScanningAlertListOptions) GetState() string {
+	if s == nil || s.State == nil {
+		return ""
+	}
+	return *s.State
+}
+
```
gmlewis, line 20072:
```suggestion
```

gmlewis:
Note that this PR is being replaced by #2934 since I was unable to edit the contents of this PR myself.

PR #2892, opened 2023-08-21
Return json.Unmarshal error when importing issues deferred

nikpivkin:
The error was not returned when the 202 status was received during the issues import.


gmlewis (changes requested):
Thank you, @nikpivkin !
Could you please add a test case that demonstrates the problem?

In the past, I've found it really handy to demonstrate `json.Unmarshal` errors by forcing the data to be something bogus for JSON like `{]`.

gmlewis:
Thank you, @gabriel-samfira !
Merging.

Issue #2893, opened 2023-08-22
Remove dependency on golang.org/x/oauth2 


WillAbides:
golang.org/x/oauth2 is a pretty big module with its own dependencies including "google.golang.org/appengine" and "cloud.google.com/go/compute/metadata", but go-github only uses a very small part of it to add a static value to the Authorization header.

The oauth2 dependency can be replaced with a simple RoundTripper that adds the Authorization header.

gmlewis:
Would you like to propose a PR, @WillAbides ?

WillAbides:
@gmlewis Thanks for the reminder. I meant to create the PR at the same time but got called to dinner.

PR #2895, opened 2023-08-22
Remove dependency on "golang.org/x/oauth2"

WillAbides:
Resolves google/go-github#2893

This removes the primary module's dependency on "golang.org/x/oauth2" and replaces it with a RoundTripper that adds an "Authorization" header with the value "Bearer <token>".

TestNewTokenClient now does an http request and checks for the "Authorization" header.

I moved `roundTripperFunc` from test code to prod because it was useful for creating the RoundTripper.


gmlewis:
Do you want to update all the files that mention `oauth2`, including documentation, while you're here?

```
$ rg -l oauth2
README.md
go.sum
scrape/go.sum
example/newreposecretwithlibsodium/go.sum
example/newreposecretwithlibsodium/go.mod
go.mod
github/github.go
github/doc.go
example/go.sum
example/go.mod
example/tokenauth/main.go
example/newfilewithappauth/main.go
github/github_test.go
```

WillAbides:
These are the remaining oauth2 mentions:

```
$ rg -l oauth2
README.md
github/github.go
github/doc.go
example/go.sum
example/go.mod
example/newfilewithappauth/main.go
scrape/go.sum
```

The first three are docs and code comments about how to authenticate. I think it's appropriate to keep oauth2 there because go-github doesn't offer an Enterprise equivalent of `NewTokenClient`.

`example/*` is similar to the doc issue except this time it's actually using oauth2 as an example. Fortunately example is its own module so we don't need to worry about extra imports from there.

`scrape/go.sum` has an oauth2 line because `scrape/go.mod` requires "github.com/google/go-github/v54 v54.0.0" which still requires oauth2. That should go away the first time `go mod tidy` is run in `scrape/` after this change is released.

gmlewis (approved):
Normally the maintainers perform the "go mod tidy", and we have been using "go mod tidy -compat=1.17" for a long while now.

But all the tests seem to be passing, so maybe we are still fine.  This will be an experiment. :smile:

Thank you, @WillAbides !
LGTM.

gmlewis:
> `scrape/go.sum` has an oauth2 line because `scrape/go.mod` requires "github.com/google/go-github/v54 v54.0.0" which still requires oauth2. That should go away the first time `go mod tidy` is run in `scrape/` after this change is released.

Thanks for the reminder!  I should do that anyway.

gmlewis:
> Thanks for the reminder! I should do that anyway.

Oh, wait - didn't I do that already in #2882 ?

gabriel-samfira (approved):

gmlewis:
Thank you, @gabriel-samfira !
Merging.

Issue #1203, opened 2019-06-21
addOptions fails to test URL if nil options are passed


gmlewis:
While increasing code coverage in #1202, I discovered that if `nil` is passed to `addOptions`, then the rest of the URL tests are not performed.

This means that a garbage URL could be used without detection if `nil` is passed as the option.
This doesn't seem correct, but this will need more investigation to see why that check was added there and if it was fixing some other problem.

AGMETEOR:
Hello @gmlewis , was this task done? Is it also a `good first issue`?

gmlewis:
This issue is still open and has not been done yet.

No, unfortunately I would not consider this to be a good first issue because it will take some investigation and testing to figure out what the correct solution is.

paladium:
Hello, I am happy to work on this issue. As I understand the effect of moving the parsing of the URL before checking for the nil value of options might have unpredictable effects?

gmlewis:
I apologize for not documenting this issue better.

Here is the problem...  In [addOptions](https://github.com/google/go-github/blob/master/github/github.go#L233-L253), the code says this right now:

```go
// addOptions adds the parameters in opt as URL query parameters to s. opt
// must be a struct whose fields may contain "url" tags.
func addOptions(s string, opts interface{}) (string, error) {
	v := reflect.ValueOf(opts)
	if v.Kind() == reflect.Ptr && v.IsNil() {
		return s, nil
	}

	u, err := url.Parse(s)
	if err != nil {
		return s, err
	}
...
}
```

What I realized is that if `opts` is `nil`, then this function returns on [line238](https://github.com/google/go-github/blob/master/github/github.go#L238) without even attempting to parse the URL.

So if the URL is can not be parsed, addOptions is still going to report success, and I'm thinking that the [lines 241-244](https://github.com/google/go-github/blob/master/github/github.go#L241-L244) should be moved up to the very top of the function.

In retrospect, this seems like a simple solution... _however_ I filed this issue because this change will require a good amount of testing, and I would like someone to test out this change "in real life" as well, with the most endpoints as possible against real GitHub. I'm not expecting the change to cause problems, but we just need someone to carefully check.

Does that make sense, @paladium ?

Thank you for volunteering!

paladium:
Yes, I understand the issue better now. I will be happy to test out this change. Is there any formal way I can document the process of real-life testing?

paladium:
I can make a separate repo, include the changed library and write e2e tests for each API endpoint by consuming the new library?

[...]

PR #2896, opened 2023-08-22
Vaidate URLs

WillAbides:
Resolves google/go-github#1203

This addresses the issue raised in google/go-github#1203 by adding a `newURLString` function that builds a url the same way they were previously built with `fmt.Sprintf` but also errors if the result is not valid.

I went this route instead of modifying `addOptions` because it seemed better to validate all urls instead of only the ones that use `addOptions`.

This turned into a pretty big diff, but that is the nature of updating boilerplate that is used in hundreds of methods.

### DIFF
--- a/github/activity_notifications.go
+++ b/github/activity_notifications.go
@@ -54,10 +54,6 @@ func (s *ActivityService) ListNotifications(ctx context.Context, opts *Notificat
 	if err != nil {
 		return nil, nil, err
 	}
-	u, err = addOptions("notifications", opts)
-	if err != nil {
-		return nil, nil, err
-	}
 	u, err = addOptions(u, opts)
 	if err != nil {
 		return nil, nil, err
--- a/github/activity_notifications.go
+++ b/github/activity_notifications.go
@@ -7,7 +7,6 @@ package github
 
 import (
 	"context"
-	"fmt"
 	"time"
 )
 
@@ -51,8 +50,15 @@ type NotificationListOptions struct {
 //
 // GitHub API docs: https://docs.github.com/en/rest/activity/notifications#list-notifications-for-the-authenticated-user
 func (s *ActivityService) ListNotifications(ctx context.Context, opts *NotificationListOptions) ([]*Notification, *Response, error) {
-	u := "notifications"
-	u, err := addOptions(u, opts)
+	u, err := newURLString("notifications")
+	if err != nil {
+		return nil, nil, err
+	}
+	u, err = addOptions("notifications", opts)
+	if err != nil {
+		return nil, nil, err
+	}
+	u, err = addOptions(u, opts)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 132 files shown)


github/activity_notifications.go
```diff
@@ -51,8 +50,11 @@ type NotificationListOptions struct {
 //
 // GitHub API docs: https://docs.github.com/en/rest/activity/notifications#list-notifications-for-the-authenticated-user
 func (s *ActivityService) ListNotifications(ctx context.Context, opts *NotificationListOptions) ([]*Notification, *Response, error) {
-	u := "notifications"
-	u, err := addOptions(u, opts)
+	u, err := newURLString("notifications")
+	if err != nil {
+		return nil, nil, err
+	}
```
gmlewis, line 56: I don't see any benefit in this case. I think we can safely revert this change.

github/activity_star.go
```diff
@@ -71,12 +73,16 @@ type ActivityListStarredOptions struct {
 // GitHub API docs: https://docs.github.com/en/rest/activity/starring#list-repositories-starred-by-a-user
 func (s *ActivityService) ListStarred(ctx context.Context, user string, opts *ActivityListStarredOptions) ([]*StarredRepository, *Response, error) {
 	var u string
+	var err error
 	if user != "" {
-		u = fmt.Sprintf("users/%v/starred", user)
+		u, err = newURLString("users/%v/starred", user)
 	} else {
-		u = "user/starred"
+		u, err = newURLString("user/starred")
```
gmlewis, line 80: This line can stay as it was before. Any time `newURLString` has only a single constant arg, that change can be reverted.

gmlewis (changes requested):
Thank you, @WillAbides .
I'm going to stop review now because I'm curious if you are able to see all the Codecov warnings that I'm able to see?

If not, I can highlight them for you, as there are a lot, as also evidenced by the significant drop in code coverage stats.  It also found some errors, so I'll let you clean those up before I continue reviewing.

Thanks!

WillAbides:
@gmlewis Thanks for having a look. I knew this would likely have a negative impact on code coverage, but decided to wait on the PR for codecov to show me the diff before worrying about it. As an aside now that I've looked at the coverage, 98.1% coverage on master is impressive.

For the urls that are just static strings, I decided to replace those as well so that we are using a consistent pattern across all methods. I don't feel strongly about that and am happy to revert those.

I'm having doubts about the general usefulness of this PR. It seems like if we catch this error in either `newURLString` or `addOptions` we are returning the same error that would have been returned from either [client.NewRequest()](https://github.com/google/go-github/blob/38ca69f37f8597421ceb02e08f2dd225cfaeb1a9/github/github.go#L435-L438) or [http.NewRequest()](https://github.com/golang/go/blob/e68c027204d410ebca5bf1a9660605f2bd737748/src/net/http/request.go#L879-L882). Maybe that wasn't the case when #1203 was filed or maybe there is an exception to this that I'm not seeing.

gmlewis:
> I'm having doubts about the general usefulness of this PR. It seems like if we catch this error in either newURLString or addOptions we are returning the same error that would have been returned from either [client.NewRequest()](https://github.com/google/go-github/blob/38ca69f37f8597421ceb02e08f2dd225cfaeb1a9/github/github.go#L435-L438) or [http.NewRequest()](https://github.com/golang/go/blob/e68c027204d410ebca5bf1a9660605f2bd737748/src/net/http/request.go#L879-L882). Maybe that wasn't the case when https://github.com/google/go-github/issues/1203 was filed or maybe there is an exception to this that I'm not seeing.

Honestly, I am totally fine if you want to call this an interesting experiment and determine that the results are not worth it.

I remember filing #1203 after discovering that an error wasn't triggered when I messed something up which sent alarms through my mind, and was hoping that someone else would give it a good investigation and come to a conclusion, which you graciously have... and thank you!

If you wish to point to this PR in that issue and close them both as "unnecessary", I'm totally fine with that and once again really appreciate your detailed investigation!

WillAbides:
Closing this as unnecessary. I also commented on #1203, but I can't actually close it myself.

PR #2903, opened 2023-08-26
Fix linting

WillAbides:
I noticed that `.golangci.yml` had some commented out and disabled linters. This PR fixes that. It also uncomments "example" from the GitHub action.

Running `go mod tidy -compat 1.17` fixed the "no go files to analyze" error linting "examples".

For the linting issues, I updated `.golangci-lint` and fixed the issues that it found.

I think most of the changes are obvious, but I want to mention a few things:
- I created two test helpers "assertNilError" and "assertWrite" because that made the error checks issues easier to fix with find/replace. I see that assert functions are generally not used in this repo. I'm happy to inline these, but my preference is to keep them as-is.
- I changed the signature of `stringifyValue` to accept a `*bytes.Buffer` instead of an `io.Writer` because the linter knows that `Buffer.Write` always returns nil errors.
- I removed quite a bit of dead code from `authorizations_test.go`


WillAbides:
Codecov didn't like that I modified an uncovered line, so I added a test for it.

gmlewis:
I'm doing a crazy amount of driving today, so will be slow to respond. I try to "approve run" of the checks when I can, but today's going to be pretty challenging, sorry.

WillAbides:
@gmlewis I appreciate how responsive you are on this repo. I have no expectation that you or anybody else be around on a weekend. I can see the lint results on my fork, so my feedback loop is just fine.

gmlewis (approved):
Wow, @WillAbides - thank you for this huge cleanup!  It is greatly appreciated!
LGTM.

gabriel-samfira (approved):
This is awesome

gmlewis:
Thank you, @gabriel-samfira !
Merging.

PR #2904, opened 2023-08-28
Update package constructors

### LINKED ISSUE #2897: Proposal: client.WithOptions
As I was working on #2895 I realized that go-github has four different funcs for creating a new `github.Client` depending on what options you want to set.

- NewClient sets the `http.Client`
- NewClientWithEnvProxy builds an `http.Client` configured with `http.ProxyFromEnvironment`
- NewTokenClient builds an `http.Client` configured to use bearer token auth.
- NewEnterpriseClient is like unto NewClient but lets you set urls that point to your GHE server.

These are all fine on their own, but you can't combine them. If you want to configure a client for your enterprise server that uses token authentication, you need to choose whether to use NewTokenClient then configure BaseURL and UploadURL or configure an `http.Client` for token auth and pass it to NewEnterpriseClient.

## Proposal

Create a new method on `*Client`: `WithOptions(opts ...ClientOption) (*Client, error)` along with options that cover the existing constructors' functionality.

The options would be:
- `WithAuthToken(token string)`
- `WithEnvProxy()`
- `WithEnterpriseURLs(baseURL, uploadURL string)`[^enterprise-note]

[^enterprise-note]: I chose this over separate WithBaseURL and WithUploadURL options because this makes it more clear that they will be modified the same way NewEnterpriseClient currently modifies urls.

Creating a new enterprise client with an auth token would look something like:

```go
client, err := github.NewClient(nil).WithOptions(
	github.WithEnterpriseURLs(myURL, myURL),
	github.WithAuthToken(myToken), 
)
```

### Update NewClient to accept options

This would involve changing NewClient's signature to:

```go
func NewClient(httpClient *http.Client, opts ...ClientOption) *Client
```

This would be a backward-compatible signature change because the only new argument is variadic. This would also probably be the best option if we could limit it to just this change. However, not all options are guaranteed to be valid, so we would need to either add an error to the signature or panic when an invalid option is passed. Neither of those are good compromises, so I moved on from this. 

### Add a "New" func

Client is the primary export of the github package, so a constructor named "New" would be a good fit semantically. If "New" doesn't work for a name, then "NewClientWithOptions" could work too.

Being a new function would also allow us to avoid the required `http.Client` argument and move that to an option called WithHTTPClient.

The Signature would look like: `New(opts ...ClientOption) (*Client, error)`

Creating a new enterprise client with an auth token would look something like:

```go
client, err := github.New(
    github.WithEnterpriseURLs(myURL, myURL),
    github.WithAuthToken(myToken), 
)
```

This seems like a compelling option to me. I prefer the actual proposal over this because this issue started with an observation that we have too many constructors. Adding a new constructor in response reminded me of [xkcd 927](https://xkcd.com/927/).

### Separate With* methods for each option

Instead of one WithOptions method, we could add WithEnterpriseURLs, WithTokenAuth, etc.

I think a single WithOptions method lends itself to better usage. This might be a better option if none of the methods returned errors so proper method chaining could be used, but that isn't the situation here.

WillAbides:
Closes #2897

After experimenting with implementing #2897, I think this ends up with better usage for end users.

This deprecates `github.NewTokenClient` and `github.NewEnterpriseClient` and replaces them with `Client.WithAuthToken` and `Client.WithEnterpriseURLs`.

Now you can create an authenticated client for an enterprise server with:

```go
client, err := github.NewClient(nil).WithAuthToken(token).WithEnterpriseURLs(baseURL,uploadURL)
```

There is no need for users import `github.com/x/oauth2` anymore.

I opted for this implementation over my `WithOptions` proposal the error returned from `WithOptions` means that most uses of go-github would have an unnecessary error check.

```go
client err := github.NewClient(nil).WithOptions(github.WithAuthToken(token))
if err != nil {
  panic(err)
}
```

vs

```go
client := github.NewClient(nil).WithAuthToken(token)
```


github/github.go
```diff
@@ -358,62 +453,15 @@ func NewClientWithEnvProxy() *Client {
...
-// If a nil httpClient is provided, a new http.Client will be used.
-//
-// Note that NewEnterpriseClient is a convenience helper only;
-// its behavior is equivalent to using NewClient, followed by setting
-// the BaseURL and UploadURL fields.
-//
-// Another important thing is that by default, the GitHub Enterprise URL format
-// should be http(s)://[hostname]/api/v3/ or you will always receive the 406 status code.
-// The upload URL format should be http(s)://[hostname]/api/uploads/.
+// Deprecated: Use NewClient(httpClient).WithOptions(WithEnterpriseURLs(baseURL, uploadURL)) instead.
```
gmlewis, line 462: We need the first two lines to make the auto-generated godoc documentation read correctly.

```suggestion
// NewEnterpriseClient returns a new GitHub API client with provided
// base URL and upload URL (often is your GitHub Enterprise hostname).
//
// Deprecated: Use NewClient(httpClient).WithOptions(WithEnterpriseURLs(baseURL, uploadURL)) instead.
```

github/github.go
```diff
@@ -358,62 +453,15 @@ func NewClientWithEnvProxy() *Client {
...
-		!strings.HasPrefix(uploadEndpoint.Host, "api.") &&
-		!strings.Contains(uploadEndpoint.Host, ".api.") {
-		uploadEndpoint.Path += "api/uploads/"
-	}
-
-	c := NewClient(httpClient)
-	c.BaseURL = baseEndpoint
-	c.UploadURL = uploadEndpoint
-	return c, nil
+	return NewClient(httpClient).WithEnterpriseURLs(baseURL, uploadURL)
```
gmlewis, line 464: This seems like a really easy thing to add a test to make codecov happy.

github/github.go
```diff
@@ -349,7 +424,27 @@ func NewClient(httpClient *http.Client) *Client {
...
+	clone := Client{
+		client:                  c.client,
+		UserAgent:               c.UserAgent,
+		BaseURL:                 c.BaseURL,
+		UploadURL:               c.UploadURL,
+		secondaryRateLimitReset: c.secondaryRateLimitReset,
+	}
+	c.clientMu.Unlock()
+	if clone.client == nil {
+		clone.client = &http.Client{}
```
gmlewis, line 442: This also seems like a really easy thing to add a test to make codecov happy.

gmlewis (changes requested):
This looks great, @WillAbides !
Just a few minor nits, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

WillAbides:
@gmlewis Updated with your suggestions.

As an aside, is there a way to get a preview of what codecov is going to report before pushing? I looked for a way to run it locally, and it appears to require an API token and membership in the right org.

gmlewis:
> As an aside, is there a way to get a preview of what codecov is going to report before pushing? I looked for a way to run it locally, and it appears to require an API token and membership in the right org.

Not that I'm aware of.

Sometimes, to get an idea of what it might report, I run this locally first:

```bash
$ cat coverage-web.sh
#!/bin/bash -ex
go test -coverprofile=cover.out ./...
go tool cover -html cover.out
```

but it is obviously not the same.

gmlewis (approved):
Awesome!  Thank you, @WillAbides !
LGTM.

WillAbides:
I made a script to preview whether codecov will fail your PR: https://gist.github.com/WillAbides/e0bbffccb34d9922cecb2ab9d66defec

gmlewis:
> I made a script to preview whether codecov will fail your PR: https://gist.github.com/WillAbides/e0bbffccb34d9922cecb2ab9d66defec

Thank you, @WillAbides !  That's fantastic!

gmlewis:
Thank you, @sridharavinash !
Merging.

Issue #2908, opened 2023-08-31
Proposal: metadata file for endpoints


WillAbides:
Contributing to and maintaining this repo could be helped by a metadata file for all of the endpoints go-github supports. It would map to [GitHub's OpenAPI descriptions](https://github.com/github/rest-api-description). In this proposal I'll call it `operations.yaml`.

An `operations.yaml` entry might look like:

```yaml
- operation_id: pulls/request-reviewers
  openapi:
    method: POST
    endpoint_url: '/repos/{owner}/{repo}/pulls/{pull_number}/requested_reviewers'
    documentation_url: 'https://docs.github.com/rest/pulls/review-requests#request-reviewers-for-a-pull-request'
    summary: Request reviewers for a pull request
  go_method: PullRequestsService.RequestReviewers
```

`operations.yaml` could be updated by a scheduled GitHub action that opens a PR when it finds a change.

Initially `operations.yaml` would be useful for `update-urls`. Instead of having to resolve every documentation url, it could use the value stored in `operations.yaml`. `update-urls` could be expanded to also update godoc with the summary field.

It would also be useful for finding new endpoints for go-github to support. Eventually we could write a generator that uses `operations.yaml` to generate the boilerplate to implement new endpoints.

## Overrides

There are probably endpoints where we don't want to use the OpenAPI data or maybe some that aren't even in the OpenAPI description. For those cases we can create an `override` section:

```yaml
- operation_id: pulls/request-reviewers
  openapi:
    method: POST
    endpoint_url: '/repos/{owner}/{repo}/pulls/{pull_number}/requested_reviewers'
    documentation_url: 'https://docs.github.com/rest/pulls/review-requests#request-reviewers-for-a-pull-request'
    summary: Request reviewers for a pull request
  override:
    documentation_url: https://some/other/url
  go_method: PullRequestsService.RequestReviewers
```

For an endpoint that doesn't exist in the OpenAPI description, we create an entry with no `openapi` section and all required fields set in `override`. Here's a fictional PUT endpoint for the same url as above.

```yaml
- operation_id: pulls/update-request-reviewers # manually created
  override:
    method: PUT
    endpoint_url: '/repos/{owner}/{repo}/pulls/{pull_number}/requested_reviewers'
    documentation_url: 'https://docs.github.com/rest/pulls/review-requestss#update-reviewers-for-a-pull-request'
    summary: Update reviewers for a pull request
  go_method: PullRequestsService.UpdateRequestReviewers
```

[...]

PR #2919, opened 2023-09-09
Use metadata to reconcile go-github with GitHub's OpenAPI descriptions

WillAbides:
see #2908

TODO before submitting for review:

- [x] Fix line-end issue on Windows
- [x] Update CONTRIBUTING.md
- [x] Add copyright notices to new files

This PR introduces a file called openapi_operations.yaml that contains details of GitHub's endpoints as described in https://github.com/github/rest-api-description. When combined with the `//meta:operation` directive, this allows us to automatically keep documentation links up to date, but more importantly it allows us to track endpoints as they are added and removed.

The change set is huge, but it is primarily made up of openapi_operations.yaml itself and changes to godoc.

"tools/cmd/metadata" is a tool to manage metadata.yaml. I'll paste its help output here for convenience:

```
Usage: metadata <command>

Flags:
  -h, --help               Show context-sensitive help.
  -C, --working-dir="."    Working directory. Should be the root of the go-github repository.

Commands:
  update-openapi
    Update openapi_operations.yaml from OpenAPI descriptions in github.com/github/rest-api-description at the given git ref.

  update-go
    Update go source code to be consistent with openapi_operations.yaml.
      - Adds and updates "// GitHub API docs:" comments for service methods.
      - Updates "//meta:operation" comments to use canonical operation names.
      - Updates formatting of "//meta:operation" comments to make sure there isn't a space between the "//" and the "meta".
      - Formats modified files with the equivalent of "go fmt".

  format
    Format whitespace in openapi_operations.yaml and sort its operations.

  unused
    List operations in openapi_operations.yaml that aren't used by any service methods.

Run "metadata <command> --help" for more information on a command.
```

I'm trying to limit the scope of this PR because its line count is already pretty high. I will make a follow-up PR with a daily scheduled workflow that runs "update-openapi" and "update-go" and creates a PR if there are changes.


gmlewis (commented):
I'll be OOO for one week, but before I go, I just wanted to point out that some of the new URLs have double slashes next to each other, e.g. `https://docs.github.com/enterprise-cloud@latest//rest/actions/cache#get-github-actions-cache-usage-for-an-enterprise`.

Also, in cases where URLs have been completely removed, it might be nice to point
to explanations as to _why_ they are being removed, such as:
* https://docs.github.com/en/actions/using-workflows/required-workflows
* https://docs.github.com/en/webhooks/replacing-github-services
* etc.

Have a great week, and thanks for doing this, @WillAbides !

WillAbides:
@gmlewis enjoy your time away.

Those double slashed URLs are the same way in the OpenAPI description. They work as-is, but I wouldn't mind normalizing the path for aesthetic reasons.

That's a good point about adding explanations to the ones that lost their doc links. I'll go through those and add some notes.

WillAbides:
I updated update-urls to only modify links that don't resolve to the same docs. This makes the change set for this PR more manageable.

gmlewis (approved):
Wow, @WillAbides - very nice.
I'm looking forward to using it.
It looks like the PR is still a "Draft" but looks pretty complete to me. What's missing?
Maybe more instructions for maintainers? It looks pretty automated, though.

WillAbides:
@gmlewis You're right about instructions. After I started to work on CONTRIBUTING.md I realized that this adds complexity for new contributors by asking them to edit a large yaml file. I also suspect that most new contributors don't do much more than skim CONTRIBUTING.md before making their first PR. I know I'm often guilty of that for new projects. A typical new contribution would be:

1. code the change
2. create the PR
3. wait for CI to be approved
4. find out you need to update metadata.yaml

I gave it a little more thought and came up with a change I think will help. People tend to copy and adapt existing code. We can move the operation mapping to method a directive comment so that new contributors will see it on other methods as they write their code.

The code comment for `RepositoriesService.Get` would become:

```
// Get fetches a repository.
//
// GitHub API docs: https://docs.github.com/en/rest/repos/repos#get-a-repository
//
//meta:operation GET /repos/{owner}/{repo}
```

`//meta:operation ...` is a directive comment so godoc ignores it and outputs the same as before.

I started coding this up thinking it would add complexity, but it actually simplifies the tooling code quite a bit.

What do you think of moving the mapping to a directive comment?

WillAbides:
I'm glad you like that idea. I have it mostly working, but it might be a few days before I have time to finish it up.

tools/metadata/openapi.go
```diff
@@ -0,0 +1,163 @@
...
+	files := make([]*openapiFile, 0, len(dir))
+	for _, d := range dir {
+		for i, pattern := range dirPatterns {
+			m := pattern.FindStringSubmatch(d.GetName())
+			if m == nil {
+				continue
+			}
+			plan := m[pattern.SubexpIndex("plan")]
+			major, _ := strconv.Atoi(m[pattern.SubexpIndex("major")])
+			minor, _ := strconv.Atoi(m[pattern.SubexpIndex("minor")])
```
gmlewis, line 133: Do we really not care here if the major and/or minor versions are not supplied?
If so, could you please add a comment before line 132 stating that we don't care?

tools/metadata/metadata_test.go
```diff
@@ -0,0 +1,23 @@
+package main
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
//
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package main
```

tools/metadata/main.go
```diff
@@ -0,0 +1,192 @@
+// Copyright 2023 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
```
gmlewis, line 6:
```suggestion
// metadata is a command-line tool used to check and update this repo.
// See CONTRIBUTING.md for details.
package main
```

gmlewis (changes requested):
This looks great, @WillAbides !
Just a few minor tweaks, please, then we'll be ready for a second LGTM+Approval before merging.

tools/metadata/openapi.go
WillAbides: I updated it to return errors

gmlewis (approved):
Thank you, @WillAbides !
LGTM.

vandanrohatgi (approved):
I'm not sure if my approval helps here (won't be offended if we decide to wait for another approval), but I have gone through some the changes and LGTM. 
Hoping this moves things along a bit.

valbeat (approved):

gmlewis:
Thank you, @vandanrohatgi and @valbeat !
Merging.

PR #2920, opened 2023-09-11
Add enterprise actions permissions endpoints and reorg files

RickleAndMortimer:
PR following #2518


RickleAndMortimer:
I'm not sure what approach I should take to resolve the duplicated code. It seems like both enterprise and organization versions of the permissions just swap the url of the endpoint. 

`
func (s *ActionsService) ListEnabledReposInOrg(ctx context.Context, owner string, opts *ListOptions) (*ActionsEnabledOnOrgRepos, *Response, error) {
	u := fmt.Sprintf("orgs/%v/actions/permissions/repositories", owner)
	u, err := addOptions(u, opts)
	if err != nil {
		return nil, nil, err
	}

	if err != nil {
		return nil, nil, err
	}

	repos := &ActionsEnabledOnOrgRepos{}
	resp, err := s.client.Do(ctx, req, repos)
	if err != nil {
		return nil, resp, err
	}

	return repos, resp, nil
}
`

`
func (s *ActionsService) ListEnabledOrgsInEnterprise(ctx context.Context, owner string, opts *ListOptions) (*ActionsEnabledOnEnterpriseOrgs, *Response, error) {
	u := fmt.Sprintf("enterprises/%v/actions/permissions/organizations", owner)
	u, err := addOptions(u, opts)
	if err != nil {
		return nil, nil, err
	}

	if err != nil {
		return nil, nil, err
	}

	orgs := &ActionsEnabledOnEnterpriseOrgs{}
	resp, err := s.client.Do(ctx, req, orgs)
	if err != nil {
		return nil, resp, err
	}

	return orgs, resp, nil
}
`

github/enterprise_actions_allowed_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func TestEnterpriseService_EditActionsAllowed(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+	input := &ActionsAllowed{GithubOwnedAllowed: Bool(true), VerifiedAllowed: Bool(false), PatternsAllowed: []string{"a/b"}}
+
+	mux.HandleFunc("/enterprises/e/actions/permissions/selected-actions", func(w http.ResponseWriter, r *http.Request) {
+		v := new(ActionsAllowed)
+		json.NewDecoder(r.Body).Decode(v)
```
WillAbides, line 59:
```suggestion
		assertNilError(t, json.NewDecoder(r.Body).Decode(v))
```

github/enterprise_actions_permissions_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+func TestEnterpriseService_EditActionsPermissions(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	input := &ActionsPermissionsEnterprise{EnabledOrganizations: String("all"), AllowedActions: String("selected")}
+
+	mux.HandleFunc("/enterprises/e/actions/permissions", func(w http.ResponseWriter, r *http.Request) {
+		v := new(ActionsPermissionsEnterprise)
+		json.NewDecoder(r.Body).Decode(v)
```
WillAbides, line 60:
```suggestion
		assertNilError(t, json.NewDecoder(r.Body).Decode(v))
```

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (a ActionsPermissionsEnterprise) String() string {
+	return Stringify(a)
+}
+
+// ActionsEnabledOnEnterpriseOrgs represents all the repositories in an enterprise for which Actions is enabled.
+type ActionsEnabledOnEnterpriseOrgs struct {
+	TotalCount    int             `json:"total_count"`
+	Organizations []*Organization `json:"organizations"`
+}
```
WillAbides, line 30:
```suggestion
// ActionsEnabledOnEnterpriseOrgs represents all the repositories in an enterprise for which Actions is enabled.
type ActionsEnabledOnEnterpriseOrgs struct {
	TotalCount    int             `json:"total_count"`
	Organizations []*Organization `json:"organizations"`
}

// ActionsPermissionsEnterprise represents a policy for allowed actions in an enterprise.
//
// GitHub API docs: https://docs.github.com/en/rest/actions/permissions
type ActionsPermissionsEnterprise struct {
	EnabledOrganizations *string `json:"enabled_organizations,omitempty"`
	AllowedActions       *string `json:"allowed_actions,omitempty"`
	SelectedActionsURL   *string `json:"selected_actions_url,omitempty"`
}

func (a ActionsPermissionsEnterprise) String() string {
	return Stringify(a)
}
```

This is kind of hacky, but it clears the duplicate code issue.

WillAbides (commented):
Just some suggestions to make it pass lint. I'll do a full review later today.

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// AddEnabledOrgInEnterprise adds an organization to the list of selected organizations that are enabled for GitHub Actions in an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#enable-a-selected-organization-for-github-actions-in-an-enterprise
```
WillAbides, line 118:
```suggestion
// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#enable-a-selected-organization-for-github-actions-in-an-enterprise
```

github/enterprise_actions_permissions.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// RemoveEnabledOrgInEnterprise removes an organization from the list of selected organizations that are enabled for GitHub Actions in an enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#disable-a-selected-organization-for-github-actions-in-an-enterprise
```
WillAbides, line 137:
```suggestion
// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#disable-a-selected-organization-for-github-actions-in-an-enterprise
```

WillAbides (changes requested):
This looks good except for a few doc urls that need to be updated to point to enterprise-cloud.

WillAbides (approved):

gmlewis:
According to our CONTRIBUTING.md:

---

## Other notes on code organization ##

Currently, everything is defined in the main `github` package, with API methods
broken into separate service objects. These services map directly to how
the [GitHub API documentation][] is organized, so use that as your guide for
where to put new methods.

Code is organized in files also based pretty closely on the GitHub API
documentation, following the format `{service}_{api}.go`. For example, methods
defined at <https://docs.github.com/en/rest/webhooks/repos> live in
[repos_hooks.go][].

[GitHub API documentation]: https://docs.github.com/en/rest
[repos_hooks.go]: https://github.com/google/go-github/blob/master/github/repos_hooks.go

---

and I would like to preserve this convention to the best of our ability.
So I'm thinking that the `ActionsService` calls should be kept in files starting with `actions_*.go`.

Although I'm trying to catch up on a large backlog, I don't want to rush this one through without thinking about it, which is going to involve looking at the official docs and trying to make sense of which API calls belong to which `Service`, and therefore belong in which source file. If you would like to do the mapping, that would be great, otherwise I think I will ponder on this one until I figure out what seems to be the correct mapping according to this repo's conventions.

gmlewis:
Ah, I see that I had the [same challenges][https://github.com/google/go-github/pull/2518#discussion_r1004485038] with #2518 as well.

gmlewis:
For reference, here are the affected endpoints and their official docs locations:

| Endpoint | Official Docs URL |
| --- | --- |
| GET orgs/%v/actions/permissions/repositories | https://docs.github.com/en/rest/actions/permissions#list-selected-repositories-enabled-for-github-actions-in-an-organization |
| PUT orgs/%v/actions/permissions/repositories | https://docs.github.com/en/rest/actions/permissions#set-selected-repositories-enabled-for-github-actions-in-an-organization |
| PUT orgs/%v/actions/permissions/repositories/%v | https://docs.github.com/en/rest/actions/permissions#enable-a-selected-repository-for-github-actions-in-an-organization |
| DELETE orgs/%v/actions/permissions/repositories/%v | https://docs.github.com/en/rest/actions/permissions#disable-a-selected-repository-for-github-actions-in-an-organization |
| GET enterprises/%v/actions/permissions/selected-actions | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#get-allowed-actions-and-reusable-workflows-for-an-enterprise |
| PUT enterprises/%v/actions/permissions/selected-actions | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#set-allowed-actions-and-reusable-workflows-for-an-enterprise |
| GET enterprises/%v/actions/permissions | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#get-github-actions-permissions-for-an-enterprise |
| PUT enterprises/%v/actions/permissions | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#set-github-actions-permissions-for-an-enterprise |
| GET enterprises/%v/actions/permissions/organizations | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#list-selected-organizations-enabled-for-github-actions-in-an-enterprise |
| PUT enterprises/%v/actions/permissions/organizations | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#set-selected-organizations-enabled-for-github-actions-in-an-enterprise |
| PUT enterprises/%v/actions/permissions/organizations/%v | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#enable-a-selected-organization-for-github-actions-in-an-enterprise |
| DELETE enterprises/%v/actions/permissions/organizations/%v | https://docs.github.com/enterprise-cloud@latest/rest/actions/permissions#disable-a-selected-organization-for-github-actions-in-an-enterprise |

Next step - figure out which filename each endpoint should live in according to our stated goals in `CONTRIBUTING.md`... more later...

gmlewis:
> Next step - figure out which filename each endpoint should live in according to our stated goals in `CONTRIBUTING.md`... more later...

After taking the time to look at all the GitHub documentation linked in the table above, it is now clear to me that all of these endpoints should be available in the `ActionsService` and all endpoints should appear in the file `actions_permissions.go`.

Seeing that there are two flavors of endpoints (those for `Orgs` and those for `Enterprises`), I'm open to two options:

* Option 1: place all endpoints in `actions_permissions.go`, or
* Option 2: split the endpoints into `actions_permissions_orgs.go` for the first four endpoints in the [table above](https://github.com/google/go-github/pull/2920#issuecomment-1728336784) and the remainder into `actions_permissions_enterprise.go`.

I'm fine with either, but I'm betting you might prefer Option 2.  (Note that all new files should have copyright year 2023 according to the Google Open Source legal team when I was there.)

I apologize for the delay, but please proceed using either option above.

Thank you, @RickleAndMortimer !

RickleAndMortimer:
I'll run through option 2, I'll get to updating this PR asap. Thanks for the update!

github/actions_permissions_enterprise.go
```diff
@@ -0,0 +1,152 @@
+// Copyright 2022 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
```

github/actions_permissions_enterprise_test.go
```diff
@@ -0,0 +1,221 @@
+// Copyright 2022 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
```

github/actions_permissions_orgs.go
```diff
@@ -0,0 +1,204 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
+// Copyright 2021 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1:
```suggestion
// Copyright 2023 The go-github AUTHORS. All rights reserved.
```

github/actions_permissions_enterprise_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestEnterpriseService_GetActionsPermissions(t *testing.T) {
```
gmlewis, line 18: Please check all your test names, because the `Service`s have changed.

gmlewis (changes requested):
Instead of completely deleting the methods that are being moved from one service to another, can you instead please mark them as:

```go
// Deprecated: please use `...` instead.
```

That way, this will not be a "Breaking API" PR.

RickleAndMortimer:
I'm planning on just having the functions dealing with organizations as written before this PR and anything enterprise-related has it's own function, so nothing should be deprecated.

github/actions_permissions_orgs.go
```diff
@@ -0,0 +1,204 @@
...
+	p := new(ActionsPermissions)
+	resp, err := s.client.Do(ctx, req, p)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return p, resp, nil
+}
+
+// ListEnabledRepos lists the selected repositories that are enabled for GitHub Actions in an organization.
```
gmlewis, line 84: I'm surprised the linter didn't catch that this name (`ListEnabledRepos`) doesn't match the method name (`ListEnabledReposInOrg`).
Please fix.

github/actions_permissions_orgs.go
```diff
@@ -0,0 +1,204 @@
...
+	repos := &ActionsEnabledOnOrgRepos{}
+	resp, err := s.client.Do(ctx, req, repos)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return repos, resp, nil
+}
+
+// SetEnabledRepos replaces the list of selected repositories that are enabled for GitHub Actions in an organization..
```
gmlewis, line 108: ditto

github/actions_permissions_orgs.go
```diff
@@ -0,0 +1,204 @@
...
+
+	resp, err := s.client.Do(ctx, req, nil)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// AddEnabledRepos adds a repository to the list of selected repositories that are enabled for GitHub Actions in an organization.
```
gmlewis, line 129: ditto

github/actions_permissions_orgs.go
```diff
@@ -0,0 +1,204 @@
...
+
+	resp, err := s.client.Do(ctx, req, nil)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// RemoveEnabledRepo removes a single repository from the list of enabled repos for GitHub Actions in an organization.
```
gmlewis, line 148: ditto

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+		fmt.Fprint(w, `{"total_count":2,"repositories":[{"id":2}, {"id": 3}]}`)
+	})
+
+	ctx := context.Background()
+	opt := &ListOptions{
+		Page: 1,
+	}
+	got, _, err := client.Actions.ListEnabledReposInOrg(ctx, "o", opt)
+	if err != nil {
+		t.Errorf("Actions.ListEnabledRepos returned error: %v", err)
```
gmlewis, line 114:
```suggestion
		t.Errorf("Actions.ListEnabledReposInOrg returned error: %v", err)
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	want := &ActionsEnabledOnOrgRepos{TotalCount: int(2), Repositories: []*Repository{
+		{ID: Int64(2)},
+		{ID: Int64(3)},
+	}}
+	if !cmp.Equal(got, want) {
+		t.Errorf("Actions.ListEnabledRepos returned %+v, want %+v", got, want)
```
gmlewis, line 122:
```suggestion
		t.Errorf("Actions.ListEnabledReposInOrg returned %+v, want %+v", got, want)
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	const methodName = "ListEnabledRepos"
```
gmlewis, line 125:
```suggestion
	const methodName = "ListEnabledReposInOrg"
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+		testMethod(t, r, "PUT")
+		testHeader(t, r, "Content-Type", "application/json")
+		testBody(t, r, `{"selected_repository_ids":[123,1234]}`+"\n")
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	ctx := context.Background()
+	_, err := client.Actions.SetEnabledReposInOrg(ctx, "o", []int64{123, 1234})
+	if err != nil {
+		t.Errorf("Actions.SetEnabledRepos returned error: %v", err)
```
gmlewis, line 154:
```suggestion
		t.Errorf("Actions.SetEnabledReposInOrg returned error: %v", err)
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	const methodName = "SetEnabledRepos"
```
gmlewis, line 157:
```suggestion
	const methodName = "SetEnabledReposInOrg"
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+	mux.HandleFunc("/orgs/o/actions/permissions/repositories/123", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "PUT")
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	ctx := context.Background()
+	_, err := client.Actions.AddEnabledReposInOrg(ctx, "o", 123)
+	if err != nil {
+		t.Errorf("Actions.AddEnabledRepos returned error: %v", err)
```
gmlewis, line 181:
```suggestion
		t.Errorf("Actions.AddEnabledReposInOrg returned error: %v", err)
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	const methodName = "AddEnabledRepos"
```
gmlewis, line 184:
```suggestion
	const methodName = "AddEnabledReposInOrg"
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+	mux.HandleFunc("/orgs/o/actions/permissions/repositories/123", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusNoContent)
+	})
+
+	ctx := context.Background()
+	_, err := client.Actions.RemoveEnabledReposInOrg(ctx, "o", 123)
+	if err != nil {
+		t.Errorf("Actions.RemoveEnabledRepo returned error: %v", err)
```
gmlewis, line 208:
```suggestion
		t.Errorf("Actions.RemoveEnabledReposInOrg returned error: %v", err)
```

github/actions_permissions_orgs_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	const methodName = "RemoveEnabledRepo"
```
gmlewis, line 211:
```suggestion
	const methodName = "RemoveEnabledReposInOrg"
```

github/orgs_actions_allowed.go
```diff
@@ -1,63 +0,0 @@
...
-}
-
-func (a ActionsAllowed) String() string {
-	return Stringify(a)
-}
-
-// GetActionsAllowed gets the actions that are allowed in an organization.
-//
-// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#get-allowed-actions-and-reusable-workflows-for-an-organization
-func (s *OrganizationsService) GetActionsAllowed(ctx context.Context, org string) (*ActionsAllowed, *Response, error) {
```
gmlewis, line 29: My point in the last review was that this (and several other) endpoints no longer exist on this `OrganizationsService`, and that it would be better for our users to change all these moved endpoints to something like the following:

```go
// GetActionsAllowed gets the actions that are allowed in an organization.
//
// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#get-allowed-actions-and-reusable-workflows-for-an-organization
// Deprecated: please use `client.Actions.GetActionsAllowed` instead.
func (s *OrganizationsService) GetActionsAllowed(ctx context.Context, org string) (*ActionsAllowed, *Response, error) {
	s2 := (*ActionsService)(s)
	return s2.GetActionsAllowed(ctx, org)
}
```

I believe this should work, but please test it out.

gmlewis (changes requested):

github/orgs_actions_allowed.go
RickleAndMortimer: That makes sense, I'll proceed with that approach then. Thanks for clarifying 👍

RickleAndMortimer:
Should I also add the deleted tests as well for the deprecated functions? seeing as they will be using the functions under the actions service I'm not sure if it is still necessary to do so

gmlewis:
> Should I also add the deleted tests as well for the deprecated functions? seeing as they will be using the functions under the actions service I'm not sure if it is still necessary to do so

Let's please leave them for now so that our Codecov doesn't artificially drop.

Also, I'd like to figure out why it is already dropped a few percent in this PR if we could nail that down...

RickleAndMortimer:
Could the loss of code coverage be from my PR being 10 commits behind? Would that fix the issue or is there something I'm missing

gmlewis:
> Could the loss of code coverage be from my PR being 10 commits behind? Would that fix the issue or is there something I'm missing

Hmmm... I doubt it but will investigate and get back to you...

gmlewis:
Every now and then, `codecov` just has weird reports, and I think this is the case.
We can safely ignore it.  Here's a screenshot:
pr-2920-2023-09-28_12-08-04

RickleAndMortimer:
Disregarding the codecov issue, Is there anything else I should include with this PR?

github/orgs_actions_allowed.go
```diff
@@ -0,0 +1,28 @@
...
+package github
+
+import (
+	"context"
+)
+
+// GetActionsAllowed gets the actions that are allowed in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#get-allowed-actions-and-reusable-workflows-for-an-organization
+// Deprecated: please use `client.Actions.GetActionsAllowed` instead
```
gmlewis, line 15: nit: I'll commit it - no action necessary.

```suggestion
// Deprecated: please use `client.Actions.GetActionsAllowed` instead.
```

github/orgs_actions_permissions.go
```diff
@@ -0,0 +1,28 @@
...
+package github
+
+import (
+	"context"
+)
+
+// GetActionsPermissions gets the GitHub Actions permissions policy for repositories and allowed actions in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#get-github-actions-permissions-for-an-organization
+// Deprecated: please use `client.Actions.GetActionsPermissions` instead
```
gmlewis, line 15:
```suggestion
// Deprecated: please use `client.Actions.GetActionsPermissions` instead.
```

github/orgs_actions_permissions.go
```diff
@@ -0,0 +1,28 @@
...
+func (s *OrganizationsService) GetActionsPermissions(ctx context.Context, org string) (*ActionsPermissions, *Response, error) {
+	s2 := (*ActionsService)(s)
+	return s2.GetActionsPermissions(ctx, org)
+}
+
+// EditActionsPermissions sets the permissions policy for repositories and allowed actions in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#set-github-actions-permissions-for-an-organization
+// Deprecated: please use `client.Actions.EditActionsPermissions` instead
```
gmlewis, line 24:
```suggestion
// Deprecated: please use `client.Actions.EditActionsPermissions` instead.
```

github/orgs_actions_allowed.go
```diff
@@ -0,0 +1,28 @@
...
+func (s *OrganizationsService) GetActionsAllowed(ctx context.Context, org string) (*ActionsAllowed, *Response, error) {
+	s2 := (*ActionsService)(s)
+	return s2.GetActionsAllowed(ctx, org)
+}
+
+// EditActionsAllowed sets the actions that are allowed in an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/actions/permissions#set-allowed-actions-and-reusable-workflows-for-an-organization
+// Deprecated: please use `client.Actions.EditActionsAllowed` instead
```
gmlewis, line 24:
```suggestion
// Deprecated: please use `client.Actions.EditActionsAllowed` instead.
```

gmlewis:
> Disregarding the codecov issue, Is there anything else I should include with this PR?

Thank you for your patience and persistence with this PR, @RickleAndMortimer !
LGTM.

As far as I can tell, this is _NOT_ a breaking-API change, which is nice.

Merging.

gmlewis (approved):

RickleAndMortimer:
I could say the same to you as well, thank you for your guidance!

PR #2924, opened 2023-09-11
Handle Encoding value of "none"

GusPrice:
When a file is over 1 MB in a repository, GitHub responds with "none" as the encoding value. Handle this accordingly.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -86,7 +86,7 @@ func (r *RepositoryContent) GetContent() (string, error) {
 		}
 		c, err := base64.StdEncoding.DecodeString(*r.Content)
 		return string(c), err
-	case "":
+	case "", "none":
 		if r.Content == nil {
 			return "", nil
 		}


GusPrice:
Ah! It seems in fact I have misunderstood! Definitely would be nice to have a better error though. I'll update my PR accordingly.

GusPrice:
Updated to give a nicer error message in that case!

github/repos_contents_test.go
```diff
@@ -54,6 +54,12 @@ func TestRepositoryContent_GetContent(t *testing.T) {
 			want:     "",
 			wantErr:  true,
 		},
+		{
+			encoding: String("none"),
+			content:  String(""),
```
WillAbides, line 59:
```suggestion
```

content will be nil when encoding is none, so leaving it nil here is a slightly more realistic test case.

WillAbides (approved):
LGTM. Just a small nit about the test case, but you don't necessarily need to take my suggestion there.

To set expectations, the primary maintainer is out until 9/18, so this PR is not likely to be merged before then.

github/repos_contents_test.go
GusPrice: good catch!

WillAbides (approved):

github/repos_contents.go
```diff
@@ -91,6 +91,8 @@ func (r *RepositoryContent) GetContent() (string, error) {
 			return "", nil
 		}
 		return *r.Content, nil
+	case "none":
+		return "", fmt.Errorf("unsupported content encoding: %v, this may occur when file size > 1 MB, if that is the case consider using DownloadContents", encoding)
```
gmlewis, line 95:
```suggestion
		return "", errors.New("unsupported content encoding: none, this may occur when file size > 1 MB, if that is the case consider using DownloadContents")
```

gmlewis (approved):
Thank you, @GusPrice and @WillAbides !
LGTM.
Merging.

Issue #2929, opened 2023-09-14
Add support for additional response fields


vandanrohatgi:
The struct for parsing [Secret Scanning Alerts](https://docs.github.com/en/free-pro-team@latest/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#get-a-secret-scanning-alert) :

https://github.com/google/go-github/blob/fd33b81221685302178ba197dbabe71326ca9051/github/secret_scanning.go#L18-L32

does not support a few fields such as :

```
"push_protection_bypassed_by": null,
"push_protection_bypassed": false,
"push_protection_bypassed_at": null,
"resolution_comment": null
```

As always happy to create a PR if this is a valid issue 😃

WillAbides:
This looks like a valid issue to me. It's also missing updated_at according to the response schema.

gmlewis:
Please note that our goal in this repo has never been to have complete 100% coverage of all possible fields in all possible responses, but instead to have all needed fields that users of this repo will want to use.

For the longest time, we never populated the `URL` or `HTMLURL` fields of any response until someone gave an argument for why they needed them. I still can't remember why they were needed, but now they have been added to many structs.

Anyway, I just want to make it clear to all contributors (especially our enthusiastic "hacktoberfest" participants) that it is _NOT_ a goal of ours to create PRs to add missing fields just for the sheer fun (or points score) of it, but that they are added as they are needed.  Thanks.

Also note, I'm not saying that you are making this PR for the fun of it, or to score points.
I just wanted to reiterate a guiding principal of this repo that probably got buried in some PR comment a decade :open_mouth: ago. :smile:

vandanrohatgi:
Hey @gmlewis ! long time no see 😄 . I agree with your policy, never liked doing stuff for the sake of it.

I wanted to create this because I had some work which involved these particular fields (and it was getting tiring working with my fork of this repo 😅 )

PR #2930, opened 2023-09-15
Add missing secret scanning alert fields

vandanrohatgi:
Closes https://github.com/google/go-github/issues/2929

- [x] go generate
- [x] go vet
- [x] go test


WillAbides (approved):
LGTM. The lint failure is a timeout. I verified locally that lint passes on this branch, but somebody with permissions will need to trigger re-running the action.

PR #2931, opened 2023-09-18
Fix golangci-lint timeout failures

gmlewis:
Ever since #2903 , the golangci-lint GitHub Actions workflow has frequently timed out.
According to this: https://github.com/golangci/golangci-lint-action/issues/308
this PR should fix that problem.


gmlewis:
Thank you, @WillAbides !
Merging.

PR #2933, opened 2023-09-19
Fix SecretScanning API by switching arguments from url to json

gmlewis:
Closes: #2871.

This PR fixes the endpoint:
https://docs.github.com/en/enterprise-server@3.5/rest/secret-scanning/secret-scanning#update-a-secret-scanning-alert
by sending its parameters in the body instead of as URL query parameters.

I was unable to edit #2871 directly, so this is a replacement.
Thank you, @jentfoo for the bug report and initial PR!


gmlewis:
Yikes, the commit history is causing problems.  Starting fresh by closing.

PR #2935, opened 2023-09-19
Remove openpgp and change CreateCommit signature

### LINKED ISSUE #2932: Transitive dependency (cloudflare/circl) making cross compiling difficult
I'm running into a slightly obscure build problem. Using bazel (and gazelle and rules_go) to build go-github worked fine with v50 both for arm64 and amd64. I recently tried to bump to v55 and found myself unable to build for amd64 via bazel.

I found that the transitive dependency `github.com/cloudflare/circl` was bumped from v1.1.0 to v1.3.3 after v52 of go-github. It seems used by `github.com/ProtonMail/go-crypto`.

The problem is that it includes some platform specific files (`*_amd64.go`) for some packages, and some headers and some assembly.

Unfortunately led to some issues trying to build via bazel with dependencies not detected:
```
external/com_github_cloudflare_circl/dh/x25519/curve_amd64.s:7: #include: open external/com_github_cloudflare_circl/math/fp25519/fp_amd64.h: no such file or directory
```
Quickly playing with `GOARCH=amd64` it seems to work fine with `go build` so I suspect it's not actually a problem but more of a side effect of bazel trying to compile a bit too much of some of the libraries used transitively.

Since CGO usually raises the bar for compiling I thought I would mention that it might not be ideal that `github.com/ProtonMail/go-crypto` uses `github.com/cloudflare/circl`:
* Depending on the architecture it might need CGO for some packages
* https://github.com/cloudflare/circl claims to aim to be an experimental lib which might maybe not ideal to depend on transitively
  > The goal of this library is to be used as a tool for experimental deployment of cryptographic algorithms targeting Post-Quantum (PQ) and Elliptic Curve Cryptography (ECC)

Is there an alternative to `github.com/ProtonMail/go-crypto` that would avoid this?

WillAbides:
closes #2932

In #2932 we discussed removing the dependency on openpgp and replacing Commit.SigningKey with an interface.

While I was working on it, I realized that this should be an option to CreateCommit instead of a Commit field. I was going to just ignore this, but adding an interface field to Commit caused issues with Stringify and the generated string tests. Instead of unwinding all that, I'm doing the easy thing first and adding a CreateCommitOptions struct. I can change approaches if necessary.


example/commitpr/main.go
```diff
@@ -135,7 +139,25 @@ func pushCommit(ref *github.Reference, tree *github.Tree) (err error) {
...
+		if e != nil {
+			return e
+		}
+		if len(keyring) != 1 {
+			return errors.New("expected exactly one key in the keyring")
+		}
+		key := keyring[0]
+		opts.Signer = github.MessageSignerFunc(func(w io.Writer, r io.Reader) error {
+			return openpgp.ArmoredDetachSign(w, key, r, nil)
+		})
```
gmlewis, line 158: TIL that you can wrap a single `func` in a type cast that can then be used as a receiver in order to satisfy an interface...
so cool, @WillAbides - thank you!

gmlewis (approved):
Awesome, @WillAbides !
LGTM.

valbeat (approved):

gmlewis:
Thank you, @valbeat !
Merging.

PR #2939, opened 2023-09-25
Allow up to maxRedirects upon receiving HTTP 301 status

k0ral:
The [`roundTripWithOptionalFollowRedirect()` method](https://github.com/google/go-github/blob/b45ef89f2dd0ec803ab8f27725bdb4941c0cbc99/github/github.go#L1561) currently takes a boolean parameter `followRedirects`, used to support HTTP 301 redirections. However, at most 1 redirection is handled at the moment (the parameter is set to `false` after the first iteration.)

I've stumbled upon multiple repositories where `https://api.github.com/repos/<OWNER>/<PROJECT>/branches/<BRANCH>` redirects twice ; as a result, the [`GetBranch()` method](https://github.com/google/go-github/blob/b45ef89f2dd0ec803ab8f27725bdb4941c0cbc99/github/repos.go#L1277) only follows the first redirection, then returns the 2nd HTTP 301 response.

This MR proposes to generalize the `followRedirects` boolean into a `maxRedirects` integer, to support an arbitrary number of redirections. This impacts multiple methods, not just `GetBranch()`.

:warning: This is a breaking change.

### DIFF
--- a/github/actions_artifacts.go
+++ b/github/actions_artifacts.go
@@ -121,10 +121,10 @@ func (s *ActionsService) GetArtifact(ctx context.Context, owner, repo string, ar
 // DownloadArtifact gets a redirect URL to download an archive for a repository.
 //
 // GitHub API docs: https://docs.github.com/en/rest/actions/artifacts#download-an-artifact
-func (s *ActionsService) DownloadArtifact(ctx context.Context, owner, repo string, artifactID int64, followRedirects bool) (*url.URL, *Response, error) {
+func (s *ActionsService) DownloadArtifact(ctx context.Context, owner, repo string, artifactID int64, maxRedirects int) (*url.URL, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/artifacts/%v/zip", owner, repo, artifactID)
 
-	resp, err := s.client.roundTripWithOptionalFollowRedirect(ctx, u, followRedirects)
+	resp, err := s.client.roundTripWithOptionalFollowRedirect(ctx, u, maxRedirects)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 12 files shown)


gmlewis (approved):
Thank you, @k0ral !
LGTM.

valbeat (approved):

gmlewis:
Thank you, @valbeat !
Merging.

WillAbides:
I know this is already merged, but I would like to suggest that since we are doing a breaking change for these methods we consider updating them to accept an options struct instead of just the maxRedirects option. This will allow future options to be added without a breaking change. If they previously used an options struct this wouldn't need to be a breaking change. We can't avoid a break now, but we can avoid future breaks.

gmlewis:
Good idea, @WillAbides - do you want to make a PR that addresses this and we will hold off on the next release until it is merged?

gmlewis:
@WillAbides - I was [requested](https://github.com/google/go-github/pull/2947#issuecomment-1756982704) to cut a new release, but just remembered that [you wanted](https://github.com/google/go-github/pull/2939#issuecomment-1747392285) to change this `maxRedirects` API to use options instead, and thereby break the API twice before a new release instead of breaking the API twice between two different major releases.

Do you want me to hold off on cutting the new release, or shall we simply go ahead and break it twice?

I'm leaning toward breaking it twice at the moment, since I like to be responsive when users want a new release and it has been about a month since the last one.

gmlewis:
I'll go ahead on the release. We can work on an improved interface for the next breaking release.

WillAbides:
@gmlewis I agree with going ahead with the release because the need to move to options is hypothetical at this point.

PR #2948, opened 2023-10-03
Escape branch string before inserting it in URL

k0ral:
When a branch name with special characters like `%` is passed to the [`RepositoryService.GetBranch()` method](https://github.com/google/go-github/blob/8cd452b85e821c32eefcc5964e91e57f9731891f/github/repos.go#L1277) as is, an error is currently returned:
```
invalid URL escape "%"
```

Passing an escaped branch name to `RepositoriesService.GetBranch()` works fine.

Here's a minimal reproducer
```golang
import (
	"context"
	"net/http"
	"testing"

	"github.com/google/go-github/v53/github"
	"github.com/stretchr/testify/require"
)

func TestGetBranch(t *testing.T) {
	t.Parallel()
	ctx := context.Background()

	githubClient := github.NewClient(&http.Client{
		Transport: http.DefaultTransport,
	})
	owner := "samstickkz"
	project := "curr"
	branchName := "Feature/authentication_90%"
	branch, response, err := githubClient.Repositories.GetBranch(ctx, owner, project, branchName, false)

	require.Nil(t, err)
	t.Logf("Response: %+v", response)
	t.Logf("Branch: %+v", branch)
}
```

This sounds like a bug to me, as I believe users of that method shouldn't be expected to escape the branch name before calling it, as this is a private implementation quirk that they shouldn't be aware of. This PR patches all branch-related methods to automatically escape the branch name in the endpoint URL. 

There's actually [a precedent for `ref` parameter](https://github.com/google/go-github/pull/2454), and I suspect that other variables are prone to the same issue. I am conservatively patching only the branch variable, as this is the only one I have tested thoroughly. Note that, contrary to `ref` use case, my tests suggest that it's okay to escape `/` characters for branch names. 

:warning: Breaking change: for users that currently _bypass_ this issue by feeding escaped strings to `RepositoryService.GetBranch()` method (and other branch-related methods), this change will be breaking.

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 )
 
@@ -1273,9 +1274,11 @@ func (s *RepositoriesService) ListBranches(ctx context.Context, owner string, re
 
 // GetBranch gets the specified branch for a repository.
 //
+// Note: the branch name is URL path escaped for you. See: https://pkg.go.dev/net/url#PathEscape .
+//
 // GitHub API docs: https://docs.github.com/en/rest/branches/branches#get-a-branch
 func (s *RepositoriesService) GetBranch(ctx context.Context, owner, repo, branch string, maxRedirects int) (*Branch, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/branches/%v", owner, repo, branch)
+	u := fmt.Sprintf("repos/%v/%v/branches/%v", owner, repo, url.PathEscape(branch))
 
 	resp, err := s.client.roundTripWithOptionalFollowRedirect(ctx, u, maxRedirects)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


gmlewis:
Honestly, I'm quite surprised that this issue has never come up before... as this repo is how many years old? 10?!?

However, I think you are right.

I'm wondering if each of the modified endpoints needs a comment in the documentation stating something like:

```
// Note that the branch name is URL path escaped for you. See: https://pkg.go.dev/net/url#PathEscape
```

gmlewis (changes requested):
Can you please add unit tests that demonstrate this new super power?

k0ral:
@gmlewis PR updated with additional comments and unit tests.

gmlewis (approved):
Thank you, @k0ral !
LGTM.

WillAbides:
I have a bit of concern about this being a subtle breaking change that could take a long time to discover after upgrading. I'm looking at ways to mitigate that. I'll be back to review later today.

WillAbides (approved):
I'm still a little hesitant about the breaking change, but after a some pondering I can't come up with anything better that doesn't add unnecessary complexity.

gmlewis:
Thank you, @WillAbides !
Merging.

PR #2951, opened 2023-10-09
Add prevent_self_review on repo environments

brekelj1:
Add prevent_self_review field on repo/deployment environments. See https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment

### DIFF
--- a/github/repos_environments.go
+++ b/github/repos_environments.go
@@ -17,7 +17,6 @@ type Environment struct {
 	Owner                  *string         `json:"owner,omitempty"`
 	Repo                   *string         `json:"repo,omitempty"`
 	EnvironmentName        *string         `json:"environment_name,omitempty"`
-	PreventSelfReview      *bool           `json:"prevent_self_review,omitempty"`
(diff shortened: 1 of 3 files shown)


github/repos_environments.go
```diff
@@ -17,6 +17,7 @@ type Environment struct {
 	Owner                  *string         `json:"owner,omitempty"`
 	Repo                   *string         `json:"repo,omitempty"`
 	EnvironmentName        *string         `json:"environment_name,omitempty"`
+	PreventSelfReview      *bool           `json:"prevent_self_review,omitempty"`
```
gmlewis, line 20: Why is this needed in the `Environment` struct?  
Using the link in your PR description, I only see it being needed in the `ProtectionRule`.

brekelj1: Good point, I thought this was the same struct used for CreateUpdateEnvironment.

gmlewis (approved):
Thank you, @brekelj1 !
LGTM.
Merging.

PR #2969, opened 2023-10-20
Move RateLimits method to a service

### LINKED ISSUE #2968: Move RateLimits method to a service
I missed the RateLimits method in #2937. It should be moved to a service struct.

Docs are at [rate-limit/rate-limit#get-rate-limit-status-for-the-authenticated-user](https://docs.github.com/rest/rate-limit/rate-limit#get-rate-limit-status-for-the-authenticated-user), so according to our existing pattern, we should move it to a new service called `RateLimitService` and name it `Get`

`Client.RateLimits` should be marked as deprecated and its body should be updated to `return c.RateLimit.Get(ctx)`

This may be a candidate for "good first issue" with the slight complication of creating a new service.

/cc #2936

nhAnik:
Closes #2968


gmlewis (commented):
Thank you, @nhAnik !
Anywhere you find `client.RateLimits` (such as `test/integration/misc_test.go`), it needs to be updated to call the new service instead.

WillAbides (commented):
This LGTM except for the lint failure. You can fix that by calling your new method instead of client.RateLimits

WillAbides (approved):
LGTM

gmlewis (approved):
Thank you, @nhAnik and @WillAbides !
LGTM.
Merging.

PR #2972, opened 2023-10-23
Lint godoc comments

WillAbides:
This adds linting for godoc comments. We had several cases where the method name didn't match the comment.

I had to make golangci-lint not use default exclusions in order to make it check godoc. There were only a couple of other default exclusions that affected go-github, so I only added those back.


gmlewis (approved):
Awesome, thank you, @WillAbides !
LGTM.
Merging.

PR #2973, opened 2023-10-24
Add Copilot endpoints

### LINKED ISSUE #2938: Support for Copilot API
Any timeline for support of the new [Copilot](https://docs.github.com/en/rest/copilot/copilot-for-business?apiVersion=2022-11-28) part of the API?

o-sama:
Closes #2938.


o-sama:
CodeCov hasn't updated as of last commit, but I think it _should_ increase the coverage to close to 100% for the new file.

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+
+type CopilotSeats struct {
+	TotalSeats int64                `json:"total_seats"`
+	Seats      []CopilotSeatDetails `json:"seats"`
+}
+
+// CopilotSeatDetails represents the details of a Copilot for Business seat.
+// Assignee can either be a User, Team, or Organization.
+type CopilotSeatDetails struct {
+	Assignee                interface{} `json:"assignee"`
```
WillAbides, line 46: It may be ok to use `*User` here. There are a few other places where we use `*User` and let users distinguish using the `Type` field. @gmlewis will know what the preferred method is.

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+type SeatBreakdown struct {
+	Total               int64 `json:"total"`
+	AddedThisCycle      int64 `json:"added_this_cycle"`
+	PendingCancellation int64 `json:"pending_cancellation"`
+	PendingInvitation   int64 `json:"pending_invitation"`
+	ActiveThisCycle     int64 `json:"active_this_cycle"`
+	InactiveThisCycle   int64 `json:"inactive_this_cycle"`
+}
+
+type CopilotSeats struct {
```
gmlewis, line 38: Missing godoc-style comment.

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	TotalSeats int64                `json:"total_seats"`
+	Seats      []CopilotSeatDetails `json:"seats"`
```
gmlewis, line 40:
```suggestion
	Seats      []*CopilotSeatDetails `json:"seats"`
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, resp, err
+	}
+
+	return seatAssignments, resp, nil
+}
+
+// RemoveCopilotTeams Removes teams from the Copilot for Business subscription for an organization
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#remove-teams-from-the-copilot-for-business-subscription-for-an-organization
+
```
WillAbides, line 187:
```suggestion
```

In case the preview isn't clear...this removes an empty line.

github/copilot.go
gmlewis, line 46:
```suggestion
type CopilotSeatDetails struct {
	// Assignee can either be a User, Team, or Organization.
	Assignee                interface{} `json:"assignee"`
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+			return fmt.Errorf("unsupported assignee type %s", v["type"].(string))
+		}
+	default:
+		return fmt.Errorf("unsupported assignee type %T", v)
+	}
+
+	return nil
+}
+
+// GetCopilotBilling Gets Copilot for Business seat information and settings for an organization
```
gmlewis, line 124:
```suggestion
// GetCopilotBilling gets Copilot for Business billing information and settings for an organization.
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+
+// SelectedTeams represents the teams selected for the Copilot for Business subscription.
+type SelectedTeams struct {
+	SelectedTeams []string `json:"selected_teams"`
+}
+
+// SelectedUsers represents the users selected for the Copilot for Business subscription.
+type SelectedUsers struct {
+	SelectedUsers []string `json:"selected_users"`
+}
```
WillAbides, line 63: These can be removed and replaced with `[]string` in the methods that take them as an argument. When you do that, please name the variable something like `teamNames` so that it is clear it takes names and not ids.

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	var copilotDetails *OrganizationCopilotDetails
+	resp, err := s.client.Do(ctx, req, &copilotDetails)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return copilotDetails, resp, nil
+}
+
+// ListCopilotSeats Gets Copilot for Business seat assignments for an organization
```
gmlewis, line 144:
```suggestion
// ListCopilotSeats lists Copilot for Business seat assignments for an organization.
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	var copilotSeats *CopilotSeats
+	resp, err := s.client.Do(ctx, req, &copilotSeats)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return copilotSeats, resp, nil
+}
+
+// AddCopilotTeams Adds teams to the Copilot for Business subscription for an organization
```
gmlewis, line 164:
```suggestion
// AddCopilotTeams adds teams to the Copilot for Business subscription for an organization.
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#add-teams-to-the-copilot-for-business-subscription-for-an-organization
```
gmlewis, line 166:
```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-teams-to-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+func (s *CopilotService) AddCopilotTeams(ctx context.Context, org string, teams SelectedTeams) (*SeatAssignments, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/selected_teams", org)
+
+	req, err := s.client.NewRequest("PUT", u, teams)
```
gmlewis, line 170:
```suggestion
	req, err := s.client.NewRequest("POST", u, teams)
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// SelectedUsers represents the users selected for the Copilot for Business subscription.
+type SelectedUsers struct {
+	SelectedUsers []string `json:"selected_users"`
+}
+
+// SeatAssignments represents the number of seats assigned.
+type SeatAssignments struct {
+	SeatsCreated int64 `json:"seats_created"`
```
gmlewis, line 67: I think this can safely be an `int`. We typically reserve `int64` for IDs, and counts we usually use `int`.

```suggestion
	SeatsCreated int `json:"seats_created"`
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// SeatCancellations represents the number of seats cancelled.
+type SeatCancellations struct {
+	SeatsCancelled int64 `json:"seats_cancelled"`
```
gmlewis, line 72: Same here.

```suggestion
	SeatsCancelled int `json:"seats_cancelled"`
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// SeatBreakdown represents the breakdown of Copilot for Business seats for the organization.
+type SeatBreakdown struct {
+	Total               int64 `json:"total"`
+	AddedThisCycle      int64 `json:"added_this_cycle"`
+	PendingCancellation int64 `json:"pending_cancellation"`
+	PendingInvitation   int64 `json:"pending_invitation"`
+	ActiveThisCycle     int64 `json:"active_this_cycle"`
+	InactiveThisCycle   int64 `json:"inactive_this_cycle"`
```
gmlewis, line 35: Using the reasoning below, let's also please change these to `int`.

```suggestion
	Total               int `json:"total"`
	AddedThisCycle      int `json:"added_this_cycle"`
	PendingCancellation int `json:"pending_cancellation"`
	PendingInvitation   int `json:"pending_invitation"`
	ActiveThisCycle     int `json:"active_this_cycle"`
	InactiveThisCycle   int `json:"inactive_this_cycle"`
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+// SeatAssignments represents the number of seats assigned.
+type SeatAssignments struct {
+	SeatsCreated int64 `json:"seats_created"`
+}
+
+// SeatCancellations represents the number of seats cancelled.
+type SeatCancellations struct {
+	SeatsCancelled int64 `json:"seats_cancelled"`
+}
+
```
WillAbides, line 74: These should have "Copilot" in the name to avoid confusion with any other type of seat in the GitHub API.

I also think that because these are used as responses, it might be better to split these into four structs: `AddCopilotTeamsResponse`, `RemoveCopilotTeamsResponse`, `AddCopilotUsersResponse` and `RemoveCopilotUsersResponse`.

/cc @gmlewis for thoughts

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+// CopilotService provides access to the Copilot-related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/
+type CopilotService service
+
+// OrganizationCopilotDetails represents the details of an organization's Copilot for Business supbscription.
+type OrganizationCopilotDetails struct {
+	SeatBreakdown         *SeatBreakdown `json:"seat_breakdown"`
```
WillAbides, line 22:
```suggestion
	SeatBreakdown         *SeatBreakdown `json:"seat_breakdown,omitempty"`
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	var seatAssignments *SeatAssignments
+	resp, err := s.client.Do(ctx, req, &seatAssignments)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return seatAssignments, resp, nil
+}
+
+// RemoveCopilotTeams Removes teams from the Copilot for Business subscription for an organization
```
gmlewis, line 184:
```suggestion
// RemoveCopilotTeams removes teams from the Copilot for Business subscription for an organization.
```

github/copilot.go
gmlewis, line 187: Please delete the blank line between the godoc and the method.

```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-teams-from-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return SeatCancellations, resp, nil
+}
+
+// AddCopilotUsers Adds users to the Copilot for Business subscription for an organization
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#add-users-to-the-copilot-for-business-subscription-for-an-organization
```
gmlewis, line 207:
```suggestion
// AddCopilotUsers sdds users to the Copilot for Business subscription for an organization.
//
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-users-to-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+func (s *CopilotService) AddCopilotUsers(ctx context.Context, org string, users SelectedUsers) (*SeatAssignments, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/selected_users", org)
+
+	req, err := s.client.NewRequest("PUT", u, users)
```
gmlewis, line 211:
```suggestion
	req, err := s.client.NewRequest("POST", u, users)
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+)
+
+// CopilotService provides access to the Copilot-related functions
+// in the GitHub API.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/
+type CopilotService service
+
+// OrganizationCopilotDetails represents the details of an organization's Copilot for Business supbscription.
+type OrganizationCopilotDetails struct {
```
WillAbides, line 21: This is called copilot-organization-details in github's schema, so it CopilotOrganizationDetails might be a better name.

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return seatAssignments, resp, nil
+}
+
+// RemoveCopilotUsers Removes users from the Copilot for Business subscription for an organization
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#remove-users-from-the-copilot-for-business-subscription-for-an-organization
```
gmlewis, line 227:
```suggestion
// RemoveCopilotUsers removes users from the Copilot for Business subscription for an organization.
//
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-users-from-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+// OrganizationCopilotDetails represents the details of an organization's Copilot for Business supbscription.
+type OrganizationCopilotDetails struct {
+	SeatBreakdown         *SeatBreakdown `json:"seat_breakdown"`
+	PublicCodeSuggestions string         `json:"public_code_suggestions"`
+	CopilotChat           string         `json:"copilot_chat"`
+	SeatManagementSetting string         `json:"seat_management_setting"`
+}
+
+// SeatBreakdown represents the breakdown of Copilot for Business seats for the organization.
+type SeatBreakdown struct {
```
WillAbides, line 29: CopilotSeatBreakdown

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, resp, err
+	}
+
+	return SeatCancellations, resp, nil
+}
+
+// GetSeatDetails Gets Copilot for Business seat assignment details for a user
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#get-copilot-for-business-seat-assignment-details-for-a-user
+func (s *CopilotService) GetSeatDetails(ctx context.Context, org string, user string) (*CopilotSeatDetails, *Response, error) {
```
gmlewis, line 248:
```suggestion
// GetSeatDetails gets Copilot for Business seat assignment details for a user.
//
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#get-copilot-for-business-seat-assignment-details-for-a-user
func (s *CopilotService) GetSeatDetails(ctx context.Context, org, user string) (*CopilotSeatDetails, *Response, error) {
```

gmlewis (changes requested):
Thank you, @o-sama - this is looking great!
Just a few tweaks, please, before we ask for a second review.

github/copilot.go
WillAbides, line 38: This might be better named `ListCopilotSeatsResponse` because it's only used as a response.

o-sama:
Thanks @gmlewis, I'll add in the changes! My bad on the http method being PUT, I was doing some other work with S3 and I think I carried that context over.

github/copilot.go
WillAbides, line 166:
```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-teams-to-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return seatAssignments, resp, nil
+}
+
+// RemoveCopilotTeams Removes teams from the Copilot for Business subscription for an organization
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#remove-teams-from-the-copilot-for-business-subscription-for-an-organization
```
WillAbides, line 186:
```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-teams-from-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
WillAbides, line 207:
```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-users-to-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
WillAbides, line 227:
```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-users-from-the-copilot-for-business-subscription-for-an-organization
```

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+func (s *CopilotService) ListCopilotSeats(ctx context.Context, org string) (*CopilotSeats, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/seats", org)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var copilotSeats *CopilotSeats
+	resp, err := s.client.Do(ctx, req, &copilotSeats)
```
gauraw10, line 156: How are we handling pagination here for more than 100 users?

github/copilot.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return SeatCancellations, resp, nil
+}
+
+// GetSeatDetails Gets Copilot for Business seat assignment details for a user
+//
+// https://docs.github.com/en/rest/copilot/copilot-for-business#get-copilot-for-business-seat-assignment-details-for-a-user
```
WillAbides, line 247:
```suggestion
// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#get-copilot-for-business-seat-assignment-details-for-a-user
```

WillAbides (changes requested):
This looks good over all. I realize I left quite a few comments, but they are mostly nit-picks about naming and comments.

gmlewis:
Thank you, @WillAbides for your review feedback and @gauraw10 for your excellent pagination question which I seem to frequently forget about until it becomes an issue later, unfortunately. It is definitely better to add pagination support up front rather than later. 😁

I've got some work to do but see that there are some questions for me, so I will try to circle back later this afternoon and address them.

github/copilot.go
gauraw10: I don't know, its right place to add this comment here, But may be we can have `GetCopilotSeatForUser` to get the info that user already onboarded before assigning the seat.

github/copilot_test.go
```diff
@@ -0,0 +1,649 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+// Test invalid JSON responses, vlaid responses are covered in the other tests
+func TestCopilotSeatDetails_UnmarshalJSON(t *testing.T) {
+	tests := map[string]struct {
+		data    string
+		want    *CopilotSeatDetails
+		wantErr bool
+	}{
```
WillAbides, line 18: This could be:

```go
tests := map[string]struct {
  name string
  data string
} {
  {
    name: "invalid JSON",
    data: `{`,
  },
...
}
  
```

Two things changed here:
- Make it a slice instead of a map for defined order when testing.
- Because this only tests invalid, the "want" and "wantErr" fields are always the same.

github/copilot_test.go
```diff
@@ -0,0 +1,649 @@
...
+			want:    &CopilotSeatDetails{},
+			wantErr: true,
+		},
+	}
+
+	for name, tc := range tests {
+		seatDetails := &CopilotSeatDetails{}
+
+		t.Run(name, func(t *testing.T) {
+			err := seatDetails.UnmarshalJSON([]byte(tc.data))
```
WillAbides, line 80: It's better to test with `json.Unmarshal` instead of directly calling `UnmarshalJSON` because that is how users will call it.

o-sama:
Apologies guys, clearly this isn't my best work. I really appreciate the review comments. I'll apply most suggestions and reply to the few that I think we should discuss more. 

@gauraw10 Similarly to @gmlewis, I tend to forget about pagination sometimes. I'll add support for that.

gmlewis:
Watching the conversations go by, I'm wondering if it might be nice to have 3 helper methods on the struct like `GetUser() (*User, bool)`, etc.

Thoughts?

o-sama:
@gmlewis Would that be while keeping `Assignee` as `interface{}`? I'm in favour of that since it allows easy checking/fetching with `if, ok`.

github/copilot.go
WillAbides: That's a good point about Team being different. I think this might be the best way to handle it after all.

gmlewis:
> @gmlewis Would that be while keeping `Assignee` as `interface{}`? I'm in favour of that since it allows easy checking/fetching with `if, ok`.

Yes, that's what I was thinking might be nice, but I'm open to ideas.

github/copilot_test.go
WillAbides: I'll defer to your judgement on want and wantErr. We may or may not need them in the future, and it doesn't hurt anything to keep them around.

o-sama:
Good call, I wanted to expand coverage for that 🙂 

What do you think about not returning errors for unmarshaling the Assignee field once we've done the type assertions? I'm not sure we would get errors there since we know the map matches the type of the struct we're unmarshaling into, and I can't think of a way to test for that.

WillAbides:
In general, I would much rather have untested error branches than trust our assumption that an error can't happen and end up with a panic. Even though codecov puts a big, ugly ❌ on your PR, it isn't a hard rule that test coverage can't decrease.

That said, I think it will be possible to cover those cases with just the right json data. Let me have a look and see if I can come up with something.

o-sama:
Ah good catch @WillAbides, it's always the commas 🙃

WillAbides:
I honed in on it because I've made that mistake so many times. It's hard remembering when going back and forth between go syntax and json.

o-sama:
Alrighty, the `type` field panic is fixed and the commas are gone. I don't think coverage has been updated though so can't be certain those cases are covered.

o-sama:
You got me to check whether VSCode has go coverage on the extension and yes it does 🙂. The `json.Marshal` error is the only uncovered line now. Also changed the invalid type test case to use `"type": []` which caused a panic that's now fixed.

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+	return nil
+}
+
+// GetUser gets the User from the CopilotSeatDetails if the assignee is a user.
+func (cp *CopilotSeatDetails) GetUser() (*User, bool) {
+	if user, ok := cp.Assignee.(*User); ok {
+		return user, true
+	}
+	return nil, false
+}
```
gmlewis, line 126:
```suggestion
func (cp *CopilotSeatDetails) GetUser() (*User, bool) { u, ok := cp.Assignee.(*User); return u, ok }
```

github/copilot.go
gauraw10: @o-sama I feel happy to help you on the paginations one and other stuff here, because I am developing the same for my own stuff.

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+
+// GetTeam gets the Team from the CopilotSeatDetails if the assignee is a team.
+func (cp *CopilotSeatDetails) GetTeam() (*Team, bool) {
+	if team, ok := cp.Assignee.(*Team); ok {
+		return team, true
+	}
+	return nil, false
+}
```
gmlewis, line 134:
```suggestion
func (cp *CopilotSeatDetails) GetTeam() (*Team, bool) { t, ok := cp.Assignee.(*Team); return t, ok }
```

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+
+// GetOrganization gets the Organization from the CopilotSeatDetails if the assignee is an organization.
+func (cp *CopilotSeatDetails) GetOrganization() (*Organization, bool) {
+	if organization, ok := cp.Assignee.(*Organization); ok {
+		return organization, true
+	}
+	return nil, false
+}
```
gmlewis, line 142:
```suggestion
func (cp *CopilotSeatDetails) GetOrganization() (*Organization, bool) { o, ok := cp.Assignee.(*Organization); return o, ok }
```

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+
+// SeatCancellations represents the number of seats cancelled.
+type SeatCancellations struct {
+	SeatsCancelled int `json:"seats_cancelled"`
+}
+
+func (cp *CopilotSeatDetails) UnmarshalJSON(data []byte) error {
+	type alias CopilotSeatDetails
+	var seatDetail alias
```
gmlewis, line 68: I'm not sure I understand the need for line 67 or why we can't have `var seatDetails CopilotSeatDetails`.
Can you please either add comments as to why these are necessary or just remove them if they are not needed?

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+	return copilotSeats, resp, nil
+}
+
+// AddCopilotTeams adds teams to the Copilot for Business subscription for an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-teams-to-the-copilot-for-business-subscription-for-an-organization
+func (s *CopilotService) AddCopilotTeams(ctx context.Context, org string, teamNames []string) (*SeatAssignments, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/selected_teams", org)
+
+	req, err := s.client.NewRequest("POST", u, teamNames)
```
gmlewis, line 192: This isn't going to work, and I'll show you why in the unit tests below...

OK, I can't stand the suspense.  :smile:  The problem is that this does not match the official documentation.
The POST request needs to look like this:

```
{"selected_teams":["engteam1","engteam2","engteam3"]}
```

but what you've written is instead going to look like this:

```
["engteam1","engteam2","engteam3"]
```

So what we need is a temporary struct, something like this:

```go
body := struct {
  SelectedTeams []string `json:"selected_teams"`
}{
  SelectedTeams: teamNames,
}
req, err := s.client.NewRequest("POST", u, body)
```

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+	return seatAssignments, resp, nil
+}
+
+// RemoveCopilotTeams removes teams from the Copilot for Business subscription for an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-teams-from-the-copilot-for-business-subscription-for-an-organization
+func (s *CopilotService) RemoveCopilotTeams(ctx context.Context, org string, teamNames []string) (*SeatCancellations, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/selected_teams", org)
+
+	req, err := s.client.NewRequest("DELETE", u, teamNames)
```
gmlewis, line 212: Same here.

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+	return seatCancellations, resp, nil
+}
+
+// AddCopilotUsers adds users to the Copilot for Business subscription for an organization
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-users-to-the-copilot-for-business-subscription-for-an-organization
+func (s *CopilotService) AddCopilotUsers(ctx context.Context, org string, userNames []string) (*SeatAssignments, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/selected_users", org)
+
+	req, err := s.client.NewRequest("POST", u, userNames)
```
gmlewis, line 232: Same here.

github/copilot.go
```diff
@@ -0,0 +1,284 @@
...
+	return seatAssignments, resp, nil
+}
+
+// RemoveCopilotUsers removes users from the Copilot for Business subscription for an organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-users-from-the-copilot-for-business-subscription-for-an-organization
+func (s *CopilotService) RemoveCopilotUsers(ctx context.Context, org string, userNames []string) (*SeatCancellations, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/copilot/billing/selected_users", org)
+
+	req, err := s.client.NewRequest("DELETE", u, userNames)
```
gmlewis, line 252: Same here.

github/copilot_test.go
```diff
@@ -0,0 +1,883 @@
...
+		return resp, err
+	})
+}
+
+func TestCopilotService_AddCopilotTeams(t *testing.T) {
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/orgs/o/copilot/billing/selected_teams", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
```
gmlewis, line 606: On all the POST methods, we need to parse the request body and make sure it looks like the official documentation says it should look like.  I know we haven't strictly followed that rule in this repo, although we have made attempts to improve this testing with the marshaling tests.

Anyway, if we had a test like that here, it would have caught the problem I referred to above.

We do have several examples of doing this, but basically, it boils down to adding something like this:

```go
testBody(t, r, `{"selected_teams":["team1","team2"]}` + "\n")
```

Please add `testBody` to all the tests that send a body.  Thank you!

gmlewis (changes requested):

github/copilot.go
o-sama: The `alias` type has the `CopilotSeatDetails` struct fields, but not its methods, this way when we call `json.Unmarshal` inside this function we don't have an infinite recursion. I'll add the comment and make the other requested changes tomorrow 🙂

github/copilot.go
o-sama: I thought so as well hence why I initially had a struct for this and for `selected_users`, but went with the review suggestion to just use a slice since I thought maybe I had overcomplicated things 😅

github/copilot_test.go
o-sama: Will do!

github/copilot.go
```diff
@@ -53,6 +53,16 @@ type CopilotSeatDetails struct {
...
 
+// SelectedTeams represents the teams selected for Copilot for Business.
+type SelectedTeams struct {
+	SelectedTeams []string `json:"selected_teams"`
+}
+
+// SelectedUsers represents the users selected for Copilot for Business.
+type SelectedUsers struct {
+	SelectedUsers []string `json:"selected_users"`
+}
```
gmlewis, line 64: Sorry, that's not what I meant.

The original feedback from @WillAbides was a good one - we shouldn't create these extra structs just for the requests.

Instead, they should be anonymous (I said "temporary" which is my fault) structs _WITHIN_ the functions where they are being used.  We use this trick many other places in this repo.

See my earlier example:

```go
body := struct {
  SelectedTeams []string `json:"selected_teams"`
}{
  SelectedTeams: teamNames,
}
req, err := s.client.NewRequest("POST", u, body)
```

github/copilot.go
```diff
@@ -186,7 +185,7 @@ func (s *CopilotService) ListCopilotSeats(ctx context.Context, org string, opts
 // AddCopilotTeams adds teams to the Copilot for Business subscription for an organization.
 //
 // GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#add-teams-to-the-copilot-for-business-subscription-for-an-organization
-func (s *CopilotService) AddCopilotTeams(ctx context.Context, org string, teamNames []string) (*SeatAssignments, *Response, error) {
+func (s *CopilotService) AddCopilotTeams(ctx context.Context, org string, teamNames *SelectedTeams) (*SeatAssignments, *Response, error) {
```
gmlewis, line 188: Sorry, that's not what I meant... this should still be `teamNames []string` and then an anonymous struct should be used _within_ this method to create the _actual_ struct being sent.  See above.

github/copilot.go
```diff
@@ -206,7 +205,7 @@ func (s *CopilotService) AddCopilotTeams(ctx context.Context, org string, teamNa
 // RemoveCopilotTeams removes teams from the Copilot for Business subscription for an organization.
 //
 // GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-teams-from-the-copilot-for-business-subscription-for-an-organization
-func (s *CopilotService) RemoveCopilotTeams(ctx context.Context, org string, teamNames []string) (*SeatCancellations, *Response, error) {
+func (s *CopilotService) RemoveCopilotTeams(ctx context.Context, org string, teamNames *SelectedTeams) (*SeatCancellations, *Response, error) {
```
gmlewis, line 208: Same here: `teamNames []string`

github/copilot.go
```diff
@@ -246,10 +245,10 @@ func (s *CopilotService) AddCopilotUsers(ctx context.Context, org string, userNa
 // RemoveCopilotUsers removes users from the Copilot for Business subscription for an organization.
 //
 // GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-for-business#remove-users-from-the-copilot-for-business-subscription-for-an-organization
-func (s *CopilotService) RemoveCopilotUsers(ctx context.Context, org string, userNames []string) (*SeatCancellations, *Response, error) {
+func (s *CopilotService) RemoveCopilotUsers(ctx context.Context, org string, users *SelectedUsers) (*SeatCancellations, *Response, error) {
```
gmlewis, line 248: Same here: `userNames []string`

gmlewis (changes requested):

github/copilot.go
gmlewis: No, I think you were correct, @WillAbides - we should be passing in a slice of strings, and let the endpoint method handle the formatting of the call internally.

o-sama:
Sorry for the delay in this, I've been abroad, I'll try to put out the last bit of changes within the next couple of days.

gmlewis (approved):
Hey @o-sama - I think this is looking great!  Thank you!
LGTM.

o-sama:
@gmlewis My pleasure. Again, apologies for the delays, between being sick and going abroad for my parents to meet their granddaughter I've not been very timely with this. Thanks for putting up with me!

gmlewis:
> @gmlewis My pleasure. Again, apologies for the delays, between being sick and going abroad for my parents to meet their granddaughter I've not been very timely with this. Thanks for putting up with me!

No worries at all, @o-sama ! Fortunately, we are not on any restrictive timelines or deadlines.  😄

gauraw10 (changes requested):
Everything looks good for me except a small change about the naming convention for the file` github-accessors.go`, I think it would be ` github_accessors.go`, what do you think @o-sama @gmlewis

gmlewis:
> Everything looks good for me except a small change about the naming convention for the file` github-accessors.go`, I think it would be ` github_accessors.go`, what do you think @o-sama @gmlewis

Thanks, @gauraw10, but there's no need to change the filename of `github-accessors.go`.

gauraw10:
I found other file name in same directory have naming conversations as `Snake case `, So was thinking to have the same format.

gmlewis:
> I found other file name in same directory have naming conventions as `Snake case `, So was thinking to have the same format.

I understand, but we will not be changing the name.  `github-accessors.go` is an auto-generated file and does not need to match the case of the other files.

zetaab:
@gmlewis I know that second lgtm is needed. However, is there somekind of way to find that another person who could give that LGTM? This has been waiting for sometime and at least we would like to use this feature.

gmlewis:
> @gmlewis I know that second lgtm is needed. However, is there somekind of way to find that another person who could give that LGTM? This has been waiting for sometime and at least we would like to use this feature.

OK, first let me resolve the merge conflicts, @zetaab - then maybe @valbeat has time to take a look at this one.

gmlewis:
@zetaab - you yourself could move this PR along by performing a review and giving it an LGTM+Approval.

github/copilot.go
```diff
@@ -0,0 +1,311 @@
...
+	TotalSeats int64                 `json:"total_seats"`
+	Seats      []*CopilotSeatDetails `json:"seats"`
+}
+
+// CopilotSeatDetails represents the details of a Copilot for Business seat.
+type CopilotSeatDetails struct {
+	// Assignee can either be a User, Team, or Organization.
+	Assignee                interface{} `json:"assignee"`
+	AssigningTeam           *Team       `json:"assigning_team,omitempty"`
+	PendingCancellationDate *string     `json:"pending_cancellation_date,omitempty"`
```
zetaab, line 49: type should be `*Timestamp` instead that it works in similar way than other time fields.

github/copilot.go
zetaab: pagination works in current version

zetaab (changes requested):
I could not find anything else than that one type change. I am currently coding against this PR and it was little bit weird that the type is not similar than other timestamps.

zetaab (approved):
actually that PendingCancellationDate field is `date` so perhaps leave it like this.

LGTM

gmlewis:
Thank you, @zetaab !
Merging.

PR #2975, opened 2023-10-25
Add referenced workflows to WorkflowRun

### LINKED ISSUE #2974: Support for referenced workflows in WorkflowRun
Hi team, please add referenced_workflows field to the WorkflowRun, see documentation https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28 

```
GET /repos/{owner}/{repo}/actions/runs/{run_id}

{
  "title": "Workflow Run",
  "description": "An invocation of a workflow",
  "type": "object",
  "properties": {
...
    "referenced_workflows": {
      "type": [
        "array",
        "null"
      ],
      "items": {
        "title": "Referenced workflow",
        "description": "A workflow referenced/reused by the initial caller workflow",
        "type": "object",
        "properties": {
          "path": {
            "type": "string"
          },
          "sha": {
            "type": "string"
          },
          "ref": {
            "type": "string"
          }
        },
        "required": [
          "path",
          "sha"
        ]
      }
    }
}
```

schaffe-cb:
Closes #2974

### DIFF
--- a/github/actions_workflow_runs.go
+++ b/github/actions_workflow_runs.go
@@ -14,39 +14,40 @@ import (
 
 // WorkflowRun represents a repository action workflow run.
 type WorkflowRun struct {
-	ID                 *int64         `json:"id,omitempty"`
-	Name               *string        `json:"name,omitempty"`
-	NodeID             *string        `json:"node_id,omitempty"`
-	HeadBranch         *string        `json:"head_branch,omitempty"`
-	HeadSHA            *string        `json:"head_sha,omitempty"`
-	RunNumber          *int           `json:"run_number,omitempty"`
-	RunAttempt         *int           `json:"run_attempt,omitempty"`
-	Event              *string        `json:"event,omitempty"`
-	DisplayTitle       *string        `json:"display_title,omitempty"`
-	Status             *string        `json:"status,omitempty"`
-	Conclusion         *string        `json:"conclusion,omitempty"`
-	WorkflowID         *int64         `json:"workflow_id,omitempty"`
-	CheckSuiteID       *int64         `json:"check_suite_id,omitempty"`
-	CheckSuiteNodeID   *string        `json:"check_suite_node_id,omitempty"`
-	URL                *string        `json:"url,omitempty"`
-	HTMLURL            *string        `json:"html_url,omitempty"`
-	PullRequests       []*PullRequest `json:"pull_requests,omitempty"`
-	CreatedAt          *Timestamp     `json:"created_at,omitempty"`
-	UpdatedAt          *Timestamp     `json:"updated_at,omitempty"`
-	RunStartedAt       *Timestamp     `json:"run_started_at,omitempty"`
-	JobsURL            *string        `json:"jobs_url,omitempty"`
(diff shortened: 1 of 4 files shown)


github/actions_workflow_runs.go
```diff
@@ -104,6 +105,12 @@ type PendingDeploymentsRequest struct {
 	Comment string `json:"comment"`
 }
 
+type ReferencedWorkflow struct {
+	Path *string `json:"path,omitempty"`
+	SHA  *string `json:"sha,omitempty"`
```
WillAbides, line 110: path and sha are required in the response schema, so they can be `string` instead of `*string`

WillAbides (changes requested):
Looks good except for changing the types to non-pointer.

github/actions_workflow_runs.go
gmlewis: Just FYI - overall, we've been very lax on the "receive-side" of responses from the GitHub server over the years in terms of what the official docs say will definitely be returned versus what optionally might be returned.

One of the reasons we've done this is first, it has not always been clear from the documentation what will or will not be sent back, and we've also discovered experientially that there are sometimes surprises depending on where a particular struct is embedded (for example, if we were to add this to a different method's return struct).

Additionally, years back we added the helper methods to get field values that are pointers which help tremendously to avoid panics if something was expected but didn't show up in the response.

So I believe our official stance is to try and be super careful about the `omitempty` on the _request_ side of the usage, but on the _response_ side, we are happy to err on the side of more fields with pointers (rather than fewer).

So if @WillAbides had not mentioned this, I would have allowed this to go through, however I'm fine also with making the change since the official docs are hopefully correct and will not change soon.

WillAbides: Thanks. That's good to read. I didn't realize the distinction between request and response fields.

I'm going to change my review to ✅ as soon as I hit submit on this comment. Then I might make a PR adding a note about this to CONTRIBUTING.md.

WillAbides (approved):

github/actions_workflow_runs.go
gmlewis: So I wrote "official stance" above and that was probably too strongly worded 😄 .

The above is my "unofficial understanding" of how we are guiding the maintenance of this repo based upon years of observation.

The final authority is the original author of this repo, Will Norris, a friend and former coworker of mine, who asked me many years ago to help maintain this repo. I'm sure he listens in, as I hear from him every now and then, but I hate to ping him unless absolutely necessary.  😄

gmlewis:
OK, so if we are leaving the structs as is, which is fine with me, then you will need to run `go generate ./...` and push the changes to this PR, @schaffe-cb - when you get a chance.  Thanks.

github/actions_workflow_runs.go
schaffe-cb: @gmlewis thanks for the walkthrough in the past :) I'm happy to make any changes necessary. The reason I made them pointers is just to follow a patten with other required fields in `WorkflowRun`.

gmlewis (approved):
Thank you, @schaffe-cb and @WillAbides !
LGTM.
Merging.

PR #2977, opened 2023-10-31
Fix branch protection request fields

### LINKED ISSUE #2976: Issue with Branch Protection fields
We found an issue with `Branch Protection fields` that were added as part of https://github.com/google/go-github/commit/4caa1d6d12a652a6656bb72c1985393bfe6203d9

```
--
  | For 'properties/checks', nil is not an array.
  | Not all subschemas of "allOf" matched.
```

The problem is with 
```
Checks []*RequiredStatusCheck `json:"checks"`
```  
it needs the `omitempty` tag added to it.

A fix is on the way

viqueen:
Fixes #2976 

The property `Checks` on `RequiredStatusChecks` type is not required according to GitHub api, so using the `omitempty` tag

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -1055,7 +1055,7 @@ type RequiredStatusChecks struct {
 	Contexts []string `json:"contexts,omitempty"`
 	// The list of status checks to require in order to merge into this
 	// branch.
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @viqueen !
LGTM.
Merging.

Issue #2965, opened 2023-10-16
Support for Repository Custom Properties


liaodaniel:
GitHub has recently released https://github.blog/changelog/2023-10-12-github-repository-custom-properties-beta/. It would be nice to have support for this feature. With the feature in beta,  we may expect API changes over time but I'll be happy to get a draft PR for now.

References:
* [Properties API docs](https://docs.github.com/en/rest/orgs/properties?apiVersion=2022-11-28)
* [Repository Custom Properties Beta Feedback](https://github.com/orgs/community/discussions/69914)

gmlewis:
Thank you, @liaodaniel !
It's yours.

PR #2986, opened 2023-11-04
Implement Custom Properties

liaodaniel:
Resolves https://github.com/google/go-github/issues/2965

### DIFF
--- a/github/orgs_properties.go
+++ b/github/orgs_properties.go
@@ -37,7 +37,9 @@ type CustomPropertyValue struct {
 
 // GetAllCustomProperties gets all custom properties that is defined for the specified organization.
 //
-// GitHub API docs: https://docs.github.com/en/rest/orgs/properties#get-all-custom-properties-for-an-organization
+// GitHub API docs: https://docs.github.com/rest/orgs/properties#get-all-custom-properties-for-an-organization
+//
+//meta:operation GET /orgs/{org}/properties/schema
 func (s *OrganizationsService) GetAllCustomProperties(ctx context.Context, org string) ([]*CustomProperty, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/properties/schema", org)
 
@@ -57,7 +59,9 @@ func (s *OrganizationsService) GetAllCustomProperties(ctx context.Context, org s
 
 // CreateOrUpdateCustomProperties creates new or updates existing custom properties that are defined for the specified organization.
 //
-// GitHub API docs: https://docs.github.com/en/rest/orgs/properties#create-or-update-custom-properties-for-an-organization
+// GitHub API docs: https://docs.github.com/rest/orgs/properties#create-or-update-custom-properties-for-an-organization
+//
+//meta:operation PATCH /orgs/{org}/properties/schema
 func (s *OrganizationsService) CreateOrUpdateCustomProperties(ctx context.Context, org string, properties []*CustomProperty) ([]*CustomProperty, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/properties/schema", org)
 
@@ -83,7 +87,9 @@ func (s *OrganizationsService) CreateOrUpdateCustomProperties(ctx context.Contex
 
 // GetCustomProperty gets a custom property that is defined for the specified organization.
 //
-// GitHub API docs: https://docs.github.com/en/rest/orgs/properties#get-a-custom-property-for-an-organization
+// GitHub API docs: https://docs.github.com/rest/orgs/properties#get-a-custom-property-for-an-organization
+//
+//meta:operation GET /orgs/{org}/properties/schema/{custom_property_name}
 func (s *OrganizationsService) GetCustomProperty(ctx context.Context, org, name string) (*CustomProperty, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/properties/schema/%v", org, name)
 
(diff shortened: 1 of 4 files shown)


github/orgs_properties.go
```diff
@@ -0,0 +1,196 @@
...
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// CustomProperty represents an organization custom property object.
+type CustomProperty struct {
+	PropertyName *string `json:"property_name,omitempty"`
```
gmlewis, line 15: This field is required, so let's please make it mandatory.

```suggestion
	PropertyName string `json:"property_name"`
```

github/orgs_properties.go
```diff
@@ -0,0 +1,196 @@
...
+	RepositoryID       int64                  `json:"repository_id"`
+	RepositoryName     string                 `json:"repository_name"`
+	RepositoryFullName string                 `json:"repository_full_name"`
+	Properties         []*CustomPropertyValue `json:"properties"`
+}
+
+// CustomPropertyValue represents a custom property value.
+type CustomPropertyValue struct {
+	PropertyName string `json:"property_name"`
+	Value        string `json:"value"`
```
gmlewis, line 35: Looking at the [official documentation](https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28#list-custom-property-values-for-organization-repositories), it looks like `value` can be `null`, so let's please make this field nullable.

The "Response schema" tab says this:

```json
      "properties": {
        "type": "array",
        "items": {
          "title": "Custom Property Value",
          "description": "Custom property name and associated value",
          "type": "object",
          "properties": {
            "property_name": {
              "type": "string",
              "description": "The name of the property"
            },
            "value": {
              "type": [
                "string",
                "null"
              ],
              "description": "The value assigned to the property"
            }
          },
          "required": [
            "property_name",
            "value"
          ]
        },
```

```suggestion
	Value        *string `json:"value,omitempty"`
```

github/orgs_properties.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// GetAllCustomProperties gets all custom properties that is defined for the specified organization.
```
gmlewis, line 38:
```suggestion
// GetAllCustomProperties gets all custom properties that are defined for the specified organization.
```

github/orgs_properties.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return customProperty, resp, nil
+}
+
+// CreateOrUpdateCustomProperty creates a new or updates an existing custom property that is defined for the specified organization.
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/properties#create-or-update-a-custom-property-for-an-organization
+//
+//meta:operation PUT /orgs/{org}/properties/schema/{custom_property_name}
+func (s *OrganizationsService) CreateOrUpdateCustomProperty(ctx context.Context, org, name string, property *CustomProperty) (*CustomProperty, *Response, error) {
```
gmlewis, line 115: Can we please change `name` here to `customPropertyName` just to make it super-clear?
I only ask because it is so common in this repo to have something like `org, repo string` that I know I would personally probably fill in the repo name here by mistake unless I see that glaringly obvious parameter name.

```suggestion
func (s *OrganizationsService) CreateOrUpdateCustomProperty(ctx context.Context, org, customPropertyName string, property *CustomProperty) (*CustomProperty, *Response, error) {
```

github/orgs_properties.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return customProperty, resp, nil
+}
+
+// RemoveCustomProperty removes a custom property that is defined for the specified organization.
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/properties#remove-a-custom-property-for-an-organization
+//
+//meta:operation DELETE /orgs/{org}/properties/schema/{custom_property_name}
+func (s *OrganizationsService) RemoveCustomProperty(ctx context.Context, org, name string) (*Response, error) {
```
gmlewis, line 137: Same here, please?

```suggestion
func (s *OrganizationsService) RemoveCustomProperty(ctx context.Context, org, customPropertyName string) (*Response, error) {
```

gmlewis (changes requested):
Thank you, @liaodaniel!
Just a few minor tweaks, please, and then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/orgs_properties.go
liaodaniel: This was a conscious change due to https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28#create-or-update-a-custom-property-for-an-organization. This API relies on `property_name` to be specified in the path instead of the field to be in the custom property object.

liaodaniel: Hmm, just to clarify. Would it be fine to update it to send the unused field `property_name` as part of the body?

gmlewis: Ah, I see what you are talking about.

How about just add a comment to this field... something like:
```go
// Although PropertyName is required for most endpoints, this struct is shared with the PUT endpoint
// where it is sent in the path, and is therefore a pointer with "omitempty".
```

liaodaniel: Thank you @gmlewis. I'll update it with something along those lines.

gmlewis (approved):
Thank you, @liaodaniel !
LGTM.

Also, can you please switch this PR out of "Draft" mode when you are ready?

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

PR #2988, opened 2023-11-04
Deprecate RepositoriesService.List

### LINKED ISSUE #2987: Unclear godoc for RepositoriesService.List
`RepositoriesService.List` is a bit different from most other go-github methods because it calls one of two endpoints depending on the options. When `user == ""` it calls `GET /user/repos` which returns all of the authenticated user's repos. When `user != ""` it calls `GET /user/{username}/repos` which returns only the public repos of the specified user regardless of the caller's permissions.

The current godoc[^1] alludes to this by listing multiple github doc links, but it is still unclear. The options struct[^2] adds to the confusion with the "affiliation" and "visibility" query parameters that are only available on `/user/repos`. It also lists "public" and "private" as options for the "type" parameter. Those are only available for `/user/repos` as well.

See #2980 for an example of this causing confusion.

Without looking at the history, I suspect these two endpoints used to be more similar and github added more options over time that made them diverge. So it made sense to combine these into one method, but if these were new endpoints now, they wouldn't have been combined into one endpoint.

I propose that we create two new methods: `RepositoriesService.ListForUser` and `RepositoriesService.ListForAuthenticatedUser` that makes this split explicit and update `RepositoriesService.List`'s godoc to make the distinction clear. Maybe we should deprecate `RepositoriesService.List` as well.

[^1]: https://github.com/google/go-github/blob/f643c0f3fab9235ece313e9b935ae9136d2bb5ab/github/repos.go#L265-L273

[^2]: https://github.com/google/go-github/blob/f643c0f3fab9235ece313e9b935ae9136d2bb5ab/github/repos.go#L173-L206

WillAbides:
Closes #2987

This deprecates `RepositoriesService.List` and replaces it with two new methods: `RepositoriesService.ListByUser` and `RepositoriesService.ListByAuthenticatedUser`. See #2987 for the justification.

### DIFF
--- a/github/examples_test.go
+++ b/github/examples_test.go
@@ -49,14 +49,14 @@ func ExampleRepositoriesService_GetReadme() {
 	fmt.Printf("google/go-github README:\n%v\n", content)
 }
 
-func ExampleRepositoriesService_List() {
+func ExampleRepositoriesService_ListByUser() {
 	client := github.NewClient(nil)
 
 	user := "willnorris"
-	opt := &github.RepositoryListOptions{Type: "owner", Sort: "updated", Direction: "desc"}
+	opt := &github.RepositoryListByUserOptions{Type: "owner", Sort: "updated", Direction: "desc"}
 
 	ctx := context.Background()
-	repos, _, err := client.Repositories.List(ctx, user, opt)
+	repos, _, err := client.Repositories.ListByUser(ctx, user, opt)
 	if err != nil {
 		fmt.Println(err)
 	}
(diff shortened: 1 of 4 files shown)


WillAbides:
To address [@gmlewis's comment](https://github.com/google/go-github/issues/2987#issuecomment-1793512970) from the issue, I think the context of having a user parameter is enough to make it clear what `ListByUser` does.

Also, in the issue I proposed List**For**User and List**For**AuthenticatedUser, but I changed it to List**By**User and List**By**AuthenticatedUser because I found precedents "RepositoryService.ListByOrg" and "IssuesService.ListByRepo".

gmlewis (approved):
Thank you, @WillAbides !
LGTM.

valbeat (approved):

PR #3002, opened 2023-11-24
Escape package names to support names which include a slash

bn4t:
Package names aren't escaped before being added to the url.
E.g. `orgs/foo/packages/container/foo/bar` instead of `orgs/foo/packages/container/foo%2Fbar`

This causes 404 errors for such packages.

### DIFF
--- a/github/orgs_packages.go
+++ b/github/orgs_packages.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"net/url"
 )
 
 // ListPackages lists the packages for an organization.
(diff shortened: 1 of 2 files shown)


gmlewis (changes requested):
Thank you, @bn4t !

For each of the modified methods, could you please add a godoc comment (since this is a change of behavior) and also a unit test that demonstrates the new path escaping?

For example:

```go
// Note that packageName is escaped for the URL path so that you don't need to.
```

bn4t:
@gmlewis Would it be fine if I just changed the existing tests to use a package name with a slash or do cou want me to write one/multiple new tests for this?

gmlewis:
> @gmlewis Would it be fine if I just changed the existing tests to use a package name with a slash or do you want me to write one/multiple new tests for this?

I think it is totally fine to modify the existing tests.  Thanks.

bn4t:
@gmlewis Okay, pushed it. Thanks for handholding me through this, this really shouldn't have taken so long. My fault for not checking the contributing guidelines in detail.

gmlewis:
> @gmlewis Okay, pushed it. Thanks for handholding me through this, this really shouldn't have taken so long. My fault for not checking the contributing guidelines in detail.

No problem, we do have a lot of automation in this repo that is not immediately obvious.

gmlewis (approved):
Thank you, @bn4t !
LGTM.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

PR #3011, opened 2023-12-01
Don't update httpClient passed to NewClient

WillAbides:
Users can accidentally leak credentials by doing something like this:

```go
pat := os.Getenv("GITHUB_TOKEN")
httpClient := &http.Client{
	Timeout: time.Second * 30,
}
ghClient := github.NewClient(httpClient).WithAuthToken(pat)

// ... do something with ghClient

httpClient.Get("https://some-sketchy-site")
```

This PR fixes that by updating `NewClient` to make a copy of httpClient so that when its transport is updated by `WithAuthToken`, the original httpClient won't have its transport updated as well.

It also updates `TestWithAuthToken` to check for leaks

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -220,6 +220,8 @@ type service struct {
 }
 
 // Client returns the http.Client used by this GitHub client.
+// This should only be used for requests to the GitHub API because
+// request headers will contain an authorization token.
 func (c *Client) Client() *http.Client {
 	c.clientMu.Lock()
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @WillAbides !
LGTM.

valbeat (approved):
LGTM!

gmlewis:
Thank you, @valbeat !
Merging.

PR #3049, opened 2024-01-12
Add suspended as option to AdminService.CreateUser()

ttomsu:
`suspended` value can be found here: https://docs.github.com/en/enterprise-server@3.11/rest/enterprise-admin/users?apiVersion=2022-11-28#create-a-user

### DIFF
--- a/github/admin_users.go
+++ b/github/admin_users.go
@@ -13,8 +13,9 @@ import (
 // createUserRequest is a subset of User and is used internally
 // by CreateUser to pass only the known fields for the endpoint.
 type createUserRequest struct {
-	Login *string `json:"login,omitempty"`
-	Email *string `json:"email,omitempty"`
+	Login     *string `json:"login,omitempty"`
+	Email     *string `json:"email,omitempty"`
+	Suspended *bool   `json:"suspended,omitempty"`
 }
 
 // CreateUser creates a new user in GitHub Enterprise.
@@ -22,12 +23,13 @@ type createUserRequest struct {
 // GitHub API docs: https://docs.github.com/enterprise-server@3.11/rest/enterprise-admin/users#create-a-user
 //
 //meta:operation POST /admin/users
-func (s *AdminService) CreateUser(ctx context.Context, login, email string) (*User, *Response, error) {
+func (s *AdminService) CreateUser(ctx context.Context, login, email string, suspended bool) (*User, *Response, error) {
 	u := "admin/users"
 
 	userReq := &createUserRequest{
-		Login: &login,
-		Email: &email,
+		Login:     &login,
+		Email:     &email,
+		Suspended: &suspended,
 	}
 
 	req, err := s.client.NewRequest("POST", u, userReq)
--- a/github/admin_users_test.go
+++ b/github/admin_users_test.go
@@ -25,7 +25,7 @@ func TestAdminUsers_Create(t *testing.T) {
 		assertNilError(t, json.NewDecoder(r.Body).Decode(v))
 
 		testMethod(t, r, "POST")
-		want := &createUserRequest{Login: String("github"), Email: String("email@domain.com")}
+		want := &createUserRequest{Login: String("github"), Email: String("email@domain.com"), Suspended: Bool(false)}
 		if !cmp.Equal(v, want) {
 			t.Errorf("Request body = %+v, want %+v", v, want)
 		}
@@ -34,7 +34,7 @@ func TestAdminUsers_Create(t *testing.T) {
 	})
 
 	ctx := context.Background()
-	org, _, err := client.Admin.CreateUser(ctx, "github", "email@domain.com")
+	org, _, err := client.Admin.CreateUser(ctx, "github", "email@domain.com", false)
 	if err != nil {
 		t.Errorf("Admin.CreateUser returned error: %v", err)
 	}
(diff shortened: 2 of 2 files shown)


github/admin_users.go
```diff
@@ -13,21 +13,23 @@ import (
...
+	Suspended *bool   `json:"suspended,omitempty"`
 }
 
 // CreateUser creates a new user in GitHub Enterprise.
 //
 // GitHub API docs: https://docs.github.com/enterprise-server@3.11/rest/enterprise-admin/users#create-a-user
 //
 //meta:operation POST /admin/users
-func (s *AdminService) CreateUser(ctx context.Context, login, email string) (*User, *Response, error) {
+func (s *AdminService) CreateUser(ctx context.Context, login, email string, suspended bool) (*User, *Response, error) {
```
gmlewis, line 26: From the official docs: https://docs.github.com/en/enterprise-server@3.11/rest/enterprise-admin/users?apiVersion=2022-11-28#create-a-user
I'm noticing that both the `email` and `suspended` fields are optional.

Since we are breaking the API with this PR, I'm wondering if it might make more sense to create a new `CreateUserOptions` struct containing these optional fields?  In addition, `login` is required, so maybe change line 16 above to `Login string `json:"login"`.

In other words, we would still use the private `createUserRequest` but we would copy fields in from `opts` where the new signature would be:

```go
unc (s *AdminService) CreateUser(ctx context.Context, login string, opts *CreateUserOptions) (*User, *Response, error) {
```

If we do this now, then we might have a bit more flexibility in the future to add new fields to the `CreateUserOptions` struct without breaking the API again.

Thoughts?

ttomsu: Hi Glenn - I like the idea to break it only once if we have to break it at all.

Taking a look at other examples around the codebase, it looks like one pattern is to pass URL path segments individually, then the whole object that would end up in the request body. Examples

* [TeamsService.CreateTeam(ctx context.Context, org string, team NewTeam)](https://github.com/google/go-github/blob/e9f52699f5e55a4791567bce668f56446e6343ef/github/teams.go#L186)
* [PullRequestService.Create(ctx context.Context, owner string, repo string, pull *NewPullRequest)](https://github.com/google/go-github/blob/e9f52699f5e55a4791567bce668f56446e6343ef/github/pulls.go#L268)

Following this pattern, it seems a more consistent option would be to export the `createUserRequest` object and pass that as the second param (after `ctx`). 

WDYT?

gmlewis: Ah, yes, I see what you are saying.  Path query parameters should be listed as separate arguments, and then you can have another parameter:

```go
func (s *AdminService) CreateUser(ctx context.Context, req CreateUserRequest) (*User, *Response, error) {
```

where:

```go
// CreateUserRequest represents the fields sent to the `CreateUser` endpoint.
// Note that `Login` is a required field.
type CreateUserRequest struct {
	Login     string `json:"login"`
	Email     *string `json:"email,omitempty"`
	Suspended *bool   `json:"suspended,omitempty"`
}
```

gmlewis:
In the future, please don't use force-push in this repo, as we always squash-and-merge at the end.  See CONTRIBUTING.md for more details.  Thanks.

github/admin_users.go
```diff
@@ -10,26 +10,25 @@ import (
...
 
-// createUserRequest is a subset of User and is used internally
-// by CreateUser to pass only the known fields for the endpoint.
-type createUserRequest struct {
-	Login *string `json:"login,omitempty"`
-	Email *string `json:"email,omitempty"`
+// CreateUserRequest represents the fields sent to the `CreateUser` endpoint.
+// Note that `Login` is a required field.
+type CreateUserRequest struct {
+	Login     *string `json:"login"`
```
gmlewis, line 16:
```suggestion
	Login     string `json:"login"`
```

github/admin_users.go
```diff
@@ -10,26 +10,25 @@ import (
...
+	Email     *string `json:"email,omitempty"`
+	Suspended *bool   `json:"suspended,omitempty"`
+}
+
+type CreateUserOptions struct {
```
gmlewis, line 21: Please delete.

github/admin_users.go
```diff
@@ -10,26 +10,25 @@ import (
...
 }
 
 // CreateUser creates a new user in GitHub Enterprise.
 //
 // GitHub API docs: https://docs.github.com/enterprise-server@3.11/rest/enterprise-admin/users#create-a-user
 //
 //meta:operation POST /admin/users
-func (s *AdminService) CreateUser(ctx context.Context, login, email string) (*User, *Response, error) {
+func (s *AdminService) CreateUser(ctx context.Context, userReq *CreateUserRequest) (*User, *Response, error) {
```
gmlewis, line 29: Since this is required information, let's please pass by value, not by reference:

```suggestion
func (s *AdminService) CreateUser(ctx context.Context, userReq CreateUserRequest) (*User, *Response, error) {
```

gmlewis (changes requested):

github/admin_users.go
ttomsu: Done.

github/admin_users.go
ttomsu: Done.

github/admin_users.go
ttomsu: Done.

ttomsu:
Apologies for the force commit, my local branch got out of whack and I didn't realize it.

ttomsu:
@gmlewis Anyone you can ping to get a second set of eyes on?

gmlewis:
All contributors to this repo are volunteers and I typically try not to ping individuals unless there is an urgent need.

fchimpan:
Hi, @ttomsu !
Thanks your implementation. LGTM! Great job!

> If we do this now, then we might have a bit more flexibility in the future to add new fields to the CreateUserOptions struct without breaking the API again.

I agree it. It's a good design.

gmlewis:
Thank you, @fchimpan !
Merging.

Issue #3056, opened 2024-01-24
Add Topics to EditChange struct


riaje:
Hi,

I want to add the missing `Topics` field for the [EditChange](https://github.com/google/go-github/blob/master/github/event_types.go#L390) `struct`.

In my case, it will allow me to retrieve the previous topics when I receive the [repository edited event](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#repository).

I already have the branch ready for review.

Thanks

gmlewis:
Sounds great, thanks, @riaje !
Our CONTRIBUTING.md guide has helpful information in it for submitting PRs if you are new to this repo.
Specifically, step 4 seems to be a commonly-missed step.

riaje:
Already read :+1: thanks for the reminder of course !
The PR is submit, it's my first contribution to an open project on GitHub so I discover the process in the same time :laughing: 
Thanks

PR #3057, opened 2024-01-24
Add Topics to EditChange struct

riaje:
Fixes: https://github.com/google/go-github/issues/3056

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -441,7 +441,7 @@ type RepoName struct {
 
 // EditTopics represents a change of repository topics.
 type EditTopics struct {
-	From *[]string `json:"from,omitempty"`
+	From []string `json:"from,omitempty"`
 }
 
 // EditSHA represents a sha change of a pull-request.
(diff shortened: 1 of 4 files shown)


github/event_types.go
```diff
@@ -438,6 +439,11 @@ type RepoName struct {
 	From *string `json:"from,omitempty"`
 }
 
+// EditTopics represents a change of repository topics.
+type EditTopics struct {
+	From *[]string `json:"from,omitempty"`
```
gmlewis, line 444: Rarely do we actually need a pointer to a slice since a slice is already a reference type, and here we don't need one (and they are a pain in the rear to use), so let's please remove it:

```suggestion
	From []string `json:"from,omitempty"`
```

gmlewis (changes requested):
Thank you, @riaje !
Just one minor tweak, please, then we should be ready to merge.

github/event_types.go
riaje: Thanks for this suggestion, interesting reason I take in note for the future :+1:

riaje:
I will fix the tests, sorry I have made my commit a bit too faster

gmlewis (approved):
Thank you, @riaje !
LGTM.
Merging after tests pass.

PR #3090, opened 2024-03-04
Fix application of pagination options on ListCopilotSeats endpoint

ryanskidmore:
The pagination options were previously being applied as the Request body, instead of applied to the URL as in other `List` endpoints. This PR ensures that these options are correctly applied to the URL and introduces additional parameters in the test case to ensure this.

### DIFF
--- a/github/copilot_test.go
+++ b/github/copilot_test.go
@@ -336,6 +336,10 @@ func TestCopilotService_ListCopilotSeats(t *testing.T) {
 
 	mux.HandleFunc("/orgs/o/copilot/billing/seats", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
+		testFormValues(t, r, values{
+			"per_page": "100",
+			"page":     "1",
+		})
 		fmt.Fprint(w, `{
 				"total_seats": 4,
 				"seats": [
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @ryanskidmore !
LGTM.

gmlewis (changes requested):
Ah, whups - my bad, codecov reminded me that we need some more tests.
I'll point you to an example.  One moment...

github/copilot_test.go
```diff
@@ -473,7 +477,7 @@ func TestCopilotService_ListCopilotSeats(t *testing.T) {
 	lastActivityAt2 := Timestamp{tmp}
 
 	ctx := context.Background()
-	got, _, err := client.Copilot.ListCopilotSeats(ctx, "o", nil)
+	got, _, err := client.Copilot.ListCopilotSeats(ctx, "o", &ListOptions{Page: 1, PerPage: 100})
```
gmlewis, line 480: In this test, we will also want to add:

```go
	opts := &ListOptions{Page: 1, PerPage: 100}
	got, _, err := client.Copilot.ListCopilotSeats(ctx, "o", opts)
...
	const methodName = "ListCopilotSeats"
		_, _, err = client.Copilot.ListCopilotSeats(ctx, "\n", opts)
		return err
	})

		got, resp, err := client.Copilot.ListCopilotSeats(ctx, "o", opts)
		if got != nil {
			t.Errorf("Copilot.ListCopilotSeats returned %+v, want nil", got)
		}
		return resp, err
	})
```

ryanskidmore: Added these in - ty for the heads up. I don't think I can trigger the codecov workflow so I'm not sure if this has increased the coverage by enough!

ryanskidmore:
> Thank you, @ryanskidmore ! LGTM.
> 

Hey @gmlewis just wanted to follow up on this. I noticed codecov is still failing, do I need to add any more tests to this?

gmlewis:
> Hey @gmlewis just wanted to follow up on this. I noticed codecov is still failing, do I need to add any more tests to this?

No, codecov has been pretty unreliable lately. You can ignore it in this case.

be0x74a (approved):
LGMT! 🚀

gmlewis:
Thank you, @be0x74a !
Merging.

PR #3094, opened 2024-03-05
Add ListRunnersOptions to support all query parameters

### LINKED ISSUE #3093: List[Organization]Runners does not support all query parameters
According to the docs [ListRunners](https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#list-self-hosted-runners-for-a-repository) and [ListOrganizationRunners](https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#list-self-hosted-runners-for-an-organization) support `name` filter as query parameter.

Right now, only `ListOptions` is used.
I will create a PR for this.

fentas:
Closes #3093

This adds `ListRunnersOptions` to be used by `ListRunners` and `ListOrganizationRunners`.
Let me know if something is missing 🙌


github/actions_runners.go
```diff
@@ -151,12 +151,18 @@ type Runners struct {
 	Runners    []*Runner `json:"runners"`
 }
 
+// ListRunnersOptions specifies the optional parameters to the ListRunners and ListOrganizationRunners methods.
+type ListRunnersOptions struct {
+	Name string `url:"name,omitempty"`
```
gmlewis, line 156:
```suggestion
	Name *string `url:"name,omitempty"`
```

gmlewis (changes requested):
Thank you, @fentas !
One tweak, please, then you will need to update your tests and run step 4 scripts listed in CONTRIBUTING.md.

Then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

I'll also change the PR title to note that this is a breaking API change.

github/actions_runners.go
fentas: Done 👍 
I hope you don't mind a side question. I looked at other `List*Options`. Some use ptr values, and others do not. What is the rule here?

gmlewis: Sure! Mandatory params are passed by value and optional params are passed by reference. So `nil` opts just uses the defaults for that endpoint.

gmlewis (approved):
Thank you, @fentas !
LGTM.

be0x74a (approved):
LGTM! 🚀

gmlewis:
Thank you, @be0x74a !
Merging.

PR #3095, opened 2024-03-11
Add Protection to Branch struct

jscaltreto:
Details regarding branch protection are provided APIs which return the "Branch With Protection" schema, such as
'[Get a branch](https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#get-a-branch)'. This allows for getting branch protection information without needing to explicitly call the "Get branch protection" API. As the latter requires `administration:read` permissions while the former only requires `contents:read`, this enables users view branch protection information without requiring elevated permissions.

The '[List branches](https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#list-branches)' API returns the `Short Branch` schema, which does not include `protection`, but since `protected` will still be `true`, a user could subsequently call 'Get a branch' to get further details.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -1574,6 +1574,14 @@ func (b *Branch) GetProtected() bool {
 	return *b.Protected
 }
 
+// GetProtection returns the Protection field.
+func (b *Branch) GetProtection() *Protection {
+	if b == nil {
+		return nil
+	}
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @jscaltreto !
LGTM.
Merging.

Issue #3101, opened 2024-03-14
Changing custom properties


R4vio:
The current library has only the possibility to GET custom properties, while the API has to option to also PATCH as can be seen here:
https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#create-or-update-custom-property-values-for-a-repository

It would be nice if this feature would be included in the library

gmlewis:
This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

HariCharan-001:
I'd like to work on this.

gmlewis:
> I'd like to work on this.

Thank you, @HariCharan-001 !   It's yours.

HariCharan-001:
Submitted a PR, please review !

HariCharan-001:
Raised a new PR !

bhanurp:
this looks to addressed in #3109

gmlewis:
> this looks to addressed in #3109

Thank you, @bhanurp !
Closing as completed.

PR #3102, opened 2024-03-15
Add CreateOrUpdateCustomPropertyValues for repositories

HariCharan-001:
Fixes: #3101

### DIFF
--- a/github/repos_properties.go
+++ b/github/repos_properties.go
@@ -10,6 +10,12 @@ import (
 	"fmt"
 )
 
+// CustomPropertyNonNullValue represents a custom property, with a non-null value.
+type CustomPropertyNonNullValue struct {
+	PropertyName string `json:"property_name"`
+	Value        string `json:"value"`
+}
+
 // GetAllCustomPropertyValues gets all custom property values that are set for a repository.
 //
 // GitHub API docs: https://docs.github.com/rest/repos/custom-properties#get-all-custom-property-values-for-a-repository
@@ -30,4 +36,31 @@ func (s *RepositoriesService) GetAllCustomPropertyValues(ctx context.Context, or
 	}
 
 	return customPropertyValues, resp, nil
+}
+
+// CreateOrUpdateCustomPropertyValues creates new or updates existing custom property values for a repository.
+//
+// GitHub API docs: https://docs.github.com/en/rest/repos/custom-properties#create-or-update-custom-property-values-for-a-repository
+//
+//meta:operation PATCH /repos/{owner}/{repo}/properties/values
+func (s *RepositoriesService) CreateOrUpdateCustomPropertyValues(ctx context.Context, org, repo string, customPropertyValues []*CustomPropertyNonNullValue) (*Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/properties/values", org, repo)
+
+	params := struct {
+		CustomPropertyValues []*CustomPropertyNonNullValue `json:"properties"`
+	}{
+		CustomPropertyValues: customPropertyValues,
+	}
+
+	req, err := s.client.NewRequest("PATCH", u, params)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.client.Do(ctx, req, nil)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
 }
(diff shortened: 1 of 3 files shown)


gmlewis (commented):
According to https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#create-or-update-custom-property-values-for-a-repository,
the `value` field can be a string or an array (of strings maybe?), or alternatively it can be `null` to actually delete that property.

In this repo, we typically try to mirror the functionality of the official documentation in order to hopefully minimize the confusion.  So a reader of the official docs should hopefully be able to understand how to use the endpoints in this repo.

By not allowing `null` (by making value a `*string`), I'm thinking this change might be confusing to a user... because they will be unable to delete a property that they created with this API.

Additionally, we need to think about how we could support either a string value or an array of strings value.

I don't have the answers at the moment.  Maybe we need to think about this one and/or see if there are any similar examples in this repo (I can't think of any off-hand).

HariCharan-001:
Oh, right, I overlooked the array part. Thanks for pointing it out!

I'll think of an approach.

HariCharan-001:
Hi @gmlewis , 

`meta:operation PATCH /repos/{owner}/{repo}/properties/values` is missing in `openapi_operations.yaml`

I did an update to it using, `script/metadata.sh update-openapi `

It bumped _**ghes**_ from 3.11 to 3.12 in the entire file and added the required **_PATCH_** end-point. But it also removed a couple of end-points.

They are
`GET /repositories/{repository_id}/environments/{environment_name}/variables`
`GET /repositories/{repository_id}/environments/{environment_name}/variables/{name}`
and variations of these with **_POST_**, **_PATCH_** and **_DELETE_**

These are being used in `github/action_variables.go` (line 87, for example, and other places in the same file)

I've checked up the github documentation and looks like the endpoint has changed to
`/repos/{owner}/{repo}/environments/{environment_name}/variables`

`{owner}` was added, but I haven't tested if the older one was deprecated or doesn't work anymore.

Should I update this in `action_variables.go` ? 
I can create a separate issue for the same.

I cant proceed to work on the current issue, without this being resolved. 

Another way is I can manually add the _**PATCH**_ path to `openapi_operations.yaml` under the `operation_overrides` section. Not sure if this is a good thing to do.

I'd like to hear your thoughts.

gmlewis:
> I'd like to hear your thoughts.

Let me see if I can update it in a separate PR that you will hopefully be able to merge in and be unblocked.

HariCharan-001:
Sure, works.

Can I do that tho? 😅
It'll be a learning experience.

gmlewis:
> Can I do that tho? 😅
> It'll be a learning experience.

You are welcome to do that, but it sounds like you already tried that and had problems, so I was going to investigate.

Our OpenAPI tools are relatively new (considering the age of this repo), and written by another contributor, and I'm not that familiar with running them myself, so I thought I would dig into it a bit and also look at the official GitHub v3 API docs to see if the removed endpoints have actually been removed or if there is some other problem.

So sure, we can investigate this in parallel if you would like, but I still better look into it from my side to get a better understanding of what exactly is going wrong.  😄

gmlewis:
OK, I'm able to duplicate your findings above... now to research the official GitHub v3 API docs to figure out what happened with these affected endpoints...

gmlewis:
Digging deeper, it looks like the endpoints for these with problems were actually moved by the official GitHub v3 API documentation and now need updating.

@HariCharan-001 - if you would like to modify these 5 endpoints to their new locations in a separate PR, that would be great!

```
$ ./script/metadata.sh update-go
could not find operation "GET /repositories/{repository_id}/environments/{environment_name}/variables" in openapi_operations.yaml
could not find operation "GET /repositories/{repository_id}/environments/{environment_name}/variables/{name}" in openapi_operations.yaml
could not find operation "POST /repositories/{repository_id}/environments/{environment_name}/variables" in openapi_operations.yaml
could not find operation "PATCH /repositories/{repository_id}/environments/{environment_name}/variables/{name}" in openapi_operations.yaml
could not find operation "DELETE /repositories/{repository_id}/environments/{environment_name}/variables/{name}" in openapi_operations.yaml
```

If not, I can do it.  Your call.

HariCharan-001:
Thanks a lot for confirming @gmlewis 😄

I'll work on it.

HariCharan-001:
Will create a new PR on top of new changes I made in #3104 
Closing this.

PR #3104, opened 2024-03-16
feat!: Update deprecated endpoints in github/action_variables.go

HariCharan-001:
Fixes:
#3103 

BREAKING-CHANGE: The following endpoints now take `owner` and `repo` names instead of an integer repo ID:
* ActionsService.ListEnvVariables
* ActionsService.GetEnvVariable
* ActionsService.CreateEnvVariable
* ActionsService.UpdateEnvVariable
* ActionsService.DeleteEnvVariable

PR #3105, opened 2024-03-16
Add CreateOrUpdateCustomPropertyValues for repositories

HariCharanK:
Fixes: 
#3101 

Followup PR to #3104 

Used `interface{}` type for `PropertyValue` as it can be string, array or null according to the GitHub documentation [here](https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#create-or-update-custom-property-values-for-a-repository).

In Go, an `interface{}` can hold any type, including an array (or slice) of strings.


gmlewis:
If you could revert all the changes that are already covered by #3104, that would be great.
Otherwise, we will wait on reviewing this PR until #3104 is merged.

Just a heads-up... we try to avoid using `interface{}` as much as possible in this repo, but in some places we simply can't avoid it.
I don't have time now to review this PR but will hopefully have time this weekend.

HariCharan-001:
The changes in this depend on changes in #3104, so we should wait until it's merged. I don't see otherwise.

`we try to avoid using interface{}`
I see, I'll update if I can do something else instead.

HariCharan-001:
resolved merge conflicts.

gmlewis:
@HariCharan-001 - as far as I can tell, this PR is not addressing the issue in #3101 which is to add a PATCH endpoint for this: https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#create-or-update-custom-property-values-for-a-repository

HariCharan-001:
@gmlewis 
are there any generic issues caused by using interface{} other than lack of type safety and minute impact on performance ?

Do you think it is necessary to avoid using interface{} in this particular scenario ? 
I could think of a couple ways, but they may affect code readability and simplicity. What's the trade-off here ?

gmlewis:
The `interface {}` issue, I think, is secondary to the [issue above](https://github.com/google/go-github/pull/3105#issuecomment-2070701811).

gmlewis:

gmlewis:
Closing PR as abandoned.

PR #3112, opened 2024-03-26
Fix pagination for ListCopilotSeats

alinbalutoiu:
The pagination must be passed as query parameters instead of request [body according to the docs](https://docs.github.com/en/rest/copilot/copilot-user-management?apiVersion=2022-11-28#list-all-copilot-seat-assignments-for-an-organization).

Currently this is silently failing but the pagination is broken if you've got lots of Copilot users.

### DIFF
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -161,8 +161,12 @@ func (s *CopilotService) GetCopilotBilling(ctx context.Context, org string) (*Co
 //meta:operation GET /orgs/{org}/copilot/billing/seats
 func (s *CopilotService) ListCopilotSeats(ctx context.Context, org string, opts *ListOptions) (*ListCopilotSeatsResponse, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/copilot/billing/seats", org)
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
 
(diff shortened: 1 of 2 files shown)


gmlewis (changes requested):
Thank you, @alinbalutoiu !

Before line 587 of `copilot_test.go`, could you please change those lines from:

```go

		_, _, err = client.Copilot.ListCopilotSeats(ctx, "\n", nil)
```

to:

```go

	opts := &ListOptions{Page: 2}
		_, _, err = client.Copilot.ListCopilotSeats(ctx, "\n", opts)
```

That should fix our code coverage problem.

alinbalutoiu:
Thank you @gmlewis, I missed that initially. Fixed it!

gmlewis (approved):
Thank you, @alinbalutoiu !
LGTM.
Merging.

alinbalutoiu:
@gmlewis do you know when this will make into a release version? It would be super helpful to be able to pull a specific version of the library with this fix.

gmlewis:
> @gmlewis do you know when this will make into a release version? It would be super helpful to be able to pull a specific version of the library with this fix.

You can always pull in a specific version using the commit hash:
https://go.dev/wiki/Modules#daily-workflow

Otherwise, we try to release this repo approximately monthly which would be in a couple-ish weeks.

Issue #3114, opened 2024-03-28
feat: Add an option to wait for primary rate limit reset and retry instead of erroring out


erezrokah:
Hi 👋 🐰 

This is mostly the same issue as https://github.com/google/go-github/issues/1920

For our use case we'd prefer to wait until the reset time instead of erroring out when the primary rate limit is reached.
The only way I found to do it at the moment is wrap each API call with a check to `github.RateLimitError`.

We can't use `roundtripper` as `checkRateLimitBeforeDo` happens before. I tried wrapping `BareDo` with my own implementation but couldn't get it to work due to how the services and client are initialized.

If `bypassRateLimitCheck` was exported I might have been able to use it to bypass the check and get use a `roundtripper` but that doesn't seem very elegant.

Can you suggestion a solution?

erezrokah:
That would work great for us

gmlewis:
OK, cool.  Would you like to submit a PR, @erezrokah , or would you prefer that I open this up to our other contributors?

erezrokah:
I can open a PR if you point me to the right direction. Is the suggestion a new pattern to the repo or an existing pattern there's a reference for?

gmlewis:
I believe this would be new functionality.  As such, I have not explored the solution space.
So if you don't have a plan in mind at this point, then maybe we should let another developer tackle this problem, as it will involve some investigation, some thought, and some experimentation to get it right.

erezrokah:
I'm ok with that no worries, wanted to make sure I'm not missing anything. I'll take a look and reach out if I'm blocked

erezrokah:
Opened a PR https://github.com/google/go-github/pull/3117, feedback welcomed

PR #3117, opened 2024-04-01
feat: Allow blocking until primary rate limit is reset

erezrokah:
Fixes https://github.com/google/go-github/issues/3114

Feedback very much welcomed, some notes:
1. I used the same approach as `bypassRateLimitCheck` with the exception that `bypassRateLimitCheck` is private, not sure  if to use another dedicated exported `requestContext` type or also make `bypassRateLimitCheck` public for consistency
2. Checking `req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) == true` is also an option (instead of `!=nil`) and might be more readable, but that is not consistent with the `bypassRateLimitCheck` check
3. A possible improvement can be to add a log message/invoke a callback to consumers so they can show an indication the sleep is happening to avoid looking like something is stuck. If that makes sense it should be in a separate PR I think as that's would be a new API/interface for this module
4. If this approach makes send I will add tests for it


github/github.go
```diff
@@ -889,6 +890,13 @@ func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, erro
 			err = aerr
 		}
 
+		rateLimitError, ok := err.(*RateLimitError)
+		if ok && req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) != nil {
+			time.Sleep(time.Until(rateLimitError.Rate.Reset.Time))
+			// re-send the request
+			return c.BareDo(ctx, req)
```
gmlewis, line 897: I think this is probably OK, but am wondering about two things:
* this recursively calls c.BareDo - at what point should it truly fail?
* The defered body close on line 876 doesn't actually happen until the recursion fully unwraps

Again, I think this is probably all fine.
Did you try testing this out, perchance?
I know the whole point is to prevent rate limit violations - but it might be nice to verify that this does what we want.

gmlewis:
It would also be nice to add one or more unit tests that would increase the code coverage for the new code.

github/github.go
erezrokah: I did test it and the retry works. Good point on the defer let me see if I can find a good solution for it. Another thing I'm considering is to only retry once

erezrokah:
> Thank you, @erezrokah ! LGTM.
> 

Thanks for the quick review and feedback, I also added a 1 second buffer to the sleep time to avoid any edge cases

gmlewis (approved):
Thanks, @erezrokah .

Still awaiting second LGTM+Approval from any other contributor to this repo before merging.

All contributions, including code reviews, by other volunteers is greatly appreciated, as we have several outstanding PRs that need reviews.  Thanks.

erezrokah:
Hi @gmlewis I'm wondering if I should update the PR to take into account context cancellation, by changing `sleepUntilResetWithBuffer` to

```go
timer := time.NewTimer(time.Until(reset) + buffer)
select {
case <-ctx.Done():
	if !timer.Stop() {
	<-timer.C
	}
	return nil, ctx.Err()
case <-timer.C:
}
```

gmlewis:
Ah, that's an excellent idea, @erezrokah - I think you are right that this should be added.

erezrokah:
> Ah, that's an excellent idea, @erezrokah - I think you are right that this should be added.

Added https://github.com/google/go-github/pull/3117/commits/22eba4c83515649fe06e82b4b7dad938a06b5044. In one case I had to wrap the context error in a rate limit error. I can also refactor the code a bit to have `checkRateLimitBeforeDo` return an `error` instead of `*RateLimitError`, but thought it would be better to get some feedback first

github/github.go
```diff
@@ -962,6 +961,19 @@ func (c *Client) checkRateLimitBeforeDo(req *http.Request, rateLimitCategory Rat
 			Header:     make(http.Header),
 			Body:       io.NopCloser(strings.NewReader("")),
 		}
+
+		if req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) != nil {
+			err := sleepUntilResetWithBuffer(req.Context(), rate.Reset.Time)
+			if err == nil {
```
gmlewis, line 967:
```suggestion
			if err := sleepUntilResetWithBuffer(req.Context(), rate.Reset.Time); err == nil {
```

github/github.go
```diff
@@ -892,7 +892,10 @@ func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, erro
 
 		rateLimitError, ok := err.(*RateLimitError)
 		if ok && req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) != nil {
-			sleepUntilResetWithBuffer(rateLimitError.Rate.Reset.Time)
+			err := sleepUntilResetWithBuffer(req.Context(), rateLimitError.Rate.Reset.Time)
+			if err != nil {
```
gmlewis, line 896:
```suggestion
			if err := sleepUntilResetWithBuffer(req.Context(), rateLimitError.Rate.Reset.Time); err != nil {
```

gmlewis (commented):
Just a couple nits, otherwise LGTM.

gmlewis (approved):
Thank you, @erezrokah !
LGTM.

be0x74a (approved):
LGTM!

gmlewis:
Thank you, @be0x74a !
Merging.

erezrokah:
Thanks for the review(s) @be0x74a and @gmlewis, when can we expect a new release to be tagged with this commit?

gmlewis:
> Thanks for the review(s) @be0x74a and @gmlewis, when can we expect a new release to be tagged with this commit?

I'm away from Internet this week so the earliest will be this Friday.

gmlewis:
This is now available here: https://github.com/google/go-github/releases/tag/v62.0.0

PR #3121, opened 2024-04-03
feat: Add missing github enterprise importer and domains meta fields

lrstanley:
Adds two of the missing fields, `github_enterprise_importer` and `domains`, which are now included with the public meta endpoint. Also fixed some minor capitalization errors.


gmlewis (approved):
Thank you, @lrstanley !
LGTM.
Merging.

PR #3129, opened 2024-04-16
feat!: Support querying organization custom roles

tomfeigin:
Added support for querying, creating, updating and deleting organization custom roles.

BREAKING CHANGE: `CreateOrUpdateCustomRoleOptions` has been renamed to `CreateOrUpdateCustomRepoRoleOptions` and `roleID` has been changed from type `string` to `int64`.

### DIFF
--- a/github/orgs_custom_roles.go
+++ b/github/orgs_custom_roles.go
@@ -108,7 +108,7 @@ func (s *OrganizationsService) CreateCustomOrgRole(ctx context.Context, org stri
 // GitHub API docs: https://docs.github.com/rest/orgs/organization-roles#update-a-custom-organization-role
 //
 //meta:operation PATCH /orgs/{org}/organization-roles/{role_id}
-func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org, roleID string, opts *CreateOrUpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
+func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org string, roleID int64, opts *CreateOrUpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/organization-roles/%v", org, roleID)
 
 	req, err := s.client.NewRequest("PATCH", u, opts)
@@ -131,7 +131,7 @@ func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org, rol
 // GitHub API docs: https://docs.github.com/rest/orgs/organization-roles#delete-a-custom-organization-role
 //
 //meta:operation DELETE /orgs/{org}/organization-roles/{role_id}
-func (s *OrganizationsService) DeleteCustomOrgRole(ctx context.Context, org, roleID string) (*Response, error) {
+func (s *OrganizationsService) DeleteCustomOrgRole(ctx context.Context, org string, roleID int64) (*Response, error) {
 	u := fmt.Sprintf("orgs/%v/organization-roles/%v", org, roleID)
 
 	req, err := s.client.NewRequest("DELETE", u, nil)
--- a/github/orgs_custom_roles.go
+++ b/github/orgs_custom_roles.go
@@ -10,6 +10,20 @@ import (
 	"fmt"
 )
 
+// OrganizationCustomRoles represents custom organization roles available in specified organization.
+type OrganizationCustomRoles struct {
+	TotalCount      *int              `json:"total_count,omitempty"`
+	CustomRepoRoles []*CustomOrgRoles `json:"roles,omitempty"`
+}
+
+// CustomOrgRoles represents custom organization role available in specified organization.
+type CustomOrgRoles struct {
+	ID          *int64   `json:"id,omitempty"`
+	Name        *string  `json:"name,omitempty"`
+	Description *string  `json:"description,omitempty"`
+	Permissions []string `json:"permissions,omitempty"`
+}
+
 // OrganizationCustomRepoRoles represents custom repository roles available in specified organization.
 type OrganizationCustomRepoRoles struct {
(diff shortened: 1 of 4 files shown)


github/orgs_custom_roles_test.go
```diff
@@ -14,6 +14,151 @@ import (
...
+	if !cmp.Equal(resp.StatusCode, 204) {
+		t.Errorf("Organizations.DeleteCustomOrgRole returned  status code %+v, want %+v", resp.StatusCode, "204")
+	}
+
+	const methodName = "DeleteCustomOrgRole"
+	testBadOptions(t, methodName, func() (err error) {
+		_, err = client.Organizations.DeleteCustomOrgRole(ctx, "\no", "8030")
+		return err
+	})
+}
```
gmlewis, line 160:
```suggestion

	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
		return client.Organizations.DeleteCustomOrgRole(ctx, "o", "8030")
	})
}
```

gmlewis (changes requested):
Thank you, @tomfeigin !
Just one addition, please, then I think we will be ready for a second LGTM+Approval before merging.

gmlewis:
(Also, there is no need to force push in this repo - we always squash&merge - see CONTRIBUTING.md for details.)

github/orgs_custom_roles.go
```diff
@@ -27,6 +41,113 @@ type CustomRepoRoles struct {
...
+	return resultingRole, resp, err
+}
+
+// UpdateCustomOrgRole updates a custom role in this organization.
+// In order to update custom roles in an organization, the authenticated user must be an organization owner.
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/organization-roles#update-a-custom-organization-role
+//
+//meta:operation PATCH /orgs/{org}/organization-roles/{role_id}
+func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org, roleID string, opts *CreateOrUpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
```
gmlewis, line 111: According to: https://docs.github.com/en/rest/orgs/organization-roles?apiVersion=2022-11-28#update-a-custom-organization-role
`roleID` should be an integer.

```suggestion
func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org string, roleID int64, opts *CreateOrUpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
```

github/orgs_custom_roles.go
```diff
@@ -27,6 +41,113 @@ type CustomRepoRoles struct {
...
+	return resultingRole, resp, err
+}
+
+// DeleteCustomOrgRole deletes an existing custom role in this organization.
+// In order to delete custom roles in an organization, the authenticated user must be an organization owner.
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/organization-roles#delete-a-custom-organization-role
+//
+//meta:operation DELETE /orgs/{org}/organization-roles/{role_id}
+func (s *OrganizationsService) DeleteCustomOrgRole(ctx context.Context, org, roleID string) (*Response, error) {
```
gmlewis, line 134: Same here: https://docs.github.com/rest/orgs/organization-roles#delete-a-custom-organization-role

```suggestion
func (s *OrganizationsService) DeleteCustomOrgRole(ctx context.Context, org string, roleID int64) (*Response, error) {
```

tomfeigin: I'll change it but it is aligned with `DeleteCustomRepoRole`

tomfeigin: I prefer to do that in another PR because here I am just introducing new stuff so it won't break existing usage

tomfeigin:
@gmlewis this specific PR is not a breaking a change as it only introduces new API wrappers (I am pretty sure 🙃 ) can you remove the label please?

github/orgs_custom_roles.go
```diff
@@ -50,21 +171,13 @@ func (s *OrganizationsService) ListCustomRepoRoles(ctx context.Context, org stri
...
-}
-
 // CreateCustomRepoRole creates a custom repository role in this organization.
 // In order to create custom repository roles in an organization, the authenticated user must be an organization owner.
 //
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/orgs/custom-roles#create-a-custom-repository-role
 //
 //meta:operation POST /orgs/{org}/custom-repository-roles
-func (s *OrganizationsService) CreateCustomRepoRole(ctx context.Context, org string, opts *CreateOrUpdateCustomRoleOptions) (*CustomRepoRoles, *Response, error) {
+func (s *OrganizationsService) CreateCustomRepoRole(ctx context.Context, org string, opts *CreateOrUpdateCustomRepoRoleOptions) (*CustomRepoRoles, *Response, error) {
```
gmlewis, line 180: @tomfeigin - if I'm not mistaken, this line and line 203 below are both breaking API changes, as code written by a user of this repo that calls one or both of these endpoints will no longer compile correctly without changing their code.  Agreed?

tomfeigin: Oh you are right, should I keep the previous naming then?

gmlewis: Well, honestly I don't have a problem making breaking API changes, which I suppose might be obvious from the current version number of this repo.  😂 

Also, I think it is always a good idea to make things clearer and easier-to-understand for users of this repo... so I think your name change here was perfectly appropriate and that's also why I recommended fixing the bad field type while we are making a breaking API change.

So I'll leave it up to you, but I think the breaking API change is appropriate and am willing to move forward with it.

gmlewis (approved):
Thank you, @tomfeigin !
LGTM.

gsaraf (approved):
Looks good to me :)

gmlewis:
Thank you, @gsaraf !
Merging.

Issue #3127, opened 2024-04-11
Use enums for the action field in GitHub Webhooks


prnvbn:
## Issue Description:

Many GitHub webhooks contain the action field, which is enumerated. For example, in the [GitHub documentation for webhook payloads and events](https://docs.github.com/en/webhooks/webhook-events-and-payloads#check_run), the check_run event includes an action property that can take on values such as "completed", "created", "requested_action", and "rerequested". However, in the library, these fields are currently represented as free-form strings. In my opinion, making these enums would be more user-friendly.

## Proposed Solution

A rudimentary example of how this can be done is as follows ([go play ground link](https://goplay.tools/snippet/gBqRSMGO0Px))

```go
package main

import (
	"encoding/json"
	"fmt"
)

type ChecRunAction string

const (
	CheckRunCreated         ChecRunAction = "created"
	CheckRunRequested       ChecRunAction = "requested"
	CheckRunRerequested     ChecRunAction = "rerequested"
	CheckRunRequestedAction ChecRunAction = "requested_action"
)

func main() {
	jsonStr := `{"hi": "created"}`

	var v struct {
		A ChecRunAction `json:"hi"`
	}
	json.Unmarshal([]byte(jsonStr), &v)
	fmt.Println(v)
}

```

prnvbn:
If you agree, I can make this change

gmlewis:
That seems safe to me.

prnvbn:
PR: https://github.com/google/go-github/pull/3136

PR #3136, opened 2024-04-21
Add enums for webhook payloads action field

prnvbn:
Relevant Issue: https://github.com/google/go-github/issues/3127

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -20,23 +20,30 @@ type RequestedAction struct {
 //
 // GitHub API docs: https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#branch_protection_rule
 type BranchProtectionRuleEvent struct {
-	Action       *string               `json:"action,omitempty"`
-	Rule         *BranchProtectionRule `json:"rule,omitempty"`
-	Changes      *ProtectionChanges    `json:"changes,omitempty"`
-	Repo         *Repository           `json:"repository,omitempty"`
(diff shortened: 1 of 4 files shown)


prnvbn:
@gmlewis @willnorris any update on the PR?

prnvbn:
this can be done progressively as well if you prefer and do it one event at a time (or in batches)

gmlewis:
@willnorris - do you have any opinions about a large breaking change like this to change a bunch of `*string`s into explicit enum values?

prnvbn:
Would be interested to know if there is a decision taken about this. I personally would prefer enums as they are self documenting the possibilities a value can have.

willnorris:
> @willnorris - do you have any opinions about a large breaking change like this to change a bunch of `*string`s into explicit enum values?

Sorry I missed your question back in August.  Personally, I'm not a fan.  The surface area of the library is already huge, so I'd be pretty sensitive to changes like this that increase it all the more. 🤷🏻

gmlewis:
> > @willnorris - do you have any opinions about a large breaking change like this to change a bunch of `*string`s into explicit enum values?
> 
> Sorry I missed your question back in August. Personally, I'm not a fan. The surface area of the library is already huge, so I'd be pretty sensitive to changes like this that increase it all the more. 🤷🏻

Thank you, @willnorris - and thank you, @prnvbn for your patience. 
I will close this PR as we will not be pursuing this change.

Issue #3152, opened 2024-04-23
Support new REST API endpoint that evaluates if private vulnerability reporting is enabled


diogoteles08:
GitHub blog announcement: https://github.blog/changelog/2024-03-08-check-if-private-vulnerability-reporting-is-enabled-via-rest-api/

New GET endpoint: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#check-if-private-vulnerability-reporting-is-enabled-for-a-repository

gmlewis:
Thank you, @diogoteles08 !
Do you want to create a PR for this, or shall I open it up to other contributors?

diogoteles08:
No problem! ^^
Well I might take some time to implement it, as I'll probably use this feature in my project. But wouldn't be able to start **right now**, so let's leave it open in case someone else can start this work sooner than me.

Sairaviteja27:
[gmlewis](https://github.com/gmlewis) I'd be happy to take a look at this and see if I can implement a REST API endpoint to check if private vulnerability reporting is enabled. Is [diogoteles08](https://github.com/diogoteles08) currently working on this? If not, I'm ready to jump in.

diogoteles08:
>  Is @diogoteles08 currently working on this? If not, I'm ready to jump in.

I'm not! Please feel free to work on this, that'd be great 😊

PR #3157, opened 2024-05-01
Add support for check private vulnerability reporting endpoint

Sairaviteja27:
fixes https://github.com/google/go-github/issues/3152


github/repos.go
```diff
@@ -2402,3 +2402,35 @@ func (s *RepositoriesService) DisablePrivateReporting(ctx context.Context, owner
 
 	return resp, nil
 }
+
+// CheckPrivateReporting represents whether private vulnerability reporting is enabled.
+type CheckPrivateReporting struct {
```
gmlewis, line 2407: This should be private since it isn't exposed anywhere to the end-user.

github/repos.go
```diff
@@ -2402,3 +2402,35 @@ func (s *RepositoriesService) DisablePrivateReporting(ctx context.Context, owner
...
+	u := fmt.Sprintf("repos/%v/%v/private-vulnerability-reporting", owner, repo)
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return false, nil, err
+	}
+
+	privateReporting := new(CheckPrivateReporting)
+	resp, err := s.client.Do(ctx, req, privateReporting)
+	if privateReporting.Enabled {
```
gmlewis, line 2427: We don't need to explicitly check for `StatusNotFound` below as far as I can tell from the docs, so this could all be simplified to just:

```suggestion
	return privateReporting.Enabled, resp, err
```

gmlewis (changes requested):
Thank you, @Sairaviteja27 !

github/repos.go
gmlewis: The only reason I called this out here is because the official documentation doesn't say that it ever returns a "404 Not Found"... hence I don't think we need it in this particular case.

github/repos.go
```diff
@@ -2402,3 +2402,31 @@ func (s *RepositoriesService) DisablePrivateReporting(ctx context.Context, owner
...
+		return false, nil, err
+	}
+
+	privateReporting := new(checkPrivateReporting)
+	resp, err := s.client.Do(ctx, req, privateReporting)
+	if privateReporting.Enabled {
+		return privateReporting.Enabled, resp, err
+	}
+
+	return false, resp, err
```
gmlewis, line 2431: Note that the zero value of a boolean in Go is `false`. Therefore these five lines can be replaced with:

```suggestion
	return privateReporting.Enabled, resp, err
```

gmlewis (approved):
Thank you, @Sairaviteja27 !
LGTM.

Sairaviteja27:
@gmlewis I think code coverage check is failing. Could you suggest if I should I add tests in other methods?

gmlewis:
> @gmlewis I think code coverage check is failing. Could you suggest if I should I add tests in other methods?

Don't worry about that.  We've had some weirdness with codecov lately and I've made two reports with the problems we've seen and don't have any resolution yet.  Thanks, though.

be0x74a (approved):

gmlewis:
Thank you, @be0x74a !
Merging.

PR #3163, opened 2024-05-07
Add Action to DeploymentStatusEvent

JoannaaKL:
In [GH docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads#deployment_status) we can see that `deployment_status` has a field `action`, that is always set to `created`. This pr adds it to json representation.


gmlewis (approved):
Thank you, @JoannaaKL !
LGTM.
Merging.

PR #3164, opened 2024-05-09
feat: Add missing values_editable_by field for organization custom properties

lrstanley:
Adds missing field `values_editable_by` for custom properties, as well as mirrors the field documentation directly from the API documentation for the other fields of custom properties.

Ref: https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28#create-or-update-custom-properties-for-an-organization


gmlewis (approved):
Thank you, @lrstanley !
LGTM.
Merging.

Issue #3137, opened 2024-04-22
Can't remove  repository ruleset's every Bypass Actors because of serialization issue


lado936:
Hello all, 
I am using go-github repository ruleset update function from [here](https://github.com/google/go-github/blob/8d4be0b2cf2b8242a0b7e3777232cd0058ae535f/github/repos_rules.go#L493)

but issue is that, if i already have bypassactors set and want to remove all of them, it not applying changes as bypassactors has omitempty set [here](https://github.com/google/go-github/blob/8d4be0b2cf2b8242a0b7e3777232cd0058ae535f/github/repos_rules.go#L391), if i understand correctly, omitempty must be removed if i want to pass empty list there. Proposing to remove omitempty if there is no case i can't see, why its needed.

lado936:
Better, if someone else will work on this, new to Golang and will need way more time i guess

gmlewis:
> Better if someone else will work on this, new to Golang and will need way more time i guess

OK, no problem.

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

Matthew-Reidy:
I can take this

PR #3174, opened 2024-05-22
Allow an empty array of BypassActors in Ruleset struct in CreateRuleset endpoint

Matthew-Reidy:
Fixes: [#3137](https://github.com/google/go-github/issues/3137)

added an extra Ruleset struct and CreateRuleset endpoint in repo_rules.go to allow the BypassActors field to be passed as nil or an empty array


gmlewis (changes requested):
It think it might be a more pleasant user experience to not introduce a new struct,
but just a new endpoint that still uses the `Ruleset` struct as input, but internally uses a non-exported struct like the one you created. 

To do this, you would need to copy over the fields to this non-exported struct and then send that instead of the original Ruleset.

Also, let's please not start using `BPActr` as a short abbreviation for `BypassActor`, as we try to make this repo as searchable as possible, and nobody is going to guess `bpactr` as a search string.

Also, all new godoc-style comments should be complete sentences with ending periods. Thanks.

Matthew-Reidy:
Do you mean something like having an anonymous struct identical to Ruleset in the new functions scope that would just copy over the values of Ruleset to?

gmlewis:
> Do you mean something like having an anonymous struct identical to Ruleset in the new functions scope that would just copy over the values of Ruleset to? I understand your idea but i'm confused about the implementation.

Yes, that is what I mean.

We have other examples of this in our repo.

Here is one of many: https://github.com/google/go-github/blob/master/github/repos.go#L546-L573

I found this by typing:

```bash
$ grep 'type [a-z]' github/*.go
```

github/repos_rules.go
```diff
@@ -507,6 +525,50 @@ func (s *RepositoriesService) UpdateRuleset(ctx context.Context, owner, repo str
...
+			Enforcement:  rs.Enforcement,
+			BypassActors: rs.BypassActors,
+			NodeID:       rs.NodeID,
+			Links:        rs.Links,
+			Conditions:   rs.Conditions,
+			Rules:        rs.Rules,
+		}
+	} else {
+		rsNoBypassActor = &RulesetNoOmitBypassActors{}
+	}
```
gmlewis, line 556:
```suggestion
	rsNoBypassActor := &RulesetNoOmitBypassActors{}

	if rs != nil {
		rsNoBypassActor = &RulesetNoOmitBypassActors{
			ID:           rs.ID,
			Name:         rs.Name,
			Target:       rs.Target,
			SourceType:   rs.SourceType,
			Source:       rs.Source,
			Enforcement:  rs.Enforcement,
			BypassActors: rs.BypassActors,
			NodeID:       rs.NodeID,
			Links:        rs.Links,
			Conditions:   rs.Conditions,
			Rules:        rs.Rules,
		}
	}
```

github/repos_rules.go
```diff
@@ -395,6 +395,24 @@ type Ruleset struct {
 	Rules        []*RepositoryRule  `json:"rules,omitempty"`
 }
 
+// RulesetNoOmitBypassActors represents a GitHub ruleset object. The struct does not omit bypassActors if the field is nil or an empty array is passed.
+type RulesetNoOmitBypassActors struct {
```
gmlewis, line 399:
```suggestion
// rulesetNoOmitBypassActors represents a GitHub ruleset object. The struct does not omit bypassActors if the field is nil or an empty array is passed.
type rulesetNoOmitBypassActors struct {
```

gmlewis (changes requested):
Yes, this is what I was talking about - the only difference is that the new struct should not be exported because the end-user never has know or worry about it.

Matthew-Reidy:
Ok. I commited your suggested changes. I didn't know rulesetNoOmitBypassActors made it non exported. Learned something new about the language. Thanks!

github/repos_rules.go
```diff
@@ -507,6 +525,48 @@ func (s *RepositoriesService) UpdateRuleset(ctx context.Context, owner, repo str
...
+//
+// This function is necessary as the UpdateRuleset function does not marschall ByPassActor if passed as nil or an empty array.
+//
+// GitHub API docs: https://docs.github.com/rest/repos/rules#update-a-repository-ruleset
+//
+//meta:operation PUT /repos/{owner}/{repo}/rulesets/{ruleset_id}
+func (s *RepositoriesService) UpdateRulesetNoBypassActor(ctx context.Context, owner, repo string, rulesetID int64, rs *Ruleset) (*Ruleset, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/rulesets/%v", owner, repo, rulesetID)
+
+	rsNoBypassActor := &RulesetNoOmitBypassActors{}
```
gmlewis, line 538:
```suggestion
	rsNoBypassActor := &rulesetNoOmitBypassActors{}
```

github/repos_rules.go
```diff
@@ -507,6 +525,48 @@ func (s *RepositoriesService) UpdateRuleset(ctx context.Context, owner, repo str
...
+
+	if rs != nil {
+		rsNoBypassActor = &RulesetNoOmitBypassActors{
```
gmlewis, line 541:
```suggestion
		rsNoBypassActor = &rulesetNoOmitBypassActors{
```

gmlewis:
Please re-run step 4 in CONTRIBUTING.md to re-generate the auto-generated files... then the tests should all pass.
You can also test locally before you push by running `go test ./...`

gmlewis (approved):
Thank you, @Matthew-Reidy !
LGTM.

github/repos_rules.go
```diff
@@ -395,6 +395,24 @@ type Ruleset struct {
 	Rules        []*RepositoryRule  `json:"rules,omitempty"`
 }
 
+// rulesetNoOmitBypassActors represents a GitHub ruleset object. The struct does not omit bypassActors if the field is nil or an empty array is passed.
+type rulesetNoOmitBypassActors struct {
```
tomfeigin, line 399: I think that you can embed the entire `Ruleset` object and shadow just the `bypass_actors`

```golang
type rulesetNoOmitBypassActors struct {
   Ruleset
   BypassActors []*BypassActor `json:"bypass_actors"`
}
```

example: https://goplay.tools/snippet/7lHZci7LW0y

@gmlewis WDYT?

Then you can use it like so:

```golang
func (s *RepositoriesService) UpdateRulesetNoBypassActor(...rs *Ruleset...
  rsNoBypassActor := &rulesetNoOmitBypassActors{}
   if rs != nil {
      rsNoBypassActor := &rulesetNoOmitBypassActors{Ruleset: *rs}
   }
```

github/repos_rules.go
```diff
@@ -507,6 +525,48 @@ func (s *RepositoriesService) UpdateRuleset(ctx context.Context, owner, repo str
 	return ruleset, resp, nil
 }
 
+// UpdateRulesetNoBypassActor updates a ruleset for the specified repository.
+//
+// This function is necessary as the UpdateRuleset function does not marschall ByPassActor if passed as nil or an empty array.
```
tomfeigin, line 530: NIT: `marschall` -> `marshal`

github/repos_rules.go
gmlewis: Sounds great, @tomfeigin - as long as we have a unit test that demonstrates this behavior (meaning a `BypassActors=nil` still gets passed through in JSON as `"bypass_actors":[]`.

lewismiddleton:
This would be useful for me as I'm tracking [a bug](https://github.com/integrations/terraform-provider-github/issues/2269) in the GitHub terraform provider that I think links to #3137.

@Matthew-Reidy are you planning on picking this up again?

Matthew-Reidy:
I can fix the spelling mistake but as for the other proposed change whats the justification? The solution in place now on this PR seems to do the same things and has testing evidence behind it. Thanks.

tomfeigin (approved):
My comments are only relevant for styling, let's merge it

gmlewis:
Thank you, @tomfeigin and @Matthew-Reidy !
@tomfeigin - after this is merged, feel free to create a PR to improve the styling if you wish.

Merging.

Issue #3175, opened 2024-05-23
WorkflowRun struct does not include `path` property


rmacklin:
The `path` property is included in the response from these GitHub API endpoints:

- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run-attempt

but it is not included here:
https://github.com/google/go-github/blob/8c1232a5960307a6383998e1aa2dd71711343810/github/actions_workflow_runs.go#L16-L51

rmacklin:
Opened a PR: https://github.com/google/go-github/pull/3176

PR #3176, opened 2024-05-23
Support `path` property in `WorkflowRun`

rmacklin:
Fixes https://github.com/google/go-github/issues/3175

The `path` property is included in the response from these GitHub API endpoints:

- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run-attempt


gmlewis (approved):
Thank you, @rmacklin !
LGTM.
Merging.

PR #3184, opened 2024-06-11
Add fields to the InstallationPermissions struct

tomfeigin:
Added the fields:
- actions_variables
- organization_custom_org_roles

They were missing from the permissions map


gmlewis (approved):
Thank you, @tomfeigin !
LGTM.
Merging.

PR #3191, opened 2024-06-16
Fix go-githubauth link in README.md

bored-engineer:
The URL is missing the schema so GitHub treats it as a relative link.


gmlewis (approved):
Thank you, @bored-engineer !
LGTM.
Merging.

PR #3197, opened 2024-07-05
feat!: Replace ListCursorOptions with ListIDPGroupsOptions in TeamsService.ListIDPGroupsInOrganization

### LINKED ISSUE #3196: ListIDPGroupsInOrganization is missing the filter parameter
`ListIDPGroupsInOrganization` function that implements the [`/orgs/ORG/team-sync/groups`](https://docs.github.com/en/enterprise-cloud@latest/rest/teams/team-sync?apiVersion=2022-11-28#list-idp-groups-for-an-organization) endpoint, doesn't have the parameter to pre-filter results:

> `q` string
>
> Filters the results to return only those that begin with the value specified by this parameter. For example, a value of `ab` will return results that begin with "ab".

For organizations that have thousands of IdP groups, this makes using this function very slow. Please see https://github.com/integrations/terraform-provider-github/issues/1809 for more context on how this affects the GitHub Terraform provider.

I'm requesting to add this parameter to the function to fully implement the API specification and enable downstream consumers.

laughedelic:
Fixes #3196.

BREAKING CHANGE: Replace `ListCursorOptions` with `ListIDPGroupsOptions` in `TeamsService.ListIDPGroupsInOrganization`

I followed the template of this PR, fixing a similar issue: https://github.com/google/go-github/pull/3094. 

Let me know if there is a better way to handle this. And I'm open to better naming suggestions.

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -796,6 +796,18 @@ func (s *TeamsService) RemoveTeamProjectBySlug(ctx context.Context, org, slug st
 	return s.client.Do(ctx, req, nil)
 }
 
+// ListIDPGroupsOptions specifies the optional parameters to the ListIDPGroupsInOrganization method.
+type ListIDPGroupsOptions struct {
+	// The number of results to include per page.
+	PerPage int `url:"per_page,omitempty"`
+
+	// Page token.
+	Page string `url:"page,omitempty"`
+
+	// Filters the results to return only those that begin with the value specified by this parameter.
+	Query string `url:"q,omitempty"`
+}
+
 // IDPGroupList represents a list of external identity provider (IDP) groups.
 type IDPGroupList struct {
 	Groups []*IDPGroup `json:"groups"`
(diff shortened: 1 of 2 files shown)


github/teams.go
```diff
@@ -796,6 +796,18 @@ func (s *TeamsService) RemoveTeamProjectBySlug(ctx context.Context, org, slug st
...
+// ListIDPGroupsOptions specifies the optional parameters to the ListIDPGroupsInOrganization method.
+type ListIDPGroupsOptions struct {
+	// The number of results to include per page.
+	PerPage int `url:"per_page,omitempty"`
+
+	// Page token.
+	Page string `url:"page,omitempty"`
+
+	// Filters the results to return only those that begin with the value specified by this parameter.
+	Query string `url:"q,omitempty"`
```
gmlewis, line 808:
```suggestion
	// Filters the results to return only those that begin with the value specified by this parameter.
	Query string `url:"q,omitempty"`
	
	ListOptions
```

gmlewis (changes requested):
Thank you, @laughedelic .
One minor change, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/teams.go
laughedelic: Hi! That's what I wanted to do at first, but unfortunately `ListOptions` doesn't fit here because its `Page` field is of type int and this endpoint uses a _string_ token. `ListCursorOptions` that was here before would do, but then it has all other cursor-related fields that this API does not accept. So does this situation warrant a new specific type? Or should we still reuse `ListCursorOptions`?

gmlewis: Oh, right!  I apologize.  Yes, please use `ListCursorOptions` in this case.  
Thank you for catching my mistake!  I appreciate it.

laughedelic: No problem. Done in e492cdd

gmlewis (approved):
Thank you, @laughedelic !
LGTM.

laughedelic:
Hi! Is there anything I can do to move this forward?

gmlewis:
> Hi! Is there anything I can do to move this forward?

If you have any coworkers or know of any other contributors to this repo (who are all volunteers, please remember), feel free to reach out to request a code review if you are comfortable with that.

arymoraes (approved):
LGTM

gmlewis:
Thank you, @arymoraes !
Merging.

PR #3200, opened 2024-07-07
feat!: Add support for multi-select Custom Properties

### LINKED ISSUE #3198: Bug: Multi-select Custom Properties are not supported
## Issue

The Github API supports organizational custom properties with the following types: `string, single_select, multi_select, true_false`. Creating a property with type `multi_select` or `true_false` breaks many of the library interactions due to unmarshalling errors

API Reference: https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28#create-or-update-custom-properties-for-an-organization

## Reproducing

- Create a custom property of type `multi_select` via the UI
- Use library to get organizational custom properties
  - Unmarshalling error
- Assign the custom property a value on a repository
-  Use the library to get the Repository
  - Unmarshalling error

arymoraes:
Fixes #3198.

BREAKING CHANGE: `CustomPropertyValue.Value` is changed from `*string` to `interface{}` to support `string` and `[]string` values.

So the problem here is that the GitHub custom properties `value` can come in either as a `string` (for all non `multi_select` properties), or as a `[]string` in a `multi_select` property.

The issue is happening because we were trying to unmarshal it but always expecting a `string` for the Value.

My solution was to make value an `interface{}` type that implements a `UnmarshalJSON` to handle both scenarios. While this works, we end up losing the auto generated `GetValue()` accessor. Another thing I was unsure is that I saw on other places and on the tests for this that we use a pointer to the value, but I changed it to the value itself on the Unmarshal call since we lose the accessor, not sure if this is correct.

Another solution I played with was using `json.RawMessage`, I could not make it fully work (I can give it another try), but at least we did lose the auto generated `GetValue` accessor there, would this be a more adequate solution?

### DIFF
--- a/github/orgs_properties.go
+++ b/github/orgs_properties.go
@@ -7,6 +7,7 @@ package github
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 )
 
@@ -40,8 +41,42 @@ type RepoCustomPropertyValue struct {
 
 // CustomPropertyValue represents a custom property value.
 type CustomPropertyValue struct {
-	PropertyName string  `json:"property_name"`
-	Value        *string `json:"value,omitempty"`
+	PropertyName string      `json:"property_name"`
+	Value        interface{} `json:"value,omitempty"`
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+// This helps us handle the fact that Value can be either a string, []string, or nil.
+func (cpv *CustomPropertyValue) UnmarshalJSON(data []byte) error {
+	type aliasCustomPropertyValue CustomPropertyValue
+	aux := &struct {
+		*aliasCustomPropertyValue
+	}{
+		aliasCustomPropertyValue: (*aliasCustomPropertyValue)(cpv),
+	}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	switch v := aux.Value.(type) {
+	case nil:
+		cpv.Value = nil
+	case string:
+		cpv.Value = v
+	case []interface{}:
+		strSlice := make([]string, len(v))
+		for i, item := range v {
+			if str, ok := item.(string); ok {
+				strSlice[i] = str
+			} else {
+				return fmt.Errorf("non-string value in string array")
+			}
+		}
+		cpv.Value = strSlice
+	default:
+		return fmt.Errorf("unexpected value type: %T", v)
+	}
+	return nil
 }
 
 // GetAllCustomProperties gets all custom properties that are defined for the specified organization.
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -4766,14 +4766,6 @@ func (c *CustomProperty) GetValuesEditableBy() string {
 	return *c.ValuesEditableBy
 }
 
-// GetValue returns the Value field if it's non-nil, zero value otherwise.
-func (c *CustomPropertyValue) GetValue() string {
-	if c == nil || c.Value == nil {
-		return ""
-	}
-	return *c.Value
-}
-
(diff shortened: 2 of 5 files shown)


gmlewis:
Yes, if I remember correctly, we handle situations like this in various ways:

* with `interface{}` like you did (e.g. [copilot](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/copilot.go#L47) )
* with `json.RawMessage` like you mention (e.g. [scim](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/scim.go#L181) , [repos_deployments](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/repos_deployments.go#L22) , [repos_rules](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/repos_rules.go#L129) , etc.)
* with multiple structs (not finding any good examples at the moment)

So I'm fine with this method. You could additionally define two new structs `CustomPropertyValueWithString` and `CustomPropertyValueWithStringSlice` that have the respective types for `Value` and then two helper functions could each return one of the new types with a `bool` (indicating if the returned pointer is actually that type or not) if you want... but I'm fine if you want to just run with the current solution too.

arymoraes:
> Yes, if I remember correctly, we handle situations like this in various ways:
> 
> * with `interface{}` like you did (e.g. [copilot](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/copilot.go#L47) )
> * with `json.RawMessage` like you mention (e.g. [scim](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/scim.go#L181) , [repos_deployments](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/repos_deployments.go#L22) , [repos_rules](https://github.com/google/go-github/blob/dd756703cb6b93cc8dc3e826799128c4e8707db6/github/repos_rules.go#L129) , etc.)
> * with multiple structs (not finding any good examples at the moment)
> 
> So I'm fine with this method. You could additionally define two new structs `CustomPropertyValueWithString` and `CustomPropertyValueWithStringSlice` that have the respective types for `Value` and then two helper functions could each return one of the new types with a `bool` (indicating if the returned pointer is actually that type or not) if you want... but I'm fine if you want to just run with the current solution too.

That makes sense, thank you for the explanation and the breakdown. I'm fine with the current solution as is if we want to go with that, at some point I did something similar to having two new structs but ended up deciding to move back to trying the `interface{}` approach.

gmlewis (approved):
Thank you, @arymoraes!
LGTM.

valbeat (approved):

gmlewis:
Thank you, @valbeat !
Merging.

github/orgs_properties.go
```diff
@@ -40,8 +41,42 @@ type RepoCustomPropertyValue struct {
...
+		aliasCustomPropertyValue: (*aliasCustomPropertyValue)(cpv),
+	}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	switch v := aux.Value.(type) {
+	case nil:
+		cpv.Value = nil
+	case string:
```
darend, line 64: Do boolean custom properties come through as a string as well?

https://docs.github.com/en/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization#adding-custom-properties

arymoraes: Yes, they come as `string` from the GitHub API:

```json
      {
        "property_name": "boolean_prop",
        "value": "true"
      },
```

Issue #3203, opened 2024-07-10
feature: Set the callback URL at app creation time


Roming22:
The homepage and webhook URLs can be set at creation time, but the callback URL cannot.
This issue is about adding the option of setting that field during the app creation process.

Roming22:
@gmlewis The PR is live and ready to be reviewed. It is my understanding that the section I am modifying is currently not covered by unit tests. However I did test the change successfully in my application.

PR #3204, opened 2024-07-10
Add support for App Callback URLs

Roming22:
Fixes: #3203

### DIFF
--- a/scrape/apps.go
+++ b/scrape/apps.go
@@ -116,7 +116,7 @@ type AppManifest struct {
 	Name *string `json:"name,omitempty"`
 	//Required. The homepage of your GitHub App.
 	URL *string `json:"url,omitempty"`
-	// The full URL of the endpoint to authenticate users via the GitHub App.
+	// The full URL(s) of the endpoint(s) to authenticate users via the GitHub App (Max: 10).
 	CallbackURLs []string `json:"callback_urls,omitempty"`
 	// Required. The configuration of the GitHub App's webhook.
 	HookAttributes map[string]string `json:"hook_attributes,omitempty"`
--- a/scrape/apps.go
+++ b/scrape/apps.go
@@ -110,6 +110,7 @@ type OAuthApp struct {
 
 // AppManifest represents a GitHub App manifest, used for preconfiguring
 // GitHub App configuration.
+// c.f. https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest
 type AppManifest struct {
 	// The name of the GitHub App.
 	Name *string `json:"name,omitempty"`
--- a/scrape/apps.go
+++ b/scrape/apps.go
@@ -117,7 +117,7 @@ type AppManifest struct {
 	//Required. The homepage of your GitHub App.
 	URL *string `json:"url,omitempty"`
 	// The full URL of the endpoint to authenticate users via the GitHub App.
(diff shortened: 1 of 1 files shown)


scrape/apps.go
```diff
@@ -115,6 +115,8 @@ type AppManifest struct {
 	Name *string `json:"name,omitempty"`
 	//Required. The homepage of your GitHub App.
 	URL *string `json:"url,omitempty"`
+	// The full URL of the endpoint to authenticate users via the GitHub App.
+	CallbackURL *string `json:"url,omitempty"`
```
gmlewis, line 119: This change makes no sense to me, based on line 117 above it. 
Note that both fields have identical JSON bindings, which means they get mapped to the same thing:

```
	URL *string `json:"url,omitempty"`
	CallbackURL *string `json:"url,omitempty"`
```

gmlewis (changes requested):
Thank you, @Roming22 !
Could I please trouble you to add the GitHub API Docs URL as a comment to this code that shows where this is documented?
I know this is the `scrape` directory, but it would be nice to have a reference to it.
Thanks again!

Roming22:
> Thank you, @Roming22 ! Could I please trouble you to add the GitHub API Docs URL as a comment to this code that shows where this is documented? I know this is the `scrape` directory, but it would be nice to have a reference to it. Thanks again!

I do not know where, or whether, it is documented. I look at the surrounding fields and made a lucky guess.

Roming22:
Found it!
Link added as a comment. Please check that it looks like the right one to you too.

scrape/apps.go
```diff
@@ -110,11 +110,14 @@ type OAuthApp struct {
 
 // AppManifest represents a GitHub App manifest, used for preconfiguring
 // GitHub App configuration.
+// c.f. https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest
```
gmlewis, line 113: Thank you for adding the link, although now I'm slightly confused.
According to this link, the field should be:

```go
CallbackURLs []string `json:"callback_urls,omitempty"`
```

Could you please try this out in your local code and compare it to the one you added and see if there is a difference and/or if both of them work? (In which case, should we actually _add_ both?!?!?)

Roming22: Both works. I haven't tested the behavior when using both.

Since `CallbackURLs` is a superset of `CallbackURL`, implementing only the former seems best to avoid user confusion.

scrape/apps.go
```diff
@@ -110,11 +110,14 @@ type OAuthApp struct {
...
 type AppManifest struct {
 	// The name of the GitHub App.
 	Name *string `json:"name,omitempty"`
 	//Required. The homepage of your GitHub App.
 	URL *string `json:"url,omitempty"`
+	// The full URL of the endpoint to authenticate users via the GitHub App.
```
gmlewis, line 119:
```suggestion
	// The full URL(s) of the endpoint(s) to authenticate users via the GitHub App (Max: 10).
```

gmlewis (approved):
Thank you, @Roming22 !
LGTM.
Merging.

Roming22:
Thanks for the support @gmlewis!

Is there an ETA for the next release?

gmlewis:
It looks like it is about time.  I'll work on it.

gmlewis:
This is now available here: https://github.com/google/go-github/releases/tag/v63.0.0

PR #3208, opened 2024-07-15
Add doc for NewPullRequest

kevinburkesegment:
I recently mixed up "Head" and "Base" and the error message was not clear. Add some guidance to help people avoid making the same mistake.

This is copied word for word from the existing Github docs.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -250,9 +250,16 @@ func (s *PullRequestsService) GetRaw(ctx context.Context, owner string, repo str
 
 // NewPullRequest represents a new pull request to be created.
 type NewPullRequest struct {
-	Title               *string `json:"title,omitempty"`
-	Head                *string `json:"head,omitempty"`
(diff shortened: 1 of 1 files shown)


gmlewis:
Thank you, @kevinburkesegment - would you like to sign the CLA so that I can merge this, or would you prefer I open up a separate PR and copy the original text from the official GitHub documentation into that one instead?

kevinburkesegment:
I believe I signed the CLA but if you're still not seeing it on your end please feel free to duplicate the work. I do not need credit, I'd just like it to be fixed so I don't make the same mistake 6 months from now. Thanks

gmlewis (approved):
> I believe I signed the CLA but if you're still not seeing it on your end please feel free to duplicate the work. I do not need credit, I'd just like it to be fixed so I don't make the same mistake 6 months from now. Thanks

Excellent, thank you, @kevinburkesegment !
LGTM.
Merging.

Issue #3213, opened 2024-07-19
Implementing Personal Access Tokens API


AbhishekAg:
I want to use personal token api, but the go implementation is missing in the go-github package.
Please help to implement the same and commit in the main repo, so that I can pull the latest codebase and use it.

Github Rest API doc:
https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#list-fine-grained-personal-access-tokens-with-access-to-organization-resources

Below CURL command is working and able to list the fined grained access tokens authorised by the organization

```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <YOUR-TOKEN>" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/orgs/ORG/personal-access-tokens
```

gmlewis:
This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

arymoraes:
Can I take this?

I had lots of fun on the previous issue.

gmlewis:
Absolutely, @arymoraes, and thank you! 
It's yours.

arymoraes:
> Absolutely, @arymoraes, and thank you! It's yours.

Awesome! Thank you again.

AbhishekAg:
@arymoraes Will PR take sometime?

PR #3215, opened 2024-07-20
Add List fine-grained personal access tokens with access to organization resources API

arymoraes:
Fixes: #3213.

### DIFF
--- a/github/orgs_personal_access_tokens.go
+++ b/github/orgs_personal_access_tokens.go
@@ -9,6 +9,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
 )
 
 // PersonalAccessToken represents the minimal representation of an organization programmatic access grant.
@@ -46,20 +48,52 @@ type PersonalAccessToken struct {
 	TokenLastUsedAt *Timestamp `json:"token_last_used_at"`
 }
 
+// ListFineGrainedPATOptions specifies optional parameters to ListFineGrainedPersonalAccessTokens.
+type ListFineGrainedPATOptions struct {
+	// The property by which to sort the results.
+	// Default: created_at
+	// Value: created_at
+	Sort string `url:"sort,omitempty"`
+
+	// The direction to sort the results by.
+	// Default: desc
+	// Value: asc, desc
+	Direction string `url:"direction,omitempty"`
+
+	// A list of owner usernames to use to filter the results.
+	Owner []string `url:"-"`
+
+	// The name of the repository to use to filter the results.
+	Repository string `url:"repository,omitempty"`
+
+	// The permission to use to filter the results.
+	Permission string `url:"permission,omitempty"`
+
+	// Only show fine-grained personal access tokens used before the given time.
+	// This is a timestamp in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.
+	LastUsedBefore string `url:"last_used_before,omitempty"`
+
+	// Only show fine-grained personal access tokens used after the given time.
+	// This is a timestamp in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.
+	LastUsedAfter string `url:"last_used_after,omitempty"`
+
+	ListOptions
+}
+
 // ListFineGrainedPersonalAccessTokens lists approved fine-grained personal access tokens owned by organization members that can access organization resources.
 // Only GitHub Apps can call this API, using the `Personal access tokens` organization permissions (read).
 //
 // GitHub API docs: https://docs.github.com/rest/orgs/personal-access-tokens#list-fine-grained-personal-access-tokens-with-access-to-organization-resources
 //
 //meta:operation GET /orgs/{org}/personal-access-tokens
-func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.Context, org string, opts *ListOptions) ([]*PersonalAccessToken, *Response, error) {
+func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.Context, org string, opts *ListFineGrainedPATOptions) ([]*PersonalAccessToken, *Response, error) {
 	u := fmt.Sprintf("orgs/%v/personal-access-tokens", org)
-	u, err := addOptions(u, opts)
+	u, err := addListFineGrainedPATOptions(u, opts)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	req, err := s.client.NewRequest(http.MethodGet, u, &opts)
+	req, err := s.client.NewRequest(http.MethodGet, u, opts)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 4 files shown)


github/orgs_personal_access_tokens.go
```diff
@@ -11,6 +11,69 @@ import (
...
+	TokenLastUsedAt *Timestamp `json:"token_last_used_at"`
+}
+
+// ListFineGrainedPersonalAccessTokens lists approved fine-grained personal access tokens owned by organization members that can access organization resources.
+// Only GitHub Apps can call this API, using the `Personal access tokens` organization permissions (read).
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/personal-access-tokens#list-fine-grained-personal-access-tokens-with-access-to-organization-resources
+//
+//meta:operation GET /orgs/{org}/personal-access-tokens
+func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.Context, org string, opts *ListOptions) ([]*PersonalAccessToken, *Response, error) {
```
gmlewis, line 55: This endpoint can take a whole bunch more options:

```
sort stringThe property by which to sort the results.Default: created_atValue: created_at
--
direction stringThe direction to sort the results by.Default: descCan be one of: asc, desc
owner arrayA list of owner usernames to use to filter the results.
repository stringThe name of the repository to use to filter the results.
permission stringThe permission to use to filter the results.
last_used_before stringOnly show fine-grained personal access tokens used before the given time. This is a timestamp in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.
last_used_after stringOnly show fine-grained personal access tokens used after the given time. This is a timestamp in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ.

sort string
The property by which to sort the results.

Default: created_at

Value: created_at

direction string
The direction to sort the results by.

Default: desc

Can be one of: asc, desc

owner array
A list of owner usernames to use to filter the results.

repository string
The name of the repository to use to filter the results.

permission string
The permission to use to filter the results.

last_used_before string
Only show fine-grained personal access tokens used before the given time. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: YYYY-MM-DDTHH:MM:SSZ.

last_used_after string
Only show fine-grained personal access tokens used after the given time. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: YYYY-MM-DDTHH:MM:SSZ.
```

So let's please create a new `type ListFineGrainedPATOptions struct { ... ListOptions }` that can take all these parameter, and make sure to add `omitempty` to each one of them since they are optional.

gmlewis (changes requested):
Thank you, @arymoraes!

github/orgs_personal_access_tokens.go
arymoraes: Good point and sorry for missing that. I added that on my latest commit.

One thing though, using the `owner` array as 

```go
Owner           []string `url:"owner,omitempty,comma"`
```

will cause the API to return 422 status code, as it expects the `[]owner=...` format, so I had to create the helper function below.

I tried to find an example on the codebase and I found `IssueListOptions`, but this one seems to work as that endpoint expects the `orgs/{ORG}/issues?labels=bug%2Cfeature` format

github/orgs_personal_access_tokens_test.go
```diff
@@ -65,7 +77,12 @@ func TestOrganizationsService_ListFineGrainedPersonalAccessTokens(t *testing.T)
 		]`)
 	})
 
-	opts := &ListOptions{Page: 2, PerPage: 2}
+	opts := &ListFineGrainedPATOptions{
+		ListOptions: ListOptions{Page: 2, PerPage: 2},
+		Sort:        "created_at",
+		Direction:   "desc",
+		Owner:       []string{"octocat"},
```
gmlewis, line 84: Please modify your test case to list multiple owners in order to demonstrate how this code behaves for this case.

arymoraes: done!

github/orgs_personal_access_tokens.go
gmlewis: @arymoraes - I'm looking all over the GitHub v3 API documentation and can't find `[]owner=...` anywhere.
Where did you find this?  It seems incredibly strange syntax to me for a query parameter.

arymoraes: I also found it very strange. I did not find it on the docs, but by trial and error (and by remembering seeing this kind of issue once). 
If you run a CURL with:

curl -L \
```
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: {TOKEN}" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/orgs/{ORG}/personal-access-tokens?owner={username}
```

It will give you a `422 status code error`:

```json
{
  "message": "Invalid request.\n\nInvalid input: `\"{username}\"` is not of type `array`.",
  "documentation_url": "https://docs.github.com/rest/orgs/personal-access-tokens#list-fine-grained-personal-access-tokens-with-access-to-organization-resources",
  "status": "422"
}
```

However:

```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer {TOKEN}" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/orgs/{ORG}/personal-access-tokens?owner[]={username}
```

Is valid and will indeed return you a list of tokens filtered by that owner

gmlewis: OK, wow, that is bizarre... so I was hoping that your unit test would show how multiple owners would appear in the resulting URL, but it doesn't... do they show up as comma-separated values?

arymoraes: Honestly, it gets even weirder if you want multiple owners.

```
personal-access-tokens?owner[]=owner1,owner2
```

doesn't work, it'll just return an empty array of tokens (maybe if thinks the owner name is `owner1,owner2`?

You have to do

```
personal-access-tokens?owner[]=owner1&owner[]=owner2
```

To be able to filter tokens from any of those two owners. This is the one that is generated with `addListFineGrainedPATOptions`.

I am finding this really weird, but this is the way it works even with curl, so I'm assuming this is indeed the way that GitHub API wants us to send the query param. I tried to search for others with the same question, but could not. I could take another look though.

I should probably improve the comment for `addListFineGrainedPATOptions` to add the multiple owners scenario, and maybe also test the URL explicitly on the unit test?

gmlewis: Wow, that's amazing... thank you for the details, @arymoraes !

Maybe could you please just add a comment explaining all this weirdness would be extremely helpful for anyone else that comes across this and tries to use it.  Thank you!

arymoraes: Done!

I added the comments to the `addListFineGrainedPATOptions` helper function declaration, as well as inside `addListFineGrainedPATOptions` when we call `addListFineGrainedPATOptions`, so whoever runs into that code knows why we are doing this.

gmlewis (approved):
Thank you, @arymoraes !
LGTM.

AbhishekAg:
@arymoraes @gmlewis Could you please check who will be the next reviewer?

gmlewis:
> @arymoraes @gmlewis Could you please check who will be the next reviewer?

In this repo, we depend upon volunteers to review.  Since you are another contributor to this repo, you are welcome to review this PR and give your feedback. If you approve, please LGTM+Approve and we can move forward with this PR. Thank you.

AbhishekAg (approved):
LGTM

gmlewis:
Thank you, @AbhishekAg !
Merging.

PR #3221, opened 2024-07-29
Update codecov-action version

gmlewis:


gmlewis:
This does _NOT_ solve the problem because codecov appears to be using a very old commit as the BASE:
https://app.codecov.io/gh/google/go-github/commit/2b8c7fa86fcd1cb4a89a44c8e7b5a49fe4be8a45

After more investigation, I found this thread:
https://community.codecov.com/t/incorrect-base-commit-of-pull-request/1779/2

However, its "solution" was solved in a private conversation. I'll try adding to the thread.

gmlewis:
According to this: https://github.com/codecov/feedback/issues/123
It may be that the codecov token being used is out-of-date?!?

gmlewis:
This issue also mentions the codecov token: https://github.com/codecov/feedback/issues/144

gmlewis:
@google-admin - when you get a chance can you please make sure that this repo's CODECOV secret repo token is up-to-date?

According to all the similar issues I could find (linked above), they say that an old CODECOV token could be the cause of it choosing very old (6-months ago) BASE commits to compare against. 

Thank you!

gmlewis:
@google-admin - when you get a chance can you please make sure that this repo's CODECOV secret repo token is up-to-date?

According to all the similar issues I could find (linked above), they say that an old CODECOV token could be the cause of it choosing very old (6-months ago) BASE commits to compare against.

Thank you!

Issue #3210, opened 2024-07-18
Add support for GitHub App Manifest flow on Organization, not User


fproulx-boostsecurity:
As described in GitHub's doc, the GitHub App Manifest flow can be executed against a user or an org (https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest#implementing-the-github-app-manifest-flow). Right now, this library only supports on a user, which is far from ideal, because it would mean that a CLI using this library would have a potential critical app be created on an end user GitHub user account. Typically GitHub Apps are for service accounts / privileged access and it does not make sense for an employee, who might leave to hold hostage the app config, potential rotating Private Key and gaining access to org he used to be part of.

Here, instead of `https://github.com/settings/apps/new`, we should have the option to set an org as the App Owner (i.e. https://github.com/organizations/ORGANIZATION/settings/apps/new)
https://github.com/google/go-github/blob/5f496dddb88539d3e4fd0a1ee30b2f139e6736a7/scrape/apps.go#L138

gmlewis:
Thank you, @fproulx-boostsecurity. Would you like to create a PR or shall I open this up to other contributors to this repo?

fproulx-boostsecurity:
@gmlewis better open the floor to others. I have other projects that keep me busy now. I noticed that OpenShift Pipelines as code was using this feature and it fell short because of that, but I hacked my way around it and overrode the URL using DevTools for today.

Would be nice for that to be addressed though so consumer of the lib would have better experience.

gmlewis:
This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

wenked:
Hi , im a new go developer and i would like to have a opportunity to work on this issue.

gmlewis:
Thank you, @wenked - it is yours.

fproulx-boostsecurity:
Awesome @wenked , looking forward to see the PR 👍

PR #3222, opened 2024-07-29
feat!: Add support for creating GitHub App with organizations

wenked:
Fixes: #3210.

**BREAKING CHANGE:** The `CreateApp` function now requires two arguments: `AppManifest` and `orgName`, to support creating apps with organizations.

The issue occurred because the create app URL path was only set up for creating personal GitHub apps.

My solution was to add an additional argument to the function to provide the necessary information for creating organizational GitHub Apps. The second argument, `AppConfig`, now includes configuration structs with an `OwnerType` that can be either personal or organizational, and you can provide the organization name by passing `OrgName` in the `AppConfig` struct.

For more details, see the issue: [https://github.com/google/go-github/issues/3210](https://github.com/google/go-github/issues/3210).

### DIFF
--- a/scrape/apps.go
+++ b/scrape/apps.go
@@ -149,11 +149,11 @@ type AppConfig struct {
 }
 
 // CreateApp creates a new GitHub App with the given manifest configuration.
-func (c *Client) CreateApp(m *AppManifest, accType *AppConfig) (*http.Response, error) {
+func (c *Client) CreateApp(m *AppManifest, appConfig *AppConfig) (*http.Response, error) {
 	path := "/settings/apps/new"
 
-	if accType.OwnerType == "organizational" {
-		path = fmt.Sprintf("/organizations/%s/settings/apps/new", accType.OrgName)
+	if appConfig.OwnerType == "organizational" {
+		path = fmt.Sprintf("/organizations/%s/settings/apps/new", appConfig.OrgName)
 	}
 
 	u, err := c.baseURL.Parse(path)
@@ -162,7 +162,6 @@ func (c *Client) CreateApp(m *AppManifest, accType *AppConfig) (*http.Response, 
 	}
 
 	body, err := json.Marshal(map[string]*AppManifest{"manifest": m})
-
 	if err != nil {
 		return nil, err
 	}
--- a/scrape/apps.go
+++ b/scrape/apps.go
@@ -148,12 +148,11 @@ type AppConfig struct {
 	OrgName string
 }
 
-// CreateApp creates a new GitHub App with the given manifest configuration.
-func (c *Client) CreateApp(m *AppManifest, appConfig *AppConfig) (*http.Response, error) {
+func (c *Client) CreateApp(m *AppManifest, ac *AppConfig) (*http.Response, error) {
 	path := "/settings/apps/new"
 
-	if appConfig.OwnerType == "organizational" {
-		path = fmt.Sprintf("/organizations/%s/settings/apps/new", appConfig.OrgName)
+	if ac.OwnerType == "organizational" {
+		path = fmt.Sprintf("/organizations/%s/settings/apps/new", ac.OrgName)
 	}
 
 	u, err := c.baseURL.Parse(path)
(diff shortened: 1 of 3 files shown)


scrape/scrape.go
```diff
@@ -128,6 +128,7 @@ func (c *Client) Authenticate(username, password, otpseed string) error {
 		values.Set("password", password)
 	}
 	resp, err := fetchAndSubmitForm(c.Client, "https://github.com/login", setPassword)
+
```
gmlewis, line 131: Please revert this change.  It is idiomatic Go to tightly couple the handling of an error with the statement that generates the error without a blank line in between.

scrape/apps.go
```diff
@@ -133,9 +134,28 @@ type AppManifest struct {
...
+	OwnerType OwnerType
+	// OrgName is the name of the organization if the OwnerType is organizational.
+	OrgName string
+}
+
+func (c *Client) CreateApp(m *AppManifest, ac *AppConfig) (*http.Response, error) {
+	path := "/settings/apps/new"
+
+	if ac.OwnerType == "organizational" {
+		path = fmt.Sprintf("/organizations/%s/settings/apps/new", ac.OrgName)
```
gmlewis, line 155: nit: for consistency in this repo when using `grep` or `ripgrep` to search for endpoint URLs:

```suggestion
		path = fmt.Sprintf("/organizations/%v/settings/apps/new", ac.OrgName)
```

scrape/apps.go
```diff
@@ -133,9 +134,28 @@ type AppManifest struct {
 	DefaultPermissions *github.InstallationPermissions `json:"default_permissions,omitempty"`
 }
 
-// CreateApp creates a new GitHub App with the given manifest configuration.
-func (c *Client) CreateApp(m *AppManifest) (*http.Response, error) {
-	u, err := c.baseURL.Parse("/settings/apps/new")
+type OwnerType string
```
gmlewis, line 137: Please add a `godoc`-style comment for this exported type.

scrape/apps.go
```diff
@@ -133,9 +134,28 @@ type AppManifest struct {
...
+
+const (
+	OwnerTypePersonal     OwnerType = "personal"
+	OwnerTypeOrganization OwnerType = "organizational"
+)
+
+type AppConfig struct {
```
gmlewis, line 144: Please add a `godoc`-style comment for this exported type.

scrape/apps.go
```diff
@@ -133,9 +134,28 @@ type AppManifest struct {
...
+	// OwnerType specifies the type of GitHub account owner (personal or organizational).
+	OwnerType OwnerType
+	// OrgName is the name of the organization if the OwnerType is organizational.
+	OrgName string
+}
+
+func (c *Client) CreateApp(m *AppManifest, ac *AppConfig) (*http.Response, error) {
+	path := "/settings/apps/new"
```
gmlewis, line 152: nit: for repo consistency:

```suggestion
	url := "/settings/apps/new"
```

gmlewis (changes requested):
Thank you, @wenked !
Please address the few minor comments.

@willnorris - since this involves `scrape`, do you have any comments or concerns regarding the breaking change in this PR?

scrape/apps.go
```diff
@@ -133,9 +134,28 @@ type AppManifest struct {
...
+)
+
+type AppConfig struct {
+	// OwnerType specifies the type of GitHub account owner (personal or organizational).
+	OwnerType OwnerType
+	// OrgName is the name of the organization if the OwnerType is organizational.
+	OrgName string
+}
+
+func (c *Client) CreateApp(m *AppManifest, ac *AppConfig) (*http.Response, error) {
```
willnorris, line 151: This AppConfig struct feels over-engineered.  In the case of creating a personal app, you can't provide a nil AppConfig, since line 154 will panic, but otherwise none of the values get used.  I *think* all you really want is the organization name, right?  So why not just pass an `orgName string` parameter, and if it's an empty string then create a personal app?

willnorris (commented):
I don't have any sense on who is using this package (not many people, looking at pkg.go.dev), so I don't have strong feelings either way on the change.  looks fine to me.

scrape/apps.go
wenked: You're right, I'll do that instead .

gmlewis (approved):
Thank you, @wenked !
LGTM.
Merging.

Issue #3229, opened 2024-08-08
Error when parsing repository that uses multiselect custom-properties


spalberg:
## Problem overview

We started using githubs [custom properties](https://docs.github.com/en/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization) recently and noticed that our [reviewdog/action-actionlint](https://github.com/reviewdog/action-actionlint) checks started to fail with the following error:
```
reviewdog: failed to build repo base HTML URL: json: cannot unmarshal array into Go struct field Repository.custom_properties of type string
```
_reviewdog/action-actionlint_ uses _reviewdog/reviewdog_ which in turn uses the current version of _google/go-github_ to parse github related json. 

## Analysis

Handling for `custom_properties` was added in this PR: https://github.com/google/go-github/pull/3065.
The field is expected to be a map of string values (https://github.com/google/go-github/pull/3065/files#diff-7a0544e44d81e972917728b13efd3fa55e7f9d1bd8edb745216494fa1e178f93R80).

The actual api response can also include non-string values. The following is part of an actual api response obtained by using the github cli (`gh api repos/OWNER/REPO`):
```jsonc
{
  // ...
  "allow_update_branch": true,
  "use_squash_pr_title_as_default": false,
  "squash_merge_commit_message": "COMMIT_MESSAGES",
  "squash_merge_commit_title": "COMMIT_OR_PR_TITLE",
  "merge_commit_message": "PR_TITLE",
  "merge_commit_title": "MERGE_MESSAGE",
  "custom_properties": {
    "org-managed-rulesets": "true",
    "technologies": [
      "helm",
      "java"
    ]
  },
  // ...
}
```

The custom property `technologies` is multi select and github seems to return the selected values as an array which causes the parser to error out. The custom property `org-managed-rulesets` is of type boolean and represented using a string by github.

Sadly, [githubs api docs](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository) are really bad when it comes to `custom_property`, so I can't say for sure if there are other output options than string or array of strings:
```jsonc
{
  // ...
  "custom_properties": {
      "type": "object",
      "description": "The custom properties that were defined for the repository. The keys are the custom property names, and the values are the corresponding custom property values.",
      "additionalProperties": true
  }
  // ...
}
```

gmlewis:
Oh!  Wow... TIL!  Thank you for the report, @spalberg !

[...]

PR #3230, opened 2024-08-10
feat!: Widen CustomProperties type to map[string]interface{} to align with GitHub API

spalberg:
Fixes: #3229.

BREAKING CHANGE: `PushEventRepository.CustomProperties` is changed from `map[string]string` to `map[string]interface{}`.

adds tests for json unmarshalling of projects


github/repos_test.go
```diff
@@ -4463,3 +4463,57 @@ func TestRepositoriesService_IsPrivateReportingEnabled(t *testing.T) {
...
...erfest"],"visibility":"public","forks":2026,"open_issues":81,"watchers":10234,"default_branch":"master","permissions":{"admin":false,"maintain":false,"push":false,"triage":false,"pull":true},"temp_clone_token":"","custom_properties":{},"organization":{"login":"google","id":1342004,"node_id":"MDEyOk9yZ2FuaXphdGlvbjEzNDIwMDQ=","avatar_url":"https://avatars.githubusercontent.com/u/1342004?v=4","gravatar_id":"","url":"https://api.github.com/users/google","html_url":"https://github.com/google","followers_url":"https://api.github.com/users/google/followers","following_url":"https://api.github.com/users/google/following{/other_user}","gists_url":"https://api.github.com/users/google/gists{/gist_id}","starred_url":"https://api.github.com/users/google/starred{/owner}{/repo}","subscriptions_url":"https://api.github.com/users/google/subscriptions","organizations_url":"https://api.github.com/users/google/orgs","repos_url":"https://api.github.com/users/google/repos","events_url":"https://api.github.com/users/google/events{/privacy}","received_events_url":"https://api.github.com/users/google/received_events","type":"Organization","site_admin":false},"network_count":2026,"subscribers_count":211}`)
```
gmlewis, line 4509: Instead of an actual response, how about if you just put one value of each type that you found, please?

gmlewis (changes requested):
Thank you, @spalberg !
Just one comment, then we should be able to merge after a second LGTM+Approval from any other contributor to this repo.

gmlewis:
Maybe @chapurlatn can review these changes?

github/repos_test.go
spalberg: I minimized the json as you suggested to have a small partial project

github/repos_test.go
```diff
@@ -4463,3 +4464,49 @@ func TestRepositoriesService_IsPrivateReportingEnabled(t *testing.T) {
...
+			wantRepository: Repository{},
+			wantErr:        true,
+		},
+		"Partial project": {
+			data:           []byte(`{"id":10270722,"name":"go-github","private":false,"owner":{"login":"google"},"created_at":"2013-05-24T16:42:58Z","license":{},"topics":["github"],"permissions":{"pull":true},"custom_properties":{},"organization":{"login":"google"}}`),
+			wantRepository: Repository{ID: Int64(10270722), Name: String("go-github"), Private: Bool(false), Owner: &User{Login: String("google")}, CreatedAt: &Timestamp{time.Date(2013, 5, 24, 16, 42, 58, 0, time.UTC)}, License: &License{}, Topics: []string{"github"}, Permissions: map[string]bool{"pull": true}, CustomProperties: map[string]interface{}{}, Organization: &Organization{Login: String("google")}},
+			wantErr:        false,
+		},
+		"With custom properties": {
+			data:           []byte(`{"custom_properties":{"boolean":"false","text":"a","single-select":"a","multi-select":["a","b","c"]}}`),
```
gmlewis, line 4490: How about this?

```suggestion
			data:           []byte(`{"custom_properties":{"boolean":false,"int":42,"key":"value","array":["a","b","c"]}}`),
```

spalberg: This would imply that there is a possible integer value when there is non (that we know of). I used the different custom property types offered by GitHub (see [here](https://docs.github.com/en/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization#:~:text=This%20can%20either%20be%20a%20text%20string%2C%20a%20single%20select%20field%2C%20a%20multi%20select%20field%2C%20or%20a%20true/false%20boolean)) and observed their representation in the api, using their label in the github ui as key for this json.

I guess it depends on whats the goal - testing that the unmarshalling works according to the type or testing the possible values that github could send for the given field. What do you think?

gmlewis: Oh, I see, well at least please change the boolean from a string to `false` without the quotes.

spalberg: GitHub returns the value for boolean type custom properties as strings ("true" or "false"). Should I remove the quotes regardless?

gmlewis: Are you sure?  We have hundreds of endpoints that return booleans, and none of them are strings.
What makes you say this?  You made a boolean custom property and it is returning a string?!?

spalberg: Yes, I made custom properties of all types and looked at the api response from github.
This is the actual response from the github api with one custom property (named accordingly) of each type configured:

```jsonc
{
  // ...
   "custom_properties": {
      "type-multi-select": [
        "a",
        "b"
      ],
      "type-single-select": "b",
      "type-text": "lala",
      "type-true-false": "true"
  },
  //...
}
```

spalberg: I'd love if they would return an actual boolean here :smile:

gmlewis: OK, bizarre.
Then it looks like there is one linting failure remaining... let me look.

gmlewis: It looks like the generated files are out-of-date if you could please update them.

spalberg: The whole integration of file generation is a really cool golang feature!
Just pushed the updated files :+1:

gmlewis (approved):
Thank you, @spalberg !
LGTM

arymoraes (approved):
LGTM

And yeah I remember when I was doing https://github.com/google/go-github/pull/3200 that a `boolean` type custom property comes as `string` from the GitHub API.

gmlewis:
Thank you, @arymoraes !
Merging.

Issue #3226, opened 2024-08-07
Unable to create Custom Repository Role without permissions


andriyun:
In the current implementation, the client allows you to request a Custom Repository Role for an Organization without specifying the required keys. 
[CreateCustomRepoRole](https://github.com/google/go-github/blob/master/github/orgs_custom_roles.go#L96) use as argument `opts *CreateOrUpdateCustomRoleOptions` which has all key as optional https://github.com/google/go-github/blob/master/github/orgs_custom_roles.go#L60

As a result, the client will allow the creation request to GitHub and will fail with an error.

In most cases, what is missing might be clear and reasonable.

But it doesn't work when creating a custom role with no permissions. In other words when CreateCustomRepoRole gets an empty array as a `Permission` key.

Due to [omitempty](https://github.com/google/go-github/blob/master/github/orgs_custom_roles.go#L64C41-L64C51) in JSON tag of the `Permission` key, an empty array as value is omitted from the final request to GitHub API, while it is required by the API

gmlewis:
It sounds like we need a new endpoint with an additional private struct which removes the `omitempty` from the existing `CreateOrUpdateCustomRoleOptions` struct, but is otherwise identical.  We've done things like this in other places.
For example, here: https://github.com/google/go-github/blob/474a48f2a16e9f8325ef89ae78312529f6646bcd/github/orgs_properties.go#L112-L116

@andriyun - would you like to make a PR to address this, or shall I open this up to other contributors to this repo?

andriyun:
I'll be glad to submit PR with changes to fix it :)

gmlewis:
Thank you, @andriyun !  It's yours.

andriyun:
Hey @gmlewis 
Sorry for late question. I didn't get your idea about creating new endpoint. I could neither find other cases you are referring too.
It's true at there are some similar constructions there with `CreateOrUpdate****Options`
However there are not used separate `Create` Update` methods.

I've added PR with changes https://github.com/google/go-github/pull/3235
Please take a look

andriyun:
I understand your point now. It makes perfect sense.

I changed the PR in such a way, that it doesn't break the current API.
However, I couldn't run tests locally without fails. The fails I see look not relevant for changes I have added

JamBalaya56562:
This looks resolved — it was fixed by your own PR #3235 ("Allow create custom repository/organization roles without permission"), merged 2024-08-19.

[...]

PR #3235, opened 2024-08-13
Allow create custom repository/organization roles without permission

andriyun:
Changes related to the issue https://github.com/google/go-github/issues/3226

### DIFF
--- a/github/orgs_custom_roles.go
+++ b/github/orgs_custom_roles.go
@@ -53,7 +53,7 @@ type CustomRepoRoles struct {
 type CreateOrUpdateOrgRoleOptions struct {
 	Name        *string  `json:"name,omitempty"`
 	Description *string  `json:"description,omitempty"`
-	Permissions []string `json:"permissions,omitempty"`
+	Permissions []string `json:"permissions"`
 }
 
 // CreateOrUpdateCustomRepoRoleOptions represents options required to create or update a custom repository role.
@@ -61,7 +61,7 @@ type CreateOrUpdateCustomRepoRoleOptions struct {
 	Name        *string  `json:"name,omitempty"`
 	Description *string  `json:"description,omitempty"`
 	BaseRole    *string  `json:"base_role,omitempty"`
-	Permissions []string `json:"permissions,omitempty"`
+	Permissions []string `json:"permissions"`
 }
 
 // ListRoles lists the custom roles available in this organization.


andriyun:
See a discussion about the approach in the relevant issue https://github.com/google/go-github/issues/3226

gmlewis (changes requested):
So this is much closer, but my point was that we need two _NEW_ endpoints so as to not affect the people who are currently not modifying the permissions (and therefore _need_ the `omitempty`).

We need a new `CreateCustomOrgRoleNoPermissions` and a new `CreateCustomRepoRoleNoPermissions` that both use the "trick" of an inline struct that removes the `omitempty` like you have done here.
The comment for each of these need to clearly state why these two endpoints are needed.

andriyun:
I don't think I understand the need of having separate endpoint.

It is clear for me that method _CreateCustomRepoRole_ will fail if you call it without any of properties of type _CreateOrUpdateCustomRepoRoleOptions_. Which is expected and makes sense after reading GitHub Rest API

However it will also fail when you call it with defined values for each of the properties where empty array is used as value for _Permission_. Which is my case exactly.

The aim of this PR is to fix the bug. While still using the same methods without breaking current API as you requested :)

Taking into account above written, what would be the benefits from having extra endpoint?

gmlewis:
So isn't it true that users will want the `omitempty` when they wish to not _modify_ existing permissions?
Or are you saying that there is never a usage for not sending the `permissions` field?
I'll go read the docs again as well.

I know we have other endpoints where we must differentiate between an omitted field and an empty field, and I thought this was one of those cases.

gmlewis:
This endpoint says that `permissions` is required: https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-roles?apiVersion=2022-11-28#create-a-custom-repository-role

However, this endpoint makes _NO MENTION_ of the `permissions` field: https://docs.github.com/en/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role

What's up with that?

andriyun:
> So isn't it true that users will want the omitempty when they wish to not modify existing permissions?

Yes it is true. In this case I would expect _Permission_ property to be _nil_, so condition will return false and parameters would be send untouched.

The intention of the changes is to rewrite type for the case when _Permissions_ property has an empty array as value

andriyun:
> However, this endpoint makes NO MENTION of the permissions field: https://docs.github.com/en/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role

This link lead me to: Get all organization roles

If you open this link https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role

You find _permission_ key as required

gmlewis:
OK, so if I'm reading the docs correctly, then the `permissions` field is _ALWAYS_ required, correct?

If so, I truly apologize for chasing you down an incorrect rabbit hole that I created.

So it looks like completely removing the `omitempty` is simply the right choice afterall, agreed?

andriyun:
NP. Glad we clarified the intention here :)

Unfortunatelly type _CreateOrUpdateCustomRepoRoleOptions_ used for both _Create_ (where all the properties are required) and _Update_ where all of them are optional 😀

So we either rewrite API or use tweaks like I suggested in the PR

gmlewis:
OK, now that we've tried all the possibilities, I'll let you decide which you think is most appropriate.
Please let mw know when you want me to take another look.
If it requires breaking the API, then so be it.  If, however, you can avoid it, that would be nice-to-have.

gmlewis:
@andriyun - can you please explain to me how what you have written is _different_ from simply removing the `omitempty` on lines 56 and 64?

I'm honestly having troubles understanding the difference.

andriyun:
Ok, I was wrong in my assumption before. And solution here is way more simpler

gmlewis (approved):
Thank you, @andriyun !
LGTM.

gmlewis:
(nit: @andriyun - In the future, please don't use force-push in this repo, as now the history of all the experiments we went through and discussed are now lost and we can't refer to them in case others want to discuss them. Also, we always squash-and-merge in this repo so there is never a concern about bloated history in the final commits.)

andriyun:
Ohhh.. ok, got you. I have previous changes saved in separate branches locally.

Do you want me to add them?

gmlewis:
No, that's OK... if someone in the future wants to dig up what we discussed, we can try and recreate then.  😂 
Thank you for offering, though! 😄

tomfeigin (approved):
LGTM

gmlewis:
Thank you, @tomfeigin !
Merging.

PR #3249, opened 2024-08-22
feat!: Fix meta service domains schema structure

### LINKED ISSUE #3248: `MetaService.Get` fails because the schema of the JSON in the API response is different from the expected in `APIMeta.Domains` type
Downstream issue: https://github.com/integrations/terraform-provider-github/issues/2357

GitHub has a meta information API, and go-github has a corresponding implementation:

- https://docs.github.com/en/rest/meta/meta?apiVersion=2022-11-28#get-github-meta-information
  - Can be called in https://api.github.com/meta
- https://pkg.go.dev/github.com/google/go-github/v63/github#MetaService
  - https://github.com/google/go-github/blob/882755a047ea3e09b2f5f4a0e888d9a12cf2852f/github/meta.go

In the response of this API endpoint, the JSON unmarshal always seems to fail for the `Domains` field -- because the type of the value is different from what is expected.

In go-github, values under `domains` property is expected to be `map[string][]string`:

https://github.com/google/go-github/blob/882755a047ea3e09b2f5f4a0e888d9a12cf2852f/github/meta.go#L73-L75

But it actually looks like this (see https://api.github.com/meta):

```
{
  ... snip ...,
  "domains": {
    "website": [ "*.github.com", ... snip ... ],
    "codespaces": [ "*.github.com", ... snip ... ],
    "copilot": [ "*.github.com", ... snip ... ],
    "packages": [ "mavenregistryv2prod.blob.core.windows.net", ... snip ... ],
    "actions": [ "*.actions.githubusercontent.com", ... snip ... ],
    "artifact_attestations": {
      "trust_domain": "",
      "services": [ "*.actions.githubusercontent.com", ... snip ... ]
    }
  }
}
```

The values under `artifact_attestations` field clearly deviates from the `[]string` type.

This incompatibility causes `MetaService.Get()` to fail with errors like `Error: json: cannot unmarshal object into Go struct field APIMeta.domains of type []string`.

lrstanley:
Fixes: #3248.

BREAKING CHANGE: `APIMeta.Domains` changed from `map` to `struct`.

This should at least be a tiny bit less prone to breaking, as it only accounts for existing schema with that API, and future additions should silently be ignored (unless GitHub breaks the contract, which they didn't _really_ do in this scenario).

- closes #3248

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -662,6 +662,14 @@ func (a *AnalysesListOptions) GetSarifID() string {
 	return *a.SarifID
 }
 
+// GetDomains returns the Domains field.
+func (a *APIMeta) GetDomains() *APIMetaDomains {
+	if a == nil {
+		return nil
+	}
+	return a.Domains
+}
+
 // GetSSHKeyFingerprints returns the SSHKeyFingerprints map if it's non-nil, an empty map otherwise.
 func (a *APIMeta) GetSSHKeyFingerprints() map[string]string {
 	if a == nil || a.SSHKeyFingerprints == nil {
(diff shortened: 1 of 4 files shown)


gmlewis:
(No need to force-push PRs in this repo, @lrstanley - as we always squash and merge.)

gmlewis (approved):
Thank you, @lrstanley !
LGTM.

tomfeigin (approved):
LGTM

gmlewis:
Thank you, @tomfeigin !
Merging.

Issue #3264, opened 2024-09-15
'Mark Thread As Done' is missing.


danhcole:
As documented in [Mark a thread as done](https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-done), it would be very nice to be able to mark threads as done via this package.  

I'm happy to put up a PR for this (and have one working locally), but wanted to open an issue first.

gmlewis:
Thank you, @danhcole !  It is yours.

PR #3265, opened 2024-09-15
Add 'mark thread as done' functionality

danhcole:
Fixes https://github.com/google/go-github/issues/3264.

### DIFF
--- a/github/activity_notifications.go
+++ b/github/activity_notifications.go
@@ -184,7 +184,7 @@ func (s *ActivityService) MarkThreadRead(ctx context.Context, id string) (*Respo
 // GitHub API docs: https://docs.github.com/rest/activity/notifications#mark-a-thread-as-done
 //
 //meta:operation DELETE /notifications/threads/{thread_id}
-func (s *ActivityService) MarkThreadDone(ctx context.Context, id string) (*Response, error) {
+func (s *ActivityService) MarkThreadDone(ctx context.Context, id int64) (*Response, error) {
 	u := fmt.Sprintf("notifications/threads/%v", id)
 
 	req, err := s.client.NewRequest("DELETE", u, nil)
--- a/github/activity_notifications.go
+++ b/github/activity_notifications.go
@@ -178,6 +178,23 @@ func (s *ActivityService) MarkThreadRead(ctx context.Context, id string) (*Respo
 	return s.client.Do(ctx, req, nil)
 }
 
+// MarkThreadDone marks the specified thread as done.
+// Marking a thread as "done" is equivalent to marking a notification in your notification inbox on GitHub as done.
+//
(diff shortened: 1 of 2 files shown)


github/activity_notifications.go
```diff
@@ -178,6 +178,23 @@ func (s *ActivityService) MarkThreadRead(ctx context.Context, id string) (*Respo
 	return s.client.Do(ctx, req, nil)
 }
 
+// MarkThreadDone marks the specified thread as done.
+// Marking a thread as "done" is equivalent to marking a notification in your notification inbox on GitHub as done.
+//
+// GitHub API docs: https://docs.github.com/rest/activity/notifications#mark-a-thread-as-done
+//
+//meta:operation DELETE /notifications/threads/{thread_id}
+func (s *ActivityService) MarkThreadDone(ctx context.Context, id string) (*Response, error) {
```
gmlewis, line 187: According to this: https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-done
`id` is an integer, so let's please change this:

```suggestion
func (s *ActivityService) MarkThreadDone(ctx context.Context, id int64) (*Response, error) {
```

gmlewis (changes requested):
Thank you, @danhcole !
One minor tweak, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/activity_notifications.go
danhcole: Sounds good to me, I'll update the PR.  The MarkThreadRead function (which I shamelessly copied from) has the same issue, would you like me to fix it as part of this PR, or put in a separate PR to keep things clean?

gmlewis: Yes, a separate PR would be great since that will be a breaking change. Thank you!

github/activity_notifications_test.go
```diff
@@ -208,6 +208,32 @@ func TestActivityService_MarkThreadRead(t *testing.T) {
...
+	client, mux, _, teardown := setup()
+	defer teardown()
+
+	mux.HandleFunc("/notifications/threads/1", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "DELETE")
+		w.WriteHeader(http.StatusResetContent)
+	})
+
+	ctx := context.Background()
+	_, err := client.Activity.MarkThreadDone(ctx, "1")
```
gmlewis, line 221: These strings will need to be changed to `int64` for tests to pass.

danhcole: Whops, fixed and tests pass locally now.

gmlewis (approved):
Thank you, @danhcole !
LGTM.

tomfeigin (approved):
LGTM

gmlewis:
Thank you, @tomfeigin!
Merging.

PR #3272, opened 2024-09-19
Add EventType consts representing the X-Github-Event header values

MaerF0x0:
What: Adds `type EventType string`  and an `EventType` const for every of the types in the maps in `github/messages.go`

Why: So that we can do switches like

```go
switch eventType {
    case github.EventTypeFork:
        // do stuff
     case github.EventTypeDiscussionComment, github.EventTypeIssueComment:
        // do something with either
}
```

Notes: 
 * It should be non-breaking change
  * Let me know if you want the EventType used more widely including implementing `type EventTyper interface{}` on each of those structs

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -43,10 +43,6 @@ const (
 	DeliveryIDHeader = "X-Github-Delivery"
 )
 
-type EventTyper interface {
-	EventType() EventType
-}
-
 // EventType corresponds to the  X-GitHub-Event: header for Webhooks events and payloads
 type EventType string
 
--- a/github/messages.go
+++ b/github/messages.go
@@ -43,82 +43,160 @@ const (
 	DeliveryIDHeader = "X-Github-Delivery"
 )
 
+type EventTyper interface {
+	EventType() EventType
+}
+
+// EventType corresponds to the  X-GitHub-Event: header for Webhooks events and payloads
+type EventType string
+
+const (
+	EventTypeBranchProtectionRule         EventType = "branch_protection_rule"
+	EventTypeCheckRun                     EventType = "check_run"
+	EventTypeCheckSuite                   EventType = "check_suite"
+	EventTypeCodeScanningAlert            EventType = "code_scanning_alert"
+	EventTypeCommitComment                EventType = "commit_comment"
+	EventTypeContentReference             EventType = "content_reference"
+	EventTypeCreate                       EventType = "create"
+	EventTypeDelete                       EventType = "delete"
+	EventTypeDependabotAlert              EventType = "dependabot_alert"
+	EventTypeDeployKey                    EventType = "deploy_key"
+	EventTypeDeployment                   EventType = "deployment"
+	EventTypeDeploymentProtectionRule     EventType = "deployment_protection_rule"
+	EventTypeDeploymentReview             EventType = "deployment_review"
+	EventTypeDeploymentStatus             EventType = "deployment_status"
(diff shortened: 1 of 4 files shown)


github/messages.go
```diff
@@ -43,80 +43,156 @@ const (
 	DeliveryIDHeader = "X-Github-Delivery"
 )
 
+type EventTyper interface {
+	EventType() EventType
+}
+
+// EventType corresponds to the  X-GitHub-Event: header for Webhooks events and payloads
```
gmlewis, line 50:
```suggestion
// EventType corresponds to the  X-GitHub-Event: header for Webhooks events and payloads.
```

github/messages.go
```diff
@@ -43,80 +43,156 @@ const (
 	DeliveryIDHeader = "X-Github-Delivery"
 )
 
+type EventTyper interface {
+	EventType() EventType
+}
```
gmlewis, line 48: Since this is not used anywhere, let's please delete it.
If someone finds this useful in their own code, then it is a simple matter to create it there instead of here.

gmlewis (changes requested):
Thank you, @MaerF0x0!
Two minor tweaks, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

gmlewis:
Also, please make sure all tests run locally by following step 4 of CONTRIBUTING.md and push any necessary changes to this PR.

gmlewis:
Also, please resolve all conflicts and push (not force-push) the changes to this PR.  Thank you!

MaerF0x0:
@gmlewis  Thanks for the updates. I will be able to get to this in the later half of October.

gmlewis:
> @gmlewis Thanks for the updates. I will be able to get to this in the later half of October. Just wanted you to know I haven't forgotten about, or abandoned this :)

Awesome!  Thank you, @MaerF0x0, for the update!
No problem. We'll keep the PR open.

github/messages.go
MaerF0x0: abcf17e3

github/messages.go
MaerF0x0: 1fe5355e

MaerF0x0:
@gmlewis  I rebased my branch on origin/master so it required a force push. 

fixed broken test in 4aa32d2d

gmlewis (commented):
Thank you, @MaerF0x0 !

I just did a quick search for a random event string in the code base, like so:

```bash
$ rg code_scanning_alert
github/repos_hooks_deliveries_test.go
148:	"code_scanning_alert":            &CodeScanningAlertEvent{},

github/messages_test.go
281:			messageType: "code_scanning_alert",

github/messages.go
52:		"code_scanning_alert":            &CodeScanningAlertEvent{},

github/orgs_audit_log_test.go
45:		"events": ["code_scanning_alert"],
143:				"events":          []interface{}{"code_scanning_alert"},

github/event_types.go
1866:// GitHub API docs: https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#code_scanning_alert
```

and noticed that `github/messages_test.go` still spells out all of these constants explicitly, which seems odd.

I'm just thinking about when the next event gets added, the developer experience of copying that string multiple places is still not fully solved.

I think the other locations are probably fine, but `messages_test.go` should probably be updated as well.

Thoughts?

gmlewis:

gmlewis:
Closing PR as abandoned.

PR #3279, opened 2024-09-21
chore: Change golangci-lint output format due to deprecation

felixlut:
As specified in the [GitHub Actions logs for the linter](https://github.com/google/go-github/actions/runs/10973176585/job/30470274170#step:4:9), the ``github-actions`` output format is deprecated, and ``colored-line-number`` should be used instead. Deprecation was done in https://github.com/golangci/golangci-lint/pull/4726


gmlewis (approved):
Thank you, @felixlut !
LGTM.
Merging.

PR #3296, opened 2024-10-01
refactor: Do not call teardown manually in tests

gmlewis:
Closes: #3293.

Due to a non-trivial merge, this PR replaces #3293.

cc: @alexandear


github/repos_commits_test.go
```diff
@@ -380,7 +375,7 @@ func TestRepositoriesService_CompareCommits(t *testing.T) {
 	}
 
 	for _, sample := range testCases {
-		client, mux, _, teardown := setup()
+		client, mux, _ := setup(t)
```
gmlewis, line 378: @alexandear - I'm noticing in this test that `setup(t)` is getting called in a loop which is definitely not what we want since the new cleanup only happens once after the entire test completes.

It looks like this test is going to need a version of `setup` that does expose the `teardown`.

Thoughts?

github/repos_commits_test.go
```diff
@@ -515,7 +508,7 @@ func TestRepositoriesService_CompareCommitsRaw_diff(t *testing.T) {
...
```
gmlewis, line 511: same here

github/repos_commits_test.go
```diff
@@ -567,7 +558,7 @@ func TestRepositoriesService_CompareCommitsRaw_patch(t *testing.T) {
...
```
gmlewis, line 561: same here

github/repos_commits_test.go
```diff
@@ -607,21 +596,19 @@ func TestRepositoriesService_CompareCommitsRaw_invalid(t *testing.T) {
 	}
 
 	for _, sample := range testCases {
-		client, _, _, teardown := setup()
+		client, _, _ := setup(t)
```
gmlewis, line 599: same here

gmlewis:
Note that I have manually checked that all commits being pulled into this PR come from already-merged PRs that have individually passed the CLA checks, so I am overriding the cla/google check and squashing and merging.

PR #3298, opened 2024-10-02
Add Webhook for Repository ruleset

unamdev0:
Added Webhook for Repository ruleset

### DIFF
--- a/github/event_types.go
+++ b/github/event_types.go
@@ -1505,6 +1505,20 @@ type RepositoryImportEvent struct {
 	Sender *User         `json:"sender,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


gmlewis (changes requested):
I notice that you are still pushing changes while I'm reviewing the code so I'm going to stop reviewing now since there are many changes that needs to be made.
Thanks.

gmlewis:
Please do _NOT_ use force-push as described in CONTRIBUTING.md, especially on large PRs like this one, since that makes it extremely difficult for the reviewer to see what has changed since the last review. Thanks.

unamdev0:
> review

Oh, I didn't I just rebased it, will make all the changes suggested and will let you know

PR #3309, opened 2024-10-05
fix: Allow removing/unsetting repository custom property

felixlut:
This change will allow a user to set `null` when calling [the create/update repository custom property endpoint](https://docs.github.com/en/enterprise-cloud@latest/rest/repos/custom-properties?apiVersion=2022-11-28#create-or-update-custom-property-values-for-a-repository). As written in the docs, doing so will remove/unset the custom property. Currently doing so is not possible

Normally I think this would count as a breaking change, since this will modify the behavior when omitting the `Value`, but I believe this is a straight up bug, as omitting the `Value` should always result in an error. So there shouldn't be any usages that's relying on this behavior. Technically it's possible, but improbable.

### DIFF
--- a/github/orgs_properties.go
+++ b/github/orgs_properties.go
@@ -42,7 +42,7 @@ type RepoCustomPropertyValue struct {
 // CustomPropertyValue represents a custom property value.
 type CustomPropertyValue struct {
 	PropertyName string      `json:"property_name"`
-	Value        interface{} `json:"value,omitempty"`
+	Value        interface{} `json:"value"`
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface.


gmlewis (approved):
Thank you, @felixlut !
LGTM.

tomfeigin (approved):

gmlewis:
Thank you, @tomfeigin !
Merging.

felixlut:
Could we get a new release with this change as well? I have some changes over the last couple of weeks that will unblock some stuff I'm working on in https://github.com/integrations/terraform-provider-github:
* https://github.com/integrations/terraform-provider-github/pull/2322
* https://github.com/integrations/terraform-provider-github/pull/2316

gmlewis:
> Could we get a new release with this change as well? 

Yes, I can work on a new release later today.

gmlewis:
> > Could we get a new release with this change as well?
> 
> Yes, I can work on a new release later today.

This has now been released here: 
https://github.com/google/go-github/releases/tag/v66.0.0

PR #3320, opened 2024-10-10
feat: Add Expired to RepositoryInvitation

milescrabill:
Used to determine whether repository invitations have expired.


gmlewis (approved):
Thank you, @milescrabill !
LGTM.
Merging.

PR #3318, opened 2024-10-09
Add Copilot Usage Summary for Organization

claystation:
Github has a Rest API to get the usage summary of Copilot within the organization. This PR implements the functionality to fetch that data.
API Docs: [Here](https://docs.github.com/en/rest/copilot/copilot-usage?apiVersion=2022-11-28#get-a-summary-of-copilot-usage-for-organization-members)

Please let me know what you think!

PR #3321, opened 2024-10-10
 Add Copilot Usage Summary for Organization

claystation:
Github has a Rest API to get the usage summary of Copilot within the organization. This PR implements the functionality to fetch that data.
API Docs: [Here](https://docs.github.com/en/rest/copilot/copilot-usage?apiVersion=2022-11-28#get-a-summary-of-copilot-usage-for-organization-members)

Please let me know what you think!

(Fixup of #3318 due to CLA signing issues)

### DIFF
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -63,6 +63,29 @@ type SeatCancellations struct {
 	SeatsCancelled int `json:"seats_cancelled"`
 }
 
+type CopilotUsageBreakdown struct {
+	Language         string `json:"language"`
+	Editor           string `json:"editor"`
+	SuggestionsCount int    `json:"suggestions_count"`
+	AcceptancesCount int    `json:"acceptances_count"`
+	LinesSuggested   int    `json:"lines_suggested"`
+	LinesAccepted    int    `json:"lines_accepted"`
+	ActiveUsers      int    `json:"active_users"`
+}
+
+type CopilotUsageSummary struct {
+	Day                   string                  `json:"day"`
+	TotalSuggestionsCount int                     `json:"total_suggestions_count"`
+	TotalAcceptancesCount int                     `json:"total_acceptances_count"`
+	TotalLinesSuggested   int                     `json:"total_lines_suggested"`
+	TotalLinesAccepted    int                     `json:"total_lines_accepted"`
+	TotalActiveUsers      int                     `json:"total_active_users"`
+	TotalChatAcceptances  int                     `json:"total_chat_acceptances"`
+	TotalChatTurns        int                     `json:"total_chat_turns"`
+	TotalActiveChatUsers  int                     `json:"total_active_chat_users"`
+	Breakdown             []CopilotUsageBreakdown `json:"breakdown"`
+}
+
 func (cp *CopilotSeatDetails) UnmarshalJSON(data []byte) error {
 	// Using an alias to avoid infinite recursion when calling json.Unmarshal
 	type alias CopilotSeatDetails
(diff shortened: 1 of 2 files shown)


claystation:
I will implement other missing API methods as well, converting to draft for now whilst i work on more!

claystation:
@gmlewis i added all missing API methods from the Github API :) Let me know what you think!

github/copilot.go
```diff
@@ -313,3 +377,107 @@ func (s *CopilotService) GetSeatDetails(ctx context.Context, org, user string) (
...
+
+	return usage, resp, nil
+}
+
+// GetEnterpriseTeamUsage gets daily breakdown of aggregated usage metrics for Copilot completions and Copilot Chat in the IDE for a team in the enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-usage?apiVersion=2022-11-28#get-a-summary-of-copilot-usage-for-an-enterprise-team
+//
+//meta:operation GET /enterprises/{enterprise}/team/{team_slug}/copilot/usage
+func (s *CopilotService) GetEnterpriseTeamUsage(ctx context.Context, enterprise string, team string, opts *CopilotUsageSummaryListOptions) ([]*CopilotUsageSummary, *Response, error) {
```
gmlewis, line 438:
```suggestion
func (s *CopilotService) GetEnterpriseTeamUsage(ctx context.Context, enterprise, team string, opts *CopilotUsageSummaryListOptions) ([]*CopilotUsageSummary, *Response, error) {
```

github/copilot.go
```diff
@@ -313,3 +377,107 @@ func (s *CopilotService) GetSeatDetails(ctx context.Context, org, user string) (
...
+
+	return usage, resp, nil
+}
+
+// GetOrganizationTeamUsage gets daily breakdown of aggregated usage metrics for Copilot completions and Copilot Chat in the IDE for a team in the organization.
+//
+// GitHub API docs: https://docs.github.com/en/rest/copilot/copilot-usage?apiVersion=2022-11-28#get-a-summary-of-copilot-usage-for-a-team
+//
+//meta:operation GET /orgs/{org}/team/{team_slug}/copilot/usage
+func (s *CopilotService) GetOrganizationTeamUsage(ctx context.Context, org string, team string, opts *CopilotUsageSummaryListOptions) ([]*CopilotUsageSummary, *Response, error) {
```
gmlewis, line 464:
```suggestion
func (s *CopilotService) GetOrganizationTeamUsage(ctx context.Context, org, team string, opts *CopilotUsageSummaryListOptions) ([]*CopilotUsageSummary, *Response, error) {
```

github/copilot.go
```diff
@@ -63,6 +65,39 @@ type SeatCancellations struct {
...
+type CopilotUsageSummary struct {
+	Day                   string                   `json:"day"`
+	TotalSuggestionsCount int                      `json:"total_suggestions_count"`
+	TotalAcceptancesCount int                      `json:"total_acceptances_count"`
+	TotalLinesSuggested   int                      `json:"total_lines_suggested"`
+	TotalLinesAccepted    int                      `json:"total_lines_accepted"`
+	TotalActiveUsers      int                      `json:"total_active_users"`
+	TotalChatAcceptances  int                      `json:"total_chat_acceptances"`
+	TotalChatTurns        int                      `json:"total_chat_turns"`
+	TotalActiveChatUsers  int                      `json:"total_active_chat_users"`
```
gmlewis, line 97: I'm wondering if any of these `int` fields might benefit from being `int64`, but then I think that that would have to be some crazy Copilot usage to overflow a 32-bit `int`.

So if you want to leave them all as `int` that is fine with me... I just wanted to throw that out there.

gmlewis (changes requested):
Thank you, @claystation !
A couple minor tweaks, please, and one question.

After that, we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

gmlewis:
Also, please run `go generate ./...` as discussed in step 4 of CONTRIBUTING.md and push (not force-push) the changes to this PR.  Thanks!

github/copilot.go
claystation: Valid question. I assumed `int` is taken as a 64 bit int as i `expected` all systems to be 64 bit.

```
The int, uint, and uintptr types are usually 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems. When you need an integer value you should use int unless you have a specific reason to use a sized or unsigned integer type. 
```

But, obviously this does not have to be true. Lets be on the safe side and just make it int64 as it will most of the time be that anyway!

github/copilot.go
gmlewis: This hasn't been resolved yet.

claystation: I was confused! You are correct, will fix!

gmlewis (approved):
Thank you, @claystation !
LGTM.

tomfeigin (approved):
LGTM

gmlewis:
Thank you, @tomfeigin !
Merging.

Issue #3328, opened 2024-10-18
Unable to filter artifacts by name in golang


muhardanz:
Unable to filter artifacts by name in golang when listing them using the package: github.com/google/go-github/v43/github

The github.Client ->Actions.ListArtifacts() only has ListOptions to support pagination. There is no reference to filter via the name query parameter

The only way to do this is to use the API directly via:
repos/<<OWNER>>/<<REPO>>/actions/artifacts?name=<<ARTIFACT_NAME>>
which does support filtering by name

gmlewis:
Can you please give an example using `curl` semantics (without including any secrets, obviously)?

Have you tried the most recent version out of curiosity?

muhardanz:
I just tried the following package instead
`github.com/google/go-github/v66/github`
This also doesn't seem allow filter by the name query string

This is the curl request that allows name filtering:
```
curl --location --request GET 'https://api.github.com/repos/foo-owner/foo-repo/actions/artifacts?name=foo' \
--header 'Authorization: Bearer <<MY_TOKEN>>'
```
This is the same api URL used by the following function: [ListArtifacts](https://github.com/google/go-github/blob/master/github/actions_artifacts.go#L58)

But I cannot see a way to pass the 'name' query string to the 'ListArtifacts' function

gmlewis:
Ah!  I see what you are saying now, thank you for the details!  We need a breaking API change to modify `opts` to a custom struct that adds `name` to `ListOptions`.

I'll throw together a PR and send it to you for review.

muhardanz:
I put in these changes locally and tested them. I can confirm that this fixes the ListArtifacts function so that it can take in the name query string. I can also confirm that it returns the correct result

PR #3330, opened 2024-10-21
feat!: Add name query parameter to ActionsService.ListArtifacts

gmlewis:
Fixes: #3328.

BREAKING CHANGE: `opts` argument to `ActionsService.ListArtifacts` changed from `ListOptions` to `ListArtifactsOptions`.

### DIFF
--- a/github/actions_artifacts.go
+++ b/github/actions_artifacts.go
@@ -50,12 +50,22 @@ type ArtifactList struct {
 	Artifacts  []*Artifact `json:"artifacts,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


gmlewis:
@muhardanz - can you please take a look at this (and possibly try it out) to see if it fixes the issue you were having?

muhardanz:
@gmlewis I can confirm that this fixes the ListArtifacts function so that it can take in the name query string. I can also confirm that it returns the correct result

gmlewis:
Thank you, @muhardanz !
Merging.

PR #3258, opened 2024-09-08
Add type SponsorshipEvent

CodeDiego15:
### Description

This pull request adds support for the `SponsorshipEvent` to the **go-github** library, addressing the issue described in [#2982](https://github.com/google/go-github/issues/2982).

#### Details

According to GitHub documentation ([SponsorshipEvent](https://docs.github.com/en/rest/overview/github-event-types?apiVersion=2022-11-28#sponsorshipevent)), the `SponsorshipEvent` is triggered when sponsorship changes occur, such as creation, cancellation, or tier changes.

This PR introduces the following changes:
- **New `SponsorshipEvent` struct**: Captures the key properties of a sponsorship event, including:
  - `Action`: The type of action performed (e.g., `created`, `tier_changed`, `cancelled`, etc.).
  - `EffectiveDate`: The date when the event or change becomes effective, applicable for events like `pending_cancellation` and `pending_tier_change`.
  - `Changes`: An optional field that captures details about changes, such as adjustments in sponsorship tier or privacy. This is represented by an embedded struct called `SponsorshipChanges`.
  - Standard fields like `Repository`, `Organization`, `Sender`, and `Installation` are also included to maintain consistency with other GitHub event types.
  
- **New `SponsorshipChanges` struct**: This struct provides additional details when the `tier` or `privacy_level` of a sponsorship is modified.
  - The `Tier` field captures tier changes using an embedded `SponsorshipTier` struct, which contains a `From` field representing the previous tier.
  - The `PrivacyLevel` field tracks any changes to the sponsorship’s privacy settings.

This new event type will allow **go-github** users to programmatically handle sponsorship-related events triggered on GitHub, improving support for GitHub Sponsors workflows.

#### Testing
- Unit tests have been added to verify the correct deserialization of the `SponsorshipEvent` struct from JSON payloads.
- The tests cover various event scenarios, including tier changes, privacy level changes, and sponsorship creation.

#### Issue Link
Resolves [#2982](https://github.com/google/go-github/issues/2982)

PR #3332, opened 2024-10-23
Add BranchProtectionConfigurationEvent and SecretScanningAlertLocationEvent

maditya:


gmlewis (changes requested):
Thank you, @maditya.
To add new events, we need to make some more changes to the following files:
* event_types_test.go
* messages.go
* messages_test.go

Please see #3258 for an example of the changes that are needed.
Thanks again!

maditya:
```
--- FAIL: TestProjectsService_DeleteProjectCard (0.00s)
    projects_test.go:611: Projects.DeleteProjectCard returned error: Delete "http://127.0.0.1:39961/api-v3/projects/columns/cards/1": net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called
```
@gmlewis this seems like an intermittent failure unrelated to the changes in this PR. Unable to reproduce it locally. Can we re-run the workflow?

github/orgs_codesecurity_configurations_test.go
```diff
@@ -382,7 +382,6 @@ func TestOrganizationsService_SetDefaultCodeSecurityConfiguration(t *testing.T)
 					"code_scanning_default_setup": "enabled"
 				}
 		}`)
-		w.WriteHeader(http.StatusOK)
```
gmlewis, line 385: The changes to this file seem wrong to me.
I think you can revert the changes to this file.

maditya: This was a warning which I fixed.

Per https://pkg.go.dev/net/http 

```
If WriteHeader is not called explicitly, the first call to Write
will trigger an implicit WriteHeader(http.StatusOK).
```

So I think we should either keep this change or explicitly call `w.WriteHeader(http.StatusOK)` before we write to `w`. Any thoughts?

gmlewis: Oh, interesting, so this is unrelated to this PR.  OK.  I'm fine to keep the change.  Thanks.

gmlewis (approved):
Thank you, @maditya !
LGTM.

tomfeigin (approved):
LGTM

gmlewis:
Thank you, @tomfeigin !
Merging.

PR #3250, opened 2024-08-26
Add Source field to RulesetRepositoryPropertyTargetParameters

rezmoss:
Fixes: #3247.

PR #3333, opened 2024-10-24
feat!: Fix `source` property on Repo Custom Properties

claystation:
BREAKING CHANGE: Change `RulesetRepositoryPropertyTargetParameters.Source` from `string` to `*string`.

Our application was failing to apply or update Github Rulesets because of the following error:

```
Invalid property /conditions: data matches no possible input. See `documentation_url`. []
```

Following the release of Go Github v65. Digging into this we saw a change in the following PR #3250.
Looking into the Github REST API the `source` attribute is not listed as required. Thus we also did not implement this on our side. The problem that surfaced was the fact that the property was still being sent to the Github API even though it was empty. 

With this PR we make it optional, a pointer, and i adapted test cases to verify that it is indeed missing when empty.
I tested this with our own fork of Go Github and it works for our case again.

Let me know what you think!

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -21270,6 +21270,14 @@ func (r *RulesetRepositoryNamesConditionParameters) GetProtected() bool {
 	return *r.Protected
 }
 
+// GetSource returns the Source field if it's non-nil, zero value otherwise.
+func (r *RulesetRepositoryPropertyTargetParameters) GetSource() string {
+	if r == nil || r.Source == nil {
+		return ""
+	}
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @claystation, and I apologize for the breakage, as I should have caught that.
LGTM.

tomfeigin (approved):

gmlewis:
Thank you, @tomfeigin !
Merging.

(The test breakage appears to be unrelated to this PR - I will investigate and fix in a separate PR.)

PR #3334, opened 2024-10-24
Add support for fetching artifact attestations at the org, repo and user level

gillisandrew:
[GitHub artifact attestations is now generally available](https://github.blog/changelog/2024-06-25-artifact-attestations-is-generally-available/)

This pull request adds support for listing attestations hosted on GitHub for a given subject digest. Those attestations can then be verified using another library such as [sigstore-go](https://github.com/sigstore/sigstore-go). I've included a basic demo as an example.


example/go.mod
```diff
@@ -1,26 +1,103 @@
...
 
 require (
 	github.com/ProtonMail/go-crypto v0.0.0-20230828082145-3c4c8a2d2371
 	github.com/bradleyfalzon/ghinstallation/v2 v2.0.4
 	github.com/gofri/go-github-ratelimit v1.0.3
 	github.com/google/go-github/v66 v66.0.0
-	golang.org/x/crypto v0.21.0
-	golang.org/x/term v0.18.0
-	google.golang.org/appengine v1.6.7
+	github.com/sigstore/sigstore-go v0.6.2
```
gmlewis, line 12: Why is `github.com/sigstore/sigstore-go` pulling in an enormous boat load of dependencies below with it?

example/verifyartifact/main.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore-go/pkg/verify"
+)
+
+var (
+	owner                   = flag.String("owner", "cli", "GitHub organization or user to scope attestation lookup by")
+	artifactDigest          = flag.String("artifact-digest", "2ce2e480e3c3f7ca0af83418d3ebaeedacee135dbac94bd946d7d84edabcdb64", "digest of the artifact")
+	artifactDigestAlgorithm = flag.String("artifact-digest-algorithm", "sha256", "The algorithm used to compute the digest of the artifact")
+	expectedIssuer          = flag.String("expected-issuer", "https://token.actions.githubusercontent.com", "Issuer of the OIDC token")
+	expectedSAN             = flag.String("expected-san", "https://github.com/cli/cli/.github/workflows/deployment.yml@refs/heads/trunk", "The expected Subject Alternative Name (SAN) of the certificate used to sign the attestation")
```
gmlewis, line 33: How does a user know where to get this?

example/verifyartifact/main.go
```diff
@@ -0,0 +1,165 @@
...
+	trustedRootJSONPath     = flag.String("trusted-root-json-path", "verifyartifact/trusted-root-public-good.json", "Path to the trusted root JSON file")
```
gmlewis, line 34: How is this file generated?

github/attestations.go
```diff
@@ -0,0 +1,22 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"encoding/json"
+)
+
+type Attestation struct {
```
gmlewis, line 12: All exported structs need Godoc-styled comments for the auto-generated docs with complete sentences and punctuation.

github/attestations.go
```diff
@@ -0,0 +1,22 @@
...
+	// The attestation's Sigstore Bundle.
+	// Refer to the sigstore bundle specification for more info:
+	// https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto
+	Bundle       *json.RawMessage `json:"bundle"`
+	RepositoryID *int             `json:"repository_id"`
```
gmlewis, line 17:
```suggestion
	RepositoryID *int64             `json:"repository_id"`
```

github/attestations.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+type AttestationsResponse struct {
```
gmlewis, line 20: Is there any reason why we wouldn't want to un-export this struct and return `[]*Attestation` as the first value from the newly-added methods?

I'm fine either way, it just seems like returning a slice might be more useful to end-users of this client library.

gmlewis (changes requested):
Thank you, @gillisandrew!
A few questions and tweaks, please.

example/go.mod
gillisandrew: The library does a lot, including signing, hashing, provisioning of short-lived certs etc. If you'd prefer not to pull in so many indirect deps in the example module I can break it out into its own go module or host the example code in a different repo.

example/verifyartifact/main.go
gillisandrew: It's written with the assumption that you've had some introduction to [how sigstore works](https://www.sigstore.dev/how-it-works).  GitHub presents these values online ([the code references this attestation](https://github.com/cli/cli/attestations/2543768)).

example/verifyartifact/main.go
gillisandrew: The json file is a snapshot of the public good trusted root hosted by sigstore that underlies the fulcio CA that GitHub uses to sign the attestations in public repos. 

https://github.com/sigstore/root-signing
https://github.com/sigstore/fulcio

github/attestations.go
gillisandrew: Ok, I'll add some.

example/verifyartifact/main.go
gmlewis: Perfect answer... thank you!
Can you please embed these answers (as `// comments`) in this file?

example/verifyartifact/main.go
gmlewis: Thank you!  This would also be an extremely valuable comment to add to this file.

github/attestations.go
gillisandrew: I had considered it, however its a brand new endpoint and GitHub saw fit to return it as a JSON object instead of an array so it felt like unwrapping it might result in unnecessary breaking changes if GitHub starts including more data in the response.

gmlewis: OK, that makes total sense. Let's keep it then.

github/attestations.go
```diff
@@ -0,0 +1,27 @@
...
+// Attestation represents an artifact attestation associated with a repository.
+// The provided bundle can be used to verify the provenance of artifacts.
+//
+// https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds
+type Attestation struct {
+	// The attestation's Sigstore Bundle.
+	// Refer to the sigstore bundle specification for more info:
+	// https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto
+	Bundle       *json.RawMessage `json:"bundle"`
+	RepositoryID *int64           `json:"repository_id"`
```
gmlewis, line 21: `json.RawMessage` is already a reference type, so the pointer is not necessary here.
Come to think of it, the `*int64` isn't necessary either. Sorry I didn't realize this before.

```suggestion
	Bundle       json.RawMessage `json:"bundle"`
	RepositoryID int64           `json:"repository_id"`
```

gmlewis (changes requested):
Thank you, @gillisandrew .
Please add unit tests for the newly-added methods.

github/orgs_attestations_test.go
```diff
@@ -0,0 +1,58 @@
...
+				RepositoryID: 2,
+				Bundle:       []byte(`{}`),
+			},
+		},
+	}
+
+	if !cmp.Equal(attestations, want) {
+		t.Errorf("Organizations.ListAttestations = %+v, want %+v", attestations, want)
+	}
+}
```
gmlewis, line 58: Please do similarly to this in all new unit tests in order to increase code coverage.

```suggestion
	const methodName = "ListAttestations"

	testBadOptions(t, methodName, func() (err error) {
		_, _, err = client.Organizations.ListAttestations(ctx, "\n", "\n", &ListOptions{})
		return err
	})

	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
		got, resp, err := client.Organizations.ListAttestations(ctx, "o", "digest", nil)
		if got != nil {
			t.Errorf("testNewRequestAndDoFailure %v = %#v, want nil", methodName, got)
		}
		return resp, err
	})
}
```

gmlewis: Please do similarly to this in all new unit tests in order to increase code coverage.

gmlewis (approved):
Thank you, @gillisandrew !
LGTM.

gillisandrew:
@gmlewis Great! Thanks for your help and feedback, really appreciate it.

tomfeigin (approved):
Nice!

gmlewis:
Thank you, @tomfeigin !
Merging.

PR #3350, opened 2024-11-18
feat: Support Copilot metrics endpoints

henriklundstrom:
Introduces structs and functions for calling the newly GA endpoints for Copilot metrics. See this post from GitHub https://github.blog/changelog/2024-10-30-github-copilot-metrics-api-ga-release-now-available/.


github/copilot.go
```diff
@@ -99,6 +99,116 @@ type CopilotUsageSummary struct {
...
 
+// CopilotMetricsListOptions represents the optional parameters to the CopilotService get metrics methods.
+type CopilotMetricsListOptions struct {
+	Since *time.Time `url:"since,omitempty"`
+	Until *time.Time `url:"until,omitempty"`
+
+	ListOptions
+}
+
+type CopilotIdeCodeCompletionsLanguage struct {
```
claystation, line 110: Please add comments for each struct explaining the struct please.

github/copilot.go
```diff
@@ -99,6 +99,116 @@ type CopilotUsageSummary struct {
...
+type CopilotDotcomPullRequests struct {
+	TotalEngagedUsers int                                    `json:"total_engaged_users"`
+	Repositories      []*CopilotDotcomPullRequestsRepository `json:"repositories"`
+}
+
+// CopilotMetrics represents Copilot usage metrics for a given day.
+type CopilotMetrics struct {
+	Date                      string                     `json:"date"`
+	TotalActiveUsers          int                        `json:"total_active_users"`
+	TotalEngagedUsers         int                        `json:"total_engaged_users"`
```
henriklundstrom, line 205: According to the https://docs.github.com/en/rest/copilot/copilot-metrics?apiVersion=2022-11-28 response schema, these two are not required. Should I change them to be pointers to represent the case where they're omitted as nil?

gmlewis: Yes, please. Anything that is optional should be a pointer type with `omitempty`.

gmlewis (changes requested):
Thank you, @henriklundstrom !
First, there are a few acronyms that need to be fixed for idiomatic Go naming conventions.

Then, before continuing my review, it seems like there is a lot of duplication in the new set of structs... are there structs (potential candidates are the ones utilizing `omitempty`) that could possibly be shared, or is there enough distinction to warrant the current number of structs?

I'm fine if the answer is the latter - I just wanted your opinion to see if you think we could take advantage of simplifying the interface by a bit of sharing (like we do with other structs). But again, if not, no big deal... then no changes necessary.

henriklundstrom:
> Thank you, @henriklundstrom ! First, there are a few acronyms that need to be fixed for idiomatic Go naming conventions.
> 
> Then, before continuing my review, it seems like there is a lot of duplication in the new set of structs... are there structs (potential candidates are the ones utilizing `omitempty`) that could possibly be shared, or is there enough distinction to warrant the current number of structs?
> 
> I'm fine if the answer is the latter - I just wanted your opinion to see if you think we could take advantage of simplifying the interface by a bit of sharing (like we do with other structs). But again, if not, no big deal... then no changes necessary.

@gmlewis I've gone through your renaming suggestions and applied them all.

I agree there are many structs being introduced and that they're somewhat repetitive, but still not actually modular, and it's not a good read... Unfortunately this is how GitHub has structured the response schema. In this case I don't see that using struct embedding or reusable structs will make it any better. Possibly, I could introduce some:
```
type CopilotModelBase struct {
	Name                    string  `json:"name"`
	IsCustomModel           bool    `json:"is_custom_model"`
	CustomModelTrainingDate *string `json:"custom_model_training_date,omitempty"`
}
```
That would cut down the size of the `CopilotIDECodeCompletionsModel`, `CopilotIDEChatModel`, `CopilotDotcomChatModel`, and `CopilotDotcomPullRequestsModel` structs, but not by much and I'm not sure it makes them any nicer to work with.

gmlewis:
@henriklundstrom - to proceed with this PR, it looks like you need to do the following steps:

* `git checkout master`
* `git pull upstream master` (or however you set up your repo to pull from the upstream repo)
* `git checkout nordnet:copilot-metrics`
* `git merge master`
* `./script/generate.sh` (see step 4 of CONTRIBUTING.md for more details)
* `git push origin nordnet:copilot-metrics`

This should pull #3352 into your PR which has the latest OpenAPI changes needed to support `copilot-metrics`.

gmlewis (approved):
Thank you, @henriklundstrom !
LGTM.

claystation (approved):
LGTM :)

henriklundstrom:
@gmlewis I've updated the branch and we have a second approval now 🚀

gmlewis:
Thank you, @claystation !
Merging.

PR #3353, opened 2024-11-19
Add missing Deployment field to workflow_job event type

AirOne70:
The `workflow_job` webhook event as described by GitHub documentation contains a `deployment` object. 
https://docs.github.com/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_job

However, the struct that defines the `workflow_job `webhook event type is missing the `deployment` object . This PR adds the missing object to the `WorkflowJobEvent` type.


gmlewis (approved):
Thank you, @AirOne70 !
LGTM.
Merging.

PR #3357, opened 2024-11-25
feat: Support resolution_comment to update alert API

na-ga:
Support to an optional comment when closing an alert

Per the documentation available here:
https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#update-a-secret-scanning-alert

### DIFF
--- a/github/secret_scanning.go
+++ b/github/secret_scanning.go
@@ -91,8 +91,6 @@ type SecretScanningAlertUpdateOptions struct {
 	Resolution *string `json:"resolution,omitempty"`
 
 	// An optional comment when closing an alert.
-	// Cannot be updated or deleted.
-	// Must be null when changing state to open.
 	ResolutionComment *string `json:"resolution_comment,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


github/secret_scanning.go
```diff
@@ -89,6 +89,11 @@ type SecretScanningAlertUpdateOptions struct {
 	// Required when the state is "resolved" and represents the reason for resolving the alert.
 	// Can be one of: "false_positive", "wont_fix", "revoked", or "used_in_tests".
 	Resolution *string `json:"resolution,omitempty"`
+
+	// An optional comment when closing an alert.
+	// Cannot be updated or deleted.
+	// Must be null when changing state to open.
```
na-ga, line 95: I copied the comment from the documentation. 
https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#update-a-secret-scanning-alert

However, I have now confirmed that comments can be added when changing to Open, 
so this comment may not be necessary.

na-ga: For me, it is better to be able to specify comments when changing to close state or open state.

When changing to open state, I think the comment should also be modified to match the actual behavior that can be specified as optional.

Therefore, how about removing the comment “Cannot be updated or deleted. Must be null when changing state to open.”

gmlewis: Sounds good to me.

gmlewis (approved):
Thank you, @na-ga and @Alexizx00789999 .
LGTM.
Merging.

PR #3369, opened 2024-12-06
Fix: Preserve HTTP Response in URL Errors

### LINKED ISSUE #3362: Propagate *http.Response when encountering a *url.Error?
Why does [this code](https://github.com/google/go-github/blame/master/github/github.go#L853-L861) return nil for the response? I'm getting 404 errors (which can be permissions errors) but I'm not finding a good way to programmatically get the status code directly from the error object.

silasalberti:
Fix: Preserve HTTP Response in URL Errors

Fixes #3362

This PR improves error handling in the BareDo method to properly preserve HTTP response information when encountering *url.Error. Changes include:

- Move response creation earlier to avoid code duplication
- Add proper nil response handling
- Preserve HTTP response information in URL errors
- Add test case TestDo_preservesResponseInURLError

The changes allow users to access HTTP status codes (like 404) even when encountering URL errors.

This PR was created with help from Devin: https://preview.devin.ai/sessions/1b2f7ce6e3b44942b3ac1f518eac7c22

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -1109,21 +1109,45 @@ func TestDo_redirectLoop(t *testing.T) {
 func TestDo_preservesResponseInHTTPError(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
+
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		http.Error(w, "Not Found", http.StatusNotFound)
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprintf(w, `{
+			"message": "Resource not found",
+			"documentation_url": "https://docs.github.com/rest/reference/repos#get-a-repository"
+		}`)
 	})
 
 	req, _ := client.NewRequest("GET", ".", nil)
-	resp, err := client.Do(context.Background(), req, nil)
+	var resp *Response
+	var data interface{}
+	resp, err := client.Do(context.Background(), req, &data)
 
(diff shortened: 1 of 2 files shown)


github/github_test.go
```diff
@@ -1106,6 +1106,27 @@ func TestDo_redirectLoop(t *testing.T) {
 	}
 }
 
+func TestDo_preservesResponseInURLError(t *testing.T) {
```
gmlewis, line 1109: I'm not convinced that this test actually creates a `url.Error`.
For examples, please see: https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/net/url/url_test.go;l=1763-1821

gmlewis (changes requested):
Thank you, @silasalberti !
This looks like a great start. I think the unit test can be improved, however.

Additionally, on line 854 I think it is also a safe location to return the new `response` instead of `nil`.

github/github_test.go
silasalberti: That's a good point. I guess the original issue was about a 404 error. Renamed the test to HTTPError and made the mock more in line with the other mocks in the file. Also, now it's returning response on line 854 too. Does that look better?

gmlewis (approved):
Ideally, tests would cover each of the return paths through `BareDo` but I'm thinking that this may be much more effort to achieve... so I think I'm fine with this level of testing.

LGTM.

tomfeigin (approved):
LGTM

gmlewis:
Thank you, @tomfeigin !
Merging.

PR #3371, opened 2024-12-09
feat: Add assignment for org role associations

### LINKED ISSUE #3364: Add support for `assignment` on teams and users assigned to organization roles
When listing teams or users assigned to an organization role the ~~`assigned`~~ `assignment` field isn't getting set and the team `parent` and user `inherited_from` are also not getting set. I assume that this is because these endpoints are being mapped back to `Team` and `User` structs instead of being a custom type.

stevehipwell:
Resolves #3364

This PR adds the `Assignment` field to `Team` & `User` and the `InheritedFrom` field to `User` when accessed in the context of an organization role.

### DIFF
--- a/github/users.go
+++ b/github/users.go
@@ -70,6 +70,14 @@ type User struct {
 	// repository. These are only populated when calling Repositories.ListCollaborators.
 	Permissions map[string]bool `json:"permissions,omitempty"`
 	RoleName    *string         `json:"role_name,omitempty"`
+
+	// Assignment identifies how a user was assigned to an organization role. Its
+	// possible values are: "direct", "indirect", "mixed". This is only populated when
+	// calling the ListUsersAssignedToOrgRole method.
+	Assignment *string `json:"assignment,omitempty"`
+	// InheritedFrom identifies the team that a user inherited their organization role
+	// from. This is only populated when calling the ListUsersAssignedToOrgRole method.
+	InheritedFrom *Team `json:"inherited_from,omitempty"`
 }
 
 func (u User) String() string {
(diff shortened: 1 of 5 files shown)


gmlewis (approved):
Thank you, @stevehipwell !
LGTM.
Merging.

github/users.go
```diff
@@ -70,6 +70,14 @@ type User struct {
...
 	Permissions map[string]bool `json:"permissions,omitempty"`
 	RoleName    *string         `json:"role_name,omitempty"`
+
+	// Assignment identifies how a user was assigned to an organization role. Its
+	// possible values are: "direct", "indirect", "mixed". This is only populated when
+	// calling the ListUsersAssignedToOrgRole method.
+	Assignment *string `json:"assignment,omitempty"`
+	// InheritedFrom identifies the team that a user inherited their organization role
+	// from. This is only populated when calling the ListUsersAssignedToOrgRole method.
+	InheritedFrom *Team `json:"inherited_from,omitempty"`
```
tomfeigin, line 80: @stevehipwell the `inherited_from` field returns for me as an array which fails the unmarshalling, are you sure you got only a single team in the API result?

stevehipwell: I think the schema might have changed as it's now looks like an array, but I'm pretty sure I tested this.

tomfeigin: I want to change this field to an array can you verify it doesn't break anything for you? I want to know if we need to support both cases

tomfeigin: @stevehipwell I opened this: https://github.com/google/go-github/pull/3460

PR #3373, opened 2024-12-09
chore: Deprecate legacy ID-based team methods

### LINKED ISSUE #3368: Mark teams functionality that's based on ID as deprecated
The teams functions that use ID should be deprecated as the REST APIs have been marked a legacy and are under a closing down notice ([ref](https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#get-a-team-legacy)).

stevehipwell:
Resolves #3368

This PR makes all legacy ID based team methods as deprecated based on the [closing down notice](https://docs.github.com/en/enterprise-cloud@latest/rest/teams/teams?apiVersion=2022-11-28#get-a-team-legacy).

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -106,6 +106,8 @@ func (s *TeamsService) ListTeams(ctx context.Context, org string, opts *ListOpti
 
 // GetTeamByID fetches a team, given a specified organization ID, by ID.
 //
+// Deprecated: Use GetTeamBySlug instead.
+//
 // GitHub API docs: https://docs.github.com/rest/teams/teams#get-a-team-by-name
 //
 //meta:operation GET /orgs/{org}/teams/{team_slug}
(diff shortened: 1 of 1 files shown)


stevehipwell:
@gmlewis the metadata generation doesn't look right. I manually updated all of the metadata (before I knew about the script) to fix the incorrect docs and URL templates for the legacy API calls but generate is putting them back.

Just a nit but is there a reason why `//meta:` isn't `// meta:`?

stevehipwell:
@gmlewis but what about the issue with linking to the incorrect docs?

https://github.com/google/go-github/blob/master/github/teams.go#L107...L112

gmlewis:
> @gmlewis but what about the issue with linking to the incorrect docs?

The links are managed by the official GitHub OpenAPI specification, and it periodically needs to be updated (which was last done here: https://github.com/google/go-github/pull/3352 ). I have not done a deep dive into this issue yet, but it may need updating again to fix the issue.

gmlewis:
I've gone ahead and created #3374. I'm not sure if that will help your PR yet or not.

gmlewis:
Once it is merged, you can update your local `master` branch with it, then merge `master` into this PR to see if it fixes the problems you are seeing.

stevehipwell:
@gmlewis I don't think #3374 will help as the docs that should be linked are already present (search for _-legacy_ at the bottom of the diff).

gmlewis (approved):
Thank you, @stevehipwell !
LGTM.
Merging.

PR #3386, opened 2024-12-13
Permit toggling rate limit check by consumers

manicminer:
Export `BypassRateLimitCheck` to allow toggling of rate limit checks by consumers.

I am handling rate-limited retries with my own RoundTripper, and the built-in checks are preventing me from doing so because it simply errors early instead of attempting a request and handling backoff separately.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -801,7 +801,7 @@ func parseTokenExpiration(r *http.Response) Timestamp {
 type requestContext uint8
 
 const (
-	bypassRateLimitCheck requestContext = iota
+	BypassRateLimitCheck requestContext = iota
 	SleepUntilPrimaryRateLimitResetWhenRateLimited
 )
 
@@ -822,7 +822,7 @@ func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, erro
 
 	rateLimitCategory := GetRateLimitCategory(req.Method, req.URL.Path)
 
-	if bypass := ctx.Value(bypassRateLimitCheck); bypass == nil {
+	if bypass := ctx.Value(BypassRateLimitCheck); bypass == nil {
 		// If we've hit rate limit, don't make further requests before Reset time.
 		if err := c.checkRateLimitBeforeDo(req, rateLimitCategory); err != nil {
 			return &Response{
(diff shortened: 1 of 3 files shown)


gmlewis (changes requested):
Thank you, @manicminer !
Right before line 804 in `github/github.go`, could you please add a comment that demonstrates how an end-user would override this feature in their own code?

DiegoDev2 (commented):
hi, did you update the unit tests [here](https://github.com/google/go-github/blob/master/github/rate_limit_test.go) ? @manicminer

DiegoDev2:
> hi, did you update the unit tests [here](https://github.com/google/go-github/blob/master/github/rate_limit_test.go) ? @manicminer
I'm sorry, I got confused

manicminer:
@DiegoDev2 Thanks for the review! I've added a go doc comment for this constant.

gmlewis (approved):
Thank you, @manicminer !
LGTM.

github/github.go
```diff
@@ -801,7 +801,11 @@ func parseTokenExpiration(r *http.Response) Timestamp {
 type requestContext uint8
 
 const (
-	bypassRateLimitCheck requestContext = iota
+	// BypassRateLimitCheck prevents a pre-emptive check for exceeded primary rate limits
+	// Specify this by providing a context with this key, e.g.
+	//   context.WithValue(context.Background(), github.BypassRateLimitCheck, true)
+	BypassRateLimitCheck requestContext = iota
```
tomfeigin, line 807: Why is `BypassRateLimitCheck` exported? Can't it remain `bypassRateLimitCheck`?

tomfeigin: Oh I missed that you wanted to trigger this from outside the library.
I suggest that there will be wrapper functions like:

```golang
func ContextWithBypassRateLimitCheck(ctx context.Context, bypass bool) context.Context {
    context.WithValue(ctx, BypassRateLimitCheck, bypass)
}
```

You don't even need to export the getter function for this value.

Anyway, as the code is now is fine

tomfeigin (approved):

manicminer:
Thanks for the reviews! Could this be merged? It looks like the codecov check is a false negative / unrelated to this PR?

SCR-20250107-lqsa

gmlewis:
Thank you, @tomfeigin and @manicminer !
Merging.

Issue #2898, opened 2023-08-24
Is NewClientWithEnvProxy necessary?


WillAbides:
I suspect that `NewClientWithEnvProxy` is redundant with `NewClient(nil)`. The reason I think this is `NewClient(nil)` uses returns a `Client` with an empty `http.Client`. `http.Client` with no transport uses `http.DefaultTransport`, and `http.DefaultTransport` is [configured to use ProxyFromEnvironment](https://github.com/golang/go/blob/f0b05092fd9bbfc698cb3bd24c6ae042e9c54616/src/net/http/transport.go#L38-L54). So, `NewClient(nil)` returns a client that has the same proxy config as `NewClientWithEnvProxy`.

There are a couple of differences though:

1. `NewClientWithEnvProxy`'s client uses a transport that is missing all the other config in `http.DefaultTransport`. While somebody might want this, I don't think it makes sense to get that out of a constructor named "NewClientWithEnvProxy"

2. Somebody could modify `http.DefaultTransport` so that it is unsuitable for use with go-github but then want to use an empty `*http.Transport` configured with ProxyFromEnvironment for go-github. This seems like an unlikely scenario, and in the case it comes up they can pass an http.Client to NewClient that is configured however they want.

This came up when working on #2897. If we agree that NewClientWithEnvProxy is unnecessary I can mark it deprecated with a note saying to use `github.NewClient(nil)` instead.

gmlewis:
So, unfortunately, I'm the one [who proposed it here](https://github.com/google/go-github/pull/2686#issuecomment-1456501390) as a workaround.

If, after reading #2363 and #2686, you are still convinced that `NewClientWithEnvProxy` is unnecessary, I'm fine to deprecate it as you said with a clear explanation as to how to solve those two issues appropriately.

nejch:
FWIW, I stumbled across this coming from https://github.com/integrations/terraform-provider-github/issues/1509, and I don't think this is really solved without the use of `NewClientWithEnvProxy`, at least configuring the environment variables didn't seem to help so far, but will do some more testing - or if anyone else has tried this feel free to add your comments.

JamBalaya56562:
Quick update for anyone landing here, since the API has changed: in v88 the constructor was reworked into the options pattern (#4201), so `NewClientWithEnvProxy` is gone and there's a `WithEnvProxy()` option in its place. A few notes covering the cases discussed above:

[...]

PR #3393, opened 2024-12-17
Make NewClientWithEnvProxy Deprecated

hangscer8:


gmlewis:
Sorry I wasn't clear before. 

I would like to get a solution that addresses the earlier problems mentioned here:
https://github.com/google/go-github/issues/2898#issuecomment-1692285018
before deprecating `NewClientWithEnvProxy`.

PR #3395, opened 2024-12-18
fix!: Remove unused `Stats` field from  `Commit` struct

### LINKED ISSUE #3394: `CommitStats` in the `Commit` struct is always nil
The `CommitStats` in the `Commit` struct is always nil.

Comparing the current `Commit` struct:

https://github.com/google/go-github/blob/c4a49db9ca5d5f737f459fa97e3b86c52339b978/github/git_commits.go#L44-L62

With the GitHub OpenAPI Spec:
Image

The `Stats` field is not on the OpenAPI Contract.

---

I know that structs can be used to represents objects that look alike a lot (hence the `CommentCount` field).

So I'm not sure if the `Stats` field is also one of these exceptions. If that's the case, it should be explained just like `CommentCount`.

acouvreur:
BREAKING CHANGE: The unused `Stats` field is removed from the `Commit` struct.

This field appears to never be populated throughout the codebase.

Closes #3394

### DIFF
--- a/github/git_commits.go
+++ b/github/git_commits.go
@@ -49,7 +49,6 @@ type Commit struct {
 	Message      *string                `json:"message,omitempty"`
 	Tree         *Tree                  `json:"tree,omitempty"`
 	Parents      []*Commit              `json:"parents,omitempty"`
-	Stats        *CommitStats           `json:"stats,omitempty"`
 	HTMLURL      *string                `json:"html_url,omitempty"`
 	URL          *string                `json:"url,omitempty"`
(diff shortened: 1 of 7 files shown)


gmlewis (approved):
Thank you, @acouvreur !
LGTM.

DiegoDev2 (approved):
Reviewing the issue this definitely deserves a merge. Good job @acouvreur

gmlewis:
Thank you, @DiegoDev2 !
Merging.

PR #3396, opened 2024-12-21
docs: Add clarifications for mergeable field in pull requests

acouvreur:
Comparing the OpenAPI spec for pull requests shows that these fields are not populated when listing (right), but they are through GET or webhook:
image

In their OpenAPI Spec they have `pull-request` and `pull-request-simple` definitions.

I think that would make sense to also have these structs instead of relying on comments for contextual field population.


github/pulls.go
```diff
@@ -28,49 +28,63 @@ type PullRequestAutoMerge struct {
...
+	ReviewCommentURL   *string               `json:"review_comment_url,omitempty"`
+	Assignee           *User                 `json:"assignee,omitempty"`
+	Assignees          []*User               `json:"assignees,omitempty"`
+	Milestone          *Milestone            `json:"milestone,omitempty"`
+	AuthorAssociation  *string               `json:"author_association,omitempty"`
+	NodeID             *string               `json:"node_id,omitempty"`
+	RequestedReviewers []*User               `json:"requested_reviewers,omitempty"`
+	AutoMerge          *PullRequestAutoMerge `json:"auto_merge,omitempty"`
+
+	// Merged is not populated by List operation.
```
gmlewis, line 62: It is unnecessary and overly verbose to individually list each field's name separately.
Instead, please simply make this a block with a single comment at the top, like this:

```suggestion
	// These fields are not populated by the List operation.
```

acouvreur:
It is now one single block comment @gmlewis

gmlewis:
Please run the scripts listed in step 4 of CONTRIBUTING.md and push the changes.  That should fix the linter and tests.

gmlewis:

acouvreur:
> I will close this PR around the end of January, 2025 as "abandoned" if there is no further response.

Done just in time :)

gmlewis (approved):
Thank you, @acouvreur !
LGTM.
Merging.

PR #3411, opened 2024-12-31
Add opt-in rate limit support on endpoints returning 302s

### LINKED ISSUE #3406: Calls based on `roundTripWithOptionalFollowRedirect` do not respect rate limits
Hi,
I am using `go-github` to pull logs from Github Actions logs and recently had issues with ratelimits on that endpoint, leading to IP blacklisting by GitHub (confirmed with the support).

There are multiple examples of exported methods that do not respect rate limit headers returned by GitHub (e.g. `GetWorkflowJobLogs`, `DownloadArtifact`, ...).

The root cause of this problem is that some endpoint of the GitHub API return a `302` code for success response (like [GetWorkflowJobLogs](https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#download-job-logs-for-a-workflow-run)).

Because go's `http.Client` automatically follows 3xx's by default, the library currently chooses to use the lower level `RoundTripper` to execute the API call instead, leaving out all the API sugar that is usually wrapping calls to `http.Client` in `BaseDo`.

Am I understanding this issue correctly? I have a contribution ready if that's the case to refactor the client to support not following 3xx's on some endpoint

---
cc @rpelliard @Letiste

wmouchere:
Fixes #3406

Some endpoints of the GitHub HTTP API return a 302 response in successful scenarios, often to redirect to archives or artifacts download links. Because the `net/http` client usually follows 302s when calling `Do`, we instead currently use the `RoundTripper` directly for those endpoint to check after each call if the current redirection is a 302. The problem is that the library supports rate limit headers only in the `Do` type of calls.

This PR adds an opt-in feature flag to enable rate limit headers support on those endpoints which end with a 302 redirection. I refactored `Client` a bit to share the core `BareDo` logic between calls that care about redirection details and calls that don't. I added an unexported `bareDoUntilFound` method that behaves similarly to the old `roundTripWithOptionalFollowRedirect` but with all the rate limit and error handling that other endpoints have.

### DIFF
--- a/github/actions_artifacts.go
+++ b/github/actions_artifacts.go
@@ -142,6 +142,14 @@ func (s *ActionsService) GetArtifact(ctx context.Context, owner, repo string, ar
 func (s *ActionsService) DownloadArtifact(ctx context.Context, owner, repo string, artifactID int64, maxRedirects int) (*url.URL, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/artifacts/%v/zip", owner, repo, artifactID)
 
+	if s.client.RateLimitRedirectionalEndpoints {
+		return s.downloadArtifactWithRateLimit(ctx, u, maxRedirects)
+	}
+
+	return s.downloadArtifactWithoutRateLimit(ctx, u, maxRedirects)
+}
+
+func (s *ActionsService) downloadArtifactWithoutRateLimit(ctx context.Context, u string, maxRedirects int) (*url.URL, *Response, error) {
 	resp, err := s.client.roundTripWithOptionalFollowRedirect(ctx, u, maxRedirects)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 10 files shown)


github/actions_artifacts.go
```diff
@@ -160,6 +168,26 @@ func (s *ActionsService) DownloadArtifact(ctx context.Context, owner, repo strin
...
+		return nil, resp, err
+	}
+	defer resp.Body.Close()
+
+	// If we received a valid Location in a 302 response
+	if url != nil {
+		return url, resp, nil
+	}
+
+	return nil, resp, fmt.Errorf("unexpected status code: %s", resp.Status)
```
gmlewis, line 188: For this file and all the other `...WithRateLimit` methods, could we place make the "happy path" the one that flows to the end of the method, as per idiomatic Go?

```suggestion
	// If we didn't receive a valid Location in a 302 response
	if url == nil {
		return nil, resp, fmt.Errorf("unexpected status code: %v", resp.Status)
	}

	return url, resp, nil
```

gmlewis (changes requested):
Thank you, @wmouchere - this is looking great!
I'm requesting one minor change across all your new methods, please.
Also, if you could increase the code coverage for the new methods (see Codecov feedback in the GitHub UI), that would be great, but not absolutely necessary, as the uncovered cases are indeed going to be rare error paths.

Additionally, I'm noticing that you use `%d` and `%s` in format strings, and it is very common in idiomatic Go to simply use `%v` everywhere unless there is a strong need to do otherwise (which then catches the eye as an indication that something unusual is going on here).  It would be nice (but not required) if you switch to `%v` if you don't mind.

gmlewis (approved):
This looks fantastic, @wmouchere !
Beautifully well-done - thank you!
LGTM.

wmouchere:
Thanks a lot for the quick review @gmlewis! 
- Changes you mentioned on the happy path flow are done 👍 
- I should have updated all the `%s` I added to `%v`, but left the `%d` where its used to format a http status. I noticed that this was the pattern used in most places in the library.
- I did my best to increase coverage by testing handling of unexpected status codes. I would have wanted to test invalid `Location` headers but `net/http` actually validates them upstream which changes the flow a bit and makes it hard to test.

Letiste (approved):

gmlewis:
Thank you, @Letiste !
Merging.

PR #3417, opened 2025-01-06
feat!: Add support for enterprise rulesets

### LINKED ISSUE #3416: Add enterprise rules support
I'd like to see support for [enterprise rules](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/rules) added.

stevehipwell:
BREAKING CHANGE: `Create*Ruleset` and `Update*Ruleset` now pass `ruleset` parameter by-value instead of by-reference.

Fixes: #3416.


github/enterprise_rules.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return ruleset, resp, nil
+}
+
+// UpdateEnterpriseRuleset updates a ruleset from the specified enterprise.
+//
+// GitHub API docs: https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/rules#update-an-enterprise-repository-ruleset
+//
+//meta:operation PUT /enterprises/{enterprise}/rulesets/{ruleset_id}
+func (s *EnterpriseService) UpdateEnterpriseRuleset(ctx context.Context, enterprise string, rulesetID int64, rs *Ruleset) (*Ruleset, *Response, error) {
```
gmlewis, line 62:
```suggestion
func (s *EnterpriseService) UpdateEnterpriseRuleset(ctx context.Context, enterprise string, rulesetID int64, rs Ruleset) (*Ruleset, *Response, error) {
```

github/orgs_rules_test.go
```diff
@@ -18,14 +18,16 @@ func TestOrganizationsService_GetAllOrganizationRulesets(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
 
-	mux.HandleFunc("/orgs/o/rulesets", func(w http.ResponseWriter, r *http.Request) {
+	org := "o"
```
gmlewis, line 21: Throughout this PR, let's please revert this change. `org := "o"` doesn't provide any additional clarity (and only increases the number of lines) because `"o"` throughout this repo always represents an organization and `"r"` always represents the name of a repo. In fact, `enterprise := "e"` doesn't really add a whole lot of value either, but you can leave it if you want since it will be one of first representations of an enterprise.

github/orgs_rules_test.go
```diff
@@ -18,14 +18,16 @@ func TestOrganizationsService_GetAllOrganizationRulesets(t *testing.T) {
...
+	mux.HandleFunc(fmt.Sprintf("/orgs/%s/rulesets", org), func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		fmt.Fprint(w, `[{
+		fmt.Fprintf(w, `[{
 			"id": 26110,
 			"name": "test ruleset",
 			"target": "branch",
 			"source_type": "Organization",
-			"source": "o",
+			"source": "%s",
```
gmlewis, line 30: Definitely, please revert.  Placing this `"%s"` so far away from its value is really not helpful.
Additionally, we attempt to use `"%v"` throughout this repo unless we want to call specific attention as to why it important _not_ to be `"%v"`.

gmlewis (changes requested):
Thank you, @stevehipwell .
In addition to the linter findings, please make the following changes:

github/enterprise_rules.go
stevehipwell: I can't see this in the lint output or in the code (this pattern was copied from the org version)?

github/orgs_rules_test.go
stevehipwell: Sure.

github/orgs_rules_test.go
stevehipwell: Sure.

github/enterprise_rules.go
gmlewis: Oh, sorry, I forgot to explain this one.  The lint output didn't show this change.
I was suggesting this change since the `rs` (in fact, maybe we should actually call it `ruleset` here instead of `rs`) parameter is required, so I removed the pointer, showing that it is required and cannot be `nil`.

stevehipwell:
@gmlewis sorry for the force pushing, it's muscle memory.

github/enterprise_rules.go
stevehipwell: Do you want this done for all of the ruleset methods then? As I said, I copied them from the existing org implementation which AFAIK were copied from the repo ones.

stevehipwell:
@gmlewis based on the last run it looks like the OpenAPI schema needs to be updated?

github/enterprise_rules.go
gmlewis: > Do you want this done for all of the ruleset methods then? As I said, I copied them from the existing org implementation which AFAIK were copied from the repo ones.

Ah!  Bummer.  It could be that I didn't catch the other ones when they were being written.

I believe we try to pass-by-value in this repo when args are required.
We also try to be consistent... but obviously I failed here.
We also try to not create breaking API changes, but I think we value consistency over breaking the API.

So, in my opinion, I think it is valuable to pass required args as value, and I think it is valuable to be consistent... so I'm leaning toward a "yes" but am open to discussion.

gmlewis:
> @gmlewis based on the last run it looks like the OpenAPI schema needs to be updated?

I'll see if I can update them locally and report back my findings.

github/enterprise_rules.go
stevehipwell: It might be worth checking which pattern is used the most? The first one I checked where a struct was being used (`EditMilestone`) was also using a pointer.

gmlewis:
> > @gmlewis based on the last run it looks like the OpenAPI schema needs to be updated?
> 
> I'll see if I can update them locally and report back my findings.

@stevehipwell - I just created #3419. Could you please see if this looks right to you?

stevehipwell:
@gmlewis how would you like me to update this PR so the metadata is correct? Usually I'd rebase the commit onto the repo changes.

gmlewis:
> @gmlewis how would you like me to update this PR so the metadata is correct? Usually I'd rebase the commit onto the repo changes.

If you `git pull` the master branch, then you can `git merge master` in this branch and simply push to the PR.

gmlewis:
> Done, but will GH give you a linear history on squash?

Squash gives us a very clean linear history of PRs. Check out this repo's `git log` to see the result.

github/enterprise_rules.go
gmlewis: To be clear, you may do the analysis if you desire to do so.

My preference is to pass-by-value for required structs so that it is clear that `nil` is not an acceptable parameter.

In those cases where you find pass-by-reference parameters, you will also have to check if `nil` is an acceptable parameter.

Meanwhile, my request stands, both for this endpoint and for `CreateEnterpriseRuleset` above, where `nil` is definitely _not_ acceptable.

stevehipwell: Sorry, I thought I'd already fixed the cases in the new code (I have now). My question was about if there needs to be a breaking change for consistency wouldn't it be best to do the whole library in one go?

stevehipwell: I've updated the other rule implementations for consistency.

github/repos_rules.go
```diff
@@ -943,25 +956,21 @@ func (s *RepositoriesService) UpdateRuleset(ctx context.Context, owner, repo str
 // GitHub API docs: https://docs.github.com/rest/repos/rules#update-a-repository-ruleset
 //
 //meta:operation PUT /repos/{owner}/{repo}/rulesets/{ruleset_id}
-func (s *RepositoriesService) UpdateRulesetNoBypassActor(ctx context.Context, owner, repo string, rulesetID int64, rs *Ruleset) (*Ruleset, *Response, error) {
+func (s *RepositoriesService) UpdateRulesetNoBypassActor(ctx context.Context, owner, repo string, rulesetID int64, rs Ruleset) (*Ruleset, *Response, error) {
```
stevehipwell, line 959: @gmlewis my understanding of the code here is that for all ruleset updates we should just use the `rulesetNoOmitBypassActors` type if the bypass actors is empty. Does that make sense and is there a documented pattern for this already established?

gmlewis: Yes, that makes sense and if you would prefer to use this pattern, that is also fine by me if you feel that makes sense for your endpoints (meaning passing a `nil` has a a documented, predefined meaning).

Note that you should document what happens in your endpoints if `nil` is passed.

And since this endpoint already documents what happens when `nil` is passed, there is really no need to make a breaking change to this endpoint.

stevehipwell: @gmlewis this has been done. Could you please re-review?

github/repos_rules.go
```diff
@@ -936,32 +965,55 @@ func (s *RepositoriesService) UpdateRuleset(ctx context.Context, owner, repo str
...
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
+
 // UpdateRulesetNoBypassActor updates a ruleset for the specified repository.
 //
 // This function is necessary as the UpdateRuleset function does not marshal ByPassActor if passed as nil or an empty array.
```
gmlewis, line 995: I think this comment needs updating since it is no longer possible to pass `nil` with this PR change.

In fact, with this deprecation notice, it is necessary to modify this endpoint at all?

gmlewis (changes requested):
One more nit... would you mind renaming the parameter `rs Ruleset` to `ruleset Ruleset` in this PR to make the docs a bit more readable, please?

Oh, and with that last comment I sent - if you still want to change the signature to the deprecated endpoint for consistency, that is fine with me... so maybe just update the comment and ignore the rest.

github/repos_rules.go
stevehipwell: Ha ha ha, I undid the changes. Do you want me to put them back?

gmlewis: SORRY!!!  You are _FAST_!!!

Your call.  I'm fine either way.

stevehipwell: I've updated for consistency and updated the comment as it refers to the default update method. I've also renamed the public facing `rs` to `ruleset`.

github/repos_rules.go
```diff
@@ -992,42 +992,42 @@ func (s *RepositoriesService) UpdateRulesetClearBypassActor(ctx context.Context,
...
-// This function is necessary as the UpdateRuleset function does not marshal ByPassActor if passed as nil or an empty array.
+// This function is necessary as the UpdateRuleset function does not marshal ByPassActor if passed as an empty array.
 //
 // Deprecated: Use UpdateRulesetClearBypassActor instead.
 //
 // GitHub API docs: https://docs.github.com/rest/repos/rules#update-a-repository-ruleset
 //
 //meta:operation PUT /repos/{owner}/{repo}/rulesets/{ruleset_id}
-func (s *RepositoriesService) UpdateRulesetNoBypassActor(ctx context.Context, owner, repo string, rulesetID int64, rs Ruleset) (*Ruleset, *Response, error) {
+func (s *RepositoriesService) UpdateRulesetNoBypassActor(ctx context.Context, owner, repo string, rulesetID int64, ruleSet Ruleset) (*Ruleset, *Response, error) {
```
gmlewis, line 1002: Sorry, one more nit:
Since GitHub doesn't capitalize "Set", let's please not either.
So: `ruleSet` => `ruleset` throughout.

stevehipwell: Sorry, I'll fix this.

gmlewis (approved):
Thank you, @stevehipwell !
LGTM.

dnwe (approved):
feature implementation looks good to me and matches the docs

gmlewis:
Thank you, @dnwe !
Merging.

PR #3418, opened 2025-01-06
Update security manager operations to use organization roles REST APIs

just-joshing:
The [organization security manager role REST APIs](https://docs.github.com/en/rest/orgs/security-managers) are being sunset (https://gh.io/security-managers-rest-api-sunset) in favor of the [organization roles REST APIs](https://docs.github.com/en/rest/orgs/organization-roles).

Instead of introducing a breaking change by removing the operations from go-github, this PR updates the operations to use the organization roles REST APIs so consumers ([such as Terraform](https://github.com/integrations/terraform-provider-github/blob/1c11053126129f2c9ca6dae160e9d3617bcba997/github/resource_github_organization_security_manager.go)) can continue to manage organization security manager teams without needing to change code.


just-joshing:
https://github.com/google/go-github/pull/3421

PR #3424, opened 2025-01-07
build(deps): Pin and group actions/*

dnwe:
Tell dependabot to group updates to the official actions owned by GitHub under the [actions](/actions/org) into a single PR and update the workflows to pin to explicit digests tracked to individual version numbers rather than just the major tag


gmlewis (approved):
Thank you, @dnwe !
LGTM.
Merging.

PR #3426, opened 2025-01-07
Add deprecation messages to security managers APIs

just-joshing:
The [organization security manager role REST APIs](https://docs.github.com/en/rest/orgs/security-managers) are being sunset (https://gh.io/security-managers-rest-api-sunset) in favor of the [organization roles REST APIs](https://docs.github.com/en/rest/orgs/organization-roles).

This change adds deprecation messages to tell consumers which client operations to use instead going forward.

### DIFF
--- a/github/orgs_security_managers.go
+++ b/github/orgs_security_managers.go
@@ -12,6 +12,8 @@ import (
 
 // ListSecurityManagerTeams lists all security manager teams for an organization.
 //
+// Deprecated: Please use `client.Organizations.ListTeamsAssignedToOrgRole` instead.
+//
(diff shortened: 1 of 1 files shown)


just-joshing:
I had previously tried updating these operations instead to use the replacement client methods under-the-hood [but got feedback](https://github.com/google/go-github/pull/3421#issuecomment-2575157830) to go the deprecation route instead.

stevehipwell (approved):
LGTM

gmlewis (approved):
Thank you, @just-joshing and @stevehipwell !
LGTM.
Merging.

PR #3428, opened 2025-01-09
Enforce toolchain requirement in generate.sh

gmlewis:
@dnwe - does this seem like a reasonable change for the use of all developers in this repo?


dnwe:
@gmlewis seems reasonable — an alternative would be to set `toolchain` directives in the go.mod files to the latest 1.22.x as the 'auto' default will pick that up too

gmlewis:
This seems like lower maintenance needed, so let's run with this for now.

PR #3438, opened 2025-01-17
Use max retry after duration for secondary rate limit

Letiste:
The current logic to pick a retry after duration in case of a secondary rate limit is to first check for the retry after header, and if not found, use the primary rate limit reset time. As the primary rate limit reset every hour, we can be blocked for as long as 1 hour when GitHub returns a secondary rate limit error without a retry after header.

According to GitHub documentation, the rate limit reset header should be used when the remaining allowed requests reached 0. Otherwise, retry in 1 min ([doc](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28#dealing-with-secondary-rate-limits)):
> If you receive a rate limit error, you should stop making requests temporarily according to these guidelines:
> - If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed.
> - If the x-ratelimit-remaining header is 0, you should not make another request until after the time specified by the x-ratelimit-reset header. The x-ratelimit-reset header is in UTC epoch seconds.
> - Otherwise, wait for at least one minute before retrying. If your request continues to fail due to a secondary rate limit, wait for an exponentially increasing amount of time between retries, and throw an error after a specific number of retries.

This PR adds an optional max wait time for the secondary rate limit that will, if specified, prevents the client to block up to 1h in some cases.


github/github.go
```diff
@@ -174,6 +174,10 @@ type Client struct {
 	rateLimits              [Categories]Rate // Rate limits for the client as determined by the most recent API calls.
 	secondaryRateLimitReset time.Time        // Secondary rate limit reset for the client as determined by the most recent API calls.
 
+	// If specified, Client will block requests for at most this duration in case of reaching a secondary
+	// rate limit
+	MaxSecondaryRateLimitRetryAfterDuration time.Duration
```
Letiste, line 179: I'm letting users specify their desired max duration. We could enforce a minimum or a default value here if we want to make sure users are not using this field incorrectly

github/github.go
```diff
@@ -920,6 +924,10 @@ func (c *Client) bareDo(ctx context.Context, caller *http.Client, req *http.Requ
 		// Update the secondary rate limit if we hit it.
 		rerr, ok := err.(*AbuseRateLimitError)
 		if ok && rerr.RetryAfter != nil {
+			// if a max duration is specified, make sure that we are waiting at most this duration
+			if c.MaxSecondaryRateLimitRetryAfterDuration > 0 && rerr.GetRetryAfter() > c.MaxSecondaryRateLimitRetryAfterDuration {
+				rerr.RetryAfter = &c.MaxSecondaryRateLimitRetryAfterDuration
+			}
```
Letiste, line 930: I wanted to enforce the max duration in the `parseSecondaryRate` function but we don't have access to the client there, and it was tedious to pass a parameter downstream so I'm mutating the error here instead. Let me know if you would prefer another solution

gmlewis:
I am extremely leery of changing the default behavior regarding rate limiting.
If you want to provide an option for your feature, I'm fine with that.

@wmouchere - please comment on this PR because of your recent work on #3411.

gmlewis (approved):
Thank you, @Letiste.
LGTM.

stevehipwell (approved):
LGTM

gmlewis:
Thank you, @stevehipwell!
Merging.

Issue #1429, opened 2020-02-13
Why are all the responses using pointers


marcofranssen:
I noticed most responses are using slices of pointers. IMHO it would be better to not use pointers in the slices.

Also I don't think we should use pointers for strings in the JSON.

It will reduce the amount of required null checks, simplify working with the strings. Furthermore a slice is already kind of using pointers.

Is there any reason for this?

Would a PR removing the pointers in those locations be accepted?

gmlewis:
Please see #79, #45, and #19. 

Thank you for asking, but no, a PR removing the pointers would not be accepted.

marcofranssen:
I get the nil checking, what I don't understand is why to apply this on `string` and `[]*SomeStruct`. In those cases I think they should be removed. For integers it is debatable as well I think as also there, why would I want to check for nil in those cases?

I Would do something like this. That simplifies codebase and consumption of it a lot.

```go
type SomeType struct {
     MyOtherObject *OtherType
     Amount               Int64
     Text                      string
     Collection           []CollectionItem
}
```

An empty string can then be treated the same as nil and 0 can also be treated similar. No more nil checks there. And I think in the case of slices it doesn't add value as well. Either there is an element in the slice or it is not. Why would you want to have nil elements in the slice?

Only the case were another struct like MyOtherObject is used I could think of user a pointer.

Am I completely missing out on something?

gmlewis:
This has been discussed at length... (whups, I forgot to include others like #537), but here are short answers:

* an empty string will not tell you if the field was completely missing in the response. In this case, a string pointer of `nil` would indicate that the field was missing.
* for slices, it prevents copying data around... sometimes these structs are large and would involve a lot of memory copying, so we stuck with a pattern.  Additionally, slices of pointers make for loop behavior much easier for everyone.  Here is an issue filed by Russ Cox on the subject: #180.

marcofranssen:
Thanks for explaining.

PR #3447, opened 2025-01-21
chore!: Add `sliceofpointers` custom linter 

gmlewis:
BREAKING CHANGE: `ListOAuthApps` now returns `([]*OAuthApp, error)` instead of `([]OAuthApp, error)`.

This PR should help developers use a consistent style by not writing `[]*string` or `[]T` in their PRs.
It should also ease the maintenance burden by not having to repeat this over and over again.


.custom-gcl.yml
```diff
@@ -0,0 +1,4 @@
+version: v1.57.0
```
ldez, line 1:
```suggestion
version: v1.62.0
```

To be consistent with https://github.com/google/go-github/blob/27bd17565a213b72d232e8e1ea820881f791f1a7/script/lint.sh#L8

ldez (changes requested):

gmlewis:
@stevehipwell - do you have time for a code review?  Thank you!

ldez:
I'm curious: what is the reasoning behind using slice of pointer of struct inside of slice of struct?

ldez (approved):

gmlewis:
> I'm curious: what is the reasoning behind using slice of pointer of struct inside of slice of struct?

- #180 
- #1429
- #537 
- #79
- #45
- #19

gmlewis:
Thank you, @ldez!
Merging.

PR #3449, opened 2025-01-23
chore: Check and fix license headers

ldez:
Add `goheader` to check the file header license.

### DIFF
--- a/github/pulls_reviewers.go
+++ b/github/pulls_reviewers.go
@@ -85,11 +85,7 @@ func (s *PullRequestsService) ListReviewers(ctx context.Context, owner, repo str
 func (s *PullRequestsService) RemoveReviewers(ctx context.Context, owner, repo string, number int, reviewers ReviewersRequest) (*Response, error) {
 	// reviewers.Reviewers may be empty if the caller wants to remove teams, but not users. Unlike AddReviewers,
 	// "reviewers" is a required param here. Reference: https://github.com/google/go-github/issues/3336
-	removeRequest := removeReviewersRequest{
-		NodeID:        reviewers.NodeID,
-		Reviewers:     reviewers.Reviewers,
-		TeamReviewers: reviewers.TeamReviewers,
-	}
+	removeRequest := removeReviewersRequest(reviewers)
 
 	if removeRequest.Reviewers == nil {
 		// GitHub accepts the empty list, but rejects null. Removing `omitempty` is not enough - we also have to promote nil to [].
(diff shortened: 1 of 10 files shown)


github/pulls_reviewers.go
```diff
@@ -85,11 +85,7 @@ func (s *PullRequestsService) ListReviewers(ctx context.Context, owner, repo str
 func (s *PullRequestsService) RemoveReviewers(ctx context.Context, owner, repo string, number int, reviewers ReviewersRequest) (*Response, error) {
 	// reviewers.Reviewers may be empty if the caller wants to remove teams, but not users. Unlike AddReviewers,
 	// "reviewers" is a required param here. Reference: https://github.com/google/go-github/issues/3336
-	removeRequest := removeReviewersRequest{
-		NodeID:        reviewers.NodeID,
-		Reviewers:     reviewers.Reviewers,
-		TeamReviewers: reviewers.TeamReviewers,
-	}
+	removeRequest := removeReviewersRequest(reviewers)
```
gmlewis, line 88: Why is this change in this PR?

ldez: because locally golangci-lint reported this element, I can remove it.

ldez: It's because I use golangci-lint v1.63.4 and the CI uses v1.62.0.

Do you want me to take it off?

gmlewis: I ask because I think a proper cleanup would be to completely get rid of the `removeRequest` variable entirely since the original param is passed by value anyway.

gmlewis: And also because `removeRequest := removeReviewersRequest(reviewers)` looks like you're calling a function, when in actuality it is performing an unnecessary cast.

gmlewis: I'm surprised that the linter didn't flag the rewritten line.  Weird.

ldez: I replace `removeRequest` with `reviewers` or I revert the change?

gmlewis: Yes, it makes sense to me to strictly use `reviewers` throughout the method and get rid of `removeRequest` entirely.  Does that make sense to you too?

ldez: We cannot change the type:

```go
type removeReviewersRequest struct {
	NodeID *string `json:"node_id,omitempty"`
	// Note the lack of omitempty! See comment in RemoveReviewers.
	Reviewers     []string `json:"reviewers"`
	TeamReviewers []string `json:"team_reviewers,omitempty"`
}
```

```go
type ReviewersRequest struct {
	NodeID        *string  `json:"node_id,omitempty"`
	Reviewers     []string `json:"reviewers,omitempty"`
	TeamReviewers []string `json:"team_reviewers,omitempty"`
}
```

The struct tags are different.

ldez: the cast is better than using the fields only because it ensures that the 2 types are identical.

gmlewis: OH!  Right!  That's the whole point of the copy... gotcha.
OK, then can you please add one more line saying why this cast is necessary (because the JSON tags are different)?

gmlewis (approved):
Thank you for doing this cleanup, @ldez!
LGTM.
Merging.

Issue #3409, opened 2024-12-30
Update unit tests so that `GODEBUG=httpmuxgo121=1` is not required


gmlewis:
Go version 1.22 introduced significant changes to the pattern syntax and matching
behavior of `http.ServerMux` which causes a large number of legacy unit tests to break.
(See https://pkg.go.dev/net/http#hdr-Compatibility-ServeMux for more information.)
As a result, testing of this repo is currently performed by setting this env variable:

```bash
export GODEBUG=httpmuxgo121=1
```

It would be greatly appreciated if an experienced Go developer would update this repo's
legacy unit tests so that `go test ./...` passes without the use of setting the `GODEBUG`
environment variable.

Ideally, no unit tests would need to be deleted, however it is possible that some tests
no longer make sense. If that is the case, it would be nice to see a detailed explanation
for any removed unit tests as to why it no longer makes sense to keep.

Thank you in advance for your contributions!  See CONTRIBUTING.md for helpful details
for contributing to this repo.

DiegoDev2:
Hello, do you have a deadline or something like that? @gmlewis

gmlewis:
Hi @DiegoDev2 , no, I don't have a timeline... this would just be a nice-to-have.
Since you responded almost immediately, would you mind performing a code review of #3410 which is related?
Would you like me to assign this issue to you?

DiegoDev2:
Yes, assign it to me, but something else is needed, I'm reviewing the PR @gmlewis

gmlewis:
> Yes, assign it to me, but something else is needed, I'm reviewing the PR [#3410](https://github.com/google/go-github/pull/3410) [@gmlewis](https://github.com/gmlewis)

Thank you, @DiegoDev2 !
It's yours.

DiegoDev2:
Well, excuse me, what files do I have to work on and were you part of the issue?

DiegoDev2:
I was a little confused, but that's it. Don't worry.

gmlewis:
I'm not sure I understand what you mean.  

Are you clear now on what I'm asking to be done in the description above?

gmlewis:
To be hopefully be more clear, #3410 did not cause the problem.

Go 1.22 caused the problem.

We would like to keep this repo on the most two recent (minor) versions of Go without any odd workarounds.
To accomplish this, we first needed to write #3410 to move this repo to Go 1.22.10 (and Go 1.23.4 for the maintenance tools).
[...]

PR #3459, opened 2025-02-05
fix: go 1.22 test breakage

amanv8060:
Use correct url escaping to fix the errors caused by mux changes.

Fixes: #3409

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,6 @@ go-github will require the N-1 major release of Go by default.
 
 ## Development
 
-Go version 1.22 introduced significant changes to the pattern syntax and matching
-behavior of `http.ServerMux` which causes a large number of legacy unit tests to break.
-(See https://pkg.go.dev/net/http#hdr-Compatibility-ServeMux for more information.)
-As a result, testing of this repo is currently performed by setting this env variable:
-
-```bash
-export GODEBUG=httpmuxgo121=1
-```
-
-An issue has been created (#3409) requesting assistance in updating all breaking legacy unit
-tests when this `GODEBUG` environment variable is not set and Go 1.23.4 or later is
-used to perform unit tests.
-
 If you're interested in using the [GraphQL API v4][], the recommended library is
 [shurcooL/githubv4][].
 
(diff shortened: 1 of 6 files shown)


amanv8060:
@gmlewis could you please approve actions test to see if all works correctly?

gmlewis:
Yes, however we don't want to remove problematic (but valid) branch names from the unit tests (as you have done in this PR)...
Instead, we want to fix the testing mux harness to properly handle all the problematic branch names.
Does that make sense?

amanv8060:
> Yes, however we don't want to remove problematic (but valid) branch names from the unit tests (as you have done in this PR)... Instead, we want to fix the testing mux harness to properly handle all the problematic branch names. Does that make sense?

I am not sure if I get what you're saying, I haven't really removed any branch names, instead in the mux urls(the one's to be called) by api, I have just [Path escaped](https://pkg.go.dev/net/url#PathEscape) them, github also does this, for eg, 

[This](https://github.com/flutter/engine/compare/main...shorebirdtech:engine:shorebird/dev) would go like this:

`https://github.com/flutter/engine/compare/commit-list?range=main...shorebirdtech%3Aengine%3Ashorebird%2Fdev`

Also see, [tests in net/http](https://github.com/golang/go/commit/3563792768b1e06da4a0cb5f946adf90d1297766#diff-fd7f79f4735e1892698cce88905ff7a09531ad0094fb065fe6a2c0dc9f823806L1448-L1464).


IMO, one thing that could be done better here is,to create a util function of sort that would take the branch name and path escape them before adding it to mux route.

gmlewis:
> I am not sure if I get what you're saying, I haven't really removed any branch names, instead in the mux urls(the one's to be called) by api, I have just [Path escaped](https://pkg.go.dev/net/url#PathEscape) them

Ah!  I see now what you are saying, @amanv8060. I apologize for the confusion I caused.

Yes, a utility function would help tremendously, I would think, and make it clearer what is happening.

Would you mind doing that as part of this PR?

amanv8060:
I would prefer to do that in seperate pr so that review becomes easier.

gmlewis (approved):
Thank you, @amanv8060 !
LGTM.

@stevehipwell - might you have time for a code review?  Thanks!

alexandear (approved):

gmlewis:
Thank you, @stevehipwell and @alexandear !
Merging.

Issue #3473, opened 2025-02-11
Support pulling only runID from Custom Deployment Protection Rule Events


pputman-clabs:
In:
https://github.com/google/go-github/blob/v69.0.0/github/actions_workflow_runs.go#L521

We require a RunID argument in order to respond back to and authorize/deny a deployment.

Currently it looks like the only way to get the runID, is from the *github.DeploymentProtectionRuleEvent.DeploymentCallbackURL
which contains the entire callback URL, not just the runID.  Since the function only needs the runID and not the string, the options are to either:

1. Use a post request directly with this URL rather than using the library.
2. Use regex to strip the runID from this URL, just to use it with this function which adds it right back into the same URL it was stripped out of.

It would be much cleaner if we could just use the callback URL as an option to the function, or if the event had a field that was just the RunID and not the entire url string

gmlewis:
What about adding a helper function that transforms the callback URL to a runID that is ready-to-use, since there are many endpoints in this file that require a `runID` as a param.

If we decide to go with an exported helper function, then I would recommend that we update the comments to all the methods that require `runID` to state that the helpful function can be used to get it.

Thoughts?

pputman-clabs:
Sounds like a great option to me!

gmlewis:
> Sounds like a great option to me!

Would you like to volunteer to write a PR, @pputman-clabs ? :smile:
Or shall I open this issue up to our large community of awesome contributors to implement?

pputman-clabs:
Sure I'll try.

pputman-clabs:
@gmlewis Can you look at https://github.com/google/go-github/pull/3476 and let me know what you think?

PR #3476, opened 2025-02-12
Add helper to get `runID` from Custom Deployment Protection Rule Event

pputman-clabs:
Added Helper Function with test to get the RunID from the DeploymentCallbackURL and convert it to an int64.

Fixes: #3473.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -1774,7 +1774,7 @@ func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 
 // Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL.
 func (e *DeploymentProtectionRuleEvent) GetRunID() (int64, error) {
-	r := regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(?P<RunID>[0-9].*)\/deployment_protection_rule$`)
+	r := regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(?P<RunID>[0-9]*)\/deployment_protection_rule$`)
 	match := r.FindStringSubmatch(*e.DeploymentCallbackURL)
 	if len(match) != 2 {
 		return -1, errors.New("no match")
--- a/github/github.go
+++ b/github/github.go
@@ -1772,7 +1772,6 @@ func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
 
-
 var r *regexp.Regexp = regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(\d+)\/deployment_protection_rule$`)
 
 // Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL.
--- a/github/github.go
+++ b/github/github.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	Version = "v69.0.0"
+	Version = "v69.1.0"
 
 	defaultAPIVersion = "2022-11-28"
 	defaultBaseURL    = "https://api.github.com/"
--- a/github/github.go
+++ b/github/github.go
@@ -1777,8 +1777,9 @@ var r *regexp.Regexp = regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(\d+)\
 // Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL.
 func (e *DeploymentProtectionRuleEvent) GetRunID() (int64, error) {
 	match := r.FindStringSubmatch(*e.DeploymentCallbackURL)
+	var ErrNoMatch = errors.New("no match found")
 	if len(match) != 2 {
-		return -1, errors.New("no match")
+		return -1, ErrNoMatch
 	}
 	runID, err := strconv.ParseInt(match[1], 10, 64)
 	if err != nil {
--- a/github/github.go
+++ b/github/github.go
@@ -1777,9 +1777,8 @@ var r *regexp.Regexp = regexp.MustCompile(`^repos/.*/actions/runs/(\d+)/deployme
 // Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL.
 func (e *DeploymentProtectionRuleEvent) GetRunID() (int64, error) {
 	match := r.FindStringSubmatch(*e.DeploymentCallbackURL)
-	var ErrNoMatch = errors.New("no match found")
 	if len(match) != 2 {
-		return -1, ErrNoMatch
+		return -1, errors.New("no match")
 	}
 	runID, err := strconv.ParseInt(match[1], 10, 64)
 	if err != nil {
--- a/github/github.go
+++ b/github/github.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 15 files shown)


github/github.go
```diff
@@ -1770,3 +1771,15 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
+
+// Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL
+func (e *DeploymentProtectionRuleEvent) GetRunID() (int64, error) {
+	r := regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(?P<RunID>[0-9].*)\/deployment_protection_rule$`)
+	match := r.FindStringSubmatch(*e.DeploymentCallbackURL)
```
gmlewis, line 1778: Line 1780 will crash if there was no match.  So you need to add this:

```go
if len(match) != 2 {
  return -1, errors.New("no match")
}

File: github/github.go [hunk 2]
```diff
@@ -1770,3 +1771,15 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
+
+// Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL
```
gmlewis, line 1775:
```suggestion
// Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL.
```

github/github.go
```diff
@@ -1770,3 +1771,15 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
...
+func (e *DeploymentProtectionRuleEvent) GetRunID() (int64, error) {
+	r := regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(?P<RunID>[0-9].*)\/deployment_protection_rule$`)
+	match := r.FindStringSubmatch(*e.DeploymentCallbackURL)
+
+	runID, err := strconv.ParseInt(match[1], 10, 64)
+	if err != nil {
+		return -1, err
```
gmlewis, line 1782: Please add a test that returns a bogus `runID`, for example `abc123` so that this line is covered.

gmlewis (changes requested):
Thank you, @pputman-clabs.

It might be nice to add a comment to the other functions that required a `runID` that this new helper can be used to easily extract the value from the event.

github/github.go
pputman-clabs: I'm curious how to do this, because if I say set abc123, then it has to match the "want" which is an int64.  So it would never match the regex, and return the Error: No Match, before it got to this code anyways.
Any suggestions on how I could approach it?

gmlewis: Actually, I neglected to reviewer your regexp yet... and it is not quite right... I'm in a meeting right now, but `[0-9].*` will match a lot more than just digits. That will be a single digit followed by anything.

gmlewis: Instead of `(?P<RunID>[0-9]*)`, how about simply `(\d+)` ?

Also, since `regexp.MustCompile` is typically an expensive operation, we pull the whole thing outside to a global variable instead of embedding it into a method.

github/github_test.go
```diff
@@ -3101,3 +3101,33 @@ func TestPtr(t *testing.T) {
...
+
+func TestDeploymentProtectionRuleEvent_GetRunID(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		want int64
+		URL   string
+	}{
+		{
+			want: 123456789,
+			URL:   fmt.Sprintf("repos/dummy-org/dummy-repo/actions/runs/%v/deployment_protection_rule", 123456789),
```
gmlewis, line 3113:
```suggestion
			URL:   "repos/dummy-org/dummy-repo/actions/runs/123456789/deployment_protection_rule",
```

github/github_test.go
```diff
@@ -3101,3 +3101,33 @@ func TestPtr(t *testing.T) {
...
+		},
+		{
+			want: ,
+			URL:   fmt.Sprintf("repos/dummy-org/dummy-repo/actions/runs/%v/deployment_protection_rule", "abc123"),
```
gmlewis, line 3117:
```suggestion
			URL:   "repos/dummy-org/dummy-repo/actions/runs/abc123/deployment_protection_rule",
```

github/github_test.go
```diff
@@ -3101,3 +3101,33 @@ func TestPtr(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+
+		e := DeploymentProtectionRuleEvent{
+			DeploymentCallbackURL: &tt.URL,
+		}
+		got, err := e.GetRunID()
+		if err != nil {
+			t.Errorf("error %v", err)
```
gmlewis, line 3127: Shouldn't the second case cause an error?

gmlewis (changes requested):

github/github.go
pputman-clabs: Added this, thanks!

github/github.go
pputman-clabs: added

github/github_test.go
pputman-clabs: changed

github/github_test.go
pputman-clabs: Added test for error and incorrect values

github/github_test.go
pputman-clabs: changed

pputman-clabs:
Okay I made the requested changes, thanks for the feedback!  Let me know if there's anything else programming wise you think I should change and if not I'll start adding the documentation for it to the other method docs.

gmlewis (changes requested):
Please revert changes to go.mod and go.sum, as this helper and its tests should not require any more dependencies than we already have.

github/github.go
```diff
@@ -1770,3 +1771,19 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
+
+var r *regexp.Regexp = regexp.MustCompile(`^repos\/.*\/.*\/actions\/runs\/(\d+)\/deployment_protection_rule$`)
```
gmlewis, line 1775: Just out of curiosity... did you try without all the backslashes?

Does this still work?

```go
var r *regexp.Regexp = regexp.MustCompile(`^repos/.*/actions/runs/(\d+)/deployment_protection_rule$`)
```

Since `.*` will slurp up the intervening slash, we really only need one of them, I'm thinking.

pputman-clabs: yes that seems to work, changed.  My thoughts were to try to make the regex as specific as possible, but yeah it looks much cleaner this way.

github/github.go
```diff
@@ -1770,3 +1771,19 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
+
+var r *regexp.Regexp = regexp.MustCompile(`^repos/.*/actions/runs/(\d+)/deployment_protection_rule$`)
+
+// Helper Function to extract the workflow RunID from the *DeploymentProtectionRuleEvent.DeploymentCallBackURL.
+func (e *DeploymentProtectionRuleEvent) GetRunID() (int64, error) {
+	match := r.FindStringSubmatch(*e.DeploymentCallbackURL)
+	var ErrNoMatch = errors.New("no match found")
```
gmlewis, line 1780: This line is not idiomatic Go for two reasons:
* there is no need to export (capitalize) it
* it is only used once and provides no more useful information by making it a separate line.

Therefore, please revert this change.

pputman-clabs:
Made requested changes, and added a comment to the actions_workflow_runs.go methods that they can use this helper function to easily pull out the RunID

github/github.go
pputman-clabs: Reverted.

github/github.go
```diff
@@ -1770,3 +1771,18 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 func (fn roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 	return fn(r)
 }
+
+var r = regexp.MustCompile(`^repos/.*/actions/runs/(\d+)/deployment_protection_rule$`)
```
gmlewis, line 1775: One more tweak, please. Since this variable is now a package global, let's please give it a descriptive name, maybe `runIDFromURLRE`?

pputman-clabs: Changed, thanks

gmlewis (approved):
Thank you, @pputman-clabs and @stevehipwell!
Merging.

PR #3477, opened 2025-02-12
fix: Add missing query params to AlertListOptions

maaarcelino:
As per the official documentation the endpoints to list alerts for both an organization and a repository accept three additional params which are not included in this SDK. Those are: `tool_guid`, `direction` and `sort`.

See:

- https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-an-organization
- https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-a-repository

This PR adds those fields and extends the tests to cover them as well.

### DIFF
--- a/github/code_scanning.go
+++ b/github/code_scanning.go
@@ -141,6 +141,15 @@ type AlertListOptions struct {
 	// The name of a code scanning tool. Only results by this tool will be listed.
 	ToolName string `url:"tool_name,omitempty"`
 
+	// The GUID of a code scanning tool. Only results by this tool will be listed.
+	ToolGUID string `url:"tool_guid,omitempty"`
+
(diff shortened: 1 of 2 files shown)


maaarcelino:
I ran the linting, testing and generation scripts. @gmlewis can you please take a look? 🙏

maaarcelino:
@gmlewis Thanks! Any word on when I can expect a new release? Is there any chance you could do a minor release sooner than later?

gmlewis:
Usually we cut a release monthly, but we've also tried to be accommodating to requests, so I can work on a minor release.

maaarcelino:
That's fantastic. Thank you so much @gmlewis! ❤️

PR #3480, opened 2025-02-15
feat: Add JSON marshal tests for dependabot alerts

raisa2010:
This PR adds some more tests for issue #55. Would appreciate a review!

Also noticed that some fields in the Github API response such as `cvss_severities` and `epss` are not supported in the existing struct. Is this expected (deprecated fields) or just new fields in the Github response that haven't been reflected in the current codebase yet? 

Thanks in advance!

### DIFF
--- a/github/dependabot_alerts_test.go
+++ b/github/dependabot_alerts_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 )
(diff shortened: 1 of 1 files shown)


gmlewis:
> Also noticed that some fields in the Github API response such as `cvss_severities` and `epss` are not supported in the existing struct. Is this expected (deprecated fields) or just new fields in the Github response that haven't been reflected in the current codebase yet?

There are several possibilities, and I would have to look up those specific cases to answer your question but don't have time at the moment. We rarely remove fields but do try to mark them as deprecated if indeed they are. Most of the time, they are added by GitHub after our endpoints are written. When someone needs them, we add a PR.

gmlewis (approved):
Thank you, @raisa2010.
LGTM.
Merging.

raisa2010:
Got it, thank you for clarifying and reviewing @gmlewis!

Issue #3461, opened 2025-02-05
Add support for `actions/hosted-runners` endpoints


stevehipwell:
The REST endpoints for GitHub-hosted larger runners have been [announced as GA](https://github.blog/changelog/2025-01-29-actions-github-hosted-larger-runner-network-configuration-rest-apis-ga/).

- [OSS Docs](https://docs.github.com/en/rest/actions/hosted-runners?apiVersion=2022-11-28)
- [Enterprise docs](https://docs.github.com/en/enterprise-cloud@latest/rest/actions/hosted-runners?apiVersion=2022-11-28)

gmlewis:
This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.
Also feel free to perform a code review on any PR that has the label "NeedsReview" - we appreciate those contributions too!
We have a `REVIEWERS` file that you could add your username to if you are willing to be contacted when reviews are needed.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

gmlewis:
Thank you, @atilsensalduz!
It's yours.  How about `actions_hosted_runners.go`?

I'm happy to hear that you read CONTRIBUTING.md. For guidance, I recommend you read through a similar collection of endpoints, such as `actions_runners.go` or another one of your choice, and read the official (linked) GitHub documentation for both the already-implemented collection and for the collection you will be implementing, and follow the already-established pattern for how you create your new endpoints along with their corresponding tests.

Make sure your tests pass locally along with the other step-4 scripts in CONTRIBUTING.md. Then, when you are happy with your results, go ahead and post a PR.

PR #3487, opened 2025-02-20
feat(actions): add support for GitHub-hosted runner API endpoints

atilsensalduz:
This pull request introduces support for the GitHub-hosted runner API endpoints in the go-github library, addressing [Issue #3461](https://github.com/google/go-github/issues/3461).

Changes Include:
* Implementation of API endpoints for managing GitHub-hosted runners.
* Support for the following operations:
* Create a new hosted runner.
* Update existing hosted runner details.
* Delete a hosted runner.
* Retrieve details of hosted runners.
* Corresponding unit tests for each operation.

Additional Information:
API Reference: [GitHub REST API: Hosted Runners](https://docs.github.com/en/rest/actions/hosted-runners).

Please let me know if any additional changes or improvements are needed!

### DIFF
--- a/github/actions_hosted_runners.go
+++ b/github/actions_hosted_runners.go
@@ -1,0 +1,333 @@
+// Copyright 2020 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// HostedRunnerPublicIP represents the details of a public IP for github-hosted runner.
+type HostedRunnerPublicIP struct {
+	Enabled bool   `json:"enabled"`
+	Prefix  string `json:"prefix"`
+	Length  int    `json:"length"`
+}
+
+// HostedRunnerMachineSpec represents the details of a particular machine specification for github-hosted runner.
+type HostedRunnerMachineSpec struct {
+	ID        string `json:"id"`
+	CPUCores  int    `json:"cpu_cores"`
+	MemoryGB  int    `json:"memory_gb"`
+	StorageGB int    `json:"storage_gb"`
+}
+
+// HostedRunner represents a single github-hosted runner with additional details.
+type HostedRunner struct {
+	ID                 *int64                   `json:"id,omitempty"`
+	Name               *string                  `json:"name,omitempty"`
+	RunnerGroupID      *int64                   `json:"runner_group_id,omitempty"`
+	Platform           *string                  `json:"platform,omitempty"`
+	Image              *HostedRunnerImageDetail `json:"image,omitempty"`
+	MachineSizeDetails *HostedRunnerMachineSpec `json:"machine_size_details,omitempty"`
+	Status             *string                  `json:"status,omitempty"`
+	MaximumRunners     *int64                   `json:"maximum_runners,omitempty"`
+	PublicIPEnabled    *bool                    `json:"public_ip_enabled,omitempty"`
+	PublicIPs          []*HostedRunnerPublicIP  `json:"public_ips,omitempty"`
+	LastActiveOn       *string                  `json:"last_active_on,omitempty"`
+}
+
+// HostedRunnerImageDetail represents the image details of a github-hosted runners.
+type HostedRunnerImageDetail struct {
+	ID   *string `json:"id,omitempty"`
+	Size *int    `json:"size,omitempty"`
+}
+
+// HostedRunners represents a collection of github-hosted runners for a organization.
+type HostedRunners struct {
+	TotalCount int             `json:"total_count"`
+	Runners    []*HostedRunner `json:"runners"`
+}
+
+// ListHostedRunners lists all the github-hosted runners for a organization.
+//
+// GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#list-github-hosted-runners-for-an-organization
+//
+//meta:operation GET /orgs/{org}/actions/hosted-runners
+func (s *ActionsService) ListHostedRunners(ctx context.Context, org string, opts *ListOptions) (*HostedRunners, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/actions/hosted-runners", org)
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	runners := &HostedRunners{}
+	resp, err := s.client.Do(ctx, req, &runners)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return runners, resp, nil
+}
+
+// HostedRunnerImage represents the image of github-hosted runners
+// To list all available images, use GET /actions/hosted-runners/images/github-owned or GET /actions/hosted-runners/images/partner
+type HostedRunnerImage struct {
+	ID      string `json:"id"`
+	Source  string `json:"source"`
+	Version string `json:"version"`
+}
+
+// CreateHostedRunnerRequest specifies body parameters to Hosted Runner configuration.
+type CreateHostedRunnerRequest struct {
+	Name           string            `json:"name"`
+	Image          HostedRunnerImage `json:"image"`
+	RunnerGroupID  int64             `json:"runner_group_id"`
+	Size           string            `json:"size"`
+	MaximumRunners int64             `json:"maximum_runners,omitempty"`
+	EnableStaticIP bool              `json:"enable_static_ip,omitempty"`
+}
+
+// CreateHostedRunner creates a github-hosted runner for an organization.
+//
+// GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#create-a-github-hosted-runner-for-an-organization
+//
+//meta:operation POST /orgs/{org}/actions/hosted-runners
+func (s *ActionsService) CreateHostedRunner(ctx context.Context, org string, request *CreateHostedRunnerRequest) (*HostedRunner, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/actions/hosted-runners", org)
+	req, err := s.client.NewRequest("POST", u, request)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	hostedRunner := new(HostedRunner)
+	resp, err := s.client.Do(ctx, req, hostedRunner)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return hostedRunner, resp, nil
+}
+
+// HostedRunnerImageSpecs represents the details of a github-hosted runner image.
+type HostedRunnerImageSpecs struct {
+	ID          string `json:"id"`
+	Platform    string `json:"platform"`
+	SizeGB      int    `json:"size_gb"`
+	DisplayName string `json:"display_name"`
+	Source      string `json:"source"`
+}
+
+// HostedRunnerImages represents the response containing the total count and details of runner images.
+type HostedRunnerImages struct {
+	TotalCount int                       `json:"total_count"`
+	Images     []*HostedRunnerImageSpecs `json:"images"`
+}
+
+// GetHostedRunnerGithubOwnedImages gets the list of GitHub-owned images available for github-hosted runners for an organization.
+//
+// GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#get-github-owned-images-for-github-hosted-runners-in-an-organization
+//
+//meta:operation GET /orgs/{org}/actions/hosted-runners/images/github-owned
+func (s *ActionsService) GetHostedRunnerGithubOwnedImages(ctx context.Context, org string) (*HostedRunnerImages, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/actions/hosted-runners/images/github-owned", org)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	hostedRunnerImages := new(HostedRunnerImages)
+	resp, err := s.client.Do(ctx, req, hostedRunnerImages)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return hostedRunnerImages, resp, nil
+}
+
+// GetHostedRunnerPartnerImages gets the list of partner images available for github-hosted runners for an organization.
+//
+// GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#get-partner-images-for-github-hosted-runners-in-an-organization
+//
+//meta:operation GET /orgs/{org}/actions/hosted-runners/images/partner
+func (s *ActionsService) GetHostedRunnerPartnerImages(ctx context.Context, org string) (*HostedRunnerImages, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/actions/hosted-runners/images/partner", org)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	hostedRunnerImages := new(HostedRunnerImages)
+	resp, err := s.client.Do(ctx, req, hostedRunnerImages)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return hostedRunnerImages, resp, nil
+}
+
+// HostedRunnerPublicIPLimits represents the static public IP limits for github-hosted runners.
+type HostedRunnerPublicIPLimits struct {
+	PublicIPs *PublicIPUsage `json:"public_ips"`
+}
+
+// PublicIPUsage provides details of static public IP limits for github-hosted runners.
+type PublicIPUsage struct {
+	Maximum      int `json:"maximum"`
+	CurrentUsage int `json:"current_usage"`
+}
+
+// GetHostedRunnerLimits gets the github-hosted runners Static public IP Limits for an organization.
+//
+// GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#get-limits-on-github-hosted-runners-for-an-organization
+//
+//meta:operation GET /orgs/{org}/actions/hosted-runners/limits
+func (s *ActionsService) GetHostedRunnerLimits(ctx context.Context, org string) (*HostedRunnerPublicIPLimits, *Response, error) {
+	u := fmt.Sprintf("orgs/%v/actions/hosted-runners/limits", org)
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	publicIPLimits := new(HostedRunnerPublicIPLimits)
+	resp, err := s.client.Do(ctx, req, publicIPLimits)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return publicIPLimits, resp, nil
+}
+
+// HostedRunnerMachineSpecs represents the response containing the total count and details of machine specs for github-hosted runners.
+type HostedRunnerMachineSpecs struct {
+	TotalCount   int                        `json:"total_count"`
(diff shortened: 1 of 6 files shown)


github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
+// Copyright 2020 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1: For new files, we use the current year, then we never change the year of an existing file.

```suggestion
// Copyright 2025 The go-github AUTHORS. All rights reserved.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+	Name               *string                  `json:"name,omitempty"`
+	RunnerGroupID      *int64                   `json:"runner_group_id,omitempty"`
+	Platform           *string                  `json:"platform,omitempty"`
+	Image              *HostedRunnerImageDetail `json:"image,omitempty"`
+	MachineSizeDetails *HostedRunnerMachineSpec `json:"machine_size_details,omitempty"`
+	Status             *string                  `json:"status,omitempty"`
+	MaximumRunners     *int64                   `json:"maximum_runners,omitempty"`
+	PublicIPEnabled    *bool                    `json:"public_ip_enabled,omitempty"`
+	PublicIPs          []*HostedRunnerPublicIP  `json:"public_ips,omitempty"`
+	LastActiveOn       *string                  `json:"last_active_on,omitempty"`
```
gmlewis, line 40: Question - would this be better as a `*Timestamp` rather than a `*string` ?

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+// HostedRunnerImageDetail represents the image details of a github-hosted runners.
```
gmlewis, line 43: In all comments, we try to use their official capitalization.

```suggestion
// HostedRunnerImageDetail represents the image details of a GitHub-hosted runners.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type HostedRunnerImageDetail struct {
+	ID   *string `json:"id,omitempty"`
+	Size *int    `json:"size,omitempty"`
+}
+
+// HostedRunners represents a collection of github-hosted runners for an organization.
```
gmlewis, line 49:
```suggestion
// HostedRunners represents a collection of GitHub-hosted runners for an organization.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type HostedRunners struct {
+	TotalCount int             `json:"total_count"`
+	Runners    []*HostedRunner `json:"runners"`
+}
+
+// ListHostedRunners lists all the github-hosted runners for an organization.
```
gmlewis, line 55:
```suggestion
// ListHostedRunners lists all the GitHub-hosted runners for an organization.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+//
+// GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#list-github-hosted-runners-for-an-organization
```
gmlewis, line 57: Since this one is a URL, we'll leave it as-is.

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+	MaximumRunners     *int64                   `json:"maximum_runners,omitempty"`
+	PublicIPEnabled    *bool                    `json:"public_ip_enabled,omitempty"`
+	PublicIPs          []*HostedRunnerPublicIP  `json:"public_ips,omitempty"`
+	LastActiveOn       *string                  `json:"last_active_on,omitempty"`
+}
+
+// HostedRunnerImageDetail represents the image details of a github-hosted runners.
+type HostedRunnerImageDetail struct {
+	ID   *string `json:"id,omitempty"`
+	Size *int    `json:"size,omitempty"`
```
gmlewis, line 46: Should this be an `*int64` ?
Could you please add a comment or update the comment on line 43 describing what the units are for this field?

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+	resp, err := s.client.Do(ctx, req, &runners)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return runners, resp, nil
+}
+
+// HostedRunnerImage represents the image of github-hosted runners
+// To list all available images, use GET /actions/hosted-runners/images/github-owned or GET /actions/hosted-runners/images/partner
```
gmlewis, line 82:
```suggestion
// HostedRunnerImage represents the image of GitHub-hosted runners.
// To list all available images, use `GET /actions/hosted-runners/images/github-owned` or
// `GET /actions/hosted-runners/images/partner`.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type CreateHostedRunnerRequest struct {
+	Name           string            `json:"name"`
+	Image          HostedRunnerImage `json:"image"`
+	RunnerGroupID  int64             `json:"runner_group_id"`
+	Size           string            `json:"size"`
+	MaximumRunners int64             `json:"maximum_runners,omitempty"`
+	EnableStaticIP bool              `json:"enable_static_ip,omitempty"`
+}
+
+// CreateHostedRunner creates a github-hosted runner for an organization.
```
gmlewis, line 99:
```suggestion
// CreateHostedRunner creates a GitHub-hosted runner for an organization.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+	hostedRunnerImages := new(HostedRunnerImages)
+	resp, err := s.client.Do(ctx, req, hostedRunnerImages)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return hostedRunnerImages, resp, nil
+}
+
+// HostedRunnerPublicIPLimits represents the static public IP limits for github-hosted runners.
```
gmlewis, line 177:
```suggestion
// HostedRunnerPublicIPLimits represents the static public IP limits for GitHub-hosted runners.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type HostedRunnerPublicIPLimits struct {
+	PublicIPs *PublicIPUsage `json:"public_ips"`
+}
+
+// PublicIPUsage provides details of static public IP limits for github-hosted runners.
```
gmlewis, line 182:
```suggestion
// PublicIPUsage provides details of static public IP limits for GitHub-hosted runners.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type PublicIPUsage struct {
+	Maximum      int `json:"maximum"`
+	CurrentUsage int `json:"current_usage"`
```
gmlewis, line 185: Should these be `int64` ?
Could you please add a comment or update the comment on line 182 describing what the units are for these fields?

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// HostedRunnerPublicIP represents the details of a public IP for github-hosted runner.
```
gmlewis, line 13:
```suggestion
// HostedRunnerPublicIP represents the details of a public IP for GitHub-hosted runner.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type HostedRunnerPublicIP struct {
+	Enabled bool   `json:"enabled"`
+	Prefix  string `json:"prefix"`
+	Length  int    `json:"length"`
```
gmlewis, line 17: Could you please add a comment stating what this field represents?
What is it the length of?

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+// HostedRunnerMachineSpec represents the details of a particular machine specification for github-hosted runner.
```
gmlewis, line 20:
```suggestion
// HostedRunnerMachineSpec represents the details of a particular machine specification for GitHub-hosted runner.
```

github/actions_hosted_runners.go
```diff
@@ -0,0 +1,333 @@
...
+type HostedRunnerMachineSpec struct {
+	ID        string `json:"id"`
+	CPUCores  int    `json:"cpu_cores"`
+	MemoryGB  int    `json:"memory_gb"`
+	StorageGB int    `json:"storage_gb"`
+}
+
+// HostedRunner represents a single github-hosted runner with additional details.
```
gmlewis, line 28:
```suggestion
// HostedRunner represents a single GitHub-hosted runner with additional details.
```

github/actions_hosted_runners_test.go
```diff
@@ -0,0 +1,789 @@
...
+	}
+
+	const methodName = "GetHostedRunnerGithubOwnedImages"
+	testBadOptions(t, methodName, func() (err error) {
+		_, _, err = client.Actions.GetHostedRunnerGithubOwnedImages(ctx, "\n")
+		return err
+	})
+
+	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
+		got, resp, err := client.Actions.GetHostedRunnerGithubOwnedImages(ctx, "o")
```
gmlewis, line 306:
```suggestion
		t.Errorf("Actions.GetHostedRunnerGitHubOwnedImages returned %+v, want %+v", hostedRunnerImages, want)
	}

	const methodName = "GetHostedRunnerGitHubOwnedImages"
	testBadOptions(t, methodName, func() (err error) {
		_, _, err = client.Actions.GetHostedRunnerGitHubOwnedImages(ctx, "\n")
		return err
	})

	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
		got, resp, err := client.Actions.GetHostedRunnerGitHubOwnedImages(ctx, "o")
```

gmlewis (changes requested):
Thank you, @atilsensalduz !
This is looking great!
There are a few minor tweaks, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/actions_hosted_runners.go
atilsensalduz: Hey @gmlewis , you want to remove the `#list-github-hosted-runners-for-an-organization` fragment since the base URL already directs to the same place, right?

github/actions_hosted_runners.go
atilsensalduz: Absolutely! I missed the Timestamp wrapper in the repository. Thank you for the review. I really appreciate it!

github/actions_hosted_runners.go
atilsensalduz: I noticed a discrepancy in the GitHub API documentation, the example response and the response schema don’t match. I based the struct definition on the response schema. Could you guide me on how to handle this situation?

Example Response: 
image

Response Schema:
image

https://docs.github.com/en/rest/actions/hosted-runners

github/actions_hosted_runners.go
gmlewis: I believe all GitHub API docs are now auto-generated, so please do not edit them.

github/actions_hosted_runners.go
gmlewis: When you find discrepancies, please contact GitHub Tech Support and report the issue, then report back here with your findings. Thank you.

We make an effort to ensure that the code works, and we obviously prefer that over matching broken API documentation. However, we also attempt to report broken API documentation in the hopes that it will be fixed.

atilsensalduz: Thanks @gmlewis , I've reported the issue. I'll share any findings as soon as I hear back.

https://github.com/github/docs/issues/36416

gmlewis (approved):
Thank you, @atilsensalduz!
LGTM.

stevehipwell (commented):
This looks great @atilsensalduz. My only question is about the distinction between the `CreateHostedRunnerRequest` and `UpdateHostedRunnerRequest`. AFAIK the idiom in the codebase is to use a single struct where possible so this differs from that pattern. Personally I like this pattern but I'm interested if @gmlewis has some overarching guidance here?

In this specific case the difference between the two models should be removed when the API catches up with the [UI behaviour](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/managing-larger-runners#changing-the-size-of-a-larger-runner). So maybe a shared `HostedRunnerRequest` struct with an error thrown if size is set for the update until the API supports it? Or maybe a comment in code to switch to aliases when the API matches. I guess the approach here probably comes back to the question above.

gmlewis:
> This looks great @atilsensalduz. My only question is about the distinction between the `CreateHostedRunnerRequest` and `UpdateHostedRunnerRequest`. AFAIK the idiom in the codebase is to use a single struct where possible so this differs from that pattern. Personally I like this pattern but I'm interested if @gmlewis has some overarching guidance here?

Good catch, @stevehipwell.

We attempt to minimize the creation of extra structs and reuse similar structs for similar use cases with comments where necessary to explain usage.

atilsensalduz:
> This looks great @atilsensalduz. My only question is about the distinction between the `CreateHostedRunnerRequest` and `UpdateHostedRunnerRequest`. AFAIK the idiom in the codebase is to use a single struct where possible so this differs from that pattern. Personally I like this pattern but I'm interested if @gmlewis has some overarching guidance here?
> 
> In this specific case the difference between the two models should be removed when the API catches up with the [UI behaviour](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/managing-larger-runners#changing-the-size-of-a-larger-runner). So maybe a shared `HostedRunnerRequest` struct with an error thrown if size is set for the update until the API supports it? Or maybe a comment in code to switch to aliases when the API matches. I guess the approach here probably comes back to the question above.

@stevehipwell  Thank you so much for the review and the excellent suggestion. I’ve implemented it along with the additional validations. I really appreciate your valuable input!

gmlewis:
@atilsensalduz - please run step 4 of CONTRIBUTING.md and make sure the linter and all tests pass locally, then push the changes to this PR. Thank you!

github/actions_hosted_runners.go
```diff
@@ -89,22 +90,52 @@ type HostedRunnerImage struct {
...
 // CreateHostedRunner creates a GitHub-hosted runner for an organization.
 //
 // GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#create-a-github-hosted-runner-for-an-organization
 //
 //meta:operation POST /orgs/{org}/actions/hosted-runners
-func (s *ActionsService) CreateHostedRunner(ctx context.Context, org string, request *CreateHostedRunnerRequest) (*HostedRunner, *Response, error) {
+func (s *ActionsService) CreateHostedRunner(ctx context.Context, org string, request *HostedRunnerRequest) (*HostedRunner, *Response, error) {
+	err := validateCreateHostedRunnerRequest(request)
+	if err != nil {
+		return nil, nil, errors.New("validation failed: " + err.Error())
```
gmlewis, line 136:
```suggestion
		return nil, nil, errors.New("validation failed: %w", err)
```

github/actions_hosted_runners.go
```diff
@@ -284,21 +315,32 @@ func (s *ActionsService) GetHostedRunner(ctx context.Context, org string, runner
...
 // UpdateHostedRunner updates a GitHub-hosted runner for an organization.
 //
 // GitHub API docs: https://docs.github.com/rest/actions/hosted-runners#update-a-github-hosted-runner-for-an-organization
 //
 //meta:operation PATCH /orgs/{org}/actions/hosted-runners/{hosted_runner_id}
-func (s *ActionsService) UpdateHostedRunner(ctx context.Context, org string, runnerID int64, updateReq UpdateHostedRunnerRequest) (*HostedRunner, *Response, error) {
+func (s *ActionsService) UpdateHostedRunner(ctx context.Context, org string, runnerID int64, updateReq HostedRunnerRequest) (*HostedRunner, *Response, error) {
+	err := validateUpdateHostedRunnerRequest(&updateReq)
+	if err != nil {
+		return nil, nil, errors.New("validation failed: " + err.Error())
```
gmlewis, line 341:
```suggestion
		return nil, nil, errors.New("validation failed: %w", err)
```

github/enterprise_actions_hosted_runners.go
```diff
@@ -41,7 +42,12 @@ func (s *EnterpriseService) ListHostedRunners(ctx context.Context, enterprise st
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/actions/hosted-runners#create-a-github-hosted-runner-for-an-enterprise
 //
 //meta:operation POST /enterprises/{enterprise}/actions/hosted-runners
-func (s *EnterpriseService) CreateHostedRunner(ctx context.Context, enterprise string, request *CreateHostedRunnerRequest) (*HostedRunner, *Response, error) {
+func (s *EnterpriseService) CreateHostedRunner(ctx context.Context, enterprise string, request *HostedRunnerRequest) (*HostedRunner, *Response, error) {
+	err := validateCreateHostedRunnerRequest(request)
+	if err != nil {
```
gmlewis, line 47:
```suggestion
	if err := validateCreateHostedRunnerRequest(request); err != nil {
```

github/enterprise_actions_hosted_runners.go
```diff
@@ -41,7 +42,12 @@ func (s *EnterpriseService) ListHostedRunners(ctx context.Context, enterprise st
...
+		return nil, nil, errors.New("validation failed: " + err.Error())
```
gmlewis, line 48:
```suggestion
		return nil, nil, errors.New("validation failed: %w", err)
```

github/enterprise_actions_hosted_runners.go
```diff
@@ -188,7 +194,12 @@ func (s *EnterpriseService) GetHostedRunner(ctx context.Context, enterprise stri
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/actions/hosted-runners#update-a-github-hosted-runner-for-an-enterprise
 //
 //meta:operation PATCH /enterprises/{enterprise}/actions/hosted-runners/{hosted_runner_id}
-func (s *EnterpriseService) UpdateHostedRunner(ctx context.Context, enterprise string, runnerID int64, updateReq UpdateHostedRunnerRequest) (*HostedRunner, *Response, error) {
+func (s *EnterpriseService) UpdateHostedRunner(ctx context.Context, enterprise string, runnerID int64, updateReq HostedRunnerRequest) (*HostedRunner, *Response, error) {
+	err := validateUpdateHostedRunnerRequest(&updateReq)
+	if err != nil {
```
gmlewis, line 199:
```suggestion
	if err := validateUpdateHostedRunnerRequest(&updateReq); err != nil {
```

github/enterprise_actions_hosted_runners.go
```diff
@@ -188,7 +194,12 @@ func (s *EnterpriseService) GetHostedRunner(ctx context.Context, enterprise stri
...
+		return nil, nil, errors.New("validation failed: " + err.Error())
```
gmlewis, line 200:
```suggestion
		return nil, nil, errors.New("validation failed: %w", err)
```

gmlewis (changes requested):
Thank you, @atilsensalduz!
After a few tweaks, please, we should be ready for a second LGTM+Approval from any other contributor to this repo.

atilsensalduz:
@gmlewis and @stevehipwell Thank you so much for reviewing and providing such valuable feedback. It was a pleasure working with you, and I’ve learned a lot throughout the process. I’m looking forward to continuing to contribute to the repository!

gmlewis (approved):
Thank you, @atilsensalduz and @stevehipwell!
LGTM.
Merging.

Issue #3466, opened 2025-02-07
UpdatePages method doesn't work on github enterprise server (CNAME argument not allowed)


gk-drw:
The update pages method (UpdatePages) doesn't work on github enterprise server, causing the error 

"400 Custom domains are not available for GitHub Pages"

An example can be seen here: https://github.com/integrations/terraform-provider-github/issues/2087

The update struct  doesn't specify "omitempty" for its json marshalling. As a result, the client will send an explicit null value. 

Unfortunately, github is inconsistent in how it treats this value on GHES vs GitHub proper (cloud) - specifying any value including `null` is not allowed in GHES.

atilsensalduz:
Hey everyone,

I’ve implemented a workaround to fix the issue, but I’m not entirely sure if it's the best approach. Since I’m still learning, I feel there might be some code smells, and I’d appreciate any feedback or suggestions for improvement. I’m just trying my best to improve and contribute!

Thanks a lot!

PR #3489, opened 2025-02-21
fix: Resolve '400 Custom domains' error on GitHub Enterprise Server

atilsensalduz:
This PR fixes an issue with the UpdatePages method, which caused the error: "400 Custom domains are not available for GitHub Pages" on GHES. The issue occurred because the update struct didn't specify the omitempty tag for JSON marshalling, which resulted in explicit null values being sent. As a result, GHES returned the error.

To resolve this, the fix adds a check for the client URL. If the URL differs from the public GitHub API, a new struct is created without the CNAME field to avoid sending null values.

Question: Should we add the omitempty tag directly to the CNAME field of the PagesUpdate struct instead of this workaround?

References:
- Issue: https://github.com/google/go-github/issues/3466

Fixes: #3466.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -16734,6 +16734,38 @@ func (p *PagesUpdate) GetSource() *PagesSource {
 	return p.Source
 }
 
+// GetBuildType returns the BuildType field if it's non-nil, zero value otherwise.
+func (p *PagesUpdateWithoutCNAME) GetBuildType() string {
+	if p == nil || p.BuildType == nil {
+		return ""
+	}
+	return *p.BuildType
+}
+
+// GetHTTPSEnforced returns the HTTPSEnforced field if it's non-nil, zero value otherwise.
+func (p *PagesUpdateWithoutCNAME) GetHTTPSEnforced() bool {
+	if p == nil || p.HTTPSEnforced == nil {
+		return false
+	}
+	return *p.HTTPSEnforced
+}
+
+// GetPublic returns the Public field if it's non-nil, zero value otherwise.
+func (p *PagesUpdateWithoutCNAME) GetPublic() bool {
+	if p == nil || p.Public == nil {
+		return false
+	}
+	return *p.Public
+}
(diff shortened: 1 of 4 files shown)


gmlewis:
What happens when a person wishes to remove a custom domain?

When this happens for other endpoints (where sending a "null" is important and has a specific meaning), we often introduce a second endpoint that has serves the other usage, and whose name highlights the difference between the two use cases.

I think that might be appropriate here.  

If you look around, you should be able to find some endpoints like that, but if you can't, let me know and I'll dig around for examples of this.

atilsensalduz:
> What happens when a person wishes to remove a custom domain?
> 
> When this happens for other endpoints (where sending a "null" is important and has a specific meaning), we often introduce a second endpoint that has serves the other usage, and whose name highlights the difference between the two use cases.
> 
> I think that might be appropriate here.
> 
> If you look around, you should be able to find some endpoints like that, but if you can't, let me know and I'll dig around for examples of this.

You're absolutely right. Thanks for guiding me toward the better approach. I've updated the PR to use a separate endpoint for updating pages specifically for GHES and to use a struct without the CNAME field to avoid the 400 error.

gmlewis (approved):
Thank you, @atilsensalduz!
LGTM.

@stevehipwell - might you have time for a coffee review? Thank you!

gmlewis (changes requested):
Whups, I spoke too soon. Please fix the lint issues by following step 4 of CONTRIBUTING.md.

atilsensalduz:
> Whups, I spoke too soon. Please fix the lint issues by following step 4 of CONTRIBUTING.md.

Ah, sorry about that! I must have missed it. 🤦  Thanks for the heads-up! I've fixed the lint issues now. I really appreciate your review and support! @gmlewis

stevehipwell (approved):
LGTM

gmlewis:
Thank you, @stevehipwell!
Merging.

PR #3495, opened 2025-02-25
Correct the `runIDFromURLRE` regex to properly match the callbackURL

pputman-clabs:
the callbackURL from the event previously didn't match with the regex, and was always causing a "no match found error".  This includes the full URL and should match.

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -3117,6 +3117,19 @@ func TestDeploymentProtectionRuleEvent_GetRunID(t *testing.T) {
 		t.Errorf("want %#v, got %#v", want, got)
 	}
 
+	want = 123456789
+	url = "https://local.dummyhost.net/repos/dummy-org/dummy-repo/actions/runs/123456789/deployment_protection_rule"
+
+	e = DeploymentProtectionRuleEvent{
+		DeploymentCallbackURL: &url,
+	}
+
+	got, _ = e.GetRunID()
+	if got != want {
+		t.Errorf("want %#v, got %#v", want, got)
+	}
+
+
 	want = -1
 	url = "https://api.github.com/repos/dummy-org/dummy-repo/actions/runs/abc123/deployment_protection_rule"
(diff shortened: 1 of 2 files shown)


gmlewis (changes requested):
I should have caught this before (sorry about that), but this repo supports both public GitHub and enterprise GitHub installations, which means that we can't hard-code a domain name in a regex.

Instead of your change, we should simply remove the leading carat (`^`) and then it should support both styles of URLs.

Please update this PR and then add a test for both styles of URLs showing that it should work on both. Thanks you, @pputman-clabs.

github/github_test.go
```diff
@@ -3117,9 +3117,21 @@ func TestDeploymentProtectionRuleEvent_GetRunID(t *testing.T) {
 		t.Errorf("want %#v, got %#v", want, got)
 	}
 
-	want = -1
-	url = "repos/dummy-org/dummy-repo/actions/runs/abc123/deployment_protection_rule"
+	want = 123456789
+	url = "https://local.dummyhost.net/repos/dummy-org/dummy-repo/actions/runs/123456789/deployment_protection_rule"
```
gmlewis, line 3121: Actually, on GitHub Enterprise, the URL comes back in the event without the base URL, like the following, so please change this line to represent that.

```suggestion
	// GitHub Enterprise returns the URL in events without the leading domain:
	url = "repos/dummy-org/dummy-repo/actions/runs/123456789/deployment_protection_rule"
```

gmlewis (changes requested):
One more tweak, please, @pputman-clabs, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

gmlewis (approved):
Thank you, @pputman-clabs!
LGTM.

gmlewis:

stevehipwell (approved):
LGTM

gmlewis:
Thank you, @stevehipwell!
Merging.

Issue #3488, opened 2025-02-21
Enterprise API responses are sometimes returning arrays of strings instead of a string


antixcode6:
### Context

go version go1.24.0 darwin/arm64
Github Enterprise REST API
go-github versions v66 and v69 tested with

### The issue
While pulling logs from the GH enterprise REST API via `client.Enterprise.GetAuditLog(ctx, enterprise-name, &opts)` it seems that there are some issues with actual API response vs the Github provided response schema for `/enterprises/{enterprise}/audit-log`

The docs state org should be a string

https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/audit-log?apiVersion=2022-11-28#get-the-audit-log-for-an-enterprise

```json
      "org": {
        "type": "string"
      }
```

However, while pulling audit logs from an enterprise account the org field may sometimes be an array of strings.

```json
{
  "@timestamp" : 1739983846824,
  "_document_id" : "<redacted>",
  "action" : "oauth_authorization.create",
  "actor" : "<redacted>",
  "actor_id" : 123456789,
  "actor_ip" : "<redacted>",
  "actor_is_bot" : false,
  "actor_location" : {
    "country_code" : "GB"
  },
  "business" : "<redacted>",
  "business_id" : 123456,
  "created_at" : 123456789,
  "external_identity_username" : "<redacted>",
  "operation_type" : "create",
  "org" : [ "<redacted>", "<redacted>", "<redacted>" ],
  "org_id" : [ 123456789, 123456789, 123456789 ],
  "request_access_security_header" : null,
  "token_scopes" : "gist,repo,workflow",
  "user" : "<redacted>",
  "user_agent" : "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36 Edg/133.0.0.0",
  "user_id" : 123456789
}
```

This is causing the following error when trying to unmarshal the response into a struct `json: cannot unmarshal array into Go struct field entryAlias.org of type string`

This error is present in v66 and v69, I've tested and was able to replicate the error with both versions. 

Is this something this package may be able to flex? I wonder if the array can be turned into a string with `strings.Join()` when an array is detected during the type switch in this function https://github.com/google/go-github/blob/master/github/github.go#L1024-L1045

Specifically it is failing here 

```go
	default:
		decErr := json.NewDecoder(resp.Body).Decode(v)
		if decErr == io.EOF {
[...]

PR #3502, opened 2025-03-05
feat!: Change AuditEntry.Org and AuditEntry.OrgID to json.RawMessage

antixcode6:
This PR is meant to resolve the issues around https://github.com/google/go-github/issues/3488

I am modifying `AuditEntry.Org` and `AuditEntry.OrgID` from `*string` and `*int64`, respectively, to both be `json.RawMessage` types this is meant to allow the consumer of this client to test the fields for being a string/int64 or being a []string/[]int64 and handle it accordingly.

This PR also modifies the tests `TestOrganizationService_GetAuditLog` `TestAuditEntry_Marshal` and `TestEnterpriseService_GetAuditLog` to ensure functionality remains after changing the above mentioned fields to `json.RawMessage`

I have also added two methods, `AuditEntry.GetOrgID` and `AuditEntry.GetOrg`,  to `skipStructMethods` and ran the code gen to remove the tests and methods to make sure we do not need to modify the generated tests/accessor methods.

### DIFF
--- a/github/orgs_audit_log.go
+++ b/github/orgs_audit_log.go
@@ -30,24 +30,24 @@ type ActorLocation struct {
 // in AdditionalFields.
 // For a list of actions see - https://docs.github.com/github/setting-up-and-managing-organizations-and-teams/reviewing-the-audit-log-for-your-organization#audit-log-actions
 type AuditEntry struct {
-	Action                   *string        `json:"action,omitempty"` // The name of the action that was performed, for example `user.login` or `repo.create`.
-	Actor                    *string        `json:"actor,omitempty"`  // The actor who performed the action.
-	ActorID                  *int64         `json:"actor_id,omitempty"`
-	ActorLocation            *ActorLocation `json:"actor_location,omitempty"`
-	Business                 *string        `json:"business,omitempty"`
-	BusinessID               *int64         `json:"business_id,omitempty"`
-	CreatedAt                *Timestamp     `json:"created_at,omitempty"`
-	DocumentID               *string        `json:"_document_id,omitempty"`
-	ExternalIdentityNameID   *string        `json:"external_identity_nameid,omitempty"`
-	ExternalIdentityUsername *string        `json:"external_identity_username,omitempty"`
-	HashedToken              *string        `json:"hashed_token,omitempty"`
-	Org                      *string        `json:"org,omitempty"`
-	OrgID                    *int64         `json:"org_id,omitempty"`
-	Timestamp                *Timestamp     `json:"@timestamp,omitempty"` // The time the audit log event occurred, given as a [Unix timestamp](http://en.wikipedia.org/wiki/Unix_time).
-	TokenID                  *int64         `json:"token_id,omitempty"`
-	TokenScopes              *string        `json:"token_scopes,omitempty"`
-	User                     *string        `json:"user,omitempty"` // The user that was affected by the action performed (if available).
-	UserID                   *int64         `json:"user_id,omitempty"`
+	Action                   *string         `json:"action,omitempty"` // The name of the action that was performed, for example `user.login` or `repo.create`.
+	Actor                    *string         `json:"actor,omitempty"`  // The actor who performed the action.
+	ActorID                  *int64          `json:"actor_id,omitempty"`
+	ActorLocation            *ActorLocation  `json:"actor_location,omitempty"`
+	Business                 *string         `json:"business,omitempty"`
+	BusinessID               *int64          `json:"business_id,omitempty"`
+	CreatedAt                *Timestamp      `json:"created_at,omitempty"`
+	DocumentID               *string         `json:"_document_id,omitempty"`
+	ExternalIdentityNameID   *string         `json:"external_identity_nameid,omitempty"`
+	ExternalIdentityUsername *string         `json:"external_identity_username,omitempty"`
+	HashedToken              *string         `json:"hashed_token,omitempty"`
+	Org                      json.RawMessage `json:"org,omitempty"`
+	OrgID                    json.RawMessage `json:"org_id,omitempty"`
+	Timestamp                *Timestamp      `json:"@timestamp,omitempty"` // The time the audit log event occurred, given as a [Unix timestamp](http://en.wikipedia.org/wiki/Unix_time).
+	TokenID                  *int64          `json:"token_id,omitempty"`
+	TokenScopes              *string         `json:"token_scopes,omitempty"`
+	User                     *string         `json:"user,omitempty"` // The user that was affected by the action performed (if available).
+	UserID                   *int64          `json:"user_id,omitempty"`
 
 	// Some events types have a data field that contains additional information about the event.
 	Data map[string]interface{} `json:"data,omitempty"`
--- a/github/enterprise_audit_log_test.go
+++ b/github/enterprise_audit_log_test.go
@@ -7,6 +7,7 @@ package github
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"testing"
@@ -60,7 +61,7 @@ func TestEnterpriseService_GetAuditLog(t *testing.T) {
 			Action:     Ptr("workflows.completed_workflow_run"),
 			Actor:      Ptr("testactor"),
 			CreatedAt:  &Timestamp{timestamp},
-			Org:        Ptr("o"),
+			Org:        json.RawMessage(`"o"`),
 			AdditionalFields: map[string]interface{}{
 				"completed_at":    "2021-03-07T00:35:08.000Z",
 				"conclusion":      "success",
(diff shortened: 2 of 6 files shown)


gmlewis:
@antixcode6 - this looks like a great start!  Why not add the new getter methods to this PR to keep it all in the same place?

I'm thinking that you can add replacements for the original `AuditEntry.GetOrgID` and `AuditEntry.GetOrg` methods with the  same result types as before (but with the added `ok` field if it is indeed the expected type), and then add two new methods with your new getters that support the alternative type(s) that the fields represent.

Thoughts?

antixcode6:
@gmlewis sounds good! I just pushed up some changes with four getters, two raw getters and two for the same result types as before with the added ok. 

Also I created a test to make sure these getters are covered.

gmlewis (approved):
Thank you, @antixcode6!
Did you want to add new getters for the other types being returned that you discovered?

Otherwise, LGTM.

antixcode6:
@gmlewis this covers both Org and OrgID which are the two I discovered to be returning arrays sometimes, otherwise I haven't noticed anything else that would need to be modified

gmlewis:
> @gmlewis this covers both Org and OrgID which are the two I discovered to be returning arrays sometimes, otherwise I haven't noticed anything else that would need to be modified

So you don't want to add getters that return `([]string, bool)` and `([]int64, bool)` respectively?

github/orgs_audit_log.go
```diff
@@ -68,6 +68,18 @@ func (a *AuditEntry) GetOrg() (org string, ok bool) {
 	return org, true
 }
 
+// GetOrgSlice returns the OrgID field, as a slice of int64, if it's valid.
+func (a *AuditEntry) GetOrgSlice() (orgID []string, ok bool) {
```
gmlewis, line 72: How about this?

```suggestion
// GetOrgIDStrings returns the OrgID field, as a slice of strings, if it's valid.
func (a *AuditEntry) GetOrgIDStrings() (orgIDs []string, ok bool) {
```

github/orgs_audit_log.go
```diff
@@ -89,6 +101,18 @@ func (a *AuditEntry) GetOrgID() (orgID int64, ok bool) {
 	return orgID, true
 }
 
+// GetOrgIDSlice returns the OrgID field, as a slice of int64, if it's valid.
+func (a *AuditEntry) GetOrgIDSlice() (orgID []int64, ok bool) {
```
gmlewis, line 105:
```suggestion
// GetOrgIDInt64s returns the OrgID field, as a slice of int64s, if it's valid.
func (a *AuditEntry) GetOrgIDIny64s() (orgIDs []int64, ok bool) {
```

github/orgs_audit_log.go
```diff
@@ -68,6 +68,18 @@ func (a *AuditEntry) GetOrg() (org string, ok bool) {
 	return org, true
 }
 
+// GetOrgSlice returns the OrgID field, as a slice of int64, if it's valid.
+func (a *AuditEntry) GetOrgSlice() (orgID []string, ok bool) {
+	if a == nil || a.OrgID == nil {
+		return []string{}, false
```
gmlewis, line 74:
```suggestion
		return nil, false
```

github/orgs_audit_log.go
```diff
@@ -68,6 +68,18 @@ func (a *AuditEntry) GetOrg() (org string, ok bool) {
...
+	}
+	if err := json.Unmarshal([]byte(a.OrgID), &orgID); err != nil {
+		return []string{}, false
```
gmlewis, line 77:
```suggestion
		return nil, false
```

github/orgs_audit_log.go
```diff
@@ -89,6 +101,18 @@ func (a *AuditEntry) GetOrgID() (orgID int64, ok bool) {
 	return orgID, true
 }
 
+// GetOrgIDSlice returns the OrgID field, as a slice of int64, if it's valid.
+func (a *AuditEntry) GetOrgIDSlice() (orgID []int64, ok bool) {
+	if a == nil || a.OrgID == nil {
+		return []int64{}, false
```
gmlewis, line 107:
```suggestion
		return nil, false
```

github/orgs_audit_log.go
```diff
@@ -89,6 +101,18 @@ func (a *AuditEntry) GetOrgID() (orgID int64, ok bool) {
...
+	}
+	if err := json.Unmarshal([]byte(a.OrgID), &orgID); err != nil {
+		return []int64{}, false
```
gmlewis, line 110:
```suggestion
		return nil, false
```

gmlewis (changes requested):

antixcode6:
@gmlewis good suggestions I've made the changes, I also had to fix something in my test case. All should be good now

github/orgs_audit_log.go
```diff
@@ -68,16 +68,16 @@ func (a *AuditEntry) GetOrg() (org string, ok bool) {
 	return org, true
 }
 
-// GetOrgSlice returns the OrgID field, as a slice of int64, if it's valid.
-func (a *AuditEntry) GetOrgSlice() (orgID []string, ok bool) {
-	if a == nil || a.OrgID == nil {
-		return []string{}, false
+// GetOrgStrings returns the Org field, as a slice of strings, if it's valid.
+func (a *AuditEntry) GetOrgStrings() (org []string, ok bool) {
```
gmlewis, line 72: Shouldn't `org` be `orgs` (plural) here since it is a slice?
Or maybe `names` if they all represent the same org?

github/orgs_audit_log.go
```diff
@@ -101,16 +101,16 @@ func (a *AuditEntry) GetOrgID() (orgID int64, ok bool) {
 	return orgID, true
 }
 
-// GetOrgIDSlice returns the OrgID field, as a slice of int64, if it's valid.
-func (a *AuditEntry) GetOrgIDSlice() (orgID []int64, ok bool) {
+// GetOrgIDInt64s returns the OrgID field, as a slice of int64, if it's valid.
```
gmlewis, line 104: Now that we don't have `GetOrgIDStrings`, this might be much nicer as `GetOrgIDs`.

github/orgs_audit_log.go
gmlewis: And if you go with `names`, then maybe it should be `GetOrgNames`?

antixcode6: Names (plural) makes the most sense as the slice returned from the API was multiple orgs not the same org multiple times

gmlewis (approved):
Thank you, @antixcode6!
LGTM.

stevehipwell (commented):
The implementation looks good, but I'm not a fan of `json.RawMessage`. Did you consider using a custom marshaller and adding `Orgs` & `OrgIDs` to take the array versions? I did this or something similar for the ruleset APIs.

gmlewis:
> The implementation looks good, but I'm not a fan of `json.RawMessage`. Did you consider using a custom marshaller and adding `Orgs` & `OrgIDs` to take the array versions? I did this or something similar for the ruleset APIs.

Do you want to create an alternative PR and then we can look at the two implementations and decide which one we prefer?

stevehipwell:
@gmlewis sorry but I don't think I have the bandwidth to pick up a PR for this now. This is probably a wider discussion than just this PR as GitHub APIs often include types that can't be directly represented by the Go type system.

After re-reading the issue I have a slightly different view. The schema is incorrect, which needs to be reported at [github/rest-api-description](https://github.com/github/rest-api-description) (but I wouldn't hold your breath as it looks as much of a ghost town as the TF provider). And for the implementation as we don't need to directly follow the GitHub API, I think the struct should have `Orgs` & `OrgIDs` fields both as arrays; these can be marshalled from either a single string or an array to result in a consistent representation where the end user doesn't need to understand how to extract the data.

antixcode6:
@gmlewis @stevehipwell I'm open to this, would it make sense to open a separate PR with with the Orgs and OrgIDs fields as part of the struct, and then we can pick the more appropriate change? I have no problem creating a new branch and PR but if it is better to work out of here I am ok with that too.

gmlewis:
> @gmlewis @stevehipwell I'm open to this, would it make sense to open a separate PR with with the Orgs and OrgIDs fields as part of the struct, and then we can pick the more appropriate change? I have no problem creating a new branch and PR but if it is better to work out of here I am ok with that too.

Sure, a separate PR makes sense to me. Thank you!

gmlewis:
Status update: my understanding is that we will not be moving forward with this PR but instead, @antixcode6 will create a new PR that has `Orgs` & `OrgIDs` fields both as arrays.

Shall I close this PR, @antixcode6?

antixcode6:
@gmlewis Hey Glenn, apologies I've been quite busy on my end but yes that is correct I will submit a new PR when I have some time this one is good to be closed

gmlewis:
Thank you, @antixcode6!
Closing.

Issue #2033, opened 2021-08-08
`package` event outdated


mahansky:
[Package](https://github.com/google/go-github/blob/master/github/packages.go) struct seems to be outdated (probably) after the migration from Docker registry to Container registry.
I can't find any information about this [GH docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#package), the example there is still showing the old format.
Here is a small comparison between the two (sent by webhook with `Packages` event selected):

<details>
  <summary>Old format</summary>

  ```json
  {
      "id": 1,
      "name": "packagename",
      "namespace": "org",
      "description": null,
      "ecosystem": "docker",
      "package_type": "docker",
      "html_url": "url",
      "created_at": "2021-06-14T20:58:10Z",
      "updated_at": "2021-06-14T20:58:10Z",
      "owner": {},
      "package_version": {
          "id": 1,
          "version": "v1.0.53",
          "name": "sha256:547ab5a9317ef70c02e7dd50dace3bb3ea5027eda4c20ffee2dcc58547a880e9",
          "description": "",
          "summary": "",
          "body": "",
          "body_html": "",
[...]

PR #3515, opened 2025-03-13
refactor!: Update package types to align with webhook event

adrian-saunders:
BREAKING CHANGE: `PackageVersion.Body` and `PackageVersion.Metadata` are both now `json.RawMessage`.

Hello,

Firstly, thank you for maintaining this project! It has made interacting with GitHub significantly easier. I know this is a little backward to raise a PR without even a comment on the original issue so sorry but I had to get some legal approval first. Anyway...

This PR is meant to better align the package types with the webhook event schema.

Fixes: #2033

Breaking changes:
* `PackageVersion.Body` is now `json.RawMessage` - I'm not sure in what scenarios this is a string or object but it can be both.
* `PackageVersion.Metadata` is now `json.json.RawMessage` - webhooks return an array here, rather than a `PackageMetadata`

I've added getters for both of these fields.

References:
Webhooks: https://github.com/octokit/webhooks/tree/main/payload-schemas/api.github.com/package
REST: https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28

Please do let me know your comments and if there is anything I haven't thought of!


github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+	HTMLURL             *string                        `json:"html_url,omitempty"`
+	TagName             *string                        `json:"tag_name,omitempty"`
+	TargetCommitish     *string                        `json:"target_commitish,omitempty"`
+	TargetOID           *string                        `json:"target_oid,omitempty"`
+	Draft               *bool                          `json:"draft,omitempty"`
+	Prerelease          *bool                          `json:"prerelease,omitempty"`
+	ContainerMetadata   *PackageEventContainerMetadata `json:"container_metadata,omitempty"`
+	DockerMetadata      *[]interface{}                 `json:"docker_metadata,omitempty"`
+	NPMMetadata         *PackageNPMMetadata            `json:"npm_metadata,omitempty"`
+	NugetMetadata       *[]PackageNugetMetadata        `json:"nuget_metadata,omitempty"`
```
gmlewis, line 67:
```suggestion
	NugetMetadata       []*PackageNugetMetadata        `json:"nuget_metadata,omitempty"`
```

github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+	Author              *User                          `json:"author,omitempty"`
+	SourceURL           *string                        `json:"source_url,omitempty"`
+	InstallationCommand *string                        `json:"installation_command,omitempty"`
+
+	// The following are NOT populated for webhook events
+	DeletedAt *Timestamp `json:"deleted_at,omitempty"`
+}
+
+// GetBody returns the body field as a string if it's valid
+func (pv *PackageVersion) GetBody() (body *string, ok bool) {
```
gmlewis, line 80:
```suggestion
func (pv *PackageVersion) GetBody() (body string, ok bool) {
```

github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+	if pv == nil || pv.Body == nil {
+		return nil, true
```
gmlewis, line 82:
```suggestion
		return "", false
```

github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+	}
+
+	if err := json.Unmarshal(pv.Body, &body); err != nil {
+		return nil, false
```
gmlewis, line 86:
```suggestion
		return "", false
```

github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+	}
+
+	return body, true
+}
+
+// GetBodyAsPackageVersionBody returns the body field as a PackageVersionBody if it's valid
+func (pv *PackageVersion) GetBodyAsPackageVersionBody() (body *PackageVersionBody, ok bool) {
+	if pv == nil || pv.Body == nil {
+		return nil, true
```
gmlewis, line 95:
```suggestion
		return nil, false
```

github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+		return nil, false
+	}
+
+	return body, true
+}
+
+// GetMetadata returns the metadata field as PackageMetadata if it's valid
+func (pv *PackageVersion) GetMetadata() (metadata *PackageMetadata, ok bool) {
+	if pv == nil || pv.Metadata == nil {
+		return nil, true
```
gmlewis, line 108:
```suggestion
		return nil, false
```

github/packages.go
```diff
@@ -85,7 +157,6 @@ type PackageFile struct {
 	MD5         *string    `json:"md5,omitempty"`
 	ContentType *string    `json:"content_type,omitempty"`
 	State       *string    `json:"state,omitempty"`
-	Author      *User      `json:"author,omitempty"`
```
gmlewis, line 88: Why is this field being removed?

github/packages.go
```diff
@@ -141,3 +212,106 @@ type PackageContainerMetadata struct {
...
+
+func (b PackageVersionBody) String() string {
+	return Stringify(b)
+}
+
+// PackageVersionBodyInfo represents the info field of a PackageVersionBody
+type PackageVersionBodyInfo struct {
+	Type       *string `json:"type,omitempty"`
+	OID        *string `json:"oid,omitempty"`
+	Mode       *int64  `json:"mode,omitempty"`
```
gmlewis, line 230: Why is "mode" an `int64`? That seems really unusual. What is a "mode"?

github/packages.go
```diff
@@ -141,3 +212,106 @@ type PackageContainerMetadata struct {
...
+	Version              *string            `json:"version,omitempty"`
+	NPMUser              *string            `json:"npm_user,omitempty"`
+	Author               *map[string]string `json:"author,omitempty"`
+	Bugs                 *map[string]string `json:"bugs,omitempty"`
+	Dependencies         *map[string]string `json:"dependencies,omitempty"`
+	DevDependencies      *map[string]string `json:"dev_dependencies,omitempty"`
+	PeerDependencies     *map[string]string `json:"peer_dependencies,omitempty"`
+	OptionalDependencies *map[string]string `json:"optional_dependencies,omitempty"`
+	Description          *string            `json:"description,omitempty"`
+	Dist                 *map[string]string `json:"dist,omitempty"`
```
gmlewis, line 286: All of the `*map[string]string` should be `map[string]string` as maps are already reference types.

Same below.
I'm going to stop my review and let you address the issues found thus far.

github/packages.go
```diff
@@ -5,21 +5,31 @@
...
+	HTMLURL     *string    `json:"html_url,omitempty"`
+	Visibility  *string    `json:"visibility,omitempty"`
+	Owner       *User      `json:"owner,omitempty"`
+	CreatedAt   *Timestamp `json:"created_at,omitempty"`
+	UpdatedAt   *Timestamp `json:"updated_at,omitempty"`
+
+	// The following are only populated for webhook events
+	Namespace      *string          `json:",omitempty"`
+	Description    *string          `json:",omitempty"`
+	Ecosystem      *string          `json:",omitempty"`
```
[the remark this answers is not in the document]
adrian-saunders: Too much copy and paste!

github/packages.go
```diff
@@ -28,28 +38,90 @@ func (p Package) String() string {
...
+	Release             *PackageRelease                `json:"release,omitempty"`
+	Manifest            *string                        `json:"manifest,omitempty"`
+	HTMLURL             *string                        `json:"html_url,omitempty"`
+	TagName             *string                        `json:"tag_name,omitempty"`
+	TargetCommitish     *string                        `json:"target_commitish,omitempty"`
+	TargetOID           *string                        `json:"target_oid,omitempty"`
+	Draft               *bool                          `json:"draft,omitempty"`
+	Prerelease          *bool                          `json:"prerelease,omitempty"`
+	ContainerMetadata   *PackageEventContainerMetadata `json:"container_metadata,omitempty"`
+	DockerMetadata      *[]interface{}                 `json:"docker_metadata,omitempty"`
```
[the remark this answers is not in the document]
adrian-saunders: Apologies! 101 of how to look dumb on the internet 😭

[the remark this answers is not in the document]
gmlewis: No worries!

github/packages.go
adrian-saunders: Makes sense, my thought process here was that `ok` represents if `Body` was valid/unmarshaling didn't fail. In this case the value is just `nil` but that's still a valid response from GitHub.

Either way, as a consumer, you'd still then have to check for nil which probably isn't that useful

github/packages.go
adrian-saunders: As far as I can tell, `PackageFile`s only exist on a webhook event and the schema doesn't have it as a property

https://github.com/octokit/webhooks/blob/main/payload-schemas/api.github.com/package/published.schema.json#L227-L238

Should have mentioned as a breaking change though!

github/packages.go
adrian-saunders: https://github.com/octokit/webhooks/blob/main/payload-schemas/api.github.com/package/updated.schema.json#L106

Not sure what it actually represents, my guess was that it represented a files mode/permission bits like `os.FileMode`

Here's an example from an event payload we've received
``` json
{
  "type":"blob"
  "oid":"6475613a1081b7e1c42b57cc3eb6ed161225a978"
  "mode":33188
  "name":"README.md"
  "path":"README.md"
  "size":NULL
  "collection":true
}
```

github/packages.go
gmlewis: Why is this field being removed?

github/packages.go
gmlewis: Why is "mode" an int64? That seems really unusual. What is a "mode"?

gmlewis:
Oh, sorry, I see your responses now in email.
I'll continue the review.

github/packages.go
gmlewis: OK, let's please leave the field in unless we can [conclusively prove](https://github.com/google/go-github/issues/3394) that it is no longer being used. Feel free to add a comment about your findings.

gmlewis (changes requested):
Thank you, @adrian-saunders!
It is unfortunate that we have so many `any`, `map` and `json.RawMessage` in this PR, but it sounds like it is pretty unavoidable due to the wide variety of possible data.

Let's please add back in the removed `Author` field, and then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/packages_test.go
```diff
@@ -238,31 +221,12 @@ func TestPackageVersion_Marshal(t *testing.T) {
...
-					PublicRepos:     Ptr(1),
-					Followers:       Ptr(1),
-					Following:       Ptr(1),
-					CreatedAt:       &Timestamp{referenceTime},
-					SuspendedAt:     &Timestamp{referenceTime},
-				},
-				Size:      Ptr(int64(1)),
-				CreatedAt: &Timestamp{referenceTime},
-				UpdatedAt: &Timestamp{referenceTime},
+				Size:        Ptr(int64(1)),
```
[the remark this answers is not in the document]
adrian-saunders: I'd expect the test to fail as the Marshal function has added extra fields that don't exist on the struct or could possibly indicate regression - maybe someone accidentally changed the tag?

What are your thoughts here @gmlewis?

github/packages.go
adrian-saunders: OK will add back

github/packages_test.go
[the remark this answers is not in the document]
gmlewis: Please take a look at the implementation of `testJSONMarshal`... it unmarshals then marshals instead of an exact match which explains the behavior.  Feel free to improve `testJSONMarshal` if you want... it is a bit of a hack, but tries to be a general solution for all marshaling without forcing the writers of our tests to get the fields in the same exact order as is defined in each `struct`.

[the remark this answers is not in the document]
adrian-saunders: Ok that makes sense, I will have a think about it/raise another PR at a later point

stevehipwell:
@gmlewis I think this is another case where it's worth considering if `json.RawMessage` is the right solution or if we need an alternative such as to widen the struct field (e.g.single to array) or add a second field (object or string depending on the response)?

gmlewis:
> @gmlewis I think this is another case where it's worth considering if `json.RawMessage` is the right solution or if we need an alternative such as to widen the struct field (e.g.single to array) or add a second field (object or string depending on the response)?

I think that is certainly a possibility. Are there examples of what you are proposing elsewhere in this repo that can be used as a guide for implementation?

If not, I think it may be too much to ask PR authors to try and understand what you are envisioning, as my personal understanding is a bit fuzzy and I would not be able to give guidance.

Maybe the best way to go about this is to open one or more new Issues with a enhancement proposals to make the developer experience better for users of this client library by implementing what you are proposing. In fact, maybe we could start with one or two specific API collections (maybe the two you've already mentioned... this PR and the other one you commented on) and see if we want to expand this idea throughout the repo based on how the changes are received?

stevehipwell (approved):
LGTM

adrian-saunders:
Thank you both for the review!

jBouyoud:
Hi @adrian-saunders,  

It looks like the `Repository` field was deleted, but it's still referenced in the documentation.  
I believe this might be a mistake — please correct me if I’m wrong.

gmlewis:
> Hi @adrian-saunders,
> 
> It looks like the `Repository` field was deleted, but it's still referenced in the documentation. I believe this might be a mistake — please correct me if I’m wrong.
> 
> cc @gmlewis

Ah, yes... specifically, `Package.Repository`.  I'm sorry for not catching that.  Would you like to make a PR to add it back in, @jBouyoud?

jBouyoud:
Yes for sure , i'll do that

jBouyoud:
@gmlewis,

Here is the [PR](https://github.com/google/go-github/pull/3539) — I hope it meets your standards!

PR #3516, opened 2025-03-15
Add validity filter to secret scanning alert list options

x-way:


gmlewis (approved):
Thank you, @x-way!
LGTM.
Merging.

Issue #3523, opened 2025-03-19
Add support for issue types


stevehipwell:
The  GitHub REST API [now supports](https://github.blog/changelog/2025-03-18-github-issues-projects-rest-api-support-for-issue-types/) managing [issue types](https://docs.github.com/en/issues/tracking-your-work-with-issues/configuring-issues/managing-issue-types-in-an-organization).

- [REST API endpoints for issue types](https://docs.github.com/en/rest/orgs/issue-types?apiVersion=2022-11-28)

gmlewis:
Thank you, @stevehipwell!

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.
Also feel free to perform a code review on any PR that has the label "NeedsReview" - we appreciate those contributions too!
We have a `REVIEWERS` file that you could add your username to if you are willing to be contacted when reviews are needed.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

atilsensalduz:
Hey guys, I'd like to work on this issue. Could you please assign it to me?

atilsensalduz:
Hey guys, I've completed my work, but I stumbled upon a GHES API version upgrade. When I ran generate.sh, it updated the GHES version from 3.15 to 3.16. Should I include this change in my PR as well?

![Image](https://github.com/user-attachments/assets/bc0ba5a5-c7a4-40f5-ba45-46fe37ddf9f1)

![Image](https://github.com/user-attachments/assets/32ed36ec-5131-47e2-b30a-13b181791f26)

gmlewis:
That should be fine, thank you for checking!

PR #3525, opened 2025-03-20
feat: Add support for Issue Types API

atilsensalduz:
This pull request introduces support for the Issue Types API to the `go-github` library, enabling users to manage issue types within their organizations. The following functions have been added:

- **ListIssueTypes**: Retrieves all issue types for a specified organization.
- **CreateIssueType**: Allows the creation of a new issue type within an organization.
- **UpdateIssueType**: Enables updating the attributes of an existing issue type.
- **DeleteIssueType**: Facilitates the deletion of an issue type from an organization.

Each function is accompanied by comprehensive tests to ensure functionality and reliability.

API References:  https://docs.github.com/en/rest/orgs/issue-types

Fixes: https://github.com/google/go-github/issues/3523

Please let me know if any additional changes or improvements are needed!


gmlewis:
It looks like the openapi stuff is a problem afterall.
I'm thinking I should make a separate PR to update openapi, then you can merge it from master and your PR can then be focused on just your changes.  I'll do that today.

github/orgs_issue_types.go
```diff
@@ -0,0 +1,98 @@
...
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+type CreateOrUpdateIssueTypesOptions struct {
```
gmlewis, line 13: Please add a godoc-style comment (with punctuation) for all exported structs that describes their purpose.

Wisma-55 (approved):

gmlewis (changes requested):
This looks great, @atilsensalduz!
Just one minor tweak, please.

github/orgs_issue_types.go
atilsensalduz: Thank you for the review @gmlewis 🙏  I've added godoc-style comment.

atilsensalduz:
Tests are passed in my local. @gmlewis could you please rerun the test?

gmlewis:
> wdyt to add small delay in here?

What I would really like to understand is _WHY_ we have any race conditions at all.
There should be a way to close down the server reliably without any use of `sleep`, and I would love to get to the bottom of it. Can you please file an issue, @atilsensalduz, and/or investigate if you feel like it?

github/orgs_issue_types.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+// CreateOrUpdateIssueTypesOptions represents the parameters for creating or updating an issue type.
+type CreateOrUpdateIssueTypesOptions struct {
+	Name        *string `json:"name"`                  // Name of the issue type. (Required.)
+	IsEnabled   *bool   `json:"is_enabled"`            // Whether or not the issue type is enabled at the organization level. (Required.)
```
gmlewis, line 16: Sorry I didn't catch this before, but let's please make these two required fields values instead of pointers.

```suggestion
	Name        string `json:"name"`                  // Name of the issue type. (Required.)
	IsEnabled   bool   `json:"is_enabled"`            // Whether or not the issue type is enabled at the organization level. (Required.)
```

gmlewis (changes requested):

github/orgs_issue_types.go
atilsensalduz: > Sorry I didn't catch this before, but let's please make these two required fields values instead of pointers.

Thanks @gmlewis , I've updated those required fields to values. I'm still learning and I'd like to understand why we prefer values instead of pointers for required fields. could you clarify the reasoning behind this choice?

gmlewis (approved):
Thank you, @atilsensalduz and @Wisma-55!
LGTM.
Merging.

github/orgs_issue_types.go
gmlewis: Oh, sure!  There have been several discussions about this topic in this repo before, but I'll give you a quick summary. We use pointers and `omitempty` for optional fields so that they are not included in the JSON when Go marshals the structs to JSON. When they are required, as in this case, we prefer to use values so that it is obvious to the user of this client library that they need to provide a value (and can't leave it as `nil` if it were a pointer).
I hope that makes sense. Please let me know if you have further questions, and I can try to dig up older conversations that go into more details.

atilsensalduz: Got it now! Thanks a lot for the clear explanation @gmlewis Really appreciate it!

Issue #3527, opened 2025-03-21
Investigate Intermittent Test Failures in GitHub Actions


atilsensalduz:
We've encountered intermittent test failures that don't seem to be tied to a specific test case. The failures appear to be related to network issues in GitHub Actions, as rerunning the pipeline usually resolves them.

```
--- FAIL: TestUsersService_DeleteKey (0.00s)
    users_keys_test.go:170: Users.DeleteKey returned error: Delete "http://127.0.0.1:40547/api-v3/user/keys/1": net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called
```

Since we've observed this issue across multiple test cases, it might be worth investigating whether it's related to how the test server is being managed or if it's an underlying issue with GitHub Actions.

Example workflow runs: 
- https://github.com/google/go-github/actions/runs/13973544105/job/39126491938
- https://github.com/google/go-github/actions/runs/13615256360/job/38057644759

gmlewis:
Thank you, @atilsensalduz!  I would like to add that ideally, any solution will _NOT_ involve `time.Sleep`, as experience has shown that using `time.Sleep` is not a great long-term solution because inevitably a system can be found that needs slightly more time than was alotted and the problem shows up again.

gmlewis:
Also note that this flakiness feels like it started around the time we switched to running all tests in parallel, but I could be mistaken.

Abiji-2020:
> Also note that this flakiness feels like it started around the time we switched to running all tests in parallel, but I could be mistaken.

Actually I have tried to run this actions with the list of `-parallel=25` to ensure is the parallel running was problem and yes it is. 

My actions : 
 - https://github.com/Abiji-2020/go-github/actions/runs/14007606090/job/39223464715

Abiji-2020:
I am thinking of restricting the number of parallel test cases running in the actions so that it may not arise again.

atilsensalduz:
Thanks for the investigation, @Abiji-2020  By default, tests run in parallel based on the CPU count. If not explicitly specified, the number of parallel executions is determined by the GOMAXPROCS environment variable. If the value of GOMAXPROCS is not explicitly set, it will be equal to the number of CPUs. I'm not sure whether the proposed values are lower than the number of CPUs on the GitHub Actions runner.

atilsensalduz:
I'm sharing my findings and proposed solution regarding the issue.

## Root Cause
[...]

PR #3529, opened 2025-03-22
fix: Isolate HTTP transports in parallel tests to prevent connection issues

atilsensalduz:
## Summary  

This PR addresses an issue where parallel tests share the default HTTP transport, leading to intermittent "connection broken" errors due to race conditions in connection reuse.  

## Root Cause  

Tests create their own clients and servers via `setup(t)`, but they unintentionally share Go's default transport when no explicit transport is specified. As a result:  
- `server.Close()` triggers `CloseIdleConnections()` on the default transport.  
- This closes all idle connections, affecting other tests running in parallel.  
- If another test is in the middle of using a connection from the shared pool, it encounters a "connection broken" error.  

## Solution  

- Ensure each test uses an **isolated HTTP transport** instead of the shared default transport.  
- This prevents unintended connection closures from affecting other tests running concurrently.  
- Improves test stability by eliminating race conditions in connection reuse.  

## References  

- Related discussion: https://github.com/google/go-github/issues/3527

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -57,9 +57,29 @@ func setup(t *testing.T) (client *Client, mux *http.ServeMux, serverURL string) 
 	// server is a test HTTP server used to provide mock API responses.
 	server := httptest.NewServer(apiHandler)
 
+	// Create a custom transport with isolated connection pool
+	transport := &http.Transport{
+		// Controls connection reuse - false allows reuse, true forces new connections for each request
+		DisableKeepAlives: false,
+		// Maximum concurrent connections per host (active + idle)
+		MaxConnsPerHost: 10,
+		// Maximum idle connections maintained per host for reuse
+		MaxIdleConnsPerHost: 5,
+		// Maximum total idle connections across all hosts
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @atilsensalduz!
LGTM.

Abiji-2020:
@gmlewis can I review it?

gmlewis:
> @gmlewis can I review it?

Absolutely!  Thank you, @Abiji-2020!  We welcome all contributions.

Abiji-2020 (approved):
✅ Looks Good.
I am just curious that can we reduce the timeout ?

gmlewis:
> ✅ Looks Good. I am just curious that can we reduce the timeout ?

Thank you, @Abiji-2020!

The timeout here is sufficiently large such that it should _NEVER_ trigger... and if it ever does, then something is seriously wrong.  I think it is good to keep the 30-second timeout so that if it ever _does_ take that long, then it will be pretty obvious and will alert us that it needs investigation as to _why_.  So let's please leave it large like this for now.

PR #3545, opened 2025-04-09
Add support for registry_package event

### LINKED ISSUE #3544: Support for registry_package events
According to the [documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads#registry_package) (and [this](https://docs.github.com/en/webhooks/webhook-events-and-payloads#package)) a Github App wishing to receive events for a new/updated package should expect to receive the package event which is the renamed registry_package event. However, what actually happens is that in the permissions section for a Github App there is only the Registry Packages event available and the webhook request title has the value `registry_package`.

Can we please add support for this?

ptzianos:
as of the time of this writing, the Github documentation describes two events related to packages. One called package and one called registry_package. According to the documentation, users should switch to using the package event instead of the registry_package event. However, when working with the actual permissions and events, there is no package event but there is a RegistryPackage event that can be enabled for a Github App. It looks like the documentation is wrong and the registry_package event was added after the package event and it's the registry_package event that users should expect to receive.

Fixes: #3544.


github/event_types.go
```diff
@@ -1467,6 +1467,26 @@ type ReleaseEvent struct {
 	Org *Organization `json:"organization,omitempty"`
 }
 
+// RegistryPackageEvent represents activity related to GitHub Packages.
+// The Webhook event name is "registry_package".
+//
+// This event is triggered when a GitHub Package is published or updated.
+//
+// GitHub API docs: https://docs.github.com/en/webhooks/webhook-events-and-payloads#registry_package
+type RegistryPackageEvent struct {
```
gmlewis, line 1476: Could we please keep these structs sorted by name?

github/event_types_test.go
```diff
@@ -8320,6 +8320,135 @@ func TestPackageEvent_Marshal(t *testing.T) {
 	testJSONMarshal(t, u, want)
 }
 
+func TestRegistryPackageEvent_Marshal(t *testing.T) {
```
gmlewis, line 8323: Ditto

github/messages_test.go
```diff
@@ -539,6 +539,11 @@ func TestParseWebHook(t *testing.T) {
 			payload:     &WorkflowRunEvent{},
 			messageType: "workflow_run",
 		},
+
```
gmlewis, line 542: Ditto, and no need for a blank line here.

github/repos_hooks_deliveries_test.go
```diff
@@ -184,6 +184,7 @@ var hookDeliveryPayloadTypeToStruct = map[string]interface{}{
 	"pull_request_target":            &PullRequestTargetEvent{},
 	"push":                           &PushEvent{},
 	"release":                        &ReleaseEvent{},
+	"registry_package":               &RegistryPackageEvent{},
```
gmlewis, line 187: Ditto

gmlewis (changes requested):
Thank you!

github/event_types.go
ptzianos: sure on it

ptzianos: fixed

github/event_types_test.go
ptzianos: fixed

github/messages_test.go
ptzianos: fixed

github/repos_hooks_deliveries_test.go
ptzianos: fixed

gmlewis (approved):
Thank you, @ptzianos!
LGTM.

@stevehipwell - might you have time for a code review? Thank you!

stevehipwell (approved):
LGTM

ptzianos:
@stevehipwell @gmlewis  do I need to fix the codecov check to get the change merged?

gmlewis:
> @stevehipwell @gmlewis thank you both for the prompt reviews, do I need to fix the codecov check to get the change merged?

No need, thank you. 

Thank you, @stevehipwell!

Merging.

PR #3554, opened 2025-04-18
Fixed headers for EventType & DeliveryID 

sparshev:
Apparently github webhook headers are actually have uppercase "H" in "GitHub", instead "Github":
```
map[Accept:*/* Content-Type:application/json User-Agent:GitHub-Hookshot/ef91aa7 X-GitHub-Delivery:****** X-GitHub-Event:workflow_job X-GitHub-Hook-ID:****** X-GitHub-Hook-Installation-Target-ID:****** X-GitHub-Hook-Installation-Target-Type:repository X-Hub-Signature:sha1=****** X-Hub-Signature-256:sha256=******]
```

### DIFF
--- a/github/messages.go
+++ b/github/messages.go
@@ -321,7 +321,7 @@ func DeliveryID(r *http.Request) string {
 func ParseWebHook(messageType string, payload []byte) (interface{}, error) {
 	eventType, ok := messageToTypeName[messageType]
 	if !ok {
-		return nil, fmt.Errorf("unknown X-Github-Event in message: %v", messageType)
+		return nil, fmt.Errorf("unknown X-GitHub-Event in message: %v", messageType)
 	}
 
 	event := Event{
--- a/github/messages.go
+++ b/github/messages.go
@@ -38,9 +38,9 @@ const (
 	// SHA256SignatureHeader is the GitHub header key used to pass the HMAC-SHA256 hexdigest.
 	SHA256SignatureHeader = "X-Hub-Signature-256"
 	// EventTypeHeader is the GitHub header key used to pass the event type.
-	EventTypeHeader = "X-Github-Event"
+	EventTypeHeader = "X-GitHub-Event"
 	// DeliveryIDHeader is the GitHub header key used to pass the unique ID for the webhook event.
-	DeliveryIDHeader = "X-Github-Delivery"
+	DeliveryIDHeader = "X-GitHub-Delivery"
 )
 
 var (
--- a/github/messages_test.go
+++ b/github/messages_test.go
@@ -607,7 +607,7 @@ func TestDeliveryID(t *testing.T) {
 	if err != nil {
 		t.Fatalf("DeliveryID: %v", err)
 	}
-	req.Header.Set("X-Github-Delivery", id)
+	req.Header.Set("X-GitHub-Delivery", id)
 
 	got := DeliveryID(req)
 	if got != id {


gmlewis:
According to claude.ai:

> RFC 7230 is part of the HTTP/1.1 specification and states that header field names are case-insensitive. This means that while header names might be written with specific capitalization in documentation (often in "Train-Case" with initial capitals like "Content-Type"), the actual implementation should treat them as case-insensitive.
The relevant quote from RFC 7230, section 3.2 is:
"Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace."

So my interpretation of this is that our casing of `GitHub` in custom header fields doesn't really matter, and the only reason that we've chosen the way it is was discussed in another issue many years ago where we decided to be self-consistent and use `Train-Case`.

So unless there is a good reason to change it (which I don't think there is but am willing to listen), I think we should close this issue.

gmlewis:
Looking through the code, we have specifically chosen to make the keys canonical in #3389.

gmlewis:
Closing due to reasons listed in #3389.

sparshev:
Hi @gmlewis,

Sorry, but I should try to protect my changes here. Unfortunately the mentioned #3389 does not completely apply here: the mentioned headers in #3389 are internal for the go-github package, the changed ones in this PR are external. I would also understand if the standard will be used by GitHub here - but it's clearly is not, so what's the point of following the standard if it makes interface to not work correctly, forcing the developer to jump through hoops?

But maybe I'm wrong here and using the provided interfaces incorrectly - maybe you can tell me what to do in the following case:

The issue I have with those headers: when I try to parse webhook (both from github.HookDelivery & incoming webhook) I need to read the headers like that:
```go
guid, ok := headers[github.DeliveryIDHeader] // Bug here due to incorrect key
if !ok {
    return fmt.Errorf("Can't find delivery GUID in headers")
}
sig, ok := headers[github.SHA256SignatureHeader]
if !ok {
    return fmt.Errorf("Can't find delivery signature in headers")
}
eventType, ok := headers[github.EventTypeHeader] // Bug here due to incorrect name
if !ok {
    return fmt.Errorf("Can't find delivery event type in headers")
}
```

Since those are used on client side - I would need to do canonicalization by [http.CanonicalHeaderKey](https://pkg.go.dev/net/http#CanonicalHeaderKey) now, so not quite sure the provided example applies here. Maybe there is a better way you could suggest?

Thank you

gmlewis:
Oh, I see.  You are trying to use these directly to look up the values.
I'm sorry I didn't understand before.  Reopening issue.
I'm pretty sure there is a `Get` call in the headers API that will do all the work for you... I'll reopen and then investigate and report back.

gmlewis:
Have you tried this? https://pkg.go.dev/net/http#Header.Get

gmlewis:
https://pkg.go.dev/net/http#Header.Get is pretty industry-standard.
I would say that if Header.Get doesn't work for you, then it is time to contact GitHub v3 API tech support and point them to this issue and see what they have to say.

sparshev:
The issue I have right now is with HookDelivery accessed by REST API - we have to use it to make sure no webhook requests are missing due to network issues.

Unfortunately [HookDelivery](https://pkg.go.dev/github.com/google/go-github/v71/github#HookDelivery) and it's internal [HookRequest](https://pkg.go.dev/github.com/google/go-github/v71/github#HookRequest) are not providing easy way to access Headers but to use map interface...

gmlewis:
> The issue I have right now is with HookDelivery accessed by REST API - we have to use it to make sure no webhook requests are missing due to network issues.
> 
> Unfortunately [HookDelivery](https://pkg.go.dev/github.com/google/go-github/v71/github#HookDelivery) and it's internal [HookRequest](https://pkg.go.dev/github.com/google/go-github/v71/github#HookRequest) are not providing easy way to access Headers but to use map interface...

Ah, I see what you mean.  Then I would suggest a different way of fixing this.
I suggest that you leave all the canonicalized keys alone - please don't change their cases.
Instead, modify the existing implementation or helper functions to internally canonicalize the externally-generated header keys so that the helper functions in this repo work correctly.

The reason I don't want to change the casing of our constant strings is not only because of the standard, but GitHub could change the capitalization of their strings every other Monday and we would have to attempt to jump through hoops to keep up-to-date with their changes.

Therefore, let us fix this once and fix it such that no matter what capitalization is sent to this client library, it can handle it and work correctly.

Would you like to work on what I have suggested, or would you prefer that we turn this into an issue (instead of a PR) and have someone else tackle this issue, @sparshev ?

sparshev:
Yeah, let's maybe turn it into an issue then - since those header keys are interfaces for the user of the library, I think they need to be good enough to use with the library data interfaces. Maybe the HookRequest itself could implement the http.Header.Get - kind of interface...

PR #3556, opened 2025-04-18
Add case-instensitive GetHeader for HookRequest & HookResponse

### LINKED ISSUE #3555: HookRequest.Headers and HookResponse.Headers maps need to be canonicalized
As discussed in #3554, attempting to access known header keys directly from the maps provided by:
* [HookRequest.Headers](https://github.com/google/go-github/blob/c099baf334af9b3362514dd9d3164c4be0097056/github/repos_hooks_deliveries.go#L45)
* [HookResponse.Headers](https://github.com/google/go-github/blob/c099baf334af9b3362514dd9d3164c4be0097056/github/repos_hooks_deliveries.go#L56)

using the [provided constant, canonicalized header keys](https://github.com/google/go-github/blob/c099baf334af9b3362514dd9d3164c4be0097056/github/messages.go#L36-L43) does not work when the external header keys do not exactly match the case of our provided constants.

There are two solutions to this problem:

Solution 1:
* canonicalize all header keys every time a `HookDelivery` struct is returned ([here](https://github.com/google/go-github/blob/c099baf334af9b3362514dd9d3164c4be0097056/github/repos_hooks_deliveries.go#L69), [here](https://github.com/google/go-github/blob/c099baf334af9b3362514dd9d3164c4be0097056/github/repos_hooks_deliveries.go#L95), and [here](https://github.com/google/go-github/blob/c099baf334af9b3362514dd9d3164c4be0097056/github/repos_hooks_deliveries.go#L116), plus six more locations).

Solution 2:
* add two new helper methods that would provide a [Header.Get](https://pkg.go.dev/net/http#Header.Get)-like experience for `HookRequest` and `HookResponse` such that the header key lookup is totally case insensitive (both for the search strings and for the contents of the map), like this:

```go
// GetHeader gets the value associated with the given key (ignoring key case).
func (h *HookRequest) GetHeader(key string) string {
  ...
}

// GetHeader gets the value associated with the given key (ignoring key case).
func (h *HookResponse) GetHeader(key string) string {
  ...
}
```

I'm open to either solution. @sparshev - do you have a preference?

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.
Also feel free to perform a code review on any PR that has the label "NeedsReview" - we appreciate those contributions too!
We have a `REVIEWERS` file that you could add your username to if you are willing to be contacted when reviews are needed.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

sparshev:
Simple implementation for case-insensitive GetHeader. I thought about caching the keys, but it seems too complicated for now with additional not predictable memory consumption. Tests are included.

Fixes #3555

### DIFF
--- a/github/repos_hooks_deliveries.go
+++ b/github/repos_hooks_deliveries.go
@@ -40,6 +40,16 @@ func (d HookDelivery) String() string {
 	return Stringify(d)
 }
 
+// getHeader common function for GetHeader funcs of HookRequest & HookResponse.
+func getHeader(headers map[string]string, key string) string {
+	for k, v := range headers {
+		if strings.EqualFold(k, key) {
+			return v
+		}
+	}
+	return ""
+}
+
 // HookRequest is a part of HookDelivery that contains
 // the HTTP headers and the JSON payload of the webhook request.
 type HookRequest struct {
@@ -49,12 +59,7 @@ type HookRequest struct {
 
 // GetHeader gets the value associated with the given key (ignoring key case).
 func (r *HookRequest) GetHeader(key string) string {
-	for k, v := range r.Headers {
-		if strings.EqualFold(k, key) {
-			return v
-		}
-	}
-	return ""
+	return getHeader(r.Headers, key)
 }
 
 func (r HookRequest) String() string {
@@ -70,12 +75,7 @@ type HookResponse struct {
 
 // GetHeader gets the value associated with the given key (ignoring key case).
 func (r *HookResponse) GetHeader(key string) string {
-	for k, v := range r.Headers {
-		if strings.EqualFold(k, key) {
-			return v
-		}
-	}
-	return ""
+	return getHeader(r.Headers, key)
 }
 
 func (r HookResponse) String() string {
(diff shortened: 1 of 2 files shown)


gmlewis (changes requested):
If you add one more test case to each of your tests that covers lines 57 and 78, respectively, then the linter will be happy.  😄

sparshev:
Oh you right! Completely forgot about negative cases)

gmlewis (approved):
Thank you, @sparshev!
LGTM.

atilsensalduz:
LGTM! Tiny nit: wdyt to abstract the common functionality into a shared helper function aims to reduce code duplication like below?

```
func getHeader(headers map[string]string, key string) string {
	for k, v := range headers {
		if strings.EqualFold(k, key) {
			return v
		}
	}
	return ""
}

func (r *HookRequest) GetHeader(key string) string {
	return getHeader(r.Headers, key)
}

func (r *HookResponse) GetHeader(key string) string {
	return getHeader(r.Headers, key)
}
```

atilsensalduz (approved):

gmlewis:
Thank you, @atilsensalduz and @sparshev!
Merging after the tests all pass.

PR #3562, opened 2025-04-24
Add support for pagination options in rules API methods

ktruedat:
Updated `GetRulesForBranch`, `GetAllRulesets`, and `GetAllRepositoryRulesets` methods to accept optional pagination parameters (`ListOptions`). Enhanced test cases to validate the use of these parameters in API requests.

### DIFF
--- a/github/repos_rules.go
+++ b/github/repos_rules.go
@@ -38,8 +38,13 @@ type rulesetClearBypassActors struct {
 // GitHub API docs: https://docs.github.com/rest/repos/rules#get-rules-for-a-branch
 //
 //meta:operation GET /repos/{owner}/{repo}/rules/branches/{branch}
-func (s *RepositoriesService) GetRulesForBranch(ctx context.Context, owner, repo, branch string) (*BranchRules, *Response, error) {
+func (s *RepositoriesService) GetRulesForBranch(ctx context.Context, owner, repo, branch string, opts *ListOptions) (*BranchRules, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/rules/branches/%v", owner, repo, branch)
+
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
@@ -55,14 +60,28 @@ func (s *RepositoriesService) GetRulesForBranch(ctx context.Context, owner, repo
 	return rules, resp, nil
 }
 
+// RepositoryListRulesetsOptions specifies optional parameters to the
+// RepositoriesService.GetAllRulesets method.
+type RepositoryListRulesetsOptions struct {
+	// IncludesParents indicates whether to include rulesets configured at the organization or enterprise level that apply to the repository.
+	IncludesParents *bool `url:"includes_parents,omitempty"`
+	ListOptions
+}
+
 // GetAllRulesets gets all the repository rulesets for the specified repository.
-// If includesParents is true, rulesets configured at the organization or enterprise level that apply to the repository will be returned.
+// By default, this endpoint will include rulesets configured at the organization or enterprise level that apply to the repository.
+// To exclude those rulesets, set the `RepositoryListRulesetsOptions.IncludesParents` parameter to `false`.
 //
 // GitHub API docs: https://docs.github.com/rest/repos/rules#get-all-repository-rulesets
 //
 //meta:operation GET /repos/{owner}/{repo}/rulesets
-func (s *RepositoriesService) GetAllRulesets(ctx context.Context, owner, repo string, includesParents bool) ([]*RepositoryRuleset, *Response, error) {
-	u := fmt.Sprintf("repos/%v/%v/rulesets?includes_parents=%v", owner, repo, includesParents)
+func (s *RepositoriesService) GetAllRulesets(ctx context.Context, owner, repo string, opts *RepositoryListRulesetsOptions) ([]*RepositoryRuleset, *Response, error) {
+	u := fmt.Sprintf("repos/%v/%v/rulesets", owner, repo)
+
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
--- a/github/orgs_rules.go
+++ b/github/orgs_rules.go
@@ -15,8 +15,13 @@ import (
 // GitHub API docs: https://docs.github.com/rest/orgs/rules#get-all-organization-repository-rulesets
 //
 //meta:operation GET /orgs/{org}/rulesets
-func (s *OrganizationsService) GetAllRepositoryRulesets(ctx context.Context, org string) ([]*RepositoryRuleset, *Response, error) {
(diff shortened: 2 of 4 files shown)


github/repos_rules.go
```diff
@@ -55,14 +60,28 @@ func (s *RepositoriesService) GetRulesForBranch(ctx context.Context, owner, repo
 	return rules, resp, nil
 }
 
+// RepositoryListRulesetsOptions specifies optional parameters to the
+// RepositoriesService.GetAllRulesets method.
+type RepositoryListRulesetsOptions struct {
+	// IncludesParents indicates whether to include rulesets configured at the organization or enterprise level that apply to the repository.
+	IncludesParents *bool `url:"includes_parents,omitempty"`
```
gmlewis, line 67: According to https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-all-repository-rulesets, the default for `includes_parents` is `true`. So if the user needs to set this to `false`, the `omitempty` will remove the zero-value from the query parameters, so this won't work. It needs to be:

```suggestion
	IncludesParents bool `url:"includes_parents"`
```

github/repos_rules.go
```diff
@@ -55,14 +60,28 @@ func (s *RepositoriesService) GetRulesForBranch(ctx context.Context, owner, repo
...
+	ListOptions
+}
+
 // GetAllRulesets gets all the repository rulesets for the specified repository.
-// If includesParents is true, rulesets configured at the organization or enterprise level that apply to the repository will be returned.
+// By default, this endpoint will include rulesets configured at the organization or enterprise level that apply to the repository.
```
gmlewis, line 72: Also note that this comment is now slightly misleading, because if the user adds `opts`, they will need to explicitly set `IncludesParents` correctly.

gmlewis (changes requested):
Also, to fix the CodeCov issues, please add `testBadOptions` to your new tests. You can see examples in other locations.

github/repos_rules.go
CrinuDimitriu-RecoLabs: Yes, you are correct, this is why I put a pointer there so omitempty will not ignore the `false` value. If I don't put a pointer there and remove `omitempty`, `IncludesParents` will always be `false` if the user doesn't explicitly set that field to true and just fills up the pagination. This also references your comment below, because with pointer and `omitempty`,  if the user adds opts, they will not need to explicitly set IncludesParents correctly. It will be just ignored in the query params, which means the Github API will treat it as `true` for `includes_parents`

github/repos_rules.go
CrinuDimitriu-RecoLabs: Please see my response above, it is related to this also, if the user doesn't explicitly add the `IncludesParents`, it will not be passed to the query params, and the default API behaviour if it's missing will be that `includes_parents` is true

github/repos_rules.go
CrinuDimitriu-RecoLabs: Please let me know what you think about it, in the case of leaving things how they are, with pointer and `omitempty`, the empty value for pointer is nil, which means if the user doesn't fill up that field, it will be omitted, whereas if they want to change the default `true` behaviour of the API, they can set the field to a `Ptr(false)` value, which will not be omitted since it's not an empty value for the pointer.
If I change it to be without a pointer and without `omitempty`, there can occur a different misleading situation:
```go
IncludesParents bool `url:"includes_parents"`
```
In this case, if the user only fills up the pagination and doesn't fill the `IncludesParents` field here, it will always be false. Maybe the user expects it to be `true` since that's what the API docs state, and doesn't fill up this field, but leaving it empty will actually set it to the empty value for bool, which is `false`, when the user expected it to be `true` by default, that may cause confusion

ktruedat: Yes, you are correct, this is why I put a pointer there so omitempty will not ignore the false value. If I don't put a pointer there and remove omitempty, IncludesParents will always be false if the user doesn't explicitly set that field to true and just fills up the pagination. This also references your comment below, because with pointer and omitempty, if the user adds opts, they will not need to explicitly set IncludesParents correctly. It will be just ignored in the query params, which means the Github API will treat it as true for includes_parents

ktruedat: Please let me know what you think about it, in the case of leaving things how they are, with pointer and `omitempty`, the empty value for pointer is nil, which means if the user doesn't fill up that field, it will be omitted, whereas if they want to change the default `true` behaviour of the API, they can set the field to a `Ptr(false)` value, which will not be omitted since it's not an empty value for the pointer.
If I change it to be without a pointer and without `omitempty`, there can occur a different misleading situation:
```go
IncludesParents bool `url:"includes_parents"`
```
In this case, if the user only fills up the pagination and doesn't fill the `IncludesParents` field here, it will always be false. Maybe the user expects it to be `true` since that's what the API docs state, and doesn't fill up this field, but leaving it empty will actually set it to the empty value for bool, which is `false`, when the user expected it to be `true` by default, that may cause confusion

gmlewis: I apologize for the incorrect statements above.  I see what you are saying.

github/repos_rules.go
gmlewis: You are right again, the comment makes sense.  Resolving.

stevehipwell (changes requested):
Do we not want tests to cover the case where we set `opts` to `nil`? I'd have assumed that the existing tests would use `nil` and new tests would have been added for the pagination pattern?

gmlewis:
Testing with `opts` being `nil` sounds like a good idea in this case.
In the general case, that would be uninteresting, but it would probably be good to show it, although we still can't see how GitHub interprets the missing query parameter... so I'm fine either way.

stevehipwell:
@gmlewis wouldn't testing without `opts` catch a regression with the method where it no longer supported `nil`?

ktruedat:
@gmlewis of course! I actually wanted to do separate tests from the start but wasn't sure about it. Will add them in a few

ktruedat:
@gmlewis @stevehipwell I included separate tests with ListOptions for every modified operation. I didn't include the full mock response in the client response since we are not really interested in the response itself, but rather testing the existence of the query params. Let me know if that's okay, thanks!

stevehipwell (changes requested):
@ktruedat is there a reason the enterprise rules have been excluded from these changes? Sorry I'm only just picking this up now, but I thought I'd already checked this.

ktruedat:
> @ktruedat is there a reason the enterprise rules have been excluded from these changes? Sorry I'm only just picking this up now, but I thought I'd already checked this.

@stevehipwell If you are referring to the rules in `enterprise_rules.go`, there are no list endpoints there, and the [documentation](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/rules?apiVersion=2022-11-28#get-an-enterprise-repository-ruleset) also does not mention a `ListEnterpriseRules` endpoint. Therefore, there is nothing to change.
image

stevehipwell (approved):
LGTM

gmlewis:
Thank you, @stevehipwell!
Merging.

gmlewis:
> Thank you so much @gmlewis and @stevehipwell , I am glad I could help here! Also, if there are any other open issues that can be fixed, feel free to reach out to me, I am more than happy to contribute to this amazing repository!

Thank _you_, @ktruedat! You are welcome to contribute to any open unassigned issue. We welcome all contributions!

Issue #3566, opened 2025-04-28
Bug: DownloadContentsWithMeta returns old content for a file


kuzminT:
When I use _DownloadContentsWithMeta_ after making changes to a file, I consistently receive outdated content. I need to wait several minutes before calling this method again to retrieve the latest changes.

Upon investigation, I discovered that the issue originates from GitHub's side. The _DownloadContentsWithMeta_ method uses a direct download link to fetch raw content from this URL: https://raw.githubusercontent.com/. If you attempt to access the raw content immediately after modifying a file, you will encounter the same problem.

However, perhaps we can implement a workaround on our side — for example, by adding some parameters to the URL or using a different method to bypass caching.

I also found a related bug reported in the GitHub community: https://github.com/orgs/community/discussions/46691. As of now, it remains unresolved.

Suggestions:

- Investigate if adding cache-busting parameters (e.g., timestamps or unique query strings) to the download URL forces GitHub to return the latest version.
- Explore using other link/API to fetch the most recent file contents.

gmlewis:
Over the years, we have witnessed several issues like this that have typically been caused by the internal architecture of the GitHub servers being eventually-consistent.

When we report issues like this to GitHub Tech Support, the answer has usually been to "try again after a few seconds".

As you pointed out, the bug is not in this repo, and there is not a good way that I know of to create a workaround within this client library, as each use case varies. For example, users of this repo frequently work around this issue by performing one of these "eventually-consistent" operations, then going off and doing other work, and coming back later to check if the original operation appears to have propagated throughout the GitHub servers, at which point they can continue.

I'll leave this issue open in case others have ideas or want to try out the proposed suggestions above (but be warned that it will not be acceptable to add `time.Sleep` calls to this client library's methods as a workaround since that is never a 100% reliable solution). Eventually we will close this issue if there are no other ideas or working solutions.

kuzminT:
I've checked the response from this GitHub API: https://docs.github.com/en/rest/repos/contents using in this method and found that the content field in the response (base64-encoded) contains the most recent changes to the file.
Would it be possible to return this content directly, or use it instead of making an additional request to the raw content by download_link?

kuzminT:
Hi @gmlewis,

I did some additional research and now understand why `download_url` is used in this method instead of the `content` field. The reason is that `download_url` allows access to the raw file content regardless of file size, while the `content` field in the GitHub API response is only populated under certain conditions:

> If the requested file's size is:
>
[...]

PR #3573, opened 2025-05-10
Improve DownloadContents and DownloadContentsWithMeta methods

kuzminT:
Fixes: #3566. 
By parsing the content field from the response, we can retrieve the most recent state of the file without making an additional request. Only when the content field is empty—such as [for object format and files exceeding the 1 MB ](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content) limit (see GitHub API docs) — do we fall back to the previous approach, which involves directory parsing and fetching the content via download_url or as raw data.
Additional tests were added to cover various cases.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -140,11 +140,8 @@ func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo,
 	dir := path.Dir(filepath)
 	filename := path.Base(filepath)
 	fileContent, _, resp, err := s.GetContents(ctx, owner, repo, filepath, opts)
-	if err != nil {
-		return nil, resp, err
-	}
 
-	if fileContent != nil {
+	if err == nil && fileContent != nil {
 		content, err := fileContent.GetContent()
 		if err == nil && content != "" {
 			return io.NopCloser(strings.NewReader(content)), resp, nil
@@ -194,11 +191,8 @@ func (s *RepositoriesService) DownloadContentsWithMeta(ctx context.Context, owne
 	dir := path.Dir(filepath)
 	filename := path.Base(filepath)
 	fileContent, _, resp, err := s.GetContents(ctx, owner, repo, filepath, opts)
-	if err != nil {
-		return nil, nil, resp, err
-	}
 
-	if fileContent != nil {
+	if err == nil && fileContent != nil {
 		content, err := fileContent.GetContent()
 		if err == nil && content != "" {
 			return io.NopCloser(strings.NewReader(content)), fileContent, resp, nil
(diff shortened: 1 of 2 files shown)


gmlewis:
Ah, I see now what is happening.  Your PR is changing how downloads work... which is probably not a good sign in that this change as written will cause compatibility problems with existing code as demonstrated in the `newTestServer` here:
https://github.com/google/go-github/blob/ef319d55764dd339eb8e56632eabf8eb4563366a/tools/metadata/main_test.go#L356-L378

It looks like you are going to have to make it so that your changes do not adversely affect existing code, if possible.

If it is not possible, then we probably are going to have to keep the legacy behavior consistent for existing users and instead add a new endpoint that behaves the way you propose, with detailed explanations in comments as to when to use the legacy mechanism and when to use your new mechanism.

Does that make sense?

kuzminT:
> It looks like you are going to have to make it so that your changes do not adversely affect existing code, if possible.

I believe I’ve finally managed to fix the problem. Take a look when you have a moment.

github/repos_contents.go
```diff
@@ -139,14 +139,23 @@ func (s *RepositoriesService) GetReadme(ctx context.Context, owner, repo string,
 func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (io.ReadCloser, *Response, error) {
 	dir := path.Dir(filepath)
 	filename := path.Base(filepath)
+	fileContent, _, resp, err := s.GetContents(ctx, owner, repo, filepath, opts)
+
```
gmlewis, line 143: Please delete this blank line so that the error checking is more tightly-coupled to the line that generated the error.

github/repos_contents.go
```diff
@@ -181,14 +190,23 @@ func (s *RepositoriesService) DownloadContents(ctx context.Context, owner, repo,
 func (s *RepositoriesService) DownloadContentsWithMeta(ctx context.Context, owner, repo, filepath string, opts *RepositoryContentGetOptions) (io.ReadCloser, *RepositoryContent, *Response, error) {
 	dir := path.Dir(filepath)
 	filename := path.Base(filepath)
+	fileContent, _, resp, err := s.GetContents(ctx, owner, repo, filepath, opts)
+
```
gmlewis, line 194: ditto

gmlewis (changes requested):
Thank you, @kuzminT!
Two tiny nits to address, please, otherwise looks good.
If you could increase the code coverage, that would be great, but not critical.

After the nits, we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/repos_contents.go
kuzminT: Fixed.

github/repos_contents.go
kuzminT: Fixed.

kuzminT: I can also add some new tests cases for these methods if we need them.

kuzminT: Alright, I believe it’s better now. Let me know if there’s anything else I should adjust.

gmlewis (approved):
Thank you, @kuzminT!
LGTM.

stevehipwell (approved):
LGTM

gmlewis:
Thank you, @stevehipwell!
Merging.

Issue #3581, opened 2025-05-29
DependencyGraph.GetSBOM() missing important SBOM fields like relationship and package externalRefs


viveksahu26:
Hey, 

I'm not sure if this is intentional or an oversight, but while using `DependencyGraph.GetSBOM()` function from the  Github client:

https://github.com/google/go-github/blob/718c304661aa42ba7a01da5b54b2ce85f2f48234/github/dependency_graph.go#L67

  

I noticed that the returned `SBOMInfo` struct is missing important fields like `relationship` section completely and part of package information such as `externalRefs`. These are critical for understanding dependency relationships and for identifying known vulnerabilities using identifiers like PURL, CPE, or SWID.

Here’s how I’m calling it:
```go
dependencyGraph, _, err := client.DependencyGraph.GetSBOM(ctx.Context, owner, repo)
```
Here’s the SBOM I get from this function: https://gist.github.com/viveksahu26/de932a7a4dcc00c6337cec1f7c287b1a

However, when I fetch the SBOM directly using GitHub’s REST API as documented here: https://docs.github.com/en/rest/dependency-graph/sboms?apiVersion=2022-11-28#export-a-software-bill-of-materials-sbom-for-a-repository

I receive a complete SBOM with all the expected details, including relationships and externalRefs:
https://gist.github.com/viveksahu26/b9b53b236fcf9a248e8e4b27ad41f98f

It seems like `go-github` might be missing logic to fully parse or include these fields from the API response. Could this be looked into?

https://github.com/google/go-github/blob/718c304661aa42ba7a01da5b54b2ce85f2f48234/github/dependency_graph.go#L43

```go
type SBOMInfo struct {
	SPDXID       *string       `json:"SPDXID,omitempty"`
	SPDXVersion  *string       `json:"spdxVersion,omitempty"`
	CreationInfo *CreationInfo `json:"creationInfo,omitempty"`

	// Repo name
	Name              *string  `json:"name,omitempty"`
	DataLicense       *string  `json:"dataLicense,omitempty"`
	DocumentDescribes []string `json:"documentDescribes,omitempty"`
	DocumentNamespace *string  `json:"documentNamespace,omitempty"`

	// List of packages dependencies
	Packages []*RepoDependencies `json:"packages,omitempty"`
}
```

viveksahu26:
Hey @gmlewis , I would love to take it :)

PR #3582, opened 2025-05-29
fix: Add missing relationship and pkg external info for SBOMs

viveksahu26:
Fixes: #3581 

This PR adds the following changes:
- It adds up Relationship section in the SBOMs.
- It also includes `externalRefs` under the package information.

TODO: Testing need to be added.

### DIFF
--- a/github/dependency_graph.go
+++ b/github/dependency_graph.go
@@ -37,7 +37,7 @@ type RepoDependencies struct {
 	ExternalRefs     []*PackageExternalRef `json:"externalRefs,omitempty"`
 }
 
-// ExternalRef allows an Package to reference an external sources of additional information,
+// PackageExternalRef allows an Package to reference an external sources of additional information,
 // like asset identifiers, or downloadable content that are relevant to the package,
 // Example for identifiers (e.g., PURL/SWID/CPE) for a package in the SBOM.
 // https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field


github/dependency_graph.go
```diff
@@ -28,12 +28,32 @@ type CreationInfo struct {
...
+	VersionInfo      *string        `json:"versionInfo,omitempty"`
+	DownloadLocation *string        `json:"downloadLocation,omitempty"`
+	FilesAnalyzed    *bool          `json:"filesAnalyzed,omitempty"`
+	LicenseConcluded *string        `json:"licenseConcluded,omitempty"`
+	LicenseDeclared  *string        `json:"licenseDeclared,omitempty"`
+	ExternalRefs     []*ExternalRef `json:"externalRefs"`
+}
+
+// ExternalRef represents an external reference (e.g., PURL/SWID/CPE) for a package in the SBOM.
+type ExternalRef struct {
```
gmlewis, line 41: Since our exported struct names are package-global, let's please add more details to the individual struct names to make them more identifiable-at-a-glance.

Question: are all 3 fields required? If not, you might consider adding `omitempty` to the optional fields and make those same fields pointers instead of values (i.e. `*string` instead of `string`).

```suggestion
// SBOMExternalRef represents an external reference (e.g., PURL/SWID/CPE) for a package in the SBOM.
type SBOMExternalRef struct {
```

github/dependency_graph.go
```diff
@@ -28,12 +28,32 @@ type CreationInfo struct {
...
+	ReferenceCategory string `json:"referenceCategory"`
+	ReferenceType     string `json:"referenceType"`
+	ReferenceLocator  string `json:"referenceLocator"`
+}
+
+// Relationship represents a relationship between two packages in the SBOM.
+type Relationship struct {
```
gmlewis, line 48:
```suggestion
// SBOMRelationship represents a relationship between two packages in the SBOM.
type SBOMRelationship struct {
```

github/dependency_graph.go
```diff
@@ -28,12 +28,32 @@ type CreationInfo struct {
...
+	// Element ID
+	SpdxElementID *string `json:"spdxElementId,omitempty"`
```
gmlewis, line 50: Since SPDX is an acronym for "Software Package Data Exchange", let's please capitalize it following idiomatic Go style:

```suggestion
	SPDXElementID *string `json:"spdxElementId,omitempty"`
```

github/dependency_graph.go
```diff
@@ -28,12 +28,32 @@ type CreationInfo struct {
...
+
+	// Related Element ID
+	RelatedSpdxElement *string `json:"relatedSpdxElement,omitempty"`
```
gmlewis, line 53:
```suggestion
	RelatedSPDXElement *string `json:"relatedSpdxElement,omitempty"`
```

gmlewis (changes requested):
Thank you, @viveksahu26!
I know this PR is still a draft, but I wanted to give some early feedback before you got too far in writing the new tests.

github/dependency_graph.go
viveksahu26: > Since our exported struct names are package-global, let's please add more details to the individual struct names to make them more identifiable-at-a-glance.

Here, do you mean to add details on stuffs that I added only, or beyond that also which have been already there. As I can see that exiting ones also lacks few details.

viveksahu26: > Question: are all 3 fields required? If not, you might consider adding omitempty to the optional fields and make those same fields pointers instead of values (i.e. *string instead of string).

Overall, the field `ExternalRefs` is optional, but if it is provided, then these 3 fields must be provided, they aren't optional.

viveksahu26: > > Here, do you mean to add details on stuffs that I added only, or beyond that also which have been already there. As I can see that exiting ones also lacks few details.
> 
> Sorry I wasn't very clear... here I was just trying to explain the reasoning for my suggestion of changing "ExternalRef" (which is rather vague) to "SBOMExternalRef" so that it is more descriptive and more tightly-coupled to the SBOM parts of the API.

Oh, here you mean to add detail on struct name so that it looks identifiable-at-a-glance. Btw, I added comments describing about each struct fields. Should i remove them, or let it be added ?

PR #3608, opened 2025-06-27
docs: Update HTTP cache references to RFC 9111 compliant implementation

bartventer:
This PR updates the documentation and example code to recommend [`bartventer/httpcache`](https://github.com/bartventer/httpcache) instead of [`gregjones/httpcache`](https://github.com/gregjones/httpcache). `bartventer/httpcache` is actively maintained and complies with [RFC 9111 HTTP Caching](https://datatracker.ietf.org/doc/html/rfc9111).

### Motivation

`bartventer/httpcache` implements [RFC 9111 §4.1](https://www.rfc-editor.org/rfc/rfc9111.html#section-4.1), including proper normalization of Vary headers and URIs when generating cache keys. This covers recommended normalization steps from [RFC 7230 §2.7.3](https://datatracker.ietf.org/doc/html/rfc7230#section-2.7.3) and [RFC 3986 §6](https://datatracker.ietf.org/doc/html/rfc3986#section-6). In contrast, `gregjones/httpcache` uses `(*url.URL).String()` as the cache key, which means it cannot support Vary-based caching and may produce unpredictable caching behaviour for requests that differ by query parameters, fragments, or other URI components.

Additionally, `bartventer/httpcache` also:
  - Supports `stale-while-revalidate` and the `immutable` cache-control extension.
  - Caches a wider range of HTTP status codes, including `200`, `203`, `301`, `304`, `308`, `404`, `405`, `410`, `414`, and `501`.
  - Adds a detailed cache status header (`X-Httpcache-Status`) for easier debugging and monitoring (`HIT`, `MISS`, `STALE`, `REVALIDATED`, `BYPASS`).
  - Maintains compatibility with the legacy `X-From-Cache` header for drop-in replacement.

### DIFF
--- a/github/billing.go
+++ b/github/billing.go
@@ -63,6 +63,46 @@ type AdvancedSecurityCommittersBreakdown struct {
 	LastPushedDate *string `json:"last_pushed_date,omitempty"`
 }
 
+// UsageReportOptions specifies optional parameters for the enhanced billing platform usage report.
+type UsageReportOptions struct {
+	// If specified, only return results for a single year. The value of year is an integer with four digits representing a year. For example, 2025.
+	// Default value is the current year.
+	Year *int `url:"year,omitempty"`
+
+	// If specified, only return results for a single month. The value of month is an integer between 1 and 12.
+	// If no year is specified the default year is used.
+	Month *int `url:"month,omitempty"`
+
+	// If specified, only return results for a single day. The value of day is an integer between 1 and 31.
+	// If no year or month is specified, the default year and month are used.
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you for sharing this, @bartventer, and nice work! It is greatly appreciated.
LGTM.
Merging.

bartventer:
> Thank you for sharing this, @bartventer, and nice work! It is greatly appreciated. LGTM. Merging.

Glad to contribute, thanks for merging @gmlewis !

Issue #3611, opened 2025-06-30
BUG: Remove Sub Issue uses incorrect endpoint


piotrolchawa-f3:
## SubIssues Bug

Recent addition for managing Sub Issues uses wrong endpoint:

`Failed to remove sub-issue: 404 Not Found, DELETE https://api.github.com/repos/REPO/OWNER/issues/ISSUE_NUMBER/sub_issues: 404 Not Found []`

https://github.com/google/go-github/blob/master/github/sub_issue.go#L51-L53

This endpoint (however inconsistent) must be singular form:

https://docs.github.com/en/rest/issues/sub-issues?apiVersion=2022-11-28#remove-sub-issue

piotrolchawa-f3:
https://github.com/google/go-github/pull/3612#issuecomment-3018107145

PR #3612, opened 2025-06-30
fix: Correct endpoint path for sub-issues in Remove method and test

piotrolchawa-f3:
fixes: https://github.com/google/go-github/issues/3611


piotrolchawa-f3:
Not filling that out, you can copy the changes and submit them yourself.

PR #3619, opened 2025-07-01
fix: Handle null assignee in Copilot Seat Billing API response

### LINKED ISSUE #3616: The Copilot Seat Billing API response for the assignee type can be "null"
There is a bug in the copilot seat details, per the spec null is a valid response type for the assignee:

```json
          "assignee": {
            "anyOf": [
              {
                "type": "null"
              },
```

https://docs.github.com/en/rest/copilot/copilot-user-management?apiVersion=2022-11-28#list-all-copilot-seat-assignments-for-an-organization

However, the code here errors with 
Code:
https://github.com/google/go-github/blob/eff2bbd83e04dfdb38a9a878f7756e305656a9e2/github/copilot.go#L47

https://github.com/google/go-github/blob/eff2bbd83e04dfdb38a9a878f7756e305656a9e2/github/copilot.go#L212

 it's getting tested here and expecting an error, as invalid json. 
This test checks for nil type, and returns an error, it should pass. 
https://github.com/google/go-github/blob/eff2bbd83e04dfdb38a9a878f7756e305656a9e2/github/copilot_test.go#L30

sanjayy-gowdaa:
### Fix: Handle `null` assignee in Copilot API response
Resolves #3616  
#### 🔧 Changes Made
- **Fixed `UnmarshalJSON` method** in `copilot.go`:
  - Added a `case nil:` branch to handle `null` `assignee` values.
  - Sets `cp.Assignee = nil` when `assignee` is `null`, instead of returning an error.
- **Updated `copilot_test.go`**:
  - Added a unit test case `"Null Assignee"` to ensure the `assignee: null` scenario is handled correctly and no error is thrown.

#### 🐛 Problem
The current implementation of `UnmarshalJSON` for `CopilotSeatDetails` throws an error when the `assignee` field is `null`, even though the [GitHub API documentation](https://docs.github.com/en/rest/copilot/copilot-user-management?apiVersion=2022-11-28#list-all-copilot-seat-assignments-for-an-organization) explicitly states that `assignee` can be `null`.

### DIFF
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -203,6 +203,9 @@ func (cp *CopilotSeatDetails) UnmarshalJSON(data []byte) error {
 	cp.PlanType = seatDetail.PlanType
 
 	switch v := seatDetail.Assignee.(type) {
+	case nil:
+		// Assignee can be null according to GitHub API specification
+		cp.Assignee = nil
 	case map[string]any:
 		jsonData, err := json.Marshal(seatDetail.Assignee)
 		if err != nil {
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -204,7 +204,9 @@ func (cp *CopilotSeatDetails) UnmarshalJSON(data []byte) error {
 
 	switch v := seatDetail.Assignee.(type) {
 	case nil:
-		// Assignee can be null according to GitHub API specification
+		// Assignee can be null according to GitHub API specification.
+		// See: https://docs.github.com/en/rest/copilot/copilot-user-management?apiVersion=2022-11-28#list-all-copilot-seat-assignments-for-an-organization
+		// Note: Copilot API is in public preview and subject to change.
 		cp.Assignee = nil
 	case map[string]any:
 		jsonData, err := json.Marshal(seatDetail.Assignee)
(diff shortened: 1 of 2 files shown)


github/copilot.go
```diff
@@ -203,6 +203,9 @@ func (cp *CopilotSeatDetails) UnmarshalJSON(data []byte) error {
 	cp.PlanType = seatDetail.PlanType
 
 	switch v := seatDetail.Assignee.(type) {
+	case nil:
+		// Assignee can be null according to GitHub API specification
```
hutson: Not sure the documentation expectations here, but should this have a link to the GitHub API documentation? (I would have also mentioned the specific API version, but the Copilot API is technically in public preview and subject to change at any time)

hutson (approved):

github/copilot.go
sanjayy-gowdaa: I've updated the comment to include the documentation link and API version as you suggested!

sanjayy-gowdaa:
Thank you, I’ll make sure to comment on the issue and get it assigned to me before starting work next time.

PR #2567, opened 2022-11-16
Add the 'require_last_push_approval' field 

noamd-legit:

PR #3623, opened 2025-07-10
fix: Add missing `require_last_push_approval` field to branch protection rule event structs

manue1:
Adds the `require_last_push_approval` field in order to be returned as part of the webhook payload object for `branch_protection_rule` as documented [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=created#branch_protection_rule).

It's being added to both the `ProtectionChanges` and `BranchProtectionRule` to account for the `created` and `edited` action types of the webhook.

_For context_:
- in https://github.com/google/go-github/pull/2567 the `require_last_push_approval` field was added originally
- request for addition of ☝🏼 field: https://github.com/google/go-github/issues/2574

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -1878,6 +1878,14 @@ func (b *BranchProtectionRule) GetRequiredStatusChecksEnforcementLevel() string 
 	return *b.RequiredStatusChecksEnforcementLevel
 }
 
+// GetRequireLastPushApproval returns the RequireLastPushApproval field if it's non-nil, zero value otherwise.
+func (b *BranchProtectionRule) GetRequireLastPushApproval() bool {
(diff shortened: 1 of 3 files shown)


manue1:
@gmlewis I'm unsure what's the release cadence like, but I was wondering if it would be possible to get a patch release out to be available to make use of this change in an official release?
Thanks in advance. 🙇🏼

gmlewis:
> @gmlewis I'm unsure what's the release cadence like, but I was wondering if it would be possible to get a patch release out to be available to make use of this change in an official release? Thanks in advance. 🙇🏼

Hi @manue1 - we should be ready for another release shortly. I would like to get #3631 in, as that should be a quick fix, then we can perform another major release.

PR #3673, opened 2025-08-12
fix: Replace 'organizations' with 'orgs' in API endpoints

zyfy29:
As noticed https://github.com/google/go-github/pull/3671#issuecomment-3179719621, it seems that github changed some organizations api endpoints in the two days.

I just fixed my implement yesterday, then did `script/metadata.sh update-openapi` to follow the change.


gmlewis:
OK, I'm seeing what you are seeing, so we will proceed with this PR. Thank you, @zyfy29!

What I'm amazed at, though, is how GitHub can make a breaking change like this without an announcement... although maybe I simply missed the announcement.

Issue #3658, opened 2025-08-04
Feature Request: configurable telemetry


aThorp96:
In a web service which may be making tens of thousands of call to Github, it's desirable to have instrumentation to identify hot code paths, bad tenants, or poor optimizations resulting in extra API calls.

Currently there are three ways to gain metrics on the API requests:
1. Wrapping every API call with a wrapper that emits the telemetry data
    Ugly, a little verbose, easy for new code to accidentally forget to wrap the API call or manually emit the telemetry data, isn't accurate when paging
2. Wrapping API client access with a wrapper which emits the telemetry data every time get the client
    Pretty clean, very easy to subvert by saving the returned client in a variable, isn't accurate when paging
3. Wrapping the entire struct and every necessary struct method 
    Very safe from accidentally using the client without metrics, boilerplate hell, isn't accurate when paging

It would be very helpful for the library to provide a mechanism to instrument the API usage with some form of telemetry. Any kind of pluggable middleware, callbacks, or hook would add a lot of utility. If there is a way to do this which is both hard to accidentally workaround and doesn't require a lot of boilerplate, that would be awesome to know about and document! 

One very simple or rudimentary option would be to add an optional field `Client.OnDo: func()` (also settable via `Client.WithCallback()` or something) which, if set, would be called during `Client.bareDo` just before `caller.Do()`. A example usage might be:
```go
func metricEmitter(org: string, repo: string) func() {
    return func() {
        metrics.emit("api-request", tags: map[string]string{"organization": org, "repository": repo})
    }

...
prs, _, err := client
  .WithCallback(metricEmitter(request.Args.Owner, request.Args.Repository))
  .PullRequests
  .List(request.Args.Owner, request.Args.Repository)
```

gmlewis:
Very interesting, @aThorp96! 
Would you like to work on adding an optional feature like what you have suggested?
If so, I'll assign this issue to you. If not, I'll open this up to our other awesome contributors.

aThorp96:
I wouldn't mind working on it, but I have not contributed to this project before and would at least want to get the community's thoughts on the design. There's a lot of ways to skin this cat with various tradeoffs 😁 

I'd also be interested if anyone has found patterns for instrumenting their API usage besides the three I've thought of

stevehipwell:
@aThorp96 would you mind expanding on what you can't get using OpenTelemetry on the client transport?

Not-Dhananjay-Mishra:
[...]

PR #3681, opened 2025-08-16
Added WithCallback Feature [#3658]

Not-Dhananjay-Mishra:
This PR implements the feature requested in [[#3658]](https://github.com/google/go-github/issues/3658) to allow configurable telemetry for the Client.

### **Modifications -**

- `WithCallback(callback func())` was added. The callback will be set by the user.
- The callback execute just before the `client.bareDo` inside `client.Do`.
- Allows for chaining because it returns the same Client instance.
- Added new Callback field in the Client struct of type `func()`, which default set to nil during `initialize()`.

### Example Code -
`client := github.NewClient(nil).WithCallback(func() {`
`    fmt.Println("Callback executed")`
`})`

This enables users to attach custom telemetry or logging kind of things to API calls in a flexible, user friendly manner.
_Btw this is my first PR_
_Happy to get any feedback or suggestions to improve it_

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -228,6 +228,7 @@ type Client struct {
 	SubIssue           *SubIssueService
 	Teams              *TeamsService
 	Users              *UsersService
+	Callback           func()
 }
 
 type service struct {
@@ -465,6 +466,7 @@ func (c *Client) initialize() {
 	c.SubIssue = (*SubIssueService)(&c.common)
 	c.Teams = (*TeamsService)(&c.common)
 	c.Users = (*UsersService)(&c.common)
+	c.Callback = nil
 }
 
 // copy returns a copy of the current client. It must be initialized before use.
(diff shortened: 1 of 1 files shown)


Not-Dhananjay-Mishra:
Corrected the typo that caused the lint issue. I’m still learning, so thank you

gmlewis (commented):
Thank you, @Not-Dhananjay-Mishra - but I don't think this fully satisfies what is being requested in #3658. 

Additionally, discussion has not completed in that issue, as there is an [outstanding request](https://github.com/google/go-github/issues/3658#issuecomment-3167600730) to the original author to further clarify their request, so it is premature to be writing any PR for that issue yet.  

Let's hold off on this PR until we get some more clarity, please.

Not-Dhananjay-Mishra:
Thanks @gmlewis for the clarification! I will hold off on the PR. I appreciate the guidance.

Issue #3666, opened 2025-08-11
Make sure exported methods and types are documented by godoc in lint rules


zyfy29:
Here says,
https://github.com/google/go-github/blob/84a094539fffabe77eee6e4d807561bf639388c4/CONTRIBUTING.md?plain=1#L95-L96

Maybe there should mention exported types as well. And it could be guaranteed by a lint rule. 

I find that [revive](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#exported) could do it, which is already enabled in this project so no new dependence will be introduced. I just checked it in my local environment and there are up to 20+ methods and types missing godoc comments.

I just want to check that if exported Vars and Consts need to be documented.

And, if you think it's a good idea, I could be of some help to implement it.

PR #3693, opened 2025-08-20
docs: Update CONTRIBUTING.md to clarify documentation requirements for exported types

zyfy29:
I noticed it in the #3666 but forgot to do in the former pr. I think it's better to notify the rule here as well.


gmlewis (approved):
Thank you, @zyfy29!
LGTM.
Merging.

PR #3701, opened 2025-08-27
Remove BillingActions for User and Organisation

Manas23601:
- Actions is now supported in enhanced billing only
- convert struct fields to pass by value from pointers for easier processing

### DIFF
--- a/github/billing.go
+++ b/github/billing.go
@@ -16,14 +16,6 @@ import (
 // GitHub API docs: https://docs.github.com/rest/billing
 type BillingService service
 
-// ActionBilling represents a GitHub Action billing.
-type ActionBilling struct {
-	TotalMinutesUsed     float64              `json:"total_minutes_used"`
-	TotalPaidMinutesUsed float64              `json:"total_paid_minutes_used"`
-	IncludedMinutes      float64              `json:"included_minutes"`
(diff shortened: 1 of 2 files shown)


gmlewis:
Just FYI - I [lost write access](https://github.com/google/go-github/pull/3689#issuecomment-3197809917) to this repo and this may cause unexpected delays for which I apologize.
You can watch that other PR to check on the status.

Manas23601:
I'm not able to do that for you. The API call redirects you to the enhanced billing platform API page. The API only returns GiTHub Actions cost.

gmlewis:
> I'm not able to do that for you. The API call redirects you to the enhanced billing platform API page. The API only returns GiTHub Actions cost.

OK, thanks. Are you running this against an Enterprise version of GitHub, or the publicly-hosted service?
If what you are seeing is against an Enterprise version, then I don't think we may need to rethink our strategy.

Manas23601:
It is the public version, I tested it on a organisation Ii created.Also, I haven't tried this on an enterprise version.

alexandear (approved):
Please update PR's description by removing "convert struct fields to pass by value from pointers for easier processing".

gmlewis:
Thank you, @alexandear!  Done.
Merging.

Manas23601:
Thank you @gmlewis

PR #3738, opened 2025-09-26
chore: Ignore hidden `.claude` settings folder

zyfy29:
ignore settings files like `.claude/settings.local.json` for claude code usage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ coverage.out
 vendor/
 .DS_Store
 .vscode
+.claude/
 # vim temp files
 *.swp
 


gmlewis:
> I'm not sure we should add this change, as it may not be relevant for every developer who doesn't use Claude.
> 
> You can ignore `.claude` locally by modifying `.git/info/exclude` in your local repo copy or by using a global `.gitignore` configuration.

While I totally agree that it may not be relevant for every developer, I also think it is a simple change that could be helpful to some devs and to me it it feels similar to .DS_Store that helps out Mac devs... so I think we can simply just merge this and move on.

gmlewis (approved):

PR #3739, opened 2025-09-27
refactor: Use `errors` package to compare and assert error types

### LINKED ISSUE #3699: Request: Use `errors.As` instead of type assertions internally
I are trying to make this library play nicely with a custom `http.Client` that converts certain error responses to `error`s at a lower level. This (rightly) breaks some things, because the error type this `Client` returns isn't an error type that go-github recognizes. One way to solve this would be to `errors.Join` the custom error type with an `*ErrorResponse` instance, so that it acts as both error types.

The problem I've faced is that go-github sometimes type-asserts `error`s down to `*ErrorResponse`, instead of using `errors.As`:

- https://github.com/google/go-github/blob/f137c94931a722223df8cc7581a2a3e953ad8d63/github/github.go#L1184
- https://github.com/google/go-github/blob/f137c94931a722223df8cc7581a2a3e953ad8d63/github/github.go#L1489
- https://github.com/google/go-github/blob/f137c94931a722223df8cc7581a2a3e953ad8d63/github/repos.go#L2370

This is _not_ a bug in the library, because our `http.Client` implementation technically breaks the API contract of `http.Client`; the doc comment above `RoundTripper.RoundTrip` says:

```go
	// RoundTrip should not attempt to interpret the response. In
	// particular, RoundTrip must return err == nil if it obtained
	// a response, regardless of the response's HTTP status code.
```

This pattern of customizing errors at the `http.Client` level is nonetheless useful to us, and it would be helpful if go-github played nicely with it. I believe the only changes required would be to use `errors.As` instead of a direct type assertion at the three locations I mentioned above. Would you be accepting of such a change?

Thanks for your time.

zyfy29:
close #3699 

follows https://go.dev/blog/go1.13-errors to handle error types with `errors` package

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -1184,8 +1184,7 @@ func (r *ErrorResponse) Error() string {
 // Is returns whether the provided error equals this error.
 func (r *ErrorResponse) Is(target error) bool {
 	var v *ErrorResponse
-	ok := errors.As(target, &v)
-	if !ok {
+	if !errors.As(target, &v) {
 		return false
 	}
 
@@ -1250,8 +1249,7 @@ func (r *RateLimitError) Error() string {
 // Is returns whether the provided error equals this error.
 func (r *RateLimitError) Is(target error) bool {
 	var v *RateLimitError
-	ok := errors.As(target, &v)
-	if !ok {
+	if !errors.As(target, &v) {
 		return false
 	}
 
@@ -1278,8 +1276,7 @@ func (*AcceptedError) Error() string {
 // Is returns whether the provided error equals this error.
 func (ae *AcceptedError) Is(target error) bool {
 	var v *AcceptedError
-	ok := errors.As(target, &v)
-	if !ok {
+	if !errors.As(target, &v) {
 		return false
 	}
 	return bytes.Equal(ae.Raw, v.Raw)
@@ -1306,8 +1303,7 @@ func (r *AbuseRateLimitError) Error() string {
 // Is returns whether the provided error equals this error.
 func (r *AbuseRateLimitError) Is(target error) bool {
 	var v *AbuseRateLimitError
-	ok := errors.As(target, &v)
-	if !ok {
+	if !errors.As(target, &v) {
 		return false
 	}
 
@@ -1341,8 +1337,7 @@ func (r *RedirectionError) Error() string {
 // Is returns whether the provided error equals this error.
 func (r *RedirectionError) Is(target error) bool {
 	var v *RedirectionError
-	ok := errors.As(target, &v)
-	if !ok {
+	if !errors.As(target, &v) {
 		return false
 	}
 
--- a/github/orgs_outside_collaborators_test.go
+++ b/github/orgs_outside_collaborators_test.go
@@ -105,8 +105,11 @@ func TestOrganizationsService_RemoveOutsideCollaborator_NonMember(t *testing.T) 
 
 	ctx := context.Background()
 	_, err := client.Organizations.RemoveOutsideCollaborator(ctx, "o", "u")
-	if !errors.As(err, new(*ErrorResponse)) {
+	var rerr *ErrorResponse
+	if !errors.As(err, &rerr) {
 		t.Error("Organizations.RemoveOutsideCollaborator did not return an error")
+	} else if rerr.Response.StatusCode != http.StatusNotFound {
+		t.Error("Organizations.RemoveOutsideCollaborator did not return 404 status code")
 	}
 }
 
@@ -122,8 +125,11 @@ func TestOrganizationsService_RemoveOutsideCollaborator_Member(t *testing.T) {
 
 	ctx := context.Background()
 	_, err := client.Organizations.RemoveOutsideCollaborator(ctx, "o", "u")
-	if !errors.As(err, new(*ErrorResponse)) {
+	var rerr *ErrorResponse
+	if !errors.As(err, &rerr) {
 		t.Error("Organizations.RemoveOutsideCollaborator did not return an error")
+	} else if rerr.Response.StatusCode != http.StatusUnprocessableEntity {
+		t.Error("Organizations.RemoveOutsideCollaborator did not return 422 status code")
 	}
 }
 
@@ -165,7 +171,10 @@ func TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMemberOrLast
 
 	ctx := context.Background()
 	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(ctx, "o", "u")
-	if !errors.As(err, new(*ErrorResponse)) {
+	var rerr *ErrorResponse
+	if !errors.As(err, &rerr) {
 		t.Error("Organizations.ConvertMemberToOutsideCollaborator did not return an error")
+	} else if rerr.Response.StatusCode != http.StatusForbidden {
+		t.Error("Organizations.ConvertMemberToOutsideCollaborator did not return 403 status code")
 	}
 }


zyfy29:
Actually I added a lint rule [go-errorlint](https://github.com/polyfloyd/go-errorlint) to do it, would you like to add it to this project? If so I could open another pr to do it because I think it's better to leave 2 commit for `refactor:` and `ci:`.

ccoVeille (approved):

gmlewis:
> Actually I added a lint rule [go-errorlint](https://github.com/polyfloyd/go-errorlint) to do it, would you like to add it to this project? If so I could open another pr to do it because I think it's better to leave 2 commit for `refactor:` and `ci:`.

That sounds great, thank you, @zyfy29!

github/github.go
```diff
@@ -923,7 +924,8 @@ func (c *Client) bareDo(ctx context.Context, caller *http.Client, req *http.Requ
 		// added to the AcceptedError and returned.
 		//
 		// Issue #1022
-		aerr, ok := err.(*AcceptedError)
+		var aerr *AcceptedError
+		ok := errors.As(err, &aerr)
```
gmlewis:
```suggestion
		if errors.As(err, &aerr) {
```

github/github.go
```diff
@@ -934,8 +936,8 @@ func (c *Client) bareDo(ctx context.Context, caller *http.Client, req *http.Requ
 			err = aerr
 		}
 
-		rateLimitError, ok := err.(*RateLimitError)
-		if ok && req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) != nil {
+		var rateLimitError *RateLimitError
+		if ok := errors.As(err, &rateLimitError); ok && req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) != nil {
```
gmlewis: Everywhere you see this pattern, please change it to the simplified version:


```suggestion
		if errors.As(err, &rateLimitError) && req.Context().Value(SleepUntilPrimaryRateLimitResetWhenRateLimited) != nil {
```

github/github_test.go
```diff
@@ -282,11 +283,15 @@ func testErrorResponseForStatusCode(t *testing.T, code int) {
...
-		if code != e.Response.StatusCode {
+	var errResp *ErrorResponse
+	var rateLimitErr *RateLimitError
+	var abuseErr *AbuseRateLimitError
+
+	switch {
+	case errors.As(err, &errResp):
+	case errors.As(err, &rateLimitErr):
+	case errors.As(err, &abuseErr):
+		if code != abuseErr.Response.StatusCode {
```
```suggestion
	var abuseErr *AbuseRateLimitError
	switch {
	case errors.Is(err, ErrorResponse):
	case errors.Is(err, RateLimitError):
	case errors.As(err, &abuseErr):
		if code != abuseErr.Response.StatusCode {
```

github/github_test.go
```diff
@@ -1157,8 +1164,8 @@ func TestDo_preservesResponseInHTTPError(t *testing.T) {
 	}
 
 	// Verify error type and access to status code
-	errResp, ok := err.(*ErrorResponse)
-	if !ok {
+	var errResp *ErrorResponse
+	if errors.As(err, &errResp) {
```
gmlewis: Please change all usages of `errors.As` to `errors.Is` where the actual value is not needed to be cast to that type, like here any many other places:


```suggestion
	if errors.Is(err, ErrorResponse) {
```

gmlewis (changes requested):

github/github.go
zyfy29: Oh, this is hard to find! Thanks!

github/github_test.go
zyfy29: In my understanding `Is` is for comparison and `As` is for type assertion. Here `ErrorResponse` is a type name so `Is` will cause a compilation error.

gmlewis: Oh, right! I'm so sorry about the bad advice. I thought we didn't need to make a pointer of the error if it doesn't need to be used, but we do. 😞

example/basicauth/main.go
```diff
@@ -43,7 +44,8 @@ func main() {
 	user, _, err := client.Users.Get(ctx, "")
 
 	// Is this a two-factor auth error? If so, prompt for OTP and try again.
-	if _, ok := err.(*github.TwoFactorAuthError); ok {
+	var twoFactorAuthError *github.TwoFactorAuthError
+	if errors.As(err, &twoFactorAuthError) {
```
gmlewis: When we do not need the value, I much prefer the single-line solution here and throughout this PR, which is what I was attempting to do previously (but unfortunately I got it wrong before). Now I've tried this out, and it works.  Please go ahead and update this PR so that any place we don't need the value, it transforms into a one-liner use of `errors.As`, like this:


```suggestion
	if errors.As(err, new(*github.TwoFactorAuthError)) {
```

zyfy29: That's a good plan!

github/github.go
```diff
@@ -1181,7 +1183,8 @@ func (r *ErrorResponse) Error() string {
 
 // Is returns whether the provided error equals this error.
 func (r *ErrorResponse) Is(target error) bool {
-	v, ok := target.(*ErrorResponse)
+	var v *ErrorResponse
+	ok := errors.As(target, &v)
```
gmlewis:
```suggestion
	if !errors.As(target, &v) {
```

github/github.go
```diff
@@ -1246,7 +1249,8 @@ func (r *RateLimitError) Error() string {
 
 // Is returns whether the provided error equals this error.
 func (r *RateLimitError) Is(target error) bool {
-	v, ok := target.(*RateLimitError)
+	var v *RateLimitError
+	ok := errors.As(target, &v)
```
gmlewis:
```suggestion
	if !errors.As(target, &v) {
```

github/github.go
```diff
@@ -1273,7 +1277,8 @@ func (*AcceptedError) Error() string {
 
 // Is returns whether the provided error equals this error.
 func (ae *AcceptedError) Is(target error) bool {
-	v, ok := target.(*AcceptedError)
+	var v *AcceptedError
+	ok := errors.As(target, &v)
```
gmlewis:
```suggestion
	if !errors.As(target, &v) {
```

github/github.go
```diff
@@ -1300,7 +1305,8 @@ func (r *AbuseRateLimitError) Error() string {
 
 // Is returns whether the provided error equals this error.
 func (r *AbuseRateLimitError) Is(target error) bool {
-	v, ok := target.(*AbuseRateLimitError)
+	var v *AbuseRateLimitError
+	ok := errors.As(target, &v)
```
gmlewis:
```suggestion
	if !errors.As(target, &v) {
```

github/github.go
```diff
@@ -1334,7 +1340,8 @@ func (r *RedirectionError) Error() string {
 
 // Is returns whether the provided error equals this error.
 func (r *RedirectionError) Is(target error) bool {
-	v, ok := target.(*RedirectionError)
+	var v *RedirectionError
+	ok := errors.As(target, &v)
```
gmlewis:
```suggestion
	if !errors.As(target, &v) {
```

github/orgs_outside_collaborators_test.go
```diff
@@ -104,10 +105,8 @@ func TestOrganizationsService_RemoveOutsideCollaborator_NonMember(t *testing.T)
 
 	ctx := context.Background()
 	_, err := client.Organizations.RemoveOutsideCollaborator(ctx, "o", "u")
-	if err, ok := err.(*ErrorResponse); !ok {
+	if !errors.As(err, new(*ErrorResponse)) {
 		t.Error("Organizations.RemoveOutsideCollaborator did not return an error")
-	} else if err.Response.StatusCode != http.StatusNotFound {
-		t.Error("Organizations.RemoveOutsideCollaborator did not return 404 status code")
```
gmlewis, line 110: Why were these lines deleted?

github/orgs_outside_collaborators_test.go
```diff
@@ -123,10 +122,8 @@ func TestOrganizationsService_RemoveOutsideCollaborator_Member(t *testing.T) {
 
 	ctx := context.Background()
 	_, err := client.Organizations.RemoveOutsideCollaborator(ctx, "o", "u")
-	if err, ok := err.(*ErrorResponse); !ok {
+	if !errors.As(err, new(*ErrorResponse)) {
 		t.Error("Organizations.RemoveOutsideCollaborator did not return an error")
-	} else if err.Response.StatusCode != http.StatusUnprocessableEntity {
-		t.Error("Organizations.RemoveOutsideCollaborator did not return 422 status code")
```
gmlewis, line 129: Same here?

github/orgs_outside_collaborators_test.go
```diff
@@ -168,9 +165,7 @@ func TestOrganizationsService_ConvertMemberToOutsideCollaborator_NonMemberOrLast
 
 	ctx := context.Background()
 	_, err := client.Organizations.ConvertMemberToOutsideCollaborator(ctx, "o", "u")
-	if err, ok := err.(*ErrorResponse); !ok {
+	if !errors.As(err, new(*ErrorResponse)) {
 		t.Error("Organizations.ConvertMemberToOutsideCollaborator did not return an error")
-	} else if err.Response.StatusCode != http.StatusForbidden {
-		t.Error("Organizations.ConvertMemberToOutsideCollaborator did not return 403 status code")
```
gmlewis, line 174: And here?

zyfy29 (commented):
Sorry for the mistakes, I found most of them are made by Intellij IDEA's quick-fix and now I've checked it again.

gmlewis:
Thank you, @stevehipwell!
Merging.

@zyfy29 - please follow up with a PR that enables linting for the usage of the `errors` package. Thank you!

PR #3740, opened 2025-09-27
refactor: Support array type for `DefaultValue` field

### LINKED ISSUE #3692: The `DefaultValue` of a `CustomProperty` is incorrect type
Since GitHub added support for multi-select it's been possible for the `DefaultValue` of a `CustomProperty` to be either a string or an array of strings; see schema below.

```json
      "default_value": {
        "oneOf": [
          {
            "type": "string"
          },
          {
            "type": "array",
            "items": {
              "type": "string"
            }
          }
        ],
        "description": "Default value of the property",
        "type": [
          "null",
          "string",
          "array"
        ]
      },
```

Based on the existing implementation of `CustomPropertyValue` where `Value` has the same type we should modify `DefaultValue` to `any`.

zyfy29:
close #3692


github/orgs_properties.go
```diff
@@ -23,8 +23,8 @@ type CustomProperty struct {
 	ValueType string `json:"value_type"`
 	// Whether the property is required.
 	Required *bool `json:"required,omitempty"`
-	// Default value of the property.
-	DefaultValue *string `json:"default_value,omitempty"`
+	// Default value of the property. Can be null, string or array of strings.
+	DefaultValue any `json:"default_value,omitempty"`
```
gmlewis: @zyfy29 - for the a `null` to be passed through to the backend, we need to remove the `omitempty`.
Also, please make sure that there is a unit test where the value is `nil` (unassigned) so we demonstrate that the `null` is passed through to the backend.

```suggestion
	DefaultValue any `json:"default_value"`
```

gmlewis (approved):
Thank you, @zyfy29!
One minor tweak, please, and then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/orgs_properties.go
zyfy29: Done. Thanks!

zyfy29:
Oh, sorry, now it should be OK

gmlewis (approved):
Thank you, @zyfy29!
LGTM.

PR #3745, opened 2025-10-01
fix return type of GetDefaultCodeSecurityConfigurations

### LINKED ISSUE #3744: `GetDefaultCodeSecurityConfigurations` returns `[]*CodeSecurityConfiguration` but should return `CodeSecurityConfigurationWithDefaultForNewRepos`
The return type of `GetDefaultCodeSecurityConfigurations` should return `CodeSecurityConfigurationWithDefaultForNewRepos` to align with the API response schema: https://docs.github.com/en/rest/code-security/configurations?apiVersion=2022-11-28#get-default-code-security-configurations 

I can create a PR for this if you agree with the problem, this would also be a breaking change I guess.

ericmort:
Modifies the return type of `GetDefaultCodeSecurityConfigurations` to match the response schema of the API.

Fixes #3744


ericmort:
Wrong commit author probably causes Github to insist on missing CLA

PR #3747, opened 2025-10-02
fix!: Change return type of `GetDefaultCodeSecurityConfigurations`

ericmort:
BREAKING CHANGE: `GetDefaultCodeSecurityConfigurations` now returns `CodeSecurityConfigurationWithDefaultForNewRepos` instead of `CodeSecurityConfiguration`.

Modifies the return type of GetDefaultCodeSecurityConfigurations to match the response schema of the API.

Fixes: #3744.

### DIFF
--- a/github/orgs_codesecurity_configurations_test.go
+++ b/github/orgs_codesecurity_configurations_test.go
@@ -167,7 +167,7 @@ func TestOrganizationsService_GetDefaultCodeSecurityConfigurations(t *testing.T)
 			}
 		},
 		{
-			"default_for_new_repos": "private",
+			"default_for_new_repos": "private_and_internal",
 			"configuration": {
 				"id":2,
 				"name":"config2",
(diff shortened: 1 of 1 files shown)


github/orgs_codesecurity_configurations_test.go
```diff
@@ -159,25 +159,32 @@ func TestOrganizationsService_GetDefaultCodeSecurityConfigurations(t *testing.T)
...
+				"name":"config1",
+				"code_scanning_default_setup": "enabled"
+			}
 		},
 		{
-			"id":2,
-			"name":"config2",
-			"private_vulnerability_reporting": "enabled"
-		}]`)
+			"default_for_new_repos": "private",
```
zyfy29, line 170: `private` is not an allowed value here. See https://docs.github.com/en/rest/code-security/configurations?apiVersion=2022-11-28#get-default-code-security-configurations

zyfy29 (changes requested):

github/orgs_codesecurity_configurations_test.go
ericmortebidbax: good catch :) fixed in latest commit

zyfy29 (approved):

gmlewis (approved):
Thank you, @ericmort and @zyfy29!
LGTM.
Merging.

PR #3751, opened 2025-10-03
feat: Add premium request usage report endpoints for organizations and users

### LINKED ISSUE #3748: Premium Requests Analytics API
announced [here](https://github.blog/changelog/2025-09-30-premium-requests-analytics-page-is-now-generally-available/), with new API endpoints released

I could be of some help to implement it. Maybe you need to do `update-openapi` firstly.

zyfy29:
close #3748

### DIFF
--- a/github/billing.go
+++ b/github/billing.go
@@ -125,18 +125,21 @@ type PremiumRequestUsageItem struct {
 	NetAmount        float64 `json:"netAmount"`
 }
 
+// PremiumRequestUsageTimePeriod represents a time period for premium request usage reports.
+type PremiumRequestUsageTimePeriod struct {
+	Year  int  `json:"year"`
+	Month *int `json:"month,omitempty"`
+	Day   *int `json:"day,omitempty"`
+}
+
 // PremiumRequestUsageReport represents the premium request usage report response.
 type PremiumRequestUsageReport struct {
-	TimePeriod struct {
-		Year  int  `json:"year"`
-		Month *int `json:"month,omitempty"`
-		Day   *int `json:"day,omitempty"`
-	} `json:"timePeriod"`
-	Organization string                     `json:"organization"`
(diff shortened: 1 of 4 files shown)


github/billing.go
```diff
@@ -74,6 +74,21 @@ type UsageReportOptions struct {
 	Hour *int `url:"hour,omitempty"`
 }
 
+// PremiumRequestUsageReportOptions specifies optional parameters
+// for the enhanced billing platform premiun request usage report.
```
gmlewis:
```suggestion
// for the enhanced billing platform premium request usage report.
```

github/billing.go
```diff
@@ -74,6 +74,21 @@ type UsageReportOptions struct {
 	Hour *int `url:"hour,omitempty"`
 }
 
+// PremiumRequestUsageReportOptions specifies optional parameters
+// for the enhanced billing platform premium request usage report.
+type PremiumRequestUsageReportOptions struct {
+	UsageReportOptions
```
alexandear: `UsageReportOptions` contains the `Hour` field. But we can't set it in the `GetPremiumRequestUsageReportOrg` according to the documentation:

Image

Maybe we should mention this in a comment.

gmlewis:
> @gmlewis my concern is that the naming in the current PR is correct in the context of the existing functions so I'm unsure if this PR is the correct place to suggest wholesale changes? If it is would we be pushing for the existing functions to be renamed too?

Oh, I see. Yes, I think we should proceed with this PR and can you please open a new issue proposing widespread wholesale change recommendations, @stevehipwell ?

github/billing.go
zyfy29: Good catch! Let's define the fields individually.

gmlewis:
Thank you, @alexandear!
Merging.

PR #3756, opened 2025-10-04
Prefer `%v` over `%d`,`%s` and add `fmtpercentv` custom linter

gmlewis:
This PR formalizes the preference of `%v` over the use of `%s` or `%d` in this repo by adding a custom linter that enforces this preference.


gmlewis:
@stevehipwell - @alexandear - @zyfy29 - might you have time for a code review?  Thank you!

tools/urlpercentv/testdata/src/has-warnings/main.go
```diff
@@ -0,0 +1,15 @@
...
+
+package main
+
+import "fmt"
+
+func main() {
+	_ = fmt.Sprintf("some/%d/url", 1)          // want `use %v instead of %d`
+	_ = fmt.Sprintf("some/%s/url", "yo")       // want `use %v instead of %s`
+	_ = fmt.Sprintf("some/%s/%d/url", "yo", 1) // want `use %v instead of %s and %d`
+	_ = fmt.Sprintf("some/%d/%s/url", 1, "yo") // want `use %v instead of %s and %d`
```
zyfy29, line 14:
```diff
+	fmt.Printf("some/%d/url", 1)               // want `use %v instead of %d`
+	fmt.Printf("some/%s/url", "yo")            // want `use %v instead of %s`
```

The linter currently only catches violations in assignment statements (e.g., `url := fmt.Sprintf("some/%s", x)`), but misses
  standalone function calls like `fmt.Printf("some/%s", x)`. I've  reproduced it locally.

tools/fmtpercentv/testdata/src/has-warnings/main.go
```diff
@@ -0,0 +1,15 @@
...
+
+package main
+
+import "fmt"
+
+func main() {
+	_ = fmt.Sprintf("some/%d/url", 1)          // want `use %v instead of %d`
+	_ = fmt.Sprintf("some/%s/url", "yo")       // want `use %v instead of %s`
+	_ = fmt.Sprintf("some/%s/%d/url", "yo", 1) // want `use %v instead of %s and %d`
+	_ = fmt.Sprintf("some/%d/%s/url", 1, "yo") // want `use %v instead of %s and %d`
```
gmlewis: > The linter currently only catches violations in assignment statements (e.g., `url := fmt.Sprintf("some/%s", x)`), but misses
> standalone function calls like `fmt.Printf("some/%s", x)`.

I suppose that it wouldn't hurt to be consistent _throughout_ the entire repo.
I originally wrote it just to catch all the URL generators, but maybe it should do the whole repo.
Should I change the name of the linter to `percentv` instead of `urlpercentv`?

gmlewis: > I suppose that it wouldn't hurt to be consistent throughout the entire repo.

I now remember why I didn't do this. It was because `%s` is perfectly valid when you have a `[]byte` and I didn't want to have to flag all these usages with a `// nolint...` comment.

I believe the vast majority of these cases are in assignment statements, although I could be wrong. I could run some experiments.

tools/urlpercentv/testdata/src/has-warnings/main.go
zyfy29: Oh, yes, I didn't notice the linter name. It seems that all URL variable is created by a assignment statement. It's enough!

zyfy29 (approved):

tools/urlpercentv/testdata/src/has-warnings/main.go
zyfy29: While it could be avoided to use `%s` when converting a `[]byte` to `string` (the only usage for the remaining `%s`). IMO since we have come to this point, it's better to lightly refactor it and ban all `%s` usage without using `// nolint...`. But it's a trade-off.

tools/fmtpercentv/testdata/src/has-warnings/main.go
gmlewis: OK, that's a fair point.  Let me investigate by removing the requirement for making an assignment and see what happens.  First, though, I'll try to take care of unnecessary `.String()` calls... because that is so rare that I don't think we need a linter for it... or maybe `golangci` already has an option for it... :eyes: no, apparently it does not... and I still don't think we need a linter for that.

gmlewis:
OK, great idea, @zyfy29!  I've renamed `urlpercentv` to `fmtpercentv` and generalized it to lint _all_ call expression format strings, not just ones involving assignments. This only flagged 3 perfectly valid lines in the entire repo, so I've gone ahead and added the `//nolint` comments to those.

PTAL.

gmlewis:
Thank you, @zyfy29!
Merging.

PR #3759, opened 2025-10-06
fix existing test case for secret_scanning_pattern_configs_test to in…

alkak95:
This PR has a fix for existing test cases in secret_scanning_pattern_configs_test.go file which were not working.


alkak95:
this PR has some issue with the email id from which I pushed the changes

Issue #3799, opened 2025-11-03
Add support for new custom properties for orgs APIs


stevehipwell:
GitHub [announced](https://github.blog/changelog/2025-10-23-organization-custom-properties-are-now-available-in-public-preview/) support for organization custom properties. This will also require support adding to [enterprise rules](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/rules) as the selector can now use organization properties.

- Enterprise [REST API](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs) to configure custom properties
- Organization [REST API](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-properties-for-orgs) to set custom properties

gmlewis:
This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.
Also feel free to perform a code review on any PR that has the label "NeedsReview" - we appreciate those contributions too!
We have a `REVIEWERS` file that you could add your username to if you are willing to be contacted when reviews are needed.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

Not-Dhananjay-Mishra:
I’d love to work on this!
I just have a couple of questions:

- Should I create two new files one for the `Enterprise REST API to configure custom properties` and another for the `Organization REST API` or should I add the implementation to the existing files? If new files are preferred, could you please suggest appropriate names for them?
- Do I also need to make any changes in enterprise_rules.go?

stevehipwell:
@Not-Dhananjay-Mishra thank you, the issue is yours.

> Should I create two new files one for the Enterprise REST API to configure custom properties and another for the Organization REST API or should I add the implementation to the existing files? If new files are preferred, could you please suggest appropriate names for them?

I think that would make sense to have two new files, but naming is hard especially when GitHub features are involved.

> Do I also need to make any changes in enterprise_rules.go?

Yes, enterprise rules can now target orgs by property.

/assign @Not-Dhananjay-Mishra

Not-Dhananjay-Mishra:
@gmlewis I am getting this error while runing `go generate ./...`
[...]

PR #3804, opened 2025-11-05
Add support for new custom properties for orgs APIs

Not-Dhananjay-Mishra:
Fixes: #3799.

This PR adds support for 

- Enterprise [REST API](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs) to configure custom properties
- Organization [REST API](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-properties-for-orgs) to set custom properties
- Add organization_property support to enterprise rules

### DIFF
--- a/github/enterprise_organization_properties.go
+++ b/github/enterprise_organization_properties.go
@@ -139,13 +139,13 @@ func (s *EnterpriseService) DeleteOrganizationCustomProperty(ctx context.Context
 	return resp, nil
 }
 
-// ListEnterpriseCustomPropertyValues lists enterprise organizations with all of their custom property values.
+// ListOrganizationCustomPropertyValues lists enterprise organizations with all of their custom property values.
 // Returns a list of organizations and their custom property values defined in the enterprise.
 //
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs#list-custom-property-values-for-organizations-in-an-enterprise
 //
 //meta:operation GET /enterprises/{enterprise}/org-properties/values
-func (s *EnterpriseService) ListEnterpriseCustomPropertyValues(ctx context.Context, enterprise string, opts *ListOptions) ([]*EnterpriseCustomPropertiesValues, *Response, error) {
+func (s *EnterpriseService) ListOrganizationCustomPropertyValues(ctx context.Context, enterprise string, opts *ListOptions) ([]*EnterpriseCustomPropertiesValues, *Response, error) {
 	u := fmt.Sprintf("enterprises/%v/org-properties/values", enterprise)
 
 	u, err := addOptions(u, opts)
@@ -167,13 +167,13 @@ func (s *EnterpriseService) ListEnterpriseCustomPropertyValues(ctx context.Conte
 	return values, resp, nil
 }
 
-// CreateOrUpdateEnterpriseCustomPropertyValues creates or updates custom property values for organizations in an enterprise.
+// CreateOrUpdateOrganizationCustomPropertyValues creates or updates custom property values for organizations in an enterprise.
 // To remove a custom property value from an organization, set the property value to null.
 //
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs#create-or-update-custom-property-values-for-organizations-in-an-enterprise
 //
 //meta:operation PATCH /enterprises/{enterprise}/org-properties/values
-func (s *EnterpriseService) CreateOrUpdateEnterpriseCustomPropertyValues(ctx context.Context, enterprise string, values EnterpriseCustomPropertyValuesRequest) (*Response, error) {
+func (s *EnterpriseService) CreateOrUpdateOrganizationCustomPropertyValues(ctx context.Context, enterprise string, values EnterpriseCustomPropertyValuesRequest) (*Response, error) {
 	u := fmt.Sprintf("enterprises/%v/org-properties/values", enterprise)
 	req, err := s.client.NewRequest("PATCH", u, values)
 	if err != nil {
--- a/github/enterprise_organization_properties_test.go
+++ b/github/enterprise_organization_properties_test.go
@@ -184,7 +184,7 @@ func TestEnterpriseService_DeleteOrganizationCustomProperty(t *testing.T) {
 	})
 }
 
-func TestEnterpriseService_ListEnterpriseCustomPropertyValues(t *testing.T) {
+func TestEnterpriseService_ListOrganizationCustomPropertyValues(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
 
@@ -201,9 +201,9 @@ func TestEnterpriseService_ListEnterpriseCustomPropertyValues(t *testing.T) {
 
 	ctx := t.Context()
 	opts := &ListOptions{Page: 1, PerPage: 10}
-	got, _, err := client.Enterprise.ListEnterpriseCustomPropertyValues(ctx, "e", opts)
+	got, _, err := client.Enterprise.ListOrganizationCustomPropertyValues(ctx, "e", opts)
 	if err != nil {
-		t.Fatalf("Enterprise.ListEnterpriseCustomPropertyValues returned error: %v", err)
+		t.Fatalf("Enterprise.ListOrganizationCustomPropertyValues returned error: %v", err)
 	}
 
 	want := []*EnterpriseCustomPropertiesValues{
@@ -217,16 +217,16 @@ func TestEnterpriseService_ListEnterpriseCustomPropertyValues(t *testing.T) {
 	}
 
 	if !cmp.Equal(got, want) {
-		t.Errorf("Enterprise.ListEnterpriseCustomPropertyValues = %+v, want %+v", got, want)
+		t.Errorf("Enterprise.ListOrganizationCustomPropertyValues = %+v, want %+v", got, want)
 	}
 
 	const methodName = "ListEnterpriseCustomPropertyValues"
 	testBadOptions(t, methodName, func() error {
-		_, _, err := client.Enterprise.ListEnterpriseCustomPropertyValues(ctx, "\n", opts)
+		_, _, err := client.Enterprise.ListOrganizationCustomPropertyValues(ctx, "\n", opts)
 		return err
 	})
 	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
-		got, resp, err := client.Enterprise.ListEnterpriseCustomPropertyValues(ctx, "e", nil)
+		got, resp, err := client.Enterprise.ListOrganizationCustomPropertyValues(ctx, "e", nil)
 		if got != nil {
 			t.Errorf("testNewRequestAndDoFailure %v = %#v, want nil", methodName, got)
 		}
@@ -234,7 +234,7 @@ func TestEnterpriseService_ListEnterpriseCustomPropertyValues(t *testing.T) {
 	})
 }
 
-func TestEnterpriseService_CreateOrUpdateEnterpriseCustomPropertyValues(t *testing.T) {
+func TestEnterpriseService_CreateOrUpdateOrganizationCustomPropertyValues(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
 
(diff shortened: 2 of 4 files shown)


Not-Dhananjay-Mishra:
@stevehipwell did i implement organization_property support to enterprise rules correctly?

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,220 @@
...
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// Property represents a custom property definition in an enterprise.
+type Property struct {
```
stevehipwell: We already have the `CustomProperty` type which can be used for these properties so you don't need to create  anew type. You can add `URL` to that as it's currently missing.

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// EnterpriseCustomPropertySchema represents the schema response for GetEnterpriseCustomPropertiesSchema.
+type EnterpriseCustomPropertySchema struct {
+	// An ordered list of the custom property defined in the enterprise.
+	Properties []*Property `json:"properties,omitempty"`
+}
+
+// PropertyValue represents the custom property values for an organization.
+type PropertyValue struct {
```
stevehipwell: We already have the `CustomPropertyValue` type so this type isn't needed.

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,220 @@
...
+	// The name of the property
+	PropertyName *string `json:"property_name,omitempty"`
+	// The value assigned to the property
+	Value *string `json:"value,omitempty"`
+}
+
+// CustomPropertiesValues represents the custom properties values for an organization within an enterprise.
+type CustomPropertiesValues struct {
```
stevehipwell:
```suggestion
type EnterpriseCustomPropertiesValues struct {
```
This type is enterprise scoped so should have the `Enterprise` prefix.

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,220 @@
...
+	// The Organization ID that the custom property values will be applied to.
+	OrganizationID *int64 `json:"organization_id,omitempty"`
+	// The names of organizations that the custom property values will be applied to.
+	OrganizationLogin *string `json:"organization_login,omitempty"`
+	// List of custom property names and associated values to apply to the organizations.
+	Properties []*PropertyValue `json:"properties,omitempty"`
+}
+
+// EnterpriseCustomPropertyValues represents the request to update custom property values for organizations within an enterprise.
+type EnterpriseCustomPropertyValues struct {
```
stevehipwell: This type either isn't needed as the above type could be used or it needs a different name such as `EnterpriseCustomPropertyValuesRequest`.

github/enterprise_custom_properties_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestOrganizationsService_GetOrganizationCustomPropertyValues(t *testing.T) {
```
stevehipwell, line 16: I think these tests are in the wrong file?

github/organizations_custom_properties_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestEnterpriseService_GetEnterpriseCustomPropertySchema(t *testing.T) {
```
stevehipwell: These tests looks like they're in the wrong file?

github/rules.go
```diff
@@ -217,6 +218,19 @@ type RepositoryRulesetRefConditionParameters struct {
 	Exclude []string `json:"exclude"`
 }
 
+// RepositoryRulesetOrganizationPropertyTargetParameters represents an organization_property name and values to be used for targeting.
+type RepositoryRulesetOrganizationPropertyTargetParameters struct {
```
stevehipwell: We already have the `RepositoryRulesetRepositoryPropertyTargetParameters` type, we should re-use this type.

Not-Dhananjay-Mishra:
@stevehipwell Thanks for the suggestions.
I have done following changes -

- Re-use struct that already present
- CustomProperty now have one more field named URL
- Change struct name as suggested 
- Swap the test file, really apologize for that mistake.
- change filename from `organizations_..go` to `orgs_..go`

gmlewis (approved):
Thank you, @Not-Dhananjay-Mishra and @stevehipwell!
LGTM.

@stevehipwell - we have a few options here:
Option A: Merge this and work on #3692 before cutting the next release
Option B: Work on #3692 before merging this

Now that I write these out, I'm betting you meant Option A, but please correct me if I'm wrong.

Either way, I will wait for your LGTM and reply before proceeding.

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+	// List of custom property names and associated values to apply to the organizations.
+	Properties []*CustomPropertyValue `json:"properties"`
+}
+
+// GetEnterpriseCustomPropertySchema gives all organization custom property definitions that are defined on an enterprise.
+//
+// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs#get-organization-custom-properties-schema-for-an-enterprise
+//
+//meta:operation GET /enterprises/{enterprise}/org-properties/schema
+func (s *EnterpriseService) GetEnterpriseCustomPropertySchema(ctx context.Context, enterprise string) (*EnterpriseCustomPropertySchema, *Response, error) {
```
stevehipwell:
```suggestion
func (s *EnterpriseService) GetOrganizationCustomPropertySchema(ctx context.Context, enterprise string) (*EnterpriseCustomPropertySchema, *Response, error) {
```
As we're in the enterprise service we don't need `Enterprise` in the name, but these are organization custom properties according to the docs. The same applies to the other methods.

stevehipwell (changes requested):
I think the naming here is hard, and a per my last comment I think the existing repository custom properties need some changes. But let's get the changes in this PR correct and then we only need to update the existing code.

In addition to my inline comment think the new files need to follow the pattern `[enterprise|orgs]_organization_properties.go`.

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+// EnterpriseCustomPropertyValuesRequest represents the request to update custom property values for organizations within an enterprise.
+type EnterpriseCustomPropertyValuesRequest struct {
+	// The names of organizations that the custom property values will be applied to.
+	// OrganizationLogin specifies the organization name when updating multiple organizations.
+	OrganizationLogin []string `json:"organization_login"`
+	// List of custom property names and associated values to apply to the organizations.
+	Properties []*CustomPropertyValue `json:"properties"`
+}
+
+// GetEnterpriseCustomPropertySchema gives all organization custom property definitions that are defined on an enterprise.
```
alexandear:
```suggestion
// GetEnterpriseCustomPropertySchema gets all organization custom property definitions that are defined on an enterprise.
```

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+	var values []*EnterpriseCustomPropertiesValues
+	resp, err := s.client.Do(ctx, req, &values)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return values, resp, nil
+}
+
+// CreateOrUpdateEnterpriseCustomPropertyValues create or update custom property values for organizations in an enterprise.
```
alexandear:
```suggestion
// CreateOrUpdateEnterpriseCustomPropertyValues creates or updates custom property values for organizations in an enterprise.
```

github/enterprise_custom_properties.go
Not-Dhananjay-Mishra: @stevehipwell I made some changes in method name - 

- GetEnterpriseCustomPropertySchema -> GetOrganizationCustomPropertySchema
- CreateOrUpdateEnterpriseCustomPropertySchema -> CreateOrUpdateOrganizationCustomPropertySchema
- GetEnterpriseCustomProperty -> GetOrganizationCustomProperty
- CreateOrUpdateEnterpriseCustomProperty -> CreateOrUpdateOrganizationCustomProperty
- DeleteEnterpriseCustomProperty -> DeleteOrganizationCustomProperty

github/orgs_custom_properties.go
```diff
@@ -0,0 +1,59 @@
...
+	var values []*CustomPropertyValue
+	resp, err := s.client.Do(ctx, req, &values)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return values, resp, nil
+}
+
+// CreateOrUpdateOrgCustomPropertyValues creates or updates custom property values for an organization.
```
alexandear: This endpoint allows us to remove a custom property value from an organization. I propose to mention this in the comment.

Image

Not-Dhananjay-Mishra: @alexandear thanks for suggestion, fixed that.

gmlewis:
Oh, sorry - @stevehipwell - I see you had comments - do you approve this PR now?

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+	return resp, nil
+}
+
+// ListEnterpriseCustomPropertyValues lists enterprise organizations with all of their custom property values.
+// Returns a list of organizations and their custom property values defined in the enterprise.
+//
+// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs#list-custom-property-values-for-organizations-in-an-enterprise
+//
+//meta:operation GET /enterprises/{enterprise}/org-properties/values
+func (s *EnterpriseService) ListEnterpriseCustomPropertyValues(ctx context.Context, enterprise string, opts *ListOptions) ([]*EnterpriseCustomPropertiesValues, *Response, error) {
```
stevehipwell:
```suggestion
func (s *EnterpriseService) ListOrganizationCustomPropertyValues(ctx context.Context, enterprise string, opts *ListOptions) ([]*EnterpriseCustomPropertiesValues, *Response, error) {
```
This name still doesn't look right, we're in the enterprise service and there is no such thing as an enterprise custom property.

github/enterprise_custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+	return values, resp, nil
+}
+
+// CreateOrUpdateEnterpriseCustomPropertyValues creates or updates custom property values for organizations in an enterprise.
+// To remove a custom property value from an organization, set the property value to null.
+//
+// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/custom-properties-for-orgs#create-or-update-custom-property-values-for-organizations-in-an-enterprise
+//
+//meta:operation PATCH /enterprises/{enterprise}/org-properties/values
+func (s *EnterpriseService) CreateOrUpdateEnterpriseCustomPropertyValues(ctx context.Context, enterprise string, values EnterpriseCustomPropertyValuesRequest) (*Response, error) {
```
stevehipwell:
```suggestion
func (s *EnterpriseService) CreateOrUpdateOrganizationCustomPropertyValues(ctx context.Context, enterprise string, values EnterpriseCustomPropertyValuesRequest) (*Response, error) {
```
Same as above.

github/orgs_custom_properties.go
```diff
@@ -0,0 +1,60 @@
...
+// To remove a custom property value from an organization, set the property value to null.
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/custom-properties-for-orgs#create-or-update-custom-property-values-for-an-organization
+//
+//meta:operation PATCH /organizations/{org}/org-properties/values
+func (s *OrganizationsService) CreateOrUpdateOrgCustomPropertyValues(ctx context.Context, org string, values OrganizationCustomPropertyValues) (*Response, error) {
```
stevehipwell: The name of this function doesn't match the function above organization/org.

Not-Dhananjay-Mishra: Thanks for suggestion, I have changed it to CreateOrUpdateOrganizationCustomPropertyValues. Do I need to change it to something else?

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Awaiting approval from @stevehipwell.

gmlewis:
Thank you, @stevehipwell!
Merging.

PR #3818, opened 2025-11-10
fix: Add `DownloadContent` to support downloads of any size

### LINKED ISSUE #2480: DownloadContents can't download files from root of repo
GitHub Apps have the ability to request access to individual files which is great for fine-grained access.

image

However, as of `v47.1.0`, this will prevent someone from being able to download a file from the root repository because of the way `DownloadContents` implements directory scanning first:

https://github.com/google/go-github/blob/8a4bdb5e400fef2e8d783d35d8f4cfd1bc79c01e/github/repos_contents.go#L129-L134

The first thing that the function does is get just the directory name with `path.Dir` but if you try to download a file like just `CODEOWNERS` which is a valid location for a `CODEOWNERS` file, then the first request to the API will create a URL like so:
```
https://api.github.com/repos/$owner/$repo/contents/
```

At the moment, there's no way to add the root directory to the list of single file paths. You cannot add a blank path and adding just `/` doesn't work either.

### LINKED ISSUE #3810: DownloadContents can't always download files in large directories
`*github.RepositoriesService` provides a special helper method called `DownloadContents` [[code link](https://github.com/google/go-github/blob/876028969d13dbdb40e9b63e9850b2e2bbbe5d6a/github/repos_contents.go#L139)] to get the contents of a file even if it's big enough that the result from `GetContents` method won't include them. `DownloadContents` works by first trying `GetContents` on the file itself, and then if that doesn't work (or doesn't seem to work, e.g. because the file is just genuinely empty), it falls back to calling `GetContents` on the containing directory, iterating over the results until it gets to the desired file, and then using the `DownloadUrl` for that desired file. (There's also another helper method called `DownloadContentsWithMeta` [[code link](https://github.com/google/go-github/blob/876028969d13dbdb40e9b63e9850b2e2bbbe5d6a/github/repos_contents.go#L188)] that does the same thing but returns a bit more information from the API calls; everything here applies to that method as well.)

The problem is, when you call `GetContents` on a directory, it only returns the first 1000 files. [[doc link](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content)] So if the initial `GetContents` call (the one for the specific file) didn't work or didn't seem to work, and if the file isn't one of the first 1000 in the directory, then this approach will fail with the error `"no file named %s found in %s"`.

The good news is, there's a simple fix: we don't actually need to call `GetContents` on the containing directory at all, because the initial call to `GetContents` on the individual file already gives the same information (including the `DownloadUrl`), even if the file is big enough that it doesn't contain the actual file contents. So we can just use that. In addition to fixing this issue, that would also fix https://github.com/google/go-github/issues/2480. (In fact, I see that this same solution has also been suggested in comments there.)

sheeeng:
Fix https://github.com/google/go-github/issues/3810.
Fix https://github.com/google/go-github/issues/2480.

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -17,7 +17,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 )
 
@@ -98,6 +97,54 @@ func (r *RepositoryContent) GetContent() (string, error) {
 	}
 }
 
+// DownloadContent downloads the content of a file using the information from
+// the RepositoryContent. This method supports files of any size and works with
+// RepositoryContent entries obtained from GetContents calls on both individual
+// files and directories.
+//
+// It returns an io.ReadCloser for the file content and an *http.Response for
+// the download request (which may be nil if content was obtained without an
+// HTTP request). It is the caller's responsibility to close the ReadCloser.
+func (r *RepositoryContent) DownloadContent(ctx context.Context, httpClient *http.Client) (io.ReadCloser, *http.Response, error) {
+	// Check if this is a file.
+	if r.Type == nil || *r.Type != "file" {
+		return nil, nil, fmt.Errorf("cannot download content for type %q (only files are supported)", r.GetType())
+	}
+
+	// Handle empty files.
+	if r.Size != nil && *r.Size == 0 {
+		return io.NopCloser(strings.NewReader("")), nil, nil
+	}
+
+	// Try to use the inline content if available.
+	if r.Content != nil && *r.Content != "" {
+		content, err := r.GetContent()
+		if err == nil {
+			return io.NopCloser(strings.NewReader(content)), nil, nil
+		}
+		// If GetContent fails (e.g., encoding "none"), fall through to use DownloadURL.
+	}
+
+	// Use the download URL.
+	if r.DownloadURL == nil || *r.DownloadURL == "" {
+		return nil, nil, errors.New("no download URL available for this file")
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "GET", *r.DownloadURL, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Return the response body even for non-2xx status codes.
+	// Callers should check resp.StatusCode to verify success.
+	return resp.Body, resp, nil
+}
+
 // GetReadme gets the Readme file for the repository.
 //
 // GitHub API docs: https://docs.github.com/rest/repos/contents#get-a-repository-readme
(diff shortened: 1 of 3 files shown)


gmlewis (commented):
My biggest concern is that we now have:
* DownloadContent
* DownloadContents
* DownloadContentsWithMeta
* GetContent
* GetContents

How is a user supposed to understand which one to choose?

sheeeng:
> My biggest concern is that we now have:
> 
> * DownloadContent
> * DownloadContents
> * DownloadContentsWithMeta
> * GetContent
> * GetContents
> 
> How is a user supposed to understand which one to choose?

Good point. I will think about it.

sheeeng:
> * DownloadContents
> * DownloadContentsWithMeta

These two are existing functions. Do we want to expose `DownloadContents` only for use?

sheeeng:
> Do we want to expose `DownloadContents` only for use?

The `DownloadContentsWithMeta` function and its references were deleted. Is that preferred?

gmlewis:
> The `DownloadContentsWithMeta` function and its references were deleted. Is that preferred?

Well, it seems we have a number of people interested in this topic based on the conversation in #2480 and #3810.

I would really like to get some feedback from the individuals who care about these endpoints before we start changing it dramatically, and the only way I can think of doing that is to start listing GitHub logins that commented, unfortunately.

cc: Please provide your feedback on how these endpoints should be handled:
* @ran-arigur
* @abatilo
* @zaataylor
* @petersondmg
* @githoober
* @JayDubyaEey

ran-arigur:
> My biggest concern is that we now have:
> 
> * DownloadContent
> * DownloadContents
> * DownloadContentsWithMeta
> * GetContent
> * GetContents
> 
> How is a user supposed to understand which one to choose?

(Full disclosure: Adding DownloadContent was my suggestion, so I can't independently say whether it will make sense to other people.)

FWIW, those methods are not all on the same type. Grouping it a bit differently:

* on RepositoriesService:
    * GetContents(...)
        * is genuinely a bit confusing, but I don't think we should change it, because it maps pretty directly to an underlying API call [[doc link](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content)] and what makes it confusing is that that API is confusing.
        * returns either a single RepositoryContent or a slice of RepositoryContents, depending on whether the path is a file or a directory.
    * DownloadContents(...), DownloadContentsWithMeta(...)
        * convenience wrappers around GetContents(...) + DownloadContent().
        * behave identically, but DownloadContentsWithMeta(...) returns more information. I assume that DownloadContents(...) was created first, and then DownloadContentsWithMeta(...) was added later because there was no backward-compatible way to have DownloadContents(...) return the additional information.
* on RepositoryContent (as returned by RepositoriesService.GetContents(...)):
    * GetContent()
        * returns the file content, if already available on the RepositoryContent. handles Base64-decoding if necessary.
        * doesn't handle cases where the RepositoryContent _doesn't_ contain the file content, e.g. if the file is >1MB (in which case it returns an error) or the RepositoryContent came from listing a directory (in which case it returns the empty string even if the file is non-empty).
    * DownloadContent()
        * equivalent to GetContent(), except that it works even if the RepositoryContent doesn't contain the file content: in that case it fetches it using the RepositoryContent's DownloadURL.

I think that, with reasonable comments on each, users will be able to find a method that does what they need, and ignore the rest.

sheeeng:
Setting this PR as draft as waiting for more feedback.

gmlewis:
Closing, as this was resolved by #4153.

Issue #3822, opened 2025-11-12
add missing mergequeue rule configuration properties


patrickcarnahan:
the merge queue configuration is missing a couple of properties that are available in the raw api:

```json
{
    "type": "merge_queue",
    "parameters": map[string]interface{}{
        "merge_method":                      "MERGE",
        "max_entries_to_build":              5,
        "min_entries_to_merge":              1,
        "max_entries_to_merge":              5,
        "min_entries_to_merge_wait_minutes": 1,
        "grouping_strategy":                 "ALLGREEN",
        "check_response_timeout_minutes":    60,
        "check_run_retries_limit":           0,
        "actor_controlled_merging":          true,
    },
}
```

the current support in the merge queue ruleset is represented by the existing struct:

```go
// MergeQueueRuleParameters represents the merge_queue rule parameters.
type MergeQueueRuleParameters struct {
	MergeMethod                  MergeQueueMergeMethod `json:"merge_method"`
	MaxEntriesToBuild            int                   `json:"max_entries_to_build"`
	MinEntriesToMerge            int                   `json:"min_entries_to_merge"`
	MaxEntriesToMerge            int                   `json:"max_entries_to_merge"`
	MinEntriesToMergeWaitMinutes int                   `json:"min_entries_to_merge_wait_minutes"`
	GroupingStrategy             MergeGroupingStrategy `json:"grouping_strategy"`
	CheckResponseTimeoutMinutes  int                   `json:"check_response_timeout_minutes"`
}
```

from this diff we need to implement two additional properties:

```go 
CheckRunRetriesLimit   int  `json:"check_run_retries_limit"`
ActorControlledMerging bool `json:"actor_controlled_merging"`
```

i'm happy to self-assign and add this functionality myself if there is reasonable expectation it will be merged.

patrickcarnahan:
went ahead and opened a pull request to add these properties since it was pretty straight-forward. i don't believe there is anything else to be done on the PR but happy to course correct as necessary.

patrickcarnahan:
mentioned this in the pr but these properties aren't documented nor are they available for most organizations or repos, so will take a different approach as these will not apply for the majority of use cases.

PR #3823, opened 2025-11-12
Add missing `MergeQueueRuleParameters` config properties

patrickcarnahan:
fixes https://github.com/google/go-github/issues/3822

This adds the missing properties `actor_controlled_merging` and `check_run_retries_limit` to the merge queue rule configuration object. The new properties are added as `omitempty` to avoid behavioral changes for existing logic that may not be setting these values. 

Happy to adjust course and remove `omitempty` and fix tests if that's the desired take.

### DIFF
--- a/github/rules.go
+++ b/github/rules.go
@@ -415,7 +415,9 @@ type UpdateRuleParameters struct {
 
 // MergeQueueRuleParameters represents the merge_queue rule parameters.
 type MergeQueueRuleParameters struct {
+	ActorControlledMerging       bool                  `json:"actor_controlled_merging,omitempty"`
 	CheckResponseTimeoutMinutes  int                   `json:"check_response_timeout_minutes"`
+	CheckRunRetriesLimit         int                   `json:"check_run_retries_limit,omitempty"`
 	GroupingStrategy             MergeGroupingStrategy `json:"grouping_strategy"`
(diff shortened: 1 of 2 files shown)


github/rules.go
```diff
@@ -415,7 +415,9 @@ type UpdateRuleParameters struct {
 
 // MergeQueueRuleParameters represents the merge_queue rule parameters.
 type MergeQueueRuleParameters struct {
+	ActorControlledMerging       bool                  `json:"actor_controlled_merging,omitempty"`
```
gmlewis, line 418: If this is an optional parameter, as indicated by the inclusion of `omitempty`, then this should be a `*bool` instead of a `bool`.

github/rules.go
```diff
@@ -415,7 +415,9 @@ type UpdateRuleParameters struct {
...
 	CheckResponseTimeoutMinutes  int                   `json:"check_response_timeout_minutes"`
+	CheckRunRetriesLimit         int                   `json:"check_run_retries_limit,omitempty"`
```
gmlewis, line 420: Same here. This should be `*int` instead of `int`.

patrickcarnahan:
@gmlewis thanks for the quick turnaround! unfortunately i should have done a little more research first as these properties are not documented nor are they available first repos. for those reasons i think the right thing to do is close this pr out for the time being. 

thanks again!

PR #3852, opened 2025-12-04
feat!: Implement Enterprise SCIM - Provision Groups & Users

elminster-aom:
BREAKING CHANGE: `SCIMEnterpriseDisplayReference.Ref` is now of type `*string`.

Related with #3813

From REST API endpoints for Enterprise SCIM docs, implements:

- [Provision a SCIM enterprise group](https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/scim#provision-a-scim-enterprise-group), POST /scim/v2/enterprises/{enterprise}/Groups

- [Provision a SCIM enterprise user](https://docs.github.com/enterprise-cloud@latest/rest/enterprise-admin/scim#provision-a-scim-enterprise-user), POST /scim/v2/enterprises/{enterprise}/Users

### Breaking API change

When provisioning Groups, the `SCIMEnterpriseDisplayReference.Ref` is optional for provisioning actions so it changed to pointer.

### DIFF
--- a/github/enterprise_scim.go
+++ b/github/enterprise_scim.go
@@ -42,7 +42,7 @@ type SCIMEnterpriseGroupAttributes struct {
 // SCIMEnterpriseDisplayReference represents a JSON SCIM (System for Cross-domain Identity Management) resource reference.
 type SCIMEnterpriseDisplayReference struct {
 	Value   string  `json:"value"`             // The local unique identifier for the member (e.g., user ID or group ID).
-	Ref     *string `json:"$ref,omitempty"`    // The URI reference to the member resource (e.g., https://api.github.com/scim/v2/Users/{id}).
+	Ref     *string `json:"$ref,omitempty"`    // The URI reference to the Members or Groups resource (e.g., /scim/v2/enterprises/{enterprise}/Users/{scim_user_id}).
 	Display *string `json:"display,omitempty"` // The display name associated with the member (e.g., user name or group name).
 }
 


github/enterprise_scim.go
```diff
@@ -33,16 +33,16 @@ type SCIMEnterpriseGroupAttributes struct {
 	DisplayName *string                           `json:"displayName,omitempty"` // Human-readable name for a group.
 	Members     []*SCIMEnterpriseDisplayReference `json:"members,omitempty"`     // List of members who are assigned to the group in SCIM provider
 	ExternalID  *string                           `json:"externalId,omitempty"`  // This identifier is generated by a SCIM provider. Must be unique per user.
```
stevehipwell:
```suggestion
	ExternalID  *string                           `json:"externalId,omitempty"`  // This identifier is generated by a SCIM provider. Must be unique per group.
```
I assume this is a typo?

elminster-aom: Good catch! thanks 😃

elminster-aom:
@gmlewis Do I need to take action regarding this drop of -0.10% for `codecov/project`?

gmlewis:
> @gmlewis Do I need to take action regarding this drop of -0.10% for `codecov/project`?

No, thank you though, @elminster-aom. The Codecov reports are mostly informational, and I'm not always convinced that they are helpful. Sometimes it will catch cases where unit tests could have added a `testBadOptions` call, for example, but in this case and many other cases they are simply a bit noisy.

gmlewis:
@elminster-aom - it looks like some conflicts have emerged from other merges. Would you like to address them, or would you like me to do that? I'm fine either way. Some people like a fun challenge, others don't. :joy: Your call.

elminster-aom:
> it looks like some conflicts have emerged from other merges. Would you like to address them, ... ?

Sure. Always the fanny option for me, of course 😁 . Thanks @gmlewis

gmlewis:
Thank you, @stevehipwell and @elminster-aom!
Merging.

PR #3874, opened 2025-12-10
fix!: Change `copilot_code_review` field names to match GitHub API

JiayangZhou:
BREAKING CHANGE: `CopilotCodeReviewRuleParameters.ReviewNewPushes` is now `ReviewOnPush`.

Rename ReviewNewPushes to ReviewOnPush with json tag review_on_push
 
Hey sorry when I reviewed the github API doc https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28 I noticed that the payload parameter we previously used was incorrect, thus I am creating this PR to correct it, sorry for the innocence! 

Edit: also added MarshalJSON tests under `func TestRepositoryRule.` currently only for `copilot_code_review` 'creation' i guess others can add if they want

Edit2:
I think you are right from the beginning, i was too naive, so for this API https://github.com/google/go-github/blob/1830689bcbcb8b16aa7a07f34f4f8a9712fc1478/github/repos_rules.go#L41 the respond body is Unmarshaling to BranchRules and it does include codepilot, 
 i verified it from here too https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-rules-for-a-branch

### DIFF
--- a/github/rules_test.go
+++ b/github/rules_test.go
@@ -701,6 +701,31 @@ func TestBranchRules(t *testing.T) {
 			})
 		}
 	})
+
+	t.Run("UnmarshalJSON_Error", func(t *testing.T) {
+		t.Parallel()
+
+		tests := []struct {
+			name string
+			json string
+		}{
+			{
+				"invalid_copilot_code_review_parameters",
+				`[{"type":"copilot_code_review","ruleset_source_type":"Repository","ruleset_source":"test/test","ruleset_id":1,"parameters":"not_an_object"}]`,
+			},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				t.Parallel()
+				got := &BranchRules{}
+				err := json.Unmarshal([]byte(tt.json), got)
+				if err == nil {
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @JiayangZhou!
LGTM.

gmlewis:
Thank you, @stevehipwell!
Merging.

mkushakov:
> I noticed that the payload parameter we previously used was incorrect

@JiayangZhou It is not your fault. GitHub just changed payload params for Copilot Code Review to move it from Pull Request property boolean to Rules root object without bumping API version last week. We encountered same issue. And since this field was in Preview, they didn't even bother to announce  breaking change API.

stevehipwell:
It's a good job they [added API versioning](https://github.blog/developer-skills/github/to-infinity-and-beyond-enabling-the-future-of-githubs-rest-api-with-api-versioning/)... 👀

tim-goto:
Hey @gmlewis would you mind creating a new release that includes this change? Thanks a lot!

gmlewis:
> Hey @gmlewis would you mind creating a new release that includes this change? Thanks a lot!

Done:
https://github.com/google/go-github/releases/tag/v81.0.0

PR #3877, opened 2025-12-11
docs: Clarify `CreateTree` semantics and `TreeEntry` usage

rogpeppe:
## Summary

This PR improves the documentation for `CreateTree` and `TreeEntry` to clarify several behaviors that are not obvious from the existing docs:

- **Merge semantics**: When `baseTree` is provided, entries are merged with that tree - paths not mentioned in `entries` are preserved from the base tree
- **Duplicate entries**: If the same path appears multiple times in entries, the last entry wins
- **File/directory deletion**: To delete an entry, set both `SHA` and `Content` to nil; this works for both files and entire directories
- **Type/Mode ignored on delete**: When deleting, the `Type` and `Mode` fields are ignored by the API; only `Path` is required
- **Content vs SHA**: Clarifies when to use `Content` (small text files) vs `SHA` (reference to existing blob), with a reference to `CreateBlob` for binary files

These behaviors were verified experimentally against the GitHub API.

## Test plan

- [x] Documentation-only change, no code changes
- [x] Existing tests pass (`go test ./github/... -run TestGitService_.*Tree`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -30,6 +30,13 @@ func (t Tree) String() string {
 // TreeEntry represents the contents of a tree structure. TreeEntry can
 // represent either a blob, a commit (in the case of a submodule), or another
 // tree.
+//
+// When used with [GitService.CreateTree], set Content for small text files,
+// or set SHA to reference an existing blob (use [GitService.CreateBlob] for
+// binary files or large content). To delete an entry, set both Content and SHA
+// to nil; the entry will be serialized with `"sha": null` which the API interprets
+// as a deletion. When deleting, the Type and Mode fields are ignored; only Path
+// is required.
 type TreeEntry struct {
 	SHA     *string `json:"sha,omitempty"`
 	Path    *string `json:"path,omitempty"`
@@ -126,6 +133,12 @@ type createTree struct {
 // CreateTree creates a new tree in a repository. If both a tree and a nested
 // path modifying that tree are specified, it will overwrite the contents of
 // that tree with the new path contents and write a new tree out.
+//
+// When baseTree is provided, entries are merged with that tree: paths not
+// mentioned in entries are preserved from the base tree. If the same path
+// appears multiple times in entries, the last entry wins. To delete an entry,
+// include a [TreeEntry] with the path and both SHA and Content set to nil.
+// Entire directories can be deleted this way.
 //
 // GitHub API docs: https://docs.github.com/rest/git/trees#create-a-tree
 //


github/git_trees.go
```diff
@@ -30,6 +30,13 @@ func (t Tree) String() string {
 // TreeEntry represents the contents of a tree structure. TreeEntry can
 // represent either a blob, a commit (in the case of a submodule), or another
 // tree.
+//
+// When used with [GitService.CreateTree], set Content for small text files,
+// or set SHA to reference an existing blob (use [GitService.CreateBlob] for
+// binary files or large content). To delete an entry, set both Content and SHA
+// to nil; the entry will be serialized with "sha": null which the API interprets
```
gmlewis:
```suggestion
// to `nil`; the entry will be serialized with `"sha": null` which the API interprets
```

github/git_trees.go
```diff
@@ -30,6 +30,13 @@ func (t Tree) String() string {
 // TreeEntry represents the contents of a tree structure. TreeEntry can
 // represent either a blob, a commit (in the case of a submodule), or another
 // tree.
+//
+// When used with [GitService.CreateTree], set Content for small text files,
+// or set SHA to reference an existing blob (use [GitService.CreateBlob] for
+// binary files or large content). To delete an entry, set both Content and SHA
```
gmlewis, line 36:
```suggestion
// or set `SHA` to reference an existing blob (use [GitService.CreateBlob] for
// binary files or large content). To delete an entry, set both `Content` and `SHA`
```

github/git_trees.go
rogpeppe: Done the second of those two.

github/git_trees.go
rogpeppe: I had a look at current precedent and it seems to me that it fits more with Go convention and existing conventions in this package not to quote field names or the Go value "nil". For example all the Get* methods mention the field in question without quoting it, and there are many examples of using "nil" in doc comments in the stdlib without quoting.

I totally agree with backquoting the JSON fragment however.

Happy to make the changes anyway if you decide otherwise.

gmlewis: > I had a look at current precedent and it seems to me that it fits more with Go convention and existing conventions in this package not to quote field names or the Go value "nil". For example all the Get* methods mention the field in question without quoting it, and there are many examples of using "nil" in doc comments in the stdlib without quoting.
> 
> I totally agree with backquoting the JSON fragment however.
> 
> Happy to make the changes anyway if you decide otherwise.

True, that's a good point. Go ahead and follow the Go convention. Thanks, @rogpeppe!

gmlewis: Following the Go convention, you would wrap `Content` and `SHA` in square brackets, right?

* `[Content]`
* `[SHA]`

rogpeppe: That wouldn't fit with the [doc links syntax](https://go.dev/doc/comment#doclinks). I guess we could use `[TreeEntry.Content]` and `[TreeEntry.SHA]` instead, but given that other doc comments in this package already refer to fields without such a qualifier, that seems a bit unnecessary to me. Happy to change to do it anyway if you think it's worthwhile.

gmlewis: OK, I'm fine with that.

To fix this properly, it would probably mean that we should make the entire repo consistent with the Go doc links syntax, which would be a huge undertaking. If you feel like filing an issue, that would be great, but let's proceed as you currently have it. Thanks, @rogpeppe!

gmlewis:
Thank you, @stevehipwell!
Merging.

PR #3879, opened 2025-12-13
feat: Add support for URL custom property value type

### LINKED ISSUE #3878: Add support for URL custom properties
GitHub has [announced](https://github.blog/changelog/2025-12-09-repository-custom-properties-graphql-api-and-url-type/#new-url-property-type) support for URL custom properties and we ought to support them.

IrvingMg:
Since `ValueType` is just a string field with no validation, I only updated the documentation comment and added a test. Let me know if I'm missing something.

Fixes #3878

### DIFF
--- a/github/enterprise_organization_properties_test.go
+++ b/github/enterprise_organization_properties_test.go
@@ -38,7 +38,7 @@ func TestEnterpriseService_GetOrganizationCustomPropertySchema(t *testing.T) {
 		Properties: []*CustomProperty{
 			{
 				PropertyName: Ptr("team"),
-				ValueType:    "string",
+				ValueType:    PropertyValueTypeString,
 				Description:  Ptr("Team name"),
 			},
 		},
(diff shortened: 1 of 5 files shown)


stevehipwell:
It might make sense to add type for `ValueType`?

IrvingMg:
> It might make sense to add type for `ValueType`?

I think it would make sense. However, as I see it, the pattern in the library is to use similar enum-like fields as plain strings with documentation comments, such as:

https://github.com/google/go-github/blob/07ddcd9c1b275afc5147c028328099596e00b958/github/repos.go#L75-L78

But if you’re okay with adding types, I could add constants like:

```go
const (
      PropertyValueTypeString      = "string"
      PropertyValueTypeSingleSelect = "single_select"
      PropertyValueTypeMultiSelect  = "multi_select"
      PropertyValueTypeTrueFalse    = "true_false"
      PropertyValueTypeURL          = "url"
  )
```
And we can decide later if we want to add similar enums for other fields across the library.

stevehipwell:
@IrvingMg this pattern is used elsewhere here.

IrvingMg:
> @IrvingMg this pattern is used elsewhere here.

I’ve added the constants. I considered using a type alias, but I believe that would  be a breaking change, and I wasn’t sure we wanted to introduce one.

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #3881, opened 2025-12-15
feat: Handle `omitzero` in `structfield` linter

Not-Dhananjay-Mishra:
Added handling for the `,omitzero` tag option in processTag (structfield.go) by stripping it from the tag value before further validation, consistent with existing `,omitempty` behavior.

### DIFF
--- a/tools/structfield/testdata/src/has-warnings/main.go
+++ b/tools/structfield/testdata/src/has-warnings/main.go
@@ -23,17 +23,22 @@ type JSONFieldType struct {
 	PointerToMap                   *map[string]string `json:"pointer_to_map,omitempty"`                      // want `change the "PointerToMap" field type to "map\[string\]string" in the struct "JSONFieldType"`
 	SliceOfInts                    []*int             `json:"slice_of_ints,omitempty"`                       // want `change the "SliceOfInts" field type to "\[\]int" in the struct "JSONFieldType"`
 
-	Count                              int        `json:"count,omitzero"`                                    // want `the "Count" field in struct "JSONFieldType" uses "omitzero" with a primitive type; remove "omitzero", as it is only allowed with structs, maps, and slices`
-	Size                               *int       `json:"size,omitzero"`                                     // want `the "Size" field in struct "JSONFieldType" uses "omitzero" with a primitive type; remove "omitzero" and use only "omitempty" for pointer primitive types`
-	PointerToSliceOfStringsZero        *[]string  `json:"pointer_to_slice_of_strings_zero,omitzero"`         // want `change the "PointerToSliceOfStringsZero" field type to "\[\]string" in the struct "JSONFieldType"`
-	PointerToSliceOfStructsZero        *[]Struct  `json:"pointer_to_slice_of_structs_zero,omitzero"`         // want `change the "PointerToSliceOfStructsZero" field type to "\[\]\*Struct" in the struct "JSONFieldType"`
-	PointerToSliceOfPointerStructsZero *[]*Struct `json:"pointer_to_slice_of_pointer_structs_zero,omitzero"` // want `change the "PointerToSliceOfPointerStructsZero" field type to "\[\]\*Struct" in the struct "JSONFieldType"`
-	PointerSliceInt                    *[]int     `json:"pointer_slice_int,omitempty"`                       // want `change the "PointerSliceInt" field type to "\[\]int" in the struct "JSONFieldType"`
-	AnyZero                            any        `json:"any_zero,omitzero"`                                 // want `the "AnyZero" field in struct "JSONFieldType" uses "omitzero"; remove "omitzero", as it is only allowed with structs, maps, and slices`
+	Count                              int                `json:"count,omitzero"`                                    // want `the "Count" field in struct "JSONFieldType" uses "omitzero" with a primitive type; remove "omitzero", as it is only allowed with structs, maps, and slices`
+	Size                               *int               `json:"size,omitzero"`                                     // want `the "Size" field in struct "JSONFieldType" uses "omitzero" with a primitive type; remove "omitzero" and use only "omitempty" for pointer primitive types`
+	PointerToSliceOfStringsZero        *[]string          `json:"pointer_to_slice_of_strings_zero,omitzero"`         // want `change the "PointerToSliceOfStringsZero" field type to "\[\]string" in the struct "JSONFieldType"`
+	PointerToSliceOfStructsZero        *[]Struct          `json:"pointer_to_slice_of_structs_zero,omitzero"`         // want `change the "PointerToSliceOfStructsZero" field type to "\[\]\*Struct" in the struct "JSONFieldType"`
+	PointerToSliceOfPointerStructsZero *[]*Struct         `json:"pointer_to_slice_of_pointer_structs_zero,omitzero"` // want `change the "PointerToSliceOfPointerStructsZero" field type to "\[\]\*Struct" in the struct "JSONFieldType"`
+	PointerSliceInt                    *[]int             `json:"pointer_slice_int,omitzero"`                        // want `change the "PointerSliceInt" field type to "\[\]int" in the struct "JSONFieldType"`
+	AnyZero                            any                `json:"any_zero,omitzero"`                                 // want `the "AnyZero" field in struct "JSONFieldType" uses "omitzero"; remove "omitzero", as it is only allowed with structs, maps, and slices`
+	StringPointerSlice                 []*string          `json:"string_pointer_slice,omitzero"`                     // want `change the "StringPointerSlice" field type to "\[\]string" in the struct "JSONFieldType"`
+	PointerToMapZero                   *map[string]string `json:"pointer__to_map_zero,omitzero"`                     // want `change the "PointerToMapZero" field type to "map\[string\]string" in the struct "JSONFieldType"`
+	SliceOfStructsZero                 []Struct           `json:"slice_of_structs_zero,omitzero"`                    // want `change the "SliceOfStructsZero" field type to "\[\]\*Struct" in the struct "JSONFieldType"`
+	StructZero                         Struct             `json:"struct_zero,omitzero"`                              // want `change the "StructZero" field type to "\*Struct" in the struct "JSONFieldType"`
 
 	AnyBoth              any     `json:"any_both,omitempty,omitzero"`                // want `the "AnyBoth" field in struct "JSONFieldType" uses "omitzero"; remove "omitzero", as it is only allowed with structs, maps, and slices`
-	NonPointerStructBoth Struct  `json:"non_pointer_struct_both,omitempty,omitzero"` // want `change the "NonPointerStructBoth" field type to "\*Struct" in the struct "JSONFieldType" because its tag uses "omitempty"`
+	NonPointerStructBoth Struct  `json:"non_pointer_struct_both,omitempty,omitzero"` // want `change the "NonPointerStructBoth" field type to "\*Struct" in the struct "JSONFieldType"`
 	PointerStringBoth    *string `json:"pointer_string_both,omitempty,omitzero"`     // want `the "PointerStringBoth" field in struct "JSONFieldType" uses "omitzero" with a primitive type; remove "omitzero" and use only "omitempty" for pointer primitive types`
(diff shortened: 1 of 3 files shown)


Not-Dhananjay-Mishra:
@gmlewis I added some test cases

**No warning -** 
- 	SliceOfZerosStructs   []*Struct         `json:"slice_of_zeros_structs,omitzero"`
- SliceOfBothOmitTags   []*Struct         `json:"slice_of_both_omit_tags,omitempty,omitzero"`

**Has warning -** 
- SliceOfZeroInts                []*int             `json:"slice_of_zero_ints,omitzero"` 
- SliceOfBothOmitTags            []*int             `json:"slice_of_both_omit_tags,omitempty,omitzero"`
- PerPageZeros  int    `url:"per_page_zeros,omitzero"`
- PerPageBoth   int    `url:"per_page_both,omitempty,omitzero"`

Not-Dhananjay-Mishra:
Just to confirm, are we going to use `omitzero` with the following conditions?

- Slice of pointer structs
- Slice of non-pointer data types (int, string)
- Pointer data types

tools/structfield/testdata/src/has-warnings/main.go
```diff
@@ -31,7 +33,9 @@ type URLFieldName struct {
...
-	Participating bool   `url:"participating,omitempty"` // want `change the "Participating" field type to "\*bool" in the struct "URLFieldType" because its tag uses "omitempty"`
+	Page          string `url:"page,omitempty"`                   // want `change the "Page" field type to "\*string" in the struct "URLFieldType" because its tag uses "omitempty"`
+	PerPage       int    `url:"per_page,omitempty"`               // want `change the "PerPage" field type to "\*int" in the struct "URLFieldType" because its tag uses "omitempty"`
+	Participating bool   `url:"participating,omitempty"`          // want `change the "Participating" field type to "\*bool" in the struct "URLFieldType" because its tag uses "omitempty"`
+	PerPageZeros  int    `url:"per_page_zeros,omitzero"`          // want `change the "PerPageZeros" field type to "\*int" in the struct "URLFieldType" because its tag uses "omitzero"`
+	PerPageBoth   int    `url:"per_page_both,omitempty,omitzero"` // want `change the "PerPageBoth" field type to "\*int" in the struct "URLFieldType" because its tag uses "omitempty and omitzero"`
```
gmlewis: OK, now here is where I'm a bit fuzzy - do we want to require using pointers for primitives when using `omitzero`? Do we want to allow `omitzero` at all for primitives?

I'm guessing the answer is "it depends on the particular field and its usage", but honestly I don't have experience with `omitzero`.

Thoughts?

gmlewis:
> Just to confirm, are we going to use `omitzero` with the following conditions?
> 
> * Slice of pointer structs
> * Slice of non-pointer data types (int, string)
> * Pointer data types

HAH!  We were typing at the same time!

Excellent questions!!!  I don't know the answers. I'd like to hear some more opinions.

tools/structfield/testdata/src/has-warnings/main.go
alexandear: `omitzero` is not valid for the `url` tag. See the documentation https://github.com/google/go-querystring/blob/59c17d91d3d41ba66bc4db00b458acc8c3220835/query/encode.go#L50

`omitzero` is only valid for the `json`. Let's fix tests and linter's code.

alexandear: I propose narrowing the scope and allowing `omitzero` only for slices, maps, and structs. These are the cases where `omitzero` reduces a bit of boilerplate.

Maybe in the future we can use `omitzero` for primitives, but I'm not sure.

Not-Dhananjay-Mishra: > I propose narrowing the scope and allowing `omitzero` only for slices, maps, and structs. These are the cases where `omitzero` reduces a bit of boilerplate.
> 
For these type of data? am i correct?

- []*Struct 
- []string/int/...
- Struct (or *Struct)
- map

Not-Dhananjay-Mishra:
Now `omitzero` is :

**Allowed with -** 
- []builtin & []*builtin
- map & *map
- Struct & *Struct
- []Struct & []*Struct

**Not Allowed with -**
- with URL Tag
- builtin & *builtin
- *[]builtin/Struct &  *[]*builtin/Struct

Not-Dhananjay-Mishra:
@alexandear could you please confirm whether the current implementation aligns with your suggestion, or if any changes are needed?

tools/structfield/testdata/src/has-warnings/main.go
```diff
@@ -22,6 +22,16 @@ type JSONFieldType struct {
 	PointerToSliceOfPointerStructs *[]*Struct         `json:"pointer_to_slice_of_pointer_structs,omitempty"` // want `change the "PointerToSliceOfPointerStructs" field type to "\[\]\*Struct" in the struct "JSONFieldType"`
 	PointerToMap                   *map[string]string `json:"pointer_to_map,omitempty"`                      // want `change the "PointerToMap" field type to "map\[string\]string" in the struct "JSONFieldType"`
 	SliceOfInts                    []*int             `json:"slice_of_ints,omitempty"`                       // want `change the "SliceOfInts" field type to "\[\]int" in the struct "JSONFieldType"`
+
+	Count                              int        `json:"count,omitzero"`                                    // want `change the "Count" field type to a slice, map, or struct in the struct "JSONFieldType" because its tag uses "omitzero"`
```
gmlewis: This message does not make sense to me.

A user is not going to change the type of a field from a primitive to a slice, map, or struct just to satisfy the "omitzero".

Instead, a user is going to REMOVE the "omitzero" when using a primitive, right?

tools/structfield/testdata/src/has-warnings/main.go
```diff
@@ -22,6 +22,16 @@ type JSONFieldType struct {
...
+	Size                               *int       `json:"size,omitzero"`                                     // want `change the "Size" field type to a slice, map, or struct in the struct "JSONFieldType" because its tag uses "omitzero"`
```
gmlewis: same here. In this case, we probably should suggest they use `omitempty` instead of `omitzero`.

tools/structfield/testdata/src/has-warnings/main.go
Not-Dhananjay-Mishra: > This message does not make sense to me.
> 
> A user is not going to change the type of a field from a primitive to a slice, map, or struct just to satisfy the "omitzero".
> 
> Instead, a user is going to REMOVE the "omitzero" when using a primitive, right?

`the Count field in struct JSONField uses "omitzero" with a primitive type; remove "omitzero"`
How is this message?

Not-Dhananjay-Mishra: or
`the Count  field in struct JSONField uses "omitzero" with a primitive type; remove "omitzero", as it is only allowed with structs, maps, and slices`

gmlewis: > or `the Count field in struct JSONField uses "omitzero" with a primitive type; remove "omitzero", as it is only allowed with structs, maps, and slices`

Yes, I like that much better, thank you, @Not-Dhananjay-Mishra!

tools/structfield/testdata/src/no-warnings/main.go
```diff
@@ -27,6 +27,20 @@ type JSONFieldType struct {
 	Exception             string            `json:"exception,omitempty"`
 	Value                 any               `json:"value,omitempty"`
 	SliceOfPointerStructs []*Struct         `json:"slice_of_pointer_structs,omitempty"`
+
+	SliceOfStrings               []string        `json:"slice_of_strings,omitzero"`
+	SliceOfPointerInts           []*int          `json:"slice_of_pointer_ints,omitzero"`
+	MapOfStringToInt             map[string]int  `json:"map_of_string_to_int,omitzero"`
+	MapOfPointerStringToInt      *map[string]int `json:"map_of_pointer_string_to_int,omitzero"`
+	StructField                  Struct          `json:"struct_field,omitzero"`
```
gmlewis: I'm having troubles with:

* `*map[...]...`
* `Struct`
* `[]Struct`

I still think these should not be allowed, but instead use:

* `map[...]...`
* `*Struct`
* `[]*Struct`

But maybe I'm just not getting the whole thing?

cc: @stevehipwell - @alexandear - @zyfy29 - please help me make sense of how we should handle `omitzero` in this repo, as I'm really struggling with this one.

Not-Dhananjay-Mishra: Should I go ahead with this :

**Allowing only** -` map[...]...` , `*Struct` , `[]*Struct` and `[]buildin` as suggested.

I tried a few small experiments with different field shapes. From what I observed, even when the data is “empty”, these values are still considered non-zero and therefore are not omitted during marshaling.

I tested with the following example.
```go
type DummyDataForOmitZero struct {
	OmitZeroWithoutPointerSlice []int  `json:"omit_zero_without_pointer_slice,omitzero"`
	OmitZeroWithPointerSlice    []*int `json:"omit_zero_with_pointer_slice,omitzero"`

	OmitZeroWithoutPointerStruct Struct  `json:"omit_zero_without_pointer_struct,omitzero"`
	OmitZeroWithPointerStruct    *Struct `json:"omit_zero_with_pointer_struct,omitzero"`

	OmitZeroWithoutPointerSliceofStruct []Struct  `json:"omit_zero_without_pointer_sliceofstruct,omitzero"`
	OmitZeroWithPointerSliceofStruct    []*Struct `json:"omit_zero_with_pointer_sliceofstruct,omitzero"`

	OmitZeroWithoutPointerMap map[string]string  `json:"omit_zero_without_pointer_map,omitzero"`
	OmitZeroWithPointerMap    *map[string]string `json:"omit_zero_with_pointer_map,omitzero"`
}
```
Empty input:
```go
		data := DummyDataForOmitZero{
			OmitZeroWithoutPointerSlice:         []int{},
			OmitZeroWithPointerSlice:            []*int{},
			OmitZeroWithoutPointerSliceofStruct: []Struct{},
			OmitZeroWithPointerSliceofStruct:    []*Struct{},
			OmitZeroWithoutPointerStruct:        Struct{},
			OmitZeroWithPointerStruct:           &Struct{},
			OmitZeroWithoutPointerMap:           map[string]string{},
			OmitZeroWithPointerMap:              &map[string]string{},
		}
```
Marshaled output:
```json
{
  "omit_zero_without_pointer_slice": [],
  "omit_zero_with_pointer_slice": [],
  "omit_zero_with_pointer_struct": {},
  "omit_zero_without_pointer_sliceofstruct": [],
  "omit_zero_with_pointer_sliceofstruct": [],
  "omit_zero_without_pointer_map": {},
  "omit_zero_with_pointer_map": {}
}
```
or should non-pointer values be allowed as well?
any opinion?

gmlewis: Awesome... thank you for the excellent experiment... that is useful to see the results.

Here's my opinion, and I am totally happy to be corrected or discuss:

I have some ideas of what should never be used in this repo (regardless the JSON tags) unless there is a REALLY GOOD reason (which, _granted_, does indeed sometimes happen):
* pointers to slices or maps (which are already reference types)
* slices of pointers to primitives
* slices of structs without the pointers (for iteration)

As for the use of `omitzero`, I don't have a good enough "feel" for them yet to really understand their use.

Looking carefully at the above experiment, only one case was removed by `omitzero`:

* OmitZeroWithoutPointerStruct Struct  `json:"omit_zero_without_pointer_struct,omitzero"`

But I'm struggling to figure out if that is a really useful case in this repo. It seems like a very targeted and specific use case... one that we might indeed find useful, but honestly I don't know where.

So in my opinion, if we still don't allow my list above, then I think we may be able to proceed.

Not-Dhananjay-Mishra: make sense.
I’ll proceed with -

**Allowing only** 
- `map[...]...`
- `*Struct`
- `[]*Struct`
- `[]builtin` (e.g. []int, []string .... )

We can adjust the linter later if necessary.

gmlewis (approved):
Thank you, @Not-Dhananjay-Mishra for this PR and for your patience while I have been trying to figure out what makes sense to me.
I'm liking this a lot.
LGTM.

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #3888, opened 2025-12-17
fix!: Add support for missing attributes in ProjectsV2 API

tmelliottjr:
BREAKING CHANGE: `AddProjectItemOptions` fields have changed types.

This PR addresses a few gaps found between the current implementation and GitHub's [ProjectsV2 OpenAPI spec](https://github.com/github/rest-api-description/tree/main/descriptions/api.github.com):

* Introduced new `ProjectV2ItemContent` and `ProjectV2ItemWithContent` structs, along with a custom `UnmarshalJSON` implementation to inspect the `content_type` returned by the API and automatically unmarshall the item's content (`Issue|PullRequest|DraftIssue`) into the correct struct.
* Added accessor methods for `ProjectV2`, `ProjectV2DraftIssue`, `ProjectV2ItemContent`, `ProjectV2ItemWithContent`, and `ProjectV2StatusUpdate` structs.
* Changed the `ContentType` field on `ProjectV2Item` from `*string` to a dedicated `*ProjectV2ItemContentType` type for improved type safety.
* Added `IsTemplate` and `LatestStatusUpdate` to the existing `ProjectV2` struct.

### DIFF
--- a/github/codespaces.go
+++ b/github/codespaces.go
@@ -167,7 +167,101 @@ type CreateCodespaceOptions struct {
 	DisplayName                *string `json:"display_name,omitempty"`
 	// RetentionPeriodMinutes represents the duration in minutes after codespace has gone idle in which it will be deleted.
 	// Must be integer minutes between 0 and 43200 (30 days).
-	RetentionPeriodMinutes *int `json:"retention_period_minutes,omitempty"`
+	RetentionPeriodMinutes *int    `json:"retention_period_minutes,omitempty"`
+	Location               *string `json:"location,omitempty"`
+}
+
+// DevContainer represents a devcontainer configuration in a repository.
+type DevContainer struct {
+	Path        string  `json:"path"`
+	Name        *string `json:"name,omitempty"`
+	DisplayName *string `json:"display_name,omitempty"`
+}
+
+// DevContainerConfigurations represents a list of devcontainer configurations in a repository.
+type DevContainerConfigurations struct {
+	Devcontainers []*DevContainer `json:"devcontainers"`
+	TotalCount    int64           `json:"total_count"`
+}
+
+// CodespaceDefaults represents default settings for a Codespace.
+type CodespaceDefaults struct {
+	Location         string  `json:"location"`
+	DevcontainerPath *string `json:"devcontainer_path,omitempty"`
+}
+
+// CodespaceDefaultAttributes represents the default attributes for codespaces created by the user with the repository.
+type CodespaceDefaultAttributes struct {
+	BillableOwner *User              `json:"billable_owner"`
+	Defaults      *CodespaceDefaults `json:"defaults"`
+}
+
+// CodespaceGetDefaultAttributesOptions represents options for getting default attributes for a codespace.
+type CodespaceGetDefaultAttributesOptions struct {
+	// Ref represents the branch or commit to check for a default devcontainer path. If not specified, the default branch will be checked.
+	Ref *string `url:"ref,omitempty"`
+	// ClientIP represents an alternative IP for default location auto-detection, such as when proxying a request.
+	ClientIP *string `url:"client_ip,omitempty"`
+}
+
+// CodespacePullRequestOptions represents options for a CodespacePullRequest.
+type CodespacePullRequestOptions struct {
+	// PullRequestNumber represents the pull request number.
+	PullRequestNumber int64 `json:"pull_request_number"`
+	// RepositoryID represents the repository ID for this codespace.
+	RepositoryID int64 `json:"repository_id"`
+}
+
+// CodespaceCreateForUserOptions represents options for creating a codespace for the authenticated user.
+type CodespaceCreateForUserOptions struct {
+	PullRequest *CodespacePullRequestOptions `json:"pull_request"`
+	// RepositoryID represents the repository ID for this codespace.
+	RepositoryID               int64   `json:"repository_id"`
+	Ref                        *string `json:"ref,omitempty"`
+	Geo                        *string `json:"geo,omitempty"`
+	ClientIP                   *string `json:"client_ip,omitempty"`
+	RetentionPeriodMinutes     *int    `json:"retention_period_minutes,omitempty"`
+	Location                   *string `json:"location,omitempty"`
+	Machine                    *string `json:"machine,omitempty"`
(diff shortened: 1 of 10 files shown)


github/projects.go
```diff
@@ -16,22 +17,65 @@ import (
...
 type ProjectsService service
 
+// ProjectV2ItemContentType represents the type of content in a ProjectV2Item.
+type ProjectV2ItemContentType string
+
+// This is the set of possible content types for a ProjectV2Item.
+const (
+	ProjectV2ItemContentTypeIssue       ProjectV2ItemContentType = "Issue"
+	ProjectV2ItemContentTypePullRequest ProjectV2ItemContentType = "PullRequest"
+	ProjectV2ItemContentTypeDraftIssue  ProjectV2ItemContentType = "DraftIssue"
```
gmlewis: Would you mind sorting these lines, please?

gmlewis (commented):
I'm wondering if we need a brand new "...WithContent" type here?
Since this PR is breaking the API anyway, can't you just do all the unmarshaling into the original "ProjectV2Item" struct?

It seems odd to me to continue to have both a "ProjectV2Item" and a "ProjectV2ItemWithContent" (and the extra maintenance burden that causes) if we can avoid it.

Thoughts?

tmelliottjr:
> It seems odd to me to continue to have both a "ProjectV2Item" and a "ProjectV2ItemWithContent" (and the extra maintenance burden that causes) if we can avoid it.

@gmlewis I had actually considered that and talked myself out of it, somehow. I'll push up some updates to consolidate these.

github/event_types.go
```diff
@@ -1148,44 +1148,6 @@ type FieldValue struct {
...
-	//   - linked_pull_requests: array of pull request objects or null
-	//   - milestone: milestone object with "id", "title", "description", etc. or null
-	//   - repository: repository object with "id", "name", "full_name", etc. or null
-	//   - reviewers: array of user objects or null
-	//   - status: object with "id", "name", "color", "description" fields (same structure as single_select) or null
-	Value any `json:"value,omitempty"`
-}
-
-// ProjectV2Item represents an item belonging to a project.
-type ProjectV2Item struct {
```
tmelliottjr, line 1174: Moved and consolidated with the newer expanded type in `projects.go`

github/event_types.go
```diff
@@ -1148,44 +1148,6 @@ type FieldValue struct {
 	To            json.RawMessage `json:"to,omitempty"`
 }
 
-// ProjectV2ItemFieldValue represents a field value of a project item.
-type ProjectV2ItemFieldValue struct {
```
tmelliottjr, line 1152: Also moved to colocate with associated ProjectsV2 types. As I understand it, this is not used in events.

github/projects.go
```diff
@@ -115,6 +154,87 @@ type ProjectV2FieldConfiguration struct {
...
+	// ProjectNodeID and ContentNodeID are used in ProjectsV2Item Webhook payloads.
+	// They may not be populated in all API responses, but are included here for completeness.
+	// See: https://docs.github.com/en/webhooks/webhook-events-and-payloads#projects_v2_item
+	ProjectNodeID *string `json:"project_node_id,omitempty"`
+	ContentNodeID *string `json:"content_node_id,omitempty"`
+}
+
+// UnmarshalJSON implements custom unmarshaling for ProjectV2Item.
+// It uses the ContentType field to determine how to unmarshal the Content field.
+func (i *ProjectV2Item) UnmarshalJSON(data []byte) error {
```
gmlewis: It is idiomatic Go to use the same letter for the receiver here as its struct name. Also, we try to reserve things like `i`,`j`,`k` for iterator indexes if possible, so let's please change this to a `p` and anywhere else you did this.


```suggestion
func (p *ProjectV2Item) UnmarshalJSON(data []byte) error {
```

github/projects.go
```diff
@@ -131,6 +251,28 @@ type ProjectV2Field struct {
 	UpdatedAt     *Timestamp                   `json:"updated_at,omitempty"`
 }
 
+// ProjectV2ItemFieldValue represents a field value of a project item.
+type ProjectV2ItemFieldValue struct {
+	ID       *int64 `json:"id,omitempty"`
+	Name     string `json:"name,omitempty"`
+	DataType string `json:"data_type,omitempty"`
```
gmlewis:
```suggestion
	Name     *string `json:"name,omitempty"`
	DataType *string `json:"data_type,omitempty"`
```

github/projects.go
```diff
@@ -330,8 +472,8 @@ type GetProjectItemOptions struct {
 // to a project. The Type must be either "Issue" or "PullRequest" (as per API docs) and
 // ID is the numerical ID of that issue or pull request.
 type AddProjectItemOptions struct {
-	Type string `json:"type,omitempty"`
-	ID   int64  `json:"id,omitempty"`
+	Type ProjectV2ItemContentType `json:"type,omitempty"`
```
gmlewis:
```suggestion
	Type *ProjectV2ItemContentType `json:"type,omitempty"`
	ID   *int64                    `json:"id,omitempty"`
```

gmlewis:
@tmelliottjr - as for the code coverage, if you could please try to add tests that cover these two lines, that would be a nice little bonus. The first one should be trivial... the second one might be trickier... and if it is too much of a pain, don't worry about it.  Thank you!

codecov-3888

gmlewis:
Thank you, @stephenotalora!
Merging.

tmelliottjr:
@gmlewis I am working on at least one other PR that may involve breaking changes. There are some updates to an existing endpoint and the addition of a new one that I'd like to add support for. Those PRs are waiting for descriptions to be published, so they're not quite ready. 

I'm curious what the release cadence looks like to better coordinate the changes?

gmlewis:
> I'm curious what the release cadence looks like to better coordinate the changes?

We are a bit ~casual~ flexible about the release cadence in this repo.

Generally, we shoot for one release per month, but sometimes we exceed that and sometimes it takes longer.
The last release was 2 weeks ago, so we are probably shooting for another release in a couple weeks (early January 2026) but again, we can be flexible if some users of this repo have pressing deadlines and prefer "official" releases to simply pulling in SHA commit hashes for the features they need.

tmelliottjr:
Wonderful! Thanks for the info.

Issue #2362, opened 2022-05-07
Codespaces


josebalius:
I am planning on implementing support for Codespaces, see: https://docs.github.com/en/rest/codespaces

Using this issue to track that work. A couple of questions:
- Do we prefer one large PR with the whole API implementation or multiple PRs?
- Do we prefer to have multiple issues?

Thanks!

gmlewis:
> * Do we prefer one large PR with the whole API implementation or multiple PRs?

I would prefer a bunch of smaller PRs if it is all the same with you.

> * Do we prefer to have multiple issues?

I'm fine with a single issue.

When we have a lot of things to work on, we often use checkboxes in a single issue like this:

- [ ] item 1
- [ ] item 2

etc... feel free to use them if you find it helpful to keep you organized.  You are welcome to use this issue or open a new one.
You can also go back and edit your first entry if you wish... totally up to you.

Thank you, @josebalius !  We appreciate your contributions!

gsmachado:
@josebalius any progress on this?

josebalius:
@gsmachado I started it but never really got into it much. Are you interested in taking over the issue or in the functionality?

gsmachado:
@josebalius I'm developing something that would need the codespaces API. If I progress with the implementation directly on the go-github repo, I will submit a PR. 😄

gsmachado:
@josebalius do you have any branch with your preliminary code?

josebalius:
I have some code locally that I haven't pushed up. Are you interested in it?

artificial-aidan:
Can you push up whatever you came up with?

artificial-aidan:
I started with some simple implementation of codespace management. I'll post the punchlist below. Happy to split it into a new issue as well.

artificial-aidan:
- [ ] Codespaces
  - [ ] List codespaces in a repository for the authenticated user
  - [ ] Create a codespace in a repository
  - [ ] List devcontainer configurations in a repository for the authenticated user
  - [ ] Get default attributes for a codespace
  - [ ] Create a codespace from a pull request
  - [ ] List codespaces for the authenticated user
  - [ ] Create a codespace for the authenticated user
  - [ ] Get a codespace for the authenticated user
  - [ ] Update a codespace for the authenticated user
  - [ ] Delete a codespace for the authenticated user
  - [ ] Export a codespace for the authenticated user
  - [ ] Get dtails about a codespace export
  - [ ] Create a repository from an unpublished codespace
[...]

PR #3890, opened 2025-12-20
feat: Add support for Codespace Machines APIs

Not-Dhananjay-Mishra:
Relates: https://github.com/google/go-github/issues/2362
This PR adds support for endpoint of [Codespace Machines](https://docs.github.com/en/enterprise-cloud@latest/rest/codespaces/machines?apiVersion=2022-11-28)

- GET /repos/{owner}/{repo}/codespaces/machines
- GET /user/codespaces/{codespace_name}/machines

### DIFF
--- a/github/codespaces_machines.go
+++ b/github/codespaces_machines.go
@@ -16,22 +16,22 @@ type CodespacesMachines struct {
 	Machines   []*CodespacesMachine `json:"machines"`
 }
 
-// ListMachinesOptions represent options for ListMachineTypesForRepository.
-type ListMachinesOptions struct {
+// ListRepoMachineTypesOptions represent options for ListMachineTypesForRepository.
+type ListRepoMachineTypesOptions struct {
 	// Ref represent the branch or commit to check for prebuild availability and devcontainer restrictions.
-	Ref *string `json:"ref,omitempty"`
+	Ref *string `url:"ref,omitempty"`
 	// Location represent the location to check for available machines. Assigned by IP if not provided.
-	Location *string `json:"location,omitempty"`
+	Location *string `url:"location,omitempty"`
 	// ClientIP represent the IP for location auto-detection when proxying a request
-	ClientIP *string `json:"client_ip,omitempty"`
+	ClientIP *string `url:"client_ip,omitempty"`
 }
 
-// ListMachineTypesForRepository lists the machine types available for a given repository based on its configuration.
+// ListRepositoryMachineTypes lists the machine types available for a given repository based on its configuration.
 //
 // GitHub API docs: https://docs.github.com/rest/codespaces/machines#list-available-machine-types-for-a-repository
 //
 //meta:operation GET /repos/{owner}/{repo}/codespaces/machines
-func (s *CodespacesService) ListMachineTypesForRepository(ctx context.Context, owner, repo string, opts *ListMachinesOptions) (*CodespacesMachines, *Response, error) {
+func (s *CodespacesService) ListRepositoryMachineTypes(ctx context.Context, owner, repo string, opts *ListRepoMachineTypesOptions) (*CodespacesMachines, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/codespaces/machines", owner, repo)
 	u, err := addOptions(u, opts)
 	if err != nil {
@@ -52,12 +52,12 @@ func (s *CodespacesService) ListMachineTypesForRepository(ctx context.Context, o
 	return machines, resp, nil
 }
 
-// ListMachineTypesForCodespace lists the machine types a codespace can transition to use.
+// ListCodespaceMachineTypes lists the machine types a codespace can transition to use.
 //
 // GitHub API docs: https://docs.github.com/rest/codespaces/machines#list-machine-types-for-a-codespace
 //
 //meta:operation GET /user/codespaces/{codespace_name}/machines
-func (s *CodespacesService) ListMachineTypesForCodespace(ctx context.Context, codespaceName string) (*CodespacesMachines, *Response, error) {
+func (s *CodespacesService) ListCodespaceMachineTypes(ctx context.Context, codespaceName string) (*CodespacesMachines, *Response, error) {
 	u := fmt.Sprintf("user/codespaces/%v/machines", codespaceName)
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
--- a/github/codespaces_machines_test.go
+++ b/github/codespaces_machines_test.go
@@ -13,13 +13,17 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
-func TestCodespacesService_ListMachineTypesForRepository(t *testing.T) {
+func TestCodespacesService_ListRepositoryMachineTypes(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
 
 	mux.HandleFunc("/repos/owner/repo/codespaces/machines", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-
+		testFormValues(t, r, values{
+			"ref":       "main",
+			"location":  "WestUs2",
+			"client_ip": "1.2.3.4",
+		})
 		fmt.Fprint(w, `{
 			"total_count": 1,
 			"machines": [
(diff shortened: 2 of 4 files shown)


github/codespaces_machines.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2025 The go-github AUTHORS. All rights reserved.
```
gmlewis, line 1: How about renaming this file to be `codespace_machines.go` (and the corresponding test file similarly) since the name of the struct is `CodespaceMachines`?

github/codespaces_machines.go
```diff
@@ -0,0 +1,89 @@
...
+
+package github
+
+import (
+	"context"
+	"fmt"
+)
+
+// Machine represent a description of the machine powering a codespace.
+type Machine struct {
```
gmlewis: Since all exported names are package-global, let's please be more specific:

```suggestion
// CodespaceMachine represent a description of the machine powering a codespace.
type CodespaceMachine struct {
```

github/codespaces_machines.go
```diff
@@ -0,0 +1,89 @@
...
+type ListMachinesOptions struct {
+	// Ref represent the branch or commit to check for prebuild availability and devcontainer restrictions.
+	Ref *string `json:"ref,omitempty"`
+	// Location represent the location to check for available machines. Assigned by IP if not provided.
+	Location *string `json:"location,omitempty"`
+	// ClientIP represent the IP for location auto-detection when proxying a request
+	ClientIP *string `json:"client_ip,omitempty"`
+}
+
+// ListMachinesTypesForRepository lists the machine types available for a given repository based on its configuration.
```
gmlewis: I think `ListMachineTypesForRepository` reads more naturally, please.

github/codespaces_machines.go
```diff
@@ -0,0 +1,89 @@
...
+	var machines *CodespaceMachines
+	resp, err := s.client.Do(ctx, req, &machines)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return machines, resp, nil
+}
+
+// ListMachinesTypesForCodespace lists the machine types a codespace can transition to use.
```
gmlewis: ditto: `ListMachineTypesForCodespace`

github/codespaces_machines.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Machine represent a description of the machine powering a codespace.
+type Machine struct {
+	Name            string `json:"name"`
+	DisplayName     string `json:"display_name"`
+	OperatingSystem string `json:"operating_system"`
+	StorageInBytes  int64  `json:"storage_in_bytes"`
+	MemoryInBytes   int64  `json:"memory_in_bytes"`
+	CPUs            int64  `json:"cpus"`
```
gmlewis: I'm thinking that this one field (`CPUs`) should be of type `int` here instead of `int64` just for convenience-of-use. Please leave the other `*InBytes` fields as `int64` as that makes total sense.

github/codespaces_machines.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+// Machine represent a description of the machine powering a codespace.
+type Machine struct {
+	Name            string `json:"name"`
+	DisplayName     string `json:"display_name"`
```
gmlewis: None of these fields need to be `omitempty`? I'm anticipating in the future that we might reuse this struct for many more API return values, at which point it is often nice to have these fields as pointers and `omitempty` because the GitHub v3 API often returns similar information for various endpoints but doesn't always fill in all possible fields for each different call.

So I'm actually thinking that maybe _all_ of these fields should be pointers with `omitempty` to be consistent with the rest of the repo.

Not-Dhananjay-Mishra:
@gmlewis Thanks for suggestions! I’ve made the changes.

github/codespaces_machines.go
Not-Dhananjay-Mishra: image

I just notice that other files also have codespace**s**. Should I rename it back to `codespaces_machines.go`?

`codespaces.go` have a struct named `Codespace`

gmlewis: WHUPS!!! MY BAD!!!  Sorry about that!

It really is "Codespaces" according to GitHub.  Please ignore my request and change it back.
Also, please change the name of the struct to `CodespacesMachines` even though that sounds weird... at least it will be consistent for searching.

Not-Dhananjay-Mishra: Done! I also found that `CodespacesMachine` was already present, so I reused it.

github/codespaces_machines.go
```diff
@@ -0,0 +1,74 @@
...
+	// ClientIP represent the IP for location auto-detection when proxying a request
+	ClientIP *string `json:"client_ip,omitempty"`
+}
+
+// ListMachineTypesForRepository lists the machine types available for a given repository based on its configuration.
+//
+// GitHub API docs: https://docs.github.com/rest/codespaces/machines#list-available-machine-types-for-a-repository
+//
+//meta:operation GET /repos/{owner}/{repo}/codespaces/machines
+func (s *CodespacesService) ListMachineTypesForRepository(ctx context.Context, owner, repo string, opts *ListMachinesOptions) (*CodespacesMachines, *Response, error) {
```
alexandear: `ListRepositoryMachineTypes` and `ListCodespaceMachineTypes` would be better names and more consistent of what we have in the repo.

E.g.,

https://github.com/google/go-github/blob/6c1bb0da6cc7af48b945cdd22863bfccfe6eea7f/github/codespaces_secrets.go#L54

github/codespaces_machines.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// CodespacesMachines represent a list of machines.
+type CodespacesMachines struct {
+	TotalCount int64                `json:"total_count"`
+	Machines   []*CodespacesMachine `json:"machines"`
+}
+
+// ListMachinesOptions represent options for ListMachineTypesForRepository.
+type ListMachinesOptions struct {
```
alexandear:
```suggestion
type ListRepoMachineTypesOptions struct {
```

github/codespaces_machines.go
```diff
@@ -0,0 +1,74 @@
...
+	// Ref represent the branch or commit to check for prebuild availability and devcontainer restrictions.
+	Ref *string `json:"ref,omitempty"`
+	// Location represent the location to check for available machines. Assigned by IP if not provided.
+	Location *string `json:"location,omitempty"`
+	// ClientIP represent the IP for location auto-detection when proxying a request
+	ClientIP *string `json:"client_ip,omitempty"`
```
alexandear:
```suggestion
	Ref *string `url:"ref,omitempty"`
	// Location represent the location to check for available machines. Assigned by IP if not provided.
	Location *string `url:"location,omitempty"`
	// ClientIP represent the IP for location auto-detection when proxying a request
	ClientIP *string `url:"client_ip,omitempty"`
```

See https://github.com/google/go-github/blob/6c1bb0da6cc7af48b945cdd22863bfccfe6eea7f/github/github.go#L256-L287

Not-Dhananjay-Mishra:
```go
type CreateCodespaceOptions struct {
	Ref *string `json:"ref,omitempty"`
	// Geo represents the geographic area for this codespace.
	// If not specified, the value is assigned by IP.
	// This property replaces location, which is being deprecated.
	// Geo can be one of: `EuropeWest`, `SoutheastAsia`, `UsEast`, `UsWest`.
	Geo                        *string `json:"geo,omitempty"`
	ClientIP                   *string `json:"client_ip,omitempty"`
	Machine                    *string `json:"machine,omitempty"`
	DevcontainerPath           *string `json:"devcontainer_path,omitempty"`
	MultiRepoPermissionsOptOut *bool   `json:"multi_repo_permissions_opt_out,omitempty"`
	WorkingDirectory           *string `json:"working_directory,omitempty"`
	IdleTimeoutMinutes         *int    `json:"idle_timeout_minutes,omitempty"`
	DisplayName                *string `json:"display_name,omitempty"`
	// RetentionPeriodMinutes represents the duration in minutes after codespace has gone idle in which it will be deleted.
	// Must be integer minutes between 0 and 43200 (30 days).
	RetentionPeriodMinutes *int    `json:"retention_period_minutes,omitempty"`
	Location               *string `json:"location,omitempty"`
}
```
Why are `url` tags not used here? This is present in `codespaces.go`.

alexandear: The `Options` prefix is mostly used for query parameters. But sometimes it represents a request body, as in `CreateCodespaceOptions` (I don't agree with this and we should refactor, but anyway).

--

You can determine if `url` or `json` is needed with this check:

```go
		testFormValues(t, r, values{
			"ref":       "main",
			"location":  "WestUs2",
			"client_ip": "1.2.3.4",
		})
```

Just add it to the `mux.HandleFunc("/repos/owner/repo/codespaces/machines", func(w http.ResponseWriter, r *http.Request) {`

See https://github.com/google/go-github/blob/6c1bb0da6cc7af48b945cdd22863bfccfe6eea7f/github/codespaces_test.go#L23-L26

github/codespaces_machines_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestCodespacesService_ListMachineTypesForRepository(t *testing.T) {
+	t.Parallel()
+	client, mux, _ := setup(t)
+
+	mux.HandleFunc("/repos/owner/repo/codespaces/machines", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+
```
alexandear: Please check query parameters:
```suggestion
		testFormValues(t, r, values{
			"ref":       "main",
			"location":  "WestUs2",
			"client_ip": "1.2.3.4",
		})

```

github/codespaces_machines.go
Not-Dhananjay-Mishra: @alexandear Thanks for suggestions! I’ve made the changes.

Issue #2821, opened 2023-06-28
GetSignaturesProtectedBranch Does Not Return ErrBranchNotProtected When it Could


cedricvanrompay-datadog:
When calling the GitHub API endpoint `https://api.github.com/repos/OWNER/REPO/branches/BRANCH/protection/required_signatures` [(documentation)](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#get-commit-signature-protection), if the branch does not have any protection rules, the GitHub API will reply with a `HTTP 404 Not Found` code and the following body:

```
{
  "message": "Branch not protected",
  "documentation_url": "https://docs.github.com/rest/branches/branch-protection#get-commit-signature-protection"
}
```

Other functions in `go-github` that may receive this kind of response usually manage it the following way:

https://github.com/google/go-github/blob/96726d8192f34c0d46e7b64b3c400962e8e64898/github/repos.go#L1347-L1349

However, `GetSignaturesProtectedBranch` does not:

https://github.com/google/go-github/blob/96726d8192f34c0d46e7b64b3c400962e8e64898/github/repos.go#L1405-L1409

I think this function should have the same behaviour, allowing callers to handle this case the same way as when calling say `GetBranchProtection`. Note that function `isBranchNotProtected` sounds like it would work without changes:

https://github.com/google/go-github/blob/96726d8192f34c0d46e7b64b3c400962e8e64898/github/repos.go#L2033-L2036

https://github.com/google/go-github/blob/96726d8192f34c0d46e7b64b3c400962e8e64898/github/repos.go#L17

I am happy to provide a Pull Request (should be trivial) for this.

gmlewis:
@cedricvanrompay-datadog - may I first say that this is one of the best issue messages I have ever read?
It is clear, to the point, and provides all the details needed with examples and a pointer to the problem.
Well done, and thank you very much!

Absolutely, I agree, and the issue is yours. Thanks again!

Not-Dhananjay-Mishra:
@gmlewis Is this issue available? I'd like to work on it.

gmlewis:
> [@gmlewis](https://github.com/gmlewis) Is this issue available? I'd like to work on it.

Yes, I think we can safely say now that this issue is yours. Thank you, @Not-Dhananjay-Mishra!

PR #3896, opened 2026-01-04
feat: GetSignaturesProtectedBranch returns `ErrBranchNotProtected` when branch is not protected

Not-Dhananjay-Mishra:
Fixes : #2821

Added this check, similar to what we already have in `GetBranchProtection`
```go
if isBranchNotProtected(err) {
	err = ErrBranchNotProtected
}
```
Now, when a branch is not protected, `GetSignaturesProtectedBranch` also returns  `ErrBranchNotProtected`

### DIFF
--- a/github/repos_test.go
+++ b/github/repos_test.go
@@ -3242,6 +3242,45 @@ func TestRepositoriesService_GetSignaturesProtectedBranch(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Thank you, @Not-Dhananjay-Mishra and @alexandear !
LGTM 
Merging.

PR #3905, opened 2026-01-08
fix: Ignore unset `AllowedMergeMethods` field

stevehipwell:
This PR fixes a bug where the merge methods for a pull request ruleset rule always need to be defined instead of being able to be unset.


github/rules.go
```diff
@@ -464,7 +464,7 @@ type RequiredDeploymentsRuleParameters struct {
 
 // PullRequestRuleParameters represents the pull_request rule parameters.
 type PullRequestRuleParameters struct {
-	AllowedMergeMethods            []PullRequestMergeMethod   `json:"allowed_merge_methods"`
+	AllowedMergeMethods            []PullRequestMergeMethod   `json:"allowed_merge_methods,omitempty"`
```
alexandear, line 467: I'm curious, can we use `omitzero` here?

stevehipwell: I'm not sure why we'd want to? This field can either be missing or it needs to have at least 1 value.

gmlewis: The unit test does show that this field is being properly omitted when empty. In hindsight, adding a second unit test demonstrating _BOTH_ behaviors would be beneficial.

gmlewis: Oh, never mind. It does indeed look like the table-driven tests also include that behavior, so I think we are all good here.

stevehipwell:
@gmlewis it'd be useful if we could get this released?

gmlewis:
@stevehipwell - this is now available in:
https://github.com/google/go-github/releases/tag/v82.0.0

stevehipwell:
Thank you very much @gmlewis !

Issue #3921, opened 2026-01-16
Bug: `AddSocialAccounts` and `DeleteSocialAccounts` of UsersService return Invalid input


Not-Dhananjay-Mishra:
According to [API docs](https://docs.github.com/en/rest/users/social-accounts?apiVersion=2022-11-28#add-social-accounts-for-the-authenticated-user) of `AddSocialAccounts` it accepts `account_urls` as body parameter. However, in the current implementation, accountURLs is passed directly.
It should instead be passed as:
```go
map[string][]string{"account_urls" : accountURLs }
```
or we can make a Struct for it (with struct it will be BREAKING CHANGE)

With this change, accountURLs will be sent in the correct format:
 `{"account_urls":["https://facebook.com/GitHub","https://www.youtube.com/@GitHub"]}`

https://github.com/google/go-github/blob/9e7e51b9b51e87a111c316124983c92a112cf474/github/users_social_accounts.go#L50-L64

same with `DeleteSocialAccounts` [API docs](https://docs.github.com/en/rest/users/social-accounts?apiVersion=2022-11-28#delete-social-accounts-for-the-authenticated-user)
https://github.com/google/go-github/blob/9e7e51b9b51e87a111c316124983c92a112cf474/github/users_social_accounts.go#L71-L79

**To reproduce the issue:**
```go
client := github.NewClient(tc)
acc, res, err := client.Users.AddSocialAccounts(ctx, []string{"https://www.example.com"})
if err != nil {
	fmt.Printf("Error adding social accounts: %v\n", err)
	return
}
fmt.Printf("Added Social Accounts: %+v\n", acc)
fmt.Printf("Response: %+v\n", res)
```
This results in the following error:
```
Error adding social accounts: POST https://api.github.com/user/social_accounts: 422 Invalid request.

Invalid input: `["https://www.example.com"]` is not of type `object`. []
```
I can provide a PR to fix it.

gmlewis:
Thank you, @Not-Dhananjay-Mishra!
It's yours.

PR #3922, opened 2026-01-17
fix: DeleteSocialAccounts and AddSocialAccounts of UsersService

Not-Dhananjay-Mishra:
Fixes: #3921 
Added an internal struct `socialAccountsRequest` and use it to send the request body instead of passing `accountURLs` directly, which previously caused an `Invalid input` error.
Since the struct is internal and the public method signature remains unchanged, this change introduces **no** breaking API changes.

### DIFF
--- a/github/users_social_accounts.go
+++ b/github/users_social_accounts.go
@@ -16,8 +16,8 @@ type SocialAccount struct {
 	URL      *string `json:"url,omitempty"`
 }
 
-// socialAccountRequest represents the request body for adding or deleting social accounts.
-type socialAccountRequest struct {
+// socialAccountsRequest represents the request body for adding or deleting social accounts.
+type socialAccountsRequest struct {
 	AccountURLs []string `json:"account_urls"`
 }
 
@@ -54,7 +54,7 @@ func (s *UsersService) ListSocialAccounts(ctx context.Context, opts *ListOptions
 //meta:operation POST /user/social_accounts
 func (s *UsersService) AddSocialAccounts(ctx context.Context, accountURLs []string) ([]*SocialAccount, *Response, error) {
 	u := "user/social_accounts"
-	req, err := s.client.NewRequest("POST", u, socialAccountRequest{AccountURLs: accountURLs})
+	req, err := s.client.NewRequest("POST", u, &socialAccountsRequest{AccountURLs: accountURLs})
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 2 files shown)


github/users_social_accounts_test.go
```diff
@@ -58,14 +57,8 @@ func TestUsersService_AddSocialAccounts(t *testing.T) {
...
 	mux.HandleFunc("/user/social_accounts", func(w http.ResponseWriter, r *http.Request) {
-		var v []string
-		assertNilError(t, json.NewDecoder(r.Body).Decode(&v))
-
 		testMethod(t, r, "POST")
-		if !cmp.Equal(v, input) {
-			t.Errorf("Request body = %+v, want %+v", v, input)
-		}
-
+		testBody(t, r, `{"account_urls":["https://twitter.com/github"]}`+"\n")
```
gmlewis: I see above that you are just using what was already here, but could you please change all of these to "example.com" for consistency?

github/users_social_accounts.go
```diff
@@ -49,7 +54,7 @@ func (s *UsersService) ListSocialAccounts(ctx context.Context, opts *ListOptions
 //meta:operation POST /user/social_accounts
 func (s *UsersService) AddSocialAccounts(ctx context.Context, accountURLs []string) ([]*SocialAccount, *Response, error) {
 	u := "user/social_accounts"
-	req, err := s.client.NewRequest("POST", u, accountURLs)
+	req, err := s.client.NewRequest("POST", u, socialAccountRequest{AccountURLs: accountURLs})
```
gmlewis: For consistency, can you please use `&socialAccountRequest{...}` here and below?
I understand that it doesn't change anything but I think it makes it actually easier to read.

Not-Dhananjay-Mishra: Thanks @gmlewis for suggestions, also should I rename that struct to socialAccount**s**Request?

github/users_social_accounts_test.go
Not-Dhananjay-Mishra: Should I change other test functions body with "example.com"?

github/users_social_accounts.go
gmlewis: > Thanks @gmlewis for suggestions, also should I rename that struct to socialAccount**s**Request?

Sure, sounds great, thanks!

github/users_social_accounts_test.go
gmlewis: > Should I change other test functions body with "example.com"?

Ah, yes please... somehow other URLs keep sneaking in and I would like to stay consistent (and never be accused of pointing to malware! :smile:

PR #3916, opened 2026-01-14
feat: Add native Go 1.23 Iterator support

merchantmoh-debug:
This PR introduces native Go 1.23 iterator support to the entire library via a new code generator.

### The Problem

Pagination in go-github currently requires verbose boilerplate:

```go
opt := &github.RepositoryListByUserOptions{ListOptions: github.ListOptions{PerPage: 10}}
for {
	repos, resp, err := client.Repositories.ListByUser(ctx, "user", opt)
	// handle err
	for _, repo := range repos { ... }
	if resp.NextPage == 0 { break }
	opt.Page = resp.NextPage
}
```

The Solution: With this PR, users can simply write:

```go
for repo, err := range client.Repositories.ListByUserIter(ctx, "user", opt) {
	if err != nil { log.Fatal(err) }
	fmt.Println(repo.GetName())
}

```

### Implementation Details

github/gen-iterators.go: A new AST-based tool (similar to gen-accessors) that scans all List* methods, identifies their pagination patterns (Page/Cursor), and generates a corresponding *Iter method returning iter.Seq2.
It handles methods using standard ListOptions embedding and explicit Page fields.
It copies the opts struct to avoid mutating the caller's options during iteration.

github/iterators.go: The generated file containing hundreds of type-safe iterators.
Metadata Handling: Updated tools/metadata to exclude the generated iterators from API operation mapping validation, as they wrap existing operations.

### Testing & Benchmarks

github/iterators_benchmark_test.go confirms that the iterator abstraction introduces negligible overhead compared to manual looping.

github/iterators_test.go verifies single-page, multi-page, and error handling scenarios.

Tests explicitly verify that the user's opts struct is NOT mutated by the iterator.

### Documentation

Added github/example_iterators_test.go to provide a clear example in the Godoc.
This modernizes the library to leverage the latest Go features, significantly improving Developer Experience (DX) by eliminating error-prone pagination boilerplate.

PR #3925, opened 2026-01-18
feat: Add iterator support via Scan and Scan2

alexandear:
Support Go iterator ranging via `Scan` and `Scan2` functions. `MustIter` and `ScanAndCollect` are convenience wrappers. This is an alternative to #3916.

Iteration is implemented only for two methods that I was able to verify manually:

- `IssuesService.ListComments` - offset-based
- `SecurityAdvisoriesService.ListRepositorySecurityAdvisoriesForOrg` - cursor (`after`)-based

One drawback of this implementation is the requirement to add the optional parameter `reqOpts ...RequestOption` to every `List*` endpoint, which is not a backward-compatible change.

Adapted from [gitlab-org/api/client-go](https://gitlab.com/gitlab-org/api/client-go/-/blob/7a08ced5cdb094baafb449335d393f8b212eff7d/pagination.go#L1).

### DIFF
--- a/example/iterpagination/main.go
+++ b/example/iterpagination/main.go
@@ -1,0 +1,60 @@
+// Copyright 2026 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// iterpagination is an example of how to use pagination with Go-native iterators.
+// It's runnable with the following command:
+//
+//	export GITHUB_AUTH_TOKEN=your_token
+//	export GITHUB_REPOSITORY_OWNER=your_owner
+//	export GITHUB_REPOSITORY_NAME=your_repo
+//	export GITHUB_REPOSITORY_ISSUE=your_issue
+//	go run .
+package main
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"strconv"
+
+	"github.com/google/go-github/v81/github"
+)
+
+func main() {
+	token := os.Getenv("GITHUB_AUTH_TOKEN")
+	if token == "" {
+		log.Fatal("Unauthorized: No token present")
+	}
+	owner := cmp.Or(os.Getenv("GITHUB_REPOSITORY_OWNER"), "google")
+	repo := cmp.Or(os.Getenv("GITHUB_REPOSITORY_NAME"), "go-github")
+	issue, _ := strconv.Atoi(os.Getenv("GITHUB_REPOSITORY_ISSUE"))
+	issue = cmp.Or(issue, 2618)
+
+	ctx := context.Background()
+	client := github.NewClient(nil).WithAuthToken(token)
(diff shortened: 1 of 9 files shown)


alexandear:
@gmlewis @merchantmoh-debug this is ready for review as a concept, as I was told [here](https://github.com/google/go-github/pull/3916#issuecomment-3755865153).

gmlewis:
I suppose that a hybrid approach might be to use Scan and Scan2 under the hood but use go:generate to make a really easy-to-use API for users of this repo?

Would there be any benefits to this approach over the other go:generate-only approach?

Not-Dhananjay-Mishra:
@alexandear @gmlewis I see Scan2 approach support `resp.After` cursor-based pagination but I couldn't find that in go:generate approach.
am i correct?

merchantmoh-debug:
> @alexandear @gmlewis I see Scan2 approach support `resp.After` cursor-based pagination but I couldn't find that in go:generate approach. am i correct?

@Not-Dhananjay-Mishra You are spot on regarding the current gap. The existing generator template primarily targeted int-based offset pagination (NextPage), but we can trivially extend it to support the cursor variations.

merchantmoh-debug:
> I suppose that a hybrid approach might be to use Scan and Scan2 under the hood but use go:generate to make a really easy-to-use API for users of this repo?
> 
> Would there be any benefits to this approach over the other go:generate-only approach?

@gmlewis I strongly believe the pure go:generate approach is superior to both the Scan runtime and a "Hybrid" approach.

On usage of "Hybrid" (Generating wrappers around Scan): If we are already running go:generate, we have the opportunity to produce "native" Go loops that call the API directly (Client -> Check Error -> Yield -> Advance). Wrapping Scan would just add a layer of runtime indirection (and potentially reflection) without any DX benefit. Generated code should be simple, readable, and "boring"—it should look exactly like the code a user would write by hand.

On Cursor Support: The Response struct actually supports three disjoint cursor models (NextPageToken, Cursor, After). A generator can statically handle this nuance by encoding the precedence logic in the template:

go
// Template logic to handle all pagination types automatically
if resp.NextPageToken != "" { opts.Page = resp.NextPageToken }
else if resp.Cursor != "" { opts.Cursor = resp.Cursor }
else if resp.After != "" { opts.After = resp.After }
I can update gen-iterators.go to support all three modes, giving us full coverage without the runtime complexity of Scan.

gmlewis:
@alexandear - @Not-Dhananjay-Mishra - @merchantmoh-debug - are we all in agreement that we prefer the approach in #3916?

If so, I believe we can close this PR and continue reviewing #3916.

@stevehipwell - @zyfy29 - do you have any comments before we proceed?

alexandear:
Yes, we can close the PR.

---

For reference, this PR is similar to https://github.com/enrichman/gh-iter that we recommend in the README: https://github.com/google/go-github#iterators-experimental

PR #3928, opened 2026-01-20
fix: Use correct type for custom property default value

### LINKED ISSUE #3926: DefaultValueStrings() returns empty slice for multi_select properties after PR #3906
I've encountered the following possibly erroneous behaviour: 

**Version:** `master` (unreleased)

**Description:**

PR #3906 introduced breaking changes. After adapting to these changes and using the new `DefaultValueStrings()` method, I've noticed that default values for custom properties of type `multi_select` always return `[], false` in my setup.

**Expected Root Cause:**

The JSON unmarshalling of `DefaultValue` (type `any`) results in `[]interface{}` instead of `[]string`, causing the type assertion at [orgs_properties.go:L67](https://github.com/google/go-github/blob/master/github/orgs_properties.go#L67) to fail.

**Expected Behavior:**

`DefaultValueStrings()` should return the default values for `multi_select` properties.

**Actual Behavior:**

Always returns `[], false`.

**Suggested Fix:**

Similar to the custom unmarshalling implemented for `CustomPropertyValue` ([orgs_properties.go:L128-L136](https://github.com/google/go-github/blob/master/github/orgs_properties.go#L128-L136)), `CustomProperty.DefaultValue` likely needs custom unmarshalling logic.

**Workaround:**

I've worked around this issue by using the "raw" value `DefaultValue` and handling both `[]string` and `[]any` types:

```go
[...]
case github.PropertyValueTypeMultiSelect:
    var v []string
    switch val := current.DefaultValue.(type) {
    case []string:
        v = val
    case []any:
        v = make([]string, len(val))
        for i, item := range val {
            str, ok := item.(string)
            if !ok {
                return nil, fmt.Errorf("failed to convert default value for custom property %s to []string: %+v", current.GetPropertyName(), current.DefaultValue)
            }
            v[i] = str
        }
    default:
        return nil, fmt.Errorf("unsupported default value type for multi_select: %T", current.DefaultValue)
    }
    [...]
    return v, nil
```

**Minimal Reproduction** (untested)

- Create Custom Property "schema" for repositories at org level (`https://github.com/organizations/<YOUR ORG>/settings/custom-properties` -> New property) of type "Multi select" with
  - some arbitrary name
  - some arbitrary options
  - "Require this property for all repositories" checked and some arbitrary default values
- Use go-github@master `GetAllCustomProperties(ctx context.Context, org string)` (https://github.com/google/go-github/blob/master/github/orgs_properties.go#L148) to fetch Custom Properties
- use `DefaultValueStrings()` on the struct representing the newly created multi select property

**Disclaimer**

I've not had the time to thoroughly investigate the issue. So the error might be caused in our implementation. :)

stevehipwell:
Fixes #3926

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,13 +42,10 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
-        with:
-          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref }}
-          fetch-depth: 0
       - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         with:
           go-version: ${{ matrix.go-version }}
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       # Get values for cache paths to be used in later steps
       - id: cache-paths


gmlewis:
cc: @tstollin

github/orgs_properties.go
```diff
@@ -64,8 +64,22 @@ func (cp CustomProperty) DefaultValueString() (string, bool) {
...
-		return s, ok
+		switch v := cp.DefaultValue.(type) {
+		case []string:
+			return v, true
+		case []any:
+			vals := make([]string, len(v))
+			for i, item := range v {
+				s, ok := item.(string)
+				if !ok {
+					return nil, false
```
gmlewis, line 75: @stevehipwell - if you want to (optionally) improve the code coverage score on this PR, you could create a JavaScript aray that mixed numbers into the strings, e.g. `["a","b", 3]` and pass that in.

stevehipwell: I had a similar idea but kept the values to a single non-string type.

stevehipwell: Interestingly the code coverage doesn't look like it re-ran?

gmlewis: I'm seeing no test cases with arrays of non-string types. Did you push the changes?

stevehipwell: It's in there, but it looks like there might be an issue with the checkout when there is more than 1 commit.

.github/workflows/tests.yml
```diff
@@ -42,10 +42,12 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```
gmlewis: Please revert, as this change is unrelated.

stevehipwell: Before I moved it you were getting annotations saying that caching had failed and this looks to have fixed it. Is there a specific reason for this order?

github/orgs_properties.go
stevehipwell: Codecov is still wrong, so trying to set the ref again.

.github/workflows/tests.yml
alexandear: @stevehipwell Actually, this should fix caching. But let's move this to a separate PR.

gmlewis: Oh, I thought this was accidental. 
I don't remember mentioning caching. Can you please provide a link that I can go read about the changes to checkout?
What should I be seeing differently in terms of PR behavior with this change?
Sorry, I'm lost... please clue me in.

stevehipwell: Is there a reason why we're not using the native setup go caching?

gmlewis: > Is there a reason why we're not using the native setup go caching?

Not particularly, but all these workflow changes should really be in a separate PR, please.

Also, you can test locally both unit tests and coverage so we don't have to keep going back and forth on this, as tests are failing again.

github/orgs_properties.go
stevehipwell: OK, I've found the culprit and reverted the changes here for a new PR.

.github/workflows/tests.yml
stevehipwell: @gmlewis I'm testing GitHub Actions, but I'll move it to a seperate PR now I know what the issue is.

alexandear: @gmlewis Documentation https://github.com/actions/setup-go#caching says that the `checkout` must be before `setup-go` for automatic caching

stevehipwell: See #3929.

stevehipwell:
This should be ready to be merged.

gmlewis (approved):
Thank you, @stevehipwell!
LGTM.

stevehipwell:
@alexandear any chance of a review?

gmlewis:
Thank you, @alexandear!
Merging.

PR #3929, opened 2026-01-20
chore: Refactor test workflow

stevehipwell:
This PR refactors the test workflow to use the native setup go caching as we use in the linting workflow. It also moves checkout before setup go, which is idiomatic and also matches the linting workflow.

These changes should stop the cache warning annotations that testing has been generating, although this could also be achieved by disabling caching for the setup go action.


gmlewis (approved):
Thank you, @stevehipwell!
LGTM.
Awaiting second LGTM+Approval from any other contibutor to this repo before merging.

PR #3935, opened 2026-01-22
chore: Enable `govet` linter with all checks

alexandear:
This PR enables all `govet` checks except `fieldalignment` and `shadow`.

`fieldalignment` suggests rearranging fields to reduce struct size in memory, while `shadow` generates many warnings about shadowing local `err` variables.


gmlewis (approved):
Thank you, @alexandear!
LGTM.
Merging.

PR #3948, opened 2026-01-29
chore: Use `example.com` instead of random URLs in tests

Not-Dhananjay-Mishra:
For reference : [comment](https://github.com/google/go-github/pull/3922#discussion_r2701170434)


Not-Dhananjay-Mishra:
@gmlewis Should I update them too?

This test contain some download URLs
https://github.com/google/go-github/blob/1c429334f46c5bb5382fd148a6bb2b227b31762c/github/actions_runners_test.go#L33-L38

The URL present in this test
https://github.com/google/go-github/blob/1c429334f46c5bb5382fd148a6bb2b227b31762c/github/admin_users_test.go#L124-L128

The URLs present in this test
https://github.com/google/go-github/blob/1c429334f46c5bb5382fd148a6bb2b227b31762c/github/meta_test.go#L28-L50

gmlewis:
@Not-Dhananjay-Mishra - I think that `https://github.com/actions/runner` can be left alone since it is a real repo on GitHub relevant to action runner.

IP addresses (since there is no particular reason to use IP addresses in this case that I can tell) can be changed to example.com and so can git.company.com.

Thanks.

Not-Dhananjay-Mishra:
This PR is ready for review 😸.

gmlewis (approved):
LGTM.

gmlewis:
Thank you, @alexandear!
Merging.

PR #3938, opened 2026-01-24
 feat(otel): Add native OpenTelemetry Transport module

merchantmoh-debug:
Adds a new optional module `github.com/google/go-github/v81/otel` which provides an `http.RoundTripper` instrumented with OpenTelemetry.

### Features
*   **Automatic Tracing:** Every API call gets a semantic span (e.g., `github/GET`).
*   **Rate Limit Observability:** Parses GitHub headers (`X-RateLimit-*`) and records them as Span Attributes (`github.rate_limit.remaining`, `github.rate_limit.reset`).
*   **Support Tracing:** Records `X-Github-Request-Id` for easy debugging with GitHub Support.

### Verification
Validated with `otel/example` app against the live API.
Dependencies are isolated in a nested module to avoid bloating core `go-github`.

PR #3949, opened 2026-01-29
chore: Format code with `golangci-lint fmt`

alexandear:
- Change `script/fmt.sh` to use the `golangci-lint fmt` command. Internally, it uses `gofmt`, `gci`, `gofumpt`, `goimports` simultaneously.
- Set `gci.sections`, `gofumpt.extra-rules` in golangci-lint config and reformat examples.

Inspired by discussion comments in #3938.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -5,7 +5,7 @@
 
 //go:generate go run gen-accessors.go
 //go:generate go run gen-stringify-test.go
-//go:generate sh ../script/metadata.sh update-go
+//go:generate ../script/metadata.sh update-go
 
 package github
 


merchantmoh-debug:
@alexandear - Genius. Very good idea.

merchantmoh-debug:
@gmlewis  - I can answer that.

Yes, the tool absolutely supports the "Old School" grouping you prefer (Single sorted bunch for all non-standard imports).

Currently, the config defines 3 sections, which forces the split you're seeing:

standard
default (Third-party)
prefix(github.com/google/go-github/) (Internal)
To restore the single "External+Internal" bunch (preserving just the 2-block layout: Standard + Everything Else), we just need to remove the prefix section.

The Config Fix (.golangci.yml):

linters-settings:
  gci:
    sections:
      - standard # Section 1: Standard Lib
      - default  # Section 2: Everything else (External + Internal sorted together)
      # - prefix(github.com/google/go-github/) <--- Removing this merges it into 'default'
This configuration tells the linter to treating everything non-standard as one group, eliminating the "visual search" through multiple bunches.

alexandear:
@gmlewis reverted. By default, `gci` groups imports into two sections: `std` and `default`.

PR #3951, opened 2026-01-30
fix: Handle HTTP 429 status code for rate limiting

abtris:
GitHub API can return either 403 or 429 for rate limiting, but the library was only checking for 403. This caused 429 responses to be treated as generic ErrorResponse instead of RateLimitError or AbuseRateLimitError, breaking retry logic and rate limit handling.

Changes:
- CheckResponse now detects 429 with X-RateLimit-Remaining: 0 as RateLimitError (primary rate limit)
- CheckResponse now detects 429 with secondary rate limit documentation_url as AbuseRateLimitError

Reference: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api


gmlewis:
Merging.

Issue #3954, opened 2026-02-01
Latest release notes are not properly formatted


Not-Dhananjay-Mishra:
<img width="1233" height="467" alt="Image" src="https://github.com/user-attachments/assets/4442998a-f4a8-4e8f-a358-0c74ca841ad1" />

**It Should be like this** 
- fix!: Simplify `Git.ListMatchingRefs` by removing `ReferenceListOptions` ([#3924](https://github.com/google/go-github/pull/3924))
BREAKING CHANGE: `Git.ListMatchingRefs` accepts ref instead of the `ReferenceListOptions`.
- refactor!: Use a struct for `Repository.Permissions` instead of `map[string]bool` ([#3936](https://github.com/google/go-github/pull/3936))
BREAKING CHANGE: `Repository.Permissions` now typed as a struct

In my option this issue occurs due to "gen-release-notes first reads the web page" because of that we can't get the full title of commits with long titles.

Why can't we use this [awesome lib](https://github.com/google/go-github) to get all the PRs merged after latest release instead of HTML scraping?

gmlewis:
First off, I apologize for the mangled release notes... that is completely my fault. I've manually fixed them.
The web scraper is also my fault and your suggestion is a good one.
I'll look into this, as it is my responsibility.

gmlewis:
The interesting side effect of changing the tool now is that anyone can run `gen-release-notes` locally and find out _ALL_ breaking and non-breaking API changes since a particular release (using `--tag v76.0.0` for example).
This might be nice if multiple releases were skipped since the last update, which is super-easy to do especially in a work environment.

Not-Dhananjay-Mishra:
> The interesting side effect of changing the tool now is that anyone can run `gen-release-notes` locally and find out _ALL_ breaking and non-breaking API changes since a particular release (using `--tag v76.0.0` for example). This might be nice if multiple releases were skipped since the last update, which is super-easy to do especially in a work environment.

That is really cool!
How about mentioning that in README.md

PR #3958, opened 2026-02-01
fix(tools): Change `gen-release-notes` to use `git` instead of scraping web

gmlewis:
Fixes: #3954.

cc: @Not-Dhananjay-Mishra


Not-Dhananjay-Mishra (approved):
Thank you @gmlewis. 
LGTM

PR #3963, opened 2026-02-03
refactor!: Use `RepositoryPermissions` struct for `User.Permissions`

### LINKED ISSUE #3947: Modeling `User.Permissions` as struct instead of `map[string]bool`
Similar to #3934 
According to the [List Collaborators Docs](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators), the permissions object has a fixed, well-defined structure:

```json
{
  "permissions": {
    "type": "object",
    "properties": {
      "pull": {
        "type": "boolean"
      },
      "triage": {
        "type": "boolean"
      },
      "push": {
        "type": "boolean"
      },
      "maintain": {
        "type": "boolean"
      },
      "admin": {
        "type": "boolean"
      }
    }
  }
}
```
Currently it is typed as `map[string]bool` (It was added in 2015 so I guess it is safe to update):
https://github.com/google/go-github/blob/1c429334f46c5bb5382fd148a6bb2b227b31762c/github/users.go#L69-L73

we can use `RepositoryPermissions` struct introduced in #3936

niveshdandyan:
BREAKING CHANGE: `User.Permissions` is now `*RepositoryPermissions` instead of `map[string]bool`.

## Summary

Replace `map[string]bool` with `*RepositoryPermissions` struct for the `User.Permissions` field.

Fixes #3947

## Changes

- Changed `User.Permissions` from `map[string]bool` to `*RepositoryPermissions`
- Updated related tests to use the struct format
- Regenerated accessor functions

## Rationale

According to the [List Collaborators API docs](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators), the permissions object has a fixed, well-defined structure with `pull`, `triage`, `push`, `maintain`, and `admin` boolean fields.

Using the existing `RepositoryPermissions` struct provides:
- Better type safety
- Consistency with `Repository.Permissions`
- IDE autocompletion for permission fields

## Breaking Change

This is a breaking API change as noted in the issue labels. Users who access `User.Permissions` will need to update their code from map access (`u.Permissions["pull"]`) to struct access (`u.GetPermissions().GetPull()`).

## Testing

All existing tests pass with the updated types.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -30622,10 +30622,10 @@ func (u *User) GetOwnedPrivateRepos() int64 {
 	return *u.OwnedPrivateRepos
 }
 
-// GetPermissions returns the Permissions map if it's non-nil, an empty map otherwise.
-func (u *User) GetPermissions() map[string]bool {
-	if u == nil || u.Permissions == nil {
-		return map[string]bool{}
+// GetPermissions returns the Permissions field.
+func (u *User) GetPermissions() *RepositoryPermissions {
+	if u == nil {
+		return nil
 	}
 	return u.Permissions
 }
--- a/github/github-accessors_test.go
+++ b/github/github-accessors_test.go
@@ -39496,10 +39496,7 @@ func TestUser_GetOwnedPrivateRepos(tt *testing.T) {
 
 func TestUser_GetPermissions(tt *testing.T) {
 	tt.Parallel()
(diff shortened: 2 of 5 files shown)


gmlewis:
Merging.

Issue #3964, opened 2026-02-03
Support native cursor-based pagination using Go 1.23 iterators


gmlewis:
Relates to: #3962, #3916, #3209, #1935, #2618, and probably others.

Native Go 1.23 iterator support was added in #3916 and we decided to only handle integer-based `NextPage` pagination in that PR.
Now we wish to add support for cursor-based pagination.

I will work on this one myself.

Not-Dhananjay-Mishra:
@gmlewis I tried running this
```go
c := 0
for e, err := range client.SecurityAdvisories.ListGlobalSecurityAdvisoriesIter(context.Background(), nil) {
	if err != nil {
		log.Fatal(err)
	}
	_ = e
	c++
}
fmt.Println("Total Global Security Advisories found ", c)
```
output - 
```
notdm@lappy:~/golang/iter$ go run main.go
Total Global Security Advisories found  30
```

Iterator is not working... because `resp.Cursor` returned by GitHub is empty. 

I believe we should replace this 
```go
if resp.Cursor == "" {
	break
}
opts.ListCursorOptions.Cursor = resp.Cursor
```
with this
```go
if resp.After == "" {
	break
}
opts.ListCursorOptions.After = resp.After
```

gmlewis:
That's weird! I need to read up on the latest official GitHub v3 API pagination docs and see what I can find.
We originally documented how cursor-based pagination should work in this repo... but maybe it has changed since then, or maybe I didn't follow the directions carefully?!?

gmlewis:
Surprisingly, this page:
https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28
no longer mentions the word "cursor" at all!!!

gmlewis:
From this example: https://github.com/octokit/octokit.js/#readme
it almost looks like we should be using:

```go
if resp.Cursor == "" {
	break
}
opts.ListCursorOptions.After = resp.Cursor
```

I'll try printing out the full response with your example and see what happens.

gmlewis:
Just as you said, @Not-Dhananjay-Mishra - the response does not even contain a "cursor"!!!

Here's what I get in the header:

```
link: <https://api.github.com/advisories?after=Y3Vyc29yOnYyOpK0MjAyNi0wMi0xMFQwMDoyNToyNFrOAAUh8g%3D%3D>; rel="next"
```

I'll send a fix PR to address this.  Thank you!

PR #3965, opened 2026-02-03
feat: Add native Go 1.23 iterator support for cursor-based pagination

gmlewis:
Fixes: #3964.

cc: @alexandear

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,7 @@ Users who have worked with protocol buffers should find this pattern familiar.
 ### Pagination ###
 
 All requests for resource collections (repos, pull requests, issues, etc.)
-support pagination. Pagination options are described in the
+support pagination. Pagination options using page numbers are described in the
 `github.ListOptions` struct and passed to the list methods directly or as an
 embedded type of a more specific list options struct (for example
 `github.PullRequestListOptions`). Pages information is available via the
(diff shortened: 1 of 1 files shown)


gmlewis:
@Not-Dhananjay-Mishra - feel free to review if you are currently online. I saw your emojis above. 😂

github/github-iterators.go
```diff
@@ -837,10 +814,10 @@ func (s *DependabotService) ListOrgAlertsIter(ctx context.Context, org string, o
 				}
 			}
 
-			if resp.NextPage == 0 {
+			if resp.Cursor == "" && resp.NextPage == 0 {
```
Not-Dhananjay-Mishra, line 817:
```go
switch {
    case resp.NextPage != 0:
		opts.ListOptions.Page = resp.NextPage
	case resp.Cursor != "":
		opts.ListCursorOptions.Cursor = resp.Cursor
	default:
		return
	}
```
Can we use this approach? I found a similar pattern in #3925 Scan2
Btw I am about to go to sleep 🙃.

gmlewis: I thought about that, but I think my mechanism here is simpler with fewer lines of code with the same behavior, so I think I would rather leave my current implementation.

github/github-iterators_test.go
```diff
@@ -1973,7 +1973,7 @@ func TestDependabotService_ListOrgAlertsIter(t *testing.T) {
 		callNum++
 		switch callNum {
 		case 1:
-			w.Header().Set("Link", `<https://api.github.com/?page=1>; rel="next"`)
+			w.Header().Set("Link", `<https://api.github.com/?cursor=yo>; rel="next"`)
 			fmt.Fprint(w, `[{},{},{}]`) // Call 1 below: return 3 items, NextPage=1, no errors
```
Not-Dhananjay-Mishra: @gmlewis is generated comment worng? Case 1 no longer returns page=1 but comment mentioned Next page=1

github/github-iterators.go
Not-Dhananjay-Mishra: Make sense 👍

github/github-iterators_test.go
gmlewis: Ah, that is an excellent point. Do you think the comments are helpful, or should I just remove all of them?

Not-Dhananjay-Mishra: In my opinion, since this is generated code and it’s test function, we can remove the comments.

gmlewis: OK, sounds good. I'll remove them.

alexandear (commented):
Should we also extend README and example_iterators_test.go?

gmlewis:
> Should we also extend README and example_iterators_test.go?

I just updated the README.md and was starting to update `example_iterators_test.go` but the cursor-based pagination example looks absolutely identical to the existing example, so I decided that it was unnecessary. Thoughts?

alexandear:
> > Should we also extend README and example_iterators_test.go?
> 
> I just updated the README.md and was starting to update `example_iterators_test.go` but the cursor-based pagination example looks absolutely identical to the existing example, so I decided that it was unnecessary. Thoughts?

Updating README is enough, thanks.

gmlewis:
Thank you, @alexandear!
Merging.

PR #3969, opened 2026-02-05
chore: Address `otel` review feedback

alexandear:
- Use `example/go.mod` instead of `example/otel/go.mod` for consistency with other examples.
- Upgrade to the latest `go.opentelemetry.io` packages.
- Add dependabot config for `otel` directory.
- Change to raw strings instead of constants for consistency in tests.

Follows #3938


otel/transport_test.go
```diff
@@ -59,9 +58,9 @@ func TestRoundTrip_Spans(t *testing.T) {
 	headers := http.Header{}
 	headers.Set("X-Ratelimit-Limit", "5000")
 	headers.Set("X-Ratelimit-Remaining", "4999")
-	headers.Set(github.HeaderRateReset, "1372700873")
+	headers.Set("X-Ratelimit-Reset", "1372700873")
```
gmlewis, line 61: Can you please remind me why we bothered to export the headers if we aren't going to be using them?

alexandear: We exported the headers not for tests, but for use in the `otel` module.

In tests, I intentionally avoid using constants. For example, if someone changes a constant string, the test will catch it. If we use constants everywhere, including in tests, such changes would go unnoticed.

That said, I do not mind using constants in tests. It is up to the reviewer to decide.

gmlewis: OK, cool.  Sounds good to me... thanks!

PR #3972, opened 2026-02-05
fix: Address `modernize.omitzero` issues

alexandear:
This PR enables the previously disabled `modernize.omitzero` check.

## Fixed issues

1. `markReadOptions.LastReadAt`:

We can safely change `omitzero` because it's an unexported struct. Now we can pass an empty `Timestamp{}` to `MarkNotificationsRead` and `MarkRepositoryNotificationsRead`, which means  marking all notifications as read.

2. `RepositoryContentResponse.Commit`:

It's a required field according to https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#create-or-update-file-contents.


github/actions_hosted_runners.go
```diff
@@ -93,7 +93,7 @@ type HostedRunnerImage struct {
 // HostedRunnerRequest specifies body parameters to Hosted Runner configuration.
 type HostedRunnerRequest struct {
 	Name           string            `json:"name,omitempty"`
-	Image          HostedRunnerImage `json:"image,omitempty"`
+	Image          HostedRunnerImage `json:"image"`
 	RunnerGroupID  int64             `json:"runner_group_id,omitempty"`
 	Size           string            `json:"size,omitempty"`
 	MaximumRunners int64             `json:"maximum_runners,omitempty"`
```
Not-Dhananjay-Mishra: `name`, `size` and `runner_group_id` are also required should we remove omitempty from them too?

stevehipwell:
If you're updating the code for omitzero, the environment created update struct had annotations removed before omitzero was available. The omitzero annotation could now be added to reviewers and one of the other fields (can't remember the name off the top of my head).

github/actions_hosted_runners.go
alexandear: Done in #3973

Not-Dhananjay-Mishra (approved):
LGTM

github/actions_hosted_runners.go
Not-Dhananjay-Mishra: @alexandear should we undo these changes here? 
I guess otherwise there will be merge conflicts in #3973.

alexandear: @Not-Dhananjay-Mishra we can't undo this because the linter will fail. I'll resolve conflicts if needed. It's not a big deal.

gmlewis:
Thank you, @alexandear, @stevehipwell, and @Not-Dhananjay-Mishra!
Merging.

PR #3973, opened 2026-02-05
fix!: `CreateHostedRunnerRequest`, `UpdateHostedRunnerRequest` instead of `HostedRunnerRequest`

alexandear:
BREAKING CHANGE: `ActionsService.CreateHostedRunner` and `EnterpriseService.CreateHostedRunner` now accept `CreateHostedRunnerRequest`; `ActionsService.UpdateHostedRunner` and `EnterpriseService.UpdateHostedRunner` now accept `UpdateHostedRunnerRequest`.

---

While working on #3972, I found that `ActionsService.UpdateHostedRunner` accepts `HostedRunnerRequest` as a body. But it's wrong according to the docs for [actions](https://docs.github.com/en/rest/actions/hosted-runners?apiVersion=2022-11-28#update-a-github-hosted-runner-for-an-organization) and [enterprise](https://docs.github.com/en/enterprise-cloud@latest/rest/actions/hosted-runners?apiVersion=2022-11-28#update-a-github-hosted-runner-for-an-enterprise).

Current:

```go
type HostedRunnerRequest struct {
	Name           string            `json:"name,omitempty"`
	Image          HostedRunnerImage `json:"image,omitempty"`
	RunnerGroupID  int64             `json:"runner_group_id,omitempty"`
	Size           string            `json:"size,omitempty"`
	MaximumRunners int64             `json:"maximum_runners,omitempty"`
	EnableStaticIP bool              `json:"enable_static_ip,omitempty"`
	ImageVersion   string            `json:"image_version,omitempty"`
}
```

Should be `image_id` (optional) instead of the `Image`. So, I created a separate struct `UpdateHostedRunnerRequest` with an optional `image_id` and the rest of the fields (all optional).

### DIFF
--- a/github/actions_hosted_runners.go
+++ b/github/actions_hosted_runners.go
@@ -89,7 +89,7 @@ type HostedRunnerImage struct {
 	// The source of the runner image. Can be one of: github, partner, custom.
 	Source string `json:"source"`
 	// The version of the runner image to deploy. This is relevant only for runners using custom images.
-	Version *string `json:"version"`
+	Version *string `json:"version,omitempty"`
 }
 
 // CreateHostedRunnerRequest specifies body parameters to create Hosted Runner configuration.


github/actions_hosted_runners_test.go
```diff
@@ -776,47 +776,6 @@ func TestActionsService_UpdateHostedRunner(t *testing.T) {
...
-					ID:      "ubuntu-latest",
-					Source:  "github",
-					Version: "latest",
-				},
-			},
-			expectedError: "validation failed: image struct should not be set directly; use the ImageVersion to specify version details",
-		},
-	}
-	for _, tt := range testCases {
-		_, _, err := client.Enterprise.UpdateHostedRunner(ctx, "o", 23, tt.request)
```
alexandear, line 814: This checks `Enterprise.UpdateHostedRunner`, but the test is `TestActionsService_UpdateHostedRunner`

github/enterprise_actions_hosted_runners_test.go
```diff
@@ -770,47 +770,6 @@ func TestEnterpriseService_UpdateHostedRunner(t *testing.T) {
...
-					ID:      "ubuntu-latest",
-					Source:  "github",
-					Version: "latest",
-				},
-			},
-			expectedError: "validation failed: image struct should not be set directly; use the ImageVersion to specify version details",
-		},
-	}
-	for _, tt := range testCases {
-		_, _, err := client.Enterprise.UpdateHostedRunner(ctx, "o", 23, tt.request)
```
alexandear, line 808: We don't need validation. All fields are optional.

alexandear:
We can't reuse structs in this case. They are different. And yes, splitting this struct into two separate use cases:
 1) fixes Update methods.
 2) reduces end-user confusion (optional-required fields).

Here are comparison with Copilot:

Field | Create | Update | Required in Create | Required in Update
-- | -- | -- | -- | --
name | ✓ | ✓ | Required | Optional
image (object with id and source) | ✓ | - | Required | -
size | ✓ | ✓ | Required | Optional
runner_group_id | ✓ | ✓ | Required | Optional
maximum_runners | ✓ | ✓ | Optional | Optional
enable_static_ip | ✓ | ✓ | Optional | Optional
image_gen | ✓ | - | Optional (default: false) | -
image_id | - | ✓ | - | Optional
image_version | - | ✓ | - | Optional (can be null)

<details><summary>Raw doc screenshots</summary>
<p>

Create body:

image
image

Update body:

image

</p>
</details>

github/actions_hosted_runners.go
```diff
@@ -101,6 +100,17 @@ type HostedRunnerRequest struct {
 	ImageVersion   string            `json:"image_version,omitempty"`
```
gmlewis: So can `image_version` now be removed from `HostedRunnerRequest`?

gmlewis: And since this is a breaking change, how about completely cleaning up `HostRunnerRequest` by removing `omitempty` from all the required fields and adding pointers to the optional fields... also to be more consistent with the rest of the repo?

alexandear: Removed `image_version`; renamed `HostedRunnerRequest` to `CreatedHostedRunnerRequest`; fixed other fields. Please recheck.

github/actions_hosted_runners.go
```diff
@@ -83,48 +83,55 @@ func (s *ActionsService) ListHostedRunners(ctx context.Context, org string, opts
...
 type HostedRunnerImage struct {
-	ID      string `json:"id"`
-	Source  string `json:"source"`
-	Version string `json:"version"`
+	// The unique identifier of the runner image.
+	ID string `json:"id"`
+	// The source of the runner image. Can be one of: github, partner, custom.
+	Source string `json:"source"`
+	// The version of the runner image to deploy. This is relevant only for runners using custom images.
+	Version *string `json:"version"`
```
gmlewis: Should this be `omitempty`?

gmlewis (approved):
Thank you, @alexandear!
One nit, otherwise LGTM.

github/actions_hosted_runners.go
alexandear: Yes.

This should be caught by the linter `structfield`.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

Issue #3976, opened 2026-02-08
Iterator generation issues caused by incorrect or unsupported pagination options


Not-Dhananjay-Mishra:
While trying new `Iter` when i run `client.Users.ListAllIter` it went into infinite loop, I found out that `client.Users.ListAll` doesn't support Page based  pagination. so i decide to find if there exist more functions with similar or some sort of problems.

**Here are my findings -**

---

1. `UsersService.ListAll` don't have `page` acc to [docs](https://docs.github.com/rest/users/users#list-users) if you run ListAllIter it will go in infinite loop since its pagination is powered exclusively by the since parameter. 
https://github.com/google/go-github/blob/c8fd3f74bcd37883274cf9d3eae49d3fe3e7f012/github/users.go#L224-L227
Solution : Replacing `UserListOptions.ListOptions` with `UserListOptions.PerPage` which will also result in removal of  ListAllIter

---

2. `PackageGetAllVersions` handles two endpoints one of them doesn't support pagination acc to [docs](https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-a-user) also for some reason it's Iter dosen't generated maybe because it dose not contain `List` as prefix. 
https://github.com/google/go-github/blob/c8fd3f74bcd37883274cf9d3eae49d3fe3e7f012/github/users_packages.go#L130-L139
Solution : Dividing PackageGetAllVersions into two separate functions, if Iter is required then add `List` to prefix of new function

---

3. `SubIssueService.ListByIssue` dosen't support cursor based pagination acc to [docs](https://docs.github.com/en/rest/issues/sub-issues?apiVersion=2022-11-28#list-sub-issues)
https://github.com/google/go-github/blob/c8fd3f74bcd37883274cf9d3eae49d3fe3e7f012/github/sub_issue.go#L71-L74
Solution : Remove `IssueListOptions.ListCursorOptions`

---

4. `RepositoriesService.ListAllTopics` also support pagination acc to [docs](https://docs.github.com/rest/repos/repos#get-all-repository-topics).
https://github.com/google/go-github/blob/c8fd3f74bcd37883274cf9d3eae49d3fe3e7f012/github/repos.go#L1928-L1931
Solution : Add ListOptions

---

5. `RepositoriesService.ListCustomDeploymentRuleIntegrations` also support pagination acc to [docs](https://docs.github.com/rest/deployments/protection-rules#list-custom-deployment-rule-integrations-available-for-an-environment).
[...]

PR #3977, opened 2026-02-08
fix!: Replace `UserListOptions.ListOptions` with `UserListOptions.PerPage`

Not-Dhananjay-Mishra:
BREAKING CHANGE: Replaced `UserListOptions.ListOptions` with `UserListOptions.PerPage` which will also result in removal of `UsersService.ListAllIter`

UsersService.ListAll don't have `page` acc to [docs](https://docs.github.com/rest/users/users#list-users) if you run ListAllIter it will go in infinite loop since its pagination is powered exclusively by the since parameter.

Related: #3976

### DIFF
--- a/github/users.go
+++ b/github/users.go
@@ -208,11 +208,9 @@ func (s *UsersService) GetHovercard(ctx context.Context, user string, opts *Hove
 // UserListOptions specifies optional parameters to the UsersService.ListAll
 // method.
 type UserListOptions struct {
-	// ID of the last user seen
-	Since *int64 `url:"since,omitempty"`
-
-	// Note: Pagination is powered exclusively by the Since parameter,
-	PerPage *int `url:"per_page,omitempty"`
+	// A user ID. Only return users with an ID greater than this ID.
+	Since   *int64 `url:"since,omitempty"`
+	PerPage *int   `url:"per_page,omitempty"`
 }
 
 // ListAll lists all GitHub users.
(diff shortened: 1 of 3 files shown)


github/users.go
```diff
@@ -212,9 +212,7 @@ type UserListOptions struct {
 	Since int64 `url:"since,omitempty"`
```
Not-Dhananjay-Mishra:
```suggestion
	Since *int64 `url:"since,omitempty"`
```
Should I make it pointer to be consistent with rest of repo?

alexandear: Yes, let's change to `*int64`.

github/users.go
alexandear: Maybe we can improve the comment:
```suggestion
	// A user ID. Only return users with an ID greater than this ID.
```

github/users.go
```diff
@@ -212,9 +212,7 @@ type UserListOptions struct {
...
 
 	// Note: Pagination is powered exclusively by the Since parameter,
```
alexandear: This comment is no longer needed and should be removed.

github/users.go
```diff
@@ -209,12 +209,10 @@ func (s *UsersService) GetHovercard(ctx context.Context, user string, opts *Hove
 // method.
 type UserListOptions struct {
```
alexandear, line 210: And please remove this line

https://github.com/google/go-github/blob/c8fd3f74bcd37883274cf9d3eae49d3fe3e7f012/.golangci.yml#L420

alexandear:
Updates #3976

PR #3981, opened 2026-02-09
fix!: Add `ListLicensesOptions` to `LicensesService.List`

alexandear:
BREAKING CHANGE: `LicensesService.List` now accepts `ListLicensesOptions` parameter for pagination.

---

1. Add missing `ListLicensesOptions` to `LicensesService.List`, see the [query parameters](https://docs.github.com/rest/licenses/licenses#get-all-commonly-used-licenses).
2. Generate iterator method `LicensesService.ListIter`.
3. Add an integration test `TestLicenses_ListIter` showing that `LicensesService.List` is called at least twice. We have `page: 1`, `per_page: 1`, and number of featured licenses: `3`. I think it's a good integration test because featured licenses change rarely.

Updates #3976

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -14910,6 +14910,14 @@ func (l *ListGlobalSecurityAdvisoriesOptions) GetUpdated() string {
 	return *l.Updated
 }
 
+// GetFeatured returns the Featured field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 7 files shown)


alexandear:
It's impossible to cover the line `return nil, nil, err`.

image

gmlewis:
Merging.

PR #3986, opened 2026-02-10
fix!: Add field `PerPage` to `OrganizationsListOptions`

alexandear:
BREAKING CHANGE: `OrganizationsListOptions` now contains only `PerPage` instead of `ListOptions`.

Updates #3976

We could build an iterator for `ListAll` using the `Since` field, but it is not very useful. `ListAllIter` would return all organizations across GitHub, which is a large result set.

### DIFF
--- a/github/orgs.go
+++ b/github/orgs.go
@@ -154,10 +154,10 @@ func (p Plan) String() string {
 // OrganizationsService.ListAll method.
 type OrganizationsListOptions struct {
 	// An organization ID. Only return organizations with an ID greater than this ID.
-	Since *int64 `url:"since,omitempty"`
+	Since int64 `url:"since,omitempty"`
 
 	// The number of results per page (max 100).
-	PerPage *int `url:"per_page,omitempty"`
+	PerPage int `url:"per_page,omitempty"`
 }
 
 // ListAll lists all organizations, in the order that they were created on GitHub.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -361,6 +361,8 @@ linters:
             - NotificationListOptions.Before # TODO: Activities
             - NotificationListOptions.Participating # TODO: Activities
             - NotificationListOptions.Since # TODO: Activities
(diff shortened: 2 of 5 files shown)


alexandear:
Created issue #3987 about failing `TestDo_rateLimit_abuseRateLimitError_xRateLimitReset`.

github/orgs.go
```diff
@@ -153,13 +153,11 @@ func (p Plan) String() string {
 // OrganizationsListOptions specifies the optional parameters to the
 // OrganizationsService.ListAll method.
 type OrganizationsListOptions struct {
-	// Since filters Organizations by ID.
-	Since int64 `url:"since,omitempty"`
+	// An organization ID. Only return organizations with an ID greater than this ID.
+	Since *int64 `url:"since,omitempty"`
```
gmlewis: So technically, what we had here was fine since these are all URL parameters and although they are of types `int64` and `int`, their default zero value is OK.

According to our README.md:

> All structs for GitHub resources use pointer values for all non-repeated fields.
> This allows distinguishing between unset fields and those set to a zero-value.

So this kind of breaking change is not needed and I think might be confusing to people.


```suggestion
	Since int64 `url:"since,omitempty"`
```

github/orgs.go
```diff
@@ -153,13 +153,11 @@ func (p Plan) String() string {
...
 
-	// Note: Pagination is powered exclusively by the Since parameter,
-	// ListOptions.Page has no effect.
-	// ListOptions.PerPage controls an undocumented GitHub API parameter.
-	ListOptions
+	// The number of results per page (max 100).
+	PerPage *int `url:"per_page,omitempty"`
```
gmlewis:
```suggestion
	PerPage int `url:"per_page,omitempty"`
```

gmlewis (approved):
Excellent! Thank you, @alexandear!
LGTM.

gmlewis:
Merging.

PR #3988, opened 2026-02-10
fix!: Add `ListOptions` to `ListDeploymentBranchPolicies` and `ListCustomDeploymentRuleIntegrations`

Not-Dhananjay-Mishra:
BREAKING CHANGE: `RepositoriesService.ListDeploymentBranchPolicies` and `RepositoriesService.ListCustomDeploymentRuleIntegrations` now accept one more parameter `ListOptions`

RepositoriesService.ListCustomDeploymentRuleIntegrations also support pagination acc to [docs](https://docs.github.com/rest/deployments/protection-rules#list-custom-deployment-rule-integrations-available-for-an-environment).
RepositoriesService.ListDeploymentBranchPolicies also support pagination acc to [docs](https://docs.github.com/rest/deployments/branch-policies#list-deployment-branch-policies).

Updates: #3976

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -231,38 +231,27 @@ linters:
             - WeeklyStats.Deletions
             - WeeklyStats.Week
           allowed-tag-types:
-            - ActivityListStarredOptions.Direction # TODO: Activities
-            - ActivityListStarredOptions.Sort # TODO: Activities
-            - AddProjectItemOptions.ID # TODO: Projects
-            - AddProjectItemOptions.Type # TODO: Projects
-            - AlertInstancesListOptions.Ref # TODO: CodeScanning
-            - AlertListOptions.Direction # TODO: CodeScanning
-            - AlertListOptions.Ref # TODO: CodeScanning
-            - AlertListOptions.Severity # TODO: CodeScanning
-            - AlertListOptions.Sort # TODO: CodeScanning
-            - AlertListOptions.State # TODO: CodeScanning
-            - AlertListOptions.ToolGUID # TODO: CodeScanning
-            - AlertListOptions.ToolName # TODO: CodeScanning
             - APIMetaArtifactAttestations.TrustDomain # TODO: Meta
-            - CommitsListOptions.Author # TODO: Repositories
-            - CommitsListOptions.Path # TODO: Repositories
-            - CommitsListOptions.SHA # TODO: Repositories
+            - ActionsCacheListOptions.Direction
+            - ActionsCacheListOptions.Key
+            - ActionsCacheListOptions.Ref
+            - ActionsCacheListOptions.Sort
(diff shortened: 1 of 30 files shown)


github/repos_deployment_branch_policies_test.go
```diff
@@ -40,7 +45,7 @@ func TestRepositoriesService_ListDeploymentBranchPolicies(t *testing.T) {
 
 	const methodName = "ListDeploymentBranchPolicies"
 	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
```
gmlewis, line 52: When you add `ListOptions` to a method that has string params, you can also typically add `testBadOptions` here before `testNewRequestAndDoFailure` and pass in a newline (`\n`) for each string parameter to cause it to fail.

github/repos_deployment_protection_rules_test.go
```diff
@@ -135,7 +140,7 @@ func TestRepositoriesService_ListCustomDeploymentRuleIntegrations(t *testing.T)
 
 	const methodName = "ListCustomDeploymentRuleIntegrations"
 	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
```
gmlewis, line 146: Same here.

gmlewis:
And now I'm left wondering why new iterators weren't created by `gen-iterators.go`?

gmlewis:
And why didn't the CI/CD catch that the iterators had not been updated?

Not-Dhananjay-Mishra:
> And now I'm left wondering why new iterators weren't created by `gen-iterators.go`?

Both of these functions return a struct not a array of struct. even when I add 
```go
var customTestJSON = map[string]string{
	"ListAllTopics":                                `{"names": []}`,
	"ListUserInstallations":                        `{"installations": []}`,
	"DeploymentBranchPolicyResponse":               `{"branch_policies": []}`,
	"ListCustomDeploymentRuleIntegrationsResponse": `{"available_custom_deployment_protection_rule_integrations": []}`,
}
```
and run `go generate ./...` iterator for them not generates

Edit : I don't think customTestJSON will work if function returns struct

gmlewis:
> Edit : I don't think customTestJSON will work if function returns struct

Ah!  OK, maybe it needs to support the return of a struct? Sounds like a feature request!

@Not-Dhananjay-Mishra - would you mind opening a new issue to add support to `gen-iterators.go` for methods that return structs if its method takes either `ListOptions` or `ListCursorOptions`, please?

gmlewis:
@Not-Dhananjay-Mishra - it will be interesting to see how #3991 affects this PR.
Please let me know if you experience any problems when merging `master` into this PR.

gmlewis:
Thank you, @alexandear!
Merging.

Issue #3990, opened 2026-02-10
Consistent use of non-pointer fields in URL option structs


Not-Dhananjay-Mishra:
As discuses in [comment](https://github.com/google/go-github/pull/3984#discussion_r2788326034) we should **NOT** use pointers for URL struct tags.
### **Example :**
Currently (wrong) - 
https://github.com/google/go-github/blob/e5024aaf1af67d70ef537fc8118d6638d1eab1f1/github/orgs_audit_log.go#L15-L21

---

Correct - 
```go
 type GetAuditLogOptions struct { 
 	Phrase  string `url:"phrase,omitempty"`
 	Include string `url:"include,omitempty"`
 	Order   string `url:"order,omitempty"`  
  
 	ListCursorOptions 
 }
```
Task :
- Update all URL option structs to remove pointer fields
- Update the `structfield` linter to catch this issue going forward

PR #3984, opened 2026-02-09
fix!: Fix pagination support for `IssuesService` list methods

maishivamhoo123:
BREAKING CHANGE: `ListCursorOptions` is removed from `IssueListOptions`.

Updates: #3976.

Removed unsupported cursor pagination IssuesService.List and IssuesService.ListByRepo by deleting ListCursorOptions from issue option structs. Updated related unit tests accordingly.

PR #3907, opened 2026-01-08
chore(deps): Switch from `gopkg.in/yaml.v3` to `go.yaml.in/yaml/v3` in /tools

scop:
The latter is a maintained fork of the former, https://github.com/yaml/go-yaml#project-status

While at it, upgrade from 3.0.1 to 3.0.4.

3.0.1 are identical, 3.0.1..3.0.2 diff between upstreams: https://gist.github.com/scop/57f94f2bf97fb6a2d01349fa59d88530

PR #3991, opened 2026-02-10
fix!: Pass `url` struct tags by value instead of by reference

gmlewis:
BREAKING CHANGE: Many `*Options` structs now pass `omitempty` URL struct fields by value instead of by reference.

Fixes: #3990.
Relates to: #3984.

Over time, I had allowed and even encouraged pointers to primitive types to be used in the URL struct tags when "omitempty" was provided. This violates the spirit of the purpose of the use of pointers for GitHub resource types as declared in this repo's README.md, and this is the first step to getting back on track to a self-consistent client library.

I apologize for the breakages that this PR causes and for all the contributors whom I wrongly steered during code reviews.

Note that timestamps can still be passed by reference (or by value if using `omitzero`).

Also note that a new `check-structfield-settings` tool is added to this repo that will report (and optionally `-fix`) any obsolete exceptions listed in the `.golangci.yml` file at the root of the repo.

### DIFF
--- a/github/github-iterators.go
+++ b/github/github-iterators.go
@@ -1312,10 +1312,9 @@ func (s *IssuesService) ListIter(ctx context.Context, all bool, opts *IssueListO
 				}
 			}
 
-			if resp.Cursor == "" && resp.NextPage == 0 {
+			if resp.NextPage == 0 {
 				break
 			}
-			opts.ListCursorOptions.Cursor = resp.Cursor
 			opts.ListOptions.Page = resp.NextPage
 		}
 	}
@@ -1375,10 +1374,9 @@ func (s *IssuesService) ListByOrgIter(ctx context.Context, org string, opts *Iss
 				}
 			}
 
-			if resp.Cursor == "" && resp.NextPage == 0 {
+			if resp.NextPage == 0 {
 				break
 			}
-			opts.ListCursorOptions.Cursor = resp.Cursor
 			opts.ListOptions.Page = resp.NextPage
 		}
 	}
@@ -3767,10 +3765,9 @@ func (s *SubIssueService) ListByIssueIter(ctx context.Context, owner string, rep
 				}
 			}
 
-			if resp.Cursor == "" && resp.NextPage == 0 {
+			if resp.NextPage == 0 {
 				break
 			}
-			opts.ListCursorOptions.Cursor = resp.Cursor
 			opts.ListOptions.Page = resp.NextPage
 		}
 	}
--- a/github/github-iterators_test.go
+++ b/github/github-iterators_test.go
@@ -2975,7 +2975,7 @@ func TestIssuesService_ListIter(t *testing.T) {
 		callNum++
 		switch callNum {
 		case 1:
-			w.Header().Set("Link", `<https://api.github.com/?cursor=yo>; rel="next"`)
+			w.Header().Set("Link", `<https://api.github.com/?page=1>; rel="next"`)
 			fmt.Fprint(w, `[{},{},{}]`)
 		case 2:
 			fmt.Fprint(w, `[{},{},{},{}]`)
@@ -3119,7 +3119,7 @@ func TestIssuesService_ListByOrgIter(t *testing.T) {
 		callNum++
 		switch callNum {
 		case 1:
-			w.Header().Set("Link", `<https://api.github.com/?cursor=yo>; rel="next"`)
+			w.Header().Set("Link", `<https://api.github.com/?page=1>; rel="next"`)
 			fmt.Fprint(w, `[{},{},{}]`)
 		case 2:
 			fmt.Fprint(w, `[{},{},{},{}]`)
@@ -8663,7 +8663,7 @@ func TestSubIssueService_ListByIssueIter(t *testing.T) {
 		callNum++
 		switch callNum {
 		case 1:
-			w.Header().Set("Link", `<https://api.github.com/?cursor=yo>; rel="next"`)
+			w.Header().Set("Link", `<https://api.github.com/?page=1>; rel="next"`)
 			fmt.Fprint(w, `[{},{},{}]`)
 		case 2:
 			fmt.Fprint(w, `[{},{},{},{}]`)
--- a/github/issues.go
+++ b/github/issues.go
@@ -125,10 +125,7 @@ type IssueListOptions struct {
 	// Since filters issues by time.
 	Since time.Time `url:"since,omitempty"`
 
-	ListCursorOptions
-
 	// Add ListOptions so offset pagination with integer type "page" query parameter is accepted
-	// since ListCursorOptions accepts "page" as string only.
 	ListOptions
 }
 
@@ -245,6 +242,7 @@ type IssueListByRepoOptions struct {
 	// Since filters issues by time.
 	Since time.Time `url:"since,omitempty"`
 
+	// ListCursorOptions specifies the optional parameters for cursor pagination.
 	ListCursorOptions
 
 	// Add ListOptions so offset pagination with integer type "page" query parameter is accepted
--- a/github/issues_test.go
+++ b/github/issues_test.go
@@ -28,22 +28,21 @@ func TestIssuesService_List_all(t *testing.T) {
 			"labels":    "a,b",
 			"sort":      "updated",
 			"direction": "asc",
-			"since":     "2002-02-10T15:30:00Z",
+			"since":     referenceTime.Format(time.RFC3339),
 			"page":      "1",
 			"per_page":  "2",
-			"before":    "foo",
-			"after":     "bar",
 		})
 		fmt.Fprint(w, `[{"number":1}]`)
 	})
 
 	opt := &IssueListOptions{
-		"all", "closed",
-		[]string{"a", "b"},
-		"updated", "asc",
-		time.Date(2002, time.February, 10, 15, 30, 0, 0, time.UTC),
-		ListCursorOptions{Before: "foo", After: "bar"},
-		ListOptions{Page: 1, PerPage: 2},
+		Filter:      "all",
+		State:       "closed",
+		Labels:      []string{"a", "b"},
+		Sort:        "updated",
+		Direction:   "asc",
+		Since:       referenceTime,
+		ListOptions: ListOptions{Page: 1, PerPage: 2},
 	}
 	ctx := t.Context()
 	issues, _, err := client.Issues.List(ctx, true, opt)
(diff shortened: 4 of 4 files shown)


gmlewis:
If any of you have time to review, that would be greatly appreciated.

tools/structfield/structfield.go
```diff
@@ -280,24 +291,41 @@ func checkAndReportInvalidTypes(structName, goFieldName string, fieldType ast.Ex
 			const msg = "change the %q field type to %q in the struct %q"
 			pass.Reportf(tokenPos, msg, goFieldName, exprToString(ft.X), structName)
 		}
-		return true
+		if ident, ok := ft.X.(*ast.Ident); ok && tagType == "url" {
+			switch ident.Name {
+			// Remove the pointer for primitives in url tags
+			case "string", "int", "int64", "float64", "bool":
+				return ident.Name, false
```
Not-Dhananjay-Mishra: Can we use `isBuiltinType` helper function?
```suggestion
            // Remove the pointer for primitives in url tags
			if isBuiltinType(ident.Name){
				return ident.Name, false
```

Not-Dhananjay-Mishra:
Also, should we update all the struct in this PR? I believe a lot of struct need to be change

tools/check-structfield-settings/main.go
```diff
@@ -0,0 +1,414 @@
+// Copyright 2026 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// check-structfield-settings reads the settings for
+// the custom `structfield` linter in ".golangci.yml" -
+// specifically, the "allowed-tag-names" and "allowed-tag-types"
+// exceptions, then scans the code repo to find all exceptions
+// that are no longer needed and reports a list.
```
alexandear, line 10: I'm curious, can we move this logic of detecting unused config entries to the `structfield` itself?

alexandear: Looks like we can't. This comment can be resolved.

script/run-check-structfield-settings.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash -e
+pushd tools/check-structfield-settings && go run . "$@"
```
alexandear: I propose the following:
```suggestion
go run -C tools/check-structfield-settings  . "$@"
```

tools/check-structfield-settings/go.mod
```diff
@@ -0,0 +1,18 @@
+module github.com/google/go-github/v82/tools/check-structfield-settings
+
+go 1.24.0
+
+require (
+	github.com/golangci/plugin-module-register v0.1.1
+	github.com/google/go-github/v82/tools/structfield v0.0.0
+	gopkg.in/yaml.v3 v3.0.1
```
alexandear: Better to use `go.yaml.in/yaml/v3`. We recently did this in #3907

tools/check-structfield-settings/go.mod
```diff
@@ -0,0 +1,18 @@
...
+)
+
+require (
+	golang.org/x/mod v0.22.0 // indirect
+	golang.org/x/sync v0.10.0 // indirect
+	golang.org/x/tools v0.29.0
```
alexandear: This is a direct dependency; let's move it to the upper `require` section.

tools/check-structfield-settings/main.go
```diff
@@ -0,0 +1,414 @@
...
+	return duplicates
+}
+
+func sortedKeys(values map[string]int) []string {
+	keys := make([]string, 0, len(values))
+	for key := range values {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+	return keys
```
alexandear: I believe we can simplify:
```suggestion
	return slices.Sorted(maps.Keys(values))
```

tools/check-structfield-settings/main.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+func removeObsoleteExceptions(configPath string, obsoleteNames, obsoleteTypes []string) error {
```
alexandear, line 302: I think we can remove obsolete exceptions manually and not use the `-fix`. The benefit of maintaining this function is negligible because it is very complex, and removals are infrequent anyway.

tools/check-structfield-settings/main.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+func diffKeys(all, used map[string]bool) []string {
+	var obsolete []string
+	for key := range all {
+		if !used[key] {
+			obsolete = append(obsolete, key)
+		}
+	}
+	sort.Strings(obsolete)
```
alexandear:
```suggestion
	slices.Sort(obsolete)
```

tools/check-structfield-settings/main.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	if *fix {
+		if err := removeObsoleteExceptions(resolvedConfig, obsoleteNames, obsoleteTypes); err != nil {
+			log.Fatalf("fix: %v", err)
+		}
+	}
+
+	if len(obsoleteNames) > 0 {
+		fmt.Println("Obsolete allowed-tag-names:")
```
alexandear, line 89: We should not mix `fmt.Println` with `log.Println`. It is better to choose one and use it consistently throughout the file.

tools/check-structfield-settings/main.go
gmlewis: Right!  I tried long and hard to do this on the original custom linter, and it turns out that the surrounding machinery flat-out prevents this!!!

So I finally got super frustrated and added this tool. It's a bummer, but the upside is that it embeds the `structfield` linter itself so we have one source-of-truth for the linter logic.

gmlewis:
> Also, should we update all the struct in this PR? I believe a lot of struct need to be change

In this PR, I updated the ones that I know are "safe" - all the pagination ones, etc. I left the others for later review on a one-by-one basis.

tools/check-structfield-settings/main.go
gmlewis: As a maintainer, this `-fix` comes in mighty handy... note that this will primarily be a tool for the maintainer to use.
I prefer to keep it, please.

tools/check-structfield-settings/main.go
gmlewis: I'll use `fmt` for tool output to stdout and `log` for output to stderr, as this is common practice.
You are right that there is currently a mix and I'll address it.  Tool output should be `fmt`.

gmlewis:
Thank you, @Not-Dhananjay-Mishra and @alexandear, for the great review feedback!
I believe I've addressed all your comments (except for `-fix` which I want to keep) and have pushed the changes.

gmlewis:
Oh!  And here's one other idea I had for the `-fix` option... we could possibly add it to `script/lint.sh` and/or `script/generate.sh` so that it would automatically alert a developer when the ".golintci.yml" file needs updating.
I'm picturing it being used in a non-destructive temporary area like `script/generate.sh` currently uses so that it could just see if there are any differences caused by running it.

Thoughts?

alexandear:
> Oh! And here's one other idea I had for the `-fix` option... we could possibly add it to `script/lint.sh` and/or `script/generate.sh` so that it would automatically alert a developer when the ".golintci.yml" file needs updating. I'm picturing it being used in a non-destructive temporary area like `script/generate.sh` currently uses so that it could just see if there are any differences caused by running it.

Great idea. Let's add it to `generate.sh`.

gmlewis:
> Great idea. Let's add it to `generate.sh`.

OK, I've incorporated it and pushed the change... amazing that the change was trivial... the comments were more difficult to write. :joy:

Here is an example run to show that it is working:

```
glenn@glenn-MacM2Pro ~/go/src/github.com/google/go-github (i3990-fix-url-tags) $ ./script/generate.sh --check
Obsolete allowed-tag-types:
  - ActivityListStarredOptions.Direction
Generated files are out of date. Please run script/generate.sh and commit the results
diff --git a/.golangci.yml b/.golangci.yml
index 11f70525..51d47ebb 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -231,7 +231,6 @@ linters:
             - WeeklyStats.Deletions
             - WeeklyStats.Week
           allowed-tag-types:
-            - ActivityListStarredOptions.Direction # TODO: Activities
             - APIMetaArtifactAttestations.TrustDomain # TODO: Meta
             - ActionsCacheListOptions.Direction
             - ActionsCacheListOptions.Key
glenn@glenn-MacM2Pro ~/go/src/github.com/google/go-github (i3990-fix-url-tags) $ ./script/generate.sh
Obsolete allowed-tag-types:
  - ActivityListStarredOptions.Direction
glenn@glenn-MacM2Pro ~/go/src/github.com/google/go-github (i3990-fix-url-tags) $ ./script/generate.sh --check
No obsolete structfield exceptions found.
```

Now I'm wondering if `-fix` should be silent?
Do you think the message:

```
$ ./script/generate.sh
Obsolete allowed-tag-types:
  - ActivityListStarredOptions.Direction
```

might be confusing to end-users?

Not-Dhananjay-Mishra:
> Now I'm wondering if `-fix` should be silent?
> 

I think we should tell the end users that script has done some changes in `.golangci.yml`.
How about `Removed {x} obsolete fields from .golangci.yml`
Thoughts?

gmlewis:
> > Now I'm wondering if `-fix` should be silent?
> 
> I think we should tell the end users that script has done some changes in `.golangci.yml`. How about `Removed {x} obsolete fields from .golangci.yml` Thoughts?

OK, I thought I saw something wrong, but I think this is what we are ending up with:

```
glenn@glenn-MacM2Pro ~/go/src/github.com/google/go-github (i3990-fix-url-tags) $ ./script/generate.sh --check
Removed 1 obsolete fields from .golangci.yml.
Generated files are out of date. Please run script/generate.sh and commit the results
diff --git a/.golangci.yml b/.golangci.yml
index f6ac98d3..51d47ebb 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -167,7 +167,6 @@ linters:
         original-url: github.com/google/go-github/v82/tools/structfield
         settings:
           allowed-tag-names:
-            - ActivityListStarredOptions.Direction
             - ActionsCacheUsageList.RepoCacheUsage # TODO: RepoCacheUsages ?
             - AuditEntry.ExternalIdentityNameID
             - AuditEntry.Timestamp
glenn@glenn-MacM2Pro ~/go/src/github.com/google/go-github (i3990-fix-url-tags) $ ./script/generate.sh
Removed 1 obsolete fields from .golangci.yml.
glenn@glenn-MacM2Pro ~/go/src/github.com/google/go-github (i3990-fix-url-tags) $ ./script/generate.sh --check
No obsolete structfield exceptions found.
```

I saw that "Removed" for the `--check` but it really is removing it in its own copy... so I think we can't avoid that easily.

Not-Dhananjay-Mishra:
> I saw that "Removed" for the `--check` but it really is removing it in its own copy... so I think we can't avoid that easily.

If we can't avoid that, I think then it's fine to remove that print statement 🙃

gmlewis:
> > I saw that "Removed" for the `--check` but it really is removing it in its own copy... so I think we can't avoid that easily.
> 
> If we can't avoid that, I think then it's fine to remove that print statement 🙃

OK, then `-fix` will be silent.  That's probably better anyway. 
Please take another look.

gmlewis:
Merging.

PR #3994, opened 2026-02-11
fix: Change cursor pagination to use `After`

gmlewis:
Fixes: #3964.

@Not-Dhananjay-Mishra pointed out in #3964 that cursor-based pagination is not working, at least on the `SecurityAdvisories.ListGlobalSecurityAdvisories` endpoint, and was correct that using the `After` field instead of the `Cursor` field solved the problem.

This has been tested now and without using a throttling transport, I quickly got rate-limited. 😅

### DIFF
--- a/test/integration/pagination_test.go
+++ b/test/integration/pagination_test.go
@@ -1,0 +1,38 @@
+// Copyright 2026 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build integration
+
+package integration
+
+import (
+	"testing"
+
+	"github.com/google/go-github/v82/github"
+)
+
+func TestSecurityAdvisories_ListGlobalSecurityAdvisories(t *testing.T) {
+	opt := &github.ListGlobalSecurityAdvisoriesOptions{
+		ListCursorOptions: github.ListCursorOptions{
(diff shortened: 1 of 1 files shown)


github/gen-iterators.go
alexandear, line 1: I propose adding an integration test to ensure that GitHub does not break the API again without us knowing.

Something like this:

```go
func TestSecurityAdvisories_ListGlobalSecurityAdvisories(t *testing.T) {
	opt := &github.ListGlobalSecurityAdvisoriesOptions{
		ListCursorOptions: github.ListCursorOptions{
			PerPage: 2,
		},
	}
	advisories, resp, err := client.SecurityAdvisories.ListGlobalSecurityAdvisories(t.Context(), opt)
	if err != nil {
		t.Fatalf("ListGlobalSecurityAdvisories returned error: %v", err)
	}

	if got, want := len(advisories), 2; got != want {
		t.Errorf("ListGlobalSecurityAdvisories returned %v advisories, want %v", got, want)
	}

	if resp.After == "" {
		t.Error("ListGlobalSecurityAdvisories returned an empty 'after' cursor")
	}

	if resp.Cursor != "" {
		t.Error("ListGlobalSecurityAdvisories returned a non-empty 'cursor' value")
	}
}
```

gmlewis:
Thank you, @alexandear! I added the integration test.
Merging.

PR #3995, opened 2026-02-12
chore: Update workflow and tools to use Go 1.26 and 1.25

gmlewis:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/google/go-github/v82
+module github.com/google/go-github/v83
 
 go 1.25.0
 
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,8 +1,8 @@
 version: v2.7.0
 plugins:
-  - module: "github.com/google/go-github/v82/tools/fmtpercentv"
+  - module: "github.com/google/go-github/v83/tools/fmtpercentv"
     path: ./tools/fmtpercentv
-  - module: "github.com/google/go-github/v82/tools/sliceofpointers"
+  - module: "github.com/google/go-github/v83/tools/sliceofpointers"
     path: ./tools/sliceofpointers
-  - module: "github.com/google/go-github/v82/tools/structfield"
+  - module: "github.com/google/go-github/v83/tools/structfield"
     path: ./tools/structfield
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,7 @@ version: "2"
 run:
   build-tags:
     - integration
+  allow-parallel-runners: true
 linters:
   enable:
     - canonicalheader
@@ -158,15 +159,15 @@ linters:
       fmtpercentv:
         type: module
         description: Reports usage of %d or %s in format strings.
-        original-url: github.com/google/go-github/v82/tools/fmtpercentv
+        original-url: github.com/google/go-github/v83/tools/fmtpercentv
       sliceofpointers:
         type: module
         description: Reports usage of []*string and slices of structs without pointers.
-        original-url: github.com/google/go-github/v82/tools/sliceofpointers
+        original-url: github.com/google/go-github/v83/tools/sliceofpointers
       structfield:
         type: module
         description: Reports mismatches between Go field and JSON, URL tag names and types.
-        original-url: github.com/google/go-github/v82/tools/structfield
+        original-url: github.com/google/go-github/v83/tools/structfield
         settings:
           allowed-tag-names:
             - ActionsCacheUsageList.RepoCacheUsage # TODO: RepoCacheUsages ?
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@
 2BFL <imqksl@gmail.com>
 413x <dedifferentiator@gmail.com>
 6543 <6543@obermui.de>
+Aaron LaBrie <bigdadbear@github.com>
 Abed Kibbe <abed.kibbe@gmail.com>
 Abhijit Hota <abhihota025@gmail.com>
 Abhinav Gupta <mail@abhinavg.net>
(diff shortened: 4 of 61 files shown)


Not-Dhananjay-Mishra (approved):
LGTM.
Just one question should we update README.md as well?
https://github.com/google/go-github/blob/1732098cb3b0f3b7eb7edf410837dc8a8cdd12c9/README.md?plain=1#L15-L16

gmlewis:
> LGTM. Just one question should we update README.md as well?
> 
> https://github.com/google/go-github/blob/1732098cb3b0f3b7eb7edf410837dc8a8cdd12c9/README.md?plain=1#L15-L16

No, that section should stay the same.

.github/workflows/tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         shell: bash
     strategy:
       matrix:
-        go-version: [1.x, 1.24.0] # test with N and the .0 release of N-1
+        go-version: [1.x, 1.25.0] # test with N and the .0 release of N-1
```
alexandear: This can be a little better:
```suggestion
        go-version: ['stable', 'oldstable'] # test with N and the .0 release of N-1
```

From https://github.com/actions/setup-go#version-aliases:

Image

gmlewis:
> Could we ship this change after the go-github release with all the breaking changes? I mean releasing **v83** first and then updating the `go.mod` to Go 1.25.

Great idea, @alexandear - placing this PR on hold until after the release of `v83.0.0`.

.github/workflows/tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         shell: bash
     strategy:
       matrix:
-        go-version: [1.x, 1.24.0] # test with N and the .0 release of N-1
+        go-version: ['stable', 'oldstable'] # test with N and the .0 release of N-1
```
stevehipwell: We could use the following command to dynamically get the actual Go versions, but I'm not sure there is much/any value (I'm just adding it here for completeness).

```shell
curl -s https://endoflife.date/api/go.json | jq -r '[sort_by(.releaseDate) | reverse | .[0:2] | .[].cycle] | join(",")'
```

.github/workflows/tests.yml
stevehipwell:
```suggestion
          - go-version: stable
```

This would be consistent.

stevehipwell:
```suggestion
          - go-version: stable
```

This would be consistent.

stevehipwell, line 41: Do we want this to be an additional run? With the current pattern we could either put all of the scenarios into `include` and remove the top level `go-version` & `platform` keys, or remove this one and calculate `update-coverage` as `${{ matrix.go-version == 'stable' && matrix.platform == 'ubuntu-latest' }}`.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/google/go-github/v82
 
-go 1.24.0
+go 1.25.0
```
stevehipwell, line 3:
```suggestion
go 1.25
```

Isn't the recommended pattern since Go `1.23` to use `go 1.x` here? Dependabot did have issues with this initially but this should be working correctly now.

alexandear: I think it should be `go 1.25.0` for consistency with the `go mod` command that adds `0`.

From https://go.dev/doc/go1.26#go-command:

> `go mod init` now defaults to a lower go version in new `go.mod` files. Running `go mod init` using a toolchain of version `1.N.X` will create a `go.mod` file specifying the Go version `go 1.(N-1).0`. Pre-release versions of 1.N will create `go.mod` files specifying `go 1.(N-2).0`. For example, the Go 1.26 release candidates will create `go.mod` files with `go 1.24.0`, and Go 1.26 and its minor releases will create go.mod files with `go 1.25.0`.

.github/workflows/tests.yml
alexandear, line 1: Let's also change `go-version: 1.x` to `go-version: 'stable'` in the `.github/workflows/linter.yml`:

https://github.com/google/go-github/blob/0f4fdc7667f794bfcfd4f09fad2d69a3390084cb/.github/workflows/linter.yml#L14

.github/workflows/tests.yml
gmlewis: I think I like the 'stable' and 'oldstable' option best since that work of figuring out the versions is done elsewhere and keeps this super simple.

.github/workflows/tests.yml
gmlewis: This seems simple and clear to me, and I'm not sure I'm understanding any benefit to change it.

gmlewis:
Note to self: #3932 can be addressed once this is merged.

.github/workflows/linter.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
     - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
       with:
-        go-version: 1.x
+        go-version: 'stable'
```
stevehipwell:
```suggestion
        go-version: stable
```
Just a nit, but this is more idiomatic.

.github/workflows/tests.yml
```diff
@@ -27,16 +27,16 @@ jobs:
...
```
stevehipwell:
```suggestion
        go-version: [stable, oldstable] # test with N and the .0 release of N-1
```
Same as above.

stevehipwell (approved):
LGTM

.github/workflows/linter.yml
gmlewis: Why do the docs show using single quotes?
https://github.com/actions/setup-go#version-aliases

stevehipwell: I have no idea (well none that I'm comfortable sharing here), but in this context they serve no purpose and generally cause tooling to churn the files. GitHub Actions uses YAML, so no quoting of strings unless required, in which case they should be `"` quotes. Expressions require strings to be quoted with `'`, sometimes expressions need to be wrapped as strings.

PR #3997, opened 2026-02-12
fix!: Add `ListCursorOptions` to `Enterprise` and `Org` code security configs

maishivamhoo123:
This PR updates `ListEnterpriseCodeSecurityConfigurationOptions` and `ListCodeSecurityConfigurationRepositoriesOptions` to embed `ListCursorOptions`.

**Changes:**
- Replaced manual pagination fields (`Before`, `After`, `PerPage`) with `ListCursorOptions` embedding.
**Reason:**
These endpoints support cursor-based pagination. Explicitly embedding `ListCursorOptions` ensures the library generates the correct `Iter` methods, preventing infinite loops and aligning with the GitHub API.

Issue Related #3976

### DIFF
--- a/github/repos_deployment_branch_policies.go
+++ b/github/repos_deployment_branch_policies.go
@@ -35,8 +35,12 @@ type DeploymentBranchPolicyRequest struct {
 // GitHub API docs: https://docs.github.com/rest/deployments/branch-policies#list-deployment-branch-policies
 //
 //meta:operation GET /repos/{owner}/{repo}/environments/{environment_name}/deployment-branch-policies
-func (s *RepositoriesService) ListDeploymentBranchPolicies(ctx context.Context, owner, repo, environment string) (*DeploymentBranchPolicyResponse, *Response, error) {
+func (s *RepositoriesService) ListDeploymentBranchPolicies(ctx context.Context, owner, repo, environment string, opts *ListOptions) (*DeploymentBranchPolicyResponse, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/environments/%v/deployment-branch-policies", owner, repo, environment)
+	u, err := addOptions(u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	req, err := s.client.NewRequest("GET", u, nil)
 	if err != nil {
--- a/github/repos_deployment_branch_policies_test.go
+++ b/github/repos_deployment_branch_policies_test.go
@@ -17,12 +17,17 @@ func TestRepositoriesService_ListDeploymentBranchPolicies(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
 
-	mux.HandleFunc("/repos/o/r/environments/e/deployment-branch-policies", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc("/repos/o/r/environments/e/deployment-branch-policies", func(w http.ResponseWriter, r *http.Request) {
+		testFormValues(t, r, values{
+			"page":     "1",
+			"per_page": "30",
+		})
 		fmt.Fprint(w, `{"total_count":2, "branch_policies":[{"id":1}, {"id": 2}]}`)
 	})
 
+	opts := &ListOptions{Page: 1, PerPage: 30}
 	ctx := t.Context()
-	got, _, err := client.Repositories.ListDeploymentBranchPolicies(ctx, "o", "r", "e")
+	got, _, err := client.Repositories.ListDeploymentBranchPolicies(ctx, "o", "r", "e", opts)
 	if err != nil {
 		t.Errorf("Repositories.ListDeploymentBranchPolicies returned error: %v", err)
 	}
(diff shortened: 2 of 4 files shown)


github/enterprise_codesecurity_configurations.go
```diff
@@ -20,14 +20,7 @@ import (
...
 // per page (max 100 per GitHub API docs).
 type ListEnterpriseCodeSecurityConfigurationOptions struct {
-	// A cursor, as given in the Link header. If specified, the query only searches for security configurations before this cursor.
-	Before string `url:"before,omitempty"`
-
-	// A cursor, as given in the Link header. If specified, the query only searches for security configurations after this cursor.
-	After string `url:"after,omitempty"`
-
-	// For paginated result sets, the number of results to include per page.
-	PerPage int `url:"per_page,omitempty"`
```
alexandear, line 30: @gmlewis maybe it is better to simply extend the `gen-iterators.go` to support generating iterator methods from the `ListEnterpriseCodeSecurityConfigurationOptions`?

alexandear: I don't like this kind of breaking change.

gmlewis: I'm not sure I understand, @alexandear - the point of using `ListOptions` and `ListCursorOptions` is to standardize the pagination handling. Since this endpoint uses cursor pagination, it makes sense to me to embed `ListCursorOptions` like we do with the other endpoints that support cursor pagination.
The next release is already going to have a large number of breaking API changes, mostly related to pagination, so I think it is fine to include this one.

alexandear: `ListCursorOptions` contains more fields than [the GitHub API expects](https://docs.github.com/en/rest/code-security/configurations?apiVersion=2022-11-28#get-code-security-configurations-for-an-enterprise). From a user's perspective, it's better to have a simpler `ListEnterpriseCodeSecurityConfigurationOptions` with only three fields: `before`, `after`, and `per_page`.

`ListCursorOptions`:
https://github.com/google/go-github/blob/1732098cb3b0f3b7eb7edf410837dc8a8cdd12c9/github/github.go#L267-L290

alexandear: Anyway, if we decide to use `ListCursorOptions`, let's simply inline it instead of `ListEnterpriseCodeSecurityConfigurationOptions`:

```go
func (s *EnterpriseService) ListCodeSecurityConfigurations(ctx context.Context, enterprise string, opts *ListCursorOptions) ([]*CodeSecurityConfiguration, *Response, error) {
```

maishivamhoo123: @alexandear Sir Thank you for the suggestion.

I feel we should keep the wrapper struct as it is, rather than inlining `ListCursorOptions`.

The reason is that `github/gen-iterators.go` relies on the wrapper to generate the iterator. Specifically, the `hasOptions` function iterates over the `Embeds` list to find a match:
https://github.com/google/go-github/blob/master/github/gen-iterators.go#L177

Since `ListCursorOptions` defines fields directly and does not *embed* itself, `hasOptions` returns `false` if we use it directly in the function signature. This causes the generator to skip creating the `Iter` method for this endpoint entirely.

To support inlining, we would likely need to refactor the generator logic itself. Given this, and the benefit of future extensibility, I believe keeping the wrapper struct is the safer approach.

Please let me know if I have misunderstood the logic or if there is a better way to handle this.

alexandear: > To support inlining, we would likely need to refactor the generator logic itself

You're correct.

Since we had already handled direct `ListOptions`, I added support for `ListCursorOptions` in #4007 for consistency.

gmlewis: Thank you, @alexandear!  I now see what you mean. :joy:
Sounds good to me.

gmlewis:
@maishivamhoo123 and @alexandear - are we agreed that this PR can be closed due to #4007?

github/enterprise_codesecurity_configurations.go
maishivamhoo123: @alexandear Sir, I understand now. Instead of forcing a wrapper struct, you updated the generator to recursively check if the struct contains the `After` field. That makes sense!

@gmlewis @alexandear Sir I have a small doubt as there are multiple recursive functions now checking these structs. Since they might be processing the same embedded structs repeatedly, would it be beneficial to apply memoization (Dynamic Programming)?

We could cache the result for each struct type. If a struct is already checked, we can return the result immediately from the cache. This might decrease the generation time.

Thank you, Sir!

gmlewis: > We could cache the result for each struct type. If a struct is already checked, we can return the result immediately from the cache. This might decrease the generation time.

You are welcome to experiment and see if you can improve the runtime of all these linter checks.
If you are successful, please open a separate PR with your improvements.

alexandear:
> @maishivamhoo123 and @alexandear - are we agreed that this PR can be closed due to #4007?

Yes. This PR is not needed anymore.

Now we have [EnterpriseService.ListCodeSecurityConfigurationsIter](https://github.com/google/go-github/blob/4d2c2a5218b2302374060defc90f60eff5da7cf5/github/github-iterators.go#L1045) and [OrganizationsService.ListCodeSecurityConfigurationsIter](https://github.com/google/go-github/blob/4d2c2a5218b2302374060defc90f60eff5da7cf5/github/github-iterators.go#L2038).

gmlewis:
Closing as obsolete.

PR #3999, opened 2026-02-12
chore: Remove unnecessary `fmt.Print` in tests

alexandear:
1. Remove `fmt.Printf` and `fmt.Println` from unit tests. Looks like they were used for debugging.
2. Configure the `forbidigo` linter to forbid `fmt.Print` statements.


gmlewis (approved):
Thank you, @alexandear!
LGTM.
Merging.

PR #4008, opened 2026-02-13
docs: Update usage instructions

alexandear:
This PR removes the usage note explaining how to import this library when Go modules are disabled.

Go modules were introduced in Go 1.11 ([released in 2018](https://go.dev/blog/go1.11)) and are now the de facto standard. Developers who use this library with Go modules disabled are typically experts and don't need guidance on how to import `go-github` into their projects.


gmlewis (approved):
Thank you, @alexandear!
LGTM.
Merging.

Issue #3989, opened 2026-02-10
Support iterator generation for methods returning structs


Not-Dhananjay-Mishra:
`gen-iterators.go` currently does not generate iterators for methods that return a struct, even when those methods accept `ListOptions` or `ListCursorOptions`.

Example:
```go
type ListSomething struct {
    Issue []Struct `json:"Issue"`
}
```
If a method: Accepts `ListOptions` or `ListCursorOptions` and Returns a struct that contains a slice field,
then `gen-iterators.go` should be able to make iterator for it.

mohak1:
Hi, I would like to help with this if no one's already working on it

gmlewis:
> Hi, I would like to help with this if no one's already working on it

Thank you, @mohak1!
It's yours.

gmlewis:
Please check out our CONTRIBUTING.md if you haven't submitted PRs for this repo in the past, and feel free to ask questions in this issue and someone will hopefully be able to help you out.

mohak1:
Thanks @gmlewis, yes it's my first time contributing to this repo. I'll go through the CONTRIBUTING.md to understand the guidelines, and I'll post questions here if I'm stuck or need clarification

gmlewis:
WHUPS!!!  I'm so sorry, @mohak1!!!  I totally forgot you were working on this one!
I thought I was solving a different problem in #4011 (now-closed issue #4010) but then discovered I was solving this one instead!

mohak1:
Ah that's alright @gmlewis, thanks for letting me know

PR #4011, opened 2026-02-13
feat: Support pagination for methods that return structs

gmlewis:
Fixes: #3989.

### DIFF
--- a/github/github-iterators.go
+++ b/github/github-iterators.go
@@ -14,6 +14,1021 @@ import (
 	"iter"
 )
 
+// ListArtifactsIter returns an iterator that paginates through all results of ListArtifacts.
+func (s *ActionsService) ListArtifactsIter(ctx context.Context, owner string, repo string, opts *ListArtifactsOptions) iter.Seq2[*Artifact, error] {
+	return func(yield func(*Artifact, error) bool) {
+		// Create a copy of opts to avoid mutating the caller's struct
+		if opts == nil {
+			opts = &ListArtifactsOptions{}
+		} else {
+			opts = Ptr(*opts)
+		}
+
+		for {
+			results, resp, err := s.ListArtifacts(ctx, owner, repo, opts)
+			if err != nil {
+				yield(nil, err)
+				return
+			}
+
+			var iterItems []*Artifact
+			if results != nil {
+				iterItems = results.Artifacts
+			}
+			for _, item := range iterItems {
+				if !yield(item, nil) {
+					return
+				}
+			}
+
+			if resp.NextPage == 0 {
+				break
+			}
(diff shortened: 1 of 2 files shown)


github/github-iterators.go
```diff
@@ -14,6 +14,1021 @@ import (
...
+
+		for {
+			items, resp, err := s.ListArtifacts(ctx, owner, repo, opts)
+			if err != nil {
+				yield(nil, err)
+				return
+			}
+
+			var iterItems []*Artifact
+			if items != nil {
```
alexandear: Could we rename `items` for clarity? For wrapped structs, it's a `result` struct. For others - `items`.

Something like:
```go
			resultList, resp, err := s.ListArtifacts(ctx, owner, repo, opts)
			if err != nil {
				yield(nil, err)
				return
			}

			var iterItems []*Artifact
			if resultList != nil {
				iterItems = resultList.Artifacts
			}
```

github/github-iterators.go
```diff
@@ -2014,13 +3974,13 @@ func (s *OrganizationsService) ListCodeSecurityConfigurationRepositoriesIter(ctx
 		}
 
 		for {
-			items, resp, err := s.ListCodeSecurityConfigurationRepositories(ctx, org, configurationID, opts)
+			resultList, resp, err := s.ListCodeSecurityConfigurationRepositories(ctx, org, configurationID, opts)
```
alexandear: For methods that return an array the first parameter should be`items`:
```suggestion
			items, resp, err := s.ListCodeSecurityConfigurationRepositories(ctx, org, configurationID, opts)
```

gmlewis: So I was starting to push back on the last request and typed out a long response why I didn't want to make the name change, but decided to see if my simple change was sufficient to satisfy your request. Now that I realize it was not, I'm going to push back.

Having separate naming as you have requested will increase the complexity of the template and make it harder to read and harder to understand, for extremely little return-on-investment.

* `github-iterators.go` will never be hand-edited - it might be looked at to see how it works
* `gen-iterators.go` is the code that will need to be maintained and it MUST be easy to read and easy to understand

Therefore, let's choose a single name that will represent _BOTH_ `items` and `resultList` and stick with that to keep the template logic as simple as possible.  Maybe just calling it `results` or even `r` is idiomatic and sufficient.

Thoughts?

alexandear: You're right about the maintainability concern. Let's stick with `results` for both. It's clear enough.

`r` may be confused with `resp` or `reader.

Not-Dhananjay-Mishra (approved):
Awesome addition to `gen-iterators.go`! 🚀 
LGTM

gmlewis:
Merging.

PR #4012, opened 2026-02-13
fix!: Remove unsupported pagination from `ListAutolinks`

maishivamhoo123:
BREAKING CHANGE: `opts *ListOptions` is removed from `RepositoriesService.ListAutoLinks`.

Related Issue :- #3976 
This PR removed the argument ListOptions from the function signature  ListAutolinks becuase according to the docs the ` ListAutolinks` doesn't supprt pegination [https://docs.github.com/en/rest/repos/autolinks?apiVersion=2022-11-28#get-all-autolinks-of-a-repository](Get All AutoLinks of Repo)


maishivamhoo123:
@gmlewis @Not-Dhananjay-Mishra @alexandear can you please review this PR?

PR #4013, opened 2026-02-13
chore: Bump golangci-lint to v2.9.0

gmlewis:
This should help to speed up the linting process.


gmlewis:
Thank you, @stevehipwell!
Merging.

PR #4014, opened 2026-02-15
fix!: Divide `PackageGetAllVersions` into two separate methods `ListPackageVersions` and `ListUserPackageVersions`

Not-Dhananjay-Mishra:
BREAKING CHANGE: `PackageGetAllVersions` is now divided into `ListPackageVersions` and `ListUserPackageVersions`.

`PackageGetAllVersions` handles two endpoints one of them doesn't support pagination acc to docs also its Iterator doesn't get generated because its method name does not contain `List` as prefix. 

- `ListPackageVersions` [docs](https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-the-authenticated-user)
- `ListUserPackageVersions` [docs](https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-a-user)

Updates : #3976

### DIFF
--- a/github/users_packages.go
+++ b/github/users_packages.go
@@ -127,8 +127,8 @@ func (s *UsersService) RestorePackage(ctx context.Context, user, packageType, pa
 	return s.client.Do(ctx, req, nil)
 }
 
-// ListPackageVersionOptions specifies the optional parameters to the UsersService.ListPackageVersions.
-type ListPackageVersionOptions struct {
+// ListPackageVersionsOptions specifies the optional parameters to the UsersService.ListPackageVersions.
+type ListPackageVersionsOptions struct {
 	// State of package either "active" or "deleted".
 	State string `url:"state,omitempty"`
 
@@ -140,7 +140,7 @@ type ListPackageVersionOptions struct {
 // GitHub API docs: https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-the-authenticated-user
 //
 //meta:operation GET /user/packages/{package_type}/{package_name}/versions
-func (s *UsersService) ListPackageVersions(ctx context.Context, packageType, packageName string, opts *ListPackageVersionOptions) ([]*PackageVersion, *Response, error) {
+func (s *UsersService) ListPackageVersions(ctx context.Context, packageType, packageName string, opts *ListPackageVersionsOptions) ([]*PackageVersion, *Response, error) {
 	u := fmt.Sprintf("user/packages/%v/%v/versions", packageType, packageName)
 	u, err := addOptions(u, opts)
 	if err != nil {
@@ -161,12 +161,12 @@ func (s *UsersService) ListPackageVersions(ctx context.Context, packageType, pac
 	return versions, resp, nil
 }
 
-// ListPackageVersionsForUser gets all versions of a package for a user.
+// ListUserPackageVersions returns package versions for a public package owned by a specified user.
 //
 // GitHub API docs: https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-a-user
 //
 //meta:operation GET /users/{username}/packages/{package_type}/{package_name}/versions
-func (s *UsersService) ListPackageVersionsForUser(ctx context.Context, user, packageType, packageName string) ([]*PackageVersion, *Response, error) {
+func (s *UsersService) ListUserPackageVersions(ctx context.Context, user, packageType, packageName string) ([]*PackageVersion, *Response, error) {
 	u := fmt.Sprintf("users/%v/packages/%v/%v/versions", user, packageType, packageName)
 
 	req, err := s.client.NewRequest("GET", u, nil)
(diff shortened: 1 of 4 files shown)


github/users_packages.go
```diff
@@ -127,22 +127,13 @@ func (s *UsersService) RestorePackage(ctx context.Context, user, packageType, pa
...
 //meta:operation GET /user/packages/{package_type}/{package_name}/versions
-//meta:operation GET /users/{username}/packages/{package_type}/{package_name}/versions
-func (s *UsersService) PackageGetAllVersions(ctx context.Context, user, packageType, packageName string, opts *PackageListOptions) ([]*PackageVersion, *Response, error) {
-	var u string
-	if user != "" {
-		u = fmt.Sprintf("users/%v/packages/%v/%v/versions", user, packageType, packageName)
-	} else {
-		u = fmt.Sprintf("user/packages/%v/%v/versions", packageType, packageName)
-	}
+func (s *UsersService) ListPackageVersions(ctx context.Context, packageType, packageName string, opts *PackageListOptions) ([]*PackageVersion, *Response, error) {
```
Not-Dhananjay-Mishra: This endpoint method accept following query parameters according to docs
Image

but currently it's `opts` field have some extra parameters
https://github.com/google/go-github/blob/e3c4fb576edf192e734c2307fb81cb55d11612cb/github/packages.go#L185-L197

should I make a new `opts` struct for it? like :
```go
type PackageVersionListOptions struct {
	// State of package either "active" or "deleted".
	State *string `url:"state,omitempty"`

	ListOptions
}
```

gmlewis: That sounds good to me.

github/users_packages.go
```diff
@@ -127,22 +127,21 @@ func (s *UsersService) RestorePackage(ctx context.Context, user, packageType, pa
 	return s.client.Do(ctx, req, nil)
 }
 
-// PackageGetAllVersions gets all versions of a package for a user. Passing the empty string for "user" will
-// get versions for the authenticated user.
-//
-// GitHub API docs: https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-a-user
+// PackageVersionListOptions specifies the optional parameters to the UsersService.ListPackageVersions.
+type PackageVersionListOptions struct {
```
alexandear: List prefix is required:

```suggestion
type ListPackageVersionOptions struct {
```

github/users_packages.go
```diff
@@ -127,22 +127,21 @@ func (s *UsersService) RestorePackage(ctx context.Context, user, packageType, pa
 	return s.client.Do(ctx, req, nil)
 }
 
-// PackageGetAllVersions gets all versions of a package for a user. Passing the empty string for "user" will
-// get versions for the authenticated user.
-//
-// GitHub API docs: https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-a-user
+// ListPackageVersionOptions specifies the optional parameters to the UsersService.ListPackageVersions.
+type ListPackageVersionOptions struct {
```
alexandear: Sorry, one more remark:
```suggestion
type ListPackageVersionsOptions struct {
```

I propose to add also `s` for consistency with other similar structs. So if anyone greps the method `ListPackageVersion` also will find the `Options` struct.

github/users_packages.go
```diff
@@ -162,6 +161,28 @@ func (s *UsersService) PackageGetAllVersions(ctx context.Context, user, packageT
 	return versions, resp, nil
 }
 
+// ListPackageVersionsForUser gets all versions of a package for a user.
+//
+// GitHub API docs: https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-a-user
+//
+//meta:operation GET /users/{username}/packages/{package_type}/{package_name}/versions
+func (s *UsersService) ListPackageVersionsForUser(ctx context.Context, user, packageType, packageName string) ([]*PackageVersion, *Response, error) {
```
alexandear: This would be better:
```suggestion
func (s *UsersService) ListUserOwnedPackageVersions(ctx context.Context, user, packageType, packageName string) ([]*PackageVersion, *Response, error) {
```

Or maybe `ListUserPackageVersions`. What do you think?

According to the endpoint's docs:
> Lists package versions for a public package owned by a specified user.

github/users_packages.go
```diff
@@ -162,6 +161,28 @@ func (s *UsersService) PackageGetAllVersions(ctx context.Context, user, packageT
 	return versions, resp, nil
 }
 
+// ListPackageVersionsForUser gets all versions of a package for a user.
```
alexandear: Let's update according to the docs:

```
// ListUserPackageVersions returns package versions for a public package owned by a specified user.
```

github/users_packages.go
Not-Dhananjay-Mishra: I think `ListUserPackageVersions` is better option

gmlewis:
Thank you, @alexandear!
Merging.

Issue #4021, opened 2026-02-18
Implement REST API Endpoint for listing fine grained personal access token requests


AbhishekAg:
The GitHub REST API endpoint for listing fine grained personal access token **requests** (GET 
/orgs/{org}/personal-access-token-requests) is currently not implemented in the go-github library. This endpoint provides valuable information about pending requests. 

API Reference:
https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#review-requests-to-access-organization-resources-with-fine-grained-personal-access-tokens

gmlewis:
Thank you, @AbhishekAg!

This would be a great PR for any new contributor to this repo or a new Go developer.
All contributions are greatly appreciated!

Feel free to volunteer for any issue and the issue can be assigned to you so that others don't attempt to duplicate the work.
Also feel free to perform a code review on any PR that has the label "NeedsReview" - we appreciate those contributions too!
We have a `REVIEWERS` file that you could add your username to if you are willing to be contacted when reviews are needed.

Please check out our [CONTRIBUTING.md](https://github.com/google/go-github/blob/master/CONTRIBUTING.md) guide to get started. (In particular, please remember to `go generate ./...` and don't use force-push to your PRs.)

Thank you!

shuv0id:
Hey! I would like to work on this issue.

gmlewis:
> Hey! I would like to work on this issue.

Thank you, @shuv0id! It's yours.

AbhishekAg:
@gmlewis @shuv0id I made the changes already. Can I raise the PR? Actually I need it little early hence completed the changes

shuv0id:
Completely fine with me if @AbhishekAg already have made the changes.

gmlewis:
OH!  Sorry, @AbhishekAg - next time please tell me that you would like to work on it and I'll immediately assign you!
Sorry, go ahead with your PR.

Sorry, @shuv0id!  That was my bad.

AbhishekAg:
Really sorry for this. I should have updated the thread.

Issue #3987, opened 2026-02-10
Fix flaky test `TestDo_rateLimit_abuseRateLimitError_xRateLimitReset`


alexandear:
Test: https://github.com/google/go-github/blob/c42250864d94287b278a04b0385ab0dcf6e5e1a0/github/github_test.go#L1851

Short logs:
```
testing .
--- FAIL: TestDo_rateLimit_abuseRateLimitError_xRateLimitReset (0.00s)
    github_test.go:1885: abuseRateLimitErr RetryAfter = 2m1.999907571s, want 2m3s
    github_test.go:1900: abuseRateLimitErr RetryAfter = 2m1.999834194s, want 2m3s

FAIL
coverage: 99.2% of statements
FAIL	github.com/google/go-github/v82/github	3.398s
	github.com/google/go-github/v82/test/fields		coverage: 0.0% of statements
?   	github.com/google/go-github/v82/test/integration	[no test files]
FAIL

```

Full logs from the [actions run](https://github.com/google/go-github/actions/runs/21865989222/job/63107390549?pr=3986#step:4:22): [logs_56960921161.zip](https://github.com/user-attachments/files/25211510/logs_56960921161.zip)

PR #4022, opened 2026-02-18
feat: Add `ListFineGrainedPersonalAccessTokenRequests` for org

AbhishekAg:
Implement support for listing requests to access organization resources via fine-grained personal access tokens (GET /orgs/{org}/personal-access-token-requests).

- Add FineGrainedPersonalAccessTokenRequests struct for request details
  (id, reason, owner, repository selection, permissions, timestamps, token info)
- Add ListFineGrainedPATRequestOptions with TokenID filter and embedded
  ListFineGrainedPATOptions (sort, direction, owner, repository, permission, last_used_before/after, pagination)
- Add OrganizationsService.ListFineGrainedPersonalAccessTokenRequests
- Add addListFineGrainedPATRequestOptions helper for owner[] and token_id[] query parameters

API docs: https://docs.github.com/en/rest/orgs/personal-access-tokens#list-requests-to-access-organization-resources-with-fine-grained-personal-access-tokens

Fixes: #4021

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -572,15 +572,12 @@ func TestNewRequest_invalidJSON(t *testing.T) {
 	c := NewClient(nil)
 
 	type T struct {
-		A map[any]any
+		F func()
 	}
 	_, err := c.NewRequest("GET", ".", &T{})
 
 	if err == nil {
 		t.Error("Expected error to be returned.")
-	}
-	if !errors.As(err, new(*json.UnsupportedTypeError)) {
-		t.Errorf("Expected a JSON error; got %#v.", err)
 	}
 }
 
--- a/github/repos_hooks_deliveries_test.go
+++ b/github/repos_hooks_deliveries_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -261,7 +262,7 @@ func TestHookDelivery_ParsePayload_invalidPayload(t *testing.T) {
 	}
 
 	_, err := d.ParseRequestPayload()
-	if err == nil || err.Error() != "json: cannot unmarshal string into Go struct field CheckRun.check_run.id of type int64" {
+	if err == nil || !strings.Contains(err.Error(), "json: cannot unmarshal") || !strings.Contains(err.Error(), "check_run.id") {
 		t.Errorf("unexpected error: %v", err)
 	}
 }
--- a/github/security_advisories_test.go
+++ b/github/security_advisories_test.go
@@ -590,7 +590,7 @@ func TestSecurityAdvisoriesService_ListRepositorySecurityAdvisoriesForOrg_Unmars
 	advisories, resp, err := client.SecurityAdvisories.ListRepositorySecurityAdvisoriesForOrg(ctx, "o", nil)
 	if err == nil {
 		t.Error("Expected unmarshal error")
-	} else if !strings.Contains(err.Error(), "json: cannot unmarshal number into Go struct field SecurityAdvisory.ghsa_id of type string") {
+	} else if !strings.Contains(err.Error(), "json: cannot unmarshal") || !strings.Contains(err.Error(), "ghsa_id") {
 		t.Errorf("ListRepositorySecurityAdvisoriesForOrg returned unexpected error: %v", err)
 	}
 	if got, want := resp.Response.StatusCode, http.StatusOK; got != want {
@@ -721,7 +721,7 @@ func TestSecurityAdvisoriesService_ListRepositorySecurityAdvisories_UnmarshalErr
 	advisories, resp, err := client.SecurityAdvisories.ListRepositorySecurityAdvisories(ctx, "o", "r", nil)
 	if err == nil {
 		t.Error("Expected unmarshal error")
-	} else if !strings.Contains(err.Error(), "json: cannot unmarshal number into Go struct field SecurityAdvisory.ghsa_id of type string") {
+	} else if !strings.Contains(err.Error(), "json: cannot unmarshal") || !strings.Contains(err.Error(), "ghsa_id") {
 		t.Errorf("ListRepositorySecurityAdvisories returned unexpected error: %v", err)
 	}
 	if got, want := resp.Response.StatusCode, http.StatusOK; got != want {


github/orgs_personal_access_tokens.go
```diff
@@ -115,6 +116,84 @@ func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.C
 	return pats, resp, nil
 }
 
+// FineGrainedPersonalAccessTokenRequests represents the details of a request to access organization resources via a fine-grained personal access token.
+type FineGrainedPersonalAccessTokenRequests struct {
```
alexandear: Why `Requests` and not `Request`?

github/orgs_personal_access_tokens.go
```diff
@@ -115,6 +116,84 @@ func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.C
...
+	// Unique identifier of the request for access via fine-grained personal access token.
+	ID *int64 `json:"id"`
```
alexandear: Why not `ID int64`? Is there a reason you used `*` for every field in the struct?

github/orgs_personal_access_tokens.go
AbhishekAg: API is `personal-access-token-requests` hence I used requests instead of `request`

github/orgs_personal_access_tokens.go
AbhishekAg: I followed https://github.com/google/go-github/blob/master/github/orgs_personal_access_tokens.go#L22

AbhishekAg: I can certainly update my codepiece to not to use pointer in `ID` and `TokenID`. The change in existing API "PersonalAccessToken" may break the consumer of this.

AbhishekAg:
@gmlewis Can you please re-run the pipeline? linter error is due to 503
```
validating openapi_operations.yaml
unexpected status code: 503 Service Unavailable
failed validating openapi_operations.yaml
validating generated files
```

github/orgs_personal_access_tokens.go
gmlewis: According to our README.md:

> ### Creating and Updating Resources ###
> 
> All structs for GitHub resources use pointer values for all non-repeated fields.
> This allows distinguishing between unset fields and those set to a zero-value.

Since `FineGrainedPersonalAccessTokenRequests` represents a GitHub resource that is returned from a method, it does make sense to use pointers and `omitempty` as a standard practice throughout this repo. That way, GitHub can remove fields during some endpoint calls and it will be obvious when a field is not provided.

AbhishekAg:
> Thank you, @AbhishekAg! Once you add `omitempty` to the pointer fields in `FineGrainedPersonalAccessTokenRequests` then I think we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.
> 
> cc: @stevehipwell - @alexandear - @zyfy29 - @Not-Dhananjay-Mishra

Done.

github/orgs_personal_access_tokens.go
stevehipwell:
```suggestion
type FineGrainedPersonalAccessTokenRequest struct {
```

The API may be called _"personal-access-token-requests"_ but it returns an array(and GitHub's naming isn't exactly well thought through).

CC @alexandear

stevehipwell (changes requested):

github/orgs_personal_access_tokens.go
AbhishekAg: Actually it is a good point. Since return is an array, the structure should be named without pluralization

github/orgs_personal_access_tokens.go
```diff
@@ -115,6 +116,84 @@ func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.C
 	return pats, resp, nil
 }
 
+// FineGrainedPersonalAccessTokenRequest represents the details of a request to access organization resources via a fine-grained personal access token.
+type FineGrainedPersonalAccessTokenRequest struct {
```
stevehipwell, line 123: The response schema for this endpoint has all of the struct fields as required so is there a reason we're using pointers?

github/orgs_personal_access_tokens.go
```diff
@@ -115,6 +116,84 @@ func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.C
...
+		return nil, nil, err
+	}
+
+	req, err := s.client.NewRequest("GET", u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var pats []*FineGrainedPersonalAccessTokenRequest
+
```
alexandear: Let's remove this empty line:
```suggestion
```

github/orgs_personal_access_tokens.go
gmlewis: @stevehipwell - @alexandear pointed this out above, and I reminded everyone about our README.md which says that for responses representing GitHub resources, we use pointers and `omitempty` so that as users we can tell when GitHub does not send a field.  Because historically, GitHub will sometimes send fields and sometimes not send fields in a response.  

So @AbhishekAg originally had values without `omitempty` and I asked him to change them to pointers and `omitempty` because this is a returned response from GitHub representing a GitHub resource. This is how we treat the rest of the repo and I feel we should be consistent.

Thoughts?

github/orgs_personal_access_tokens.go
AbhishekAg: I made the changes by renaming to `FineGrainedPersonalAccessTokenRequest`

github/orgs_personal_access_tokens.go
AbhishekAg: I think all fields are not mandatory. Example: reason may be null when no reason is provided while requesting for fine grained PAT.
That's why we have added `omitempty` in every field

AbhishekAg:
Pipeline failure due to 
```
--- FAIL: TestRepositoriesService_RemoveCollaborator_invalidUser (0.00s)
    testing.go:1712: race detected during execution of test
--- FAIL: TestRepositoriesService_UploadReleaseAssetFromRelease_AbsoluteTemplate (0.04s)
    testing.go:1712: race detected during execution of test
```

gmlewis:
> Pipeline failure due to
> 
> ```
> --- FAIL: TestRepositoriesService_RemoveCollaborator_invalidUser (0.00s)
>     testing.go:1712: race detected during execution of test
> --- FAIL: TestRepositoriesService_UploadReleaseAssetFromRelease_AbsoluteTemplate (0.04s)
>     testing.go:1712: race detected during execution of test
> ```

Unfortunately, we appear to be having some flaky tests, which @alexandear also noticed and filed an issue: #3987.

I'm not sure at what point this started getting flaky, but it feels like it is pretty recent.

I reran the failed job.

github/orgs_personal_access_tokens.go
alexandear: According to the [response schema](https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#list-requests-to-access-organization-resources-with-fine-grained-personal-access-tokens), all fields are mandatory:

```yaml
    "required": [
      "id",
      "reason",
      "owner",
      "repository_selection",
      "repositories_url",
      "permissions",
      "created_at",
      "token_id",
      "token_name",
      "token_expired",
      "token_expires_at",
      "token_last_used_at"
    ]
  }
}
```

As I recall, we try to strictly follow the schema, and sometimes, if there are reported bugs, we can work around them using `omitempty`. If GitHub doesn’t send some fields, that’s a mistake on GitHub’s side.

I agree with @stevehipwell and suggest removing `omitempty` and pointers from all fields.

github/orgs_personal_access_tokens.go
```diff
@@ -115,6 +116,84 @@ func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.C
...
+	ListFineGrainedPATOptions
+}
+
+// ListFineGrainedPersonalAccessTokenRequests lists requests to access organization resources via fine-grained personal access tokens.
+// Only GitHub Apps can call this API, using the `Personal access tokens` organization permissions (read).
+//
+// GitHub API docs: https://docs.github.com/rest/orgs/personal-access-tokens#list-requests-to-access-organization-resources-with-fine-grained-personal-access-tokens
+//
+//meta:operation GET /orgs/{org}/personal-access-token-requests
+func (s *OrganizationsService) ListFineGrainedPersonalAccessTokenRequests(ctx context.Context, org string, opts *ListFineGrainedPATRequestOptions) ([]*FineGrainedPersonalAccessTokenRequest, *Response, error) {
```
alexandear: options name = method name + "Options" suffix

`s` is missing:

```suggestion
func (s *OrganizationsService) ListFineGrainedPersonalAccessTokenRequests(ctx context.Context, org string, opts *ListFineGrainedPATRequestsOptions) ([]*FineGrainedPersonalAccessTokenRequest, *Response, error) {
```

github/orgs_personal_access_tokens.go
```diff
@@ -115,6 +116,84 @@ func (s *OrganizationsService) ListFineGrainedPersonalAccessTokens(ctx context.C
...
+	// Date and time when the associated fine-grained personal access token was last used for authentication.
+	TokenLastUsedAt *Timestamp `json:"token_last_used_at,omitempty"`
+}
+
+// ListFineGrainedPATRequestOptions specifies optional parameters to ListFineGrainedPersonalAccessTokenRequests.
+type ListFineGrainedPATRequestOptions struct {
+	// TokenID filters results by the given fine-grained personal access token IDs.
+	TokenID []int64 `url:"-"`
+
+	ListFineGrainedPATOptions
```
alexandear: `ListFineGrainedPATOptions` also contains `TokenID`: according to the doc
https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#list-fine-grained-personal-access-tokens-with-access-to-organization-resources

So, let's add `TokenID` to the `ListFineGrainedPATOptions`, update `addListFineGrainedPATOptions`, and use `ListFineGrainedPATOptions` here instead of `ListFineGrainedPATRequestOptions`.

gmlewis:
> If GitHub doesn’t send some fields, that’s a mistake on GitHub’s side.

Yes, and over the last 10+ years, we've seen that a fair amount... mostly in the early days (hence the stance in our README.md).
Granted that GitHub was not even providing response schemas in the early days.

OK, so your argument is that now that GitHub _IS_ providing response schemas, and we are building a new endpoint based upon that schema, that we should initially attempt to honor it and fix bugs as they occur.

I totally understand that argument.

As a maintainer, we wish that everything could be 100% consistent and cut-and-dry and make perfect sense.

So my next thought is... now that GitHub response schemas are prolific, do we now go back and clean up every struct this repo generates for GitHub resources and make the fields match the schemas?!?

Are we opening up an enormous can of worms?

Thoughts?

alexandear:
> So my next thought is... now that GitHub response schemas are prolific, do we now go back and clean up every struct this repo generates for GitHub resources and make the fields match the schemas?!?
> 
> Are we opening up an enormous can of worms?

We can do this in the future, but we need to make sure it won't break anything. So we should limit these changes to endpoints that have integration tests. Or maybe we need to add more integration tests, but it's hard for some endpoints (e.g., enterprise or org-related).

gmlewis:
OK, @AbhishekAg, let's run with the recommendations from @stevehipwell and @alexandear.
I apologize for steering you in the wrong direction.

github/orgs_personal_access_tokens.go
AbhishekAg: I agree.

github/orgs_personal_access_tokens.go
AbhishekAg: Used `ListFineGrainedPATOptions`

github/orgs_personal_access_tokens.go
AbhishekAg: Completed the changes

AbhishekAg:
I completed all the requested changes.

github/orgs_personal_access_tokens.go
AbhishekAg: Done

AbhishekAg:
@gmlewis can we please merge these changes if no more comments? All reviewers have approved the changes.

gmlewis:
Thank you, @AbhishekAg, @stevehipwell, and @alexandear!
Merging.

PR #4028, opened 2026-02-21
feat!: Support workflow dispatch run details in response

### LINKED ISSUE #4027: Workflow dispatch API now returns run IDs
GitHub [announced](https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/), When you trigger a workflow using the [GitHub Actions workflow dispatch API endpoint](https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event), you’ll now have the option to receive metadata in the response that helps you map your request to the corresponding workflow run.

Affected method - 
https://github.com/google/go-github/blob/c02c318a0f8f294a34e782be7111753d3b032b23/github/actions_workflows.go#L189-L218

LinaKACI-pro:
BREAKING CHANGE: `CreateWorkflowDispatchEventByID` and `CreateWorkflowDispatchEventByFileName` now return `*WorkflowDispatchRunDetails`.

## Summary

Resolves #4027

- Added `ReturnRunDetails` field to `CreateWorkflowDispatchEventRequest` to support the new `return_run_details` API parameter
- Added `CreateWorkflowDispatchEventResponse` struct with `WorkflowRunID`, `RunURL`, and `HTMLURL` fields
- Updated `CreateWorkflowDispatchEventByID` and `CreateWorkflowDispatchEventByFileName` to return `(*CreateWorkflowDispatchEventResponse, *Response, error)` instead of `(*Response, error)`
- Updated max inputs comment from 10 to 25 to match current API docs
- Added tests for both `ReturnRunDetails: true` (200 OK with body) and omitted (204 No Content) cases

## Breaking change

The method signatures for `CreateWorkflowDispatchEventByID` and `CreateWorkflowDispatchEventByFileName` changed from returning `(*Response, error)` to `(*CreateWorkflowDispatchEventResponse, *Response, error)`.

## API reference

https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event

## Manual test

Tested against a real repository with a `workflow_dispatch` workflow using the following program:
To reproduce, create a test program in a separate directory and use `go.mod` `replace` to point to the local changes:

```bash
mkdir /tmp/test_dispatch && cd /tmp/test_dispatch
go mod init test_dispatch
go mod edit -replace github.com/google/go-github/v83=/path/to/your/local/go-github
go mod tidy
```

Then run:
```bash
go run main.go -owner <owner> -repo <repo> -workflow <workflow.yml> -token <token>
```

<details>
<summary>Test program</summary>

```go
package main

import (
	"bytes"
	"context"
	"encoding/json"
	"flag"
	"fmt"
	"io"
	"net/http"
	"os"

	"github.com/google/go-github/v83/github"
)

type loggingTransport struct{}

func (t *loggingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	fmt.Printf("\n--- REQUEST ---\n")
	fmt.Printf("%s %s\n", req.Method, req.URL)
	for k, v := range req.Header {
		fmt.Printf("%s: %s\n", k, v)
	}
	if req.Body != nil {
		bodyBytes, _ := io.ReadAll(req.Body)
		req.Body.Close()
		var pretty bytes.Buffer
		if json.Indent(&pretty, bodyBytes, "", "  ") == nil {
			fmt.Printf("Body:\n%s\n", pretty.String())
		} else {
			fmt.Printf("Body: %s\n", string(bodyBytes))
		}
		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
	}

	resp, err := http.DefaultTransport.RoundTrip(req)
	if err != nil {
		return resp, err
	}

	fmt.Printf("\n--- RESPONSE ---\n")
	fmt.Printf("Status: %s\n", resp.Status)
	if resp.Body != nil {
		bodyBytes, _ := io.ReadAll(resp.Body)
		resp.Body.Close()
		if len(bodyBytes) > 0 {
			var pretty bytes.Buffer
			if json.Indent(&pretty, bodyBytes, "", "  ") == nil {
				fmt.Printf("Body:\n%s\n", pretty.String())
			} else {
				fmt.Printf("Body: %s\n", string(bodyBytes))
			}
		} else {
			fmt.Println("Body: (empty)")
		}
		resp.Body = io.NopCloser(bytes.NewReader(bodyBytes))
	}
	fmt.Println()

	return resp, nil
}

func main() {
	owner := flag.String("owner", "", "Repository owner")
	repo := flag.String("repo", "", "Repository name")
	workflow := flag.String("workflow", "", "Workflow filename")
	ref := flag.String("ref", "main", "Git ref (default: main)")
	token := flag.String("token", "", "GitHub token (or set GITHUB_TOKEN)")
	flag.Parse()

	if *owner == "" || *repo == "" || *workflow == "" {
		fmt.Println("Usage: test_dispatch -owner <owner> -repo <repo> -workflow <file> [-ref <branch>] [-token <token>]")
		os.Exit(1)
	}

	t := *token
	if t == "" {
		t = os.Getenv("GITHUB_TOKEN")
	}
	if t == "" {
		fmt.Println("Error: provide -token flag or set GITHUB_TOKEN env var")
		os.Exit(1)
	}

	httpClient := &http.Client{Transport: &loggingTransport{}}
	client := github.NewClient(httpClient).WithAuthToken(t)

	fmt.Println("=== Test 1: ReturnRunDetails = true ===")
	event := github.CreateWorkflowDispatchEventRequest{
		Ref:              *ref,
		ReturnRunDetails: github.Ptr(true),
		Inputs:           map[string]any{"name": "go-github-test"},
	}

	resp, _, err := client.Actions.CreateWorkflowDispatchEventByFileName(
		context.Background(), *owner, *repo, *workflow, event,
	)
	if err != nil {
		os.Exit(1)
	}

	fmt.Printf("\nWorkflow Run ID: %d\n", resp.GetWorkflowRunID())
	fmt.Printf("Run URL: %s\n", resp.GetRunURL())
	fmt.Printf("HTML URL: %s\n", resp.GetHTMLURL())

	fmt.Println("\n=== Test 2: ReturnRunDetails omitted ===")
	event2 := github.CreateWorkflowDispatchEventRequest{
		Ref:    *ref,
		Inputs: map[string]any{"name": "go-github-test-no-details"},
	}

	resp2, _, err := client.Actions.CreateWorkflowDispatchEventByFileName(
		context.Background(), *owner, *repo, *workflow, event2,
	)
	if err != nil {
		os.Exit(1)
	}

	fmt.Printf("\nWorkflowRunID is nil: %v\n", resp2.WorkflowRunID == nil)
	fmt.Printf("RunURL is nil: %v\n", resp2.RunURL == nil)
	fmt.Printf("HTMLURL is nil: %v\n", resp2.HTMLURL == nil)

	fmt.Println("\nAll tests passed!")
}
```

</details>

### DIFF
--- a/github/actions_workflows_test.go
+++ b/github/actions_workflows_test.go
@@ -260,7 +260,7 @@ func TestActionsService_CreateWorkflowDispatchEventByID(t *testing.T) {
 		t.Errorf("Actions.CreateWorkflowDispatchEventByID returned error: %v", err)
 	}
 
-	want := &CreateWorkflowDispatchEventResponse{
+	want := &WorkflowDispatchRunDetails{
 		WorkflowRunID: Ptr(int64(1)),
 		RunURL:        Ptr("https://api.github.com/repos/o/r/actions/runs/1"),
 		HTMLURL:       Ptr("https://github.com/o/r/actions/runs/1"),
@@ -321,7 +321,7 @@ func TestActionsService_CreateWorkflowDispatchEventByFileName(t *testing.T) {
 		t.Errorf("Actions.CreateWorkflowDispatchEventByFileName returned error: %v", err)
 	}
 
-	want := &CreateWorkflowDispatchEventResponse{
+	want := &WorkflowDispatchRunDetails{
 		WorkflowRunID: Ptr(int64(1)),
 		RunURL:        Ptr("https://api.github.com/repos/o/r/actions/runs/1"),
 		HTMLURL:       Ptr("https://github.com/o/r/actions/runs/1"),
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -7182,30 +7182,6 @@ func (c *CreateWorkflowDispatchEventRequest) GetReturnRunDetails() bool {
 	return *c.ReturnRunDetails
 }
 
-// GetHTMLURL returns the HTMLURL field if it's non-nil, zero value otherwise.
-func (c *CreateWorkflowDispatchEventResponse) GetHTMLURL() string {
-	if c == nil || c.HTMLURL == nil {
-		return ""
-	}
-	return *c.HTMLURL
-}
-
-// GetRunURL returns the RunURL field if it's non-nil, zero value otherwise.
-func (c *CreateWorkflowDispatchEventResponse) GetRunURL() string {
-	if c == nil || c.RunURL == nil {
-		return ""
-	}
-	return *c.RunURL
-}
-
-// GetWorkflowRunID returns the WorkflowRunID field if it's non-nil, zero value otherwise.
-func (c *CreateWorkflowDispatchEventResponse) GetWorkflowRunID() int64 {
-	if c == nil || c.WorkflowRunID == nil {
-		return 0
-	}
-	return *c.WorkflowRunID
-}
-
 // GetCreated returns the Created field if it's non-nil, zero value otherwise.
 func (c *CreationInfo) GetCreated() Timestamp {
 	if c == nil || c.Created == nil {
@@ -31716,6 +31692,30 @@ func (w *WorkflowDispatchEvent) GetWorkflow() string {
 		return ""
 	}
 	return *w.Workflow
+}
+
+// GetHTMLURL returns the HTMLURL field if it's non-nil, zero value otherwise.
+func (w *WorkflowDispatchRunDetails) GetHTMLURL() string {
+	if w == nil || w.HTMLURL == nil {
+		return ""
+	}
+	return *w.HTMLURL
+}
+
+// GetRunURL returns the RunURL field if it's non-nil, zero value otherwise.
+func (w *WorkflowDispatchRunDetails) GetRunURL() string {
+	if w == nil || w.RunURL == nil {
+		return ""
+	}
+	return *w.RunURL
+}
+
+// GetWorkflowRunID returns the WorkflowRunID field if it's non-nil, zero value otherwise.
+func (w *WorkflowDispatchRunDetails) GetWorkflowRunID() int64 {
+	if w == nil || w.WorkflowRunID == nil {
+		return 0
+	}
+	return *w.WorkflowRunID
 }
 
 // GetCheckRunURL returns the CheckRunURL field if it's non-nil, zero value otherwise.
--- a/github/github-accessors_test.go
+++ b/github/github-accessors_test.go
@@ -9406,39 +9406,6 @@ func TestCreateWorkflowDispatchEventRequest_GetReturnRunDetails(tt *testing.T) {
 	c.GetReturnRunDetails()
 }
 
-func TestCreateWorkflowDispatchEventResponse_GetHTMLURL(tt *testing.T) {
-	tt.Parallel()
-	var zeroValue string
-	c := &CreateWorkflowDispatchEventResponse{HTMLURL: &zeroValue}
-	c.GetHTMLURL()
-	c = &CreateWorkflowDispatchEventResponse{}
-	c.GetHTMLURL()
-	c = nil
-	c.GetHTMLURL()
-}
-
-func TestCreateWorkflowDispatchEventResponse_GetRunURL(tt *testing.T) {
-	tt.Parallel()
-	var zeroValue string
-	c := &CreateWorkflowDispatchEventResponse{RunURL: &zeroValue}
-	c.GetRunURL()
-	c = &CreateWorkflowDispatchEventResponse{}
-	c.GetRunURL()
-	c = nil
-	c.GetRunURL()
-}
-
-func TestCreateWorkflowDispatchEventResponse_GetWorkflowRunID(tt *testing.T) {
-	tt.Parallel()
-	var zeroValue int64
-	c := &CreateWorkflowDispatchEventResponse{WorkflowRunID: &zeroValue}
-	c.GetWorkflowRunID()
-	c = &CreateWorkflowDispatchEventResponse{}
-	c.GetWorkflowRunID()
-	c = nil
-	c.GetWorkflowRunID()
-}
-
 func TestCreationInfo_GetCreated(tt *testing.T) {
 	tt.Parallel()
 	var zeroValue Timestamp
@@ -40954,6 +40921,39 @@ func TestWorkflowDispatchEvent_GetWorkflow(tt *testing.T) {
 	w.GetWorkflow()
 	w = nil
 	w.GetWorkflow()
+}
+
+func TestWorkflowDispatchRunDetails_GetHTMLURL(tt *testing.T) {
+	tt.Parallel()
+	var zeroValue string
+	w := &WorkflowDispatchRunDetails{HTMLURL: &zeroValue}
+	w.GetHTMLURL()
+	w = &WorkflowDispatchRunDetails{}
+	w.GetHTMLURL()
+	w = nil
+	w.GetHTMLURL()
+}
+
+func TestWorkflowDispatchRunDetails_GetRunURL(tt *testing.T) {
+	tt.Parallel()
+	var zeroValue string
+	w := &WorkflowDispatchRunDetails{RunURL: &zeroValue}
+	w.GetRunURL()
+	w = &WorkflowDispatchRunDetails{}
+	w.GetRunURL()
+	w = nil
+	w.GetRunURL()
+}
+
+func TestWorkflowDispatchRunDetails_GetWorkflowRunID(tt *testing.T) {
+	tt.Parallel()
+	var zeroValue int64
+	w := &WorkflowDispatchRunDetails{WorkflowRunID: &zeroValue}
+	w.GetWorkflowRunID()
+	w = &WorkflowDispatchRunDetails{}
+	w.GetWorkflowRunID()
+	w = nil
+	w.GetWorkflowRunID()
 }
 
 func TestWorkflowJob_GetCheckRunURL(tt *testing.T) {


rmacklin:
Nice, thanks for working on this - I saw the GitHub Changelog post about this yesterday and was planning to contribute this today but you beat me to it! I'm excited to remove our project's workaround for getting the run URL once this is merged and released!

github/actions_workflows.go
```diff
@@ -202,19 +213,25 @@ func (s *ActionsService) CreateWorkflowDispatchEventByID(ctx context.Context, ow
...
 	}
 
-	return s.client.Do(ctx, req, nil)
+	dispatchResponse := new(CreateWorkflowDispatchEventResponse)
+	resp, err := s.client.Do(ctx, req, dispatchResponse)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return dispatchResponse, resp, nil
```
stevehipwell:
```suggestion
	if dispatchResponse.WorkflowRunID != nil {	
		return dispatchResponse, resp, nil
	}
	
	return nil, resp, nil
```
Do we not want to only return the object if it's been populated?

gmlewis: So if there had been no error on line 230 above, then _something_ was returned, and we can expose whatever was returned. I think we use this pattern throughout this repo... very little attempt is made (for the majority of endpoints) to check the usefulness of the response, but we simply pass it along to the user. I think this is a valid implementation.

Thoughts, @stevehipwell?

LinaKACI-pro: Thanks for the suggestion! I initially went with returning an empty `&CreateWorkflowDispatchEventResponse{}` so callers always get a non-nil value and don't need to nil-check before using it. But `nil` is a clearer signal that no run details were returned. Happy to change it if you prefer that approach.

stevehipwell: CC @gmlewis

gmlewis: OK, hmmm... I'm still thinking that the roundtrip at this point was actually successful...
and if no `WorkflowRunID` was returned, then _SOMETHING_ seems wrong, and is a strong signal that GitHub tech support should be contacted.  Whereas if the roundtrip was successful yet we returned `nil`, the user would know absolutely _nothing_ and would not even be able to see what was _actually_ returned.

From my experience working with this repo, there have been many times where people have posted issues saying "Why am I not seeing what I'm expecting to be seeing when I call endpoint X?" At which point, I usually have to point them to https://github.com/gmlewis/go-httpdebug so that it is much easier to debug what was returned and also try out calling the GitHub v3 API directly using `curl` and avoid this client library entirely to really understand what is happening.

So my biggest concern is a maintenance and support issue...

If we _were_ to hide what was returned because at this moment in time we think that the response doesn't look right, and GitHub changes something on their end, then the user would have no visibility and we contributors to this repo will be reading new issues being opened as to why this endpoint is masking what _was_ returned.

So I think I still feel strongly about this.

Thoughts?

alexandear: The object is not populated - it's a GitHub issue. I propose leaving it as is, without checking `dispatchResponse.WorkflowRunID` for `nil`. This would be consistent with what we already have.

stevehipwell: The API returns nothing unless a switch is set (the new feature being supported). So if the switch isn't set I wouldn't want to get an empty struct, I'd want `nil`. This logic could be driven by the input rather than the response. Otherwise IMHO this should be a net new function.

gmlewis: > The API returns nothing unless a switch is set (the new feature being supported). So if the switch isn't set I wouldn't want to get an empty struct, I'd want `nil`. This logic could be driven by the input rather than the response. Otherwise IMHO this should be a net new function.

Can we try this out and see what exactly is returned in this case?

alexandear: @stevehipwell I see your point. In that case, maybe we can explicitly check for `ReturnRunDetails`:

```go
	if event.ReturnRunDetails == nil || !*event.ReturnRunDetails {
		return nil, resp, nil
	}

	return dispatchResponse, resp, nil
```

gmlewis: If it really is an empty struct, then sure... can we verify this?

LinaKACI-pro: I tested the code against the real API (with the code i wrote on the description in **Manual Test**):

With `return_run_details: true` -> `200 OK` with a populated body:
```json
{
  "workflow_run_id": ***,
  "run_url": "https://api.github.com/repos/.../actions/runs/***",
  "html_url": "https://github.com/.../actions/runs/***"
}
```

With `return_run_details:false` -> `204 No Content` with an empty body (`len(body) == 0`).

So the API returns zero bytes.

stevehipwell: FYI [announcement](https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/) confirming that this is the correct behaviour.

LinaKACI-pro: Thanks! So should I go with @alexandear's input-based check?

```go
if event.ReturnRunDetails == nil || !*event.ReturnRunDetails {
    return nil, resp, nil
}

return dispatchResponse, resp, nil
```

gmlewis: Can you please do an experiment for me (I'm on my phone)... Instead of: 

```go
x := new(...)
... Unmarshal(..., x)
```

Could you please try: 

```go
var x *Thing
... Unmarshal(, &x)
```

and see if `x` is still  `nil` in the case of the 204?

I prefer this style anyway. 😁

LinaKACI-pro: It's still `nil` with : `var dispatchResponse *CreateWorkflowDispatchEventResponse` :

It's passed as `&dispatchResponse` to [`Do` method](https://github.com/google/go-github/blob/master/github/github.go#L1056C1-L1077C2), which calls `json.NewDecoder(resp.Body).Decode(v)`

When the API returns `204 No Content`, the body is empty so `Decode` returns `io.EOF`, the pointer is never written to, so `dispatchResponse` stays `nil`.

```go
func (c *Client) Do(ctx context.Context, req *http.Request, v any) (*Response, error) {
	resp, err := c.BareDo(ctx, req)
	if err != nil {
		return resp, err
	}
	defer resp.Body.Close()

	switch v := v.(type) {
	case nil:
	case io.Writer:
		_, err = io.Copy(v, resp.Body)
	default:
		if decErr == io.EOF {
		}
		if decErr != nil {
			err = decErr
		}
	}
	return resp, err
}
```

#### manual test and output : 

```go
  resp1, httpResp1, _ := client.Actions.CreateWorkflowDispatchEventByFileName(ctx, owner, repo, workflow,
      github.CreateWorkflowDispatchEventRequest{Ref: "main"})
  fmt.Printf("without ReturnRunDetails:\n")
  fmt.Printf("  status: %s, content-type: %q, content-length: %d\n",
      httpResp1.Status, httpResp1.Header.Get("Content-Type"), httpResp1.ContentLength)
  fmt.Printf("  dispatchResponse: %v\n", resp1)

  resp2, httpResp2, _ := client.Actions.CreateWorkflowDispatchEventByFileName(ctx, owner, repo, workflow,
      github.CreateWorkflowDispatchEventRequest{Ref: "main", ReturnRunDetails: github.Ptr(true)})
  fmt.Printf("with ReturnRunDetails=true:\n")
  fmt.Printf("  status: %s, content-type: %q, content-length: %d\n",
      httpResp2.Status, httpResp2.Header.Get("Content-Type"), httpResp2.ContentLength)
  fmt.Printf("  dispatchResponse: {WorkflowRunID: %d, RunURL: %s, HTMLURL: %s}\n",
      *resp2.WorkflowRunID, *resp2.RunURL, *resp2.HTMLURL)
```

```sh
  Output:
  without ReturnRunDetails:
    status: 204 No Content, content-type: "", content-length: 0
    dispatchResponse: <nil>
  with ReturnRunDetails=true:
    status: 200 OK, content-type: "application/json; charset=utf-8", content-length: -1
    dispatchResponse: {...}
```

I hope i correctly understood the context and hope this helps :)

gmlewis: > I prefer this style anyway. 😁

WHUPS!!!  I totally messed that up on my phone.  I meant instead of:

```go
x := new(...)
... Do(..., x)
```

to please try: 

```go
var x *Thing
... Do(, &x)
```

and see if you get `nil` for the 204.  Now I'm going to go see what you wrote. 😅

gmlewis: > It's passed as `&dispatchResponse` to [`Do` method](https://github.com/google/go-github/blob/master/github/github.go#L1056C1-L1077C2), which calls `json.NewDecoder(resp.Body).Decode(v)`
> 
> When the API returns `204 No Content`, the body is empty so `Decode` returns `io.EOF`, the pointer is never written to, so `dispatchResponse` stays `nil`.

COOL!  You figured out what I wrote!  Now let me see if you changed the code... because, if I understand correctly, we don't need a special test if this solves the problem with a `204` response because @stevehipwell will get a `nil`!!!  :joy:

github/actions_workflows.go
```diff
@@ -202,19 +213,25 @@ func (s *ActionsService) CreateWorkflowDispatchEventByID(ctx context.Context, ow
...
+func (s *ActionsService) createWorkflowDispatchEvent(ctx context.Context, url string, event *CreateWorkflowDispatchEventRequest) (*CreateWorkflowDispatchEventResponse, *Response, error) {
 	req, err := s.client.NewRequest("POST", url, event)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return s.client.Do(ctx, req, nil)
+	dispatchResponse := new(CreateWorkflowDispatchEventResponse)
+	resp, err := s.client.Do(ctx, req, dispatchResponse)
```
gmlewis: OK, so here is my suggestion - @stevehipwell - thoughts?


```suggestion
	var dispatchResponse *CreateWorkflowDispatchEventResponse
	resp, err := s.client.Do(ctx, req, &dispatchResponse)
```

LinaKACI-pro: yep that's what i did, want me to commit ?

gmlewis: Yes, please!  I _MUCH_ prefer that solution, as it is more idiomatic Go in my opinion, and gives us the result we all want!!!  :joy:

LinaKACI-pro: Yes :smile: and that's more accurate! I just commited! @gmlewis

github/actions_workflows.go
```diff
@@ -51,9 +51,20 @@ type CreateWorkflowDispatchEventRequest struct {
...
 	// Default: Any default properties configured in the workflow file will be used when `inputs` are omitted.
 	Inputs map[string]any `json:"inputs,omitempty"`
+	// ReturnRunDetails specifies whether the response should include
+	// the workflow run ID and URLs.
+	ReturnRunDetails *bool `json:"return_run_details,omitempty"`
+}
+
+// CreateWorkflowDispatchEventResponse represents the response from creating
+// a workflow dispatch event when ReturnRunDetails is set to true.
+type CreateWorkflowDispatchEventResponse struct {
```
Not-Dhananjay-Mishra: Would `WorkflowDispatchRunDetails` be a better name?
I believe it aligns well with `ReturnRunDetails`

stevehipwell (approved):
LGTM

github/actions_workflows.go
gmlewis: Yes, I like that. Thanks, @Not-Dhananjay-Mishra!
@LinaKACI-pro - could you please make this change before we merge?

LinaKACI-pro: Sure, done 👍

gmlewis:
No need to force push, just FYI.

gmlewis:
Thank you, everyone! 
Merging.

LinaKACI-pro:
Thank you everyone!🚀

rmacklin:
@gmlewis Just curious - do you know when you'll cut a new release that includes this?

gmlewis:
> @gmlewis Just curious - do you know when you'll cut a new release that includes this?

We typically make releases about once per month (which would mean the next one is in about 3 weeks), but if people ask for a release earlier than that, we can usually accomodate such requests. Right now we have a bunch of PRs in-flight and it might be nice to get those in... feel free to help with code reviews if you have time, and please let me know if you need a release earlier than in 3 weeks.

rmacklin:
Sounds good! I'm definitely eager for this one but certainly understand waiting to get some other in-flight work across the finish line. Thanks for the answer! (and for all the hard work maintaining this library!)

sunggun-yu:
Yeah, I'm eager for this feature! It will make my proj way easier. Looking forward to it!!!

gmlewis:
> Yeah, I'm eager for this feature! It will make my proj way easier. Looking forward to it!!!

OK, two requests sound good enough to me. I'll work on a release today or tomorrow after I take a look at what PRs are currently in-flight and hopefully close-to-merging.

gmlewis:
@rmacklin and @sunggun-yu - this is now released here:
https://github.com/google/go-github/releases/tag/v84.0.0

rmacklin:
Awesome, thank you!!

sunggun-yu:
Thank you!

github/actions_workflows.go
```diff
@@ -51,9 +51,20 @@ type CreateWorkflowDispatchEventRequest struct {
 	// Ref is required when creating a workflow dispatch event.
 	Ref string `json:"ref"`
 	// Inputs represents input keys and values configured in the workflow file.
-	// The maximum number of properties is 10.
+	// The maximum number of properties is 25.
 	// Default: Any default properties configured in the workflow file will be used when `inputs` are omitted.
 	Inputs map[string]any `json:"inputs,omitempty"`
+	// ReturnRunDetails specifies whether the response should include
+	// the workflow run ID and URLs.
+	ReturnRunDetails *bool `json:"return_run_details,omitempty"`
```
rmacklin, line 59: PSA: I just learned that in the new 2026-03-10 version of the GitHub API, the `return_run_details` parameter has been removed and the endpoint will always returns `200` with the workflow run details in the response body.

Source: https://docs.github.com/en/rest/about-the-rest-api/breaking-changes?apiVersion=2026-03-10#version-2026-03-10:~:text=Removes%20the%20return_run_details%20parameter.%20The%20endpoint%20now%20always%20returns%20200%20with%20the%20workflow%20run%20details%20in%20the%20response%20body.

I also see there's already a general issue about the new API version: https://github.com/google/go-github/issues/4077 - so mostly just commenting here to add visibility that this is one of the affected methods.

PR #4034, opened 2026-02-24
chore: Use `go:fix inline` for deprecated ptr funcs

alexandear:
`Bool`, `Int`, `Int64`, and `String` were replaced by `Ptr` in #3355. This PR adds `//go:fix inline` directive to these functions, allowing users to replace the deprecated ones across the codebase using the simple command:

```
go fix -inline ./...
```

See https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/inline#hdr-Analyzer_inline


gmlewis (approved):
Thank you, @alexandear!
LGTM.
Merging.

PR #4037, opened 2026-02-25
feat: Add support for team `type` field

### LINKED ISSUE #4033: Add support for team `type` field
The teams API endpoints now return a `type` field to distinguish between organization and enterprise teams (see [announcement](https://github.blog/changelog/2026-02-23-enterprise-team-support-in-organization-apis/)). We need to add support for this.

munlicode:
Adds `type` field to `Team` struct and updates tests for listing and getting team(s).

Note: I updated test to use `assertNoDiff`, because it was printing pointers when `TestCopilotService_ListCopilotSeats` was failing. Since it is not really helpful for debugging, I changed it to print difference instead of just comparing if it is equal.

Resolves #4033

### DIFF
--- a/github/admin.go
+++ b/github/admin.go
@@ -82,7 +82,7 @@ func (m Enterprise) String() string {
 
 // UpdateUserLDAPMapping updates the mapping between a GitHub user and an LDAP user.
 //
-// GitHub API docs: https://docs.github.com/enterprise-server@3.19/rest/enterprise-admin/ldap#update-ldap-mapping-for-a-user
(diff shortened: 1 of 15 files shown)


maishivamhoo123:
Since the API also supports "enterprise", would it make sense to add a test case covering that value as well?

gmlewis:
> Since the API also supports "enterprise", would it make sense to add a test case covering that value as well?

That's optional (since marshaling/unmarshaling is already covered by your previous changes) but feel free to add if you wish for documentation purposes. It's your call.

gmlewis:
Thank you, @alexandear!
Merging.

PR #4042, opened 2026-02-26
chore: Improve `testJSONMarshal`

### LINKED ISSUE #2699: about testJSONMarshal's behavior

alexandear:
This PR rewrites `testJSONMarshal`, so that it properly verifies field marshaling and unmarshaling. Unmarshaling is more complex, as it requires adding multiple options to work around limitations in some of our structs.

I removed tests for structs with only use `url` field tags, since they are unnecessary.

As #3519 has had no activity since April, I decided to complete this work.

Original author: @exageraldo.
Applied review comments from: @stevehipwell

Fixes #2699
Closes #3519

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -163,6 +163,10 @@ func testBody(t *testing.T, r *http.Request, want string) {
 
 // testJSONMarshal tests both JSON marshaling and unmarshaling of a value by comparing
 // the marshaled output with the expected JSON string.
+//
+// This is the recommended function for most use cases.
+// It performs a round-trip test that ensures both marshaling (Go value to JSON)
+// and unmarshaling (JSON to Go value) work correctly and produce semantically equivalent results.
 func testJSONMarshal[T any](t *testing.T, v T, want string, opts ...cmp.Option) {
 	t.Helper()
 
@@ -177,6 +181,9 @@ func testJSONMarshal[T any](t *testing.T, v T, want string, opts ...cmp.Option) 
 //   - White space differences
 //   - Key ordering in objects
 //   - Numeric type differences (e.g., int vs float with same value)
+//
+// In most cases, use testJSONMarshal instead.
+// Only use this function in rare cases where you need to test marshaling behavior in isolation.
 func testJSONMarshalOnly[T any](t *testing.T, v T, want string) {
 	t.Helper()
 
@@ -205,6 +212,9 @@ func testJSONMarshalOnly[T any](t *testing.T, v T, want string) {
 
 // testJSONUnmarshalOnly tests JSON unmarshaling by parsing the JSON string
 // and comparing the result with the expected value.
+//
+// In most cases, use testJSONMarshal instead.
+// Only use this function in rare cases where you need to test unmarshaling behavior in isolation.
 func testJSONUnmarshalOnly[T any](t *testing.T, want T, v string, opts ...cmp.Option) {
 	t.Helper()
 


github/actions_runner_groups_test.go
```diff
@@ -554,7 +554,7 @@ func TestRunnerGroup_Marshal(t *testing.T) {
 		Inherited:                Ptr(true),
 		AllowsPublicRepositories: Ptr(true),
 		RestrictedToWorkflows:    Ptr(false),
-		SelectedWorkflows:        []string{},
+		SelectedWorkflows:        []string(nil),
```
gmlewis: Why not delete this line entirely, since this is effectively its zero-value?
Same here and everywhere that `[]string(nil)` is used in this PR?

github/github_test.go
```diff
@@ -163,28 +163,167 @@ func testBody(t *testing.T, r *http.Request, want string) {
...
+
+// jsonRawMessageComparator returns a cmp.Option that compares json.RawMessage
+// values by their semantic JSON content rather than byte-for-byte equality.
+func jsonRawMessageComparator() cmp.Option {
+	return cmp.Comparer(func(x, y json.RawMessage) bool {
+		if len(x) == 0 && len(y) == 0 {
+			return true
+		}
+		if len(x) == 0 || len(y) == 0 {
+			return false
```
```suggestion
		if len(x) == 0 || len(y) == 0 {
			return len(x) == 0 && len(y) == 0
```

github/meta_test.go
```diff
@@ -58,13 +58,13 @@ func TestAPIMeta_Marshal(t *testing.T) {
...
 		"actions":["a"],
-    "actions_macos":["example.com/1", "example.com/2"],
+		"actions_macos":["example.com/1", "example.com/2"],
 		"dependabot":["d"],
 		"ssh_key_fingerprints":{"a":"f"},
 		"ssh_keys":["k"],
 		"api":["a"],
 		"web":["w"],
-		"domains":{"website":["*.github.com","*.github.dev","*.github.io","*.example.com/assets","*.example.com"],"artifact_attestations":{"trust_domain":"","services":["*.actions.github.com","tuf-repo.github.com","fulcio.github.com","timestamp.github.com"]}}
+		"domains":{"website":["*.github.com","*.github.dev","*.github.io","*.example.com/assets","*.example.com"],"artifact_attestations":{"services":["*.actions.github.com","tuf-repo.github.com","fulcio.github.com","timestamp.github.com"]}}
```
gmlewis: Can we please get rid of these third-party URLs while we're editing this?

github/repos_community_health_test.go
```diff
@@ -134,7 +134,14 @@ func TestRepositoriesService_GetCommunityHealthMetrics(t *testing.T) {
...
 func TestMetric_Marshal(t *testing.T) {
 	t.Parallel()
-	testJSONMarshal(t, &Metric{}, "{}")
+	testJSONMarshal(t, &Metric{}, `{
+		"name": null,
+		"key": null,
+		"spdx_id": null,
+		"url": null,
+		"html_url": null,
+		"node_id": null
```
gmlewis, line 143: It seems odd that we don't have `omitempty`s on all these.

github/repos_community_health_test.go
```diff
@@ -159,7 +166,15 @@ func TestMetric_Marshal(t *testing.T) {
...
 	t.Parallel()
-	testJSONMarshal(t, &CommunityHealthFiles{}, "{}")
+	testJSONMarshal(t, &CommunityHealthFiles{}, `{
+		"code_of_conduct": null,
+		"code_of_conduct_file": null,
+		"contributing": null,
+		"issue_template": null,
+		"pull_request_template": null,
+		"license": null,
+		"readme": null
```
gmlewis, line 176: It seems odd that we don't have `omitempty`s on all these as well, etc.

github/repos_contents_test.go
```diff
@@ -1190,11 +1190,11 @@ func TestRepositoryContentResponse_Marshal(t *testing.T) {
 
 func TestRepositoryContentFileOptions_Marshal(t *testing.T) {
 	t.Parallel()
-	testJSONMarshal(t, &RepositoryContentFileOptions{}, "{}")
+	testJSONMarshal(t, &RepositoryContentFileOptions{}, `{"content": null}`)
 
 	r := &RepositoryContentFileOptions{
 		Message: Ptr("type"),
-		Content: []byte{1},
+		Content: []byte("1"),
```
gmlewis: I'm not sure I understand this change. It's valid to send raw binary data in `Content` such as `[0x01]` like this used to have... Did this change just make it easier for the testing code?  If so, I'm totally fine with the change... I'm mostly curious, and didn't want users of this client library to think that they couldn't send raw bytes (which will get base64-encoded as demonstrated by the unit test).

github/search_test.go
```diff
@@ -1031,7 +1007,7 @@ func TestCommitResult_Marshal(t *testing.T) {
 		Commit:      &Commit{SHA: Ptr("test")},
 		Author:      &User{ID: Ptr(int64(1))},
 		Committer:   &User{ID: Ptr(int64(1))},
-		Parents:     []*Commit{},
+		Parents:     []*Commit(nil),
```
gmlewis: Same here... can this line simply be deleted since this is the zero value for the slice?

github/repos_community_health_test.go
alexandear: It's by design. According to the [returned schema](https://docs.github.com/en/rest/metrics/community?apiVersion=2022-11-28#get-community-profile-metrics), all these fields are required but can be `null`:

```json
{
  "title": "Community Profile",
  "description": "Community Profile",
  "type": "object",
  "properties": {
    "health_percentage": {
      "type": "integer",
      "examples": [
        100
      ]
    },
    "description": {
      "type": [
        "string",
        "null"
      ],
      "examples": [
        "My first repository on GitHub!"
      ]
    },
    "documentation": {
      "type": [
        "string",
        "null"
      ],
      "examples": [
        "example.com"
      ]
    },
    "files": {
      "type": "object",
      "properties": {
        "code_of_conduct": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "Code Of Conduct Simple",
              "description": "Code of Conduct Simple",
              "type": "object",
              "properties": {
                "url": {
                  "type": "string",
                  "format": "uri",
                  "examples": [
                    "https://api.github.com/repos/github/docs/community/code_of_conduct"
                  ]
                },
                "key": {
                  "type": "string",
                  "examples": [
                    "citizen_code_of_conduct"
                  ]
                },
                "name": {
                  "type": "string",
                  "examples": [
                    "Citizen Code of Conduct"
                  ]
                },
                "html_url": {
                  "type": [
                    "string",
                    "null"
                  ],
                  "format": "uri",
                  "examples": [
                    "https://github.com/github/docs/blob/main/CODE_OF_CONDUCT.md"
                  ]
                }
              },
              "required": [
                "url",
                "key",
                "name",
                "html_url"
              ]
            }
          ]
        },
        "code_of_conduct_file": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "Community Health File",
              "type": "object",
              "properties": {
                "url": {
                  "type": "string",
                  "format": "uri"
                },
                "html_url": {
                  "type": "string",
                  "format": "uri"
                }
              },
              "required": [
                "url",
                "html_url"
              ]
            }
          ]
        },
        "license": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "License Simple",
              "description": "License Simple",
              "type": "object",
              "properties": {
                "key": {
                  "type": "string",
                  "examples": [
                    "mit"
                  ]
                },
                "name": {
                  "type": "string",
                  "examples": [
                    "MIT License"
                  ]
                },
                "url": {
                  "type": [
                    "string",
                    "null"
                  ],
                  "format": "uri",
                  "examples": [
                    "https://api.github.com/licenses/mit"
                  ]
                },
                "spdx_id": {
                  "type": [
                    "string",
                    "null"
                  ],
                  "examples": [
                    "MIT"
                  ]
                },
                "node_id": {
                  "type": "string",
                  "examples": [
                    "MDc6TGljZW5zZW1pdA=="
                  ]
                },
                "html_url": {
                  "type": "string",
                  "format": "uri"
                }
              },
              "required": [
                "key",
                "name",
                "url",
                "spdx_id",
                "node_id"
              ]
            }
          ]
        },
        "contributing": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "Community Health File",
              "type": "object",
              "properties": {
                "url": {
                  "type": "string",
                  "format": "uri"
                },
                "html_url": {
                  "type": "string",
                  "format": "uri"
                }
              },
              "required": [
                "url",
                "html_url"
              ]
            }
          ]
        },
        "readme": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "Community Health File",
              "type": "object",
              "properties": {
                "url": {
                  "type": "string",
                  "format": "uri"
                },
                "html_url": {
                  "type": "string",
                  "format": "uri"
                }
              },
              "required": [
                "url",
                "html_url"
              ]
            }
          ]
        },
        "issue_template": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "Community Health File",
              "type": "object",
              "properties": {
                "url": {
                  "type": "string",
                  "format": "uri"
                },
                "html_url": {
                  "type": "string",
                  "format": "uri"
                }
              },
              "required": [
                "url",
                "html_url"
              ]
            }
          ]
        },
        "pull_request_template": {
          "anyOf": [
            {
              "type": "null"
            },
            {
              "title": "Community Health File",
              "type": "object",
              "properties": {
                "url": {
                  "type": "string",
                  "format": "uri"
                },
                "html_url": {
                  "type": "string",
                  "format": "uri"
                }
              },
              "required": [
                "url",
                "html_url"
              ]
            }
          ]
        }
      },
      "required": [
        "code_of_conduct",
        "code_of_conduct_file",
        "license",
        "contributing",
        "readme",
        "issue_template",
        "pull_request_template"
      ]
    },
    "updated_at": {
      "type": [
        "string",
        "null"
      ],
      "format": "date-time",
      "examples": [
        "2017-02-28T19:09:29Z"
      ]
    },
    "content_reports_enabled": {
      "type": "boolean",
      "examples": [
        true
      ]
    }
  },
  "required": [
    "health_percentage",
    "description",
    "documentation",
    "files",
    "updated_at"
  ]
}
```

We can change this in a separate PR. Because this PR touches only tests.

github/repos_community_health_test.go
alexandear: Same as above.

github/meta_test.go
alexandear: I'm not sure which URLs you consider third‑party here. Could you clarify which ones you want to remove?

@Not-Dhananjay-Mishra recently changed the third party URLs in https://github.com/google/go-github/pull/3948/changes#diff-7ef9cd4ccd04bdd6ea1767f55b9df3e3c8866a233e9a9bf62bbfdc5b33e59a9a

gmlewis: So I visited one of these: "tuf-repo.github.com" which was added in #3249.
The web page (hosted by GitHub) points to a repo that returns a 404 which may possibly be a private GitHub repo.

I'm not saying that these are malicious, but wouldn't this be a clever way to spread malware or a virus?

I don't see the need to keep anything that starts with `"*.github.com"` when we could fully standardize to `"*.example.com"`.

Thoughts?

github/repos_contents_test.go
alexandear: Reverted.

github/meta_test.go
alexandear: OK, got it. Changed to `example.com`; others removed.

gmlewis:
So moving forward, do developers need to add `testJSONMarshal` or `testJSONMarshalData` or `testJSONUnmarshalData` or a combination of the latter, or how will they decide?

alexandear:
> So moving forward, do developers need to add `testJSONMarshal` or `testJSONMarshalData` or `testJSONUnmarshalData` or a combination of the latter, or how will they decide?

That's a good question, and it highlights a somewhat misleading API. I've slightly improved the test functions to make their intent clearer.

In general, developers should use `testJSONMarshal`, and only in rare cases use `testJSONMarshalOnly` or `testJSONUnmarshalOnly`.

gmlewis:
> > So moving forward, do developers need to add `testJSONMarshal` or `testJSONMarshalData` or `testJSONUnmarshalData` or a combination of the latter, or how will they decide?
> 
> That's a good question, and it highlights a somewhat misleading API. I've slightly improved the test functions to make their intent clearer.
> 
> In general, developers should use `testJSONMarshal`, and only in rare cases use `testJSONMarshalOnly` or `testJSONUnmarshalOnly`.

Do you want to give some guidance or recommendations in either CONTRIBUTING.md or in the comments of these functions?

gmlewis (approved):
Excellent!  Thank you, @alexandear!
LGTM.

github/github_test.go
```diff
@@ -161,32 +161,101 @@ func testBody(t *testing.T, r *http.Request, want string) {
...
+
+	testJSONMarshalOnly(t, v, want)
+	testJSONUnmarshalOnly(t, v, want, opts...)
+}
+
+// testJSONMarshalOnly tests JSON marshaling by comparing the marshaled output with the expected JSON string.
+//
+// This function compares JSON by unmarshaling both values into any and using cmp.Diff.
+// This means the comparison ignores:
+//   - Whitespace differences
```
Not-Dhananjay-Mishra: How about using a space in between? 
https://go.dev//wiki/Spelling
```suggestion
//   - White space differences
```

gmlewis: Wow... I don't remember ever seeing that doc before... thank you, @Not-Dhananjay-Mishra!
OK, well if that has been declared as idiomatic Go spelling, then I can't argue (even though I was about to argue that the spelling "whitespace" has been around for as long as I can remember).
So sure. 

Here are other locations:

```
$ rg -i whitespace
CONTRIBUTING.md
267:- `format` - formats whitespace in `openapi_operations.yaml` and sorts its

tools/metadata/main.go
35:     "format_help": `Format whitespace in openapi_operations.yaml and sort its operations.`,
```

alexandear: Thank you. Fixed by #4047

gmlewis:
Very nice, @alexandear, thank you! 
Merging.

PR #4048, opened 2026-02-27
chore: Simplify form value assertions in tests

alexandear:
Sometimes we have query keys that contain multiple values and need to verify them, for example in `TestOrganizationsService_ListFineGrainedPersonalAccessTokens`. So I added `testFormValuesList` instead of modifying `testFormValues`, since changing it would require many updates.

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -135,7 +135,7 @@ func testFormValues(t *testing.T, r *http.Request, want values) {
 	}
 }
 
-// testFormValuesMany checks that the request form values match the expected values.
+// testFormValuesList checks that the request form values match the expected values.
 // It allows for multiple values per key.
 func testFormValuesList(t *testing.T, r *http.Request, want url.Values) {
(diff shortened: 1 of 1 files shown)


github/github_test.go
```diff
@@ -119,16 +119,30 @@ func testMethod(t *testing.T, r *http.Request, want string) {
...
+		wantValues.Add(k, v)
+	}
+
+	assertNilError(t, r.ParseForm())
+	if got := r.Form; !cmp.Equal(got, wantValues) {
+		t.Errorf("Request query parameters: %v, want %v", got, wantValues)
 	}
+}
+
+// testFormValuesMany checks that the request form values match the expected values.
```
```suggestion
// testFormValuesList checks that the request form values match the expected values.
```

gmlewis: I was going to commit this suggestion, but apparently I'm not allowed to in this PR.  :smile:

PR #4056, opened 2026-03-02
fix!: Change `id` from `int64` to `string` in `ActivityService.MarkThreadDone`

### LINKED ISSUE #3269: Some API calls that are defined as taking `int`s in the GitHub spec are strings in this library
See https://github.com/google/go-github/pull/3265#discussion_r1761345637 for context.

Happy to take the work to fix this, filing an issue for tracking.

Question about ergonomics, though... the current library (as well as the GitHub API spec[0]) return the thread ID as a string.  Changing the library to match the spec makes it a bit more annoying to do something like 'get a thread -> call an API on that thread', since this will now require converting the returned string to an int.

Given that this 'problem' is present in the API spec, I'm inclined to match the spec, but want to double check that this is the desired behavior of this project.

[0] From https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#get-a-thread
```json
{
  "title": "Thread",
  "description": "Thread",
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    ...
```

alexandear:
BREAKING CHANGE: `ActivityService.MarkThreadDone` accepts `string` `id` instead of `int64`.

---

This is for consistency with `GetThread`, `MarkThreadRead`, `GetThreadSubscription`, which all accept an id as `string` and not `int64`. `ListNotifications` and `ListRepositoryNotifications` return `Notification` with `ID` as `string`.

Fixes #3269

### DIFF
--- a/github/activity_notifications.go
+++ b/github/activity_notifications.go
@@ -186,7 +186,7 @@ func (s *ActivityService) MarkThreadRead(ctx context.Context, id string) (*Respo
 // GitHub API docs: https://docs.github.com/rest/activity/notifications#mark-a-thread-as-done
 //
 //meta:operation DELETE /notifications/threads/{thread_id}
-func (s *ActivityService) MarkThreadDone(ctx context.Context, id int64) (*Response, error) {
+func (s *ActivityService) MarkThreadDone(ctx context.Context, id string) (*Response, error) {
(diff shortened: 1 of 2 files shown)


munlicode:
LGTM

Issue #4074, opened 2026-03-09
Inconsistency in creating and updating a custom organization role


munlicode:
```go
// CreateOrUpdateOrgRoleOptions represents options required to create or update a custom organization role.
type CreateOrUpdateOrgRoleOptions struct {
	Name        *string  `json:"name,omitempty"`
	Description *string  `json:"description,omitempty"`
	Permissions []string `json:"permissions"`
	BaseRole    *string  `json:"base_role,omitempty"`
}
```

It seems like it does not follow documentation, because `CreateCustomOrgRole` should expect at least `name` and `permissions`, and `UpdateCustomOrgRole` does not demand any of those to be present.

Is it intentional?

Documentation references:
- https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role
- https://docs.github.com/enterprise-cloud@latest/rest/orgs/organization-roles#update-a-custom-organization-role

gmlewis:
In general, mismatches with the official GitHub v3 API docs are not intentional unless a comment explicitly says otherwise. 

Any inconsistencies found you are welcome to fix. 

Thank you, @munlicode!

PR #4075, opened 2026-03-09
fix!: Resolve inconsistent options for `create` and `update` on custom org role

munlicode:
BREAKING CHANGE: `GetOrgRole`, `CreateCustomOrgRole`, and `UpdateCustomOrgRole` have new params and return values.

- Split structs for `CreateCustomOrgRole` and `UpdateCustomOrgRole` into `CreateCustomOrgRoleRequest` and `UpdateCustomOrgRoleRequest`.
- Rename the `CustomOrgRoles` struct definition to `CustomOrgRole`, updating its usage in `GetOrgRole`, `CreateCustomOrgRole`, and `UpdateCustomOrgRole`.
- Set `name` required body parameter on create operation.
- Set all body parameters to be optional on update operation.

Fixes: #4074.

### DIFF
--- a/github/orgs_organization_roles.go
+++ b/github/orgs_organization_roles.go
@@ -29,7 +29,7 @@ type CustomOrgRole struct {
 	BaseRole    *string       `json:"base_role,omitempty"`
 }
 
-// CreateCustomOrgRoleRequest represents options required to create a custom organization role.
+// CreateCustomOrgRoleRequest represents body parameters required to create a custom organization role.
 type CreateCustomOrgRoleRequest struct {
 	Name        string   `json:"name"`
 	Description *string  `json:"description,omitempty"`


github/orgs_organization_roles.go
```diff
@@ -89,7 +97,7 @@ func (s *OrganizationsService) GetOrgRole(ctx context.Context, org string, roleI
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/orgs/organization-roles#create-a-custom-organization-role
 //
 //meta:operation POST /orgs/{org}/organization-roles
-func (s *OrganizationsService) CreateCustomOrgRole(ctx context.Context, org string, opts *CreateOrUpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
+func (s *OrganizationsService) CreateCustomOrgRole(ctx context.Context, org string, opts *CreateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
```
alexandear: `CreateOrgRoleOptions` are body parameters, but we typically use the `Options`  suffix for query parameters. Also, a request body is required when creating a custom org role.
Additionally, we are returning a role, not roles.

Therefore, I propose the following renamings:

`CreateOrgRoleOptions` -> `CreateCustomOrgRoleRequest`
`CustomOrgRoles` -> `CustomOrgRole`

```suggestion
func (s *OrganizationsService) CreateCustomOrgRole(ctx context.Context, org string, request CreateCustomOrgRoleRequest) (*CustomOrgRole, *Response, error) {
```

We're breaking this endpoint anyway, so let’s do it in the best possible way.

github/orgs_organization_roles.go
```diff
@@ -112,7 +120,7 @@ func (s *OrganizationsService) CreateCustomOrgRole(ctx context.Context, org stri
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/orgs/organization-roles#update-a-custom-organization-role
 //
 //meta:operation PATCH /orgs/{org}/organization-roles/{role_id}
-func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org string, roleID int64, opts *CreateOrUpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
+func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org string, roleID int64, opts *UpdateOrgRoleOptions) (*CustomOrgRoles, *Response, error) {
```
alexandear: Same as above:


```suggestion
func (s *OrganizationsService) UpdateCustomOrgRole(ctx context.Context, org string, roleID int64, request UpdateCustomOrgRoleRequest) (*CustomOrgRole, *Response, error) {
```

github/orgs_organization_roles.go
```diff
@@ -29,11 +29,19 @@ type CustomOrgRoles struct {
...
 }
 
-// CreateOrUpdateOrgRoleOptions represents options required to create or update a custom organization role.
-type CreateOrUpdateOrgRoleOptions struct {
+// CreateOrgRoleOptions represents options required to create a custom organization role.
+type CreateOrgRoleOptions struct {
+	Name        string   `json:"name"`
+	Permissions []string `json:"permissions"`
+	Description *string  `json:"description,omitempty"`
+	BaseRole    *string  `json:"base_role,omitempty"`
```
alexandear: Image

Let's rearrange fields as they appear in the doc (and add the comment for the `BaseRole`):

```suggestion
	Name        string   `json:"name"`
	Description *string  `json:"description,omitempty"`
	Permissions []string `json:"permissions"`
	BaseRole    *string  `json:"base_role,omitempty"` // Can be one of: read, triage, write, maintain, admin
```

github/orgs_organization_roles.go
```diff
@@ -29,11 +29,19 @@ type CustomOrgRoles struct {
...
+}
+
+// UpdateOrgRoleOptions represents options required to update a custom organization role.
+type UpdateOrgRoleOptions struct {
 	Name        *string  `json:"name,omitempty"`
+	Permissions []string `json:"permissions,omitempty"`
 	Description *string  `json:"description,omitempty"`
-	Permissions []string `json:"permissions"`
 	BaseRole    *string  `json:"base_role,omitempty"`
 }
```
alexandear:
```suggestion
type UpdateCustomOrgRoleRequest struct {
	Name        *string  `json:"name,omitempty"`
	Description *string  `json:"description,omitempty"`
	Permissions []string `json:"permissions,omitempty"`
	BaseRole    *string  `json:"base_role,omitempty"` // Can be one of: none, read, triage, write, maintain, admin
}
```

github/orgs_organization_roles.go
munlicode: Would it not be confusing to have `request` and `req` ?

```go

func (s *OrganizationsService) CreateCustomOrgRole(ctx context.Context, org string, request CreateCustomOrgRoleRequest) (*CustomOrgRole, *Response, error) {
	u := fmt.Sprintf("orgs/%v/organization-roles", org)

	req, err := s.client.NewRequest("POST", u, request)
	...
```

If it is fine, I am ready to rename `opts` to `request`.

alexandear: We can call it `Body`, but that would require significant refactoring.

And the `Request` prefix is common across the code base:

https://github.com/google/go-github/blob/09946e5e2594a1473a65ae588ab9ed39831aef9c/github/repos.go#L2324
https://github.com/google/go-github/blob/09946e5e2594a1473a65ae588ab9ed39831aef9c/github/orgs_network_configurations.go#L132
https://github.com/google/go-github/blob/09946e5e2594a1473a65ae588ab9ed39831aef9c/github/actions_runners.go#L87

alexandear: The parameter name can be simply `create` instead of `request`.

munlicode: Then I will rename to `request`.

munlicode:
@alexandear thank you for suggestions. I have applied them, can you please check?

github/orgs_organization_roles.go
```diff
@@ -29,12 +29,20 @@ type CustomOrgRoles struct {
 	BaseRole    *string       `json:"base_role,omitempty"`
 }
 
-// CreateOrUpdateOrgRoleOptions represents options required to create or update a custom organization role.
-type CreateOrUpdateOrgRoleOptions struct {
-	Name        *string  `json:"name,omitempty"`
+// CreateCustomOrgRoleRequest represents options required to create a custom organization role.
```
alexandear:
```suggestion
// CreateCustomOrgRoleRequest represents body parameters required to create a custom organization role.
```

github/orgs_organization_roles.go
```diff
@@ -29,12 +29,20 @@ type CustomOrgRoles struct {
...
+type CreateCustomOrgRoleRequest struct {
+	Name        string   `json:"name"`
 	Description *string  `json:"description,omitempty"`
 	Permissions []string `json:"permissions"`
-	BaseRole    *string  `json:"base_role,omitempty"`
+	BaseRole    *string  `json:"base_role,omitempty"` // Can be one of: read, triage, write, maintain, admin
+}
+
+// UpdateCustomOrgRoleRequest represents options required to update a custom organization role.
```
alexandear:
```suggestion
// UpdateCustomOrgRoleRequest represents body parameters to update a custom organization role.
```

Issue #4070, opened 2026-03-06
Improve GitHub Actions performance and also local formatting/testing/linting


gmlewis:
I'm really kinda getting tired of waiting for `golintci` to finish in this repo.
Yes, we have a huge number of linters in addition to custom ones.
There's a new kid on the block that claims 2x performance boost: https://revive.run/
It might be nice if someone wants to investigate switching this repo over from golintci to revive and perform benchmarks on "Before vs After" so we can see if this would be a beneficial switch.
Obviously, we would want the same coverage, as all of our linters are in place for good reason.

stevehipwell:
@gmlewis do you mean `golangci-lint`? If so I'm not sure they're comparable as `golangci-lint` is an aggregator not an actual linter. `golangci-lint` does have support for [revivie](https://golangci-lint.run/docs/linters/configuration/#revive) so if you could replace some of the other linters called by `golangci-lint` with `revive` that may improve performance.

FYI we currently have two unnecessary formatters configured, `gofmt` (unnecessary as `gofumpt` is a superset) & `gci` (unnecessary as we're using `goimports` and don't have a complicated import policy that would need this mainly by virtue of having very few dependencies). Removing these two formatters speeds up formatting the whole repo by just over 15%.

gmlewis:
Thank you, @stevehipwell!  Yes, any speedups in our CI/CD pipeline, both on GitHub Actions and when running locally would be really nice, so feel free to make PRs for performance gains in formatting/testing/linting.

stevehipwell:
Just looking at the `revive` repo and it looks like this is @alexandear area of expertise so I'd be really interested in his view. Off the top of my head I'd think that we'd still want `golangci-lint` but that we might be able to replace some of the existing formatters with `revive`.

stevehipwell:
On the subject of linters, shouldn't we have `linters.default` set to `none` so only the explicitly enabled ones are used? Also I think `copyloopvar` may be unnecessary as we're `> 1.22`?

alexandear:
> On the subject of linters, shouldn't we have `linters.default` set to `none` so only the explicitly enabled ones are used?

A missing `linters.default` setting means we get the default value `linters.default: standard`, which includes linters `errcheck`, `govet`, `ineffassign`, `staticcheck`, `unused`. We can explicitly enable these linters and then add `linters.default: none` to the configuration.

> Also I think `copyloopvar` may be unnecessary as we're `> 1.22`?

[...]

PR #4094, opened 2026-03-13
chore: Remove duplicated formatters

alexandear:
The formatters `gofmt` and `goimports` can be removed because we have more powerful `gofumpt` and `gci`.

Updates #4070

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -421,9 +421,7 @@ issues:
 formatters:
   enable:
     - gci
-    - gofmt
     - gofumpt
-    - goimports
   settings:
     gofumpt:
       extra-rules: true


stevehipwell:
@alexandear do we have a need for `gci`, and if not is there a reason to pick it over `goimports`? The main reason I ask is because we have a single non-stdlib import in the non-test code and there are only 2 in the main codebase; so it doesn't seem like this is an area we need much power?

alexandear:
> @alexandear do we have a need for `gci`, and if not is there a reason to pick it over `goimports`? The main reason I ask is because we have a single non-stdlib import in the non-test code and there are only 2 in the main codebase; so it doesn't seem like this is an area we need much power?

Yes, we do.

`goimports` allows the following:

```go
import (
	"fmt"

	"net/http"

	"os"

	"github.com/google/go-github/v84/github"

	"google.golang.org/appengine"

	"google.golang.org/appengine/log"
)
```

`gci` does not. It formats to:

```go
import (
	"fmt"
	"net/http"
	"os"

	"github.com/google/go-github/v84/github"
	"google.golang.org/appengine"
	"google.golang.org/appengine/log"
)
```

You can verify this by running:

```
./bin/custom-gcl fmt --enable=goimports --no-config

./bin/custom-gcl fmt --enable=gci --no-config
```

stevehipwell:
@alexandear fair enough, I was only looking at the core codebase.

PR #4098, opened 2026-03-16
Remove deprecated permission property from NewTeam struct  in CreateTeam.

### LINKED ISSUE #4077: New API Version
A new API version just dropped: https://github.blog/changelog/2026-03-12-rest-api-version-2026-03-10-is-now-available/

What kind of a path forward do you envision for this?

Should the library provide support for bott versions for a while going forward? Or would you want to wait before adopting the new API version?

I could be interested to implement some of the changes, but having a discussion about How sounds reasonable to me first ☺️

maishivamhoo123:
Fixes #4077

Removed the deprecated permission property from the NewTeam struct request payload when a team is created.

Affected endpoints:

POST /orgs/{org}/teams

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,11 +4,12 @@ run:
     - integration
   allow-parallel-runners: true
 linters:
+  default: none
   enable:
     - canonicalheader
-    - copyloopvar
     - dogsled
     - dupl
+    - errcheck
     - errorlint
     - fmtpercentv
     - forbidigo
@@ -17,7 +18,7 @@ linters:
     - goheader
     - gosec
     - govet
-    - intrange
+    - ineffassign
     - misspell
     - modernize
     - musttag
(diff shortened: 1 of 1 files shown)


gmlewis:
@maishivamhoo123 - we have a challenge removing deprecated fields in this client library because we also support GitHub Enterprise users who often use this repo with older versions of GitHub Enterprise that still use deprecated fields.

To support the new API, we won't necessarily be removing deprecated fields, but we may be marking fields as deprecated and adding support for new/modified endpoints, if that makes sense.

maishivamhoo123:
@gmlewis Thanks for the clarification. I’ll keep this in mind for my next PR and make sure deprecated fields are handled appropriately. Thank you !

gmlewis:
> @gmlewis Thanks for the clarification. I’ll keep this in mind for my next PR and make sure deprecated fields are handled appropriately. Thank you !

Sounds good. Closing since we can't remove deprecated fields for GitHub Enterprise users.

Issue #4103, opened 2026-03-17
Clean up unnecessary use of `new` and `&SomeStruct{}`


gmlewis:
Time and again, new PRs show up with non-idiomatic Go code using `new` or `&SomeStruct{}` when a zero-value pointer is preferred and often actually prevents bugs as demonstrated by a recent PR (maybe some day I'll search for it if there is demand).

Frequently, PR authors are simply copy/pasting other examples found in this repo, and therefore all usage should be cleaned up.

PR #4106, opened 2026-03-17
chore: Remove unnecessary use of `new` and `&SomeStruct{}` and add new `extraneousnew` custom linter

gmlewis:
Fixes: #4103.

### DIFF
--- a/tools/extraneous-new/extraneous_new.go
+++ b/tools/extraneous-new/extraneous_new.go
@@ -27,30 +27,19 @@ type ExtraneousNewPlugin struct {
 	ignoredMethods map[string]bool
 }
 
-// Settings is the configuration for the extraneousnew linter.
-type Settings struct {
-	// IgnoredMethods causes the linter to skip linting individual `Receiver.Methods`.
-	IgnoredMethods []string `json:"ignored-methods" yaml:"ignored-methods"`
-}
-
 // New returns an analysis.Analyzer to use with golangci-lint.
 func New(cfg any) (register.LinterPlugin, error) {
-	settings := Settings{}
+	ignoredMethods := map[string]bool{}
 	if cfg != nil {
 		if data, ok := cfg.(map[string]any); ok {
 			if ignored, ok := data["ignored-methods"].([]any); ok {
 				for _, m := range ignored {
 					if s, ok := m.(string); ok {
-						settings.IgnoredMethods = append(settings.IgnoredMethods, s)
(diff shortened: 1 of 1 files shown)


gmlewis:

alexandear (commented):
Could we implement a linter to automatically detect that? I don’t think it would be hard to do.

---

nit: typo in the title: `unncessary` -> `unnecessary `

gmlewis:
> Could we implement a linter to automatically detect that? I don’t think it would be hard to do.

> I also agree with @alexandear that a linter to catch these would be useful. It looks like a linter for this would be general purpose, so it might be worth looking to see if one is already being worked on.

I think that is a great idea, but I think that this might actually be one of the most challenging linters that we would have, and there are several reasons for this:

* there are many valid uses of `new` and `&SomeStruct{}` in this repo (that I obviously didn't change) that the linter should allow
* there are very specific scenarios that the linter would need to address

Now, having said that, I think it would be a really fun challenge to come up with such a linter, and I HIGHLY doubt that anything exists out in the wild that will target only these very specific scenarios. And yes, I've been told that I'm very opinionated. :joy:

So, challenge accepted... I'll work on it in this PR since it makes sense to have it verify that I did the 3-hour manual cleanup properly. :smile:

gmlewis:
I've gone ahead and implemented the new `extraneousnew` custom linter (with AI assistance this time) and applied it to this repo. It found one more instance that I had manually missed. :tada:

tools/extraneous-new/extraneous_new.go
```diff
@@ -0,0 +1,308 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	ignoredMethods := map[string]bool{}
+	for _, m := range settings.IgnoredMethods {
+		ignoredMethods[m] = true
+	}
```
alexandear: We can remove `Settings`:


```suggestion
// New returns an analysis.Analyzer to use with golangci-lint.
func New(cfg any) (register.LinterPlugin, error) {
	ignoredMethods := map[string]bool{}
	if cfg != nil {
		if data, ok := cfg.(map[string]any); ok {
			if ignored, ok := data["ignored-methods"].([]any); ok {
				for _, m := range ignored {
					if s, ok := m.(string); ok {
						ignoredMethods[s] = true
					}
				}
			}
		}
	}
```

gmlewis:
Merging.

PR #4119, opened 2026-03-23
chore: Use `golangci-lint-action`; remove `newreposecretwithlibsodium`

alexandear:
- Update GitHub Actions linter workflow to use official [`golangci/golangci-lint-action`](https://github.com/golangci/golangci-lint-action) instead of `./script/lint.sh`. It automatically builds custom golangci-lint cache binaries with `~/.cache/golangci-lint`, and runs golangci-lint for each module.
- Remove the libsodium-based example because it requires custom library installation, and golangci-lint fails due to this. I think `newreposecretwithxcrypto` should be enough.

The cached lint step takes only 40 seconds:

image

Closes #4118
Updates #4070


.github/workflows/linter.yml
```diff
@@ -13,13 +13,14 @@ jobs:
...
       with:
-        path: |
-          bin/golangci-lint
-          bin/custom-gcl
-        key: ${{ runner.os }}-${{ runner.arch }}-tools-${{ hashFiles('.custom-gcl.yml', 'tools/**/go.mod', 'tools/**/go.sum', 'tools/**/*.go') }}
-    - run: script/lint.sh
+        version: v2.10.1 # sync with version in .custom-gcl.yml
+        experimental: "automatic-module-directories"
+    - name: Check OpenAPI
+      run: ./script/metadata.sh update-openapi --validate
```
gmlewis, line 23: Does this mean that is it going to check the `openapi` YAML file every time? If so, I have a concern with this.
That YAML file gets updated frequently (daily?) on the GitHub side, and gets a new hash every time it changes, which would mean that we would be out-of-date with it _most_ of them time.

As it currently stands, we have the freedom and flexibility to update our YAML file when it benefits us to do so, and I would like to continue with this luxury if possible.

alexandear: Yes. I didn't change the previous behaviour.
`.script/lint.sh` validates OpenAPI every time when running on GitHub Actions:

image

https://github.com/google/go-github/actions/runs/23317729164/job/67821480217#step:5:21

Issue #3983, opened 2026-02-09
Improve and simplify build scripts


alexandear:
We can simplify the scripts added in #2928 to make them more robust and maintainable. Proposed improvements are as follows.

### script/generate.sh

The current [script/generate.sh
](https://github.com/google/go-github/blob/5bdc95411b6370ffc769c784402004ed5da4d0f0/script/generate.sh) serves two purposes:

1) Generating Go code by running `go generate` and `go mod tidy`.
2) Verifying whether the first step produces any diff.

Currently, the second step is overly complex:

1) Creating a temporary git worktree
2) Copying current files into it
3) Making a commit with the current state
4) Running generate.sh to regenerate files
5) Checking if any files changed

While this approach avoids modifying the local working directory, the complexity often leads to unexpected issues like #3982.

I propose removing the Git worktree and commit logic. We can simply run `go generate`, `git diff`, `git restore`.

<details><summary>Proposed new script/generate.sh</summary>
<p>

```sh
#!/bin/sh
#/ script/generate.sh runs go generate on all modules in this repo.
#/ `script/generate.sh --check` checks that the generated files are up to date.

set -e

CDPATH="" cd -- "$(dirname -- "$0")/.."

CHECK_MODE=""
if [ "$1" = "--check" ]; then
  CHECK_MODE=1
fi

MOD_DIRS="$(git ls-files '*go.mod' | xargs dirname | sort)"

for dir in $MOD_DIRS; do
  (
    cd "$dir"
    go generate ./...
    go mod tidy
  )
done

if [ -n "$CHECK_MODE" ]; then
  if [ -n "$(git status --porcelain)" ]; then
    msg="Generated files are out of date. Please run script/generate.sh and commit the results"
    if [ -n "$GITHUB_ACTIONS" ]; then
      echo "::error ::$msg"
    else
      echo "$msg" 1>&2
    fi
    git diff
    git restore .
    exit 1
  fi
fi

```

</p>
</details> 

### script/metadata.sh

The [script/metadata.sh](https://github.com/google/go-github/blob/5bdc95411b6370ffc769c784402004ed5da4d0f0/script/generate.sh) currently builds and executes `tools/metadata` with various parameters. It can be replaced with `go run` for consistency with [go run gen-accessors.go, go run gen-iterators.go](https://github.com/google/go-github/blob/0938d8c0f08b4ec777a88e6a8bcb09bfe306942c/github/github.go#L6-L8).

gmlewis:
[...]

PR #4120, opened 2026-03-24
chore: Simplify `generate.sh` by removing `git worktree` and using generator-based check 

Not-Dhananjay-Mishra:
Added `-check` (env-driven via `CHECK=1`) support to:
- gen-accessors.go
- gen-iterators.go
- gen-stringify-test.go

Simplified script/generate.sh to:
- run `go generate ./...`
- optionally enable check mode via `CHECK=1`

Related : #3983


script/generate.sh
```diff
@@ -9,46 +9,9 @@ set -e
...
-      fi
-      git diff
-      exit 1
-    }
-  )
-  exit 0
+  export CHECK=1
 fi
 
-MOD_DIRS="$(git ls-files '*go.mod' | xargs dirname | sort)"
```
alexandear, line 45: Previously, the script ran `go generate ./...` with `go mod tidy` for every module directory.

Now, it runs only `go generate ./...` for the root module.

alexandear: `go mod tidy` is essential as it catches outdated `go.mod/go.sum` after adding a dependency.

Not-Dhananjay-Mishra: ~Do we need to run `go mod tidy` and `go generate./...` for every module directory?~

Not-Dhananjay-Mishra: > `go mod tidy` is essential as it catches outdated `go.mod/go.sum` after adding a dependency.

Should we run `go mod tidy` in every module directory, and `go generate ./...` from the root? Or should `go generate` also be run in every module directory?

alexandear: `go mod tidy` in every module dir, `go generate` in the root only.

Not-Dhananjay-Mishra: > `go mod tidy` in every module dir, `go generate` in the root only.

Thanks for clarification, Just one more doubt
If CHECK = 1 : we only check if `go.mod/go.sum` are outdated or not (don't make any change to them, with `go mod tidy -diff`)
else : we do `go mod tidy` and update them

github/gen-accessors.go
```diff
@@ -40,6 +40,7 @@ const (
 
 var (
 	verbose = flag.Bool("v", false, "Print verbose log messages")
+	check   = flag.Bool("check", false, "check whether generated files are up to date")
```
alexandear:
```suggestion
	check   = flag.Bool("check", false, "Check whether generated files are up to date")
```

github/gen-accessors.go
```diff
@@ -233,6 +238,23 @@ func (t *templateData) dump() error {
 			return fmt.Errorf("format.Source:\n%v\n%v", buf.String(), err)
 		}
 
+		if isCheck() {
+			logf("Checking %v...", filename)
+			old, err := os.ReadFile(filename)
+			if err != nil {
+				if os.IsNotExist(err) {
```
alexandear: From https://pkg.go.dev/os#IsNotExist:

> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

github/gen-accessors.go
```diff
@@ -233,6 +238,23 @@ func (t *templateData) dump() error {
...
+					return fmt.Errorf("Missing file: %v\n", t.filename)
```
alexandear:
```suggestion
					return fmt.Errorf("missing file: %v\n", t.filename)
```

script/generate.sh
```diff
@@ -8,47 +8,28 @@ set -e
...
+        exit 1
+      fi
+    else
+      go mod tidy
+    fi
   )
 done
-script/run-check-structfield-settings.sh -fix
+
+script/run-check-structfield-settings.sh -fix
```
alexandear: please return back EOF

Not-Dhananjay-Mishra: Could you please clarify what you mean by “return back EOF”?

alexandear: `EOF` - end of file. GitHub warns about missing EOF:

image

"return back" - "revert the change"

alexandear (approved):
LGTM

github/gen-accessors.go
```diff
@@ -40,6 +42,7 @@ const (
 
 var (
 	verbose = flag.Bool("v", false, "Print verbose log messages")
+	check   = flag.Bool("check", false, "Check whether generated files are up to date")
```
gmlewis: Please swap these lines so that the flags are in sorted order (`-check` before `-v`).

github/gen-iterators.go
```diff
@@ -35,6 +37,7 @@ const (
 
 var (
 	verbose = flag.Bool("v", false, "Print verbose log messages")
+	check   = flag.Bool("check", false, "Check whether generated files are up to date")
```
gmlewis: Same here, please.

github/gen-iterators.go
```diff
@@ -599,8 +606,25 @@ func (t *templateData) dump() error {
...
+			if !bytes.Equal(old, clean) {
+				return fmt.Errorf("Generated files are out of date. Please run go generate ./... and commit the results")
+			}
+			return nil
+		}
+
 		logf("Writing %v...", filename)
-		return os.WriteFile(filename, clean, 0o644)
+		err = os.WriteFile(filename, clean, 0o644)
+		return err
```
gmlewis: Please revert this change as it is unnecessary.

github/gen-stringify-test.go
```diff
@@ -37,6 +39,7 @@ const (
 
 var (
 	verbose = flag.Bool("v", false, "Print verbose log messages")
+	check   = flag.Bool("check", false, "Check whether generated files are up to date")
```
gmlewis: Same here, please.

gmlewis (changes requested):
Thank you, @Not-Dhananjay-Mishra and @alexandear!
Just a few minor tweaks, please, then we can merge.

github/gen-stringify-test.go
Not-Dhananjay-Mishra: Thanks @gmlewis
Done the changes.

gmlewis (approved):
LGTM.
Merging.

PR #4125, opened 2026-04-01
fix: Limit webhook payload size in `ValidatePayloadFromBody`

YooLCD:
`ValidatePayloadFromBody` reads the request body with `io.ReadAll` without any size limit, allowing an attacker to send an arbitrarily large payload and exhaust server memory.

This PR caps the body at 25 MB (GitHub's documented webhook payload limit) using `io.LimitReader` before reading, and returns an error if the limit is exceeded.

Both `application/json` and `application/x-www-form-urlencoded` code paths are covered. A test with an unbounded reader is added for each.

### DIFF
--- a/github/messages_test.go
+++ b/github/messages_test.go
@@ -206,15 +206,9 @@ func (b *badReader) Read([]byte) (int, error) {
 
 func (b *badReader) Close() error { return errors.New("bad reader") }
 
-// infiniteReader is an io.Reader that returns zeros indefinitely.
-type infiniteReader struct{}
+type readerFunc func([]byte) (int, error)
 
-func (infiniteReader) Read(p []byte) (int, error) {
-	for i := range p {
-		p[i] = 0
-	}
-	return len(p), nil
-}
+func (f readerFunc) Read(p []byte) (int, error) { return f(p) }
 
 func TestValidatePayload_BadRequestBody(t *testing.T) {
 	t.Parallel()
@@ -252,7 +246,12 @@ func TestValidatePayload_OversizedBody(t *testing.T) {
 		t.Run(fmt.Sprintf("test #%v", i), func(t *testing.T) {
 			t.Parallel()
 			// Simulate a reader that reports more than maxPayloadSize bytes.
-			oversized := &fixedSizeReader{remaining: maxPayloadSize + 1}
+			oversized := io.LimitReader(readerFunc(func(p []byte) (int, error) {
+				for i := range p {
+					p[i] = 0
+				}
+				return len(p), nil
+			}), maxPayloadSize+1)
 			req := &http.Request{
 				Header: http.Header{"Content-Type": []string{tt.contentType}},
 				Body:   io.NopCloser(oversized),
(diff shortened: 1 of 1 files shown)


github/messages_test.go
```diff
@@ -228,6 +239,31 @@ func TestValidatePayload_BadRequestBody(t *testing.T) {
...
+	for i, tt := range tests {
+		t.Run(fmt.Sprintf("test #%v", i), func(t *testing.T) {
+			t.Parallel()
+			// Simulate a reader that reports more than maxPayloadSize bytes.
+			oversized := io.LimitReader(infiniteReader{}, maxPayloadSize+1)
+			req := &http.Request{
+				Header: http.Header{"Content-Type": []string{tt.contentType}},
+				Body:   io.NopCloser(oversized),
+			}
+			if _, err := ValidatePayload(req, nil); err == nil {
```
alexandear: Could we check that the specific error "webhook payload exceeds maximum allowed size" is returned?

YooLCD: Done! I've updated the test to assert that the returned error message is exactly "webhook payload exceeds maximum allowed size".

github/messages.go
```diff
@@ -199,9 +203,12 @@ func ValidatePayloadFromBody(contentType string, readable io.Reader, signature s
 	switch contentType {
 	case "application/json":
 		var err error
-		if body, err = io.ReadAll(readable); err != nil {
+		if body, err = io.ReadAll(io.LimitReader(readable, maxPayloadSize+1)); err != nil {
```
alexandear: These lines duplicate. I propose moving them to a helper function:

```go
func readPayloadBody(readable io.Reader) ([]byte, error) {
	body, err := io.ReadAll(io.LimitReader(readable, maxPayloadSize+1))
	if err != nil {
		return nil, err
	}
	if len(body) > maxPayloadSize {
		return nil, errors.New("webhook payload exceeds maximum allowed size")
	}
	return body, nil
}
```

github/messages.go
```diff
@@ -199,9 +203,12 @@ func ValidatePayloadFromBody(contentType string, readable io.Reader, signature s
...
 			return nil, err
 		}
+		if int64(len(body)) > maxPayloadSize {
```
alexandear: A conversion to `int64` is not needed:
```suggestion
		if len(body) > maxPayloadSize {
```

`25 * 1024 * 1024 + 1` easily fits in `int` on supported platforms.

github/messages_test.go
```diff
@@ -228,6 +239,35 @@ func TestValidatePayload_BadRequestBody(t *testing.T) {
...
+	}{
+		{contentType: "application/json"},
+		{contentType: "application/x-www-form-urlencoded"},
+	}
+
+	for i, tt := range tests {
+		t.Run(fmt.Sprintf("test #%v", i), func(t *testing.T) {
+			t.Parallel()
+			// Simulate a reader that reports more than maxPayloadSize bytes.
+			oversized := io.LimitReader(infiniteReader{}, maxPayloadSize+1)
```
alexandear:
```suggestion
			oversized := &fixedSizeReader{remaining: maxPayloadSize + 1}
```

And the `fixedSizeReader` should be declared below the `TestValidatePayload_OversizedBody`:
```go
// fixedSizeReader is an io.Reader that returns exactly remaining bytes, then EOF.
type fixedSizeReader struct {
	remaining int64
}

func (r *fixedSizeReader) Read(p []byte) (int, error) {
	if r.remaining == 0 {
		return 0, io.EOF
	}

	n := min(int64(len(p)), r.remaining)
	for i := 0; i < int(n); i++ {
		p[i] = 0
	}
	r.remaining -= n
	return int(n), nil
}
```

YooLCD: Done, thank you for the suggestions!

gmlewis:
@YooLCD - in the future, please do not use force push as explained in CONTRIBUTING.md. Thanks.

YooLCD:
I pushed one more commit to fix a CI error. Also, I sincerely apologize for the force pushes earlier. I'll be more careful next time.

gmlewis (approved):
Very nice. Thank you, @YooLCD and @alexandear!
LGTM.
Merging.

PR #4146, opened 2026-04-10
fix: Add missing `User` fields

alexandear:
See response schema https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user


github/users.go
```diff
@@ -20,6 +20,7 @@ type UsersService service
 type User struct {
```
Not-Dhananjay-Mishra, line 20: It’s missing one more field `business_plus`. Should we add that as well?

Image

alexandear: Added, thanks.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis (approved):
LGTM.
Merging.

PR #4148, opened 2026-04-10
chore: Simplify `redundantptr` custom linter

gmlewis:
This reduces the maintenance burden for the `redundantptr` linter by removing an unnecessary fake file with the current version number built into its path.


gmlewis:

munlicode (approved):

munlicode:
LGTM,

Though I wonder, if comments should be left as they are?


```go
_ = Ptr(name) // want `replace github.Ptr\(name\) with &name`
```

Is there a reason for leaving it this way?

alexandear (approved):

gmlewis:
> LGTM,
> 
> Though I wonder, if comments should be left as they are?
> 
> ```go
> _ = Ptr(name) // want `replace github.Ptr\(name\) with &name`
> ```
> 
> Is there a reason for leaving it this way?

Yes, because it only flags the `Ptr` function in a "github" package.

gmlewis:
Thank you, @munlicode and @alexandear!
Merging.

PR #4150, opened 2026-04-13
fix: Reject URL path segments containing ".." in all request methods

mohammadmseet-hue:
## Summary

The existing `ErrPathForbidden` check in `GetContents` (added in #2805) correctly blocks path traversal via the `path` parameter, with the comment:

> path must not contain '..' due to auth vulnerability issue

However, the same validation was missing in two places:

1. **`NewRequest`, `NewFormRequest`, `NewUploadRequest`** — `owner`, `repo`, `org`, `user`, `enterprise`, and `team` parameters are interpolated into URL paths via `fmt.Sprintf` before being passed to these methods. A value like `x/../../../admin` in any of these parameters causes the resolved URL to reach unintended API endpoints, and the client's `Authorization` header is sent along with the request.

2. **`CreateFile`, `UpdateFile`, `DeleteFile`** — these use the same `repos/%v/%v/contents/%v` URL pattern as `GetContents` but lacked the `..` check on the `path` parameter.

### Fix

- Adds a centralized `containsDotDotPathSegment` check in all three request-building methods (`NewRequest`, `NewFormRequest`, `NewUploadRequest`). This protects all current and future API methods from path traversal via any URL-interpolated parameter.
- The check only matches `..` as a complete path segment (not substrings like `file..txt`) and ignores query strings.
- Adds the same `strings.Contains(path, "..")` guard to `CreateFile`, `UpdateFile`, and `DeleteFile` for defense in depth (consistent with `GetContents`).

### Example

```go
// owner parameter traversal — before this fix, the auth token is sent to /admin/users
client.Repositories.GetReadme(ctx, "x/../../../admin", "users", nil)
// resolved URL: https://api.github.com/admin/users/readme
// Authorization: Bearer <token> leaked to unintended endpoint

// After this fix: returns ErrPathForbidden
```

### Impact

Any application using go-github that passes user-controlled input as `owner`, `repo`, or `org` parameters could have its authentication token sent to unintended GitHub API endpoints. This affects all API methods that interpolate these parameters into URL paths.

## Test plan

- [x] Added `TestContainsDotDotPathSegment` — unit tests for the helper function
- [x] Added `TestNewRequest_pathTraversal` — verifies `NewRequest` rejects `..` path segments
- [x] Added `TestNewFormRequest_pathTraversal` — verifies `NewFormRequest` rejects `..` path segments
- [x] Added `TestNewUploadRequest_pathTraversal` — verifies `NewUploadRequest` rejects `..` path segments
- [x] Added `TestRepositoriesService_CreateFile_PathWithParent` — verifies `CreateFile` rejects `..` in path
- [x] Added `TestRepositoriesService_UpdateFile_PathWithParent` — verifies `UpdateFile` rejects `..` in path
- [x] Added `TestRepositoriesService_DeleteFile_PathWithParent` — verifies `DeleteFile` rejects `..` in path
- [x] Full test suite passes (`go test ./github/`)

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -158,6 +158,10 @@ const (
 
 var errNonNilContext = errors.New("context must be non-nil")
 
+// ErrPathForbidden is returned when a URL path contains ".." as a path
+// segment, which could allow path traversal attacks.
+var ErrPathForbidden = errors.New("path must not contain '..' due to auth vulnerability issue")
+
 // A Client manages communication with the GitHub API.
 type Client struct {
 	clientMu              sync.Mutex   // clientMu protects the client during calls that modify the CheckRedirect func.
@@ -552,21 +556,6 @@ func WithVersion(version string) RequestOption {
 	}
 }
 
-// urlContainsDotDotPathSegment reports whether urlStr contains ".." as a path
-// segment (e.g. "a/../b"). It does not match ".." embedded within a segment
-// (e.g. "file..txt"). The check is performed only on the path portion of the
-// URL, ignoring any query string or fragment.
-func urlContainsDotDotPathSegment(urlStr string) (bool, error) {
-	if !strings.Contains(urlStr, "..") {
-		return false, nil
-	}
-	u, err := url.Parse(urlStr)
-	if err != nil {
-		return false, err
-	}
-	return slices.Contains(strings.Split(u.Path, "/"), ".."), nil
-}
-
 // NewRequest creates an API request. A relative URL can be provided in urlStr,
 // in which case it is resolved relative to the BaseURL of the Client.
 // Relative URLs should always be specified without a preceding slash. If
@@ -577,10 +566,8 @@ func (c *Client) NewRequest(method, urlStr string, body any, opts ...RequestOpti
 		return nil, fmt.Errorf("baseURL must have a trailing slash, but %q does not", c.BaseURL)
 	}
 
-	if hasDotDot, err := urlContainsDotDotPathSegment(urlStr); err != nil {
+	if err := checkURLPathTraversal(urlStr); err != nil {
 		return nil, err
-	} else if hasDotDot {
-		return nil, ErrPathForbidden
 	}
 
 	u, err := c.BaseURL.Parse(urlStr)
@@ -629,10 +616,8 @@ func (c *Client) NewFormRequest(urlStr string, body io.Reader, opts ...RequestOp
 		return nil, fmt.Errorf("baseURL must have a trailing slash, but %q does not", c.BaseURL)
 	}
 
-	if hasDotDot, err := urlContainsDotDotPathSegment(urlStr); err != nil {
+	if err := checkURLPathTraversal(urlStr); err != nil {
 		return nil, err
-	} else if hasDotDot {
-		return nil, ErrPathForbidden
 	}
 
 	u, err := c.BaseURL.Parse(urlStr)
@@ -659,6 +644,25 @@ func (c *Client) NewFormRequest(urlStr string, body io.Reader, opts ...RequestOp
 	return req, nil
 }
 
+// checkURLPathTraversal returns ErrPathForbidden if urlStr contains ".." as a
+// path segment (e.g. "a/../b"), preventing path traversal attacks. It does not
+// match ".." embedded within a segment (e.g. "file..txt"). The check is
+// performed only on the path portion of the URL, ignoring any query string or
+// fragment.
+func checkURLPathTraversal(urlStr string) error {
+	if !strings.Contains(urlStr, "..") {
+		return nil
+	}
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return err
+	}
+	if slices.Contains(strings.Split(u.Path, "/"), "..") {
+		return ErrPathForbidden
+	}
+	return nil
+}
+
 // NewUploadRequest creates an upload request. A relative URL can be provided in
 // urlStr, in which case it is resolved relative to the UploadURL of the Client.
 // Relative URLs should always be specified without a preceding slash.
@@ -667,10 +671,8 @@ func (c *Client) NewUploadRequest(urlStr string, reader io.Reader, size int64, m
 		return nil, fmt.Errorf("uploadURL must have a trailing slash, but %q does not", c.UploadURL)
 	}
 
-	if hasDotDot, err := urlContainsDotDotPathSegment(urlStr); err != nil {
+	if err := checkURLPathTraversal(urlStr); err != nil {
 		return nil, err
-	} else if hasDotDot {
-		return nil, ErrPathForbidden
 	}
 
 	u, err := c.UploadURL.Parse(urlStr)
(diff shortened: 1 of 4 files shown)


github/github.go
```diff
@@ -551,6 +551,27 @@ func WithVersion(version string) RequestOption {
...
+	path := urlStr
+	if i := strings.IndexByte(path, '?'); i >= 0 {
+		path = path[:i]
+	}
+	for _, seg := range strings.Split(path, "/") {
+		if seg == ".." {
+			return true
+		}
+	}
+	return false
```
gmlewis:
```suggestion
	return slices.Contains(strings.Split(path, "/"), "..")
```

gmlewis (changes requested):
Thank you, @mohammadmseet-hue!
One minor tweak, please, then we should be ready for a second LGTM+Approval from any other contributor to this repo before merging.

github/github.go
mohammadmseet-hue: Thanks for your quick reply, applying now

gmlewis (approved):

github/github.go
```diff
@@ -551,6 +552,22 @@ func WithVersion(version string) RequestOption {
...
+	if !strings.Contains(urlStr, "..") {
+		return false
+	}
+	// Only check the path portion, ignore query string.
+	path := urlStr
+	if i := strings.IndexByte(path, '?'); i >= 0 {
+		path = path[:i]
+	}
+	return slices.Contains(strings.Split(path, "/"), "..")
+}
```
stevehipwell:
```suggestion
func containsDotDotPathSegment(urlStr string) bool {
	if !strings.Contains(urlStr, "..") {
		return false
	}
	
	u, err := url.Parse(urlStr)
	if err != nil {
		// Not a valid URL so no path segment
		return false
	}
	
	return slices.Contains(u.Path, "..")
}
```

Is there a reason you using `url.Parse()` (which we use elsewhere in this package)?

I'd also suggest changing the name from `containsDotDotPathSegment` to `urlContainsDotDotPathSegment` and changing the signature to return an error if the URL can't be parsed.

github/github_test.go
```diff
@@ -786,6 +786,72 @@ func TestNewRequest_errorForNoTrailingSlash(t *testing.T) {
 	}
 }
 
+func TestContainsDotDotPathSegment(t *testing.T) {
```
stevehipwell: Is it intentional to not have any full URLs in here with a scheme, user or fragment?

stevehipwell (changes requested):

mohammadmseet-hue:
Thanks for the review @stevehipwell!

Good call on using `url.Parse()` — cleaner and consistent with the rest of the package.

Small note on the code suggestion: `slices.Contains(u.Path, "..")` won't compile since `u.Path` is a `string`, not a slice — used `slices.Contains(strings.Split(u.Path, "/"), "..")` instead.

Applied all your suggestions:
- Switched to `url.Parse()` for path extraction
- Renamed to `urlContainsDotDotPathSegment`
- Changed signature to return `(bool, error)`
- Added full URL test cases (scheme, userinfo, fragment)

github/github.go
```diff
@@ -607,6 +629,12 @@ func (c *Client) NewFormRequest(urlStr string, body io.Reader, opts ...RequestOp
 		return nil, fmt.Errorf("baseURL must have a trailing slash, but %q does not", c.BaseURL)
 	}
 
+	if hasDotDot, err := urlContainsDotDotPathSegment(urlStr); err != nil {
+		return nil, err
+	} else if hasDotDot {
+		return nil, ErrPathForbidden
```
alexandear: Please move the declaration `ErrPathForbidden` to this file.

github/github.go
alexandear: This logic duplicates in NewRequest and NewFormRequest. We should extract into a helper.

github/repos_contents.go
```diff
@@ -278,6 +278,10 @@ func (s *RepositoriesService) GetContents(ctx context.Context, owner, repo, path
 //
 //meta:operation PUT /repos/{owner}/{repo}/contents/{path}
 func (s *RepositoriesService) CreateFile(ctx context.Context, owner, repo, path string, opts *RepositoryContentFileOptions) (*RepositoryContentResponse, *Response, error) {
+	if strings.Contains(path, "..") {
+		return nil, nil, ErrPathForbidden
+	}
```
alexandear: This is already handled by `s.client.NewRequest`. Do we need it here? The same for other RepositoriesService functions: GetContents, UpdateFile, DeleteFile.

github/github.go
```diff
@@ -551,6 +552,21 @@ func WithVersion(version string) RequestOption {
 	}
 }
 
+// urlContainsDotDotPathSegment reports whether urlStr contains ".." as a path
+// segment (e.g. "a/../b"). It does not match ".." embedded within a segment
+// (e.g. "file..txt"). The check is performed only on the path portion of the
+// URL, ignoring any query string or fragment.
+func urlContainsDotDotPathSegment(urlStr string) (bool, error) {
```
alexandear: Let's declare this unexported function after the `NewFormRequest` for readability.

alexandear (changes requested):

mohammadmseet-hue:
Thanks for the review @alexandear! Applied all 4 changes:

1. Moved `ErrPathForbidden` to `github.go`
2. Simplified to `checkURLPathTraversal(urlStr string) error` — eliminates the duplicated 4-line calling pattern across `NewRequest`/`NewFormRequest`/`NewUploadRequest`
3. Moved the function after `NewFormRequest` for readability
4. Removed redundant `..` checks from `GetContents`, `CreateFile`, `UpdateFile`, `DeleteFile` — `NewRequest` already handles it centrally

Net result: -101 lines added, +47 lines — cleaner and less duplication.

github/github.go
```diff
@@ -561,6 +566,10 @@ func (c *Client) NewRequest(method, urlStr string, body any, opts ...RequestOpti
 		return nil, fmt.Errorf("baseURL must have a trailing slash, but %q does not", c.BaseURL)
 	}
 
+	if err := checkURLPathTraversal(urlStr); err != nil {
+		return nil, err
+	}
+
 	u, err := c.BaseURL.Parse(urlStr)
```
alexandear, line 573: Here we are calling `url.Parse` twice. Could we eliminate the redundant parse?

mohammadmseet-hue: I believe the double `url.Parse` is actually necessary here. The fast path (`strings.Contains`) exits without any parsing for 99%+ of calls — `url.Parse` inside `checkURLPathTraversal` only runs when `..` is present, and in that case we almost always return `ErrPathForbidden` before reaching `c.BaseURL.Parse`.

We also can't move the check after `c.BaseURL.Parse` — it resolves `..` segments during URL resolution (e.g. `repos/x/../../../admin` → `/admin`), so the `..` would be gone and the check would never trigger. The traversal detection has to happen on the raw input before resolution.

The only case where both parses run is something like `repos/owner/file..txt` (contains `..` substring but not as a path segment) — edge case with negligible cost.

alexandear (approved):

stevehipwell (approved):
LGTM

munlicode (approved):

gmlewis:
Thank you, all!
Merging.

PR #4151, opened 2026-04-14
feat!: Refactor request context

### LINKED ISSUE #4127: Request context question
The current implementation of the client has the `NewRequest` method which uses `http.NewRequest` and then we use `withContext` which just wraps `req.WithContext` to copy the context. Is there a compelling reason why this pattern is used instead of using `http.NewRequestWithContext` which would save having to copy every request just to set the context which is known when the request is created?

stevehipwell:
BREAKING CHANGE: All internal calls now provide `Context` via the `Request` itself.

Closes #4127

This PR makes the following changes:

- Switch to using `http.NewRequestWithContext` so we no longer need the `withContext` helper
- Refactor all HTTP calls to use `client.BareDo` to get benefits such as rate limit support

### DIFF
--- a/github/actions_artifacts.go
+++ b/github/actions_artifacts.go
@@ -91,13 +91,13 @@ func (s *ActionsService) ListArtifacts(ctx context.Context, owner, repo string, 
 		return nil, nil, err
 	}
 
-	req, err := s.client.NewRequest("GET", u, nil)
+	req, err := s.client.NewRequest(ctx, "GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	var artifactList *ArtifactList
-	resp, err := s.client.Do(ctx, req, &artifactList)
+	resp, err := s.client.Do(req, &artifactList)
 	if err != nil {
 		return nil, resp, err
 	}
@@ -117,13 +117,13 @@ func (s *ActionsService) ListWorkflowRunArtifacts(ctx context.Context, owner, re
 		return nil, nil, err
 	}
 
-	req, err := s.client.NewRequest("GET", u, nil)
+	req, err := s.client.NewRequest(ctx, "GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	var artifactList *ArtifactList
-	resp, err := s.client.Do(ctx, req, &artifactList)
+	resp, err := s.client.Do(req, &artifactList)
 	if err != nil {
 		return nil, resp, err
 	}
@@ -139,13 +139,13 @@ func (s *ActionsService) ListWorkflowRunArtifacts(ctx context.Context, owner, re
 func (s *ActionsService) GetArtifact(ctx context.Context, owner, repo string, artifactID int64) (*Artifact, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/artifacts/%v", owner, repo, artifactID)
 
-	req, err := s.client.NewRequest("GET", u, nil)
+	req, err := s.client.NewRequest(ctx, "GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	var artifact *Artifact
-	resp, err := s.client.Do(ctx, req, &artifact)
+	resp, err := s.client.Do(req, &artifact)
 	if err != nil {
 		return nil, resp, err
 	}
@@ -188,12 +188,12 @@ func (s *ActionsService) downloadArtifactWithoutRateLimit(ctx context.Context, u
 }
 
 func (s *ActionsService) downloadArtifactWithRateLimit(ctx context.Context, u string, maxRedirects int) (*url.URL, *Response, error) {
-	req, err := s.client.NewRequest("GET", u, nil)
+	req, err := s.client.NewRequest(ctx, "GET", u, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	url, resp, err := s.client.bareDoUntilFound(ctx, req, maxRedirects)
+	url, resp, err := s.client.bareDoUntilFound(req, maxRedirects)
 	if err != nil {
 		return nil, resp, err
 	}
@@ -215,10 +215,10 @@ func (s *ActionsService) downloadArtifactWithRateLimit(ctx context.Context, u st
 func (s *ActionsService) DeleteArtifact(ctx context.Context, owner, repo string, artifactID int64) (*Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/actions/artifacts/%v", owner, repo, artifactID)
 
-	req, err := s.client.NewRequest("DELETE", u, nil)
+	req, err := s.client.NewRequest(ctx, "DELETE", u, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	return s.client.Do(ctx, req, nil)
+	return s.client.Do(req, nil)
 }
--- a/github/github.go
+++ b/github/github.go
@@ -1037,9 +1037,6 @@ func (c *Client) bareDo(caller *http.Client, req *http.Request) (*Response, erro
 // are supposed to read and close the response's Body. If rate limit is exceeded
 // and reset time is in the future, BareDo returns *RateLimitError immediately
 // without making a network API call.
-//
-// The provided ctx must be non-nil, if it is nil an error is returned. If it is
-// canceled or times out, ctx.Err() will be returned.
 func (c *Client) BareDo(req *http.Request) (*Response, error) {
 	return c.bareDo(c.client, req)
 }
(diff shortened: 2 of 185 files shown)


github/actions_artifacts.go
alexandear, line 100: It's a bit odd that both `NewRequest` and `Do` accept `ctx`.

stevehipwell: It is, but the logic for per request rate limiting uses the context. I could refactor it to use `req.Context()` but that seems like additional overhead when the context is already available to be passed in.

stevehipwell:
@gmlewis I'm going to completely rebase this PR as I need to split it into 3.

stevehipwell:
#4152 replaces the app engine removal part of this PR.

github/actions_artifacts.go
stevehipwell: I've made this change.

stevehipwell:
@gmlewis I've rebased this PR and it should be ready for review now.

stevehipwell:
@gmlewis I had to update the `extraneousnew` linter now that `client.Do` only has 2 args, could you please pay particular attention to that change as I don't have the full context there.

github/github.go
```diff
@@ -1048,8 +1040,8 @@ func (c *Client) bareDo(ctx context.Context, caller *http.Client, req *http.Requ
 //
 // The provided ctx must be non-nil, if it is nil an error is returned. If it is
 // canceled or times out, ctx.Err() will be returned.
-func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, error) {
-	return c.bareDo(ctx, c.client, req)
+func (c *Client) BareDo(req *http.Request) (*Response, error) {
```
gmlewis, line 1040: The comment for this method needs updating - you can remove lines 1040-1042.

github/github.go
```diff
@@ -1058,8 +1050,8 @@ func (c *Client) BareDo(ctx context.Context, req *http.Request) (*Response, erro
 //
 // The provided ctx must be non-nil, if it is nil an error is returned. If it is
 // canceled or times out, ctx.Err() will be returned.
-func (c *Client) bareDoIgnoreRedirects(ctx context.Context, req *http.Request) (*Response, error) {
-	return c.bareDo(ctx, c.clientIgnoreRedirects, req)
+func (c *Client) bareDoIgnoreRedirects(req *http.Request) (*Response, error) {
```
gmlewis, line 1047: Same for this one.

github/github.go
```diff
@@ -1071,8 +1063,8 @@ var errInvalidLocation = errors.New("invalid or empty Location header in redirec
 //
 // The provided ctx must be non-nil, if it is nil an error is returned. If it is
 // canceled or times out, ctx.Err() will be returned.
-func (c *Client) bareDoUntilFound(ctx context.Context, req *http.Request, maxRedirects int) (*url.URL, *Response, error) {
-	response, err := c.bareDoIgnoreRedirects(ctx, req)
+func (c *Client) bareDoUntilFound(req *http.Request, maxRedirects int) (*url.URL, *Response, error) {
```
gmlewis, line 1057: ...and this one.

github/github.go
```diff
@@ -1122,8 +1114,8 @@ func (c *Client) bareDoUntilFound(ctx context.Context, req *http.Request, maxRed
 //
 // The provided ctx must be non-nil, if it is nil an error is returned. If it
 // is canceled or times out, ctx.Err() will be returned.
-func (c *Client) Do(ctx context.Context, req *http.Request, v any) (*Response, error) {
-	resp, err := c.BareDo(ctx, req)
+func (c *Client) Do(req *http.Request, v any) (*Response, error) {
```
gmlewis, line 1105: ...and this one.

gmlewis (changes requested):
In case a force-push is performed, I'm going to stop my review here before examining the remaining 182 files.

stevehipwell:
@gmlewis I won't force push this, I only force pushed to realign after we split the PR up and I did provide a warning that I was doing it. Could you possibly review the linter, the only real changes are there and in `github.go`.

gmlewis:
> @gmlewis I won't force push this, I only force pushed to realign after we split the PR up and I did provide a warning that I was doing it. Could you possibly review the linter, the only real changes are there and in `github.go`.

OK, the linter LGTM.

stevehipwell:
@gmlewis I've made the requested docs changes.

github/github_test.go
```diff
@@ -1296,18 +1296,6 @@ func TestDo(t *testing.T) {
 	}
 }
 
-func TestDo_nilContext(t *testing.T) {
```
gmlewis, line 1299: Couldn't you keep this test but move the `nil` from the `Do` up into the `NewRequest` ?

gmlewis (changes requested):
One question, otherwise LGTM.

github/github_test.go
stevehipwell: I couldn't test that here as the nil check on context is done when creating the request, not on calling `client.Do`. I could have added tests for nil context in the `NewRequest` (and similar) function, but it'd just be testing the stdlib `http.NewRequestWithContext` behaviour.

gmlewis (approved):
LGTM

Not-Dhananjay-Mishra (approved):
LGTM 🚀

stevehipwell:
@gmlewis how do you want me to resolve the conflict?

gmlewis:
@stevehipwell - apologies, but could you please resolve conflicts before I merge?
Oh, and we need to hear from @alexandear too.

stevehipwell:
> @gmlewis how do you want me to resolve the conflict?

@gmlewis I'd normally rebase to fix a conflict like this, but I know you don't like the commits to be modified.

gmlewis:
> @gmlewis I'd normally rebase to fix a conflict like this, but I know you don't like the commits to be modified

1. Go to your `master` branch and `git pull`
2. Go to your branch and `git merge master`
3. Manually resolve conflicts or `git mergetool` if you have already set up a tool like `kdiff3`
4. `git push ...` to this PRs branch

stevehipwell:
@gmlewis merge commit it is, was just checking.

gmlewis:
Although if you have a better mechanism, I am fine with that in this huge PR.

stevehipwell:
@gmlewis assuming that you're squash merging and you'd rather the commits aren't modified, a merge commit is likely to be the best compromise.

stevehipwell:
@gmlewis it should be ready again.

stevehipwell:
@alexandear I think we're just waiting on you now?

alexandear (approved):

stevehipwell:
@gmlewis I think we're ready to merge now.

github/rate_limit.go
```diff
@@ -76,7 +76,12 @@ func (r RateLimits) String() string {
 //
 //meta:operation GET /rate_limit
 func (s *RateLimitService) Get(ctx context.Context) (*RateLimits, *Response, error) {
-	req, err := s.client.NewRequest("GET", "rate_limit", nil)
+	// This resource is not subject to rate limits.
+	if !s.client.DisableRateLimitCheck {
```
alexandear, line 80: Why is this added? Before this PR we don't have `if !s.client.DisableRateLimitCheck`

alexandear: If I comment out the check

Image

tests still pass:

```
❯ go test -failfast -count=1 ./...
ok      github.com/google/go-github/v85/github  2.472s
?       github.com/google/go-github/v85/test/fields     [no test files]
?       github.com/google/go-github/v85/test/integration        [no test files]
```

stevehipwell: It short circuits the context modification as it's never going to be retrieved with `s.client.DisableRateLimitCheck` set to `true`.

alexandear: I propose to move this change to a separate PR and add a test.

stevehipwell: I don't want to remove this change as it's demonstrably "correct". @alexandear if you have a suggestion for how to test this I'm happy to add a test, but it doesn't look testable to me.

alexandear: Should be something like the following:
```go
func TestRateLimits_bypassRateLimitCheckContext(t *testing.T) {
	t.Parallel()
	tests := []struct {
		name                  string
		disableRateLimitCheck bool
		wantBypassValue       any
	}{
		{
			name:                  "DisableRateLimitCheck disabled",
			disableRateLimitCheck: false,
			wantBypassValue:       true,
		},
		{
			name:                  "DisableRateLimitCheck enabled",
			disableRateLimitCheck: true,
			wantBypassValue:       nil,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			client, mux, _ := setup(t)
			client.DisableRateLimitCheck = tt.disableRateLimitCheck

			mux.HandleFunc("/rate_limit", func(w http.ResponseWriter, r *http.Request) {
				testMethod(t, r, "GET")
				fmt.Fprint(w, `{"resources":{}}`)
			})

			_, resp, err := client.RateLimit.Get(t.Context())
			if err != nil {
				t.Errorf("RateLimits returned error: %v", err)
			}

			if got, want := resp.Request.Context().Value(BypassRateLimitCheck), tt.wantBypassValue; got != want {
				t.Errorf("Request context bypass value = %v, want %v", got, want)
			}
		})
	}
}
```

stevehipwell: Thanks @alexandear, I missed the round tip of the context. I've added your test.

stevehipwell:
@alexandear I think this is waiting for your approval.

gmlewis:
> @alexandear I think this is waiting for your approval.

Yes, please, since you made a request after your last approval. Thanks.

alexandear:
Already approved. LGTM.

Issue #3915, opened 2026-01-13
Question on `DownloadReleaseAsset` redirect logic


stevehipwell:
I'm just looking at the implementation of `DownloadReleaseAsset` based on some code that calls is with a copy of the GitHub client as the `followRedirectsClient` and I'm trying to understand why the redirect is always disabled? Based on what I've seen on GitHub it looks like some assets were being server from AWS (9 years ago) which in turn was having issues with the headers.

I'm wondering if this implementation is outdated? At a minimum shouldn't redirects to `https://release-assets.githubusercontent.com` (or `github.com`) be handled by the primary client if no additional client is passed in? 

I think we'd also need to look at how GHES and GHEC work, but I'd expect that they both need authentication to access artifacts.

CC @gmlewis

gmlewis:
Without doing a deep dive into the history of this code using "Blame" to follow the thread of development over the years, I would guess that it is not enabled by default for backward compatibility, such that only those who wish to actually follow the redirects will do so. If a client just wants to know what the link is, they wouldn't enable this option.

So if changes were made to this behavior, we would have to make it clear that there is a breaking change, and still be able to support those who do not wish to follow the redirects, but instead only wish to know that they exist.

stevehipwell:
@gmlewis I saw mention of AWS and S3 failing when the GH token was in the headers but the linked GH docs appear to be missing. I agree that just getting the URL may be useful functionality but I'd expect a function called `DownloadReleaseAsset` to actually download the release asset. Would you be happy if I "fixed" `DownloadReleaseAsset` and added a new `GetReleaseDownloadURL`.

Also is there a reason why we don't copy the client to disable redirects rather than lock the active client and disable redirects on it?

gmlewis:
> Also is there a reason why we don't copy the client to disable redirects rather than lock the active client and disable redirects on it?

Yes, absolutely there was a leak of client credentials when we used to copy the client, and I would have to go back and search for the issue to explain it more fully.

As for your other question, I will have to devote more time into looking back at the history of `DownloadReleaseAsset` to answer your question about "fixing" it before I can confidently answer. I know it has had quite a history of it being "fixed" many times over the years, so this will probably take some time.

stevehipwell:
> Yes, absolutely there was a leak of client credentials when we used to copy the client, and I would have to go back and search for the issue to explain it more fully.

[...]

PR #4161, opened 2026-04-16
chore: Switch legacy redirect handling to new pattern

stevehipwell:
This PR switches the legacy redirect pattern that modifies the global client to use the new pattern introduced in #3411.

Related to #3915

### DIFF
--- a/github/migrations.go
+++ b/github/migrations.go
@@ -9,8 +9,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
-	"strings"
 )
 
 // MigrationService provides access to the migration related functions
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @stevehipwell!
Do we have any integration tests associated with these endpoints so that they can be validated that they functionally still work? If not, do you think this would be a good use of integration tests?
Otherwise, LGTM.

stevehipwell:
@gmlewis `TestRepositories_DownloadReleaseAsset` covers `DownloadReleaseAsset`. I don't think migration would be something to include in the integration tests due to permissions.

github/migrations.go
```diff
@@ -186,29 +184,18 @@ func (s *MigrationService) MigrationArchiveURL(ctx context.Context, org string,
 	if err != nil {
 		return "", err
 	}
-
 	req.Header.Set("Accept", mediaTypeMigrationsPreview)
```
Not-Dhananjay-Mishra, line 187:
```suggestion
```
Should we remove this? I can't find`application/vnd.github.wyandotte-preview+json` header in docs. same applies for other methods too in this file.

Image

stevehipwell: These have been left in for GHES support, so I think this is a separate discussion.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis (approved):
Thank you, @stevehipwell and @Not-Dhananjay-Mishra.
LGTM.
Merging.

PR #4149, opened 2026-04-11
feat: Add `DownloadCopilotMetrics` helper method

MasterOogway1466:
Added DownloadCopilotMetrics helper method to simulate previous implementation of viewing copilot usage metrics. Also, commented above the older function definitions to highlight that the methods are no longer in use due to the change in API endpoint. Fixes #4136

PR #4177, opened 2026-04-23
feat: Add typed Copilot metrics download helpers

kyungseopk1m:
Fixes #4136

## Summary
The new Copilot usage metrics endpoints (`/copilot/metrics/reports/*`) return download links whose payloads do not match the `CopilotMetrics` shape, so the existing `DownloadCopilotMetrics` helper fails to decode live data from the eight `Get*MetricsReport` methods on public GitHub.

This PR adds four typed helpers and deprecates the legacy one:

| Link returned by | Use |
|---|---|
| `Get{Enterprise,Organization}DailyMetricsReport` | `DownloadDailyMetrics` |
| `Get{Enterprise,Organization}MetricsReport` | `DownloadPeriodicMetrics` |
| `Get{Enterprise,Organization}UsersDailyMetricsReport` | `DownloadUserDailyMetrics` |
| `Get{Enterprise,Organization}UsersMetricsReport` | `DownloadUserPeriodicMetrics` |

User metrics payloads are newline-delimited JSON; aggregate payloads are JSON objects (1-day) or a `{ day_totals: [...] }` wrapper (28-day).

## Payload schemas
GitHub's public OpenAPI and REST documentation only describe the top-level response (`download_links`, `report_start_day`/`end_day`) and do not publish the schema of the downloaded files. The struct definitions in this PR are based on the live-data structs posted by @Roemer in https://github.com/google/go-github/pull/4149#issuecomment-4287547798 and the official Copilot usage metrics documentation. Follow-up PRs can adjust any field names that differ from what live endpoints emit.

## Notes
- `DownloadCopilotMetrics` is kept with a `Deprecated:` comment so GitHub Enterprise Server installations that may still serve the legacy shape continue to work.
- The four new helpers are registered in `tools/metadata/metadata.go` skip list, following the same pattern as the existing `DownloadCopilotMetrics` (helper methods without direct GitHub v3 API correspondence).
- Most of the added lines are auto-generated accessors via `script/generate.sh` for the new typed structs.

## Testing
- `script/fmt.sh`, `script/generate.sh`, `script/lint.sh`, `script/test.sh` all pass locally.
- Each helper covers success, HTTP status error, invalid URL, invalid scheme, malformed JSON; the NDJSON helpers additionally cover empty-body → nil slice behaviour.

cc @gmlewis

### DIFF
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -854,45 +854,56 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
 
 // CopilotMetricsPullRequests represents pull request totals in a Copilot metrics report.
 type CopilotMetricsPullRequests struct {
-	TotalReviewed                       int     `json:"total_reviewed"`
-	TotalCreated                        int     `json:"total_created"`
-	TotalCreatedByCopilot               int     `json:"total_created_by_copilot"`
-	TotalReviewedByCopilot              int     `json:"total_reviewed_by_copilot"`
-	TotalMerged                         int     `json:"total_merged"`
-	MedianMinutesToMerge                float64 `json:"median_minutes_to_merge"`
-	TotalSuggestions                    int     `json:"total_suggestions"`
-	TotalAppliedSuggestions             int     `json:"total_applied_suggestions"`
-	TotalMergedCreatedByCopilot         int     `json:"total_merged_created_by_copilot"`
-	MedianMinutesToMergeCopilotAuthored float64 `json:"median_minutes_to_merge_copilot_authored"`
-	TotalCopilotSuggestions             int     `json:"total_copilot_suggestions"`
-	TotalCopilotAppliedSuggestions      int     `json:"total_copilot_applied_suggestions"`
-	MedianMinutesToMergeCopilotReviewed float64 `json:"median_minutes_to_merge_copilot_reviewed"`
-	TotalMergedReviewedByCopilot        int     `json:"total_merged_reviewed_by_copilot"`
+	TotalReviewed                       *int     `json:"total_reviewed,omitempty"`
+	TotalCreated                        *int     `json:"total_created,omitempty"`
+	TotalCreatedByCopilot               *int     `json:"total_created_by_copilot,omitempty"`
+	TotalReviewedByCopilot              *int     `json:"total_reviewed_by_copilot,omitempty"`
+	TotalMerged                         *int     `json:"total_merged,omitempty"`
+	MedianMinutesToMerge                *float64 `json:"median_minutes_to_merge,omitempty"`
+	TotalSuggestions                    *int     `json:"total_suggestions,omitempty"`
+	TotalAppliedSuggestions             *int     `json:"total_applied_suggestions,omitempty"`
+	TotalMergedCreatedByCopilot         *int     `json:"total_merged_created_by_copilot,omitempty"`
+	MedianMinutesToMergeCopilotAuthored *float64 `json:"median_minutes_to_merge_copilot_authored,omitempty"`
+	TotalCopilotSuggestions             *int     `json:"total_copilot_suggestions,omitempty"`
+	TotalCopilotAppliedSuggestions      *int     `json:"total_copilot_applied_suggestions,omitempty"`
+	MedianMinutesToMergeCopilotReviewed *float64 `json:"median_minutes_to_merge_copilot_reviewed,omitempty"`
+	TotalMergedReviewedByCopilot        *int     `json:"total_merged_reviewed_by_copilot,omitempty"`
 }
 
 // CopilotMetricsCodeActivity captures the code-generation activity counts and lines-of-code (LOC)
 // suggestion totals shared across the per-IDE, per-feature, per-language, and per-model breakdowns
 // in a Copilot metrics report.
 type CopilotMetricsCodeActivity struct {
-	CodeGenerationActivityCount int `json:"code_generation_activity_count"`
-	CodeAcceptanceActivityCount int `json:"code_acceptance_activity_count"`
-	LOCSuggestedToAddSum        int `json:"loc_suggested_to_add_sum"`
-	LOCSuggestedToDeleteSum     int `json:"loc_suggested_to_delete_sum"`
-	LOCAddedSum                 int `json:"loc_added_sum"`
-	LOCDeletedSum               int `json:"loc_deleted_sum"`
+	CodeGenerationActivityCount *int `json:"code_generation_activity_count,omitempty"`
+	CodeAcceptanceActivityCount *int `json:"code_acceptance_activity_count,omitempty"`
+	LOCSuggestedToAddSum        *int `json:"loc_suggested_to_add_sum,omitempty"`
+	LOCSuggestedToDeleteSum     *int `json:"loc_suggested_to_delete_sum,omitempty"`
+	LOCAddedSum                 *int `json:"loc_added_sum,omitempty"`
+	LOCDeletedSum               *int `json:"loc_deleted_sum,omitempty"`
+}
+
+// CopilotMetricsChatPanel captures per-mode chat panel interaction counts shared across daily,
+// periodic, and user metrics reports. Each field is a subset of user_initiated_interaction_count
+// attributed to that chat panel mode.
+type CopilotMetricsChatPanel struct {
+	ChatPanelAgentMode   *int `json:"chat_panel_agent_mode,omitempty"`
+	ChatPanelAskMode     *int `json:"chat_panel_ask_mode,omitempty"`
+	ChatPanelCustomMode  *int `json:"chat_panel_custom_mode,omitempty"`
+	ChatPanelEditMode    *int `json:"chat_panel_edit_mode,omitempty"`
+	ChatPanelUnknownMode *int `json:"chat_panel_unknown_mode,omitempty"`
 }
 
 // CopilotMetricsIDE represents per-IDE aggregate totals in a Copilot metrics report.
 type CopilotMetricsIDE struct {
 	IDE                           string `json:"ide"`
-	UserInitiatedInteractionCount int    `json:"user_initiated_interaction_count"`
+	UserInitiatedInteractionCount *int   `json:"user_initiated_interaction_count,omitempty"`
 	CopilotMetricsCodeActivity
 }
 
 // CopilotMetricsFeature represents per-feature aggregate totals in a Copilot metrics report.
 type CopilotMetricsFeature struct {
 	Feature                       string `json:"feature"`
-	UserInitiatedInteractionCount int    `json:"user_initiated_interaction_count"`
+	UserInitiatedInteractionCount *int   `json:"user_initiated_interaction_count,omitempty"`
 	CopilotMetricsCodeActivity
 }
 
@@ -914,7 +925,7 @@ type CopilotMetricsLanguageModel struct {
 type CopilotMetricsModelFeature struct {
 	Model                         string `json:"model"`
 	Feature                       string `json:"feature"`
-	UserInitiatedInteractionCount int    `json:"user_initiated_interaction_count"`
+	UserInitiatedInteractionCount *int   `json:"user_initiated_interaction_count,omitempty"`
 	CopilotMetricsCodeActivity
 }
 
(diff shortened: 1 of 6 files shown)


github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	LocSuggestedToAddSum          int    `json:"loc_suggested_to_add_sum"`
+	LocSuggestedToDeleteSum       int    `json:"loc_suggested_to_delete_sum"`
+	LocAddedSum                   int    `json:"loc_added_sum"`
+	LocDeletedSum                 int    `json:"loc_deleted_sum"`
+}
+
+// CopilotMetricsCliVersion represents the last known Copilot CLI version seen in a metrics report.
+type CopilotMetricsCliVersion struct {
+	SampledAt  *Timestamp `json:"sampled_at,omitempty"`
+	CliVersion string     `json:"cli_version"`
```
gmlewis: Please add a new "initialism" to line 414 of `tools/structfield/structfield.go` to make "CLI" a recognized acronym, and change this line:


```suggestion
	CLIVersion string     `json:"cli_version"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	MonthlyActiveCopilotCloudAgentUsers int                              `json:"monthly_active_copilot_cloud_agent_users"`
+	UserInitiatedInteractionCount       int                              `json:"user_initiated_interaction_count"`
+	CodeGenerationActivityCount         int                              `json:"code_generation_activity_count"`
+	CodeAcceptanceActivityCount         int                              `json:"code_acceptance_activity_count"`
+	TotalsByIDE                         []*CopilotMetricsIDE             `json:"totals_by_ide,omitempty"`
+	TotalsByFeature                     []*CopilotMetricsFeature         `json:"totals_by_feature,omitempty"`
+	TotalsByLanguageFeature             []*CopilotMetricsLanguageFeature `json:"totals_by_language_feature,omitempty"`
+	TotalsByLanguageModel               []*CopilotMetricsLanguageModel   `json:"totals_by_language_model,omitempty"`
+	TotalsByModelFeature                []*CopilotMetricsModelFeature    `json:"totals_by_model_feature,omitempty"`
+	TotalsByCli                         *CopilotMetricsCli               `json:"totals_by_cli,omitempty"`
```
gmlewis:
```suggestion
	TotalsByCLI                         *CopilotMetricsCLI               `json:"totals_by_cli,omitempty"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	EnterpriseID                  string                           `json:"enterprise_id"`
+	UserInitiatedInteractionCount int                              `json:"user_initiated_interaction_count"`
+	CodeGenerationActivityCount   int                              `json:"code_generation_activity_count"`
+	CodeAcceptanceActivityCount   int                              `json:"code_acceptance_activity_count"`
+	TotalsByIDE                   []*CopilotUserMetricsIDE         `json:"totals_by_ide,omitempty"`
+	TotalsByFeature               []*CopilotMetricsFeature         `json:"totals_by_feature,omitempty"`
+	TotalsByLanguageFeature       []*CopilotMetricsLanguageFeature `json:"totals_by_language_feature,omitempty"`
+	TotalsByLanguageModel         []*CopilotMetricsLanguageModel   `json:"totals_by_language_model,omitempty"`
+	TotalsByModelFeature          []*CopilotMetricsModelFeature    `json:"totals_by_model_feature,omitempty"`
+	TotalsByCli                   *CopilotMetricsCli               `json:"totals_by_cli,omitempty"`
```
gmlewis:
```suggestion
	TotalsByCLI                   *CopilotMetricsCLI               `json:"totals_by_cli,omitempty"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	UsedAgent                     bool                             `json:"used_agent"`
+	UsedChat                      bool                             `json:"used_chat"`
+	UsedCli                       bool                             `json:"used_cli"`
```
gmlewis:
```suggestion
	UsedCLI                       bool                             `json:"used_cli"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	UserLogin                     string                           `json:"user_login"`
+	UserInitiatedInteractionCount int                              `json:"user_initiated_interaction_count"`
+	CodeGenerationActivityCount   int                              `json:"code_generation_activity_count"`
+	CodeAcceptanceActivityCount   int                              `json:"code_acceptance_activity_count"`
+	TotalsByIDE                   []*CopilotUserMetricsIDE         `json:"totals_by_ide,omitempty"`
+	TotalsByFeature               []*CopilotMetricsFeature         `json:"totals_by_feature,omitempty"`
+	TotalsByLanguageFeature       []*CopilotMetricsLanguageFeature `json:"totals_by_language_feature,omitempty"`
+	TotalsByLanguageModel         []*CopilotMetricsLanguageModel   `json:"totals_by_language_model,omitempty"`
+	TotalsByModelFeature          []*CopilotMetricsModelFeature    `json:"totals_by_model_feature,omitempty"`
+	TotalsByCli                   *CopilotMetricsCli               `json:"totals_by_cli,omitempty"`
```
gmlewis:
```suggestion
	TotalsByCLI                   *CopilotMetricsCLI               `json:"totals_by_cli,omitempty"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	UsedAgent                     bool                             `json:"used_agent"`
+	UsedChat                      bool                             `json:"used_chat"`
+	UsedCli                       bool                             `json:"used_cli"`
```
gmlewis:
```suggestion
	UsedCLI                       bool                             `json:"used_cli"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	CodeGenerationActivityCount   int    `json:"code_generation_activity_count"`
+	CodeAcceptanceActivityCount   int    `json:"code_acceptance_activity_count"`
+	LocSuggestedToAddSum          int    `json:"loc_suggested_to_add_sum"`
+	LocSuggestedToDeleteSum       int    `json:"loc_suggested_to_delete_sum"`
+	LocAddedSum                   int    `json:"loc_added_sum"`
+	LocDeletedSum                 int    `json:"loc_deleted_sum"`
+}
+
+// CopilotMetricsCliVersion represents the last known Copilot CLI version seen in a metrics report.
+type CopilotMetricsCliVersion struct {
```
gmlewis:
```suggestion
// CopilotMetricsCLIVersion represents the last known Copilot CLI version seen in a metrics report.
type CopilotMetricsCLIVersion struct {
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	SampledAt  *Timestamp `json:"sampled_at,omitempty"`
+	CliVersion string     `json:"cli_version"`
+}
+
+// CopilotMetricsCliTokenUsage represents Copilot CLI token totals in a metrics report.
+type CopilotMetricsCliTokenUsage struct {
```
gmlewis: There are lots of "Cli"s in this PR in Go names... please change _ALL_ of them to "CLI" for the Go names. I'm going to stop commenting on them, but I would appreciate if you find them all and fix them.

gmlewis (changes requested):
Thank you, @kyungseopk1m!
I'm stopping the review now, but please address the concerns and I will continue later when I have more time.

kyungseopk1m:
Thanks for the review, @gmlewis! I've addressed all the points in 4cd1270 on this branch.

gmlewis (approved):
Thank you, @kyungseopk1m!
LGTM.

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+// CopilotMetricsIDE represents per-IDE aggregate totals in a Copilot metrics report.
+type CopilotMetricsIDE struct {
+	IDE                           string `json:"ide"`
+	UserInitiatedInteractionCount int    `json:"user_initiated_interaction_count"`
+	CodeGenerationActivityCount   int    `json:"code_generation_activity_count"`
+	CodeAcceptanceActivityCount   int    `json:"code_acceptance_activity_count"`
+	LocSuggestedToAddSum          int    `json:"loc_suggested_to_add_sum"`
+	LocSuggestedToDeleteSum       int    `json:"loc_suggested_to_delete_sum"`
+	LocAddedSum                   int    `json:"loc_added_sum"`
+	LocDeletedSum                 int    `json:"loc_deleted_sum"`
```
alexandear: LOC is the abbreviation for "Lines of code":
```suggestion
	LOCSuggestedToAddSum          int    `json:"loc_suggested_to_add_sum"`
	LOCSuggestedToDeleteSum       int    `json:"loc_suggested_to_delete_sum"`
	LOCAddedSum                   int    `json:"loc_added_sum"`
	LOCDeletedSum                 int    `json:"loc_deleted_sum"`
```

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+// CopilotMetricsLanguageModel represents per-language-model totals in a Copilot metrics report.
+type CopilotMetricsLanguageModel struct {
+	Language                    string `json:"language"`
+	Model                       string `json:"model"`
+	CodeGenerationActivityCount int    `json:"code_generation_activity_count"`
+	CodeAcceptanceActivityCount int    `json:"code_acceptance_activity_count"`
+	LocSuggestedToAddSum        int    `json:"loc_suggested_to_add_sum"`
+	LocSuggestedToDeleteSum     int    `json:"loc_suggested_to_delete_sum"`
+	LocAddedSum                 int    `json:"loc_added_sum"`
+	LocDeletedSum               int    `json:"loc_deleted_sum"`
```
alexandear: These fields are the same for `CopilotMetricsIDE`, `CopilotMetricsLanguageFeature`, `CopilotMetricsModelFeature`. Maybe we can move them to a separate struct?

github/copilot.go
```diff
@@ -890,3 +851,348 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+		return nil, r, err
+	}
+	defer resp.Body.Close()
+
+	var metrics *CopilotDailyMetrics
+	if err := json.NewDecoder(resp.Body).Decode(&metrics); err != nil {
+		return nil, r, err
+	}
+
+	return metrics, r, nil
```
alexandear, line 1137: I'm curious: can we use the same implementation style for download metrics endpoints as for other endpoints in the repo?

Something like:

```suggestion
	req, err := s.client.NewRequest("GET", url, nil)
	if err != nil {
		return nil, nil, err
	}

	var metrics *CopilotDailyMetrics
	resp, err := s.client.Do(ctx, req, &metrics)
	if err != nil {
		return nil, resp, err
	}

	return metrics, resp, nil
```

kyungseopk1m: The pattern matches DownloadCopilotMetrics (copilot.go:837) and DownloadContentsWithMeta (repos_contents.go:178) — both download from a server-returned URL using raw http.NewRequestWithContext + s.client.client.Do(req), and I wanted these four to stay consistent with that. Two of them are also NDJSON, which s.client.Do(ctx, req, &v) can't decode in a single call. Happy to revisit if you'd prefer otherwise.

kyungseopk1m:
@alexandear Done in 90c12088 — Loc* → LOC* (added LOC to the initialism map) and CopilotMetricsCodeActivity embedded in the six smaller per-breakdown structs. Kept CopilotDailyMetrics, CopilotUserDailyMetrics, CopilotUserPeriodicMetrics inline since embedding a 6-field chunk in the middle of a 20+ field struct didn't help readability. Let me know if you'd prefer them uniform too.

Not-Dhananjay-Mishra (commented):
I am curious why aren't we using `*` + `omitempty` for all fields since there is no official response schema present, so we don't know which fields will always be returned.
example - `DailyActiveCLIUsers` is present in struct but it is absent from [comment](https://github.com/google/go-github/pull/4149#issuecomment-4287547798).

and some fields like `chat_panel_agent_mode`, `chat_panel_ask_mode`, `chat_panel_custom_mode`, `chat_panel_edit_mode` and `chat_panel_unknown_mode` are missing from structs but present in [Data available in Copilot usage metrics docs](https://docs.github.com/en/copilot/reference/copilot-usage-metrics/copilot-usage-metrics)

kyungseopk1m:
Fair points, thanks. Will switch the int/bool fields to `*T` + `omitempty` to match `CopilotMetrics`, and add the missing `chat_panel_*` fields.

github/copilot.go
alexandear: @kyungseopk1m thanks for the explanation.

@gmlewis is it possible to hit rate limits when downloading metrics? The raw `http.NewRequestWithContext + s.client.client.Do(req)` approach doesn't handle that. That's why I proposed using our custom HTTP client.

gmlewis: Great question. Sorry, but I don't know.

This question should probably be sent to the official GitHub tech support team.

gmlewis (approved):
Awaiting reviews from @alexandear and @Not-Dhananjay-Mishra before merging.

github/copilot.go
```diff
@@ -863,10 +820,14 @@ func (s *CopilotService) GetOrganizationUsersMetricsReport(ctx context.Context,
...
-// GetEnterpriseUsersMetricsReport, GetOrganizationDailyMetricsReport, GetOrganizationMetricsReport,
-// GetOrganizationUsersDailyMetricsReport, GetOrganizationUsersMetricsReport.
+// and decodes it as a []*CopilotMetrics.
+//
+// Deprecated: the payloads served at the download links returned by the new
+// Get*MetricsReport endpoints on GitHub.com do not match the CopilotMetrics shape
+// (see https://github.com/google/go-github/issues/4136). Use DownloadDailyMetrics,
+// DownloadPeriodicMetrics, DownloadUserDailyMetrics, or DownloadUserPeriodicMetrics
+// depending on which Get*MetricsReport produced the link. This method is retained
+// for GitHub Enterprise Server installations that may still serve the legacy shape.
```
alexandear: Let's improve a bit:

```suggestion
// Deprecated: Use DownloadDailyMetrics,
// DownloadPeriodicMetrics, DownloadUserDailyMetrics, DownloadUserPeriodicMetrics instead.
// The payloads served at the download links returned by the new
// Get*MetricsReport endpoints on GitHub.com do not match the CopilotMetrics shape
// (see https://github.com/google/go-github/issues/4136).
// This method is retained
// for GitHub Enterprise Server installations that may still serve the legacy shape.
```

github/copilot.go
```diff
@@ -890,3 +851,344 @@ func (s *CopilotService) DownloadCopilotMetrics(ctx context.Context, url string)
...
+	for {
+		var rec *CopilotUserPeriodicMetrics
+		if err := dec.Decode(&rec); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, r, err
+		}
+		records = append(records, rec)
+	}
```
alexandear: I propose moving this logic to a separate, unexported helper function. Because it's used in all DownloadMetrics functions.

kyungseopk1m: @alexandear Thanks for the suggestion.
Done in b05624b3 — pulled the NDJSON loop into decodeNDJSONMetrics[T any]. Left the single-object download helpers alone since there's nothing shared there.

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Awaiting review from @Not-Dhananjay-Mishra before merging.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

PR #4178, opened 2026-04-24
fix: Handle string-typed reviewer `ID` in Ruleset API responses

mmelodyRTR:
## Summary
This PR addresses a JSON unmarshaling error in `RulesetReviewer.ID` when interacting with the GitHub Repository Ruleset API.

While the GitHub API documentation specifies that the reviewer ID should be an integer, the API currently returns these IDs as quoted strings (e.g., `"id": "123456"`). This causes the standard Go JSON decoder to fail when unmarshaling into the `*int64` field.

## Related Issues
This issue was identified via the GitHub Terraform Provider: [integrations/terraform-provider-github/issues/3214](https://github.com/integrations/terraform-provider-github/issues/3214)

## Changes
- Implemented a custom `UnmarshalJSON` for `RulesetReviewer` in `github/rules.go`.
- The unmarshaler robustly handles both numeric (float64) and string-quoted IDs, ensuring compatibility with both the documented and actual API behavior.
- Maintained the `*int64` type for `RulesetReviewer.ID` to preserve consistency with the rest of the library and avoid breaking changes for existing users.
- Added a regression test case to `TestRepositoryRule` and a new comprehensive unit test suite `TestRulesetReviewer_UnmarshalJSON` in `github/rules_test.go`.

## Verification
- Verified the API behavior using `gh api` against a live repository.
- Ran all tests in `github/rules_test.go`, which now pass with the fix.


mmelodyRTR:
**Full disclosure:** I'm not a go developer and have used agentic ai/coding to drive this implementation. Will take any feedback on board in an effort to provide a _proper_ go solution in line with repository standards

gmlewis (approved):
Thank you, @mmelodyRTR!
LGTM.

gmlewis:
@mmelodyRTR - if you want to make the PR even better, Codecov is reporting that the following line is not covered by a unit test. You could ask the agent to see if it can come up with a unit test that would cover the failure on line 498:

uncovered

mmelodyRTR:
Thanks for the feedback @gmlewis - added a test case as requested 🫡

github/rules_test.go
```diff
@@ -1172,3 +1201,64 @@ func TestRepositoryRule(t *testing.T) {
...
+			name:    "invalid_type_id",
+			json:    `{"id": {}, "type": "Team"}`,
+			wantErr: true,
+		},
+		{
+			name:    "malformed_json",
+			json:    `{"id":`,
+			wantErr: true,
+		},
+	}
```
Not-Dhananjay-Mishra, line 1241:
```suggestion
		},
		{
			name:    "invalid_type",
			json:    `{"id": 123456, "type": 123456}`,
			wantErr: true,
		},
	}
```
This test case might cover those lines.

Image

Not-Dhananjay-Mishra (approved):
LGTM

github/rules_test.go
gmlewis: Codecov still claims it is not covered, but that's OK. We can proceed anyway.

gmlewis:
Thank you, @mmelodyRTR and @Not-Dhananjay-Mishra!

mmelodyRTR:
Thanks folks :)

PR #4187, opened 2026-05-01
chore: Fix `otel` module name

### LINKED ISSUE #4185: Can't import the `otel` package
I'm trying to import the [`otel`](https://github.com/google/go-github/tree/master/otel) package from this repository into a Go program, but I'm unable to do so. I tried the following commands:

```console
❯ go get github.com/google/go-github/v85/otel
go: module github.com/google/go-github/v85@upgrade found (v85.0.0), but does not contain package 
github.com/google/go-github/v85/otel

❯ go get github.com/google/go-github/otel
go: downloading github.com/google/go-github/otel v0.0.0-20260430203823-b7b1f86d59fa
go: github.com/google/go-github/otel@upgrade (v0.0.0-20260430203823-b7b1f86d59fa) requires github.com/google/go-github/otel@v0.0.0-20260430203823-b7b1f86d59fa: parsing go.mod:
        module declares its path as: github.com/google/go-github/v85/otel
                but was required as: github.com/google/go-github/otel
```

It looks like the issue is caused by missing `otel/vX.Y.Z` Git tags in the repository. According to the ["Mapping versions to commits"](https://go.dev/ref/mod#vcs-version) documentation:

> If a module is defined in a subdirectory within the repository, that is, the module subdirectory portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash. For example, the module golang.org/x/tools/gopls is defined in the gopls subdirectory of the repository with root path golang.org/x/tools. The version v0.4.0 of that module must have the tag named gopls/v0.4.0 in that repository.

So for `github.com/google/go-github/v85/otel`, Go expects tags like `otel/vX.Y.Z`. The `otel/` prefix is mandatory.

Since the `otel` package can't currently be imported, I'm wondering if anyone is actually using it. If not, perhaps it could be removed? It has been added in #3938.

alexandear:
Fixes: #4185
Closes: #4186

I think the issue is the `otel` module name: it should be `github.com/google/go-github/otel/v85`, not `github.com/google/go-github/v85/otel`.

See modules and tags in https://github.com/thepudds/nested-module-example

### DIFF
--- a/example/go.mod
+++ b/example/go.mod
@@ -7,8 +7,8 @@ require (
 	github.com/bradleyfalzon/ghinstallation/v2 v2.18.0
 	github.com/gofri/go-github-pagination v1.0.1
 	github.com/gofri/go-github-ratelimit/v2 v2.0.2
+	github.com/google/go-github/otel/v85 v85.0.0-00010101000000-000000000000
 	github.com/google/go-github/v85 v85.0.0
-	github.com/google/go-github/v85/otel v0.0.0-00010101000000-000000000000
 	github.com/sigstore/sigstore-go v1.1.4
 	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.43.0
 	go.opentelemetry.io/otel/sdk v1.43.0
@@ -90,4 +90,4 @@ require (
 // Use version at HEAD, not the latest published.
 replace github.com/google/go-github/v85 => ../
 
-replace github.com/google/go-github/v85/otel => ../otel
+replace github.com/google/go-github/otel/v85 => ../otel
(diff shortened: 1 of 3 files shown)


gmlewis:
Then maybe we don't need the `go.work`?
I would love to NOT add that file.

example/go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	github.com/bradleyfalzon/ghinstallation/v2 v2.18.0
 	github.com/gofri/go-github-pagination v1.0.1
 	github.com/gofri/go-github-ratelimit/v2 v2.0.2
+	github.com/google/go-github/otel/v85 v85.0.0-00010101000000-000000000000
```
gmlewis: Can you please try changing this to `v85.0.0`?

alexandear: Done

alexandear:
> Then maybe we don't need the `go.work`? I would love to NOT add that file.

Yes. I think we don't need to add the `go.work`.

gmlewis (approved):
Awesome, thank you, @alexandear!
LGTM.
Merging.

PR #2389, opened 2022-06-18
Add MembersCanForkPrivateRepositories field to Organization struct

lmt-swallow:
This PR adds `MembersCanForkPrivateRepositories` field to `Organization` struct.

PR #4188, opened 2026-05-02
feat: Add `deploy_keys_enabled_for_repositories` and secret scanning custom link fields to `Organization` struct

JamBalaya56562:
This PR adds three fields to the `Organization` struct that are returned by `GET /orgs/{org}` and accepted by `PATCH /orgs/{org}`, but are not currently represented in go-github:

- `deploy_keys_enabled_for_repositories` (`*bool`) — controls whether deploy keys may be added and used for repositories in the organization. This is documented as a writable parameter in the [Update an organization](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#update-an-organization) REST docs.
- `secret_scanning_push_protection_custom_link_enabled` (`*bool`) — toggles whether a custom link is shown to contributors blocked by secret scanning push protection.
- `secret_scanning_push_protection_custom_link` (`*string`) — the URL displayed to contributors blocked by secret scanning push protection.

The latter two are not currently listed in the REST docs, but are observed in real `GET` and `PATCH` responses.

Verification: I confirmed against a real organization (`aletheia-works`) that `GET /orgs/{org}` and `PATCH /orgs/{org}` return all three of these fields. The new fields are placed alongside semantically related existing fields (`MembersCanForkPrivateRepos` and `SecretScanningValidityChecksEnabled`).

The accessor and stringify test files were regenerated with `go generate ./...`. `go test ./...` passes.

This follows the same minimal-diff pattern as #2389, which added `MembersCanForkPrivateRepositories`.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -22950,6 +22950,14 @@ func (o *Organization) GetDependencyGraphEnabledForNewRepos() bool {
 	return *o.DependencyGraphEnabledForNewRepos
 }
 
+// GetDeployKeysEnabledForRepositories returns the DeployKeysEnabledForRepositories field if it's non-nil, zero value otherwise.
+func (o *Organization) GetDeployKeysEnabledForRepositories() bool {
+	if o == nil || o.DeployKeysEnabledForRepositories == nil {
+		return false
+	}
+	return *o.DeployKeysEnabledForRepositories
+}
+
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.
Merging.

PR #4189, opened 2026-05-06
github: stop cross-host redirects to prevent bearer token leakage

### LINKED ISSUE #4190: github: WithAuthToken leaks bearer token on cross-host redirects
WithAuthToken wraps the Transport with a RoundTripper that injects
Authorization: Bearer TOKEN on every RoundTrip call. When http.Client
follows a redirect to a different host, Go strips the Authorization
header — but the wrapped Transport immediately re-adds it, so the token
is forwarded to the redirect destination.

The GitHub API redirects several endpoints cross-host:
- Repository archive downloads redirect to codeload.github.com
- Release asset downloads redirect to objects.githubusercontent.com

Any application using WithAuthToken that follows such redirects, or that
accepts user-provided URLs, will leak its GitHub token to the redirect
destination.

Fix tracked in google/go-github#4189.

evilgensec:
WithAuthToken wraps the Transport with a RoundTripper that unconditionally
injects Authorization: Bearer TOKEN on every RoundTrip call, including
intermediate calls that http.Client makes when following redirects.
Go's http.Client strips the Authorization header before building a
cross-host redirect request, but the wrapped Transport immediately
re-adds it, forwarding the token to every host in the redirect chain.

The GitHub API redirects certain endpoints to different hosts. Repository
archive downloads are redirected to codeload.github.com and release
asset downloads are redirected to objects.githubusercontent.com. Any
application that calls an endpoint that redirects cross-host, or that
accepts user-provided URLs, will leak its GitHub token to the redirect
destination.

Fix: add a CheckRedirect hook in WithAuthToken that returns
http.ErrUseLastResponse when the redirect target host differs from the
initial request host. This surfaces the 3xx to the caller instead of
following it, preventing token injection into cross-host requests. The
hook composes correctly with any CheckRedirect already set on the
underlying http.Client.

Same-host redirects continue to work as before.

New tests:
- cross-host redirect does not leak token: verifies 302 is returned and
  the redirect destination does not receive the Authorization header.
- same-host redirect is followed: verifies backward compatibility for
  same-host redirects.

Fixes google/go-github#4190

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -373,6 +373,25 @@ func (c *Client) WithAuthToken(token string) *Client {
 			return transport.RoundTrip(req)
 		},
 	)
+	// Prevent the bearer token from being forwarded to a different host on
+	// redirect. The Transport above re-injects the Authorization header on
+	// every RoundTrip call, including the intermediate calls that http.Client
+	// makes when following redirects. Go's http.Client strips Authorization
+	// for cross-host redirects, but the Transport immediately adds it back,
+	// so the token would be sent to the redirect destination. Returning
+	// http.ErrUseLastResponse surfaces the 3xx to the caller instead.
+	callerRedirect := c2.client.CheckRedirect
+	c2.client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		if callerRedirect != nil {
+			if err := callerRedirect(req, via); err != nil {
+				return err
+			}
+		}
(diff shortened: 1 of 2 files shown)


gmlewis:
@evilgensec - I believe this issue was already resolved by #4171.
cc: @mohammadmseet-hue - do you agree?

evilgensec:
Confirmed duplicate of #4171 — closing. Thanks for the pointer!

PR #4191, opened 2026-05-07
fix: Limit HTTP error response body reads to prevent OOM

evilgensec:
## Problem

Two paths in `github.go` read HTTP error response bodies with no size limit:

```go
// CheckResponse — fires on every non-2xx API response
data, err := io.ReadAll(r.Body)

// AcceptedError handler in bareDo — fires on 202 Accepted
b, readErr := io.ReadAll(resp.Body)
```

A server returning a very large error body causes the client process to exhaust memory. The webhook payload path already applies a 25 MiB limit via `io.LimitReader` in `messages.go`; the error-response paths had no equivalent guard.

## Fix

Wrap both reads with `io.LimitReader` capped at `maxErrorBodySize` (1 MiB):

```go
data, err := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySize))
b, readErr := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
```

All existing tests pass.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -920,6 +920,11 @@ const (
 	SleepUntilPrimaryRateLimitResetWhenRateLimited
 )
 
+// maxErrorBodySize is the maximum number of bytes read from an HTTP error
+// response body. Limits memory allocation when a server returns an
+// unexpectedly large error body.
+const maxErrorBodySize = 1 * 1024 * 1024 // 1 MiB
+
 // bareDo sends an API request using `caller` http.Client passed in the parameters
 // and lets you handle the api response. If an error or API Error occurs, the error
 // will contain more information. Otherwise, you are supposed to read and close the
@@ -997,7 +1002,7 @@ func (c *Client) bareDo(caller *http.Client, req *http.Request) (*Response, erro
 		// Issue #1022
 		var aerr *AcceptedError
 		if errors.As(err, &aerr) {
-			b, readErr := io.ReadAll(resp.Body)
+			b, readErr := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
 			if readErr != nil {
 				return response, readErr
 			}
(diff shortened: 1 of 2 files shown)


github/github.go
```diff
@@ -373,6 +373,25 @@ func (c *Client) WithAuthToken(token string) *Client {
...
+		if callerRedirect != nil {
+			if err := callerRedirect(req, via); err != nil {
+				return err
+			}
+		}
+		if len(via) > 0 && req.URL.Host != via[0].URL.Host {
+			return http.ErrUseLastResponse
+		}
+		return nil
+	}
```
alexandear: This is off-topic from #4189 and should be removed.

github/github.go
```diff
@@ -997,7 +1021,7 @@ func (c *Client) bareDo(caller *http.Client, req *http.Request) (*Response, erro
 		// Issue #1022
 		var aerr *AcceptedError
 		if errors.As(err, &aerr) {
-			b, readErr := io.ReadAll(resp.Body)
+			b, readErr := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
```
alexandear, line 1005: Is it possible to add a test?

evilgensec: added a test as requested

github/github.go
gmlewis: This in unresolved and should be removed.

evilgensec: Done.

github/github.go
alexandear: @evilgensec I don't see the test. Did you push the changes?

evilgensec: @alexandear sorry for the confusion, should be visible now.

alexandear (approved):

gmlewis (approved):
Thank you, @evilgensec and @alexandear!
LGTM.
Merging.

PR #4192, opened 2026-05-07
feat: Add repo download contents sentinel errors

stevehipwell:
This PR adds Sentinel errors to the `RepositoriesService.DownloadContents` & `RepositoriesService.DownloadContentsWithMeta` functions. This is a follow-up to #4153 that improves the ergonomics. These sentinels are required to correctly handle the different paths in these functions without falling back to writing logic against the error messages which is brittle (and are changing in the next release).

@gmlewis since working on #4153 I've found more issues with the pattern it replaced such as always returning `file` as the type for a submodule due to the directory loop pattern and a "bug" in the 2022 API version. This combined with a lack of Sentinel errors makes using the current released version painful, would it be possible to get a release out once this PR is merged?

### DIFF
--- a/github/repos_contents.go
+++ b/github/repos_contents.go
@@ -20,6 +20,15 @@ import (
 	"strings"
 )
 
+// ErrContentsDirectory indicates that the contents are not available for a directory.
+var ErrContentsDirectory = errors.New("contents not available for directory")
+
+// ErrContentsSubmodule indicates that the contents are not available for a submodule.
+var ErrContentsSubmodule = errors.New("contents not available for submodule")
+
+// ErrContentsNoDownloadURL indicates that the contents download URL is empty, which may occur when file size > 100 MB.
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @stevehipwell!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging and I will follow up with a new release shortly thereafter.

gmlewis:
> @gmlewis since working on #4153 I've found more issues with the pattern it replaced such as always returning `file` as the type for a submodule due to the directory loop pattern and a "bug" in the 2022 API version. This combined with a lack of Sentinel errors makes using the current released version painful, would it be possible to get a release out once this PR is merged?

@stevehipwell - this is now released here:
https://github.com/google/go-github/releases/tag/v86.0.0

PR #4197, opened 2026-05-07
test: Fix invalid JSON payloads in actions workflow runs tests

### LINKED ISSUE #4196: test: Actions GetWorkflowRun tests contain structurally invalid JSON payloads
While developing a performance optimization PR that introduces strict `json.Unmarshal` decoding, the CI runner surfaced a latent bug in the `go-github` test suite.

In `actions_workflow_runs_test.go`, the mock HTTP server endpoints for `TestActionsService_GetWorkflowRunAttempt` and `TestActionsService_GetWorkflowRunByID` are returning structurally invalid JSON payloads with double closing braces (`}}`).

The current `Client.Do` implementation uses `json.NewDecoder(resp.Body).Decode()`, which is a lazy streaming parser. It reads until the first valid `}` and silently ignores the trailing garbage byte. Any future PR that moves the client toward stricter, high-throughput parsing (like `json.Unmarshal`) will instantly fail these tests.

I have isolated this bug and will open an atomic PR to fix it so it doesn't block future optimizations.

merchantmoh-debug:
Fixes #4196

### Description
Removes the invalid trailing `}}` braces from the mock HTTP server JSON payloads in `actions_workflow_runs_test.go`.

### Motivation
This bug was silently masked by `json.NewDecoder()`. By correcting the test payloads, we ensure the tests are generating valid JSON, clearing the way for stricter memory-pooling and performance optimizations in the core client.

### Verification
- [x] Validated JSON structure.

### DIFF
--- a/github/actions_workflow_runs_test.go
+++ b/github/actions_workflow_runs_test.go
@@ -109,7 +109,7 @@ func TestActionsService_GetWorkflowRunByID(t *testing.T) {
 
 	mux.HandleFunc("/repos/o/r/actions/runs/29679449", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		fmt.Fprint(w, `{"id":399444496,"run_number":296,"created_at":"2019-01-02T15:04:05Z","updated_at":"2020-01-02T15:04:05Z"}}`)
+		fmt.Fprint(w, `{"id":399444496,"run_number":296,"created_at":"2019-01-02T15:04:05Z","updated_at":"2020-01-02T15:04:05Z"}`)
 	})
 
 	ctx := t.Context()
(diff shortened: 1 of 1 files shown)


gmlewis (approved):
Awesome, thank you, @merchantmoh-debug!
LGTM.

I'm looking at the surrounding code and it is not obvious to me as to how we can prevent bugs like this in the future.
Do you have any ideas, @merchantmoh-debug - or should we just consider this to be a one-off bug fix?

I'm totally fine either way.

merchantmoh-debug:
@gmlewis 

We could actually eliminate this class of bugs entirely by introducing a simple `testMockJSON(t, w, payload)` helper.

Right now, the mock servers use `fmt.Fprint(w, "...")` which just treats the JSON as a raw, unvalidated string literal. Because it's just a string, Go has no way of knowing if the JSON is structurally sound until a strict parser trips over it.

If we use a `testMockJSON` helper instead, it could strictly enforce `json.Valid([]byte(payload))` before writing to the `ResponseWriter`. That way, if a future developer accidentally hardcodes a typo or an extra brace, the mock server will instantly fail the test itself, rather than silently serving bad data to the client.

If you like the idea, I could write an AST script to automatically migrate all the intended JSON mock payloads across the test suite over to this new helper (intelligently skipping the tests that intentionally use invalid JSON for error checking). 

It would be a pretty big diff touching a lot of files though, so if you'd rather avoid the noise, we can absolutely just consider this a one-off bug fix. I'm totally fine either way.

gmlewis:
> It would be a pretty big diff touching a lot of files though, so if you'd rather avoid the noise, we can absolutely just consider this a one-off bug fix. I'm totally fine either way.

OK, excellent. Let's go ahead and get this in, and we can address the other ideas later if we decide it would be really beneficial.

PR #4201, opened 2026-05-08
feat!: Refactor client constructor to use options pattern

### LINKED ISSUE #3870: Refactor `WithEnterpriseURLs` helper
The current implementation of the `WithEnterpriseURLs` helper doesn't function as expected and doesn't appear to support GitHub Enterprise Cloud. For example testing the path of a `url.URL` can't assume that there will be slashes at either end. It also looks like this helper is expected to work correctly with standard GitHub URLs, so maybe it should be renamed to `WithURLs` and have the implementation fixed?

### LINKED ISSUE #3915: Question on `DownloadReleaseAsset` redirect logic

stevehipwell:
BREAKING CHANGE: Clients are now constructed with a nicer builder pattern. See docs for details.

This PR refactors the client to be immutable and to use the with options pattern for construction. This is a significant change but it should reduce the overall complexity and support better UX around new capabilities and backwards compatibility (e.g. `github.WithAPIVersion("2026-03-10")` or `github.WithGHESVersion("3.15")`).

- `Client` is immutable
  - `NewClient` is the only way to create a new `Client`
  - After a `Client` is created if you need to create a new one you can use `Clone`
  - You no longer need to create multiple clients and their associated child structs to chain helpers
- `Client` configuration is explicit
  - Options are handled in a specific order to limit the potential for misconfiguration
  - Passing invalid values into `NewClient` causes an early error instead of waiting for a failure or silently doing nothing
- Fixed implementation bugs
  - `clientMu` was unnecessary
  - Proxy from env pattern lost default transport configuration
- Improved test coverage?

Closes #3870 
Closes #3915

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,12 @@ if err != nil {
 }
 ```
 
+To support more advanced use cases; you can use the `github.WithTransport` option to provide a
+custom `http.RoundTripper` that handles authentication for you, or the `github.WithHTTPClient`
+option to provide a custom `http.Client`. As an example; you can use the `oauth2.Transport`
+from the [golang.org/x/oauth2](https://pkg.go.dev/golang.org/x/oauth2) package to handle OAuth
+token refreshing for you.
+
 Note that when using an authenticated Client, all calls made by the client will
 include the specified OAuth token. Therefore, authenticated clients should
 almost never be shared between different users.
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ go-github will require the N-1 major release of Go by default.
 
 [support-policy]: https://golang.org/doc/devel/release.html#policy
 
-## Development
+## Development ##
 
 If you're interested in using the [GraphQL API v4][], the recommended library is
 [shurcooL/githubv4][].
@@ -66,7 +66,10 @@ Construct a new GitHub client, then use the various services on the client to
 access different parts of the GitHub API. For example:
 
 ```go
-client := github.NewClient(nil)
+client, err := github.NewClient()
+if err != nil {
+	// Handle error.
+}
 
 // list all organizations for user "willnorris"
 orgs, _, err := client.Organizations.List(context.Background(), "willnorris", nil)
@@ -75,7 +78,10 @@ orgs, _, err := client.Organizations.List(context.Background(), "willnorris", ni
 Some API methods have optional parameters that can be passed. For example:
 
 ```go
-client := github.NewClient(nil)
+client, err := github.NewClient()
+if err != nil {
+	// Handle error.
+}
 
 // list public repositories for org "github"
 opt := &github.RepositoryListByOrgOptions{Type: "public"}
@@ -95,12 +101,15 @@ For more sample code snippets, head over to the
 
 ### Authentication ###
 
-Use the `WithAuthToken` method to configure your client to authenticate using an
+Use the `github.WithAuthToken` options method to configure your client to authenticate using an
 OAuth token (for example, a [personal access token][]). This is what is needed
 for a majority of use cases aside from GitHub Apps.
 
 ```go
-client := github.NewClient(nil).WithAuthToken("... your access token ...")
+client, err := github.NewClient(github.WithAuthToken("... your access token ..."))
+if err != nil {
+	// Handle error.
+}
 ```
 
 Note that when using an authenticated Client, all calls made by the client will
@@ -146,7 +155,10 @@ func main() {
 	}
 
 	// Use installation transport with client.
-	client := github.NewClient(&http.Client{Transport: itr})
+	client, err := github.NewClient(github.WithTransport(itr))
+	if err != nil {
+		// Handle error.
+	}
 
 	// Use client...
 }
@@ -186,11 +198,14 @@ func main() {
 	// InstallationTokenSource has the mechanism to refresh the token when it expires.
 	httpClient := oauth2.NewClient(context.Background(), installationTokenSource)
 
-	client := github.NewClient(httpClient)
+	client, err := github.NewClient(github.WithHTTPClient(httpClient))
+	if err != nil {
+		// Handle error.
+	}
 }
 ```
 
-*Note*: In order to interact with certain APIs, for example writing a file to a repo, one must generate an installation token
+_Note_: In order to interact with certain APIs, for example writing a file to a repo, one must generate an installation token
 using the installation ID of the GitHub app and authenticate with the OAuth method mentioned above. See the examples.
 
 ### Rate Limiting ###
@@ -296,9 +311,10 @@ import (
 	_ "github.com/bartventer/httpcache/store/memcache" //  Register the in-memory backend
 )
 
-client := github.NewClient(
-	httpcache.NewClient("memcache://"),
-).WithAuthToken(os.Getenv("GITHUB_TOKEN"))
+client, err := github.NewClient(github.WithHTTPClient(httpcache.NewClient("memcache://")), github.WithAuthToken(os.Getenv("GITHUB_TOKEN")))
+if err != nil {
+	// Handle error.
+}
 ```
 
 Alternatively, the [bored-engineer/github-conditional-http-transport](https://github.com/bored-engineer/github-conditional-http-transport)
@@ -334,7 +350,10 @@ embedded type of a more specific list options struct (for example
 `github.Response` struct.
 
 ```go
-client := github.NewClient(nil)
+client, err := github.NewClient()
+if err != nil {
+	// Handle error.
+}
 
 opt := &github.RepositoryListByOrgOptions{
 	ListOptions: github.ListOptions{PerPage: 10},
@@ -372,7 +391,10 @@ To handle rate limiting issues, make sure to use a rate-limiting transport.
 To use these methods, simply create an iterator and then range over it, for example:
 
 ```go
-client := github.NewClient(nil)
+client, err := github.NewClient()
+if err != nil {
+	// Handle error.
+}
 var allRepos []*github.Repository
 
 // create an iterator and start looping through all the results
@@ -389,7 +411,10 @@ Alternatively, if you wish to use an external package, there is `enrichman/gh-it
 Its iterator will handle pagination for you, looping through all the available results.
 
 ```go
-client := github.NewClient(nil)
+client, err := github.NewClient()
+if err != nil {
+	// Handle error.
+}
 var allRepos []*github.Repository
 
 // create an iterator and start looping through all the results
@@ -465,12 +490,14 @@ implementing preview features of the GitHub API, we've adopted the following
 versioning policy:
 
 * We increment the **major version** with any incompatible change to
-	non-preview functionality, including changes to the exported Go API surface
-	or behavior of the API.
+  non-preview functionality, including changes to the exported Go API surface
+  or behavior of the API.
+
 * We increment the **minor version** with any backwards-compatible changes to
-	functionality, as well as any changes to preview functionality in the GitHub
-	API. GitHub makes no guarantee about the stability of preview functionality,
-	so neither do we consider it a stable part of the go-github API.
+  functionality, as well as any changes to preview functionality in the GitHub
+  API. GitHub makes no guarantee about the stability of preview functionality,
+  so neither do we consider it a stable part of the go-github API.
+
 * We increment the **patch version** with any backwards-compatible bug fixes.
 
 Preview functionality may take the form of entire methods or simply additional
--- a/example/basicauth/main.go
+++ b/example/basicauth/main.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"strings"
 
@@ -41,8 +42,7 @@ func main() {
 
 	client, err := github.NewClient(github.WithHTTPClient(tp.Client()))
 	if err != nil {
-		fmt.Printf("\nerror: %v\n", err)
-		return
+		log.Fatalf("error: %v", err)
 	}
 	ctx := context.Background()
 	user, _, err := client.Users.Get(ctx, "")
@@ -56,8 +56,7 @@ func main() {
 	}
 
 	if err != nil {
-		fmt.Printf("\nerror: %v\n", err)
-		return
+		log.Fatalf("error: %v", err)
 	}
 
 	fmt.Printf("\n%v\n", github.Stringify(user))
--- a/example/basicauth/main.go
+++ b/example/basicauth/main.go
@@ -39,7 +39,11 @@ func main() {
 		Password: strings.TrimSpace(string(password)),
 	}
 
-	client := github.NewClient(tp.Client())
+	client, err := github.NewClient(github.WithHTTPClient(tp.Client()))
+	if err != nil {
+		fmt.Printf("\nerror: %v\n", err)
+		return
+	}
 	ctx := context.Background()
 	user, _, err := client.Users.Get(ctx, "")
 
--- a/example/contents/main.go
+++ b/example/contents/main.go
@@ -50,7 +50,11 @@ func main() {
 
 	fmt.Printf("\nDownloading %v/%v/%v at ref %v to %v...\n", owner, repo, repoPath, ref, outputPath)
 
-	client := github.NewClient(nil)
+	client, err := github.NewClient()
+	if err != nil {
+		fmt.Printf("Error creating GitHub client: %v\n", err)
+		os.Exit(1)
+	}
 
 	rc, _, err := client.Repositories.DownloadContents(context.Background(), owner, repo, repoPath, &github.RepositoryContentGetOptions{Ref: ref})
 	if err != nil {
--- a/example/contents/main.go
+++ b/example/contents/main.go
@@ -16,6 +16,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -52,27 +53,23 @@ func main() {
 
 	client, err := github.NewClient()
 	if err != nil {
-		fmt.Printf("Error creating GitHub client: %v\n", err)
-		os.Exit(1)
+		log.Fatalf("Error creating GitHub client: %v", err)
 	}
 
 	rc, _, err := client.Repositories.DownloadContents(context.Background(), owner, repo, repoPath, &github.RepositoryContentGetOptions{Ref: ref})
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
+		log.Fatalf("Error downloading contents: %v", err)
 	}
 	defer rc.Close()
 
 	f, err := os.Create(outputPath) //#nosec G703 -- path is validated above
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
+		log.Fatalf("Error creating output file: %v", err)
 	}
 	defer f.Close()
 
 	if _, err := io.Copy(f, rc); err != nil {
-		fmt.Printf("Error: %v\n", err)
-		os.Exit(1)
+		log.Fatalf("Error writing to output file: %v", err)
 	}
 
 	fmt.Println("Download completed.")
--- a/example/ratelimit/main.go
+++ b/example/ratelimit/main.go
@@ -12,6 +12,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"log"
 
 	"github.com/gofri/go-github-pagination/githubpagination"
 	"github.com/gofri/go-github-ratelimit/v2/github_ratelimit"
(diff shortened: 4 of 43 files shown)


example/basicauth/main.go
```diff
@@ -39,7 +39,11 @@ func main() {
 		Password: strings.TrimSpace(string(password)),
 	}
 
-	client := github.NewClient(tp.Client())
+	client, err := github.NewClient(github.WithHTTPClient(tp.Client()))
+	if err != nil {
+		fmt.Printf("\nerror: %v\n", err)
+		return
```
gmlewis:
```suggestion
		log.Fatalf("error: %v", err)
```

example/contents/main.go
```diff
@@ -50,7 +50,11 @@ func main() {
 
 	fmt.Printf("\nDownloading %v/%v/%v at ref %v to %v...\n", owner, repo, repoPath, ref, outputPath)
 
-	client := github.NewClient(nil)
+	client, err := github.NewClient()
+	if err != nil {
+		fmt.Printf("Error creating GitHub client: %v\n", err)
+		os.Exit(1)
```
gmlewis:
```suggestion
		log.Fatalf("Error creating GitHub client: %v", err)
```

example/ratelimit/main.go
```diff
@@ -37,7 +37,11 @@ func main() {
 	paginator := githubpagination.NewClient(rateLimiter,
 		githubpagination.WithPerPage(100), // default to 100 results per page
 	)
-	client := github.NewClient(paginator)
+	client, err := github.NewClient(github.WithHTTPClient(paginator))
+	if err != nil {
+		fmt.Printf("Error creating GitHub client: %v\n", err)
+		return
```
gmlewis:
```suggestion
		log.Fatalf("Error creating GitHub client: %v", err)
```

example/tokenauth/main.go
```diff
@@ -25,7 +25,10 @@ func main() {
 	fmt.Println()
 
 	ctx := context.Background()
-	client := github.NewClient(nil).WithAuthToken(string(token))
+	client, err := github.NewClient(github.WithAuthToken(string(token)))
+	if err != nil {
+		log.Fatalf("Error creating GitHub client: %v\n", err)
```
gmlewis:
```suggestion
		log.Fatalf("Error creating GitHub client: %v", err)
```

example/uploadreleaseassetfromrelease/main.go
```diff
@@ -24,7 +24,10 @@ func main() {
 	}
 
 	ctx := context.Background()
-	client := github.NewClient(nil).WithAuthToken(token)
+	client, err := github.NewClient(github.WithAuthToken(token))
+	if err != nil {
+		log.Fatalf("Error creating GitHub client: %v\n", err)
```
gmlewis:
```suggestion
		log.Fatalf("Error creating GitHub client: %v", err)
```

example/verifyartifact/main.go
```diff
@@ -78,7 +78,10 @@ func main() {
 	}
 
 	ctx := context.Background()
-	client := github.NewClient(nil).WithAuthToken(token)
+	client, err := github.NewClient(github.WithAuthToken(token))
+	if err != nil {
+		log.Fatalf("Error creating GitHub client: %v\n", err)
```
gmlewis:
```suggestion
		log.Fatalf("Error creating GitHub client: %v", err)
```

test/fields/fields.go
```diff
@@ -40,9 +40,19 @@ func main() {
 	token := os.Getenv("GITHUB_AUTH_TOKEN")
 	if token == "" {
 		fmt.Print("!!! No OAuth token. Some tests won't run. !!!\n\n")
-		client = github.NewClient(nil)
+		c, err := github.NewClient()
+		if err != nil {
+			fmt.Printf("Error creating GitHub client: %v\n", err)
+			os.Exit(1)
```
gmlewis:
```suggestion
			log.Fatalf("Error creating GitHub client: %v", err)
```

test/fields/fields.go
```diff
@@ -40,9 +40,19 @@ func main() {
...
+		}
+		client = c
 	} else {
-		client = github.NewClient(nil).WithAuthToken(token)
+		c, err := github.NewClient(github.WithAuthToken(token))
+		if err != nil {
+			fmt.Printf("Error creating GitHub client with token: %v\n", err)
+			os.Exit(1)
```
gmlewis:
```suggestion
			log.Fatalf("Error creating GitHub client with token: %v", err)
```

test/integration/github_test.go
```diff
@@ -24,9 +24,20 @@ import (
 var client, auth = sync.OnceValues(func() (*github.Client, bool) {
 	token := os.Getenv("GITHUB_AUTH_TOKEN")
 	if token == "" {
-		return github.NewClient(nil), false
+		c, err := github.NewClient()
+		if err != nil {
+			fmt.Printf("Error creating GitHub client: %v\n", err)
+			os.Exit(1)
```
gmlewis:
```suggestion
			log.Fatalf("Error creating GitHub client: %v", err)
```

test/integration/github_test.go
```diff
@@ -24,9 +24,20 @@ import (
...
+		}
+		return c, false
+	}
+
+	c, err := github.NewClient(github.WithAuthToken(token))
+	if err != nil {
+		fmt.Printf("Error creating GitHub client with token: %v\n", err)
+		os.Exit(1)
```
gmlewis:
```suggestion
		log.Fatalf("Error creating GitHub client with token: %v", err)
```

tools/metadata/main.go
```diff
@@ -57,6 +57,7 @@ type rootCmd struct {
 
 	// for testing
 	GithubURL string `kong:"hidden,default='https://api.github.com'"`
+	UploadURL string `kong:"hidden,default='https://uploads.github.com'"`
```
gmlewis, line 60: This seems unrelated. Are you addressing two issues in this PR?

gmlewis (changes requested):
Thank you, @stevehipwell.

Although I like the overall pattern, I have a few concerns with this PR.
First, the various fields within `Client` were originally exported individually, on a case-by-case basis, as users had need to access them. Suddenly unexporting all fields may wreak havoc.

Second, the `clientMu` was indeed needed to solve a race condition found when copying clients that were currently in-flight, if I'm remembering correctly.

Can we keep the nice builder pattern you've created here without unexporting all the fields and without removing the mutex that is protecting the client?

stevehipwell:
@gmlewis if you look at the code you'll see that `clientMu` is doing nothing and doesn't even wrap the relevant concerns in the copy function. It used to technically serve a purpose when some of the functions mutated the client, but that no longer happens, although as I said above it wasn't "correct" even for that. I also double checked this with 2 different agents, which both agreed with my summary above.

Unexporting the values and putting them behind the options pattern to make the client immutable is core I the whole pattern. Could you provide some examples where these need changing and `Clone` is the wrong choice? For example I added the disable rate limit option to the client, it's only exported as there wasn't an existing pattern so that was the simplest approach. I'm happy to add getters for the remaining unexpired fields, but I wasn't sure they were actually needed.

example/basicauth/main.go
stevehipwell: I intentionally kept the existing pattern for the individual examples rather than standardising them all.

tools/metadata/main.go
stevehipwell: The old pattern allowed you to leave the client in an invalid state and would just error when you used an empty URL. The new pattern checks for errors early so there is no uncertainty later as to why a call failed.

example/basicauth/main.go
gmlewis: No, let's please clean them all up and standardize them since you are touching every one of these in this PR.

tools/metadata/main.go
gmlewis: Gotcha, OK, thanks.

gmlewis:
> @gmlewis if you look at the code you'll see that `clientMu` is doing nothing and doesn't even wrap the relevant concerns in the copy function. It used to technically serve a purpose when some of the functions mutated the client, but that no longer happens, although as I said above it wasn't "correct" even for that. I also double checked this with 2 different agents, which both agreed with my summary above.
> 
> Unexporting the values and putting them behind the options pattern to make the client immutable is core I the whole pattern. Could you provide some examples where these need changing and `Clone` is the wrong choice? For example I added the disable rate limit option to the client, it's only exported as there wasn't an existing pattern so that was the simplest approach. I'm happy to add getters for the remaining unexpired fields, but I wasn't sure they were actually needed.

OK, I see. That sounds good to me.

If you wouldn't mind please cleaning up all the non-idiomatic failures in the examples where I have marked them, that would be greatly appreciated, then we should be ready for a second LGTM+Approval.

gmlewis:
If you could also please investigate the 8 missed lines in code coverage in `github/github.go` in this PR, that would also be great. Thanks!

README.md
```diff
@@ -465,12 +490,14 @@ implementing preview features of the GitHub API, we've adopted the following
...
+  or behavior of the API.
+
 * We increment the **minor version** with any backwards-compatible changes to
-	functionality, as well as any changes to preview functionality in the GitHub
-	API. GitHub makes no guarantee about the stability of preview functionality,
-	so neither do we consider it a stable part of the go-github API.
+  functionality, as well as any changes to preview functionality in the GitHub
+  API. GitHub makes no guarantee about the stability of preview functionality,
+  so neither do we consider it a stable part of the go-github API.
+
```
alexandear, line 506: Why were these lines changed? It looks like they are not related to the PR's topic.

README.md
```diff
@@ -186,11 +198,14 @@ func main() {
...
-	client := github.NewClient(httpClient)
+	client, err := github.NewClient(github.WithHTTPClient(httpClient))
+	if err != nil {
+		// Handle error.
+	}
 }
 ```
 
-*Note*: In order to interact with certain APIs, for example writing a file to a repo, one must generate an installation token
+_Note_: In order to interact with certain APIs, for example writing a file to a repo, one must generate an installation token
```
alexandear, line 214: Unrelated change:
```suggestion
*Note*: In order to interact with certain APIs, for example writing a file to a repo, one must generate an installation token
```

README.md
```diff
@@ -20,7 +20,7 @@ go-github will require the N-1 major release of Go by default.
 
 [support-policy]: https://golang.org/doc/devel/release.html#policy
 
-## Development
+## Development ##
```
alexandear, line 23: Unrelated:
```suggestion
## Development
```

github/github.go
```diff
@@ -337,113 +334,271 @@ func addOptions[P structPtr[T], T any](s string, opts P) (string, error) {
...
 	}
-	return c2, nil
 }
 
-// initialize sets default values and initializes services.
-func (c *Client) initialize() {
-	if c.client == nil {
+// WithSecondaryRateLimitOptions returns a ClientOptionsFunc that configures the Client
+// secondary rate limits.
+func WithSecondaryRateLimitOptions(maxRetryAfterDuration time.Duration) ClientOptionsFunc {
```
alexandear: "Options" seems redundant:
```suggestion
func WithSecondaryRateLimit(maxRetryAfterDuration time.Duration) ClientOptionsFunc {
```

github/github.go
```diff
@@ -337,113 +334,271 @@ func addOptions[P structPtr[T], T any](s string, opts P) (string, error) {
...
+// use [WithAuthToken] or provide an http.Client that performs authentication
+// (e.g., using golang.org/x/oauth2) to [WithHTTPClient]. For GitHub
+// Enterprise, use [WithEnterpriseURLs] to set the base and upload URLs. If no
+// http.Client is provided, a default one will be used, but it is recommended
+// to provide a custom http.Client with an appropriate timeout for production
+// environments.
+//
+// Note: When using a nil httpClient, the default client has no timeout set.
+// This may not be suitable for production environments. It is recommended to
+// provide a custom http.Client with an appropriate timeout.
```
alexandear: `NewClient` no longer accepts an `http.Client` argument or `nil` `http.Client`. This note is now misleading and should be rewritten.

github/github_test.go
```diff
@@ -3632,13 +4265,16 @@ func TestClientCopy_leak_transport(t *testing.T) {
...
 		_, _ = fmt.Fprintf(w, `{"login": "%v"}`, accessToken)
 	}))
-	clientPreconfiguredWithURLs, err := NewClient(nil).WithEnterpriseURLs(srv.URL, srv.URL)
+	clientPreconfiguredWithURLs := mustNewClient(t, WithEnterpriseURLs(srv.URL, srv.URL))
+
+	aliceClient, err := clientPreconfiguredWithURLs.Clone(WithAuthToken("alice"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	bobClient, err := clientPreconfiguredWithURLs.Clone(WithAuthToken("bob"))
```
alexandear, line 4426: If we change `clientPreconfiguredWithURLs` to `aliceClient`, the test will fail:
```suggestion
	bobClient, err := aliceClient.Clone(WithAuthToken("bob"))
```

```
=== RUN   TestClientCopy_leak_transport
=== PAUSE TestClientCopy_leak_transport
=== CONT  TestClientCopy_leak_transport
    /Users/alexandear/src/github.com/google/go-github/github/github_test.go:4291: diff mismatch (-want +got):
          string(
        - 	"Bearer bob",
        + 	"Bearer alice",
          )
--- FAIL: TestClientCopy_leak_transport (0.00s)
```

But it shouldn't fail because we are overwriting Alice's auth token with Bob's one.

github/github_test.go
```diff
@@ -427,123 +447,155 @@ func assertWrite(t *testing.T, w io.Writer, data []byte) {
...
-	}
+		customClient := &http.Client{Timeout: 10 * time.Second}
+		opts := clientOptions{}
+		err := WithHTTPClient(customClient)(&opts)
+		if err != nil {
+			t.Errorf("WithHTTPClient errored: %v", err)
+		}
+
+		if opts.httpClient == nil {
+			t.Error("httpClient is nil")
```
```suggestion
			t.Fatal("httpClient is nil")
```

github/github_test.go
```diff
@@ -427,123 +447,155 @@ func assertWrite(t *testing.T, w io.Writer, data []byte) {
...
-	c := NewClient(nil)
-	c2 := c.Client()
-	if c.client == c2 {
-		t.Error("Client returned same http.Client, but should be different")
-	}
+		customClient := &http.Client{Timeout: 10 * time.Second}
+		opts := clientOptions{}
+		err := WithHTTPClient(customClient)(&opts)
+		if err != nil {
+			t.Errorf("WithHTTPClient errored: %v", err)
```
```suggestion
			t.Fatalf("WithHTTPClient errored: %v", err)
```

github/github_test.go
```diff
@@ -427,123 +447,155 @@ func assertWrite(t *testing.T, w io.Writer, data []byte) {
...
-		httpClient := &http.Client{}
-		client := NewClient(httpClient).WithAuthToken(token)
-		validate(t, client.Client(), token)
-		// make sure the original client isn't setting auth headers now
-		validate(t, httpClient, "")
+
+		opts := clientOptions{}
+		err := WithUserAgent("")(&opts)
+		if err != nil {
+			t.Errorf("WithUserAgent errored: %v", err)
```
```suggestion
			t.Fatalf("WithUserAgent errored: %v", err)
```

github/github_test.go
alexandear: The problem is that `Clone(WithAuthToken())` silently uses the original client's token instead of the new one. When cloning a client that already has an auth token set and providing a different token via `WithAuthToken`, the new token is ignored.

github/github_test.go
```diff
@@ -640,34 +692,615 @@ func TestWithEnterpriseURLs(t *testing.T) {
...
+			name:                  "maxRetryAfterDuration is 1 minute",
+			maxRetryAfterDuration: time.Minute,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			opts := clientOptions{}
+			err := WithSecondaryRateLimitOptions(tt.maxRetryAfterDuration)(&opts)
+			if err != nil {
+				t.Errorf("WithSecondaryRateLimitOptions errored: %v", err)
```
```suggestion
				t.Fatalf("WithSecondaryRateLimitOptions errored: %v", err)
```

README.md
stevehipwell: I've modified the README and these lines weren't consistently formatted. @gmlewis has asked me to fixup files where I've made changes, so I suggest we keep this for the same reason.

README.md
stevehipwell: I've modified the README and these lines weren't consistently formatted. @gmlewis has asked me to fixup files where I've made changes, so I suggest we keep this for the same reason.

README.md
stevehipwell: I've modified the README and these lines weren't consistently formatted. @gmlewis has asked me to fixup files where I've made changes, so I suggest we keep this for the same reason.

github/github.go
stevehipwell: I've created a single options function here for the secondary rate limit both because I think there may be other options required and because the singular property being set is long. I could rename it to remove the `Options` suffix, but would that not be confusing given that the secondary rate limit will be enabled even if this isn't passed in?

github/github.go
stevehipwell: Good spot. I'd be tempted to add a `WithTimeout` options function to allow this to be set without having to pass in a full HTTP client.

github/github_test.go
stevehipwell: Isn't this exactly the same behaviour as the current version?

github/github.go
stevehipwell: I added `WithTimeout` and updated this comment.

stevehipwell:
@gmlewis I've addressed you review comments and should have fixed the test coverage. I've also added exclusions for the examples and other code that shouldn't be be part of the coverage report which should make it easier to see what code should have been tested but hasn't been.

github/examples_test.go
```diff
@@ -118,7 +130,10 @@ func ExamplePullRequestsService_Create() {
 	// github.NewClient() instead of nil. See the following documentation for more
```
Not-Dhananjay-Mishra: That comment feel misleading since `github.NewClient()` no longer takes `nil`. Should we update that?

stevehipwell:
@alexandear I'm still waiting on a reply on a couple of threads you opened?

alexandear (approved):

github/github_test.go
alexandear: Yeah, you're right. This can be fixed in a separate PR.

github/github.go
alexandear: Can be also:
```go
func WithMaxSecondaryRateLimitMaxRetryAfterDuration(d time.Duration) ClientOptionsFunc {
```

This would be consistent with other options functions.

stevehipwell: Yeah, I agree that this would be consistent.

stevehipwell: Done.

stevehipwell:
@gmlewis all review comments should now have been addressed.

gmlewis:
> @gmlewis all review comments should now have been addressed.

Thanks, @stevehipwell.
It looks like conflicts must be resolved before we can move forward.

stevehipwell:
@gmlewis the UI was showing me it could be resolved on the PR so I left it for you to decide. Would you like me to merge `master` into my branch and push that up?

gmlewis:
> @gmlewis the UI was showing me it could be resolved on the PR so I left it for you to decide. Would you like me to merge `master` into my branch and push that up?

Yes, please.

stevehipwell:
@gmlewis done.

github/github.go
```diff
@@ -337,113 +334,292 @@ func addOptions[P structPtr[T], T any](s string, opts P) (string, error) {
...
 	}
-	return c2, nil
 }
 
-// initialize sets default values and initializes services.
-func (c *Client) initialize() {
-	if c.client == nil {
+// WithMaxSecondaryRateLimitMaxRetryAfterDuration returns a ClientOptionsFunc
+// that configures the Client secondary rate limit max retry after duration.
+func WithMaxSecondaryRateLimitMaxRetryAfterDuration(duration time.Duration) ClientOptionsFunc {
```
alexandear: Is the second `Max` intentional here?

```suggestion
func WithMaxSecondaryRateLimitRetryAfterDuration(duration time.Duration) ClientOptionsFunc {
```

`o.maxSecondaryRateLimitRetryAfterDuration` -> `WithMaxSecondaryRateLimitRetryAfterDuration`

stevehipwell: No, thanks for the spot. Fixed.

alexandear (approved):

Not-Dhananjay-Mishra (approved):
LGTM 🚀

gmlewis (approved):
Very nice, @stevehipwell... thank you!
LGTM.

Merging.

PR #4204, opened 2026-05-09
fix: Handle array-typed org/org_id in AuditEntry JSON unmarshaling

nghiack7:
## Problem

The GitHub Enterprise audit-log API sometimes returns `org` as a JSON array of strings and `org_id` as a JSON array of integers, deviating from the documented scalar types. This causes `GetAuditLog` / `GetAuditLog` (Enterprise) to return an unmarshal error:

```
json: cannot unmarshal array into Go struct field entryAlias.org of type string
```

Reported on v66 and v69. See #3488.

## Root cause

`AuditEntry.UnmarshalJSON` delegates to `json.Unmarshal` via a `type entryAlias AuditEntry` alias. When the API returns `"org": ["a", "b"]` the decoder fails because `Org` is typed `*string`.

## Fix

In `UnmarshalJSON`, capture `org` and `org_id` as `json.RawMessage` before the main decode pass, then normalise each:

- `string` arrays are joined with `", "` (preserves all org names)
- `int64` arrays use the first element (mirrors how callers would use it)
- Scalar values pass through unchanged

Two unexported helpers (`unmarshalStringOrStringArray`, `unmarshalInt64OrInt64Array`) encapsulate the logic so the core unmarshal flow stays readable.

## Tests

Added table-driven tests in `orgs_audit_log_test.go`:

| test | payload shape | expected result |
|---|---|---|
| `org_as_scalar_string` | `"org":"myorg"` | `"myorg"` |
| `org_as_single_element_array` | `"org":["myorg"]` | `"myorg"` |
| `org_as_multi_element_array` | `"org":["org1","org2","org3"]` | `"org1, org2, org3"` |
| `org_id_as_scalar` | `"org_id":42` | `42` |
| `org_id_as_array` | `"org_id":[42,43,44]` | `42` (first element) |

All existing audit-log tests continue to pass.

### DIFF
--- a/github/orgs_audit_log.go
+++ b/github/orgs_audit_log.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 )
 
 // GetAuditLogOptions sets up optional parameters to query audit-log endpoint.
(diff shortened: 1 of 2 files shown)


nghiack7:
Added tests to cover the error paths and edge cases flagged by Codecov (68.57% → should now be above 90%):

- **Empty `org_id` array** → `OrgID` is `nil` (no panic)
- **Object-typed `org`** → error from `unmarshalStringOrStringArray`
- **Object-typed `org_id`** → error from `unmarshalInt64OrInt64Array`
- **Explicit JSON `null`** for both fields → both remain `nil`, no error

All existing tests still pass.

nghiack7:
@google-cla recheck

gmlewis:
> @google-cla recheck

Unfortunately, the CLA check is still failing. Sometimes it helps to close a PR and create a brand new one using only the email address that was used to sign the CLA (or alternatively changing the commits and then performing a force-push).
Please investigate when you get a chance, @nghiack7.

danyalahmed1995:
@nghiack7 Just in case it helps, the CLA check usually depends on the email attached to the commits.

A few things you can try:
- Check the commit email with `git log --format='%h %ae %an'`
- Make sure that email matches the one used for the Google CLA
- If needed, update local git config with `git config user.email "your-cla-email@example.com"`
- Re-author the commits with the correct email, then force-push the branch
- If the bot still does not refresh, try an empty commit after fixing the email: `git commit --allow-empty -m "chore: trigger CLA recheck"` and push again

If that still fails, the cleanest path is probably what Glenn suggested: recreate the PR from a branch whose commits were authored with the CLA-approved email.

PR #4209, opened 2026-05-11
fix!: Support omitting `EnforceAdmins`

conorevans:
https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2026-03-10

image

I think it's worth noting that this does change the default behaviour of the request. Because it previously would send `false` even if not explicitly set by user, this would toggle the setting to true (as not null). Whereas here, not setting the field will send null, and thus true.

You could flip that logic by having it be `DoNotEnforceAdmins *bool` but then there would need to be a translation between the ProtectionRequest that the user uses and the one that is JSON-marshalled to the API.

With that in mind I am happy for you to adapt this PR to a different approach if it seems better to you.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -28678,12 +28678,12 @@ func (p *ProtectionRequest) GetBlockCreations() bool {
 	return *p.BlockCreations
 }
 
-// GetEnforceAdmins returns the EnforceAdmins field.
+// GetEnforceAdmins returns the EnforceAdmins field if it's non-nil, zero value otherwise.
 func (p *ProtectionRequest) GetEnforceAdmins() bool {
-	if p == nil {
+	if p == nil || p.EnforceAdmins == nil {
 		return false
 	}
-	return p.EnforceAdmins
+	return *p.EnforceAdmins
 }
 
 // GetLockBranch returns the LockBranch field if it's non-nil, zero value otherwise.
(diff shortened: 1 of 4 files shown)


gmlewis (changes requested):
Thank you, @conorevans!
Please run the `./script/lint.sh` script from step 4 of CONTRIBUTING.md and push the changes to this PR.
Then once you sign the CLA, we can get a second LGTM+Approval from any other contributor to this repo before merging.

conorevans:
Hi @gmlewis thanks for taking a look so quickly! I wanted to see if the PR was palatable before doing too much. I also thought I might be able to squeeze the change through just using UI, not realise there was a gen script to run. Have run that, squashed, and re-committed in signed format. Hopefully good to go now :)

gmlewis:
I apologize, I should have also mentioned running `./script/generate.sh` and pushing the results.
All the details are in CONTRIBUTING.md, under "step 4".

conorevans:
I was using an older version of the library (v72) in the scenario where it wasn't working; when I run against `conorevans/go-github` I get

```
PUT https://api.github.com/repos/foobar/foobar/branches/main/protection: 422 Invalid request.\n\n\"enforce_admins\" wasn't supplied. []
```

And if I submit it as `false` using go-github/v86 it does indeed uncheck the permission

image

Thanks for your time @glenjamin and apologies for not catching the outdated version. At least I'm set up for contributing in the future 😛

PR #4210, opened 2026-05-11
fix: Close `httptest` server to prevent test flakiness

gmlewis:
This PR should address the test flakiness demonstrated here:
https://github.com/google/go-github/actions/runs/25682256160/job/75396628089

The `httptest.Server` instances in `TestWithAuthToken` were never explicitly closed, relying on garbage collection finalizers to clean up. The GC finalizer calls `Close() → CloseIdleConnections()` on the transport, which races with the HTTP client's connection pool, producing:

> net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called

This was especially likely under `-race` and on the Linux runner. Adding `defer srv.Close()` to both `httptest.NewServer` calls causes the server is shut down deterministically before the test returns, preventing the broken-connection race.

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -477,7 +477,7 @@ func TestWithAuthToken(t *testing.T) {
 			gotReq = true
 			headerVal = r.Header.Get("Authorization")
 		}))
-		defer srv.Close()
+		t.Cleanup(srv.Close)
 		_, err := c.Get(srv.URL)
 		assertNilError(t, err)
 		if !gotReq {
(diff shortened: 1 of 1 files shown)


gmlewis:

github/github_test.go
```diff
@@ -477,6 +477,7 @@ func TestWithAuthToken(t *testing.T) {
 			gotReq = true
 			headerVal = r.Header.Get("Authorization")
 		}))
+		defer srv.Close()
```
alexandear: Slightly better:
```suggestion
		t.Cleanup(srv.Close)
```

github/github_test.go
```diff
@@ -517,6 +518,7 @@ func TestWithAuthToken(t *testing.T) {
 			gotReq = true
 			_, ifAuthorizationSet = r.Header["Authorization"]
 		}))
+		defer srv.Close()
```
```suggestion
		t.Cleanup(srv.Close)
```

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #4211, opened 2026-05-13
fix: Use value receiver for `MarshalJSON`

alexandear:
Consistently use value receivers for `MarshalJSON` methods.

The issue is clearly illustrated by the following program, where `panic` is never called. However, if we change `p := Person{"Bob"}` to `p := &Person{"Bob"}`, everything works as expected, and the panic occurs.

```go
package main

import (
	"encoding/json"
	"fmt"
)

type Person struct {
	Name string
}

func (p *Person) MarshalJSON() ([]byte, error) {
	panic("unexpected")
}

func main() {
	p := Person{"Bob"}
	b, err := json.Marshal(p)
	if err != nil {
		panic(err)
	}
	fmt.Printf("%s\n", b)
}
```

The issue exists only in `encoding/json` v1. The experimental `encoding/json/v2` package does not have this issue.

See https://go.dev/issue/22967 for more details.

I think this is a non-breaking change.

### DIFF
--- a/github/git_trees.go
+++ b/github/git_trees.go
@@ -66,7 +66,7 @@ type treeEntryWithFileDelete struct {
 }
 
 // MarshalJSON implements the json.Marshaler interface.
-func (t *TreeEntry) MarshalJSON() ([]byte, error) {
+func (t TreeEntry) MarshalJSON() ([]byte, error) {
 	if t.SHA == nil && t.Content == nil {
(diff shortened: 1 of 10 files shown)


munlicode (approved):
LGTM

gmlewis (approved):
Thank you, @alexandear and @munlicode!
LGTM.
Merging.

Issue #4117, opened 2026-03-19
Missing iterators


stevehipwell:
It looks like the current auto generated iterators are only being created for the list commands, but the GitHub API uses pagination for other endpoints such as [get a commit](https://docs.github.com/en/rest/commits/commits?apiVersion=2026-03-10#get-a-commit) (`client.Repositories.GetCommit()`).

gmlewis:
Yes, that was a conscious design decision:

```
		if !fd.Name.IsExported() || !strings.HasPrefix(fd.Name.Name, "List") {
			continue
		}
```

The plan was to make this feature extremely consistent across the entire repo.
So if we wanted to make iteration available for other methods, we would make a breaking change and force it to conform to this standard.

stevehipwell:
> Yes, that was a conscious design decision

Can you remember the reasons? Wouldn't creating an iterator function for any exported function that has one of the list options as an argument also be consistent?

gmlewis:
Yes, it was just so that users can easily look for `List*Iter` methods, and if the method starts with `List*` it might have an iterator.

> Can you remember the reasons? Wouldn't creating an iterator function for any exported function that has one of the list options as an argument also be consistent?

Yes, that is absolutely also a viable alternative design decision.

We had others weigh in on this decision previously while we were originally creating the feature...

cc: @alexandear - @zyfy29 - @Not-Dhananjay-Mishra - @munlicode - Does anyone have any opinions about changing our design decisions made for the current set of iterators to accomodate this alternate design?

I'm not married to the current approach but I do believe we discussed this previously.

munlicode:
I think that creating a function, as @stevehipwell suggested, would be great, since it makes the library more resilient in case function, for some reason, needs to attain its name.

We could potentially add verbose command that would show what functions do not have `List*Iter` or `List*` naming style, so that it is just clear what functions are not consistent in naming, though it is more of an enchantment.

Not-Dhananjay-Mishra:
I have a question even if we allow iterator generation for methods without a List* prefix (like `GetCommit`), it returns a `*RepositoryCommit`, which is not a slice. Instead, it contains two slice fields `Parents []*Commit` and `Files []*CommitFile`. How will we handle this case?With the current implementation, only structs with a single slice field are supported. I guess there will be more cases like this.

stevehipwell:
@Not-Dhananjay-Mishra when you page the commit API you get the same object back but with the next set of files. Wouldn't that be the implementation here?

[...]

PR #4229, opened 2026-05-14
refactor!: Change `GetRulesForBranch` to `ListRulesForBranch`

gmlewis:
BREAKING CHANGE: `RepositoriesService.GetRulesForBranch` is now `RepositoriesService.ListRulesForBranch`.

Relates to: #4117.

### DIFF
--- a/github/repos_rules.go
+++ b/github/repos_rules.go
@@ -8,6 +8,7 @@ package github
 import (
 	"context"
 	"fmt"
+	"iter"
 )
 
 // ListRulesForBranch gets all the repository rules that apply to the specified branch.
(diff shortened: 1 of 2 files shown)


gmlewis:

gmlewis:
Note that I'm not terribly concerned about the code coverage report in this PR since the JSON custom unmarshaler is tested separately.

github/repos_rules.go
```diff
@@ -37,6 +38,147 @@ func (s *RepositoriesService) GetRulesForBranch(ctx context.Context, owner, repo
 	return rules, resp, nil
 }
 
+// ListRulesForBranchIter returns an iterator that paginates through all results of ListRulesForBranch.
+//
+// Note that since [BranchRules] contains a large number of slices, this iterator
+// returns type `any` and it is therefore the responsibility of the caller to perform a
+// type switch to determine what item is being returned for each iteration.
+func (s *RepositoriesService) ListRulesForBranchIter(ctx context.Context, owner, repo, branch string, opts *ListOptions) iter.Seq2[any, error] {
```
stevehipwell, line 46: I can see why this iterator isn't being generated, but do you need to add some configuration to block the generator or will it see that this already exists?

I'd also like a go refactoring it once it's merged as I think we could make the API a bit more specific and remove `any`.

stevehipwell (commented):
This looks good, just a minor question.

github/repos_rules.go
gmlewis: > I can see why this iterator isn't being generated, but do you need to add some configuration to block the generator or will it see that this already exists?

No, the generator already skips this one due to the `BranchRules` struct having many slices and one is not called out as "the" slice to use.

> I'd also like a go refactoring it once it's merged as I think we could make the API a bit more specific and remove `any`.

OK, sure.

PR #4230, opened 2026-05-16
feat: Add enterprise app installation lookup

### LINKED ISSUE #4212: Add support for getting a GitHub App enterprise installation
GitHub has [announced](https://github.blog/changelog/2026-05-13-new-enterprise-installation-api-now-in-public-preview/) support for apps to discover if they're installed in an enterprise and if so what the installation ID is. We should support this endpoint.

danyalahmed1995:
## Summary

Adds support for looking up a GitHub App installation for an enterprise.

This implements the new `AppsService.FindEnterpriseInstallation` helper for:

```http
GET /enterprises/{enterprise}/installation
```

The method follows the existing app installation lookup pattern already used for organization, repository, and user installations.

## Details

- Added `FindEnterpriseInstallation(ctx, enterprise)` to `AppsService`
- Uses the enterprise installation endpoint:
  - `GET /enterprises/{enterprise}/installation`
- Reuses the existing `Installation` model
- Keeps the implementation aligned with the existing installation lookup helpers
- Adds coverage for:
  - successful enterprise installation lookup
  - bad options handling
  - request/do failure handling

## Testing

```bash
gofmt -w github/apps.go github/apps_test.go
go test ./github -run 'Apps|Installation'
go test ./...
```

All tests passed locally.

Fixes #4212

### DIFF
--- a/github/apps.go
+++ b/github/apps.go
@@ -457,6 +457,15 @@ func (s *AppsService) FindOrganizationInstallation(ctx context.Context, org stri
 	return s.getInstallation(ctx, fmt.Sprintf("orgs/%v/installation", org))
 }
 
+// FindEnterpriseInstallation finds the enterprise's installation information.
+//
+// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/apps/apps#get-an-enterprise-installation-for-the-authenticated-app
+//
+//meta:operation GET /enterprises/{enterprise}/installation
+func (s *AppsService) FindEnterpriseInstallation(ctx context.Context, enterprise string) (*Installation, *Response, error) {
+	return s.getInstallation(ctx, fmt.Sprintf("enterprises/%v/installation", enterprise))
+}
+
 // FindRepositoryInstallation finds the repository's installation information.
 //
 // GitHub API docs: https://docs.github.com/rest/apps/apps?apiVersion=2022-11-28#get-a-repository-installation-for-the-authenticated-app
--- a/github/apps.go
+++ b/github/apps.go
@@ -459,7 +459,7 @@ func (s *AppsService) FindOrganizationInstallation(ctx context.Context, org stri
 
 // FindEnterpriseInstallation finds the enterprise's installation information.
 //
-// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/apps/apps#get-an-enterprise-installation-for-the-authenticated-app
+// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/apps/apps?apiVersion=2022-11-28#get-an-enterprise-installation-for-the-authenticated-app
 //
 //meta:operation GET /enterprises/{enterprise}/installation
 func (s *AppsService) FindEnterpriseInstallation(ctx context.Context, enterprise string) (*Installation, *Response, error) {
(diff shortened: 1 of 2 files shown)


github/apps.go
```diff
@@ -457,6 +457,15 @@ func (s *AppsService) FindOrganizationInstallation(ctx context.Context, org stri
 	return s.getInstallation(ctx, fmt.Sprintf("orgs/%v/installation", org))
 }
 
+// FindEnterpriseInstallation finds the enterprise's installation information.
```
gmlewis, line 460: The GitHub API docs call this "Get an enterprise installation for the authenticated app" so instead of "FindEnterpriseInstallation", can we use "GetEnterpriseInstallation"?

Wow... I see all the other ones in this file are called "Find*"... I'm not sure how I let that one slide by in code reviews. 😞 

OK, let's take this to a vote from our dedicated team of reviewers... which do you like best?

* Option A: Just name this one endpoint "GetEnterpriseInstallation"
* Option B: Leave this one endpoint "FindEnterpriseInstallation"
* Option C: Rename (breaking API change) ALL methods whose documentation URL starts with "get-" to "Get*" instead of "Find*" in this file
* Option D: Something else?

Not-Dhananjay-Mishra: I vote for Option C.
Because in my opinion, method needs to be consistent with GitHub API docs.

danyalahmed1995: Yeah, I agree Option C would be the cleanest long-term naming direction if the API is ever revisited more broadly.

For this PR, I used `FindEnterpriseInstallation` to stay consistent with the existing installation lookup methods in this file and avoid making a breaking API change as part of a small endpoint addition.

Let's wait for the other maintainers' votes as well.

alexandear: I vote for "Option B".

And "Option C" for the future PRs.

github/apps.go
```diff
@@ -457,6 +457,15 @@ func (s *AppsService) FindOrganizationInstallation(ctx context.Context, org stri
...
+//
+// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/apps/apps#get-an-enterprise-installation-for-the-authenticated-app
```
alexandear:
```suggestion
// GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/apps/apps?apiVersion=2022-11-28#get-an-enterprise-installation-for-the-authenticated-app
```

Because `s.getInstallation` works with the API version `2022-11-18`.

gmlewis:
OK, we have:
* 2 votes for "Option C"
* 1 vote for "Option B in this PR with a followup PR of Option C"

Does anyone else want to vote?

munlicode (approved):

github/apps.go
munlicode: same as @alexandear.

I think getting done with this PR, and then making breaking API change in the next PR would be better.

gmlewis (approved):
OK, sounds good to me. 
Merging.
Maybe someone can file an issue to address Option C, please.

danyalahmed1995:
@gmlewis Yes i think that would be a good idea .

gmlewis:
> Maybe someone can file an issue to address Option C, please.

No issue is necessary, as the thread above documents it just fine.
This is addressed in #4243.

PR #4225, opened 2026-05-14
feat: Add agent tasks API support

danyalahmed1995:
## Summary

Adds typed client support for GitHub's Agent Tasks REST API.

This implements the documented public-preview Agent Tasks endpoints:

- `GET /agents/repos/{owner}/{repo}/tasks`
- `POST /agents/repos/{owner}/{repo}/tasks`
- `GET /agents/repos/{owner}/{repo}/tasks/{task_id}`
- `GET /agents/tasks`
- `GET /agents/tasks/{task_id}`

Closes #4213.

## Changes

- Added `AgentTasksService` and registered it on `Client`.
- Added Agent Tasks request/response types for tasks, sessions, artifacts, list options, and create options.
- Added methods for listing, creating, and retrieving agent tasks.
- Added tests covering:
  - endpoint paths
  - HTTP methods
  - request bodies
  - response parsing
  - query parameters
  - pagination
  - preview API headers
- Added OpenAPI operation metadata for the new preview endpoints.
- Regenerated accessors and iterators.

## Notes

The Agent Tasks API is currently in public preview, so this PR keeps the implementation scoped to the fields and endpoints currently documented by GitHub.

## Validation

Ran the repository validation flow from `CONTRIBUTING.md`:

- `script/fmt.sh`
- `script/test.sh ./...`
- `script/lint.sh`
- `script/generate.sh`
- `git diff --check`

`script/test.sh` with the default race configuration could not run on this Windows machine because `cgo`/`gcc` was not available. I reran the full test script with the non-race fallback across all modules, and it passed.

PR #4236, opened 2026-05-18
refactor: Add constants for API versions

alexandear:
From https://github.com/google/go-github/pull/4225#discussion_r3252783500

cc: @stevehipwell

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -39,10 +39,15 @@ const (
 	HeaderRateUsed      = "X-Ratelimit-Used"
 	HeaderRequestID     = "X-Github-Request-Id"
 
-	defaultAPIVersion = "2022-11-28"
-	defaultBaseURL    = "https://api.github.com/"
-	defaultUserAgent  = "go-github" + "/" + Version
-	uploadBaseURL     = "https://uploads.github.com/"
+	// https://docs.github.com/en/rest/about-the-rest-api/api-versions#about-api-versioning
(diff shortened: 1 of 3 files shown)


stevehipwell (approved):
LGTM

github/github.go
```diff
@@ -39,10 +39,15 @@ const (
 	HeaderRateUsed      = "X-Ratelimit-Used"
 	HeaderRequestID     = "X-Github-Request-Id"
 
-	defaultAPIVersion = "2022-11-28"
-	defaultBaseURL    = "https://api.github.com/"
-	defaultUserAgent  = "go-github" + "/" + Version
-	uploadBaseURL     = "https://uploads.github.com/"
+	// https://docs.github.com/en/rest/about-the-rest-api/api-versions#about-api-versioning
+	defaultAPIVersion = api20221128
+	latestAPIVersion  = api20260310
```
gmlewis, line 44: Why do we need `latestAPIVersion` if it is never used anywhere and just exists to be a maintenance burden to update as simply being informative?

alexandear: It's for the future use. But I can remove it for now.

I thought it's better to add a constant instead of commenting that 2026 is the latest API.

stevehipwell: It might help provide context (for agents) too.

gmlewis (approved):

PR #4240, opened 2026-05-19
feat: Add support for setting client URLs

stevehipwell:
This PR adds support for setting the URLs without making changes to the path prefix as happens when using `github.WithEnterpriseURLs`. This function is useful when mocking the GitHub API as without it all test servers need modifying to add the GHAS prefix.

@gmlewis sorry but this might be worthy of a patch release, I discovered my oversight when I was wiring the new version up in the TF provider and while I was able to get the tests working I needed to modify the paths.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -444,34 +444,18 @@ func WithAuthToken(token string) ClientOptionsFunc {
 func WithURLs(baseURL, uploadURL *string) ClientOptionsFunc {
 	return func(o *clientOptions) error {
 		if baseURL != nil {
-			if *baseURL == "" {
-				return errors.New("base url must not be empty")
-			}
-
-			b, err := url.Parse(*baseURL)
+			b, err := parseURL(*baseURL)
 			if err != nil {
 				return fmt.Errorf("invalid base url: %w", err)
-			}
-
-			if !strings.HasSuffix(b.Path, "/") {
-				b.Path += "/"
 			}
 
 			o.baseURL = b
 		}
 
 		if uploadURL != nil {
-			if *uploadURL == "" {
-				return errors.New("upload url must not be empty")
-			}
-
-			u, err := url.Parse(*uploadURL)
+			u, err := parseURL(*uploadURL)
 			if err != nil {
 				return fmt.Errorf("invalid upload url: %w", err)
-			}
-
-			if !strings.HasSuffix(u.Path, "/") {
-				u.Path += "/"
 			}
 
 			o.uploadURL = u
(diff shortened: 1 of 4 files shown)


github/github.go
```diff
@@ -438,6 +438,49 @@ func WithAuthToken(token string) ClientOptionsFunc {
...
+			}
+
+			u, err := url.Parse(*uploadURL)
+			if err != nil {
+				return fmt.Errorf("invalid upload url: %w", err)
+			}
+
+			if !strings.HasSuffix(u.Path, "/") {
+				u.Path += "/"
+			}
```
gmlewis: All of this looks like almost identical code elsewhere. Could this be a helper function?

gmlewis (changes requested):

github/github.go
stevehipwell: Yeah, it could be.

stevehipwell: Done.

github/github_test.go
```diff
@@ -4417,7 +4506,7 @@ func TestClientCopy_leak_transport(t *testing.T) {
 		accessToken := r.Header.Get("Authorization")
 		_, _ = fmt.Fprintf(w, `{"login": "%v"}`, accessToken)
 	}))
-	clientPreconfiguredWithURLs := mustNewClient(t, WithEnterpriseURLs(srv.URL, srv.URL))
+	clientPreconfiguredWithURLs := mustNewClient(t, WithURLs(&srv.URL, &srv.URL))
```
gmlewis, line 4509: Why is `TestWithEnterpriseURLs` calling `WithURLs`?

gmlewis (changes requested):

github/github_test.go
stevehipwell: It's not, but GitHub does make it look like it is (I wasted a not insignificant amount of time on this yesterday)! This is a completely separate test where the new `WithURLs` is a better fit.

gmlewis: Ah, I see.  My bad.  Sorry for the noise.

gmlewis (approved):
LGTM.
Awaiting second LGTM+Approval from any other contributor to the repo before merging.

munlicode (approved):
LGTM

gmlewis:
Thank you, @munlicode!
Merging.

stevehipwell:
@gmlewis would it be possible to get a patch out for this?

gmlewis:
> @gmlewis would it be possible to get a patch out for this?

I may be able to do one this weekend.

gmlewis:
> > @gmlewis would it be possible to get a patch out for this?
> 
> I may be able to do one this weekend.

@stevehipwell - this is now available here:
https://github.com/google/go-github/releases/tag/v88.0.0

Issue #4237, opened 2026-05-19
Move compatibility for non github.com targets to `github.With<THING>` options


stevehipwell:
I propose that all compatibility functionality not required for the github.com target be removed from the default client setup and only be included via a `github.With<THING>` option. This (hopefully) would make the default client surface area simpler to understand. The justification for this is that you already need to use `github.WithEnterpriseURLs` to target non github.com. I propose the following changes to adopt this pattern, with the potential for other capabilities (such as enabling/disabling preview endpoints) to be added later.

- Default to the latest API version (`2026-03-10`)
  - Functions that require a specific version now check against the client version and error if not valid
- Add `github.WithAPIVersion` to explicitly set a version
  - Set at the client/service level
- Add `github.WithAdvancedServer` to enable GHAS support
  - Better validation of inputs
  - Support setting version
  - Add compatibility accordingly
- Add `github.WithEnterpriseCloud` to enable GHEC support
  - Better validation of inputs
- Deprecate `github.WithEnterpriseURLs`

I'd appreciate any feedback on this proposal.

cc: @gmlewis - @alexandear - @zyfy29 - @Not-Dhananjay-Mishra - @munlicode

gmlewis:
That sounds like a good plan to me.
The only thing that I'm a bit confused about is this:
> Functions that require a specific version now check against the client version and error if not valid

It was my understanding that endpoints that need a specific version can state that version explicitly so that the end user doesn't even need to worry about API versioning, and so we would handle all this for them in this client library.

alexandear:
The idea in general is great, and I fully support it.

> Default to the latest API version
> Add github.WithAPIVersion to explicitly set a version

My only concern is migrating to the latest API version. This will be done in #4077. It requires a LOT of work.

stevehipwell:
@gmlewis at the moment we set the version explicitly, which works fine for github.com where all of the version are currently present. But this doesn't provide an explicit error if the API version is unavailable and doesn't quite fit with an explicit version pattern. I was thinking that endpoints requiring a version check would do something like below:

```go
minVersion := api20260310
if s.version < api20260310 {
	return VersionNotSupportedErr
}
```

@alexandear using the latest version could be separated from the other changes. We could still manage the client version in cases where it makes sense (such as GHAS), but leave the current defaults (effectively unset) until we were in a position to make the latest version default.

danyalahmed1995:
I think it may be safer to split the API version work from the enterprise / compatibility option cleanup.

One possible path:

[...]

PR #4246, opened 2026-05-21
feat: Add client API version support

stevehipwell:
~~This PR makes the API version used by the client configurable via the new `github.WithAPIVersion` option function. There is also a new `Client.CheckAPIVersion` method to validate if the client is currently configured to support a minimum version. I've used this new pattern in the service methods which require the latest API version.~~

This PR moves the API version selection logic into the client and adds support for setting a min and a max version to constrain the default value. The intended behaviour is that the library maintainers explicitly set the client default version in code and add overrides to the functions where a non-default version is required. Library consumers will then be able to constrain the max and min versions so they get a sentinel error response when they attempt to call an unsupported endpoint.

As it currently stands no behaviour has been changed and this PR just introduces the primitives to extend in future PRs. The first of which I suggest adds `github.WithAdvancedServer` to set the URLs with correct suffixes and the max version to `2022-11-28`.

This is the first part of the implementation for #4237.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -159,9 +159,6 @@ const (
 	mediaTypeContentAttachmentsPreview = "application/vnd.github.corsair-preview+json"
 )
 
-// apiVersionRegexp is a regular expression to validate API version strings.
-var apiVersionRegexp = regexp.MustCompile(`^\d{4}-\d{2}-\d{2}$`)
-
 // ErrPathForbidden is returned when a URL path contains ".." as a path
 // segment, which could allow path traversal attacks.
 var ErrPathForbidden = errors.New("path must not contain '..' due to auth vulnerability issue")
--- a/github/github.go
+++ b/github/github.go
@@ -161,6 +161,9 @@ const (
 	mediaTypeContentAttachmentsPreview = "application/vnd.github.corsair-preview+json"
 )
 
+// apiVersionRegexp is a regular expression to validate API version strings.
+var apiVersionRegexp = regexp.MustCompile(`^\d{4}-\d{2}-\d{2}$`)
+
 // ErrPathForbidden is returned when a URL path contains ".." as a path
 // segment, which could allow path traversal attacks.
 var ErrPathForbidden = errors.New("path must not contain '..' due to auth vulnerability issue")
@@ -177,6 +180,9 @@ type Client struct {
 
 	// Base URL for uploading files.
 	uploadURL *url.URL
+
+	// API version to set in the X-Github-Api-Version header.
+	apiVersion string
 
 	// User agent used when communicating with the GitHub API.
 	userAgent string
@@ -347,6 +353,7 @@ type clientOptions struct {
 	httpClient                              *http.Client
 	transport                               http.RoundTripper
 	timeout                                 *time.Duration
+	apiVersion                              *string
 	userAgent                               *string
 	envProxy                                bool
 	token                                   *string
@@ -401,6 +408,27 @@ func WithTimeout(timeout time.Duration) ClientOptionsFunc {
 		}
 
 		o.timeout = &timeout
+		return nil
+	}
+}
+
+// WithAPIVersion returns a ClientOptionsFunc that sets the API version for a
+// Client. The API version should be in the format "YYYY-MM-DD" as specified by
+// GitHub's API versioning scheme. If not set, the default API version will be
+// used.
+// Warning: Setting the API version to anything other than [defaultAPIVersion]
+// is not recommended and may cause compatibility issues with this package.
+func WithAPIVersion(apiVersion string) ClientOptionsFunc {
+	return func(o *clientOptions) error {
+		if apiVersion == "" {
+			return errors.New("api version must not be empty")
+		}
+
+		if !apiVersionRegexp.MatchString(apiVersion) {
+			return errors.New("invalid api version")
+		}
+
+		o.apiVersion = &apiVersion
 		return nil
 	}
 }
@@ -609,6 +637,12 @@ func newClient(opts clientOptions) (*Client, error) {
 		CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse },
 	}
 
+	if opts.apiVersion != nil {
+		c.apiVersion = *opts.apiVersion
+	} else {
+		c.apiVersion = defaultAPIVersion
+	}
+
 	if opts.userAgent != nil {
 		c.userAgent = *opts.userAgent
 	} else {
@@ -684,6 +718,18 @@ func newClient(opts clientOptions) (*Client, error) {
 	return c, nil
 }
 
+// APIVersion returns the API version set in the X-Github-Api-Version header
+// for the client.
+func (c *Client) APIVersion() string {
+	return c.apiVersion
+}
+
+// CheckAPIVersion checks if the client's API version is compatible with the
+// provided minimum version.
+func (c *Client) CheckAPIVersion(minVersion string) bool {
+	return minVersion <= c.apiVersion
+}
+
 // UserAgent returns the User-Agent header value for the client.
 func (c *Client) UserAgent() string {
 	return c.userAgent
@@ -718,6 +764,7 @@ func (c *Client) Clone(opts ...ClientOptionsFunc) (*Client, error) {
 	}
 
 	o := clientOptions{
+		apiVersion:                              &c.apiVersion,
 		userAgent:                               &c.userAgent,
 		baseURL:                                 Ptr(*c.baseURL),
 		uploadURL:                               Ptr(*c.uploadURL),
@@ -814,7 +861,7 @@ func (c *Client) NewRequest(ctx context.Context, method, urlStr string, body any
 	if c.userAgent != "" {
 		req.Header.Set("User-Agent", c.userAgent)
 	}
-	req.Header.Set(headerAPIVersion, defaultAPIVersion)
+	req.Header.Set(headerAPIVersion, c.apiVersion)
 
 	for _, opt := range opts {
 		opt(req)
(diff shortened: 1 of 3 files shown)


github/private_registries.go
```diff
@@ -348,7 +378,13 @@ func (s *PrivateRegistriesService) UpdateOrganizationPrivateRegistry(ctx context
 func (s *PrivateRegistriesService) DeleteOrganizationPrivateRegistry(ctx context.Context, org, secretName string) (*Response, error) {
 	u := fmt.Sprintf("orgs/%v/private-registries/%v", org, secretName)
 
-	req, err := s.client.NewRequest(ctx, "DELETE", u, nil, WithVersion(api20260310))
+	minVersion := api20260310
+	reqOpts := []RequestOption{}
+	if !s.client.CheckAPIVersion(minVersion) {
+		reqOpts = append(reqOpts, WithVersion(minVersion))
+	}
```
gmlewis: How about making a helper function for all of these identical blocks?

```go
reqOpts := s.client.requestAPIMinVersionOpts()
```

gmlewis (changes requested):

github/private_registries.go
gmlewis: Or you could even make it private since nobody extermally should need it,.

gmlewis: In fact, I'm wondering if any user of this repo would ever need `CheckAPIVersion`?
Maybe it could _also_ be private?

gmlewis: Or, alternatively, maybe this is better:

```go
reqOpts := s.client.requestAPIMinVersionOpts(api20260310)
```

stevehipwell: @gmlewis I had similar thoughts about `CheckAPIVersion` and came to the conclusion that as the client can be used independently it's valid to be exported. I think the helper function makes sense.

github/github.go
```diff
@@ -405,6 +412,27 @@ func WithTimeout(timeout time.Duration) ClientOptionsFunc {
 	}
 }
 
+// WithAPIVersion returns a ClientOptionsFunc that sets the API version for a
+// Client. The API version should be in the format "YYYY-MM-DD" as specified by
+// GitHub's API versioning scheme. If not set, the default API version will be
+// used.
+// Warning: Setting the API version to anything other than [defaultAPIVersion]
+// is not recommended and may cause compatibility issues with this package.
+func WithAPIVersion(apiVersion string) ClientOptionsFunc {
```
Not-Dhananjay-Mishra: @stevehipwell can an example usage of this be added?

stevehipwell: > // Warning: Setting the API version to anything other than [defaultAPIVersion]
> // is not recommended and may cause compatibility issues with this package.

I've included this to be used for testing or as an escape hatch, I don't think we want to add an example until we're actually supporting running the client with the non-default version.

stevehipwell:
@gmlewis @Not-Dhananjay-Mishra I've reworked this whole PR (see updated description), please could you take another look?

gmlewis (approved):
Is `WithAPIVersion` going to be added in a follow-up PR?

stevehipwell:
> Is WithAPIVersion going to be added in a follow-up PR?

@gmlewis it could be, but I'm not sure what the purpose of it would be or the actual logic required to implement it "correctly".

While working through these changes I realised that the `go-github` code couples an implementation to an API version. So by setting an immutable default version that can be overridden on a per endpoint basis I kept the existing logic but made it easier to change how this works in the future. Then by adding min and max versions we can support constrained use cases (e.g. GHAS) without increasing the maintenance burden. It also leaves the door open for more complex patterns in the future (e.g. branching logic), if it's required.

gmlewis:
> > Is WithAPIVersion going to be added in a follow-up PR?
> 
> @gmlewis it could be, but I'm not sure what the purpose of it would be or the actual logic required to implement it "correctly".
> 
> While working through these changes I realised that the `go-github` code couples an implementation to an API version. So by setting an immutable default version that can be overridden on a per endpoint basis I kept the existing logic but made it easier to change how this works in the future. Then by adding min and max versions we can support constrained use cases (e.g. GHAS) without increasing the maintenance burden. It also leaves the door open for more complex patterns in the future (e.g. branching logic), if it's required.

Yes, my original understanding was that there was no need to concern the user over API versions at all, and we would simply update this client library as the versions changed. I don't even understand why we need a "min" or "max". I thought we would just keep each endpoint in sync with the versions that the GitHub server supplies.

stevehipwell:
> Yes, my original understanding was that there was no need to concern the user over API versions at all, and we would simply update this client library as the versions changed. I don't even understand why we need a "min" or "max". I thought we would just keep each endpoint in sync with the versions that the GitHub server supplies.

The simple first improvement is that you get a sentinel error when the constraints are set and you attempt to use an unsupported endpoint. The second benefit is to be able to design a pattern that allows us to update endpoints for DotCom without being blocked by GHAS; I think we could take a look at the get content functionality to see what this would look like.

As a more abstract concern, I predict that the next new API version is going to be along sooner than 4 years from now.

IMHO the version setting functionality (currently `WithVersion`) is the thing that likely needs to change to really make use of the new version constraints as the current pattern isn't expressive enough to model the actual system.

stevehipwell:
@gmlewis where are we with this?

github/github_test.go
```diff
@@ -3126,6 +3144,106 @@ func TestDo_noContent(t *testing.T) {
...
+			client.apiVersionMin = tt.versionMin
+			client.apiVersionMax = tt.versionMax
+
+			req, _ := http.NewRequestWithContext(t.Context(), "GET", ".", nil)
+			req.Header.Set(headerAPIVersion, tt.version)
+
+			err := client.checkRequestAPIVersionBeforeDo(req)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal("Expected error to be returned, got nil.")
```
alexandear: for consistency:
```suggestion
					t.Fatal("Expected error to be returned, got nil")
```

github/github_test.go
```diff
@@ -3126,6 +3144,106 @@ func TestDo_noContent(t *testing.T) {
...
+				}
+				if !errors.Is(err, ErrUnsupportedAPIVersion) {
+					t.Errorf("Expected ErrUnsupportedAPIVersion; got %#v.", err)
```
```suggestion
					t.Errorf("Expected ErrUnsupportedAPIVersion; got %#v", err)
```

github/github_test.go
```diff
@@ -3126,6 +3144,106 @@ func TestDo_noContent(t *testing.T) {
...
+	c := mustNewClient(t)
+	c.apiVersionMin = api20221128
+	c.apiVersionMax = api20221128
+
+	req, _ := http.NewRequestWithContext(t.Context(), "GET", ".", nil)
+	req.Header.Set(headerAPIVersion, api20260310)
+
+	_, err := c.bareDo(c.client, req)
+	if err == nil {
+		t.Fatal("Expected error to be returned, got nil.")
```
```suggestion
		t.Fatal("Expected error to be returned, got nil")
```

github/github_test.go
```diff
@@ -3126,6 +3144,106 @@ func TestDo_noContent(t *testing.T) {
...
+	}
+	if !errors.Is(err, ErrUnsupportedAPIVersion) {
+		t.Errorf("Expected ErrUnsupportedAPIVersion; got %#v.", err)
```
```suggestion
		t.Errorf("Expected ErrUnsupportedAPIVersion; got %#v", err)
```

github/github.go
```diff
@@ -1143,6 +1164,28 @@ const (
 // unexpectedly large error body.
 const maxErrorBodySize = 1 * 1024 * 1024 // 1 MiB
 
+// ErrUnsupportedAPIVersion is returned when the API version specified in the
+// request is not supported by the client.
+var ErrUnsupportedAPIVersion = errors.New("unsupported api version")
```
alexandear, line 1169:
```suggestion
var ErrUnsupportedAPIVersion = errors.New("unsupported API version")
```

stevehipwell: Shouldn't errors be lowercase?

stevehipwell:
@alexandear I fixed up the consistency items and applied the pattern to the whole test file.

alexandear (approved):

gmlewis:
> > Yes, my original understanding was that there was no need to concern the user over API versions at all, and we would simply update this client library as the versions changed. I don't even understand why we need a "min" or "max". I thought we would just keep each endpoint in sync with the versions that the GitHub server supplies.
> 
> The simple first improvement is that you get a sentinel error when the constraints are set and you attempt to use an unsupported endpoint. 

I'm confused here because I don't understand how a user could even possibly attempt to use an unsupported endpoint if we maintain this repo properly and keep the methods matching the API versions documented by the official GitHub v3 API.

> The second benefit is to be able to design a pattern that allows us to update endpoints for DotCom without being blocked by GHAS; I think we could take a look at the get content functionality to see what this would look like.

Maybe I simply don't understand how we could be blocked by GHAS.

> As a more abstract concern, I predict that the next new API version is going to be along sooner than 4 years from now.
> 
> IMHO the version setting functionality (currently `WithVersion`) is the thing that likely needs to change to really make use of the new version constraints as the current pattern isn't expressive enough to model the actual system.

Possibly... how much do we want to speculate here? I don't know.

I would really appreciate it if we get some sort of helpful feedback from our other reviewers here because I don't have a good grasp on what is needed and/or useful/helpful to users of this library at this point for versionining, or what the pain points are.

stevehipwell:
@gmlewis basically GHES and DotCom don't follow the same release cycle. So if you use one of the new 2026 only endpoints in the latest package release with GHES you'll get an error as the 2026 version hasn't been released to the latest GHES yet. This will be the case until all GHES instances have been updated to a version with support for 2026. Before this happens 2022 may have been removed from DotCom, but more likely we want to use 2026 for some of the endpoints we still need to support on GHES.

gmlewis:
Thank you, @alexandear!
Are your review comments all resolved, @Not-Dhananjay-Mishra?

Not-Dhananjay-Mishra (approved):

gmlewis:
Merging.

PR #4253, opened 2026-05-25
test: Add attestation marshal tests

KirtiRamchandani:
Problem
Artifact attestation payload types are shared by organization, repository, and user attestation APIs, but the shared resource structs did not have dedicated JSON marshal coverage tracked by #55.

Root cause
Existing service tests exercise API decoding paths, while the resource-level marshal round-trip tests skipped Attestation and AttestationsResponse.

Solution
Add marshal tests for empty and populated Attestation and AttestationsResponse values. The populated cases use the existing json.RawMessage comparator so Sigstore bundle JSON is compared semantically.

Tests run
- gofmt -w github/attestations_test.go
- go test ./github -run 'TestAttestation(sResponse)?_Marshal'
- go test ./github
- go test ./...
- git diff --cached --check

Linked issue
Part of #55

### DIFF
--- a/github/attestations_test.go
+++ b/github/attestations_test.go
@@ -1,4 +1,4 @@
-// Copyright 2024 The go-github AUTHORS. All rights reserved.
+// Copyright 2026 The go-github AUTHORS. All rights reserved.
 //
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
--- a/github/attestations_test.go
+++ b/github/attestations_test.go
@@ -1,0 +1,71 @@
+// Copyright 2024 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package github
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestAttestation_Marshal(t *testing.T) {
+	testJSONMarshalOnly(t, &Attestation{}, `{"bundle": null, "repository_id": 0}`)
+
+	u := &Attestation{
+		Bundle:       json.RawMessage(`{"mediaType":"application/vnd.dev.sigstore.bundle.v0.3+json"}`),
+		RepositoryID: 1,
+	}
+
+	want := `{
+		"bundle": {
+			"mediaType": "application/vnd.dev.sigstore.bundle.v0.3+json"
+		},
+		"repository_id": 1
+	}`
+
+	testJSONMarshal(t, u, want, cmpJSONRawMessageComparator())
+}
+
+func TestAttestationsResponse_Marshal(t *testing.T) {
+	testJSONMarshal(t, &AttestationsResponse{}, `{"attestations": null}`)
+
+	u := &AttestationsResponse{
+		Attestations: []*Attestation{
+			{
+				Bundle:       json.RawMessage(`{"verificationMaterial":{"certificate":{"rawBytes":"abc"}}}`),
+				RepositoryID: 1,
+			},
+			{
+				Bundle:       json.RawMessage(`{"dsseEnvelope":{"payload":"def"}}`),
+				RepositoryID: 2,
+			},
+		},
+	}
+
+	want := `{
+		"attestations": [
+			{
+				"bundle": {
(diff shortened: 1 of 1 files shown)


github/attestations_test.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2024 The go-github AUTHORS. All rights reserved.
```
gmlewis:
```suggestion
// Copyright 2026 The go-github AUTHORS. All rights reserved.
```

gmlewis (changes requested):

KirtiRamchandani:
Thanks, updated the new test file copyright year to 2026 and reran the focused marshal test plus diff check.

github/attestations_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+package github
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestAttestation_Marshal(t *testing.T) {
+	testJSONMarshalOnly(t, &Attestation{}, `{"bundle": null, "repository_id": 0}`)
```
Not-Dhananjay-Mishra, line 14:
```suggestion
func TestAttestation_Marshal(t *testing.T) {
	t.Parallel()
	testJSONMarshalOnly(t, &Attestation{}, `{"bundle": null, "repository_id": 0}`)
```

github/attestations_test.go
```diff
@@ -0,0 +1,71 @@
...
+			"mediaType": "application/vnd.dev.sigstore.bundle.v0.3+json"
+		},
+		"repository_id": 1
+	}`
+
+	testJSONMarshal(t, u, want, cmpJSONRawMessageComparator())
+}
+
+func TestAttestationsResponse_Marshal(t *testing.T) {
+	testJSONMarshal(t, &AttestationsResponse{}, `{"attestations": null}`)
```
Not-Dhananjay-Mishra, line 32:
```suggestion
func TestAttestationsResponse_Marshal(t *testing.T) {
	t.Parallel()
	testJSONMarshal(t, &AttestationsResponse{}, `{"attestations": null}`)
```

alexandear (commented):
No need for `Attestation` and `AttestationsResponse` JSON marshal tests. We should only add tests for types with custom `MarshalJSON` or `UnmarshalJSON` implementations. For example:

- https://github.com/google/go-github/blob/d40cb978086391bed8115ca709affc454b9a5bee/github/repos_environments.go#L156
- https://github.com/google/go-github/blob/d40cb978086391bed8115ca709affc454b9a5bee/github/projects.go#L205

@gmlewis what do you think? These tests, generated by an LLM, don't actually validate anything meaningful and only add maintenance overhead.

gmlewis:
> No need for `Attestation` and `AttestationsResponse` JSON marshal tests. We should only add tests for types with custom `MarshalJSON` or `UnmarshalJSON` implementations. For example:
> 
> * https://github.com/google/go-github/blob/d40cb978086391bed8115ca709affc454b9a5bee/github/repos_environments.go#L156
> * https://github.com/google/go-github/blob/d40cb978086391bed8115ca709affc454b9a5bee/github/projects.go#L205
> 
> @gmlewis what do you think? These tests, generated by an LLM, don't actually validate anything meaningful and only add maintenance overhead.

@alexandear - that's a good point. I'm wondering if #55 should simply be closed now and we should stop "advertising" this open-ended invitation for PRs.  This one issue was the sole target for previous Hacktoberfest contributions, and maybe it is no longer really needed.  Thoughts?

alexandear:
> This one issue was the sole target for previous Hacktoberfest contributions, and maybe it is no longer really needed. Thoughts?

I vote for closing the issue.

Also, we can simply write a tool to generate marshal/unmarshal tests. But I'm not sure if we need.

gmlewis:
> > This one issue was the sole target for previous Hacktoberfest contributions, and maybe it is no longer really needed. Thoughts?
> 
> I vote for closing the issue.
> 
> Also, we can simply write a tool to generate marshal/unmarshal tests. But I'm not sure if we need.

Agreed.  Thank you, @alexandear, and thank you, @KirtiRamchandani for your understanding. Closing.

PR #4289, opened 2026-06-09
test: Replace `&x` variables with inline `Ptr(value)` calls

alexandear:
This PR refactors tests. It replaces the pattern of declaring a temporary variable only to take its address:

```
name := "User One"
// ...
field: &name,
```

with an inline `Ptr()` call:

```
field: Ptr("User One"),
```

I created a custom golangci-lint linter to detect such cases, but it is not worth adding it because it increases the maintenance overhead.

### DIFF
--- a/github/copilot_test.go
+++ b/github/copilot_test.go
@@ -1332,13 +1332,11 @@ func TestCopilotService_GetEnterpriseMetrics(t *testing.T) {
 		t.Errorf("Copilot.GetEnterpriseMetrics returned error: %v", err)
 	}
 
-	totalActiveUsers := 24
-	totalEngagedUsers := 20
 	want := []*CopilotMetrics{
 		{
 			Date:              "2024-06-24",
-			TotalActiveUsers:  &totalActiveUsers,
-			TotalEngagedUsers: &totalEngagedUsers,
+			TotalActiveUsers:  Ptr(24),
+			TotalEngagedUsers: Ptr(20),
 			CopilotIDECodeCompletions: &CopilotIDECodeCompletions{
 				TotalEngagedUsers: 20,
 				Languages: []*CopilotIDECodeCompletionsLanguage{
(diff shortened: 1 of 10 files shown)


gmlewis (commented):
I find this PR harder to read and reason about than what we had previously, and I do not see the benefit. I personally would prefer to close this PR but I'm willing to listen to any arguments as to why we should proceed with this.

alexandear:
> I find this PR harder to read and reason about than what we had previously, and I do not see the benefit. 

Previously in [this discussion](https://github.com/google/go-github/pull/4145#discussion_r3061348086) we had slightly different code, where the variable was used twice.

```go
lastActiveOn := &Timestamp{time.Date(2023, 4, 26, 15, 23, 37, 0, time.UTC)}
// ...
LastActiveOn: Ptr(lastActiveOn)
// ...
LastActiveOn: Ptr(lastActiveOn)
```

But in this PR variables are used only once.

> I personally would prefer to close this PR but I'm willing to listen to any arguments as to why we should proceed with this.

The main arguments for `Ptr(value)` inline:
- Eliminates the "temp var just for address" anti-pattern - the variable name adds no semantic meaning
- Consistent with how the rest of the codebase already uses `Ptr()`
- Reduces noise: one line instead of two

gmlewis (approved):
OK, excellent points, @alexandear ... thank you!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

stevehipwell (approved):
LGTM

@alexandear I assume that the intention here is to replace `Ptr` with `new` when we upgrade to 1.26?

alexandear:
> @alexandear I assume that the intention here is to replace `Ptr` with `new` when we upgrade to 1.26?

Yes, it is.

The command `go fix -newexpr ./...` will do this automatically.

gmlewis:
Thank you, @alexandear, @Not-Dhananjay-Mishra, and @stevehipwell!
Merging.

PR #4297, opened 2026-06-12
fix!: Fix `LicenseStatus` response and `Supportkey` type

### LINKED ISSUE #4296: Incorrect return and data type in license api
While developing for api endpoint `/manage/v1/config/license` we detected 2 issues, which are not consistent with api docs
https://docs.github.com/en/enterprise-server@3.19/rest/enterprise-admin/manage-ghes?apiVersion=2022-11-28#get-the-enterprise-license-information. We opened support case, and verified with github support our findings.
1. The response is advertised as: [{...}] but actual response is {...}
2. supportKey is actual a bool return.

Example response:
curl -u api_key:xxxx https://hostname:8443/manage/v1/config/license
{"advancedSecurityEnabled":false, "advancedSecuritySeats":0, "clusterSupport":false, "company":"yes", "croquetSupport":true, "customTerms":false, "evaluation":false, "expireAt":"2029-01-01T07:59:59Z", "insightsEnabled":false, "insightsExpireAt":"0001-01-01T00:00:00Z", "learningLabEvaluationExpires":"0001-01-01T00:00:00Z", "learningLabSeats":0, "perpetual":false, "referenceNumber":"zyz", "seats":100, "sshAllowed":true, "supportKey":false, "unlimitedSeating":false}

s4mur4i:
BREAKING CHANGE: `LicenseStatus.SupportKey` type changed from `*string` to `*bool` and `License` return type is no longer a slice.

Closes: https://github.com/google/go-github/issues/4296

License status response was implemented according to documentation, which is not according to actual response.

Fix:
Corrected the response
Fixed incorrect data type for SupportKey
Fixed tests

### DIFF
--- a/github/enterprise_manage_ghes_config.go
+++ b/github/enterprise_manage_ghes_config.go
@@ -89,7 +89,9 @@ type LicenseStatus struct {
 	ReferenceNumber              *string    `json:"referenceNumber,omitempty"`
 	Seats                        *int       `json:"seats,omitempty"`
 	SSHAllowed                   *bool      `json:"sshAllowed,omitempty"`
-	SupportKey                   *bool      `json:"supportKey,omitempty"`
+	// SupportKey is documented as a string, but the actual response is a bool.
+	// TODO: Remove this note once GitHub corrects the schema documentation.
+	SupportKey       *bool `json:"supportKey,omitempty"`
 	UnlimitedSeating             *bool      `json:"unlimitedSeating,omitempty"`
 }
 
--- a/github/enterprise_manage_ghes_config.go
+++ b/github/enterprise_manage_ghes_config.go
@@ -89,7 +89,7 @@ type LicenseStatus struct {
 	ReferenceNumber              *string    `json:"referenceNumber,omitempty"`
 	Seats                        *int       `json:"seats,omitempty"`
 	SSHAllowed                   *bool      `json:"sshAllowed,omitempty"`
-	SupportKey                   *bool `json:"supportKey,omitempty"`
+	SupportKey                   *bool      `json:"supportKey,omitempty"`
 	UnlimitedSeating             *bool      `json:"unlimitedSeating,omitempty"`
 }
 
--- a/github/enterprise_manage_ghes_config.go
+++ b/github/enterprise_manage_ghes_config.go
@@ -70,8 +70,8 @@ type InitialConfigOptions struct {
 }
 
 // LicenseStatus is a struct to hold the response from the License API.
-// SupportKey is documented as string but is actual a bool.
-// TODO: Remove comment after github updates schema documentation
+// SupportKey is documented as a string, but the actual response is a bool.
+// TODO: Remove this note once GitHub corrects the schema documentation.
 type LicenseStatus struct {
 	AdvancedSecurityEnabled      *bool      `json:"advancedSecurityEnabled,omitempty"`
 	AdvancedSecuritySeats        *int       `json:"advancedSecuritySeats,omitempty"`
@@ -89,9 +89,7 @@ type LicenseStatus struct {
 	ReferenceNumber              *string    `json:"referenceNumber,omitempty"`
 	Seats                        *int       `json:"seats,omitempty"`
 	SSHAllowed                   *bool      `json:"sshAllowed,omitempty"`
-	// SupportKey is documented as a string, but the actual response is a bool.
-	// TODO: Remove this note once GitHub corrects the schema documentation.
-	SupportKey       *bool `json:"supportKey,omitempty"`
+	SupportKey                   *bool `json:"supportKey,omitempty"`
 	UnlimitedSeating             *bool      `json:"unlimitedSeating,omitempty"`
 }
 
--- a/github/enterprise_manage_ghes_config.go
+++ b/github/enterprise_manage_ghes_config.go
@@ -70,6 +70,8 @@ type InitialConfigOptions struct {
 }
 
 // LicenseStatus is a struct to hold the response from the License API.
+// SupportKey is documented as string but is actual a bool.
+// TODO: Remove comment after github updates schema documentation
 type LicenseStatus struct {
 	AdvancedSecurityEnabled      *bool      `json:"advancedSecurityEnabled,omitempty"`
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @s4mur4i!
LGTM.

github/enterprise_manage_ghes_config.go
```diff
@@ -354,14 +354,14 @@ func (s *EnterpriseService) InitialConfig(ctx context.Context, license, password
 // GitHub API docs: https://docs.github.com/enterprise-server@3.21/rest/enterprise-admin/manage-ghes#get-the-enterprise-license-information
 //
 //meta:operation GET /manage/v1/config/license
-func (s *EnterpriseService) License(ctx context.Context) ([]*LicenseStatus, *Response, error) {
+func (s *EnterpriseService) License(ctx context.Context) (*LicenseStatus, *Response, error) {
```
alexandear, line 363: Let's add a comment to the function that GitHub documentation is wrong about the response.

github/enterprise_manage_ghes_config.go
```diff
@@ -87,7 +87,7 @@ type LicenseStatus struct {
 	ReferenceNumber              *string    `json:"referenceNumber,omitempty"`
 	Seats                        *int       `json:"seats,omitempty"`
 	SSHAllowed                   *bool      `json:"sshAllowed,omitempty"`
-	SupportKey                   *string    `json:"supportKey,omitempty"`
+	SupportKey                   *bool      `json:"supportKey,omitempty"`
```
alexandear: Let's add a comment that GitHub documentation is wrong about the actual type.

github/enterprise_manage_ghes_config.go
s4mur4i: Their last response was:
```
We’ll treat this as a documentation/schema mismatch and route it for review so the example can be corrected to match the actual GHES behavior.
```
Eventually it should be corrected, in that case wouldn't that cause confusion?

github/enterprise_manage_ghes_config.go
s4mur4i: Their last response was:
```
We’ll treat this as a documentation/schema mismatch and route it for review so the example can be corrected to match the actual GHES behavior.
```
Eventually it should be corrected, in that case wouldn't that cause confusion?

alexandear: We can mitigate this by adding another comment `TODO: remove the comment below when GitHub corrects the schema` or something similar.

s4mur4i: I added comment PTAL.

alexandear (approved):

github/enterprise_manage_ghes_config.go
```diff
@@ -354,14 +356,16 @@ func (s *EnterpriseService) InitialConfig(ctx context.Context, license, password
 // GitHub API docs: https://docs.github.com/enterprise-server@3.21/rest/enterprise-admin/manage-ghes#get-the-enterprise-license-information
 //
 //meta:operation GET /manage/v1/config/license
-func (s *EnterpriseService) License(ctx context.Context) ([]*LicenseStatus, *Response, error) {
+// Current Docs shouw incorrect return type. They list as [{...}] but actual is {...}
+// TODO: Remove comment after github updates schema documentation
```
alexandear: I propose adding a note:
```go
// License gets the current license information for the GitHub Enterprise instance.
//
// NOTE: The GitHub documentation incorrectly shows the return type as a list ([{...}]),
// but the actual response is a single object ({...}).
// TODO: Remove this note once GitHub corrects the schema documentation.
//
// GitHub API docs: https://docs.github.com/enterprise-server@3.21/rest/enterprise-admin/manage-ghes#get-the-enterprise-license-information
//
//meta:operation GET /manage/v1/config/license
```

github/enterprise_manage_ghes_config.go
alexandear:
```suggestion
	// SupportKey is documented as a string, but the actual response is a bool.
	// TODO: Remove this note once GitHub corrects the schema documentation.
	SupportKey       *bool `json:"supportKey,omitempty"`
```

github/enterprise_manage_ghes_config.go
s4mur4i: I cannot accept this, should I just update with commit?

s4mur4i: Added proposed note

github/enterprise_manage_ghes_config.go
```diff
@@ -70,6 +70,8 @@ type InitialConfigOptions struct {
 }
 
 // LicenseStatus is a struct to hold the response from the License API.
+// SupportKey is documented as string but is actual a bool.
+// TODO: Remove comment after github updates schema documentation
```
alexandear: This is no longer needed.

alexandear (approved):

s4mur4i:
@gmlewis I added a commit for fixing the lint findings. Hope it will be good like this.

github/enterprise_manage_ghes_config.go
gmlewis: @s4mur4i - please remove these two lines.

s4mur4i: The inline version was failing lint, so I moved it to header par, similar to licensestatus.
As I see lint output that part doesnt fail.

alexandear: We have plenty of inline comments for fields - they all passed linting. E.g.,

https://github.com/google/go-github/blob/0eacd3434ac048cfd1f8bb9da8e0824ef03ced86/github/billing.go#L131-L132

s4mur4i: I changed it back, Can linter be run again please.

alexandear:
The linter is failing because of formatting issues. You need to run `./script/fmt.sh`

s4mur4i:
> The linter is failing because of formatting issues. You need to run `./script/fmt.sh`

Ran it, commited results.

gmlewis:
Thank you, @s4mur4i and @alexandear!
Merging.

PR #4300, opened 2026-06-14
refactor: Fix parameters in method endpoints

alexandear:
This PR adds the `paramcheck` custom linter that detects issues with body and query parameters. Changes are non-breaking.

The `paramcheck` does the following:

- suggests renaming a body parameter to `body`
- reports body parameters passed by pointer, because body parameters are usually required
- reports body parameter types with an `Options` suffix, because `Options` are used for query parameters
- suggests renaming a query struct to `opts`
- reports query parameter passed by value, because query parameters are usually optional
- reports query parameter types without an `Options` suffix


gmlewis:
I'm wondering if we should standardize on an "Options" suffix for query (URL) parameters at the same time since we are standardizing on a "Request" suffix for body parameters?

If so, it seems to me like this new linter is the perfect place to put this.

Thoughts?

alexandear:
> I'm wondering if we should standardize on an "Options" suffix for query (URL) parameters at the same time since we are standardizing on a "Request" suffix for body parameters?
> 
> If so, it seems to me like this new linter is the perfect place to put this.
> 
> Thoughts?

Agree, good idea. I'll extend the linter.
Also should think of a better name for the linter.

alexandear:
@gmlewis I extended the linter, ready for review.

The linter found a few more problems, created #4301.

.custom-gcl.yml
```diff
@@ -6,6 +6,8 @@ plugins:
     path: ./tools/fmtpercentv
   - module: "github.com/google/go-github/v88/tools/redundantptr"
     path: ./tools/redundantptr
+  - module: "github.com/google/go-github/v88/tools/paramcheck"
```
gmlewis, line 7: Please keep these modules sorted.

.golangci.yml
```diff
@@ -28,6 +28,7 @@ linters:
     - paralleltest
     - perfsprint
     - redundantptr
+    - paramcheck
```
gmlewis: Please keep these linters sorted.

gmlewis (changes requested):
Thank you, @alexandear!
Two tiny nits, otherwise LGTM.

gmlewis (approved):

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

PR #4315, opened 2026-06-20
fix!: Send request body in SCIM update methods

### LINKED ISSUE #4301: Fix `SCIMService.UpdateProvisionedOrgMembership` and `SCIMService.UpdateAttributeForSCIMUser`
Two `SCIMService` methods in github/scim.go have implementations that don't match the GitHub API schema:
- [SCIMService.UpdateProvisionedOrgMembership](https://github.com/google/go-github/blob/98eb342cb0e99491962bc2090577d2f8b07e6ef5/github/scim.go#L170-L183)
- [SCIMService.UpdateAttributeForSCIMUser](https://github.com/google/go-github/blob/98eb342cb0e99491962bc2090577d2f8b07e6ef5/github/scim.go#L205-L218)

Both are breaking changes to fix.

Problems:

- Parameters are passed as query parameters, but must be in the body
- Missing return type
- `UpdateAttributeForSCIMUserOptions.Operations` should be an array of objects
- maybe some other; see documentation for these endpoints

## `UpdateProvisionedOrgMembership`

Current signature:

```go
func (s *SCIMService) UpdateProvisionedOrgMembership(ctx context.Context, org, scimUserID string, opts *SCIMUserAttributes) (*Response, error)
```

Expected signature:

```go
func (s *SCIMService) UpdateProvisionedOrgMembership(ctx context.Context, org, scimUserID string, body UpdateProvisionedOrgMembershipRequest) (*SCIMUserAttributes, *Response, error)
```

## `UpdateAttributeForSCIMUser`

Current signature:

```go
func (s *SCIMService) UpdateAttributeForSCIMUser(ctx context.Context, org, scimUserID string, opts *UpdateAttributeForSCIMUserOptions) (*Response, error)
```

Expected signature:

```go
func (s *SCIMService) UpdateAttributeForSCIMUser(ctx context.Context, org, scimUserID string, body UpdateAttributeForSCIMUserRequest) (*SCIMUserAttributes, *Response, error)
```

JamBalaya56562:
BREAKING CHANGE: `UpdateProvisionedOrgMembership` and `UpdateAttributeForSCIMUser` params and return values changed.

`SCIMService.UpdateProvisionedOrgMembership` and `UpdateAttributeForSCIMUser` passed their parameters through `addOptions`, which only serializes `url`-tagged fields. The SCIM types are `json`-tagged, so nothing was actually sent and both calls were effectively no-ops. They also discarded the user object the API returns.

Both methods now send the payload as the request body and return the updated `*SCIMUserAttributes`, following the existing `ProvisionAndInviteSCIMUser` pattern.

The body is taken by value via dedicated request types:

- new `UpdateProvisionedOrgMembershipRequest`
- `UpdateAttributeForSCIMUserRequest` (renamed from `UpdateAttributeForSCIMUserOptions`), whose `Operations` field is now a slice (`[]*UpdateAttributeForSCIMUserOperations`) with the `json:"Operations"` key required by the SCIM PatchOp schema, matching the existing enterprise SCIM implementation.

The `//nolint:paramcheck` workaround is removed. Tests verify the request body wire format and the decoded response.

This is a breaking API change.

Fixes: #4301.

### DIFF
--- a/github/scim.go
+++ b/github/scim.go
@@ -172,7 +172,6 @@ type UpdateProvisionedOrgMembershipRequest struct {
 	Schemas     []string         `json:"schemas,omitempty"`     // (Optional.)
 	ExternalID  *string          `json:"externalId,omitempty"`  // (Optional.)
 	Groups      []string         `json:"groups,omitempty"`      // (Optional.)
-	Roles       []*SCIMUserRole  `json:"roles,omitempty"`       // (Optional, GHES only.)
 	Active      *bool            `json:"active,omitempty"`      // (Optional.)
 }
 
--- a/github/scim.go
+++ b/github/scim.go
@@ -162,32 +162,48 @@ func (s *SCIMService) GetSCIMProvisioningInfoForUser(ctx context.Context, org, s
 	return user, resp, nil
 }
 
+// UpdateProvisionedOrgMembershipRequest represents the body of a request to
+// update a provisioned organization membership.
+type UpdateProvisionedOrgMembershipRequest struct {
+	UserName    string           `json:"userName"`              // Configured by the admin. Could be an email, login, or username. (Required.)
+	Name        SCIMUserName     `json:"name"`                  // (Required.)
+	DisplayName *string          `json:"displayName,omitempty"` // The name of the user, suitable for display to end-users. (Optional.)
+	Emails      []*SCIMUserEmail `json:"emails"`                // User emails. (Required.)
+	Schemas     []string         `json:"schemas,omitempty"`     // (Optional.)
+	ExternalID  *string          `json:"externalId,omitempty"`  // (Optional.)
+	Groups      []string         `json:"groups,omitempty"`      // (Optional.)
+	Roles       []*SCIMUserRole  `json:"roles,omitempty"`       // (Optional, GHES only.)
+	Active      *bool            `json:"active,omitempty"`      // (Optional.)
+}
+
 // UpdateProvisionedOrgMembership updates a provisioned organization membership.
 //
 // GitHub API docs: https://docs.github.com/enterprise-cloud@latest/rest/scim/scim?apiVersion=2022-11-28#update-a-provisioned-organization-membership
 //
 //meta:operation PUT /scim/v2/organizations/{org}/Users/{scim_user_id}
-func (s *SCIMService) UpdateProvisionedOrgMembership(ctx context.Context, org, scimUserID string, opts *SCIMUserAttributes) (*Response, error) { //nolint:paramcheck // see https://github.com/google/go-github/issues/4301
+func (s *SCIMService) UpdateProvisionedOrgMembership(ctx context.Context, org, scimUserID string, body UpdateProvisionedOrgMembershipRequest) (*SCIMUserAttributes, *Response, error) {
 	u := fmt.Sprintf("scim/v2/organizations/%v/Users/%v", org, scimUserID)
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/scim.go
```diff
@@ -162,32 +162,48 @@ func (s *SCIMService) GetSCIMProvisioningInfoForUser(ctx context.Context, org, s
...
+// update a provisioned organization membership.
+type UpdateProvisionedOrgMembershipRequest struct {
+	UserName    string           `json:"userName"`              // Configured by the admin. Could be an email, login, or username. (Required.)
+	Name        SCIMUserName     `json:"name"`                  // (Required.)
+	DisplayName *string          `json:"displayName,omitempty"` // The name of the user, suitable for display to end-users. (Optional.)
+	Emails      []*SCIMUserEmail `json:"emails"`                // User emails. (Required.)
+	Schemas     []string         `json:"schemas,omitempty"`     // (Optional.)
+	ExternalID  *string          `json:"externalId,omitempty"`  // (Optional.)
+	Groups      []string         `json:"groups,omitempty"`      // (Optional.)
+	Roles       []*SCIMUserRole  `json:"roles,omitempty"`       // (Optional, GHES only.)
```
Not-Dhananjay-Mishra:
```suggestion

```
Roles is not present in docs
Image

JamBalaya56562: Thanks @Not-Dhananjay-Mishra, good catch — confirmed. GitHub's GHEC OpenAPI description (both `2022-11-28` and `2026-03-10`) defines the `PUT /scim/v2/organizations/{org}/Users/{scim_user_id}` request body with only `userName`, `name`, `displayName`, `emails`, `schemas`, `externalId`, `groups` and `active`; `roles` only appears in the enterprise SCIM schemas (`/scim/v2/enterprises/...`), not the org SCIM endpoints. Dropped the `Roles` field in 5c74f864.

github/scim_test.go
```diff
@@ -291,46 +286,105 @@ func TestSCIMService_UpdateProvisionedOrgMembership(t *testing.T) {
...
+	wantBody := map[string]any{
+		"schemas": []any{SCIMSchemasURINamespacesPatchOp},
+		"Operations": []any{
+			map[string]any{
+				"op":    "replace",
+				"path":  "displayName",
+				"value": "NewDisplayName",
+			},
+		},
+	}
```
Not-Dhananjay-Mishra, line 353:
```suggestion

```
Can we remove this? as `body` and `wantBody` are same.

JamBalaya56562: Thanks @Not-Dhananjay-Mishra! They do hold the same data, but `wantBody` is intentionally a `map[string]any` rather than reusing `body`. If we pass the struct to `testJSONBody`, the body is marshaled and then unmarshaled back through the *same* json tags, so a wrong tag — e.g. `operations` instead of the SCIM-required capital `Operations` — would round-trip cleanly and the test would still pass. The map pins the exact wire keys (`Operations` / `schemas`), so it guards against regressing that casing fix. I left a comment above it explaining this; happy to reword if it's unclear, but I'd prefer to keep the wire-format assertion. WDYT?

Not-Dhananjay-Mishra (approved):

gmlewis (approved):
Thank you, @JamBalaya56562 and @Not-Dhananjay-Mishra!
LGTM.
Merging.

PR #4318, opened 2026-06-22
fix: Set `GetBody` on uploads for HTTP/2 retry

### LINKED ISSUE #2113: http2 failure in `UploadAsset`
We are using `ghr` to upload binary assets from Github to CI.

Currently I am seeing an increased failure rate:

```
Failed to upload one of assets: one of the goroutines failed: failed to upload asset: /tmp/jl_A641Uc/Clang_assert.v12.0.1.x86_64-linux-gnu-cxx11.tar.gz: failed to upload release asset: /tmp/jl_A641Uc/Clang_assert.v12.0.1.x86_64-linux-gnu-cxx11.tar.gz: Post https://uploads.github.com/repos/JuliaBinaryWrappers/Clang_assert_jll.jl/releases/49230189/assets?name=Clang_assert.v12.0.1.x86_64-linux-gnu-cxx11.tar.gz: http2: Transport: cannot retry err [stream error: stream ID 117; REFUSED_STREAM] after Request.Body was written; define Request.GetBody to avoid this error
```

AFAIK `ghr` uses `UploadAsset`: https://github.com/tcnksm/ghr/blob/93494744e6930d13e5c052eac9ca64a6027779a4/github.go#L184 from https://github.com/tcnksm/ghr/blob/93494744e6930d13e5c052eac9ca64a6027779a4/github.go#L209

and the error looks like https://github.com/golang/go/issues/25009

JamBalaya56562:
Fixes #2113

## Problem

`NewUploadRequest` passed the body to `http.NewRequestWithContext` without setting `Request.GetBody`. Unlike the JSON `NewRequest` path (where the Go stdlib auto-populates `GetBody` for `*bytes.Buffer`), the upload path wraps the reader in `uploadRequestBodyReader` to hide concrete types, so `GetBody` stayed nil. As a result `net/http2` could not rewind the body and asset uploads failed with:

```
http2: Transport: cannot retry err ... after Request.Body was written; define Request.GetBody to avoid this error
```

when a stream was refused (REFUSED_STREAM / GOAWAY / idle-connection close). This has been reported by several downstream tools (ghr, goreleaser, go-release-action).

## Fix

Set `req.GetBody` when the reader implements **both** `io.Seeker` (to record the body's starting offset) and `io.ReaderAt`. `GetBody` returns an independent `io.SectionReader` (re-wrapped in `uploadRequestBodyReader` to preserve the existing race-safety behaviour), so each call yields a fresh copy **without** disturbing the original body's read position — honoring the `http.Request.GetBody` contract.

This is satisfied by `*os.File` (used by `UploadReleaseAsset`), `*bytes.Reader` and `*strings.Reader`. Readers that cannot provide an independent, rewindable view are intentionally left without `GetBody` so large uploads are never buffered in memory.

This is a backwards-compatible change (no signature changes).

## Tests

- `TestNewUploadRequest_setsGetBodyForSeekableReader` — asserts `GetBody` is set, returns identical bytes on repeated calls, and does not consume the original `req.Body`.
- `TestNewUploadRequest_noGetBodyWithoutReaderAt` — asserts `GetBody` stays nil for a non-seekable reader and for a seek-only reader without `io.ReaderAt`.

`go test -race ./github/`, `go vet`, and golangci-lint all pass.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -937,6 +937,29 @@ func (c *Client) NewUploadRequest(ctx context.Context, urlStr string, reader io.
 
 	req.ContentLength = size
 
+	// Set GetBody so HTTP/2 transports can retry the request when a stream is
+	// refused (REFUSED_STREAM / GOAWAY) after the body has been written. Without
+	// GetBody, net/http2 fails with "cannot retry ... after Request.Body was
+	// written". See https://github.com/google/go-github/issues/2113.
+	//
+	// http.Request.GetBody must return an independent copy of the body, so we
+	// require io.ReaderAt (to read without moving the original body's position)
+	// in addition to io.Seeker (to record the body's starting offset). This is
+	// satisfied by *os.File, *bytes.Reader and *strings.Reader. Readers that
+	// cannot provide an independent, rewindable view are left without GetBody so
+	// that large uploads are never buffered in memory.
+	seeker, isSeeker := reader.(io.Seeker)
+	readerAt, isReaderAt := reader.(io.ReaderAt)
+	if isSeeker && isReaderAt {
+		offset, err := seeker.Seek(0, io.SeekCurrent)
+		if err != nil {
+			return nil, err
+		}
+		req.GetBody = func() (io.ReadCloser, error) {
+			return io.NopCloser(uploadRequestBodyReader{Reader: io.NewSectionReader(readerAt, offset, size)}), nil
+		}
+	}
+
 	if mediaType == "" {
 		mediaType = defaultMediaType
 	}
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -6,6 +6,7 @@
 package github
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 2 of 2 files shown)


github/github.go
```diff
@@ -937,6 +937,29 @@ func (c *Client) NewUploadRequest(ctx context.Context, urlStr string, reader io.
...
+	// in addition to io.Seeker (to record the body's starting offset). This is
+	// satisfied by *os.File, *bytes.Reader and *strings.Reader. Readers that
+	// cannot provide an independent, rewindable view are left without GetBody so
+	// that large uploads are never buffered in memory.
+	seeker, isSeeker := reader.(io.Seeker)
+	readerAt, isReaderAt := reader.(io.ReaderAt)
+	if isSeeker && isReaderAt {
+		offset, err := seeker.Seek(0, io.SeekCurrent)
+		if err != nil {
+			return nil, err
```
gmlewis, line 956: Can you please investigate if one of the test utils might help to increase the code coverage of this line in a unit test?

Image

gmlewis (changes requested):
Thank you, @JamBalaya56562!

github/github.go
JamBalaya56562: Good idea — `openTestFile` does the trick: closing the file before the call makes its `Seek(0, io.SeekCurrent)` fail, which exercises the offset-probe error branch (no new helper needed). Added `TestNewUploadRequest_returnsErrorWhenSeekFails` in 373b0b5; that line is covered now.

gmlewis (approved):
Excellent. Thank you, @JamBalaya56562!
LGTM.

stevehipwell (approved):
LGTM

gmlewis:
Thank you, @stevehipwell!
Merging.

PR #4202, opened 2026-05-08
docs: Extend CONTRIBUTING.md with code guidelines

jlaportebot:
## Summary

This PR extends CONTRIBUTING.md with a comprehensive section on code guidelines and conventions, addressing issue #4023.

## Changes

Added a new "Code Guidelines" section that documents:

1. **Naming Conventions**
   - File names ({service}_{api}.go pattern)
   - Receiver names (consistently use 's')
   - Request option types
   - Request body types
   - Response types
   - Method and variable naming
   - Common variable names (owner, repo, org, user, team, project)
   - Enterprise and organization scoped methods
   - Common structs (ListOptions, ListCursorOptions, UploadOptions, Response)

2. **JSON Tags for Request Bodies**
   - Required fields: non-pointer types without omitempty
   - Optional fields: pointer types with omitempty
   - Use omitzero for structs and slices

3. **URL Tags for Query Parameters**
   - All fields should be non-pointer types with url tags
   - Use omitempty to omit zero values

4. **Pagination**
   - Offset-based pagination (ListOptions)
   - Cursor-based pagination (ListCursorOptions)
   - Pagination best practices (embedding options, handling nil, using iterators)

5. **Common Types**
   - ID fields: always int64
   - Node ID fields: always string
   - Timestamp fields: use Timestamp type
   - Boolean fields: always *bool

6. **Generation Patterns**
   - Generated accessors
   - Generated iterators
   - Generated documentation
   - Linter rules (sliceofpointers, structfield)

## Benefits

- Helps new contributors understand the codebase better
- Reduces review burden by making expectations explicit
- Improves code consistency across the project
- Documents patterns that are frequently discussed in reviews

## Testing

This is a documentation-only change. No code changes or tests needed.

Closes #4023

PR #4319, opened 2026-06-22
feat: Add remaining Projects v2 endpoints

### LINKED ISSUE #3715: Add support for projects v2
REST API support for projects v2 has been [announced](https://github.blog/changelog/2025-09-11-a-rest-api-for-github-projects-sub-issues-improvements-and-more/).

JamBalaya56562:
Fixes #3715

## Summary

`ProjectsService` already covers projects (get/list), fields (get/list), and items (full CRUD). Comparing the current code against `openapi_operations.yaml`, eight Projects v2 REST operations were present in the spec but not yet implemented. This PR adds them, completing the Projects v2 REST surface:

| Group | Organization | User |
| --- | --- | --- |
| Create draft item | `CreateOrganizationProjectDraftItem` | `CreateUserProjectDraftItem` |
| Add field | `AddOrganizationProjectField` | `AddUserProjectField` |
| Create view | `CreateOrganizationProjectView` | `CreateUserProjectView` |
| List view items | `ListOrganizationProjectViewItems` | `ListUserProjectViewItems` |

## Notes

- **Reuse:** drafts and view items return `ProjectV2Item`; fields return `ProjectV2Field`; view items use the existing `ListProjectItemsOptions` (before/after cursor pagination). New types: `ProjectV2View` plus `CreateProjectV2DraftItemRequest`, `AddProjectV2FieldRequest`, `CreateProjectV2ViewRequest` (and the `single_select`/`iteration` field helper types). Request bodies are passed by value with a `Request` suffix, per the `paramcheck` convention.
- **User path quirk (mirrors the upstream spec):** the user-owned *draft* and *view* create endpoints identify the user by the numeric `user_id` in the path (`POST /user/{user_id}/projectsV2/{project_number}/drafts` and `POST /users/{user_id}/projectsV2/{project_number}/views`), unlike the `username`-based paths used by the other user endpoints. These methods therefore take a `userID int64`, consistent with `Users.GetByID`.
- No changes to `openapi_operations.yaml` (the operations were already present); generated accessors/iterators were regenerated via `script/generate.sh`.

## Testing

Added success + error-path tests for all eight methods (including a single_select field, an iteration field, and view-item pagination). `go test -race ./github/`, `go vet`, `go generate ./... --check`, and golangci-lint all pass.


github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+		return nil, resp, err
+	}
+
+	return field, resp, nil
+}
+
+// ProjectV2View represents a view in a project.
+type ProjectV2View struct {
+	ID         *int64  `json:"id,omitempty"`
+	Number     *int    `json:"number,omitempty"`
```
gmlewis: Should `Number` also be `*int64`?

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+type ProjectV2ViewSortBy struct {
+	// FieldID is the ID of the field to sort by.
+	FieldID *int64
+	// Direction is the sort direction, one of "asc" or "desc".
+	Direction *string
+}
+
+// UnmarshalJSON implements custom unmarshaling for the [field_id, direction]
+// tuple. The field_id may be encoded as either a number or a string.
+func (s *ProjectV2ViewSortBy) UnmarshalJSON(data []byte) error {
```
gmlewis, line 911: It should be pretty straightforward to get 100% code coverage on this method in a unit test

Image

.

gmlewis (changes requested):
Thank you, @JamBalaya56562!

github/projects.go
JamBalaya56562: I kept it as `*int`: GitHub's OpenAPI types both `projects-v2.number` and `projects-v2-view.number` as plain `integer` (no `format: int64`), so by the usual convention they map to `*int` — matching the existing `ProjectV2.Number`. The `*int64` fields here (`ID`, `FieldID`, `IssueFieldID`, `visible_fields`, …) are the ones the spec marks `format: int64`. Happy to switch to `*int64` if you'd rather keep it consistent with the IDs, though.

github/projects.go
JamBalaya56562: Good call — added subtests for the non-array, unsupported / non-numeric-string `field_id`, and non-string `direction` branches in 75c7c47. `ProjectV2ViewSortBy.UnmarshalJSON` (and `MarshalJSON`) are at 100% now.

gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
 
 	return s.client.Do(req, nil)
 }
+
+// CreateProjectV2DraftItemRequest specifies the parameters to create a draft item in a project.
+type CreateProjectV2DraftItemRequest struct {
+	// Title is the title of the draft issue item to create. (Required.)
+	Title *string `json:"title,omitempty"`
```
stevehipwell:
```suggestion
	Title string `json:"title"`
```
As title is required it shouldn't be a pointer.

stevehipwell (changes requested):
I think we need to fix the struct types to not use pointers for required fields. I've commented on the first one.

github/projects.go
JamBalaya56562: Done — thanks! Made the required request fields non-pointer values (and dropped `omitempty`): `CreateProjectV2DraftItemRequest.Title`, `CreateProjectV2ViewRequest.Name`/`Layout` (all in the OpenAPI `required` arrays), plus `ProjectV2FieldSingleSelectOption.Name` (an option's required display label). I kept `AddProjectV2FieldRequest` all-pointer since its body is a `oneOf` where fields are only conditionally required, and left the optional/response fields as-is. e5df91b

stevehipwell (changes requested):
There still appear to be fields which are required in the responses that are being defined as pointers (e.g. many of the view fields).

github/projects_test.go
alexandear, line 1: Please use `cmp.Diff`, `cmp.Equal`, `testJSONMarshal` to simplify tests.

JamBalaya56562:
@stevehipwell thanks. I've kept the response types pointer-based on purpose, to stay consistent with go-github's convention: response structs use `*T` + `omitempty` for every field, even ones that are always present. That's what the `Repository` example in CONTRIBUTING.md shows (`ID *int64`, `NodeID *string`, `Owner *User`), what the sibling `ProjectV2`/`ProjectV2DraftIssue`/`ProjectV2FieldOption` in this file do, and what `User`, `Repository`, etc. do throughout the package. The value-for-required convention applies to request bodies, which I've updated (`Title`, `Name`/`Layout`, single-select option `Name`). Making only `ProjectV2View` use value types would make it inconsistent with `ProjectV2` right beside it. Happy to switch if you'd prefer to change the response convention package-wide, but that feels like a separate change beyond this PR.

github/projects_test.go
JamBalaya56562: Good call — switched the new draft/field/view/view-item tests to `cmp.Diff(want, got)` against a fully-populated `want` (which now also checks every decoded field) in 365ebfb.

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
 
 	return s.client.Do(req, nil)
 }
+
+// CreateProjectV2DraftItemRequest specifies the parameters to create a draft item in a project.
+type CreateProjectV2DraftItemRequest struct {
+	// Title is the title of the draft issue item to create. (Required.)
```
alexandear:
```suggestion
	// Title is the title of the draft issue item to create.
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+	Title string `json:"title"`
+	// Body is the body content of the draft issue item to create. (Optional.)
```
alexandear:
```suggestion
	// Body is the body content of the draft issue item to create.
```

stevehipwell:
@JamBalaya56562 I think the current required pattern is to only use pointer types for optional fields. See #4202 (closed) where the consensus from the maintainer/reviewers was that required response body fields should be non-pointer types.

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return item, resp, nil
+}
+
+// ProjectV2FieldSingleSelectOption represents an option to create for a single_select project field.
+type ProjectV2FieldSingleSelectOption struct {
+	// Name is the display name of the option. (Required.)
```
alexandear:
```suggestion
	// Name is the display name of the option.
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+
+// UnmarshalJSON implements custom unmarshaling for the [field_id, direction]
+// tuple. The field_id may be encoded as either a number or a string.
+func (s *ProjectV2ViewSortBy) UnmarshalJSON(data []byte) error {
+	var tuple []json.RawMessage
+	if err := json.Unmarshal(data, &tuple); err != nil {
+		return err
+	}
+	if len(tuple) != 2 {
+		return fmt.Errorf("ProjectV2ViewSortBy: expected a [field_id, direction] tuple, got %v elements", len(tuple))
```
alexandear:
```suggestion
		return fmt.Errorf("expected a [field_id, direction] tuple, got %v elements", len(tuple))
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+	}
+
+	var fieldID int64
+	if err := json.Unmarshal(tuple[0], &fieldID); err != nil {
+		// The OpenAPI schema allows the field_id to be a string as well.
+		var str string
+		if err2 := json.Unmarshal(tuple[0], &str); err2 != nil {
+			return fmt.Errorf("ProjectV2ViewSortBy: invalid field_id: %w", err)
```
alexandear:
```suggestion
			return fmt.Errorf("invalid field_id: %w", err2)
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+		}
+		fieldID, err = strconv.ParseInt(str, 10, 64)
+		if err != nil {
+			return fmt.Errorf("ProjectV2ViewSortBy: invalid field_id %q: %w", str, err)
```
alexandear:
```suggestion
			return fmt.Errorf("invalid field_id %q: %w", str, err)
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+		}
+	}
+
+	var direction string
+	if err := json.Unmarshal(tuple[1], &direction); err != nil {
+		return fmt.Errorf("ProjectV2ViewSortBy: invalid direction: %w", err)
```
alexandear:
```suggestion
		return fmt.Errorf("invalid direction: %w", err)
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+// MarshalJSON implements custom marshaling to the [field_id, direction] tuple.
+func (s ProjectV2ViewSortBy) MarshalJSON() ([]byte, error) {
+	return json.Marshal([]any{s.FieldID, s.Direction})
+}
+
+// CreateProjectV2ViewRequest specifies the parameters to create a project view.
+type CreateProjectV2ViewRequest struct {
+	// Name is the view's display name. (Required.)
+	Name string `json:"name"`
+	// Layout is the view's layout. One of: table, board, roadmap. (Required.)
```
alexandear:
```suggestion
	// Layout is the view's layout. One of: table, board, roadmap.
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+}
+
+// MarshalJSON implements custom marshaling to the [field_id, direction] tuple.
+func (s ProjectV2ViewSortBy) MarshalJSON() ([]byte, error) {
+	return json.Marshal([]any{s.FieldID, s.Direction})
+}
+
+// CreateProjectV2ViewRequest specifies the parameters to create a project view.
+type CreateProjectV2ViewRequest struct {
+	// Name is the view's display name. (Required.)
```
alexandear:
```suggestion
	// Name is the view's display name.
```

JamBalaya56562:
Thanks @stevehipwell, and apologies for pushing back earlier — you're right. I checked #4202 and the consensus there is indeed that required response fields should be non-pointer (the merged CONTRIBUTING.md hasn't caught up yet, which is what threw me off). I've made `ProjectV2View`'s required scalar fields non-pointer (`id`, `number`, `name`, `layout`, `node_id`, `project_url`, `html_url`) in 5044ba0, keeping the optional `filter`, the nested `creator`, the timestamps and the slice fields as-is. Let me know if you'd like me to extend it to `creator`/timestamps too.

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+	UpdatedAt       *Timestamp `json:"updated_at,omitempty"`
+}
+
+// ProjectV2ViewSortBy represents a single sort criterion of a project view.
+//
+// On the wire it is encoded as a [field_id, direction] tuple rather than an
+// object, so it has custom JSON (un)marshaling.
+type ProjectV2ViewSortBy struct {
+	// FieldID is the ID of the field to sort by.
+	FieldID *int64
```
alexandear: Maybe it makes sense:
```suggestion
	FieldID *int64 `json:"-"`
```

github/projects.go
```diff
@@ -728,3 +729,331 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+type ProjectV2View struct {
+	ID         int64   `json:"id"`
+	Number     int     `json:"number"`
+	Name       string  `json:"name"`
+	Layout     string  `json:"layout"`
+	NodeID     string  `json:"node_id"`
+	ProjectURL string  `json:"project_url"`
+	HTMLURL    string  `json:"html_url"`
+	Creator    *User   `json:"creator,omitempty"`
+	Filter     *string `json:"filter,omitempty"`
```
alexandear: Why are these fields without comments and the rest with?

github/projects.go
JamBalaya56562: Done in bcc07a6.

github/projects.go
JamBalaya56562: Done in bcc07a6 — also switched to wrap `err2` (the actual string-parse error) here, thanks for catching that.

github/projects.go
JamBalaya56562: Done in bcc07a6.

github/projects.go
JamBalaya56562: Done in bcc07a6.

github/projects.go
JamBalaya56562: Done in bcc07a6.

github/projects.go
JamBalaya56562: Done in bcc07a6.

github/projects.go
JamBalaya56562: Done in 2aee530f.

github/projects.go
JamBalaya56562: Done in 2aee530f.

github/projects.go
JamBalaya56562: Done in 2aee530f.

github/projects.go
JamBalaya56562: Done — added `json:"-"` to `FieldID` and `Direction` since they're handled by the custom Marshal/UnmarshalJSON. 362c31a

github/projects.go
JamBalaya56562: Good point — I removed the per-field comments to make `ProjectV2View` consistent (matching the sibling `ProjectV2`, which has none). Happy to add a comment to every field instead if you'd prefer. 362c31a

github/projects.go
```diff
@@ -728,3 +729,327 @@ func (s *ProjectsService) DeleteUserProjectItem(ctx context.Context, username st
...
+	resp, err := s.client.Do(req, &field)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return field, resp, nil
+}
+
+// ProjectV2View represents a view in a project.
+type ProjectV2View struct {
```
Not-Dhananjay-Mishra, line 880:
```json
  "required": [
    "id",
    "number",
    "name",
    "layout",
    "node_id",
    "project_url",
    "html_url",
    "creator",
    "created_at",
    "updated_at",
    "visible_fields",
    "sort_by",
    "group_by",
    "vertical_group_by"
  ]
```
These fields are required we do not use `omitempty` and pointer with them.
Still some required field use `omitempty` like `group_by`, `vertical_group_by` etc. ([As said in comment](https://github.com/google/go-github/pull/4319#issuecomment-4779127818))

JamBalaya56562: Done — in `ProjectV2View` the remaining required fields (`creator`, `created_at`, `updated_at`, `visible_fields`, `sort_by`, `group_by`, `vertical_group_by`) are now non-pointer values without `omitempty`, matching the convention that only optional fields are pointers. The only field left as a pointer is `filter`, which is optional. Accessors were regenerated accordingly. 10bea9e

(The branch was also rebased onto the latest `master` to clear the merge conflict, hence the force-push.)

github/projects_test.go
```diff
@@ -1432,3 +1434,396 @@ func TestProjectV2ItemContent_Marshal(t *testing.T) {
...
+	testNewRequestAndDoFailure(t, methodName, client, func() (*Response, error) {
+		got, resp, err := client.Projects.ListUserProjectViewItems(ctx, "u", 1, 2, opts)
+		if got != nil {
+			t.Errorf("testNewRequestAndDoFailure %v = %#v, want nil", methodName, got)
+		}
+		return resp, err
+	})
+}
+
+func TestProjectV2ViewSortBy_UnmarshalJSON(t *testing.T) {
```
alexandear, line 1748: Please refactor this test in the same way as in #4323

JamBalaya56562: Done in ac5bf32 — the success case now uses `testJSONUnmarshalOnly` like #4323. The error-case subtests stay manual since the helper expects a successful unmarshal.

github/projects_test.go
```diff
@@ -1431,3 +1433,388 @@ func TestProjectV2ItemContent_Marshal(t *testing.T) {
...
+
+	t.Run("non-string direction is an error", func(t *testing.T) {
+		t.Parallel()
+		var got ProjectV2ViewSortBy
+		if err := json.Unmarshal([]byte(`[123,456]`), &got); err == nil {
+			t.Error("Unmarshal of a numeric direction = nil error, want error")
+		}
+	})
+
+	t.Run("round trips through MarshalJSON", func(t *testing.T) {
```
alexandear, line 1802: it's better to use `testJSONMarshal` here

JamBalaya56562: Done — switched the round-trip subtest to `testJSONMarshal`. Thanks!

alexandear (approved):

gmlewis:
Thank you, @alexandear!

@stevehipwell and @Not-Dhananjay-Mishra - have all your requests been addressed?

Not-Dhananjay-Mishra (approved):
LGTM

stevehipwell (commented):
Some of the pointer fields still look like they might be required despite the docs, but I've not had a chance to dig deeper into the schemas so I'm not objecting to this PR being merged.

gmlewis:
> Some of the pointer fields still look like they might be required despite the docs, but I've not had a chance to dig deeper into the schemas so I'm not objecting to this PR being merged.

What do you all think about having a linter that actually grabs the official schemas and checks whether each field is required or not and can make recommendations on its findings so that we match the official schemas as closely as possible?

Obviously, there have been times when we have had to explicitly override the official docs due to what was actually discovered "in the wild", but these could be the (hopefully few) exceptions (like our other tools accomodate exceptions that have the eventual goal of getting cleaned up over time).

gmlewis:
Whups, I meant to cc: @stevehipwell - @alexandear - @Not-Dhananjay-Mishra on that last question. ⬆️

stevehipwell:
@gmlewis I was thinking about the possibility of that too, I wasn't sure if it'd been tried before. I was also thinking about if it was possible to use the example payloads from the schema docs in our tests.

JamBalaya56562:
@stevehipwell good call — I went and dug into the OpenAPI schemas for the request bodies this PR adds. Here's what I found:

- **`POST .../drafts`** (`CreateProjectV2DraftItemRequest`): schema is `required: [title]` → `Title` is already a non-pointer `string`, and `body` is optional.
- **`POST .../views`** (`CreateProjectV2ViewRequest`): schema is `required: [name, layout]` → both `Name` and `Layout` are already non-pointer `string`; `filter`/`visible_fields` are optional.
- **`POST .../fields`** (`AddProjectV2FieldRequest`): this is the type whose fields are all pointers, and its schema is a `oneOf` of four variants:
  - `[issue_field_id]`
  - `[name, data_type]`
  - `[name, data_type, single_select_options]`
  - `[name, data_type, iteration_configuration]`

  So no field is unconditionally required — `name`/`data_type` aren't required in the `issue_field_id` variant — which is why they stay pointers, consistent with how we handle other conditional/`oneOf` bodies.

So the required scalars are already non-pointer, and the only remaining pointers are the conditionally-required `oneOf` fields. Happy to adjust if you read the schemas differently.

gmlewis:
OK, so @JamBalaya56562 - I think we are now good-to-go for this PR, right? Am I missing anything, or can we go ahead and merge this one now?

PR #4321, opened 2026-06-23
fix: Stop test/fields from reporting null-valued fields as missing

### LINKED ISSUE #576: tests/fields: False positive for fields with null values, reported as "missing" incorrectly.
There appears to be a bug in the `./tests/fields` test/utility, causing some fields to be falsely reported as "missing".

For example, in the following test case:

```Go
{"repos/google/go-github/issues/1", &github.Issue{}},
```

The real GitHub API response from https://api.github.com/repos/google/go-github/issues/1 includes:

```JSON
...
    "site_admin": false
    }
  ],
  "milestone": null,
  "comments": 2,
  "created_at": "2013-05-30T18:35:59Z",
  "updated_at": "2013-08-20T22:47:14Z",
...
```

Note that the value of `milestone` is `null`.

The corresponding struct field `Issue.Milestone` exists:

https://github.com/google/go-github/blob/d9c241b9c95cbd5996418f70223328a37cfb9bb3/github/issues.go#L42

But the output from the fields test incorrectly reports:

```
...
*github.Issue missing field for key: milestone (example value: <nil>)
...
```

I think it's because the null value gets dropped when JSON marshalling the `Issue` struct, because of `,omitempty` option.

JamBalaya56562:
## What problem are you solving?

Fixes #576. The `test/fields` developer utility falsely reported mapped struct fields as "missing" whenever the GitHub API returned `null` for them (e.g. `Issue.Milestone` for `"milestone": null`).

## Root cause

`testType` decoded the raw JSON into the struct and then **re-marshaled** the struct to build the set of known keys. Since every go-github field is a `,omitempty` pointer, a `null` value decodes to a nil pointer and is dropped on re-marshal, so the key was absent from the comparison set and incorrectly flagged as missing.

## Fix

Derive the expected top-level JSON keys directly from the struct type via reflection (`jsonFieldNames`), which is value-independent and therefore never drops `,omitempty` fields. The comparison is extracted into a `missingFields` function and covered by a network-free unit test (`fields_test.go`), including a regression test for the null-`milestone` case and a check that genuinely unmapped keys are still reported.

This keeps the tool's existing top-level scope; recursing into nested structs/slices is out of scope for this fix.

> Note: prepared with AI assistance (Claude Code); all changes were reviewed and verified locally (`go build`/`go test`/`go vet`/`gofmt` and the repo's custom `golangci-lint`).

### DIFF
--- a/test/fields/fields.go
+++ b/test/fields/fields.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"reflect"
+	"slices"
 	"strings"
 
 	"github.com/google/go-github/v88/github"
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.
Awaiting second LGTM+Approval from any other contributor before merging.

alexandear:
Looks like we're not using the `test/field` utility. I'm not sure whether we need to fix issues for it.

Currently, `test/field` checks only these structs:

```
		{"users/octocat", &github.User{}},
		{"user", &github.User{}},
		{"users/willnorris/keys", &[]github.Key{}},
		{"orgs/google-test", &github.Organization{}},
		{"repos/google/go-github", &github.Repository{}},
		{"repos/google/go-github/issues/1", &github.Issue{}},
		{"/gists/9257657", &github.Gist{}},
```

But we have many more structs.

The proper fix would be to extend the test cases and ensure that it runs regularly, or to simply delete the tool. @gmlewis what do you think?

stevehipwell:
> Since every go-github field is a ,omitempty pointer, a null value decodes to a nil pointer and is dropped on re-marshal, so the key was absent from the comparison set and incorrectly flagged as missing.

I'm not sure if it matters, but the original premise doesn't appear to be correct as not all fields are pointers.

gmlewis:
@alexandear - @stevehipwell - Honestly, I don't remember ever running this tool before.

I just ran it, and I get this:

```
*github.User missing field for key: bio (example value: <nil>)
*github.User missing field for key: hireable (example value: <nil>)
*github.User missing field for key: twitter_username (example value: <nil>)
*github.User missing field for key: email (example value: <nil>)
*github.User missing field for key: hireable (example value: <nil>)
*github.User missing field for key: notification_email (example value: <nil>)
*github.User missing field for key: company (example value: <nil>)
*github.User missing field for key: twitter_username (example value: <nil>)
*github.Organization missing field for key: location (example value: <nil>)
*github.Organization missing field for key: blog (example value: <nil>)
*github.Organization missing field for key: archived_at (example value: <nil>)
*github.Organization missing field for key: company (example value: <nil>)
*github.Organization missing field for key: twitter_username (example value: <nil>)
*github.Organization missing field for key: email (example value: <nil>)
*github.Repository missing field for key: forks (example value: 2244)
*github.Repository missing field for key: mirror_url (example value: <nil>)
*github.Repository missing field for key: temp_clone_token (example value: )
*github.Issue missing field for key: type (example value: <nil>)
*github.Issue missing field for key: issue_field_values (example value: [])
*github.Issue missing field for key: pinned_comment (example value: <nil>)
*github.Issue missing field for key: milestone (example value: <nil>)
*github.Issue missing field for key: performed_via_github_app (example value: <nil>)
*github.Issue missing field for key: timeline_url (example value: https://api.github.com/repos/google/go-github/issues/1/timeline)
*github.Issue missing field for key: active_lock_reason (example value: <nil>)
error: GET https://api.github.com/gists/9257657: 502 Server Error []
```

which makes me wonder.  I guess this is going to take a bit more research.

gmlewis:
Thank you, @JamBalaya56562, but we've decided to remove this obsolete tool, and therefore closed the issue and this PR.

PR #4322, opened 2026-06-23
chore: Remove test/fields utility

### LINKED ISSUE #576: tests/fields: False positive for fields with null values, reported as "missing" incorrectly.

alexandear:
The [`test/fields`](https://github.com/google/go-github/blob/728cdb1d018e0cc2f98850bc7fef24eee970d2cb/test/fields/fields.go#L17) utility was created 11 years ago in https://github.com/google/go-github/commit/3aaa6f118996781361f2fa6552b90e31bc953f1f. It has not been actively extended since, and it is currently unused. It has several problems:

- Covers only 10 API struct responses, while we have more than 100 structs
- Has false positives for fields with null values; see #576 
- Is not run regularly on CI
- Is not run by maintainers, reviewers, or contributors
- Is a simple app for manual use only, not a `testing`-compatible functions

So I propose we get rid of it. Instead, we should consider extending the integration tests with new cases and scenarios.

Closes #576
Closes #4321


gmlewis (approved):
OK, thanks, @alexandear.
I agree. LGTM.
Merging.

PR #4325, opened 2026-06-23
chore: Replace `time.Date` with `referenceTime`

### LINKED ISSUE #4324: Use `referenceTime` and `referenceTimeStr` throughout repo
See: https://github.com/google/go-github/pull/4323/files/5002e2ef1ddebf0d41402222891b9601adddf4de#diff-e31b0e93efc82d9c991c0bdc7e3f2f4d03d7a29a2de5d3956c937dbbc59a40b5

gmlewis:
Fixes: #4324.
Closes: #4326.


gmlewis:

stevehipwell (commented):
@gmlewis sorry for being dense, but I think I'm missing how the `+referenceTimeStr+` string works and why it needs to use a \` instead of `"`?

github/actions_workflow_runs_test.go
```diff
@@ -165,8 +164,8 @@ func TestActionsService_GetWorkflowRunAttempt(t *testing.T) {
 		ID:         Ptr(int64(399444496)),
 		RunNumber:  Ptr(296),
 		RunAttempt: Ptr(3),
-		CreatedAt:  &Timestamp{time.Date(2019, time.January, 2, 15, 4, 5, 0, time.UTC)},
-		UpdatedAt:  &Timestamp{time.Date(2020, time.January, 2, 15, 4, 5, 0, time.UTC)},
+		CreatedAt:  &referenceTimestamp,
+		UpdatedAt:  &referenceTimestamp,
```
alexandear: With this refactoring here, we lose field-mapping discrimination.

`CreatedAt` and `UpdatedAt` were distinct before the refactoring.

A decode bug that swapped `created_at`/`updated_at` would previously fail the test, but now the test asserts the two are equal, so the swap is invisible.

I propose reverting here, and for all structs that have two or more timestamps.

```suggestion
		CreatedAt:  &Timestamp{time.Date(2019, time.January, 2, 15, 4, 5, 0, time.UTC)},
		UpdatedAt:  &Timestamp{time.Date(2020, time.January, 2, 15, 4, 5, 0, time.UTC)},
```

github/git_commits_test.go
```diff
@@ -202,16 +201,15 @@ func TestGitService_CreateCommit_WithSigner(t *testing.T) {
 	client, mux, _ := setup(t)
 
 	signature := "my voice is my password"
-	date := time.Date(2017, time.May, 4, 0, 3, 43, 0, time.FixedZone("CEST", 2*3600))
```
alexandear, line 205: This date here is intentional. It checks `commit.Author.GetDate().Format("-0700")))` in the `createSignatureMessage`. Let's revert all changes to tests in the file.

gmlewis:
> @gmlewis sorry for being dense, but I think I'm missing how the `+referenceTimeStr+` string works and why it needs to use a ` instead of `"`?

Oh, it just keeps the double quote characters in the string.

github/rate_limit_test.go
```diff
@@ -43,18 +42,18 @@ func TestRateLimits(t *testing.T) {
...
-			"graphql": {"limit":4,"remaining":3,"used":1,"reset":1372700875},
-			"integration_manifest": {"limit":5,"remaining":4,"used":1,"reset":1372700876},
-			"source_import": {"limit":6,"remaining":5,"used":1,"reset":1372700877},
-			"code_scanning_upload": {"limit":7,"remaining":6,"used":1,"reset":1372700878},
-			"actions_runner_registration": {"limit":8,"remaining":7,"used":1,"reset":1372700879},
-			"scim": {"limit":9,"remaining":8,"used":1,"reset":1372700880},
-			"dependency_snapshots": {"limit":10,"remaining":9,"used":1,"reset":1372700881},
-			"code_search": {"limit":11,"remaining":10,"used":1,"reset":1372700882},
-			"audit_log": {"limit": 12,"remaining":11,"used":1,"reset":1372700883},
-			"dependency_sbom": {"limit": 100,"remaining":100,"used":0,"reset":1372700884}
```
alexandear, line 57: All these times are different on purpose. Let's revert.

github/actions_workflow_runs_test.go
gmlewis: Well, I thought about that and decided that this could only happen for old code if the labels were changed, so I wasn't concerned about it.

I thought that the readability gains were worth it. 

Hey, here's an idea: 

We could fix _all_ reported issues by adding helper functions that add an offset (seconds) to the reference time: 

```go
refTime(0)
refTimeStr(1)
etc...
```

Thoughts?

alexandear: Yeah, `refTime(0)` solves the regression in the PR.

But offsets in seconds are cryptic. `refTime(0)`, `refTime(1)` at a call site tells you nothing about why those two times differ. It's just "CreatedAt != UpdatedAt". The original `time.Date(2019,`  and `time.Date(2020,` was arguably more self-documenting. I prefer named distinct constants over magic offsets if possible.

gmlewis:
OK, I took a slightly different approach and used unix-seconds-since-epoch for all cases where different timestamps are valuable. I believe that this approach combines the best of both worlds.

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #4333, opened 2026-06-26
fix: AuditEntry fields `org` and `org_id` can be an array

### LINKED ISSUE #3488: Enterprise API responses are sometimes returning arrays of strings instead of a string

alexandear:
The problem is clearly described in #3488.
This PR is based on #4204, which lacks the author's CLA from May 9.

Fixes #3488
Closes #4204


gmlewis (changes requested):
Thank you, @alexandear!
You could probably easily add a unit test that sends `data = "{]"` to this function to get it to fail and improve the code coverage.

gmlewis:
codecov-4333

gmlewis (approved):
Thank you, @alexandear!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

PR #4335, opened 2026-06-26
feat!: Replace actions env secret endpoints

stevehipwell:
BREAKING CHANGE: `ActionsService` methods involving secrets have new params and return values.

This PR updates the actions environment secrets functions to use the documented and supported endpoints, it also changes all actions secret create/update functions to use a value parameter (#3644).

### DIFF
--- a/example/newreposecretwithxcrypto/main.go
+++ b/example/newreposecretwithxcrypto/main.go
@@ -143,21 +143,21 @@ func addRepoSecret(ctx context.Context, client *github.Client, owner, repo, secr
 	return nil
 }
 
-func encryptSecretWithPublicKey(publicKey *github.PublicKey, secretName, secretValue string) (*github.EncryptedSecret, error) {
+func encryptSecretWithPublicKey(publicKey *github.PublicKey, secretName, secretValue string) (github.EncryptedSecret, error) {
 	decodedPublicKey, err := base64.StdEncoding.DecodeString(publicKey.GetKey())
 	if err != nil {
-		return nil, fmt.Errorf("base64.StdEncoding.DecodeString was unable to decode public key: %v", err)
+		return github.EncryptedSecret{}, fmt.Errorf("base64.StdEncoding.DecodeString was unable to decode public key: %v", err)
 	}
 
 	boxKey := [32]byte(decodedPublicKey)
 	encryptedBytes, err := box.SealAnonymous([]byte{}, []byte(secretValue), &boxKey, crypto_rand.Reader)
 	if err != nil {
-		return nil, fmt.Errorf("box.SealAnonymous failed with error %w", err)
+		return github.EncryptedSecret{}, fmt.Errorf("box.SealAnonymous failed with error %w", err)
 	}
 
 	encryptedString := base64.StdEncoding.EncodeToString(encryptedBytes)
 	keyID := publicKey.GetKeyID()
-	encryptedSecret := &github.EncryptedSecret{
+	encryptedSecret := github.EncryptedSecret{
 		Name:           secretName,
 		KeyID:          keyID,
 		EncryptedValue: encryptedString,
--- a/example/newreposecretwithxcrypto/main.go
+++ b/example/newreposecretwithxcrypto/main.go
@@ -144,13 +144,12 @@ func addRepoSecret(ctx context.Context, client *github.Client, owner, repo, secr
 
 	encryptedString := base64.StdEncoding.EncodeToString(encryptedBytes)
 	keyID := publicKey.GetKeyID()
-	encryptedSecret := github.EncryptedSecret{
-		Name:           secretName,
+	req := github.SecretRequest{
 		KeyID:          keyID,
 		EncryptedValue: encryptedString,
 	}
 
-	if _, err := client.Actions.CreateOrUpdateRepoSecret(ctx, owner, repo, encryptedSecret); err != nil {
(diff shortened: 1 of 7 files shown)


example/newreposecretwithxcrypto/main.go
```diff
@@ -143,21 +143,21 @@ func addRepoSecret(ctx context.Context, client *github.Client, owner, repo, secr
 	return nil
 }
 
-func encryptSecretWithPublicKey(publicKey *github.PublicKey, secretName, secretValue string) (*github.EncryptedSecret, error) {
+func encryptSecretWithPublicKey(publicKey *github.PublicKey, secretName, secretValue string) (github.EncryptedSecret, error) {
```
gmlewis: What value is there to changing the return type from a pointer to a value?
It is pretty standard practice in Go to return `nil` when there is an error.
Why would we change that for this method?

stevehipwell: I'm not sure it makes any difference in this code if we're returning nil or an empty struct, AFAIK both could be considered idiomatic Go. I did this just before logging off for the weekend so I'm on my phone and can't see if this is a linter error? If it is I'll fix it. FYI I don't this example is particularly idiomatic Go given the unnecessary function separation.

gmlewis: Yeah, there are linter and test errors in this PR.

During code reviews at Google, I was told to not return a value struct during an error because a `nil` value was perfectly valid. So from my experience with the Go team, it is more idiomatic to return a `nil` pointer to a struct than to return an empty value struct when errors are possible.

stevehipwell: Thanks, I'll fix it up when I get a few mins. I'll probably just get rid of the function call as that'd remove the whole issue.

github/actions_secrets.go
```diff
@@ -246,9 +246,9 @@ func (s *ActionsService) putSecret(ctx context.Context, url string, body *Encryp
 // GitHub API docs: https://docs.github.com/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret
 //
 //meta:operation PUT /repos/{owner}/{repo}/actions/secrets/{secret_name}
-func (s *ActionsService) CreateOrUpdateRepoSecret(ctx context.Context, owner, repo string, body *EncryptedSecret) (*Response, error) {
-	if body == nil {
-		return nil, errors.New("encrypted secret must be provided")
+func (s *ActionsService) CreateOrUpdateRepoSecret(ctx context.Context, owner, repo string, body EncryptedSecret) (*Response, error) {
+	if body.Name == "" {
+		return nil, errors.New("secret name must be provided")
```
alexandear: We usually don't add such checks to other endpoint methods. 

The field `Name` is without `omitempty`, which means "required".

stevehipwell: It's used to construct the URL, which is why I think the previous signature had nil checks (although an uninitialized struct would still cause issues).

alexandear: Now I understand. In that way, we should add a new parameter.

github/actions_secrets.go
```diff
@@ -246,9 +246,9 @@ func (s *ActionsService) putSecret(ctx context.Context, url string, body *Encryp
 // GitHub API docs: https://docs.github.com/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret
 //
 //meta:operation PUT /repos/{owner}/{repo}/actions/secrets/{secret_name}
-func (s *ActionsService) CreateOrUpdateRepoSecret(ctx context.Context, owner, repo string, body *EncryptedSecret) (*Response, error) {
-	if body == nil {
-		return nil, errors.New("encrypted secret must be provided")
+func (s *ActionsService) CreateOrUpdateRepoSecret(ctx context.Context, owner, repo string, body EncryptedSecret) (*Response, error) {
```
alexandear:
```suggestion
func (s *ActionsService) CreateOrUpdateRepoSecret(ctx context.Context, owner, repo, secretName string, body CreateOrUpdateRepoSecretRequest) (*Response, error) {
```

```
type CreateOrUpdateRepoSecretRequest struct {
	KeyID                 string          `json:"key_id"`
	EncryptedValue        string          `json:"encrypted_value"`
}
```

Because https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-a-repository-secret

Image

alexandear: The similar for other endpoints

github/actions_secrets.go
stevehipwell: A new parameter or a new request type (as per below)?

github/actions_secrets.go
stevehipwell: Yeah, I'd considered doing this but didn't want to shave the yak, but I think that boat has sailed now. How about `SecretRequest` (for repo and env) and `OrgSecretRequest`; I'd like to think that this endpoint will eventually be replaced with separate endpoints for create and update. The Dependabot endpoints could use the same types, although I'm not sure we want to make that change in this PR?

alexandear: Currently, these `ActionsService` endpoints are completely broken. We can fix them whenever needed. And adding new structs won't break Dependabot endpoints.

stevehipwell: @alexandear this should be fixed now.

alexandear (approved):

gmlewis (approved):
LGTM.
Merging.

stevehipwell:
@gmlewis variables PR to follow.

PR #4338, opened 2026-06-27
build: Exclude generated files from golangci-lint

JamBalaya56562:
Excludes the large auto-generated files from golangci-lint's linters and formatters in `.golangci.yml`:

- `github/github-accessors.go` / `github/github-accessors_test.go`
- `github/github-iterators.go` / `github/github-iterators_test.go`
- `github/github-stringify_test.go`

They are added to `linters.exclusions.paths` and `formatters.exclusions.paths`. All carry the `// Code generated ... DO NOT EDIT.` header and are covered by `check-generated`, `go build`, and `go test`.

This did not measurably change lint/format wall-clock time:

| | linters (`run`) | formatters (`fmt`) |
| --- | --- | --- |
| before | ~15–17 s | ~6–7 s |
| after | ~16–17 s | ~6–7 s |

golangci-lint loads and type-checks whole packages, so the excluded files are still parsed; `exclusions.paths` only suppresses their reported issues/formatting (`lax` mode already suppresses generated-file issues for the linters, so the net-new effect is on the formatters).

Re #4070

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -557,6 +557,12 @@ linters:
       # Because fmt.Sprint(reset.Unix())) is more readable than strconv.FormatInt(reset.Unix(), 10).
       - linters: [perfsprint]
         text: fmt.Sprint.* can be replaced with faster strconv.FormatInt
+
+    # Exclude the large auto-generated files (script/generate.sh output) from linting.
+    paths:
+      - github/github-accessors.*\.go$
+      - github/github-iterators.*\.go$
+      - github/github-stringify_test\.go$
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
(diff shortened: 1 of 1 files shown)


gmlewis:
@JamBalaya56562 - according to the description above, this PR makes zero difference.

And in fact, for developers who actually are tweaking the `tools` in this repo, this PR would actually PREVENT them from seeing linter warnings for the newly-modified files that they are working on.

So I'm thinking that this PR is NOT desireable and should be closed.

Or am I missing something?

gmlewis:
Also, it certainly does NOT fix #4070.

JamBalaya56562:
You're right on both counts — thanks.

The numbers show no wall-clock improvement, and more importantly your second point is valid: excluding the generated files would hide linter warnings on the regenerated output for anyone working on the `gen-*.go` generators. That's a real downside, so this isn't worth doing. Closing.

Agreed that it doesn't fix #4070 either — I've dropped that keyword.

For context on #4070: this generated-file exclusion was the last open item there (the performance-relevant work — custom-gcl caching #4116/#4119, formatter/linter cleanup #4094/#4097 — is already merged, and the revive/gci changes were declined). Since this last item turns out to be undesirable rather than just neutral, I think #4070 itself can be closed, but I'll leave that to you.

PR #4360, opened 2026-07-03
refactor!: Pass `HookConfig` by value and rename `EditHookConfiguration` to `UpdateHookConfiguration`

JamBalaya56562:
Part of #3644.

Two related changes to the webhook-configuration methods:

**1. Pass `HookConfig` by value.** `HookConfig` is shared as the PATCH request body by three methods, so `paramcheck` flags them together and they must be converted in a single PR:

- `RepositoriesService.UpdateHookConfiguration`
- `OrganizationsService.UpdateHookConfiguration`
- `AppsService.UpdateHookConfig`

Each method's `body` parameter changes from `*HookConfig` to `HookConfig`, and `HookConfig` is removed from the `paramcheck` `body-allowed-pointer-types` allow-list in `.golangci.yml`.

**2. Rename `EditHookConfiguration` to `UpdateHookConfiguration`** (on `RepositoriesService` and `OrganizationsService`). The GitHub API docs name this operation "Update a webhook configuration for a repository/organization", and `AppsService.UpdateHookConfig` already uses the "Update" verb. This is consistent with the method naming used elsewhere in the #3644 refactor (e.g. `EditRelease` → `UpdateRelease`).

The `HookConfig` struct is unchanged (all four fields are optional), so no generated accessors change. Cross-checked against the GitHub webhook-config schema — its only fields are `url`, `content_type`, `secret`, `insecure_ssl`, all already present.

BREAKING CHANGE: `EditHookConfiguration` is renamed to `UpdateHookConfiguration` on `RepositoriesService` and `OrganizationsService`; these methods and `AppsService.UpdateHookConfig` now take `HookConfig` by value instead of by pointer.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -229,7 +229,6 @@ linters:
             - EnterpriseSecurityAnalysisSettings
             - ExternalGroup
             - Hook
-            - HookConfig
             - ImpersonateUserOptions
             - Import
             - InstallationTokenListRepoOptions
--- a/github/apps_hooks.go
+++ b/github/apps_hooks.go
@@ -36,7 +36,7 @@ func (s *AppsService) GetHookConfig(ctx context.Context) (*HookConfig, *Response
 // GitHub API docs: https://docs.github.com/rest/apps/webhooks?apiVersion=2022-11-28#update-a-webhook-configuration-for-an-app
 //
 //meta:operation PATCH /app/hook/config
-func (s *AppsService) UpdateHookConfig(ctx context.Context, body *HookConfig) (*HookConfig, *Response, error) {
+func (s *AppsService) UpdateHookConfig(ctx context.Context, body HookConfig) (*HookConfig, *Response, error) {
 	req, err := s.client.NewRequest(ctx, "PATCH", "app/hook/config", body)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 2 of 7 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

gmlewis:
@JamBalaya56562 - I really appreciate your efforts to make my job easier - thank you! When you write the "BREAKING CHANGE: ..." string in the description above, it would be awesome if you could make that a single, succint, stand-alone sentence, as that will get copied verbatim to this repo's git history (see `git log`), which will in-turn get copied to the release notes for the next release.

So, for example, you wrote "of the three methods above" which I cannot use as a single, stand-alone sentence as it is currently written.  I also like to keep them as short and succint as possible because nobody likes to read huge amounts of fluffy text, especially when trying to answer the question "I just upgraded this repo - how do I fix all the breaking changes in my own code?!?!?"

If you could please tweak that one line in the description to make it summarize the changes in a single short sentence, that would be fantastic... thank you!

JamBalaya56562:
Thanks, @gmlewis — that makes complete sense, and I'll keep it in mind going forward. Updated the description so the `BREAKING CHANGE:` line is a single, stand-alone sentence:

> BREAKING CHANGE: `EditHookConfiguration` is renamed to `UpdateHookConfiguration` on `RepositoriesService` and `OrganizationsService`; these methods and `AppsService.UpdateHookConfig` now take `HookConfig` by value instead of by pointer.

Happy to trim it further if you'd like it shorter.

gmlewis:
@JamBalaya56562 - that looks great... thank you!

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #4361, opened 2026-07-04
refactor!: Pass `DeploymentRequest` and `DeploymentStatusRequest` by value

JamBalaya56562:
This is a follow-up to #3644, converting `repos_deployments.go` to the established value-parameter pattern:

- `CreateDeployment` and `CreateDeploymentStatus` now take their request bodies (`DeploymentRequest`, `DeploymentStatusRequest`) by value instead of by pointer.
- The schema-required fields are now non-pointers with `omitempty` removed: `DeploymentRequest.Ref` (`ref` is required by the create-a-deployment request schema) and `DeploymentStatusRequest.State` (`state` is required by the create-a-deployment-status request schema).
- Both types are removed from the paramcheck allowlist in `.golangci.yml` (`custom-gcl run ./...` reports 0 issues).
- A second commit adds the missing optional `TargetURL` field to `DeploymentStatusRequest` (`target_url` is present in the OpenAPI request schema; the API docs recommend `log_url`, which replaces it).

Both method names already match the GitHub docs operation names ("Create a deployment", "Create a deployment status"), so no renames are needed.

Updates #3644.

BREAKING CHANGE: `CreateDeployment` and `CreateDeploymentStatus` now take `DeploymentRequest` and `DeploymentStatusRequest` by value instead of by pointer; the required fields `DeploymentRequest.Ref` and `DeploymentStatusRequest.State` are now `string` instead of `*string`, and `DeploymentRequest.RequiredContexts` is now `[]string` instead of `*[]string`.

### DIFF
--- a/github/repos_deployments.go
+++ b/github/repos_deployments.go
@@ -32,15 +32,17 @@ type Deployment struct {
 
 // DeploymentRequest represents a deployment request.
 type DeploymentRequest struct {
-	Ref                   string    `json:"ref"`
-	Task                  *string   `json:"task,omitempty"`
-	AutoMerge             *bool     `json:"auto_merge,omitempty"`
-	RequiredContexts      *[]string `json:"required_contexts,omitempty"`
-	Payload               any       `json:"payload,omitempty"`
-	Environment           *string   `json:"environment,omitempty"`
-	Description           *string   `json:"description,omitempty"`
-	TransientEnvironment  *bool     `json:"transient_environment,omitempty"`
-	ProductionEnvironment *bool     `json:"production_environment,omitempty"`
+	Ref       string  `json:"ref"`
+	Task      *string `json:"task,omitempty"`
+	AutoMerge *bool   `json:"auto_merge,omitempty"`
+	// RequiredContexts is the status contexts to verify against commit status checks.
+	// If nil, all unique contexts are verified; an empty slice bypasses checking entirely.
+	RequiredContexts      []string `json:"required_contexts,omitzero"`
+	Payload               any      `json:"payload,omitempty"`
+	Environment           *string  `json:"environment,omitempty"`
+	Description           *string  `json:"description,omitempty"`
+	TransientEnvironment  *bool    `json:"transient_environment,omitempty"`
+	ProductionEnvironment *bool    `json:"production_environment,omitempty"`
 }
 
 // DeploymentsListOptions specifies the optional parameters to the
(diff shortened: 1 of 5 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/repos_deployments.go
```diff
@@ -32,7 +32,7 @@ type Deployment struct {
 
 // DeploymentRequest represents a deployment request.
 type DeploymentRequest struct {
-	Ref                   *string   `json:"ref,omitempty"`
+	Ref                   string    `json:"ref"`
 	Task                  *string   `json:"task,omitempty"`
 	AutoMerge             *bool     `json:"auto_merge,omitempty"`
 	RequiredContexts      *[]string `json:"required_contexts,omitempty"`
```
alexandear: Please also fix this:
```suggestion
	RequiredContexts      []string `json:"required_contexts,omitempty"`
```

And do not forget to remove the line from `.golangci.yml`:

```yaml
            - DeploymentRequest.RequiredContexts # TODO: Deployments
```

github/repos_deployments.go
```diff
@@ -239,7 +242,7 @@ func (s *RepositoriesService) GetDeploymentStatus(ctx context.Context, owner, re
 // GitHub API docs: https://docs.github.com/rest/deployments/statuses?apiVersion=2022-11-28#create-a-deployment-status
 //
 //meta:operation POST /repos/{owner}/{repo}/deployments/{deployment_id}/statuses
-func (s *RepositoriesService) CreateDeploymentStatus(ctx context.Context, owner, repo string, deployment int64, body *DeploymentStatusRequest) (*DeploymentStatus, *Response, error) {
+func (s *RepositoriesService) CreateDeploymentStatus(ctx context.Context, owner, repo string, deployment int64, body DeploymentStatusRequest) (*DeploymentStatus, *Response, error) {
```
alexandear:
```suggestion
func (s *RepositoriesService) CreateDeploymentStatus(ctx context.Context, owner, repo string, deploymentID int64, body DeploymentStatusRequest) (*DeploymentStatus, *Response, error) {
```

github/repos_deployments.go
JamBalaya56562: Done in ff556397 — changed `RequiredContexts` to a plain `[]string` and removed the structfield exception from `.golangci.yml`.

One deviation from the suggestion: I used `omitzero` instead of `omitempty`, because an explicit empty array is meaningful for this endpoint — the API docs say "To bypass checking entirely, pass an empty array." With `omitempty` an empty slice could never be serialized, whereas `omitzero` keeps `nil` = omitted and `[]string{}` = sent as `[]` (the same pattern as `SelectedRepositoryIDs`/`SelectedUsernames` elsewhere in the repo). I also added a short doc comment explaining the nil vs. empty semantics. Happy to switch to `omitempty` if you prefer.

github/repos_deployments.go
JamBalaya56562: Done in ff556397, thank you!

github/repos_deployments.go
```diff
@@ -32,15 +32,17 @@ type Deployment struct {
...
-	Payload               any       `json:"payload,omitempty"`
-	Environment           *string   `json:"environment,omitempty"`
-	Description           *string   `json:"description,omitempty"`
-	TransientEnvironment  *bool     `json:"transient_environment,omitempty"`
-	ProductionEnvironment *bool     `json:"production_environment,omitempty"`
+	Ref       string  `json:"ref"`
+	Task      *string `json:"task,omitempty"`
+	AutoMerge *bool   `json:"auto_merge,omitempty"`
+	// RequiredContexts is the status contexts to verify against commit status checks.
+	// If nil, all unique contexts are verified; an empty slice bypasses checking entirely.
```
alexandear: Is there a reason you added the comment only for that field, and not for the rest of the struct?

JamBalaya56562: Good point — no real reason. I removed the comment in afa76ce5 to keep the struct consistent with its other undocumented fields.

github/repos_deployments.go
```diff
@@ -170,9 +170,12 @@ type DeploymentStatus struct {
 	URL            *string    `json:"url,omitempty"`
 }
 
-// DeploymentStatusRequest represents a deployment request.
+// DeploymentStatusRequest represents a deployment status request.
 type DeploymentStatusRequest struct {
-	State          *string `json:"state,omitempty"`
+	State string `json:"state"`
+	// TargetURL is the target URL to associate with this status.
+	// It's recommended to use LogURL instead, which replaces TargetURL.
```
alexandear, line 177: Same as https://github.com/google/go-github/pull/4361#discussion_r3524914469

alexandear (approved):

PR #4363, opened 2026-07-04
fix: Scope token auth to configured origins

### LINKED ISSUE #4366: WithAuthToken authorizes requests outside configured hosts
## Summary

`WithAuthToken` installs a transport wrapper that adds `Authorization: Bearer <token>` to outgoing requests. When callers build a request with an absolute URL outside the client's configured API or upload origins, that wrapper still adds the bearer token before sending the request.

## Expected behavior

The token configured through `WithAuthToken` should only be attached to requests targeting the client's configured API or upload origins. Requests to other origins should be sent without this transport-managed authorization header.

## Reproduction

This can be reproduced with `httptest` servers by configuring a client with trusted API/upload URLs, then issuing requests using absolute URLs for another server and checking whether that server receives an `Authorization` header.

## Proposed fix

PR #4363 limits token injection to the configured API and upload origins, and includes regression coverage for trusted API/upload requests and cross-host absolute URL requests.

huynhtrungcsc:
## Summary

- Send WithAuthToken credentials only to configured API and upload origins.
- Apply the same origin policy to redirects and cloned clients.
- Add regression coverage for absolute URLs, redirect chains, default ports, and Clone.

Fixes #4366


alexandear (commented):
Could you create an issue first with the problem you're trying to resolve, and once we've confirmed it, we can proceed with a review?

huynhtrungcsc:
Thanks for the guidance. I opened an issue describing the behavior this PR addresses:

- #4366

I also added focused coverage for the host-boundary helper and the configured-host authorization behavior. I’m happy to adjust the PR based on the maintainers’ preferred direction.

gmlewis (approved):
Thank you, @huynhtrungcsc.

I understand that this is a hypothetical situation and has not been reported as an actual problem "in the wild" but I also understand that with the increased proliferation of "vibe coding" and not properly understanding how to use packages, that this kind of change will help reduce the possibility of users of this package from shooting themselves in the foot.

LGTM.

alexandear:
@huynhtrungcsc please update the PR's description with "Fixes #4366" and remove the unnecessary "Tests" section.

huynhtrungcsc:
Updated the PR description to include "Fixes #4366" and removed the Tests section. Thank you.

alexandear (commented):
It looks like this PR was created with AI assistance, so I reviewed it with AI as well. Here's what I found:

1. github/github.go:638 — Clone() silently loses authentication (CONFIRMED, reproduced). The token closure captures c and checks the original client's URLs; Clone at github.go:783 copies the wrapped transport without re-binding it. NewClient(WithAuthToken(...)).Clone(WithEnterpriseURLs(...)) — a pattern that worked before this change — now sends every request unauthenticated. This is the must-fix.

2. github/github.go:723 — no default-port normalization (CONFIRMED). https://ghe.corp:443/ in config vs https://ghe.corp/... in a request URL (same RFC 6454 origin) fails the match and drops the token deterministically.

3. github/github.go:718 — three divergent boundary predicates (CONFIRMED). sameOrigin (scheme+host, case-insensitive) vs bareDoUntilFound:1391 and checkRedirectHost:2182 (host-only, case-sensitive). Concrete effect: a same-host https→http redirect is followed but not authenticated — no data, no error. Consolidate into one shared predicate.

4. github/repos_contents.go:200 — DownloadContents on private repos (PLAUSIBLE). The download_url fetch (raw.githubusercontent.com / GHES raw host) goes through the token transport and now loses the header; usually rescued by the embedded ?token=, but worth a deliberate decision and a test whose fake download_url is cross-origin.

5. github/repos_releases.go:328 — DownloadReleaseAsset + client.Client() (PLAUSIBLE). The documented private-repo pattern no longer authenticates cross-host redirect targets. A fix for github.com's pre-signed S3 URLs, a possible regression for GHES storage topologies.

6. github/github.go:258 — stale docs/comments (CONFIRMED). Client(), WithAuthToken (:436), and the bareDoUntilFound rationale comment (:1389) all still describe unconditional token attachment.

7. github/github.go:636 — wasted work in the hot path (cleanup). Early-exit before req.Clone, and hoist the "Bearer "+token string out of the closure.

8. github/github_test.go:714 — redundant tests / dead nil checks (cleanup). Fold TestSameOrigin's two unique cases into TestClientShouldAuthorizeRequest and drop the unreachable nil branches.

Findings 1–3 all point at one design question worth settling before merging: should the closure capture resolved URL values, or should the origin check live at a client level (e.g. in BareDo) where the correct client — including clones — is always in hand? Fixing it there resolves the Clone bug and gives the redirect guards a single predicate to share.

huynhtrungcsc:
Thanks for the detailed review. I addressed the confirmed regression and consistency issues in the latest push:

- Fixed `Clone()` by preserving the `WithAuthToken` token in the client config and re-wrapping the clone's transport against the clone's configured API/upload URLs, instead of reusing the original client's auth wrapper.
- Normalized origin matching with a shared `sameOrigin` helper, including default port handling (`https://host` == `https://host:443`).
- Updated the redirect guards to use the same origin predicate and reject cross-origin redirects consistently.
- Updated the `Client()` / `WithAuthToken` comments and cleaned up the hot path by avoiding request cloning unless the token is actually attached.
- Added regression coverage for the clone behavior and default-port origin matching, and simplified the helper tests.

I left the DownloadContents / DownloadReleaseAsset cases unchanged in this PR because those are broader product-behavior decisions around cross-origin GitHub/GHES download URLs. I’m happy to follow the maintainers’ preferred direction there if you want this PR to cover those paths too.

Verified locally with:

```bash
go test ./github -run "TestWithAuthToken|TestShouldAuthorizeURL|TestBareDoUntilFound_RejectsCrossOriginRedirect|TestRoundTripWithOptionalFollowRedirect_RejectsCrossOriginRedirect" -count=1
go test ./... -count=1
```

gmlewis (changes requested):
Same comment here about your code coverage dropping. Please fix.

huynhtrungcsc:
Thanks, I added coverage for the remaining helper branches so the new code paths are exercised directly, including nil URL/configured origins, default HTTP port handling, and non-HTTP schemes.

Verified locally with:

```bash
go test ./github -run "TestShouldAuthorizeURL|TestWithAuthTokenCloneAuthorizesCloneHosts" -count=1
go test ./... -count=1
```

github/github.go
```diff
@@ -647,6 +631,33 @@ func newClient(opts clientOptions) (*Client, error) {
 		c.uploadURL, _ = url.Parse(uploadBaseURL)
 	}
 
+	c.transport = c.client.Transport
```
stevehipwell: Wy has this code been moved? It makes it much harder to review the actual changes?

stevehipwell (changes requested):
I don't think we should be making these changes without actually discussing the issue as there are other related issues that should be looked at together with this. When I rewrote the client auth code I did so to keep it aligned with the existing behaviours, if we're looking to change that I'd like to make sure we carefully consider how to do it.

huynhtrungcsc:
Thanks, that makes sense. I agree this should not be changed piecemeal if there are related auth, redirect, and download-url behaviors that need to be considered together.

I will stop pushing implementation changes on this PR for now and use #4366 as the place to settle the expected behavior first. Once there is maintainer consensus on the boundary this should enforce, I can either adjust this PR to match that direction or close it if a different approach is preferred.

JamBalaya56562:
No action needed while this is paused — just noting one inconsistency that looks wrong regardless of where #4366 lands:

`shouldAuthorizeRequest` allows `baseURL` or `uploadURL`, but the redirect guards compare only against `baseURL`: `sameOrigin(newURL, c.baseURL)` in `bareDoUntilFound` and `sameOrigin(target, c.baseURL)` in `checkRedirectHost`. A hop that stays within the configured upload origin is still treated as cross-host. If the allowed-origin set is base + upload, those two want the same set.

huynhtrungcsc:
Updated the implementation to use the same configured-origin set for token injection and redirect handling. Clone now rebinds authentication to the clone's API and upload URLs. Added coverage for direct absolute URLs, same-origin redirects, cross-origin redirect chains, default ports, and clone behavior.

Validated with script/fmt.sh, script/test.sh, and script/lint.sh.

gmlewis (approved):

stevehipwell (changes requested):
@gmlewis as per my previous comment I don't think we should take this PR without a proper discussion.

gmlewis:
> @gmlewis as per my previous comment I don't think we should take this PR without a proper discussion.

OK, for posterity, this discussion is happening in #4366.

gmlewis:
Thanks for your patience, @huynhtrungcsc — I've settled the direction in #4366 (https://github.com/google/go-github/issues/4366#issuecomment-5717105620) and will implement it in one new PR that closes this one, carrying your `Clone` re-binding and default-port fixes and crediting you as co-author so none of your work is lost.

PR #4364, opened 2026-07-04
fix!: Scope auth transports to allowed origins

### LINKED ISSUE #4365: Auth transports add credentials after cross-origin redirects
## Summary

`BasicAuthTransport` and `UnauthenticatedRateLimitedTransport` add credentials in `RoundTrip` for each outgoing request. When Go's `net/http` client follows a redirect to a different origin, the redirected request can pass through the same transport and receive those credentials again.

## Expected behavior

Credentials added by these transports should only be sent to the original origin. Redirected requests that cross origin should continue without these transport-managed credentials.

## Reproduction

This can be reproduced with two `httptest` servers: one server returns a redirect to a second server, and the second server records whether the redirected request contains the transport-added authorization headers.

## Proposed fix

PR #4364 avoids adding these credentials on cross-origin redirect requests and includes regression coverage for both affected transports.

huynhtrungcsc:
## Summary

- Scope BasicAuthTransport and UnauthenticatedRateLimitedTransport credentials to AllowedOrigins.
- Default the allowlist to the public GitHub API and upload origins.
- Suppress Authorization and X-GitHub-OTP outside the allowlist.
- Cover direct absolute URLs and multi-hop redirect chains.

Fixes #4365


alexandear (commented):
Could you create an issue first with the problem you're trying to resolve, and once we've confirmed it, we can proceed with a review?

huynhtrungcsc:
Thanks for the guidance. I opened an issue describing the behavior this PR addresses:

- #4365

I also added focused coverage for the redirect-origin helper so the new behavior is covered more directly. I’m happy to adjust the PR based on the maintainers’ preferred direction.

gmlewis (approved):
Thank you, @huynhtrungcsc!
LGTM with the same comments I made in #4363.

github/github.go
```diff
@@ -1981,6 +1981,20 @@ func setCredentialsAsHeaders(req *http.Request, id, secret string) *http.Request
...
+func isCrossOriginRedirect(req *http.Request) bool {
+	if req == nil || req.URL == nil || req.Response == nil || req.Response.Request == nil || req.Response.Request.URL == nil {
+		return false
+	}
+	return !sameRedirectOrigin(req.URL, req.Response.Request.URL)
+}
+
+func sameRedirectOrigin(a, b *url.URL) bool {
+	if a == nil || b == nil {
+		return false
```
gmlewis, line 2003: For complete Codecov results, please add a simple unit test that exercises this code path.
Thank you, @huynhtrungcsc!

gmlewis (changes requested):

github/github.go
huynhtrungcsc: Added direct unit coverage for sameRedirectOrigin, including nil, same-origin, case-insensitive, and different-origin cases. Thank you.

gmlewis (approved):

gmlewis (changes requested):
Since you added `defaultPort` (is this function really necessary?!?), your code coverage went down, so if you want to keep it, it needs its own unit test.

huynhtrungcsc:
Added direct unit coverage for `defaultPort` in c5be7ac.

Local verification:

- `go test ./github -run "TestIsCrossOriginRedirect|TestSameRedirectOrigin|TestDefaultPort|TestUnauthenticatedRateLimitedTransport_doesNotAuthorizeCrossOriginRedirect|TestBasicAuthTransport_doesNotAuthorizeCrossOriginRedirect" -count=1`
- `go test ./github -coverprofile=/tmp/go-github-4364.cover -count=1` now reports 100% coverage for `isCrossOriginRedirect`, `sameRedirectOrigin`, and `defaultPort`.

stevehipwell (changes requested):
I have the same issue with this as I do with #4363, this need further discussion around the actual context rather than a rushed PR.

huynhtrungcsc:
Understood. I agree this belongs in the same broader discussion as #4363 before changing transport behavior.

I will pause further implementation changes here as well and use #4365 to clarify the expected redirect/auth boundary first. Once that direction is settled, I can update this PR accordingly or close it if maintainers prefer a different approach.

JamBalaya56562:
No action needed while this is paused — just noting the part I think has to change once #4366 settles:

`isCrossOriginRedirect` compares `req.URL` against `req.Response.Request.URL` and returns false when `req.Response` is nil, so the guard only fires on a redirect. A direct absolute-URL request through `BasicAuthTransport` or `UnauthenticatedRateLimitedTransport` still gets credentials attached — the same leak as #4366, via a different transport.

Making it origin-based rather than hop-based needs the configured origins, which neither transport can see (both are user-constructed via `github.NewClient(tp.Client())` and neither struct has a URL field). That likely means an explicit allowed-origins field on both — new public API, so possibly its own PR rather than a change here.

huynhtrungcsc:
Reworked the transport guard to check every request against a fixed origin allowlist instead of comparing redirect hops. This covers direct absolute URLs and multi-hop redirect chains, and suppresses Basic auth, OTP, and OAuth app credentials outside AllowedOrigins.

Validated with script/fmt.sh, script/test.sh, and script/lint.sh.

gmlewis (approved):

stevehipwell (changes requested):
@gmlewis as per my previous comment I don't think we should take this PR without a proper discussion.

gmlewis:
> @gmlewis as per my previous comment I don't think we should take this PR without a proper discussion.

OK, for posterity, this discussion is happening in #4365.

gmlewis:
Thanks for your patience, @huynhtrungcsc — same direction as #4366 (https://github.com/google/go-github/issues/4366#issuecomment-5717105620): the new PR will carry your `AllowedOrigins` design for both transports, credit you as co-author, and close this one.

Issue #4334, opened 2026-06-26
Metadata question


stevehipwell:
I happened to be re-implementing some acceptance tests the other day and while doing so I noticed that the signatures for the actions secret functions didn't look right. Tracking this back into the OpenAPI doc shows that the 5 functions below are tracking APIs from GHES v3.7 that are no longer documented (see #4335 for my proposed fix for this specific case).

- `ActionsService.GetEnvPublicKey`
- `ActionsService.ListEnvSecrets`
- `ActionsService.GetEnvSecret`
- `ActionsService.CreateOrUpdateEnvSecret`
- `ActionsService.DeleteEnvSecret`

Looking at the OpenAPI metadata I was expecting these endpoints to be marked as deprecated, but this doesn't appear to be the case. I can also see unused endpoints from old GHES versions still in the metadata.

My primary questions are as follows:

- Should endpoints be removed once they're no longer supported by GitHub and are no longer in use (e.g. there are some projects v1 endpoints still in there)?
- Should endpoints be marked as deprecated once they're no longer supported by GitHub (e.g. deprecated or not in supported GHES versions)?
- If the endpoints were correctly marked as deprecated would this automatically mark the functions as deprecated?
  - If this is the case do we have a process for updating functions that just need to change endpoint and/or signature?

stevehipwell:
CC @gmlewis

gmlewis:
> * Should endpoints be removed once they're no longer supported by GitHub and are no longer in use (e.g. there are some projects v1 endpoints still in there)?

This topic has been discussed many times in this repo over the years.
The bottom line is that we have challenges removing deprecated endpoints in this repo because after a release someone will say "Hey, I'm still running GitHub Enterprise version X.Y.Z and we still rely on these endpoints that you just removed!" at which point we tell them to pin this repo to a version that supports it, but then they say, "Yes, but version D.E.F of go-github fixed a bug in method G.H which we also really need!"
So the easiest thing to do over the years is to NOT remove deprecated endpoints so that our dear Enterprise users (which there are WAY more of than I had ever imagined there would be) do not get upset.

> * Should endpoints be marked as deprecated once they're no longer supported by GitHub (e.g. deprecated or not in supported GHES versions)?

Yes, they should, and lately we have been attempting to do that.

> * If the endpoints were correctly marked as deprecated would this automatically mark the functions as deprecated?

Yes, our "tools/metadata" program does this.

[...]

PR #4367, opened 2026-07-06
chore: Add deprecated flag for unused metadata

stevehipwell:
This PR add a new `--deprecated` flag to the metadata unused tool call to only list unused deprecated operations.

Related to #4334

### DIFF
--- a/tools/metadata/main.go
+++ b/tools/metadata/main.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 
 	"github.com/alecthomas/kong"
 	"github.com/google/go-github/v88/github"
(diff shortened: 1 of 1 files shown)


tools/metadata/main.go
```diff
@@ -44,7 +45,8 @@ consistent with the SHA listed in "openapi_commit". This is run in CI as a conve
 changes to openapi_operations.yaml.
 `,
 
-	"output_json_help": `Output JSON.`,
+	"unused_deprecated_help": `Onbly list deprecated operations in openapi_operations.yaml that aren't used by any service methods.`,
```
gmlewis:
```suggestion
	"unused_deprecated_help": `Only list deprecated operations in openapi_operations.yaml that aren't used by any service methods.`,
```

gmlewis (changes requested):
Thank you, @stevehipwell - one tiny nit, otherwise LGTM.

cc: @alexandear - @Not-Dhananjay-Mishra

tools/metadata/main.go
stevehipwell: Done

alexandear (approved):

gmlewis (approved):

PR #4371, opened 2026-07-07
fix: Enable submitting empty allowlist for actions permissions patterns

### LINKED ISSUE #4359: BUG: GitHub Actions Permissions do not allow emptying a once-set allowlist
Originating finding: https://github.com/integrations/terraform-provider-github/issues/2680

It seems that there isn't currently a way to set `PatternsAllowed` of `ActionsAllowed` to an empty list to override any previous additions to the allowlist since it uses `omitempty` [1](https://github.com/google/go-github/blob/master/github/actions_permissions_orgs.go#L39)

The API doesn't look it has any limitation on sending an empty Array: https://docs.github.com/en/rest/actions/permissions?apiVersion=2022-11-28#set-allowed-actions-and-reusable-workflows-for-an-organization--parameters

I can try to submit a patch for this in the next few weeks, if nobody gets to it faster ^^

deiga:
Resolves #4359

This PR changes `PatternsAllowed` to use `omitzero` instead of `omitempty`. That enables submitting an empty List to the API and thus override any currently set allowlisted patterns.

### DIFF
--- a/github/actions_permissions_orgs.go
+++ b/github/actions_permissions_orgs.go
@@ -34,9 +34,12 @@ type ActionsEnabledOnOrgRepos struct {
 //
 // GitHub API docs: https://docs.github.com/rest/actions/permissions?apiVersion=2022-11-28
 type ActionsAllowed struct {
-	GithubOwnedAllowed *bool    `json:"github_owned_allowed,omitempty"`
-	VerifiedAllowed    *bool    `json:"verified_allowed,omitempty"`
-	PatternsAllowed    []string `json:"patterns_allowed,omitzero"` // omitzero is used to allow empty array to be sent in the request body, to configure the organization to allow no 3rd-party actions. It needs to still omit a `nil` value, so that it's possible to patch the other values without changing the allowed actions.
+	GithubOwnedAllowed *bool `json:"github_owned_allowed,omitempty"`
+	VerifiedAllowed    *bool `json:"verified_allowed,omitempty"`
+	// PatternsAllowed uses `omitzero` to allow an empty array to be sent in the request body in order to
+	// configure the organization to allow no 3rd-party actions. It needs to still omit a `nil` value so
+	// that it's possible to patch the other values without changing the allowed actions.
+	PatternsAllowed []string `json:"patterns_allowed,omitzero"`
 }
 
 func (a ActionsAllowed) String() string {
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @deiga!
LGTM.

github/actions_permissions_orgs.go
```diff
@@ -36,7 +36,7 @@ type ActionsEnabledOnOrgRepos struct {
 type ActionsAllowed struct {
 	GithubOwnedAllowed *bool    `json:"github_owned_allowed,omitempty"`
 	VerifiedAllowed    *bool    `json:"verified_allowed,omitempty"`
-	PatternsAllowed    []string `json:"patterns_allowed,omitempty"`
+	PatternsAllowed    []string `json:"patterns_allowed,omitzero"`
```
alexandear: Let's add a comment about distinguishing passing an empty slice `[]string{}` and passing `nil` slice and why we need both.

alexandear (approved):

stevehipwell (approved):
LGTM

github/actions_permissions_orgs.go
deiga: @alexandear Done. I hope it's placed correctly 😬

github/actions_permissions_orgs.go
```diff
@@ -36,7 +36,7 @@ type ActionsEnabledOnOrgRepos struct {
 type ActionsAllowed struct {
 	GithubOwnedAllowed *bool    `json:"github_owned_allowed,omitempty"`
 	VerifiedAllowed    *bool    `json:"verified_allowed,omitempty"`
-	PatternsAllowed    []string `json:"patterns_allowed,omitempty"`
+	PatternsAllowed    []string `json:"patterns_allowed,omitzero"` // omitzero is used to allow empty array to be sent in the request body, to configure the organization to allow no 3rd-party actions. It needs to still omit a `nil` value, so that it's possible to patch the other values without changing the allowed actions.
```
gmlewis:
```suggestion
	// PatternsAllowed uses `omitzero` allow an empty array to be sent in the request body in order to
	// configure the organization to allow no 3rd-party actions. It needs to still omit a `nil` value so
	// that it's possible to patch the other values without changing the allowed actions.
	PatternsAllowed    []string `json:"patterns_allowed,omitzero"`
```

gmlewis (changes requested):

gmlewis (approved):

PR #4377, opened 2026-07-10
chore: Fix flaky tests with deterministic runs

gmlewis:
This PR addresses a recent race failure on Windows demonstrated here:
https://github.com/gmlewis/go-github/actions/runs/29060026853/job/86259654619

The crash was fatal error: bad g->status in ready — a Go runtime bug on Windows + -race, triggered when parallel tests with goroutines/timers complete under heavy load.

Root cause of flakiness: Two rate-limit tests relied on wall-clock timing:
- abortSleepContextCancelled: spawned a goroutine running client.Do + used time.After(10s) timeouts (2 real timers active)
- abortSleepContextCancelledClientLimit: used context.WithTimeout(10ms) — a tiny window racing with goroutine scheduling

Fix — made both fully deterministic with zero real-time dependencies:

1. abortSleepContextCancelled: Removed the errCh goroutine + time.After timeouts. Now calls client.Do synchronously; a tiny goroutine just waits for the handler's requestReceived signal then calls cancel(). The HTTP transport is synchronous, so the handler always completes before the rate-limit sleep begins — cancel fires either just before or during the sleep, both handled identically by sleepUntilResetWithBuffer's select.
2. abortSleepContextCancelledClientLimit: Replaced WithTimeout(10ms) with a pre-cancelled context (WithCancel + immediate cancel()). checkRateLimitBeforeDo → sleepUntilResetWithBuffer sees ctx.Done() already closed and returns instantly — no timing race at all.

Also fixed a copy-paste bug: "Expected 1 requests" → "Expected 0 requests".

### DIFF
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -2839,6 +2839,18 @@ func TestDo_rateLimit_sleepUntilClientResetLimit(t *testing.T) {
 }
 
 // Ensure sleep is aborted when the context is cancelled.
+//
+// The test verifies that when a request receives a 403 rate-limit response,
+// the client begins sleeping until the reset time, and that canceling the
+// context during (or just before) that sleep aborts it and returns
+// context.Canceled.
+//
+// Determinism: The handler signals via requestReceived once it has run,
+// guaranteeing exactly one request before cancellation. A tiny goroutine
+// waits for that signal and then calls cancel, so the context is cancelled
+// after the handler returns but independently of wall-clock timing. There
+// are no time.After timeouts or time.Sleep delays: the test either completes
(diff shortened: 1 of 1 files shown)


gmlewis:

stevehipwell (approved):
LGTM

@gmlewis as an aside did you consider using `synctest` to keep the timeout pattern?

gmlewis:
> @gmlewis as an aside did you consider using `synctest` to keep the timeout pattern?

Yes, in fact the agent started out fixing this flaky test by using synctest but then discovered it could completely remove all asynchrony and make the tests fully deterministic without changing their primary objectives without it.

PR #4378, opened 2026-07-10
refactor!: Pass `TemplateRepoRequest` by value in `Repositories.CreateFromTemplate`

JamBalaya56562:
BREAKING CHANGE: `RepositoriesService.CreateFromTemplate` now passes `body` by value and `Name` is now required and passed by value.

Towards #3644.

`Repositories.CreateFromTemplate` now takes its `TemplateRepoRequest` body by value instead of by pointer, and the required `name` property becomes a non-pointer `string` (its `omitempty` is dropped), so the required body is enforced at compile time.

The OpenAPI schema for `POST /repos/{template_owner}/{template_repo}/generate` lists `name` as the only required property, and its properties (`owner`, `name`, `description`, `include_all_branches`, `private`) map 1:1 to the struct — no fields are missing, so nothing is added.

`TemplateRepoRequest` is removed from the `paramcheck` `body-allowed-pointer-types` allowlist in `.golangci.yml`, and the generated accessors are regenerated (`GetName` now returns the value directly).

### DIFF
--- a/github/repos.go
+++ b/github/repos.go
@@ -620,13 +620,11 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
 
 // TemplateRepoRequest represents a request to create a repository from a template.
 type TemplateRepoRequest struct {
-	// Name is required when creating a repo.
-	Name        string  `json:"name"`
-	Owner       *string `json:"owner,omitempty"`
-	Description *string `json:"description,omitempty"`
-
-	IncludeAllBranches *bool `json:"include_all_branches,omitempty"`
-	Private            *bool `json:"private,omitempty"`
+	Name               string  `json:"name"`
+	Owner              *string `json:"owner,omitempty"`
+	Description        *string `json:"description,omitempty"`
+	IncludeAllBranches *bool   `json:"include_all_branches,omitempty"`
+	Private            *bool   `json:"private,omitempty"`
 }
 
 // CreateFromTemplate generates a repository from a template.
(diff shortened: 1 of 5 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/repos.go
```diff
@@ -621,7 +621,7 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
 // TemplateRepoRequest represents a request to create a repository from a template.
 type TemplateRepoRequest struct {
 	// Name is required when creating a repo.
```
alexandear: This comment is redundant and can be removed.
```suggestion
```

github/repos.go
```diff
@@ -621,7 +621,7 @@ func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repo
...
-	Name        *string `json:"name,omitempty"`
+	Name        string  `json:"name"`
 	Owner       *string `json:"owner,omitempty"`
 	Description *string `json:"description,omitempty"`
```
alexandear: Let's remove this empty line:
```suggestion
```

github/repos.go
JamBalaya56562: Done, removed the redundant comment.

github/repos.go
JamBalaya56562: Done, removed the empty line.

alexandear (approved):

PR #4382, opened 2026-07-15
refactor!: Pass `CreateDeploymentBranchPolicyRequest` and `UpdateDeploymentBranchPolicyRequest` by value

JamBalaya56562:
BREAKING CHANGE: `RepositoriesService.CreateDeploymentBranchPolicy` and `UpdateDeploymentBranchPolicy` now take `body` by value and the required `Name` field is of type `string`.

Towards #3644.

The create and update operations use different request schemas, so this replaces the shared `DeploymentBranchPolicyRequest` with two dedicated request types, both passed by value instead of by pointer:

| Method | Schema | `required` | Properties |
| --- | --- | --- | --- |
| `CreateDeploymentBranchPolicy` (POST) | `deployment-branch-policy-name-pattern-with-type` | `[name]` | `name`, `type` |
| `UpdateDeploymentBranchPolicy` (PUT) | `deployment-branch-policy-name-pattern` | `[name]` | `name` |

`CreateDeploymentBranchPolicyRequest` has a required non-pointer `Name` and an optional `Type`; `UpdateDeploymentBranchPolicyRequest` has only the required non-pointer `Name`, so the update method no longer exposes a `type` field the endpoint does not accept.

`DeploymentBranchPolicyRequest` is removed from the `paramcheck` `body-allowed-pointer-types` allowlist in `.golangci.yml`, and the generated accessors are regenerated (`GetName` returns the value directly).

### DIFF
--- a/github/repos_deployment_branch_policies.go
+++ b/github/repos_deployment_branch_policies.go
@@ -24,10 +24,15 @@ type DeploymentBranchPolicyResponse struct {
 	BranchPolicies []*DeploymentBranchPolicy `json:"branch_policies,omitempty"`
 }
 
-// DeploymentBranchPolicyRequest represents a deployment branch policy request.
-type DeploymentBranchPolicyRequest struct {
+// CreateDeploymentBranchPolicyRequest represents a request to create a deployment branch policy.
+type CreateDeploymentBranchPolicyRequest struct {
 	Name string  `json:"name"`
 	Type *string `json:"type,omitempty"`
+}
+
+// UpdateDeploymentBranchPolicyRequest represents a request to update a deployment branch policy.
+type UpdateDeploymentBranchPolicyRequest struct {
+	Name string `json:"name"`
 }
 
 // ListDeploymentBranchPolicies lists the deployment branch policies for an environment.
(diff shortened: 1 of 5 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/repos_deployment_branch_policies.go
```diff
@@ -105,7 +105,7 @@ func (s *RepositoriesService) CreateDeploymentBranchPolicy(ctx context.Context,
 // GitHub API docs: https://docs.github.com/rest/deployments/branch-policies?apiVersion=2022-11-28#update-a-deployment-branch-policy
 //
 //meta:operation PUT /repos/{owner}/{repo}/environments/{environment_name}/deployment-branch-policies/{branch_policy_id}
-func (s *RepositoriesService) UpdateDeploymentBranchPolicy(ctx context.Context, owner, repo, environment string, branchPolicyID int64, body *DeploymentBranchPolicyRequest) (*DeploymentBranchPolicy, *Response, error) {
+func (s *RepositoriesService) UpdateDeploymentBranchPolicy(ctx context.Context, owner, repo, environment string, branchPolicyID int64, body DeploymentBranchPolicyRequest) (*DeploymentBranchPolicy, *Response, error) {
```
alexandear: The body contains only the `name` according to the spec. We can't use `DeploymentBranchPolicyRequest` as `body` here.

Image

alexandear (changes requested):
I propose to add two separate structs: `CreateDeploymentBranchPolicyRequest` and `UpdateDeploymentBranchPolicy`.

github/repos_deployment_branch_policies.go
JamBalaya56562: Done in `df5d6b5a` — `UpdateDeploymentBranchPolicy` now takes a dedicated `UpdateDeploymentBranchPolicyRequest` with only `name`, matching the update schema.

JamBalaya56562:
@alexandear you're right — thanks. Done in `df5d6b5a`: I split the shared type into `CreateDeploymentBranchPolicyRequest` (required `name` + optional `type`) and `UpdateDeploymentBranchPolicyRequest` (required `name` only), so each matches its schema and the update body no longer carries `type`.

I named the update type `UpdateDeploymentBranchPolicyRequest` (with the `Request` suffix) since `UpdateDeploymentBranchPolicy` is already the method name — let me know if you'd prefer a different name.

alexandear (approved):

PR #4394, opened 2026-07-16
refactor!: Pass `SarifAnalysis` by value

JamBalaya56562:
BREAKING CHANGE: `CodeScanningService.UploadSarif` now takes `body` by value and its required fields are no longer pointers.

Towards #3644.

This has two commits:

1. **`refactor!`** — `CodeScanning.UploadSarif` now takes its `SarifAnalysis` body by value instead of by pointer, and the three required properties become non-pointer strings (their `omitempty` is dropped). The OpenAPI schema for `POST /repos/{owner}/{repo}/code-scanning/sarifs` lists `commit_sha`, `ref`, and `sarif` as `required`, so `SarifAnalysis.CommitSHA`, `.Ref`, and `.Sarif` are now `string`. `SarifAnalysis` is removed from the `paramcheck` `body-allowed-pointer-types` allowlist in `.golangci.yml`.

2. **`feat`** — adds the optional `Validate *bool` (`validate`) property, which the same endpoint accepts but `SarifAnalysis` was missing.

`SarifAnalysis` is the dedicated body of that single endpoint (the response is a separate `SarifID`), so there is no shared-type / split concern. The generated accessors are regenerated.

### DIFF
--- a/github/code_scanning.go
+++ b/github/code_scanning.go
@@ -215,8 +215,7 @@ type SarifAnalysis struct {
 	CheckoutURI *string    `json:"checkout_uri,omitempty"`
 	StartedAt   *Timestamp `json:"started_at,omitempty"`
 	ToolName    *string    `json:"tool_name,omitempty"`
-	// Validate indicates whether the SARIF file will be validated according to the code scanning specification.
-	Validate *bool `json:"validate,omitempty"`
+	Validate    *bool      `json:"validate,omitempty"`
 }
 
 // CodeScanningAlertState specifies the state of a code scanning alert.
--- a/github/code_scanning.go
+++ b/github/code_scanning.go
@@ -209,9 +209,9 @@ type ScanningAnalysis struct {
 //
 // GitHub API docs: https://docs.github.com/rest/code-scanning?apiVersion=2022-11-28
 type SarifAnalysis struct {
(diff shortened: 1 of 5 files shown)


github/code_scanning.go
```diff
@@ -209,12 +209,14 @@ type ScanningAnalysis struct {
...
-	CommitSHA   *string    `json:"commit_sha,omitempty"`
-	Ref         *string    `json:"ref,omitempty"`
-	Sarif       *string    `json:"sarif,omitempty"`
+	CommitSHA   string     `json:"commit_sha"`
+	Ref         string     `json:"ref"`
+	Sarif       string     `json:"sarif"`
 	CheckoutURI *string    `json:"checkout_uri,omitempty"`
 	StartedAt   *Timestamp `json:"started_at,omitempty"`
 	ToolName    *string    `json:"tool_name,omitempty"`
+	// Validate indicates whether the SARIF file will be validated according to the code scanning specification.
```
alexandear: Please remove this comment. Or add comments for all fields.

JamBalaya56562: Done in `7b0f7e7d` — removed it. The other fields have no comments, so the field name and json tag should speak for themselves.

alexandear (approved):

PR #4395, opened 2026-07-17
refactor!: Rename `NewPullRequest` to `CreatePullRequest` and pass it by value

JamBalaya56562:
BREAKING CHANGE: `NewPullRequest` is renamed to `CreatePullRequest`, `PullRequests.Create` now takes it by value, and `CreatePullRequest.Head` and `CreatePullRequest.Base` are now `string`.

Towards #3644.

`PullRequests.Create` now takes its request body by value instead of by pointer, the type is renamed from `NewPullRequest` to `CreatePullRequest` (the `New` prefix reads as a constructor function; `Create` matches the create-a-pull-request operation), and the two required properties become non-pointer strings (their `omitempty` is dropped): the OpenAPI schema for `POST /repos/{owner}/{repo}/pulls` lists `head` and `base` as `required`, so `CreatePullRequest.Head` and `CreatePullRequest.Base` are now `string`.

`title` is not in the schema's `required` list (an issue number can be supplied via `issue` instead of a title), so `Title` and the other optional fields stay pointers. The schema's eight properties map 1:1 to the struct — nothing is missing, so nothing is added.

`CreatePullRequest` is the dedicated body of that single endpoint (the response is the separate `PullRequest` type), so there is no shared-type / split concern. The type is removed from the `paramcheck` `body-allowed-pointer-types` allowlist in `.golangci.yml`, the generated accessors are regenerated (`GetHead`/`GetBase` return the value directly), and the `example/commitpr` sample and `ExampleClient_pullRequests` are updated to match.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -258,8 +258,8 @@ func (s *PullRequestsService) GetRaw(ctx context.Context, owner, repo string, nu
 	return buf.String(), resp, nil
 }
 
-// NewPullRequest represents a new pull request to be created.
-type NewPullRequest struct {
+// CreatePullRequest represents a request to create a pull request.
+type CreatePullRequest struct {
 	Title *string `json:"title,omitempty"`
 	// The name of the branch where your changes are implemented. For
 	// cross-repository pull requests in the same network, namespace head with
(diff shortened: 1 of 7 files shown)


github/pulls.go
```diff
@@ -282,7 +282,7 @@ type NewPullRequest struct {
 // GitHub API docs: https://docs.github.com/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request
 //
 //meta:operation POST /repos/{owner}/{repo}/pulls
-func (s *PullRequestsService) Create(ctx context.Context, owner, repo string, body *NewPullRequest) (*PullRequest, *Response, error) {
+func (s *PullRequestsService) Create(ctx context.Context, owner, repo string, body NewPullRequest) (*PullRequest, *Response, error) {
```
alexandear: I propose to rename `NewPullRequest` to `CreatePullRequest`:
```suggestion
func (s *PullRequestsService) Create(ctx context.Context, owner, repo string, body CreatePullRequest) (*PullRequest, *Response, error) {
```

The `New` prefix usually means a "constructor" function.

JamBalaya56562: Done in `bd7f6b01` — renamed `NewPullRequest` to `CreatePullRequest` (type, tests, examples, and the `example/commitpr` sample), and updated the PR title/description accordingly.

alexandear (approved):

PR #4396, opened 2026-07-18
refactor!: Split `IssueRequest` into `CreateIssueRequest` & `UpdateIssueRequest` and pass by value

jvm986:
BREAKING CHANGE: `IssueService.Edit` is renamed to `IssueService.Update` to match the underlying "update an issue" endpoint.

`IssuesService.Create` now takes a dedicated `CreateIssueRequest` by value (with a required non-pointer `Title`).
`IssuesService.Updated` takes `UpdateIssueRequest` by value.

Towards #3644.

### DIFF
--- a/github/issues.go
+++ b/github/issues.go
@@ -123,17 +123,21 @@ type CreateIssueRequest struct {
 // It is separate from Issue above because otherwise Labels
 // and Assignee fail to serialize to the correct JSON.
 type IssueRequest struct {
-	Title    *string   `json:"title,omitempty"`
-	Body     *string   `json:"body,omitempty"`
-	Labels   *[]string `json:"labels,omitempty"`
-	Assignee *string   `json:"assignee,omitempty"`
-	State    *string   `json:"state,omitempty"`
+	Title *string `json:"title,omitempty"`
+	Body  *string `json:"body,omitempty"`
+	// Labels: nil leaves existing labels unchanged; a non-nil slice replaces
+	// them, so []string{} clears all labels.
+	Labels   []string `json:"labels,omitzero"`
+	Assignee *string  `json:"assignee,omitempty"`
+	State    *string  `json:"state,omitempty"`
 	// StateReason can be `completed`, `not_planned`, `duplicate` or `reopened`.
 	StateReason *string `json:"state_reason,omitempty"`
 	// DuplicateIssueID is required when state_reason is `duplicate`.
-	DuplicateIssueID *int                      `json:"duplicate_issue_id,omitempty"`
-	Milestone        *int                      `json:"milestone,omitempty"`
-	Assignees        *[]string                 `json:"assignees,omitempty"`
+	DuplicateIssueID *int `json:"duplicate_issue_id,omitempty"`
+	Milestone        *int `json:"milestone,omitempty"`
+	// Assignees: nil leaves existing assignees unchanged; a non-nil slice
+	// replaces them, so []string{} clears all assignees.
+	Assignees        []string                  `json:"assignees,omitzero"`
 	Type             *string                   `json:"type,omitempty"`
 	IssueFieldValues []*IssueRequestFieldValue `json:"issue_field_values,omitempty"`
 }
--- a/github/issues.go
+++ b/github/issues.go
@@ -491,12 +491,12 @@ func (s *IssuesService) Create(ctx context.Context, owner, repo string, body Cre
 	return i, resp, nil
 }
 
-// Edit (update) an issue.
+// Update an issue.
 //
 // GitHub API docs: https://docs.github.com/rest/issues/issues?apiVersion=2022-11-28#update-an-issue
 //
 //meta:operation PATCH /repos/{owner}/{repo}/issues/{issue_number}
-func (s *IssuesService) Edit(ctx context.Context, owner, repo string, number int, body IssueRequest) (*Issue, *Response, error) {
+func (s *IssuesService) Update(ctx context.Context, owner, repo string, number int, body IssueRequest) (*Issue, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/issues/%v", owner, repo, number)
 	req, err := s.client.NewRequest(ctx, "PATCH", u, body)
 	if err != nil {
--- a/github/issues.go
+++ b/github/issues.go
@@ -108,7 +108,6 @@ func (i Issue) IsPullRequest() bool {
 // It is separate from Issue above because otherwise Labels
 // and Assignee fail to serialize to the correct JSON.
 type CreateIssueRequest struct {
-	// Title is required when creating an issue.
 	Title            string                    `json:"title"`
 	Body             *string                   `json:"body,omitempty"`
 	Labels           []string                  `json:"labels,omitempty"`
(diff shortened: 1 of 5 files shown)


gmlewis (approved):
Thank you, @jvm986!
LGTM.

cc: @stevehipwell - @alexandear - @Not-Dhananjay-Mishra - @JamBalaya56562 

Oh, and @jvm986 and @JamBalaya56562 - feel free to "cc:" each other on each PR to be extra careful that work is not duplicated.

jvm986:
@gmlewis, @JamBalaya56562 

What are your thoughts on the `CreateIssueRequest` question?

> IssueRequest is shared by Create (API requires title) and Edit (all optional), so I only changed the param to a value — promoting Title to a required string would break Edit. Should we split into a dedicated CreateIssueRequest (like https://github.com/google/go-github/pull/4382 did for deployment branch policies), or is the shared type fine here?

gmlewis:
> @gmlewis, @JamBalaya56562
> 
> What are your thoughts on the `CreateIssueRequest` question?
> 
> > IssueRequest is shared by Create (API requires title) and Edit (all optional), so I only changed the param to a value — promoting Title to a required string would break Edit. Should we split into a dedicated CreateIssueRequest (like #4382 did for deployment branch policies), or is the shared type fine here?

I think there was a discussion for #4382 where the change of name was deemed valuable, so I think we should also address that here in the breaking change PR to remain consistent.

Thank you for bringing this up!  I appreciate it.

github/issues.go
```diff
@@ -104,7 +104,22 @@ func (i Issue) IsPullRequest() bool {
...
+	Milestone        *int                      `json:"milestone,omitempty"`
+	Assignees        []string                  `json:"assignees,omitempty"`
+	Type             *string                   `json:"type,omitempty"`
+	IssueFieldValues []*IssueRequestFieldValue `json:"issue_field_values,omitempty"`
+}
+
+// IssueRequest represents a request to edit an issue.
 // It is separate from Issue above because otherwise Labels
 // and Assignee fail to serialize to the correct JSON.
 type IssueRequest struct {
```
Not-Dhananjay-Mishra: Can we update `IssueRequest` struct too. 
It has one missing field `duplicate_issue_id` and we should use `[]string` instead of `*[]string` for `Labels` and `Assignees`

github/issues.go
```diff
@@ -479,7 +494,7 @@ func (s *IssuesService) Create(ctx context.Context, owner, repo string, body *Is
 // GitHub API docs: https://docs.github.com/rest/issues/issues?apiVersion=2022-11-28#update-an-issue
 //
 //meta:operation PATCH /repos/{owner}/{repo}/issues/{issue_number}
-func (s *IssuesService) Edit(ctx context.Context, owner, repo string, number int, body *IssueRequest) (*Issue, *Response, error) {
+func (s *IssuesService) Edit(ctx context.Context, owner, repo string, number int, body IssueRequest) (*Issue, *Response, error) {
```
Not-Dhananjay-Mishra: Can we rename this to `Update`? as API docs mentioned it as ["Update an issue"](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue--code-samples)

github/issues.go
jvm986: Nice find on `duplicate_issue_id`, will add it.

Regarding `Labels` and `Assignees` the logic here is that the update endpoint treats `"labels": []` as "clear all labels" (distinct from omitting the field which leaves them unchanged). `[]string` isn't able to distinguish between omission and an explicit empty array.

github/issues.go
jvm986: Makes sense to me, will update after feedback from others.

gmlewis: > Makes sense to me, will update after feedback from others.

SGTM. Thank you, @Not-Dhananjay-Mishra and @jvm986!

I figure that while we are actually breaking the API, we should always attempt to make it as easy-to-understand and easy-to-use as possible, and this seems to fit into that category. :smile:

github/issues.go
Not-Dhananjay-Mishra: @gmlewis could you help me with this? I assumed that we don't use `*[]buildin` in this repo. 

> Regarding `Labels` and `Assignees` the logic here is that the update endpoint treats `"labels": []` as "clear all labels" (distinct from omitting the field which leaves them unchanged). `[]string` isn't able to distinguish between and an explicit empty array.

`omitzero` seems like a good fit here instead of `omitempty`, since it omits only `nil` slices while preserving an explicit empty slice.

gmlewis: > `omitzero` seems like a good fit here instead of `omitempty`, since it omits only `nil` slices while preserving an explicit empty slice.

Yes, exactly right, @Not-Dhananjay-Mishra - this is what `omitzero` is for. So we can finally get rid of the `*[]string`, change it to `[]string` and then use `omitzero` on it so that `nil` is different from `[]string{}`.

Not-Dhananjay-Mishra: Thanks @gmlewis.
@jvm986, you can change `*[]string` to `[]string` for `Labels` and `Assignees`, and use `omitzero` instead of `omitempty`.

gmlewis: > Thanks @gmlewis. @jvm986, you can change `*[]string` to `[]string` for `Labels` and `Assignees`, and use `omitzero` instead of `omitempty`.

Yes, please, and just be absolutely sure to include a table-driven unit test where _BOTH_ `nil` and `[]string{}` are passed to ensure that the generated JSON matches our expections, and preferrably add a comment that explains when to use one versus the other. Thank you, @jvm986!

github/issues.go
```diff
@@ -104,19 +104,40 @@ func (i Issue) IsPullRequest() bool {
...
-type IssueRequest struct {
-	Title    *string   `json:"title,omitempty"`
-	Body     *string   `json:"body,omitempty"`
-	Labels   *[]string `json:"labels,omitempty"`
-	Assignee *string   `json:"assignee,omitempty"`
-	State    *string   `json:"state,omitempty"`
-	// StateReason can be 'completed' or 'not_planned'.
-	StateReason      *string                   `json:"state_reason,omitempty"`
+type CreateIssueRequest struct {
+	// Title is required when creating an issue.
```
alexandear: This comment can be removed.

github/issues.go
```diff
@@ -474,12 +495,12 @@ func (s *IssuesService) Create(ctx context.Context, owner, repo string, body *Is
...
 }
 
-// Edit (update) an issue.
+// Update an issue.
 //
 // GitHub API docs: https://docs.github.com/rest/issues/issues?apiVersion=2022-11-28#update-an-issue
 //
 //meta:operation PATCH /repos/{owner}/{repo}/issues/{issue_number}
-func (s *IssuesService) Edit(ctx context.Context, owner, repo string, number int, body *IssueRequest) (*Issue, *Response, error) {
+func (s *IssuesService) Update(ctx context.Context, owner, repo string, number int, body IssueRequest) (*Issue, *Response, error) {
```
alexandear: Let's rename this as well:
```suggestion
func (s *IssuesService) Update(ctx context.Context, owner, repo string, number int, body UpdateIssueRequest) (*Issue, *Response, error) {
```

jvm986: Good call, as it's a breaking change anyway.

Not-Dhananjay-Mishra (approved):
LGTM 🚀
Just update Title and Description.

gmlewis:
Thank you, @jvm986, @alexandear, and @Not-Dhananjay-Mishra!
Merging.

PR #4399, opened 2026-07-21
refactor!: Rename `AutolinkOptions` to `CreateAutolinkRequest`, `AddAutolink` to `CreateAutolink`, and pass the body by value

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for `RepositoriesService.AddAutolink`.

### What changed

- `AutolinkOptions` is renamed to `CreateAutolinkRequest`, following the `...Request` convention for request bodies. The old name was in the `body-allowed-wrong-names` allowlist.
- `AddAutolink` is renamed to `CreateAutolink` to match the docs operation name, [Create an autolink reference for a repository](https://docs.github.com/rest/repos/autolinks?apiVersion=2022-11-28#create-an-autolink-reference-for-a-repository). The doc comment already read "creates an autolink reference".
- The request body is now passed by value, and both `AutolinkOptions` entries are removed from the `.golangci.yml` allowlists.
- `key_prefix` and `url_template` are required by the API, so `KeyPrefix` and `URLTemplate` become non-pointer `string` without `omitempty`. `is_alphanumeric` is optional (it defaults to `true`), so `IsAlphanumeric` stays `*bool`.

### Notes

- The request schema has exactly these three properties, so nothing was missing from the Go struct and no field needed to be added.
- The response type `Autolink` is a separate struct and is unchanged — this is a request-only type, so there is no create/update split to consider.
- `test/integration/repos_test.go` builds under the root module with `-tags integration`, so it is updated too. Its comparison needed a small fix rather than a mechanical rename: it compared `actionlink.KeyPrefix` (`*string`) against the request field, which is now a `string`, so it goes through `GetKeyPrefix()`/`GetURLTemplate()` instead.

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (`CreateAutolink` and the regenerated accessors at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entries).

Updates #3644

BREAKING CHANGE: AutolinkOptions is renamed to CreateAutolinkRequest with non-pointer KeyPrefix and URLTemplate; RepositoriesService.AddAutolink is renamed to CreateAutolink and now takes the request body by value.

cc @jvm986 — flagging for #3644 coordination; this one is in the `repos` service, so it shouldn't overlap with the `Issues` work.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -206,7 +206,6 @@ linters:
             - ActionsVariable
             - AddProjectItemOptions
             - AuthorizationRequest
-            - AutolinkOptions
             - CodeScanningAlertState
             - CodespaceCreateForUserOptions
             - ConfigApplyOptions
@@ -276,7 +275,6 @@ linters:
           # TODO: fix and remove these exceptions.
           body-allowed-wrong-names:
             - AddProjectItemOptions
-            - AutolinkOptions
             - CheckSuitePreferenceOptions
             - CodespaceCreateForUserOptions
             - ConfigApplyOptions
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -2862,30 +2862,6 @@ func (a *Autolink) GetURLTemplate() string {
 	return *a.URLTemplate
 }
 
-// GetIsAlphanumeric returns the IsAlphanumeric field if it's non-nil, zero value otherwise.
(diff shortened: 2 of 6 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

alexandear (approved):

PR #4400, opened 2026-07-22
refactor!: Rename `EditLabel` to `UpdateLabel`, Split `Label` into `CreateLabelRequest` & `UpdateLabelRequest` and pass by value

jvm986:
Continues the request-body-by-value work in #3644, for `IssuesService.Label`.

BREAKING CHANGE: `IssuesService.CreateLabel` now takes a `CreateLabelRequest` by value (with a required non-pointer `Name`). `IssuesService.EditLabel` renamed to `UpdateLabel`, taking an `UpdateLabelRequest` by value instead of a `*Label` (with all fields optional).

## Notes
- The update request takes `new_name`, not `name` which appears to have been a latent bug.

cc @JamBalaya56562 flagging for #3644 coordination

### DIFF
--- a/github/issues_labels.go
+++ b/github/issues_labels.go
@@ -25,6 +25,20 @@ func (l Label) String() string {
 	return Stringify(l)
 }
 
+// CreateLabelRequest represents a request to create a label.
+type CreateLabelRequest struct {
+	Name        string  `json:"name"`
+	Color       *string `json:"color,omitempty"`
+	Description *string `json:"description,omitempty"`
+}
+
+// UpdateLabelRequest represents a request to update a label.
+type UpdateLabelRequest struct {
+	NewName     *string `json:"new_name,omitempty"`
+	Color       *string `json:"color,omitempty"`
+	Description *string `json:"description,omitempty"`
+}
+
 // ListLabels lists all labels for a repository.
 //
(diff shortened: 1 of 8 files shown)


github/issues_labels.go
```diff
@@ -25,6 +25,20 @@ func (l Label) String() string {
...
 
+// CreateLabelRequest represents a request to create a label.
+type CreateLabelRequest struct {
+	Name        string  `json:"name"`
+	Color       *string `json:"color,omitempty"`
+	Description *string `json:"description,omitempty"`
+}
+
+// UpdateLabelRequest represents a request to update a label.
+type UpdateLabelRequest struct {
```
alexandear: `CreateLabelRequest`, `UpdateLabelRequest` are too broad. Let's add a prefix:

```suggestion
// IssueCreateLabelRequest represents a request to create a label.
type CreateLabelRequest struct {
	Name        string  `json:"name"`
	Color       *string `json:"color,omitempty"`
	Description *string `json:"description,omitempty"`
}

// IssueUpdateLabelRequest represents a request to update a label.
type UpdateLabelRequest struct {
```

alexandear (changes requested):
Please also fix the `Label` struct according to the schema.

github/issues_labels.go
jvm986: I left the verb at the start, as that seems to align more with the rest of the repository. Always happy to be challenged here.

alexandear (approved):

PR #4401, opened 2026-07-22
refactor!: Split `CreateOrUpdateCustomRepoRoleOptions` into `CreateCustomRepoRoleRequest` and `UpdateCustomRepoRoleRequest` and pass by value

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for the custom repository role endpoints on `OrganizationsService`.

`CreateOrUpdateCustomRepoRoleOptions` was a single type shared by both the create and update methods, but the two endpoints don't share the same required fields:

| Field | Create (POST) | Update (PATCH) |
|---|---|---|
| `name` | **required** | optional |
| `base_role` | **required** | optional |
| `permissions` | **required** | optional |
| `description` | optional | optional |

So the shared type is split into two request types, matching the split done for deployment branch policies in #4382:

- **`CreateCustomRepoRoleRequest`** — `Name` and `BaseRole` become non-pointer `string` (no `omitempty`) since they're required; `Permissions` stays required; `Description` stays optional (`*string`).
- **`UpdateCustomRepoRoleRequest`** — every field is optional, so all stay pointers (with `Permissions` gaining `omitempty`).

`CreateCustomRepoRole` and `UpdateCustomRepoRole` now take their body by value, and both entries are removed from the `.golangci.yml` allowlists. The method verbs already match the docs operation names ([Create](https://docs.github.com/en/rest/orgs/custom-roles?apiVersion=2022-11-28#create-a-custom-repository-role) / [Update](https://docs.github.com/en/rest/orgs/custom-roles?apiVersion=2022-11-28#update-a-custom-repository-role)), so no method rename is needed. The response type `CustomRepoRoles` is unchanged.

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (both methods and the regenerated accessors at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entries).

Updates #3644

BREAKING CHANGE: CreateOrUpdateCustomRepoRoleOptions is split into CreateCustomRepoRoleRequest (with non-pointer Name and BaseRole) and UpdateCustomRepoRoleRequest; OrganizationsService.CreateCustomRepoRole and UpdateCustomRepoRole now take these request types by value.

cc @jvm986 — flagging for #3644 coordination; this is in the `orgs` service, so it shouldn't overlap with the `Issues` work.

### DIFF
--- a/github/orgs_custom_repository_roles.go
+++ b/github/orgs_custom_repository_roles.go
@@ -43,7 +43,7 @@ type UpdateCustomRepoRoleRequest struct {
 	Name        *string  `json:"name,omitempty"`
 	Description *string  `json:"description,omitempty"`
 	BaseRole    *string  `json:"base_role,omitempty"`
-	Permissions []string `json:"permissions,omitempty"`
+	Permissions []string `json:"permissions,omitzero"`
 }
 
 // RepoFineGrainedPermission represents a fine-grained permission that can be used in a custom repository role.
--- a/github/orgs_custom_repository_roles.go
+++ b/github/orgs_custom_repository_roles.go
@@ -30,12 +30,20 @@ type CustomRepoRoles struct {
 	UpdatedAt   *Timestamp    `json:"updated_at,omitempty"`
 }
 
-// CreateOrUpdateCustomRepoRoleOptions represents options required to create or update a custom repository role.
-type CreateOrUpdateCustomRepoRoleOptions struct {
+// CreateCustomRepoRoleRequest represents the parameters to create a custom repository role.
+type CreateCustomRepoRoleRequest struct {
+	Name        string   `json:"name"`
+	Description *string  `json:"description,omitempty"`
+	BaseRole    string   `json:"base_role"`
+	Permissions []string `json:"permissions"`
+}
+
+// UpdateCustomRepoRoleRequest represents the parameters to update a custom repository role.
+type UpdateCustomRepoRoleRequest struct {
 	Name        *string  `json:"name,omitempty"`
 	Description *string  `json:"description,omitempty"`
 	BaseRole    *string  `json:"base_role,omitempty"`
-	Permissions []string `json:"permissions"`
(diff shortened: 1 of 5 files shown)


github/orgs_custom_repository_roles.go
```diff
@@ -30,12 +30,20 @@ type CustomRepoRoles struct {
...
+	Permissions []string `json:"permissions"`
+}
+
+// UpdateCustomRepoRoleRequest represents the parameters to update a custom repository role.
+type UpdateCustomRepoRoleRequest struct {
 	Name        *string  `json:"name,omitempty"`
 	Description *string  `json:"description,omitempty"`
 	BaseRole    *string  `json:"base_role,omitempty"`
-	Permissions []string `json:"permissions"`
+	Permissions []string `json:"permissions,omitempty"`
```
gmlewis: Should this one be `omitzero` to differentiate between `nil` and `[]string{}`?

JamBalaya56562: Good catch — done in db0c11e2a4ca407d899ffb93e2123ff956ae2f49.

You're right that `omitempty` collapses the two cases: it drops the field for both `nil` and `[]string{}`, so there was no way to send `"permissions": []` and clear the additional permissions on a role. With `omitzero` only the nil slice is omitted:

| value | `omitempty` | `omitzero` |
|---|---|---|
| `nil` | `{}` | `{}` |
| `[]string{}` | `{}` | `{"permissions":[]}` |

So nil now means "leave permissions untouched" and `[]string{}` means "clear them", which lines up with `ActionsAllowed.PatternsAllowed` and `DeploymentRequest.RequiredContexts`.

Left `CreateCustomRepoRoleRequest.Permissions` as a plain `json:"permissions"` since it's required on create and should always be sent.

gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

jvm986 (approved):

PR #4414, opened 2026-07-27
feat: Add `search_type` support to issue search

### LINKED ISSUE #4415: Add search_type support to issue search
The [search issues and pull requests](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests) endpoint documents a `search_type` query parameter:

> The type of search to perform on issues. When not specified, the default is lexical search. `semantic` performs a pure semantic (vector) search using embedding-based understanding. `hybrid` combines semantic search with lexical search for best results. Only applies to issue searches (`/search/issues`). Can be one of: `semantic`, `hybrid`.

`SearchOptions` has no field for it, so `Search.Issues` can only do lexical search. The response for these searches also carries a top-level `search_type`, which `IssuesSearchResult` doesn't expose, so callers can't tell which engine actually ran (the API falls back to lexical when a query isn't eligible for semantic).

I'm updating the [GitHub MCP server](https://github.com/github/github-mcp-server)'s `search_issues` tool to support semantic search, and we'd like to update the request and response types here rather than wrapping the HTTP transport to rewrite the query string and re-parse the response body.

Opened #4414 with the change.

kelsey-myers:
Fixes #4415 

The [search issues and pull requests](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests) endpoint supports a `search_type` query parameter (`semantic` or `hybrid`, defaulting to lexical), and returns the type that actually ran in the response body. Neither is currently exposed, so `Search.Issues` can only do lexical search, and callers can't tell when the API has fallen back to lexical for an ineligible query.

Adds `SearchType` to `SearchOptions` (mirroring `AdvancedSearch`) and to `IssuesSearchResult`.

I'm updating the [GitHub MCP server](https://github.com/github/github-mcp-server)'s `search_issues` tool to support semantic search, and we'd like to use these rather than wrapping the HTTP transport to rewrite the query string and re-parse the response body.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -21534,6 +21534,14 @@ func (i *IssuesSearchResult) GetIssues() []*Issue {
 	return i.Issues
 }
 
+// GetSearchType returns the SearchType field if it's non-nil, zero value otherwise.
+func (i *IssuesSearchResult) GetSearchType() string {
+	if i == nil || i.SearchType == nil {
+		return ""
+	}
+	return *i.SearchType
+}
+
 // GetTotal returns the Total field if it's non-nil, zero value otherwise.
 func (i *IssuesSearchResult) GetTotal() int {
 	if i == nil || i.Total == nil {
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @kelsey-myers!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

PR #4422, opened 2026-07-29
feat: Add `MetaService.ListAPIVersions`

### LINKED ISSUE #4420: Add MetaService.ListAPIVersions for GET /versions
`GET /versions` returns the REST API versions the server supports, and `MetaService` has no method for it:

https://docs.github.com/rest/meta/meta?apiVersion=2022-11-28#get-all-api-versions

The response is a JSON array of date strings, so it maps to `[]string` without adding a type. That matches `GitignoresService.List`, which backs the `gitignore/get-all-templates` operation and has the same signature and lack of pagination.

Note this is unrelated to the `apiVersionMin`/`apiVersionMax` constants in `github.go`, which are client side and stay hardcoded.

I plan to send a PR for this.

kyungseopk1m:
Fixes #4420

Adds `MetaService.ListAPIVersions` for `GET /versions`, operation `meta/get-all-versions`.

The 200 response schema is an array of `string` with `format: date`, so the method returns `[]string` and no new type is introduced. `GitignoresService.List` has the same signature for the same kind of operation.

Named `ListAPIVersions` rather than `Get...` because the repo uses `List` for operations returning multiple resources, and `MetaService` holds several unrelated resources so a bare `List` would be ambiguous.

The operation is already in `openapi_operations.yaml`, so `go generate ./...` produces no changes.

### DIFF
--- a/github/meta.go
+++ b/github/meta.go
@@ -139,6 +139,27 @@ func (c *Client) APIMeta(ctx context.Context) (*APIMeta, *Response, error) {
 	return c.Meta.Get(ctx)
 }
 
+// ListAPIVersions returns the API versions supported by the GitHub REST API,
+// as dates in the YYYY-MM-DD format.
+//
+// GitHub API docs: https://docs.github.com/rest/meta/meta?apiVersion=2022-11-28#get-all-api-versions
+//
+//meta:operation GET /versions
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @kyungseopk1m!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

PR #4406, opened 2026-07-24
refactor!: Rename `PullRequestReviewDismissalRequest` to `PullRequestDismissReviewRequest`, add `PullRequestSubmitReviewRequest`, and pass review request bodies by value

JamBalaya56562:
Continues the request-body-by-value work in #3644, for the review dismissal and submission endpoints on `PullRequestsService`.

### Dismiss a review

- `message` is [required](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#dismiss-a-review-for-a-pull-request), so `Message` becomes a non-pointer `string` (no `omitempty`), and `DismissReview` takes the body by value. The type is removed from the `body-allowed-pointer-types` allowlist.
- Per review: the type is renamed to `PullRequestDismissReviewRequest` to match the method name, and the previously unmodeled optional `event` parameter is added as `Event *string`.

### Submit a review

Per review, `SubmitReview` no longer reuses `*PullRequestReviewRequest` — whose `node_id`, `commit_id` and `comments` fields aren't part of the [submit-a-review schema](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#submit-a-review-for-a-pull-request) — and instead takes a new `PullRequestSubmitReviewRequest` by value, modeling that schema exactly: `Event` is required (non-pointer `string`), `Body` is optional (`*string`). `CreateReview` keeps the shared `PullRequestReviewRequest`, so its allowlist entry remains.

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite, and `custom-gcl` (no `paramcheck` findings).

Updates #3644

BREAKING CHANGE: PullRequestReviewDismissalRequest is renamed to PullRequestDismissReviewRequest with non-pointer Message; PullRequestsService.DismissReview takes it by value; PullRequestsService.SubmitReview now takes a new PullRequestSubmitReviewRequest (non-pointer Event) by value instead of *PullRequestReviewRequest.

cc @jvm986 — flagging for #3644 coordination; this is in the `pulls` service, so it shouldn't overlap with the `Issues` work.

PR #4425, opened 2026-07-31
refactor!: Pass `UpdateConnectedExternalGroup` request body by value via new `UpdateConnectedExternalGroupRequest`

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for `TeamsService.UpdateConnectedExternalGroup`.

The method reused the `ExternalGroup` **response** type as its request body, but per the [docs](https://docs.github.com/en/enterprise-cloud@latest/rest/teams/external-groups?apiVersion=2022-11-28#update-the-connection-between-an-external-group-and-a-team) the endpoint accepts exactly one body parameter, `group_id`, and it's required. The other four fields (`group_name`, `updated_at`, `teams`, `members`) are response-only and are ignored by the server, so the signature didn't tell callers what to fill in. The existing test shows this: it builds the request with only `GroupID` set, while the response expectation populates all five fields.

So this adds a dedicated request type, following the same approach as `PullRequestSubmitReviewRequest` in #4406:

```go
type UpdateConnectedExternalGroupRequest struct {
	GroupID int64 `json:"group_id"`
}
```

`GroupID` is a non-pointer `int64` since it's required, the body is passed by value, and the type is removed from the `body-allowed-pointer-types` allowlist.

Notes:

- `ExternalGroup` is unchanged — it stays the response type for `GetExternalGroup`, `ListExternalGroups`, `ExternalGroupList` and the iterator, so its fields keep their pointer semantics.
- The method verb `Update` already matches the docs operation name, so no rename.

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (`UpdateConnectedExternalGroup` and the generated `GetGroupID` at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entry).

Updates #3644

BREAKING CHANGE: TeamsService.UpdateConnectedExternalGroup now takes a new UpdateConnectedExternalGroupRequest (with non-pointer GroupID) by value instead of *ExternalGroup.

cc @jvm986 — flagging for #3644 coordination; this is in the `teams` service, so it shouldn't overlap with the `Issues` work.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -222,7 +222,6 @@ linters:
             - DependencyGraphSnapshot
             - EncryptedSecret
             - EnterpriseSecurityAnalysisSettings
-            - ExternalGroup
             - Hook
             - ImpersonateUserOptions
             - Import
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -43062,6 +43062,14 @@ func (u *UpdateCodespaceOptions) GetRecentFolders() []string {
 	return u.RecentFolders
 }
 
+// GetGroupID returns the GroupID field.
+func (u *UpdateConnectedExternalGroupRequest) GetGroupID() int64 {
+	if u == nil {
+		return 0
+	}
+	return u.GroupID
+}
+
 // GetBaseRole returns the BaseRole field if it's non-nil, zero value otherwise.
 func (u *UpdateCustomOrgRoleRequest) GetBaseRole() string {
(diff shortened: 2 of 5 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM

Not-Dhananjay-Mishra (approved):

PR #4432, opened 2026-08-05
refactor!: Pass LDAP mapping request bodies by value via new `UpdateUserLDAPMappingRequest` and `UpdateTeamLDAPMappingRequest`

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for the two LDAP mapping endpoints on `AdminService`.

Both methods reused their **response** types as request bodies, but per the [GHES docs](https://docs.github.com/en/enterprise-server@3.21/rest/enterprise-admin/ldap?apiVersion=2022-11-28) each endpoint accepts exactly one body parameter, `ldap_dn`, and it's required:

- `UpdateUserLDAPMapping` took `*UserLDAPMapping` — 17 all-pointer fields, most of them server-generated (`avatar_url`, `events_url`, …).
- `UpdateTeamLDAPMapping` took `*TeamLDAPMapping` — 10 all-pointer fields, same story.

The existing tests show this: both build the request with only `LDAPDN` set, while the response expectations populate more fields. So this adds dedicated request types, following the same approach as `PullRequestSubmitReviewRequest` (#4406) and `UpdateConnectedExternalGroupRequest` (#4425):

```go
type UpdateUserLDAPMappingRequest struct {
	LDAPDN string `json:"ldap_dn"`
}

type UpdateTeamLDAPMappingRequest struct {
	LDAPDN string `json:"ldap_dn"`
}
```

`LDAPDN` is a non-pointer `string` since it's required, both bodies are passed by value, and both old entries are removed from the `body-allowed-pointer-types` allowlist. The two methods live in the same file and follow the identical pattern, so they're converted together.

Notes:

- `UserLDAPMapping` and `TeamLDAPMapping` are unchanged — they stay the response types, so their fields keep pointer semantics.
- The method verbs already match the docs operation names ("Update LDAP mapping for a user/team"), so no method rename.
- The `team int64` parameter is renamed to `teamID` for clarity (non-breaking).

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (both methods and the generated `GetLDAPDN` accessors at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entries).

Updates #3644

BREAKING CHANGE: AdminService.UpdateUserLDAPMapping and UpdateTeamLDAPMapping now take new UpdateUserLDAPMappingRequest and UpdateTeamLDAPMappingRequest (with non-pointer LDAPDN) by value instead of *UserLDAPMapping and *TeamLDAPMapping.

cc @jvm986 — flagging for #3644 coordination; this is in the `admin` service, so it shouldn't overlap with the `Issues` work.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -258,13 +258,11 @@ linters:
             - Subscription
             - TeamAddTeamMembershipOptions
             - TeamAddTeamRepoOptions
-            - TeamLDAPMapping
             - TeamProjectOptions
             - UpdateCodespaceOptions
             - UpdateDefaultSetupConfigurationOptions
             - UpdateProjectItemOptions
             - User
-            - UserLDAPMapping
             - UserSuspendOptions
             - WorkflowsPermissionsOpt
           # Body type names exempt from the "Options" suffix rule.
--- a/github/admin.go
+++ b/github/admin.go
@@ -36,6 +36,12 @@ func (m TeamLDAPMapping) String() string {
 	return Stringify(m)
 }
 
+// UpdateTeamLDAPMappingRequest represents a request to update the mapping
+// between a GitHub team and an LDAP group.
+type UpdateTeamLDAPMappingRequest struct {
+	LDAPDN string `json:"ldap_dn"`
+}
+
 // UserLDAPMapping represents the mapping between a GitHub user and an LDAP user.
 type UserLDAPMapping struct {
 	ID         *int64  `json:"id,omitempty"`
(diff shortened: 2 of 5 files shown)


jvm986 (approved):
Looks good to me!

gmlewis (approved):
Thank you, @JamBalaya56562 and @jvm986!
LGTM.
Merging.

PR #4433, opened 2026-08-06
refactor!: Rename `UsersService.Edit` to `Update`, introduce `UserUpdateRequest`, and pass request by value

ManavSharma142:
BREAKING CHANGE: `UsersService.Edit` is now `Update` and uses a new `UserUpdateRequest` body value.

Relates to: #3644.

Rename UsersService.Edit to Update for consistency with GitHub API naming.
Introduce a dedicated UserUpdateRequest type containing only fields accepted by the update endpoint, and pass it by value instead of a pointer.

PR #4434, opened 2026-08-06
refactor!: Pass `UpdatePreReceiveHook` request body by value via new `UpdatePreReceiveHookRequest`

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for `RepositoriesService.UpdatePreReceiveHook` (GHES).

The method reused the `PreReceiveHook` **response** type as its request body, but per the [GHES docs](https://docs.github.com/en/enterprise-server@3.21/rest/enterprise-admin/repo-pre-receive-hooks?apiVersion=2022-11-28#update-pre-receive-hook-enforcement-for-a-repository) the endpoint accepts exactly one body parameter, `enforcement` — `id`, `name` and `configuration_url` are server-generated. The existing test illustrates the problem: it sent an empty `&PreReceiveHook{}` as the body, since the reused type gives no hint of what to fill in.

So this adds a dedicated request type, following the same approach as #4406, #4425 and #4432:

```go
type UpdatePreReceiveHookRequest struct {
	Enforcement *string `json:"enforcement,omitempty"`
}
```

`Enforcement` stays a pointer with `omitempty` since the docs don't mark it required. The body is passed by value, and the type's entry is removed from the `body-allowed-pointer-types` allowlist. The test now sends a meaningful body (`Enforcement: Ptr("enabled")`) instead of an empty struct.

Notes:

- `PreReceiveHook` is unchanged — it stays the response type for `ListPreReceiveHooks`, `GetPreReceiveHook` and this method, so its fields keep pointer semantics. (Its `ConfigURL` entry in the `structfield` allowlist is response-side debt and is untouched.)
- The method verb `Update` already matches the docs operation name, so no rename.

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (`UpdatePreReceiveHook` and the generated `GetEnforcement` at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entry).

Updates #3644

BREAKING CHANGE: RepositoriesService.UpdatePreReceiveHook now takes a new UpdatePreReceiveHookRequest by value instead of *PreReceiveHook.

cc @jvm986 — flagging for #3644 coordination; this is in the `repos` service (GHES pre-receive hooks), so it doesn't overlap with the `Issues` work, nor with #4433 (`UsersService`).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -237,7 +237,6 @@ linters:
             - PagesUpdate
             - PagesUpdateWithoutCNAME
             - PendingDeploymentsRequest
-            - PreReceiveHook
             - ProtectionRequest
             - PublishCodespaceOptions
             - PullRequestBranchUpdateOptions
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -43582,6 +43582,14 @@ func (u *UpdateOrganizationPrivateRegistry) GetVisibility() *PrivateRegistryVisi
 	return u.Visibility
 }
 
+// GetEnforcement returns the Enforcement field if it's non-nil, zero value otherwise.
+func (u *UpdatePreReceiveHookRequest) GetEnforcement() string {
+	if u == nil || u.Enforcement == nil {
+		return ""
+	}
+	return *u.Enforcement
+}
+
 // GetArchived returns the Archived field if it's non-nil, zero value otherwise.
 func (u *UpdateProjectItemOptions) GetArchived() bool {
 	if u == nil || u.Archived == nil {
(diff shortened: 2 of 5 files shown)


gmlewis:
cc: @ManavSharma142

gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

ManavSharma142 (approved):
LGTM

jvm986 (approved):
Looks good to me

gmlewis:
Thanks, all!  Merging.

PR #4436, opened 2026-08-06
feat: Add stacked pull request endpoints

### LINKED ISSUE #4435: Support for new stacked pull requests endpoints
GitHub now have endpoints for Stacked Pull Requests.
docs - https://docs.github.com/en/rest/pulls/stacks?apiVersion=2022-11-28

Bortlesboat:
Fixes #4435

Add typed `PullRequestsService` support for GitHub's five stacked pull request
REST endpoints:

- list repository stacks, optionally filtered by pull request;
- create and get a stack;
- append pull requests to a stack; and
- unstack the remaining unmerged pull requests.

The change keeps pull-request membership metadata separate from the endpoint
stack resource, adds typed request and list-option types, handles both
updated-stack and `204 No Content` unstack responses, and regenerates accessors
and iterators. Request methods,
paths, query parameters, bodies, response decoding, and failure paths have
focused coverage.

API contract: https://docs.github.com/en/rest/pulls/stacks?apiVersion=2022-11-28

Validation:

- `script/fmt.sh`
- `script/lint.sh` across all 11 modules
- `script/test.sh -covermode atomic ./...` across all 11 modules
- `git diff --check`

The local Go configuration has CGO disabled, so a race build was not available;
the full non-race suite passed.

AI assistance: OpenAI Codex helped draft the API bindings, tests, generated-file
workflow, and PR description. I reviewed the rendered diff and validation
evidence, understand every submitted line, and take responsibility for the
contribution and review follow-up.


gmlewis (approved):
Thank you, @Bortlesboat!
LGTM.

github/pulls.go
```diff
@@ -136,10 +136,20 @@ type PullRequestStack struct {
...
+	// NodeID is the global node ID of the stack.
+	NodeID *string `json:"node_id,omitempty"`
+	// URL is the API URL of the stack.
+	URL *string `json:"url,omitempty"`
+	// Open reports whether the stack contains any open pull requests.
+	Open *bool `json:"open,omitempty"`
+	// CreatedAt is the time the stack was created.
+	CreatedAt *Timestamp `json:"created_at,omitempty"`
+	// PullRequests contains the pull requests in the stack, from bottom to top.
+	PullRequests []*PullRequest `json:"pull_requests,omitempty"`
```
Not-Dhananjay-Mishra:
```go
// PullRequest represents a GitHub pull request on a repository.
type PullRequest struct {
	ID                *int64     `json:"id,omitempty"`
	Number            *int       `json:"number,omitempty"`
    ...
	Head  *PullRequestBranch `json:"head,omitempty"`
	Base  *PullRequestBranch `json:"base,omitempty"`
	Stack *PullRequestStack  `json:"stack,omitempty"`
}
```

`PullRequestStack` used with `PullRequest` struct, but the schema of`PullRequest.Stack`([docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request)) differs from the schema used by the [stacked pull request endpoints](https://docs.github.com/en/rest/pulls/stacks?apiVersion=2022-11-28). I think we should introduce a new struct instead of reusing `PullRequestStack`.

github/pulls_stacks_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+func TestCreatePullRequestStackRequest_marshal(t *testing.T) {
+	t.Parallel()
+	testJSONMarshal(t, &CreatePullRequestStackRequest{PullRequests: []int{101, 102}}, `{"pull_requests":[101,102]}`)
+}
+
+func TestAddPullRequestsToStackRequest_marshal(t *testing.T) {
+	t.Parallel()
+	testJSONMarshal(t, &AddPullRequestsToStackRequest{PullRequests: []int{103}}, `{"pull_requests":[103]}`)
+}
```
Not-Dhananjay-Mishra: These test can be removed, because we only write `_marshal`/`_unmarshal` tests for structs that implement custom `Marshal`/`Unmarshal` methods.

github/pulls_stacks_test.go
```diff
@@ -0,0 +1,314 @@
...
+	client, mux, _ := setup(t)
+
+	mux.HandleFunc("/repos/o/r/stacks", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testFormValues(t, r, values{
+			"pull_request": "101",
+			"page":         "2",
+			"per_page":     "50",
+		})
+		fmt.Fprintf(w, "[%s]", testPullRequestStackResponse())
```
Not-Dhananjay-Mishra:
```suggestion
		fmt.Fprintf(w, "[%v]", testPullRequestStackResponse())
```

github/pulls.go
Bortlesboat: Split the endpoint resource into `PullRequestStackDetails`; `PullRequest.Stack` remains the existing membership-only `PullRequestStack`. All five endpoint methods and the generated iterator now return the new type.

github/pulls_stacks_test.go
Bortlesboat: Removed the standalone marshal/unmarshal tests. The endpoint tests still cover each request body through `testJSONBody` and each response through real decode assertions.

github/pulls_stacks_test.go
Bortlesboat: Changed the list response formatting to `%v`.

Bortlesboat:
All three review items are addressed in signed commit `295dd70`, and their threads are resolved. Local `script/test.sh -covermode atomic ./...`, `script/lint.sh`, `script/fmt.sh`, and `script/generate.sh --check` all pass. GitHub initially held the new `tests`, `linter`, and workflow-security runs at `action_required`; a maintainer approved them, and the full current upstream matrix is now green.

github/pulls_stacks.go
```diff
@@ -0,0 +1,164 @@
...
+	URL *string `json:"url,omitempty"`
+	// Base is the branch the entire stack ultimately targets.
+	Base *PullRequestStackBase `json:"base"`
+	// Open reports whether the stack contains any open pull requests.
+	Open *bool `json:"open,omitempty"`
+	// CreatedAt is the time the stack was created.
+	CreatedAt *Timestamp `json:"created_at,omitempty"`
+	// PullRequests contains the pull requests in the stack, from bottom to top.
+	PullRequests []*PullRequest `json:"pull_requests,omitempty"`
+}
```
Not-Dhananjay-Mishra, line 159:
```suggestion
type PullRequestStackDetails struct {
	// ID is the ID of the stack.
	ID int64 `json:"id"`
	// Number is the number of the stack.
	Number int `json:"number"`
	// NodeID is the global node ID of the stack.
	NodeID string `json:"node_id"`
	// URL is the API URL of the stack.
	URL string `json:"url"`
	// Base is the branch the entire stack ultimately targets.
	Base *PullRequestStackBase `json:"base"`
	// Open reports whether the stack contains any open pull requests.
	Open bool `json:"open"`
	// CreatedAt is the time the stack was created.
	CreatedAt Timestamp `json:"created_at"`
	// PullRequests contains the pull requests in the stack, from bottom to top.
	PullRequests []*PullRequest `json:"pull_requests"`
}
```
All fields are required according to response schema and we don't use `,omitempty` with required field [CONTRIBUTING.md](https://github.com/google/go-github?tab=contributing-ov-file#response-bodies)
```json
    "title": "Pull Request Stack Minimal",
    "type": "object",
    "required": [
      "id",
      "number",
      "node_id",
      "url",
      "base",
      "open",
      "created_at",
      "pull_requests"
    ],
```

github/pulls_stacks.go
```diff
@@ -0,0 +1,164 @@
...
+	// ID is the ID of the stack.
+	ID *int64 `json:"id,omitempty"`
+	// Number is the number of the stack.
+	Number *int `json:"number,omitempty"`
+	// NodeID is the global node ID of the stack.
+	NodeID *string `json:"node_id,omitempty"`
+	// URL is the API URL of the stack.
+	URL *string `json:"url,omitempty"`
+	// Base is the branch the entire stack ultimately targets.
+	Base *PullRequestStackBase `json:"base"`
```
Not-Dhananjay-Mishra: `PullRequestStackBase` has one extra field `sha` which is not present in response schema. so let's implement a new struct instead.
```go
// PullRequestStackBase represents the base of a stacked pull request's stack:
// the branch the entire stack ultimately targets.
type PullRequestStackBase struct {
	Ref string `json:"ref"`
	SHA string `json:"sha"`
}
```
response schema 
```json
      "base": {
        "type": "object",
        "required": [
          "ref"
        ],
        "properties": {
          "ref": {
            "type": "string"
          }
        }
      }
```

github/pulls_stacks.go
```diff
@@ -0,0 +1,164 @@
...
+	// PullRequests contains the pull requests in the stack, from bottom to top.
+	PullRequests []*PullRequest `json:"pull_requests,omitempty"`
+}
+
+// ListStacks lists pull request stacks in a repository.
+//
+// GitHub API docs: https://docs.github.com/rest/pulls/stacks?apiVersion=2022-11-28#list-pull-request-stacks
+//
+//meta:operation GET /repos/{owner}/{repo}/stacks
+func (s *PullRequestsService) ListStacks(ctx context.Context, owner, repo string, opts *PullRequestListStacksOptions) ([]*PullRequestStackDetails, *Response, error) {
```
Not-Dhananjay-Mishra: Response schema of [List pull request stacks](https://docs.github.com/en/rest/pulls/stacks?apiVersion=2022-11-28#list-pull-request-stacks) and [Create a pull request stack](https://docs.github.com/en/rest/pulls/stacks?apiVersion=2022-11-28#create-a-pull-request-stack) ... differ slightly.
`pull_requests` object in the [List pull request stacks](https://docs.github.com/en/rest/pulls/stacks?apiVersion=2022-11-28#list-pull-request-stacks) response contains fewer fields than the response from the other endpoint. So we can't use `[]*PullRequestStackDetails` here.

Bortlesboat:
Thanks, both differences check out against the OpenAPI description. Pushed 5ad53ca.

`base` is `{ref}` only on all five stack endpoints, so it no longer reuses
`PullRequestStackBase` — that type stays as-is for `PullRequest.Stack`, where
`base` really is `{ref, sha}`. The stack endpoints now use a new
`PullRequestStackRef`.

On the list endpoint: the difference is in `pull_requests`. List items are
`{number, state, draft, merged_at, head{ref, sha}}`, while create/get/add/unstack
return `Pull Request Stack Pull Request`, which is `Pull Request Minimal`
(`id`, `number`, `url`, `head`/`base` with a nested `repo`) plus `node_id`,
`title`, `state`, `merged_at`, `draft`, `html_url`, and `user`. So `ListStacks`
now returns `[]*PullRequestStackMinimal` and the other four return
`*PullRequestStackDetails`. The `head`/`base` shapes differ too (list has no
`repo`), hence the separate branch types.

Also dropped `omitempty` and the pointers on the required fields per your other
comment. The old test fixture was sending the list-shaped `pull_requests` for
every endpoint, so there are two fixtures now.

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+// PullRequestsService.ListStacks method.
+type PullRequestListStacksOptions struct {
+	// PullRequest filters stacks to the stack containing this pull request number.
+	PullRequest int `url:"pull_request,omitempty"`
+
+	ListOptions
+}
+
+// CreatePullRequestStackRequest represents a request to create a pull request stack.
+type CreatePullRequestStackRequest struct {
```
Not-Dhananjay-Mishra, line 23:
```suggestion
type PullRequestCreateStackRequest struct {
```
I believe this would be better name

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+	// PullRequests is an ordered list of pull request numbers from the bottom of the stack to the top.
+	PullRequests []int `json:"pull_requests"`
+}
+
+// AddPullRequestsToStackRequest represents a request to append pull requests to a stack.
+type AddPullRequestsToStackRequest struct {
```
Not-Dhananjay-Mishra, line 29:
```suggestion
type PullRequestsAddToStackRequest struct {
```

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+	Open bool `json:"open"`
+	// CreatedAt is the time the stack was created.
+	CreatedAt Timestamp `json:"created_at"`
+	// PullRequests contains the pull requests in the stack, from bottom to top.
+	PullRequests []*PullRequestStackPullRequest `json:"pull_requests"`
+}
+
+// PullRequestStackPullRequest represents a pull request in a stack returned by
+// PullRequestsService.CreateStack, GetStack, AddToStack, and Unstack.
+type PullRequestStackPullRequest struct {
```
Not-Dhananjay-Mishra: How is `StackPullRequest`? `PullRequestStackPullRequest` feels repetitive.
```suggestion
type StackPullRequest struct {
```

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+	Open bool `json:"open"`
+	// CreatedAt is the time the stack was created.
+	CreatedAt Timestamp `json:"created_at"`
+	// PullRequests contains the pull requests in the stack, from bottom to top.
+	PullRequests []*PullRequestStackMinimalPullRequest `json:"pull_requests"`
+}
+
+// PullRequestStackMinimalPullRequest represents a pull request in a stack
+// returned by PullRequestsService.ListStacks.
+type PullRequestStackMinimalPullRequest struct {
```
Not-Dhananjay-Mishra: How is `StackMinimalPullRequest`? `PullRequestStackMinimalPullRequest` feels repetitive.
```suggestion
type StackMinimalPullRequest struct {
```

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+	Draft bool `json:"draft"`
+	// MergedAt is the time the pull request was merged, or nil if it is unmerged.
+	MergedAt *Timestamp `json:"merged_at"`
+	// Head is the branch the pull request merges from.
+	Head *PullRequestStackMinimalBranch `json:"head"`
+}
+
+// PullRequestStackMinimalBranch represents the head branch of a pull request
+// returned by PullRequestsService.ListStacks.
+type PullRequestStackMinimalBranch struct {
```
Not-Dhananjay-Mishra:
```suggestion
type PullRequestStackHead struct {
```

Not-Dhananjay-Mishra (commented):
A few struct name suggestions.

Bortlesboat:
Took the stutter fixes in 7b58560, but kept the `PullRequestStack` prefix
rather than dropping to bare `Stack*`:

- `PullRequestStackPullRequest` -> `PullRequestStackEntry`
- `PullRequestStackMinimalPullRequest` -> `PullRequestStackMinimalEntry`
- `PullRequestStackMinimalBranch` -> `PullRequestStackMinimalHead`

Reason for not using `StackPullRequest`/`StackMinimalPullRequest`: every other
stack type in the package is prefixed (`PullRequestStack`, `PullRequestStackBase`,
`PullRequestStackDetails`, `PullRequestStackMinimal`, ...), so those two would be
the only unprefixed ones in the group. `Entry` drops the repetition without
splitting the naming. Happy to go bare if you'd still rather.

I went with `PullRequestStackMinimalHead` over `PullRequestStackHead` because the
detailed responses have a head too, in `PullRequestStackBranch` — an unqualified
`PullRequestStackHead` would be ambiguous about which one it is.

On the two request types, I'd lean toward leaving them: `Create<Resource>Request`
is what the package mostly does (19 types — `CreateIssueRequest`,
`CreateGistCommentRequest`, `CreateHostedRunnerRequest`, ...) against 4 with a
service prefix, and those 4 are prefixed by service name
(`ActionsCreateVariableRequest`, `SecretScanningCreateCustomPatternsRequest`). If
you do want the service-prefixed form here, `PullRequestsCreateStackRequest` and
`PullRequestsAddToStackRequest` would at least match each other — the suggestions
as written are singular for one and plural for the other.

github/pulls_stacks_test.go
```diff
@@ -0,0 +1,364 @@
...
+	client, mux, _ := setup(t)
+
+	mux.HandleFunc("/repos/o/r/stacks", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		testFormValues(t, r, values{
+			"pull_request": "101",
+			"page":         "2",
+			"per_page":     "50",
+		})
+		fmt.Fprintf(w, "[%v]", testPullRequestStackMinimalResponse())
```
alexandear, line 132: Please avoid helper functions like `testPullRequestStackMinimalResponse` and `testPullRequestStackDetails`. They make the test less readable.

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return stacks, resp, nil
+}
+
+// CreateStack creates a pull request stack from an ordered list of pull request numbers.
+//
+// GitHub API docs: https://docs.github.com/rest/pulls/stacks?apiVersion=2022-11-28#create-a-pull-request-stack
+//
+//meta:operation POST /repos/{owner}/{repo}/stacks
+func (s *PullRequestsService) CreateStack(ctx context.Context, owner, repo string, body CreatePullRequestStackRequest) (*PullRequestStackDetails, *Response, error) {
```
alexandear, line 192: Why not?
```suggestion
func (s *PullRequestsService) CreateStack(ctx context.Context, owner, repo string, body CreatePullRequestStackRequest) (*PullRequestStack, *Response, error) {
```

Bortlesboat: Because `PullRequestStack` is a different resource — it's the stack *membership*
summary embedded in `PullRequest.Stack`, not the stack itself.

Comparing the two schemas:

|  | `PullRequest.stack` | stack endpoints |
|---|---|---|
| props | `base`, `id`, `number`, `position`, `size` | `base`, `id`, `number`, `node_id`, `url`, `open`, `created_at`, `pull_requests` |
| required | `base` | all eight |
| `base` | `{ref, sha}` | `{ref}` |

Only `base`, `id` and `number` overlap. `size`/`position` exist only on the
embedded one, `node_id`/`url`/`open`/`created_at`/`pull_requests` only on the
endpoint response, and even `base` differs — the embedded one carries a `sha`,
the endpoint one doesn't.

This is also what @Not-Dhananjay-Mishra asked for earlier in this PR
(https://github.com/google/go-github/pull/4436#discussion_r3733143386): the
first version did reuse `PullRequestStack`, and the request was to split it
precisely because the schemas differ.

If you two would rather converge on a different name than
`PullRequestStackDetails` I'm happy to rename — but I don't think it can be the
same type as `PullRequestStack`.

Bortlesboat:
Refreshed against current master in 064c6e70. The two request types now use the suggested names, and each endpoint test contains its response and expected-value fixtures directly. Accessors and iterators are regenerated.

Formatting, generated-file freshness, all 11 modules' non-race tests, and the custom lint checks pass. The focused stack endpoint tests also pass after replacing the copied pointer helpers with the current Go idiom. Local CGO remains disabled, so race coverage is left to CI.

The separate endpoint-resource and embedded pull-request membership types are preserved for the schema distinction discussed above.

gmlewis (approved):

github/pulls_stacks.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// PullRequestCreateStackRequest represents a request to create a pull request stack.
+type PullRequestCreateStackRequest struct {
+	// PullRequests is an ordered list of pull request numbers from the bottom of the stack to the top.
+	PullRequests []int `json:"pull_requests"`
+}
+
+// PullRequestsAddToStackRequest represents a request to append pull requests to a stack.
+type PullRequestsAddToStackRequest struct {
```
alexandear: For consistency:

```suggestion
type PullRequestAddToStackRequest struct {
```

Bortlesboat: > For consistency:

Changed to `PullRequestAddToStackRequest` in fd025036 and regenerated the accessor and its tests. Both request types now use singular `PullRequest`. Formatting, generation freshness, all 11 modules' non-race tests, and the custom lint checks pass. Native CGO remains disabled; local generator builds used `-buildvcs=false` because Go's VCS discovery selected the parent home directory for this linked worktree.

alexandear (approved):

gmlewis:
Thank you, @Bortlesboat, @Not-Dhananjay-Mishra , and @alexandear!
Merging.

PR #4438, opened 2026-08-06
refactor!: Rename `EditMilestone` to `UpdateMilestone`, split `Milestone` request bodies into `CreateMilestoneRequest` and `UpdateMilestoneRequest`, and pass by value

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for the milestone endpoints on `IssuesService`.

`CreateMilestone` and `EditMilestone` reused the 16-field `Milestone` **response** type as their request bodies, but per the [docs](https://docs.github.com/en/rest/issues/milestones?apiVersion=2022-11-28) the endpoints accept only four parameters — the other twelve fields (`url`, `number`, `creator`, `open_issues`, `created_at`, …) are server-generated. And the two schemas differ:

| Field | Create (POST) | Update (PATCH) |
|---|---|---|
| `title` | **required** | optional |
| `state` | optional | optional |
| `description` | optional | optional |
| `due_on` | optional | optional |

So this splits the request body into two dedicated types, following the same approach as #4382, #4401 and the `Issues`-service splits in #4396/#4400:

- **`CreateMilestoneRequest`** — `Title` is a non-pointer `string` (no `omitempty`) since it's required; the rest stay pointers.
- **`UpdateMilestoneRequest`** — every field is optional, so all stay pointers.

`EditMilestone` is renamed to `UpdateMilestone` to match the docs operation name, [Update a milestone](https://docs.github.com/en/rest/issues/milestones?apiVersion=2022-11-28#update-a-milestone) (same convention as `EditLabel`→`UpdateLabel` in #4400). Both methods now take their body by value, and `Milestone` is removed from the `body-allowed-pointer-types` allowlist. The `Milestone` response type itself is unchanged.

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (both methods and the eight generated accessors at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entry).

Updates #3644

BREAKING CHANGE: IssuesService.CreateMilestone now takes a new CreateMilestoneRequest (with non-pointer Title) by value, and IssuesService.EditMilestone is renamed to UpdateMilestone and takes a new UpdateMilestoneRequest by value, instead of *Milestone.

cc @jvm986 — this one is in the `Issues` service, which you'd been working through; since it's been a couple of weeks I picked up the next item, but happy to coordinate again whenever you resume.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -232,7 +232,6 @@ linters:
             - Key
             - LockIssueOptions
             - MaintenanceOptions
-            - Milestone
             - Organization
             - PagesUpdate
             - PagesUpdateWithoutCNAME
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -11326,6 +11326,38 @@ func (c *CreateJITConfigRequest) GetWorkFolder() string {
 	return *c.WorkFolder
 }
 
+// GetDescription returns the Description field if it's non-nil, zero value otherwise.
+func (c *CreateMilestoneRequest) GetDescription() string {
+	if c == nil || c.Description == nil {
+		return ""
+	}
+	return *c.Description
+}
+
+// GetDueOn returns the DueOn field if it's non-nil, zero value otherwise.
+func (c *CreateMilestoneRequest) GetDueOn() Timestamp {
+	if c == nil || c.DueOn == nil {
+		return Timestamp{}
+	}
+	return *c.DueOn
+}
+
+// GetState returns the State field if it's non-nil, zero value otherwise.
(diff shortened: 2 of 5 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

gmlewis:
Closing and reopening to force all the GitHub Actions workflows to fire properly after today's GitHub outage.

jvm986 (approved):
Great work! Thanks for taking care!

PR #4446, opened 2026-08-12
refactor!: Update teams create & update patterns

stevehipwell:
BREAKING CHANGE: `CreateTeam` and `UpdateTeamBy*` now take new body params by value.

This PR updates the teams create and update patterns to use the new request naming with seperate requests for create and update and adds support for linking parent teams by slug.

On the create path we've added the new `ParentTeamSlug` field and removed the unsupported `LDAPDN` field. On the update path we've removed added the new `ParentTeamSlug` field and a `RemoveParentTeam` field to replace the boolean in the method signature, as well as making `Name` optional and removing the unsupported `Maintainers`, `RepoNames` & `LDAPDN` fields.

### DIFF
--- a/github/teams.go
+++ b/github/teams.go
@@ -267,6 +267,7 @@ func (r UpdateTeamRequest) String() string {
 	return Stringify(r)
 }
 
+// MarshalJSON implements the json.Marshaler interface.
 func (r UpdateTeamRequest) MarshalJSON() ([]byte, error) {
 	type alias UpdateTeamRequest
 	if !r.RemoveParentTeam {
--- a/github/teams.go
+++ b/github/teams.go
@@ -7,6 +7,7 @@ package github
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 )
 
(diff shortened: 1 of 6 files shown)


gmlewis (approved):
Thank you, @stevehipwell!
LGTM.

cc: @alexandear - @Not-Dhananjay-Mishra

gmlewis:
@stevehipwell - please see if you can improve the codecov results for this PR or let me know if you need assistance.

github/teams.go
```diff
@@ -218,51 +226,85 @@ func (s *TeamsService) CreateTeam(ctx context.Context, org string, body NewTeam)
...
+	return updateTeamRequestNoParent{
+		Name:                r.Name,
+		Description:         r.Description,
+		Privacy:             r.Privacy,
+		NotificationSetting: r.NotificationSetting,
+		Permission:          r.Permission,
+		ParentTeamID:        nil,
+		ParentTeamSlug:      nil,
 	}
 }
```
alexandear: I propose to use a custom `MarshalJSON` method instead of the current approach with `body`.

```go
func (r UpdateTeamRequest) MarshalJSON() ([]byte, error) {
	type alias UpdateTeamRequest
	if !r.RemoveParentTeam {
		return json.Marshal(alias(r))
	}
	return json.Marshal(&struct {
		alias
		ParentTeamID   *int64  `json:"parent_team_id"`
		ParentTeamSlug *string `json:"parent_team_slug"`
	}{
		alias: alias(r),
	})
}

// ...

func (s *TeamsService) UpdateTeamByID(ctx context.Context, orgID, teamID int64, body UpdateTeamRequest) (*Team, *Response, error) 
{
	u := fmt.Sprintf("organizations/%v/team/%v", orgID, teamID)
	req, err := s.client.NewRequest(ctx, "PATCH", u, body)
	if err != nil {
		return nil, nil, err
	}
```

It's an established pattern across the project.

stevehipwell: @alexandear the `body()` pattern did feel a bit ugly, but it was a quick fix and an improvement on the existing code. But on reflection the `MarshalJSON()` pattern would be much better and easier to understand. The one issue is the `testJSONBody()` function which will now need to support passing in `go-cmp` options so we can ignore the non-marshaled fields.

stevehipwell:
> @stevehipwell - please see if you can improve the codecov results for this PR or let me know if you need assistance.

@gmlewis that was an oversight on my behalf as it looks like non-exported types don't get stringify tests generated, however with the change suggested by @alexandear that whole type will be removed anyway.

alexandear (approved):

gmlewis (approved):
LGTM.
Merging.

PR #4455, opened 2026-08-17
fix: Keep the `public` toggle on repository visibility rules

yordis:
`RepositoryVisibilityRuleParameters` models only `internal` and `private`, but the repository policy API has a third toggle, `public`. Because `encoding/json` discards unknown keys, the flag is lost on read, and since `PUT /orgs/{org}/rulesets/{ruleset_id}` is a full replace, writing back a ruleset the SDK just read silently resets public visibility to `false`.

Found while adding the `repository` ruleset target to the Terraform provider: integrations/terraform-provider-github#3598.

## Verified against the live API

Against a real `target: repository` ruleset on a github.com organization on the Team plan, using a program that fetches the same ruleset twice: once over plain `net/http` to capture the wire bytes, and once through `Organizations.GetRepositoryRuleset`. Org name and ruleset ID redacted.

On `v90.0.0`:

```
1. raw HTTP body      {"public":false,"internal":true,"private":true}
2. after SDK decode   {"internal":true,"private":true}
3. struct value       {Internal:true Private:true}
```

The server sends the key and the SDK drops it. Since `UpdateRepositoryRuleset` is a full replace, a read-modify-write built from that value omits `public` and the API resolves it to `false`.

With this patch applied:

```
1. raw HTTP body      {"public":false,"internal":true,"private":true}
2. after SDK decode   {"internal":true,"private":true,"public":false}
3. struct value       {Internal:true Private:true Public:false}
```

Then a write round trip through `Organizations.UpdateRepositoryRuleset`, toggling only `Public` on the decoded struct:

```
--- write phase, original public=false ---
PUT public=true  -> raw GET {"public":true,"internal":true,"private":true}
PUT public=false -> raw GET {"public":false,"internal":true,"private":true}
```

The API accepts the field on write and echoes it back, and the "Restrict visibility" section of the org policy UI showed Public checked while it was `true`. After restoring, the full ruleset body diffed equal to a backup taken before the run, ignoring `updated_at`.

## Verified in this repo

- `go build ./...`
- `go test ./... -count=1`
- `script/fmt.sh` and `script/lint.sh`, including the generated-files check

`github-accessors.go` and `github-accessors_test.go` are regenerated output.

Test coverage added for the round trip: the `repository_visibility` cases in `TestRepositoryRule` now carry `public: true` in both the marshal and unmarshal tables, which fails without the struct field.

For context on why this was missed: `repository_visibility` does not appear anywhere in `github/rest-api-description` (zero matches in both `api.github.com.json` and `ghec.json` as of today), so there is no generated schema to check the field list against. The name and JSON shape used here come from the live responses above.


alexandear:
Please take a look https://github.com/google/go-github/blob/master/CONTRIBUTING.md#assisted-contributions

yordis:
@alexandear what about it?

gmlewis (approved):
Thank you, @yordis!
LGTM.

yordis:
@gmlewis aye, let me know, as soon as this is merged, I will follow up on fixing the terraform provider, just so my paranoia is checked 😅

stevehipwell (approved):
LGTM

alexandear:
> @alexandear what about it?

Sorry, I should have been more specific. I was pointing at the "Assisted contributions" section:

> Tools that help draft or generate code are allowed, but contributors remain responsible for understanding and testing every line they submit. If an AI/LLM tool helped with code, tests, or prose, briefly disclose what it did in the PR description. Review the rendered diff and description yourself before requesting review, remove irrelevant generated changes, and keep review comments specific and actionable.

The PR description reads as AI-assisted to me, and this is unnecessary:

image

If a tool helped here, could you add a oneline note about it? Not blocking the review - just keeping us consistent with the guideline.

alexandear (approved):

gmlewis:
Thank you, @yordis, @stevehipwell, and @alexandear!
Merging.

@yordis - you can address @alexandear's comments after the merge.

PR #4460, opened 2026-08-19
feat: Add Copilot 3rd-party agent metrics

### LINKED ISSUE #4459: Add Copilot totals_by_3rd_party_agent metrics
GitHub added an optional `totals_by_3rd_party_agent` array to Copilot usage metrics reports (enterprise, organization, and per-user, 1-day and 28-day):

https://github.blog/changelog/2026-08-07-copilot-usage-metrics-api-adds-agent-app-activity/

Docs: https://docs.github.com/en/copilot/reference/copilot-usage-metrics/copilot-usage-metrics#agent-apps-metrics-fields

This field is not modeled in `CopilotDailyMetrics`, `CopilotUserDailyMetrics`, or `CopilotUserPeriodicMetrics`, so downloaded reports drop the agent-app breakdown.

I'd like to add a `CopilotMetricsThirdPartyAgent` type and wire it up on those structs, with tests. `session_count` would be a pointer so per-user reports can omit it.

Please assign this to me if that sounds good.

Tens1des:
Adds `totals_by_3rd_party_agent` to Copilot usage metrics report structs.

GitHub changelog: https://github.blog/changelog/2026-08-07-copilot-usage-metrics-api-adds-agent-app-activity/

- New `CopilotMetricsThirdPartyAgent` type (`agent_name`, `agent_id`, `user_initiated_interaction_count`, optional `session_count`)
- Field wired on `CopilotDailyMetrics`, `CopilotUserDailyMetrics`, and `CopilotUserPeriodicMetrics`
- Tests for daily and user-daily downloads; accessors regenerated via `script/generate.sh`

Fixes #4459

### DIFF
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -1008,6 +1008,22 @@ type CopilotMetricsCLI struct {
 	LastKnownCLIVersion *CopilotMetricsCLIVersion    `json:"last_known_cli_version,omitempty"`
 }
 
+// CopilotMetricsThirdPartyAgent represents per-agent app totals in a Copilot usage metrics report.
+//
+// AgentID is the stable grouping key. AgentName is for display and can change.
+// UserInitiatedInteractionCount is agent-app job starts; it is not the same as
+// the top-level user_initiated_interaction_count on the parent metrics object.
+// SessionCount is included on aggregated enterprise and organization reports
+// and omitted from per-user reports.
+//
+// GitHub API docs: https://docs.github.com/en/copilot/reference/copilot-usage-metrics/copilot-usage-metrics#agent-apps-metrics-fields
+type CopilotMetricsThirdPartyAgent struct {
+	AgentName                     string `json:"agent_name"`
+	AgentID                       string `json:"agent_id"`
+	UserInitiatedInteractionCount *int   `json:"user_initiated_interaction_count,omitempty"`
+	SessionCount                  *int   `json:"session_count,omitempty"`
+}
+
 // CopilotDailyMetrics represents the payload downloaded from a 1-day Copilot usage metrics report.
 type CopilotDailyMetrics struct {
 	Day                                 string  `json:"day"`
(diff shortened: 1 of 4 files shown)


gmlewis (approved):
Thank you, @Tens1des!
LGTM.

github/copilot.go
```diff
@@ -1008,6 +1008,22 @@ type CopilotMetricsCLI struct {
...
+// the top-level user_initiated_interaction_count on the parent metrics object.
+// SessionCount is included on aggregated enterprise and organization reports
+// and omitted from per-user reports.
+//
+// GitHub API docs: https://docs.github.com/en/copilot/reference/copilot-usage-metrics/copilot-usage-metrics#agent-apps-metrics-fields
+type CopilotMetricsThirdPartyAgent struct {
+	AgentName                     string `json:"agent_name"`
+	AgentID                       string `json:"agent_id"`
+	UserInitiatedInteractionCount *int   `json:"user_initiated_interaction_count,omitempty"`
+	SessionCount                  *int   `json:"session_count,omitempty"`
```
stevehipwell, line 1024: I thought these were unmarshal only types so I'm unsure why they need `omitempty`?

stevehipwell (changes requested):

github/copilot.go
Tens1des: Good question — `omitempty` only affects marshaling, so it's not required for Unmarshal itself.

I kept it to match the existing Copilot metrics types and the CONTRIBUTING guidance for response bodies (optional pointer fields use `omitempty`), and so nil fields round-trip cleanly via `testJSONMarshal` / user re-marshaling without emitting `null`.

Happy to drop it if you'd prefer these tags to be unmarshal-only.

gmlewis: > I thought these were unmarshal only types so I'm unsure why they need `omitempty`?

Right - it is common practice in this repo to use pointers and `omitempty` for unmarshal-only types, especially when structs are used for many different endpoints. Then, when the GitHub v3 API servers do not provide fields, it is easy to see when they are not included.

I think it is fine to run with this without modification.

stevehipwell: @gmlewis I'm not sure I follow? The official docs for JSON struct tags say `omitempty` & `omitzero` are ignored during unmarshalling, and this type is unmarshal only. I also don't think the struct tag should be conflated with the use of a pointer.

gmlewis: > @gmlewis I'm not sure I follow? The official docs for JSON struct tags say `omitempty` & `omitzero` are ignored during unmarshalling, and this type is unmarshal only. I also don't think the struct tag should be conflated with the use of a pointer.

Ah, I see exactly what you are saying, and you are right, @stevehipwell.

I think the biggest challenge in this repo is to determine which fields are unmarshal-only, so I wrote #4478 to analyze the situation.

I still like the point made by @Tens1des regarding:
> and so `nil` fields round-trip cleanly via `testJSONMarshal` / `user` re-marshaling without emitting `null`.

So I think we should go ahead with this PR and merge as-is.

gmlewis: @alexandear or @Not-Dhananjay-Mishra - do you have any opinions in this matter?

alexandear: We can remove the `omitempty` in another PR.

PR #4462, opened 2026-08-20
refactor!: Rename `ImpersonateUserOptions` to `CreateUserImpersonationRequest` and pass by value

jvm986:
Continues the request-body-by-value work in #3644, this time for `AdminService.CreateUserImpersonation`.
The type had an `Options` suffix despite being a POST body, and was passed by pointer even though its only parameter, `scopes`, is [required](https://docs.github.com/en/enterprise-server@3.21/rest/enterprise-admin/users?apiVersion=2022-11-28#create-an-impersonation-oauth-token). It's renamed after the operation, passed by value, and `Scopes` drops `omitempty`. Removed from both `paramcheck` allowlists (`body-allowed-pointer-types` and `body-allowed-wrong-names`).

BREAKING CHANGE: `AdminService.CreateUserImpersonation` now takes a new `CreateUserImpersonationRequest` (with non-pointer Scopes) by value instead of `*ImpersonateUserOptions`.

cc @JamBalaya56562  @ManavSharma142 @stevehipwell

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -223,7 +223,6 @@ linters:
             - EncryptedSecret
             - EnterpriseSecurityAnalysisSettings
             - Hook
-            - ImpersonateUserOptions
             - Import
             - InstallationTokenListRepoOptions
             - InstallationTokenOptions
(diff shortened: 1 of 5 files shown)


stevehipwell (approved):
LGTM

gmlewis (approved):
Thank you, @jvm986 and @stevehipwell!
LGTM.
Merging.

PR #4463, opened 2026-08-20
chore: Bump golang-ci-lint version to v2.13.1

gmlewis:
This should fix panics found in golang-ci-lint such as the ones here:
https://github.com/google/go-github/actions/runs/32389013094


gmlewis:
Relates to: #4462.

PR #4475, opened 2026-08-21
refactor!: Pass `PendingDeploymentsRequest` and `ReviewCustomDeploymentProtectionRuleRequest` by value

jvm986:
Continues the request-body-by-value work in #3644, this time for the two deployment-review endpoints on `ActionsService`.

`PendingDeploymentsRequest` and `ReviewCustomDeploymentProtectionRuleRequest` were already correctly shaped — every field is a required non-pointer with no `omitempty`, and both already use the `Request` suffix — so the pointer parameter was the only defect. Both are now passed by value and removed from the `body-allowed-pointer-types` allowlist; no type or field changes.

BREAKING CHANGE: `ActionsService.PendingDeployments` and `ActionsService.ReviewCustomDeploymentProtectionRule` now take `PendingDeploymentsRequest` and `ReviewCustomDeploymentProtectionRuleRequest` by value instead of by pointer.

cc @JamBalaya56562 @ManavSharma142 @stevehipwell

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -233,7 +233,6 @@ linters:
             - Organization
             - PagesUpdate
             - PagesUpdateWithoutCNAME
-            - PendingDeploymentsRequest
             - ProtectionRequest
             - PublishCodespaceOptions
             - PullRequestBranchUpdateOptions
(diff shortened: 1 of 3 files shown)


alexandear (approved):

gmlewis (approved):
Thank you, @jvm986 and @alexandear!
LGTM.
Merging.

PR #4348, opened 2026-07-02
feat!: Refactor dependabot secrets to pass request by value

stevehipwell:
BREAKING CHANGE: `DependabotService` methods involving secrets have new params and return values.

This PR updates the Dependabot secret update/create functions to use a value parameter (see https://github.com/google/go-github/issues/3644). This aligns Dependabot secrets with the changes made to actions secrets (https://github.com/google/go-github/pull/4335) & actions variables (https://github.com/google/go-github/pull/4346). It also aligns the names of the actions secret and variable request types.

PR #4346, opened 2026-07-01
feat!: Refactor actions variables to pass request by value

stevehipwell:
BREAKING CHANGE: `ActionsService` methods involving variables have new params and return values.

This PR updates the actions variable create & update functions to use a value parameter (see https://github.com/google/go-github/issues/3644). This aligns actions variables with the changes made to actions secrets in #4335.

PR #4476, opened 2026-08-21
chore: Remove stale `paramcheck` allowlist entries

jvm986:
Follow-up to #4346 and #4348.

Those PRs converted the actions-variable and dependabot-secret request bodies to pass by value, which were the only places `ActionsVariable` and `DependabotEncryptedSecret` were used as request bodies. Their entries in the `paramcheck` `body-allowed-pointer-types` allowlist became dead but weren't removed — #4348 dropped that type's `structfield` entries but not its `paramcheck` one.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -203,7 +203,6 @@ linters:
           # Body type names exempt from the "pass by value, not by pointer" rule.
           # TODO: fix and remove these exceptions.
           body-allowed-pointer-types:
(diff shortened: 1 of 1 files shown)


alexandear (approved):

gmlewis (approved):
Thank you, @jvm986 and @alexandear!
LGTM.
Merging.

PR #4477, opened 2026-08-21
refactor!: Split `Key` request bodies into `CreateDeployKeyRequest`, `CreateUserKeyRequest` and `CreateSSHSigningKeyRequest` and pass by value

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for the three key-creation endpoints that shared the `Key` **response** type as their request body.

Per the docs, each endpoint accepts only a small request schema — the other six `Key` fields (`id`, `url`, `verified`, `created_at`, `added_by`, `last_used`) are server-generated — and the schemas differ per resource:

| Method | Endpoint | `key` | `title` | `read_only` |
|---|---|---|---|---|
| `RepositoriesService.CreateKey` | [Create a deploy key](https://docs.github.com/en/rest/deploy-keys/deploy-keys?apiVersion=2022-11-28#create-a-deploy-key) | **required** | optional | optional |
| `UsersService.CreateKey` | [Create a public SSH key](https://docs.github.com/en/rest/users/keys?apiVersion=2022-11-28#create-a-public-ssh-key-for-the-authenticated-user) | **required** | optional | — |
| `UsersService.CreateSSHSigningKey` | [Create a SSH signing key](https://docs.github.com/en/rest/users/ssh-signing-keys?apiVersion=2022-11-28#create-a-ssh-signing-key-for-the-authenticated-user) | **required** | optional | — |

So this splits the body into three per-resource request types — `CreateDeployKeyRequest` (with `ReadOnly`), `CreateUserKeyRequest` and `CreateSSHSigningKeyRequest` — following the same approach as #4406/#4425/#4432/#4434/#4444. `Key` is a non-pointer `string` in all three since it's required everywhere. The methods keep their names (the `Create` verb already matches the docs, and renaming only the create methods would break consistency with the surrounding `Get`/`List`/`DeleteKey` families).

Also fixed in passing, since the signature change touches the exact lines: the failure case in `TestUsersService_CreateSSHSigningKey` had a pre-existing copy-paste bug — it called `Users.CreateKey` (a different method) with `methodName = "CreateKey"`; it now exercises `CreateSSHSigningKey` as intended.

Notes:

- The `Key` response type is unchanged — it stays the response for `ListKeys`/`GetKey`/both create methods, so its fields keep pointer semantics. `SSHSigningKey` is untouched.
- `test/integration/users_test.go` builds under the root module with `-tags integration`, so its `Users.CreateKey` call is updated too (verified with `go vet -tags integration`).

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (all three methods and the seven generated accessors at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entry).

Updates #3644

BREAKING CHANGE: RepositoriesService.CreateKey, UsersService.CreateKey and UsersService.CreateSSHSigningKey now take new CreateDeployKeyRequest, CreateUserKeyRequest and CreateSSHSigningKeyRequest (with non-pointer Key) by value instead of *Key.

cc @jvm986 — flagging for #3644 coordination; this covers the `Key` type (repos/users key endpoints), so it doesn't overlap with #4475.

### DIFF
--- a/github/repos_keys.go
+++ b/github/repos_keys.go
@@ -60,12 +60,19 @@ func (s *RepositoriesService) GetKey(ctx context.Context, owner, repo string, id
 	return key, resp, nil
 }
 
+// CreateDeployKeyRequest represents a request to create a deploy key.
+type CreateDeployKeyRequest struct {
+	Title    *string `json:"title,omitempty"`
+	Key      string  `json:"key"`
+	ReadOnly *bool   `json:"read_only,omitempty"`
+}
+
 // CreateKey adds a deploy key for a repository.
 //
 // GitHub API docs: https://docs.github.com/rest/deploy-keys/deploy-keys?apiVersion=2022-11-28#create-a-deploy-key
 //
 //meta:operation POST /repos/{owner}/{repo}/keys
-func (s *RepositoriesService) CreateKey(ctx context.Context, owner, repo string, body *Key) (*Key, *Response, error) {
+func (s *RepositoriesService) CreateKey(ctx context.Context, owner, repo string, body CreateDeployKeyRequest) (*Key, *Response, error) {
 	u := fmt.Sprintf("repos/%v/%v/keys", owner, repo)
 
 	req, err := s.client.NewRequest(ctx, "POST", u, body)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -227,7 +227,6 @@ linters:
             - InstallationTokenListRepoOptions
             - InstallationTokenOptions
             - IssueImportRequest
-            - Key
             - LockIssueOptions
             - MaintenanceOptions
             - Organization
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -22350,6 +22350,14 @@ func (k *Key) GetCreatedAt() Timestamp {
 	return *k.CreatedAt
 }
 
+// GetEnabled returns the Enabled field if it's non-nil, zero value otherwise.
+func (k *Key) GetEnabled() bool {
+	if k == nil || k.Enabled == nil {
+		return false
+	}
+	return *k.Enabled
+}
+
 // GetID returns the ID field if it's non-nil, zero value otherwise.
 func (k *Key) GetID() int64 {
 	if k == nil || k.ID == nil {
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -11414,6 +11414,30 @@ func (c *CreateCustomRepoRoleRequest) GetPermissions() []string {
 	return c.Permissions
 }
 
+// GetKey returns the Key field.
+func (c *CreateDeployKeyRequest) GetKey() string {
+	if c == nil {
+		return ""
+	}
+	return c.Key
+}
+
+// GetReadOnly returns the ReadOnly field if it's non-nil, zero value otherwise.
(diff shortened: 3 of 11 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/repos_keys.go
```diff
@@ -60,12 +60,19 @@ func (s *RepositoriesService) GetKey(ctx context.Context, owner, repo string, id
...
+	ReadOnly *bool   `json:"read_only,omitempty"`
+}
+
 // CreateKey adds a deploy key for a repository.
 //
 // GitHub API docs: https://docs.github.com/rest/deploy-keys/deploy-keys?apiVersion=2022-11-28#create-a-deploy-key
 //
 //meta:operation POST /repos/{owner}/{repo}/keys
-func (s *RepositoriesService) CreateKey(ctx context.Context, owner, repo string, body *Key) (*Key, *Response, error) {
+func (s *RepositoriesService) CreateKey(ctx context.Context, owner, repo string, body CreateDeployKeyRequest) (*Key, *Response, error) {
```
alexandear, line 75: The returned type `Key` lacks the `enabled` field:

```
{
  "title": "Deploy Key",
  "description": "An SSH key granting access to a single repository.",
  "type": "object",
  "properties": {
    "id": {
      "type": "integer"
    },
    "key": {
      "type": "string"
    },
    "url": {
      "type": "string"
    },
    "title": {
      "type": "string"
    },
    "verified": {
      "type": "boolean"
    },
    "created_at": {
      "type": "string"
    },
    "read_only": {
      "type": "boolean"
    },
    "added_by": {
      "type": [
        "string",
        "null"
      ]
    },
    "last_used": {
      "type": [
        "string",
        "null"
      ],
      "format": "date-time"
    },
    "enabled": {
      "type": "boolean"
    }
  },
  "required": [
    "id",
    "key",
    "url",
    "title",
    "verified",
    "created_at",
    "read_only"
  ]
}
```

I think we should extend the `Key` or add a new struct/

alexandear: Check also other methods that return the `Key` struct.

JamBalaya56562: Added in 188c6abb49c4 — `Enabled *bool` with `omitempty`.

I checked both schemas before choosing between your two options: `enabled` is deploy-key specific (the authenticated-user SSH key response schema doesn't have it), which matches how `Key` already carries other deploy-only fields like `added_by`. So extending the shared `Key` with an optional pointer seemed more consistent than introducing a separate response struct — splitting the response types would also change the return type of six methods, which felt out of scope for this request-side PR.

> Check also other methods that return the `Key` struct.

Done — all six (`Repositories.ListKeys`/`GetKey`/`CreateKey` and `Users.ListKeys`/`GetKey`/`CreateKey`) return the same `Key`, so the new field applies to all of them; for the user-key endpoints it simply stays `nil` since that schema doesn't include `enabled`.

gmlewis (approved):

gmlewis:
@alexandear - do you approve now?

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #4478, opened 2026-08-21
feat: Add tools/list-return-structs

gmlewis:
This tool directly answers the questions raised here:
https://github.com/google/go-github/pull/4460#discussion_r3822379041

This tool shows that there are 615 structs in this repo that are used strictly as return values and are never included as input parameters to `opts` or `body` values, and are therefore only unmarshaled in the non-test Go source files.

Out of these 615 return-only structs, 487 of them include at least one field that uses the `json` "omitempty" or "omitzero" option.

As correctly pointed out by @stevehipwell, the `omitempty` or `omitzero` in these 487 structs are unnecessary and can be removed.

Although, as @Tens1des pointed out, the `omitempty` is nice to have there when using `testJSONMarshal` so that a `null` is not provided in the output.

So I'm a bit torn on this issue and am leaning toward keeping the `omitempty`/`omitzero` for unmarshal-only structs.

### DIFF
--- a/tools/list-return-structs/main.go
+++ b/tools/list-return-structs/main.go
@@ -1,0 +1,382 @@
+// Copyright 2026 The go-github AUTHORS. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Command list-return-structs scans the Go sources of the `github` package and
+// reports the struct types that are used strictly as return values — that is,
+// structs that are only ever "unmarshaled" from the GitHub API server and are
+// never "marshaled" and sent back to it.
+//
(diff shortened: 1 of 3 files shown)


Not-Dhananjay-Mishra:
@gmlewis So, are we going to remove `omitempty`/`omitzero` from unmarshal-only structs? and use pointer only for optional field in unmarshal-only structs?

gmlewis:
> @gmlewis So, are we going to remove `omitempty`/`omitzero` from unmarshal-only structs? and use pointer only for optional field in unmarshal-only structs?

I haven't decided and haven't heard any great arguments for one versus the other.
It's fine with me to continue with the status quo, and I don't think we need a concerted effort to remove them.

So I guess the answer right now is "No".

PR #1363, opened 2020-01-03
Re-populate HTTP response body of ErrorResponse

haya14busa:
Fixes #1136

PR #4486, opened 2026-08-25
fix: Close the original response body on non-2xx responses

### LINKED ISSUE #4484: bug: Original response body is never closed
**go-github version:** v87.0.0 (also present on `master` at time of writing)
**Go version:** 1.26

## Summary

`CheckResponse` replaces `r.Body` with a buffered copy on every non-2xx response, *before* `bareDo` registers its deferred `resp.Body.Close()`. The deferred close therefore closes the in-memory replacement (`io.NopCloser`, a no-op) and the original body returned by `http.Client.Do` is never closed by anyone.

The unclosed body itself is symptomless: `CheckResponse` happens to read the original to EOF, so the connection is still reused, which is why this has gone unnoticed. It becomes observable when two conditions hold together:

1. **The `http.Client` has `Timeout` set.** `net/http` must then cancel the request at the deadline. For known stdlib transports it uses `context.WithDeadline` (no extra goroutine), but for any other `RoundTripper` it falls back to a legacy path (`setRequestCancel`) that wraps the body in `cancelTimerBody` and parks a watcher goroutine, which is stopped **only by `Close`**, not by read-to-EOF ([client.go#L985-L1003 in Go 1.26.5](https://github.com/golang/go/blob/go1.26.5/src/net/http/client.go#L985-L1003)).
2. **The transport is a non-stdlib `RoundTripper`.** This is the case for effectively every authenticated client: go-github's own `WithAuthToken` wraps the transport in a `roundTripperFunc` ([github.go#L581](https://github.com/google/go-github/blob/v87.0.0/github/github.go#L581)), and the common alternatives (`oauth2.Transport`, `ghinstallation`, `otelhttp.NewTransport`) are wrappers too.

With both in place, every non-2xx API response parks one `net/http.setRequestCancel.func4` goroutine for the full remainder of the client timeout.

The population is bounded (each goroutine self-terminates at the deadline), so this never accumulates into an unbounded leak, but any test suite using `goleak` (or similar) will flag these goroutines intermittently, and under a sustained error load a `Timeout`-sized window of goroutines stays resident (see Impact below).

## The relevant code

`bareDo` ([github.go#L1196-L1198 @ v87.0.0](https://github.com/google/go-github/blob/v87.0.0/github/github.go#L1196-L1198)):

```go
err = CheckResponse(resp)
if err != nil {
	defer resp.Body.Close()   // resp.Body is already the replacement here
```

`CheckResponse` ([github.go#L1711-L1722 @ v87.0.0](https://github.com/google/go-github/blob/v87.0.0/github/github.go#L1711-L1722)):

```go
data, err := io.ReadAll(io.LimitReader(r.Body, maxErrorBodySize))
...
// Re-populate error response body because GitHub error responses are often
// undocumented and inconsistent.
// Issue #1136, #540.
r.Body = io.NopCloser(bytes.NewBuffer(data))
```

The body swap is intentional (#1136, #540) so callers can re-read error payloads; the bug is only that the original `ReadCloser` is dropped without being closed.

## Reproduction

Self-contained, no credentials; `go run .` with the go.mod below.

```go
// go.mod:
//   module ghleak
//   go 1.25
//   require github.com/google/go-github/v87 v87.0.0
package main

import (
	"context"
	"fmt"
	"net/http"
	"net/http/httptest"
	"runtime"
	"strings"
	"time"

	"github.com/google/go-github/v87/github"
)

func cancelTimerGoroutines() int {
	buf := make([]byte, 1<<20)
	n := runtime.Stack(buf, true)
	return strings.Count(string(buf[:n]), "net/http.setRequestCancel.func4")
}

// wrapper stands in for any custom RoundTripper: oauth2.Transport,
// otelhttp.NewTransport, ghinstallation, etc. With a wrapped transport,
// net/http's Client.Timeout falls back to its legacy cancel path, which
// parks a goroutine per request until the body is closed or the timeout
// elapses (net/http.setRequestCancel).
type wrapper struct{ base http.RoundTripper }

func (w wrapper) RoundTrip(r *http.Request) (*http.Response, error) { return w.base.RoundTrip(r) }

func main() {
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if strings.Contains(r.URL.Path, "/repos/o/found") {
			w.Write([]byte(`{"id": 1, "name": "found"}`)) // 200
			return
		}
		w.WriteHeader(404)
		w.Write([]byte(`{"message": "Not Found"}`))
	}))
	defer srv.Close()

	httpClient := &http.Client{
		Timeout:   10 * time.Second,
		Transport: wrapper{http.DefaultTransport},
	}
	gh, err := github.NewClient(
		github.WithHTTPClient(httpClient),
		github.WithEnterpriseURLs(srv.URL, srv.URL),
	)
	if err != nil {
		panic(err)
	}
	ctx := context.Background()

	for range 5 {
		if _, _, err := gh.Repositories.Get(ctx, "o", "found"); err != nil {
			panic(err)
		}
	}
	time.Sleep(100 * time.Millisecond)
	fmt.Printf("after 5 responses with status 200: %d cancel-timer goroutines\n", cancelTimerGoroutines())

	for range 5 {
		if _, _, err := gh.Repositories.Get(ctx, "o", "missing"); err == nil {
			panic("expected 404 error")
		}
	}
	time.Sleep(100 * time.Millisecond)
	fmt.Printf("after 5 responses with status 404: %d cancel-timer goroutines\n", cancelTimerGoroutines())

	time.Sleep(11 * time.Second)
	fmt.Printf("after the 10s client timeout elapses: %d cancel-timer goroutines\n", cancelTimerGoroutines())
}
```

Output:

```
after 5 responses with status 200: 0 cancel-timer goroutines
after 5 responses with status 404: 5 cancel-timer goroutines
after the 10s client timeout elapses: 0 cancel-timer goroutines
```

One goroutine parked per error response; 2xx responses (closed properly by `Do`) park none; all of them self-terminate when the client timeout elapses.

## Impact

Each parked goroutine self-terminates when the client timeout elapses, so the steady-state population under sustained errors is bounded at roughly `error responses/sec x Client.Timeout`, at a few KB of stack each (plus the retained response/timer). For example, 10 error responses/sec with a 10s timeout holds ~100 goroutines resident (well under 1 MB); it takes extremes like 1,000/sec with an 80s timeout to reach ~80k goroutines / hundreds of MB. So in production this is mostly invisible; the practical cost is that `goleak.VerifyTestMain` (or similar leak detection code) in downstream test suites intermittently fails on `net/http.setRequestCancel.func4` whenever the process exits within `Timeout` of the last non-2xx response, which is how we found it.

Notably, the library's own pre-emptive rate-limit check (`checkRateLimitBeforeDo` / `checkSecondaryRateLimitBeforeDo`) already dampens the rate-limited flavor of this: once a primary/secondary limit is recorded, subsequent calls short-circuit with a synthetic response and no network request, so they don't leak. The responses that do leak are the ones that always hit the network: 404s, non-rate-limit 403s, and 5xx.

## How we hit it

Our integration suite runs `goleak.VerifyTestMain`. Tests that exercise GitHub error paths (4xx/5xx from a mock server) intermittently failed with batches of:

```
Goroutine NNN in state select, with net/http.setRequestCancel.func4 on top of the stack:
net/http.setRequestCancel.func4()
        .../src/net/http/client.go:407
```

whenever the process exited within `Timeout` seconds of the last non-2xx GitHub response.

## Suggested fix

Capture the original body before `CheckResponse` gets a chance to replace it:

```go
origBody := resp.Body
err = CheckResponse(resp)
if err != nil {
	defer origBody.Close()
	...
```

(or have `CheckResponse` close `r.Body` itself right after `io.ReadAll`, before reassigning it). `bareDoIgnoreRedirects`/other callers of `CheckResponse` may want the same treatment.

Disclaimer: I haven't actually tested the suggested fix resolves the issue but it should.

yavorl:
Fixes #4484

`CheckResponse` substitutes `r.Body` with a re-readable `NopCloser` copy on every
non-2xx response (#1363). Since #1772, the error-path `defer resp.Body.Close()` in
`bareDo` is registered after that substitution, so the deferred close releases the
copy and the network body is never closed. `CopilotService.fetchMetricsReport` and `RepositoriesService.downloadReleaseAssetFromURL` have the same pattern.

Consequences of the unclosed network body:

- With an `http.Client` that has `Timeout` set and a wrapped transport (which
  includes clients built via `WithAuthToken`), every non-2xx response parks one
  `net/http.setRequestCancel.func4` goroutine for the remainder of the timeout.
  Bounded, but it intermittently fails `goleak`-checked test suites downstream.
- Error bodies larger than `maxErrorBodySize` are only partially drained, so the
  connection is additionally lost. 1Mb seems generous so it can be a non-issue.

The fix captures the network body before calling `CheckResponse` and closes that
instead. The #1363 behavior (re-readable error bodies) is
unchanged: the substitute is left open for callers.

The regression test wraps the test client's transport to observe the network
body and asserts it is closed after an error response. It fails on `master` and
passes with this change; the full package test suite passes.

commit 1: a failing regression test
commit 2: the fix; capture the original body & close it
commit 3: similar regression tests for the other 2 call sites
commit 4: fix the 2 remaining callers & update the `CheckResponse()` godoc


github/github_test.go
```diff
@@ -2296,6 +2296,50 @@ func TestDo_httpError(t *testing.T) {
 	}
 }
 
+// closeRecorder flags when the response body handed back by the transport
+// is closed.
+type closeRecorder struct {
+	io.ReadCloser
+	closed *bool
+}
```
yavorl, line 2304: having this round tripper is also a precondition for the bug. Not a necessary precondition for the test because we're asserting that the underlying cause can't happen - Close() not being called.

gmlewis (approved):
Ah, I see what you are saying, @yavorl. 
I would like to cover all possible unclosed bodies, if possible, so that this issue can truly be recorded as solved. What are your thoughts as to what it would take to accomplish this?

LGTM.

yavorl:
> Ah, I see what you are saying, @yavorl. I would like to cover all possible unclosed bodies, if possible, so that this issue can truly be recorded as solved. What are your thoughts as to what it would take to accomplish this?
> 
> LGTM.
> 
> cc: @stevehipwell - @alexandear - @Not-Dhananjay-Mishra

The fix is the same, I have to see whether the test pattern translates to the other invocations. Would you like me to try it in this PR or a separate follow-up?

gmlewis:
Well, I'm fine either way. If you want to keep it separate in a new PR, that is fine with me too. Thank you, @yavorl!

yavorl:
Done it in the same PR. Added a comment as well to the godoc of `CheckResponse()`

gmlewis (approved):
Thank you, @yavorl!
LGTM.

github/github.go
```diff
@@ -1282,9 +1282,13 @@ func (c *Client) bareDo(caller *http.Client, req *http.Request) (*Response, erro
 		c.rateMu.Unlock()
 	}
 
+	// CheckResponse substitutes r.Body with a re-readable copy on error
+	// responses, so capture the network body first: it is the one that must
+	// be closed.
+	origBody := resp.Body
```
alexandear, line 1287: Could we centralize the fix to one place? I think it's possible to patch only `CheckResponse`.

alexandear (changes requested):

github/github.go
alexandear: The patch is something like this:
```
diff --git a/github/copilot.go b/github/copilot.go
index adbfae32..9dd24eb8 100644
--- a/github/copilot.go
+++ b/github/copilot.go
@@ -1212,12 +1212,8 @@ func (s *CopilotService) fetchMetricsReport(ctx context.Context, url string) (*h
 		return nil, nil, err
 	}
 
-	// CheckResponse substitutes resp.Body with a re-readable copy on error
-	// responses, so capture the original body first: it is the one that must
-	// be closed.
-	origBody := resp.Body
 	if err := CheckResponse(resp); err != nil {
-		_ = origBody.Close()
+		resp.Body.Close()
 	}
 
diff --git a/github/copilot_test.go b/github/copilot_test.go
index ee1f804a..4ddbab41 100644
--- a/github/copilot_test.go
+++ b/github/copilot_test.go
@@ -3212,8 +3212,8 @@ func TestCopilotService_DownloadDailyMetrics(t *testing.T) {
 	}
 }
 
-// CheckResponse substitutes resp.Body with a re-readable copy on error
-// responses; fetchMetricsReport must still close the original body it replaces.
+// CheckResponse closes the original resp.Body and substitutes a re-readable
+// copy on error responses; fetchMetricsReport must not leak the original body.
 func TestCopilotService_fetchMetricsReport_closesOriginalBodyOnErrorResponse(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
diff --git a/github/github.go b/github/github.go
index efb4c677..bfa0bd10 100644
--- a/github/github.go
+++ b/github/github.go
@@ -1282,13 +1282,9 @@ func (c *Client) bareDo(caller *http.Client, req *http.Request) (*Response, erro
 		c.rateMu.Unlock()
 	}
 
-	// CheckResponse substitutes r.Body with a re-readable copy on error
-	// responses, so capture the network body first: it is the one that must
-	// be closed.
-	origBody := resp.Body
 	err = CheckResponse(resp)
 	if err != nil {
-		defer origBody.Close()
+		defer resp.Body.Close()
 		// Special case for AcceptedErrors. If an AcceptedError
 		// has been encountered, the response's payload will be
 		// added to the AcceptedError and returned.
@@ -1795,11 +1791,9 @@ func (e *Error) UnmarshalJSON(data []byte) error {
 // API error responses are expected to have response
 // body, and a JSON response body that maps to [ErrorResponse].
 //
-// On error responses other than 202 Accepted, CheckResponse consumes r.Body
-// and replaces it with an in-memory copy so that the error body can be
-// re-read. Closing r.Body after CheckResponse returns therefore closes only
-// the copy: to release the original body and its underlying connection,
-// capture r.Body before the call and close the captured body instead.
+// On error responses other than 202 Accepted, CheckResponse consumes and
+// closes r.Body, then replaces it with an in-memory copy so that the error
+// body can be re-read by the caller.
 //
 // The error type will be *[RateLimitError] for rate limit exceeded errors,
 // *[AcceptedError] for 202 Accepted status codes,
@@ -1822,9 +1816,11 @@ func CheckResponse(r *http.Response) error {
 			errorResponse = &ErrorResponse{Response: r}
 		}
 	}
-	// Re-populate error response body because GitHub error responses are often
-	// undocumented and inconsistent.
+	// Close the original body and re-populate the error response body with an
+	// in-memory copy, because GitHub error responses are often undocumented and
+	// inconsistent and the caller may need to re-read them.
 	// Issue #1136, #540.
+	_ = r.Body.Close()
 	r.Body = io.NopCloser(bytes.NewBuffer(data))
 	switch {
 	case r.StatusCode == http.StatusUnauthorized && strings.HasPrefix(r.Header.Get(headerOTP), "required"):
diff --git a/github/github_test.go b/github/github_test.go
index 776abaf4..77b346c6 100644
--- a/github/github_test.go
+++ b/github/github_test.go
@@ -2308,8 +2308,8 @@ func (r *closeRecorder) Close() error {
 	return r.ReadCloser.Close()
 }
 
-// CheckResponse substitutes resp.Body with a re-readable copy on error
-// responses; the network body it replaces must still be closed.
+// CheckResponse closes the original resp.Body and substitutes a re-readable
+// copy on error responses; the network body must not leak.
 func TestDo_closesOriginalBodyOnErrorResponse(t *testing.T) {
 	t.Parallel()
 	client, mux, _ := setup(t)
diff --git a/github/repos_releases.go b/github/repos_releases.go
index 3361a5d7..889afb6d 100644
--- a/github/repos_releases.go
+++ b/github/repos_releases.go
@@ -375,12 +375,8 @@ func (s *RepositoriesService) downloadReleaseAssetFromURL(ctx context.Context, f
 	if err != nil {
 		return nil, err
 	}
-	// CheckResponse substitutes resp.Body with a re-readable copy on error
-	// responses, so capture the original body first: it is the one that must
-	// be closed.
-	origBody := resp.Body
 	if err := CheckResponse(resp); err != nil {
-		_ = origBody.Close()
+		_ = resp.Body.Close()
 		return nil, err
 	}
 	return resp.Body, nil
diff --git a/github/repos_releases_test.go b/github/repos_releases_test.go
index 4fb85c75..4208513d 100644
--- a/github/repos_releases_test.go
+++ b/github/repos_releases_test.go
@@ -537,9 +537,9 @@ func TestRepositoriesService_DownloadReleaseAsset_FollowRedirectToError(t *testi
 	}
 }
 
-// CheckResponse substitutes resp.Body with a re-readable copy on error
-// responses; downloadReleaseAssetFromURL must still close the original body it
-// replaces. Unlike its sibling tests, the recorder wraps the follow-redirects
+// CheckResponse closes the original resp.Body and substitutes a re-readable
+// copy on error responses; downloadReleaseAssetFromURL must not leak the
+// original body. Unlike its sibling tests, the recorder wraps the follow-redirects
 // client's transport: that client, not the library client, performs the
 // redirected request, so wrapping the library client would only ever observe
 // the first hop's correctly-closed redirect response and never the leak.

```

yavorl: The reason I didn't do it initially is that this is a behaviour change, callers might start seeing errors where there were none before.

Reason being - Response.Body is only guaranteed to be an io.ReadCloser. Custom transports, test doubles, and direct CheckResponse callers routinely supply non-Transport bodies - so the only close semantics a public API may rely on are io.Closer's, which states that a second Close is undefined behaviour. So if someone was diligent about this before and they closed the body themselves, theoretically they can start seeing errors or panics.

yavorl: That said, if you would still prefer to go with that approach after my argument. I'd be happy to apply the patch.

alexandear: Fair point. Thanks for the explanation.

alexandear (approved):

github/github.go
yavorl: Thank you for the review & thank you for the merge @gmlewis

gmlewis:
Thank you, @yavorl and @alexandear!
Merging.

PR #4492, opened 2026-08-26
fix: Handle string-valued error responses

n-g:
GitHub sometimes serves repeated errors, sometimes single strings.


gmlewis:
@n-g - when you resolve the Google CLA issues, we can move forward with this PR.

PR #4493, opened 2026-08-26
refactor!: Rename `EditComment` to `UpdateComment` on `PullRequestsService`, split review comment request bodies, and pass by value

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for the review comment endpoints on `PullRequestsService` — the pulls counterpart of #4444.

**Commit 1 (`refactor!:`)** — `CreateComment` and `EditComment` reused the 27-field `PullRequestComment` **response** type as their request bodies, and `EditComment`'s doc comment had to warn:

> A non-nil comment.Body must be provided. Other comment fields should be left nil.

Per the [docs](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28), the two schemas differ, so the body is split (same approach as #4382/#4401/#4438):

| Field | Create (POST) | Update (PATCH) |
|---|---|---|
| `body` | **required** | **required** (only parameter) |
| `commit_id`, `path` | **required** | — |
| `line`, `start_line`, `start_side` | conditionally required ("required unless …") → stay pointers | — |
| `side`, `in_reply_to`, `subject_type` | optional | — |
| `position` | optional, deprecated → kept with a `// Deprecated:` note | — |

- **`CreatePullRequestCommentRequest`** — `Body`/`CommitID`/`Path` are non-pointer since they're unconditionally required; the conditionally-required line fields stay pointers (same treatment as `title` in #4395). Note the request tag is `in_reply_to`, not the response's `in_reply_to_id`.
- **`UpdatePullRequestCommentRequest`** — just `Body string`.

`EditComment` is renamed to `UpdateComment` to match the docs operation name (same convention as #4444). `CreateCommentInReplyTo` already builds its own minimal body internally and is unchanged. `PullRequestComment` is removed from the `body-allowed-pointer-types` allowlist.

**Commit 2 (`feat:`)** — while cross-checking the **response** schema (prompted by the review feedback on #4444 and #4477, where missing response fields were caught by reviewers), I found three properties missing from `PullRequestComment` and added them up front: `BodyHTML`, `BodyText`, and `Links` (`_links`, modeled with a new `PullRequestCommentLinks` type reusing the existing `PRLink`).

Verified with `go build ./...`, `go vet -tags integration ./test/integration/`, `gofmt`, the full `./github/` test suite (`CreateComment`/`UpdateComment` and the generated accessors at 100%), and `custom-gcl` (no `paramcheck` findings after removing the allowlist entry).

Updates #3644

[...]

PR #4444, opened 2026-08-08
refactor!: Rename `EditComment` to `UpdateComment` on `IssuesService`, pass a new `IssueCommentRequest` by value, and add missing `IssueComment` fields

JamBalaya56562:
Continues the request-body-by-value work in #3644, this time for the issue comment endpoints on `IssuesService`.

`CreateComment` and `EditComment` reused the `IssueComment` **response** type as their request bodies, but per the [docs](https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28) both endpoints accept exactly one parameter, `body`, and it's **required** in both. Every other field (`id`, `user`, `reactions`, `created_at`, …) is server-generated. The clearest evidence is in the code itself — `EditComment`'s doc comment had to warn:

> A non-nil comment.Body must be provided. Other comment fields should be left nil.

The new request type makes that warning unnecessary by construction:

```go
type IssueCommentRequest struct {
	Body string `json:"body"`
}
```

Unlike the milestone split in #4438, the create and update schemas here are **identical**, so a single shared request type is used rather than a create/update pair (same reasoning as `HookConfig` in #4360). `Body` is a non-pointer `string` since it's required in both operations.

`EditComment` is renamed to `UpdateComment` to match the docs operation name, [Update an issue comment](https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#update-an-issue-comment) (same convention as #4438/#4400). Both methods now take the body by value, and `IssueComment` is removed from the `body-allowed-pointer-types` allowlist.

## Missing response fields

Three fields that the issue-comment response schema returns but `IssueComment` did not have were added during review (thanks @Not-Dhananjay-Mishra). These additions are purely additive — they are not part of the breaking change:

- `PerformedViaGithubApp *App` — same shape as the existing five uses of `performed_via_github_app` in the codebase (`Issue`, `IssueEvent`, `Timeline`, …).
- `Pin *PinnedIssueComment` — new type with `PinnedAt *Timestamp` / `PinnedBy *User`, matching the "Pinned Issue Comment" schema.
- `Minimized *MinimizedIssueComment` — new type with `Reason *string`, matching the "Minimized Issue Comment" schema.

All three are nullable in the response, so they are pointers with `omitempty`. The rest of the `IssueComment` response type is unchanged.

[...]

PR #4496, opened 2026-08-28
refactor!: Split `RepositoryComment` request bodies and pass by value

JamBalaya56562:
Continues the request-body-by-value work in #3644 — the commit-comments counterpart of #4444 (`IssueComment`, issues) and #4493 (`PullRequestComment`, pulls), completing the comment-endpoint family.

**Commit 1 (`refactor!:`)** — `RepositoriesService.CreateComment` and `UpdateComment` reused the 12-field `RepositoryComment` **response** type as their request bodies. Per the [create](https://docs.github.com/rest/commits/comments?apiVersion=2022-11-28#create-a-commit-comment) / [update](https://docs.github.com/rest/commits/comments?apiVersion=2022-11-28#update-a-commit-comment) docs the two request schemas differ, so this splits them into dedicated types (same shape as #4493):

| | `body` | `path` | `position` | `line` |
|---|---|---|---|---|
| create | **required** | optional | optional | optional (deprecated — "use position instead") |
| update | **required** | — | — | — |

```go
type CreateCommitCommentRequest struct {
	Body     string  `json:"body"`
	Path     *string `json:"path,omitempty"`
	Position *int    `json:"position,omitempty"`
	// Deprecated: Use Position instead.
	Line *int `json:"line,omitempty"`
}

type UpdateCommitCommentRequest struct {
	Body string `json:"body"`
}
```

Both methods now take these by value (`Body` required → non-pointer, no `omitempty`), and `RepositoryComment` is removed from the `paramcheck` allowlist in `.golangci.yml`. No method renames are needed — `Create`/`Update` already match the docs operation names. Alongside:

- the now-stale `// User-mutable fields` / `// User-initialized fields` markers on `RepositoryComment` are dropped and `Body`'s tag normalized to `json:"body,omitempty"`, making it a plain response type like `IssueComment` / `PullRequestComment`;
- the `{comment_id}` parameter is named `commentID` consistently across `GetComment` / `UpdateComment` / `DeleteComment` (matching the issues/pulls comment files);
- the request types also gain the deprecated `line` create parameter, which the old shared type could not send at all.

**Commit 2 (`feat:`, non-breaking)** — cross-checking the full [commit-comment response schema](https://docs.github.com/rest/commits/comments?apiVersion=2022-11-28#get-a-commit-comment) against the struct turned up two missing fields, added here with regenerated accessors: `AuthorAssociation` and `Line`.

BREAKING CHANGE: `RepositoriesService.CreateComment` now takes `CreateCommitCommentRequest` and `RepositoriesService.UpdateComment` takes `UpdateCommitCommentRequest`, both passed by value instead of `*RepositoryComment`.

Verified with `go build ./...`, `gofmt`, `go vet -tags integration ./test/integration/`, the full `./github/` test suite (all `repos_comments.go` methods at 100% coverage), generator idempotency, and `custom-gcl` (0 issues with the allowlist entry removed).

Updates #3644.

cc @jvm986

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -236,7 +236,6 @@ linters:
             - PullRequestReviewsEnforcementUpdate
             - Repository
             - RepositoryAddCollaboratorOptions
-            - RepositoryComment
             - RepositoryContentFileOptions
             - RepositoryCreateForkOptions
             - RequiredStatusChecksRequest
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -11582,6 +11582,38 @@ func (c *CreateCodespaceOptions) GetWorkingDirectory() string {
 	return *c.WorkingDirectory
 }
 
+// GetBody returns the Body field.
+func (c *CreateCommitCommentRequest) GetBody() string {
+	if c == nil {
+		return ""
+	}
+	return c.Body
+}
+
+// GetLine returns the Line field if it's non-nil, zero value otherwise.
+func (c *CreateCommitCommentRequest) GetLine() int {
+	if c == nil || c.Line == nil {
+		return 0
+	}
+	return *c.Line
+}
+
+// GetPath returns the Path field if it's non-nil, zero value otherwise.
+func (c *CreateCommitCommentRequest) GetPath() string {
+	if c == nil || c.Path == nil {
+		return ""
+	}
+	return *c.Path
+}
+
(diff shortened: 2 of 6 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

alexandear (approved):

gmlewis:
Merging.

PR #4518, opened 2026-09-04
feat: Add ThrottledAt field to HookDelivery

### LINKED ISSUE #4517: HookDelivery is missing the throttled_at field returned by the API
The `hook-delivery` and `hook-delivery-item` schemas in GitHub's OpenAPI description include `throttled_at` (string, date-time, nullable — "Time when the webhook delivery was throttled."). The live API returns it: listing deliveries for a webhook on my fork after a real push event gives a body containing `"throttled_at": null` alongside the other fields.

`HookDelivery` in `github/repos_hooks_deliveries.go` does not declare the field, so it is dropped on decode. Verified on `master` (v91) through the public API path:

```go
client, _ := NewClient(WithAuthToken(token))

// raw body via the same client: throttled_at is present
req, _ := client.NewRequest(ctx, "GET", "repos/OWNER/REPO/hooks/HOOK_ID/deliveries?per_page=1", nil)
var raw []map[string]any
client.Do(req, &raw)
_, ok := raw[0]["throttled_at"] // true

// typed path: throttled_at is gone
ds, _, _ := client.Repositories.ListHookDeliveries(ctx, "OWNER", "REPO", HOOK_ID, &ListCursorOptions{PerPage: 1})
b, _ := json.Marshal(ds[0])
// {"id":...,"guid":...,"delivered_at":...,"redelivery":false,"duration":0,"status":...,"status_code":502,"event":"push","repository_id":...}
```

Same pattern as #4503. I'd like to add `ThrottledAt *Timestamp` to `HookDelivery` and cover it in the existing tests. PR to follow.

MarcHuang168:
Adds the `throttled_at` field to `HookDelivery`. It is documented in the `hook-delivery` / `hook-delivery-item` OpenAPI schemas, and confirmed present in the live list-deliveries and get-delivery responses for a real push event on my fork; it was previously dropped on decode.

Note on verification: the drop was observed end-to-end through `ListHookDeliveries` on the live API (raw body has the key, decoded struct does not). For `GetHookDelivery` the live response also carries the key, but its value was `null` on my fork (the delivery was not throttled), so the typed path can't show a difference there; both methods decode into the same `HookDelivery` struct, and the non-null case is covered by the updated unit tests.

Updated the existing `ListHookDeliveries` and `GetHookDelivery` tests to cover the new field.

Fixes #4517

AI assistance: Claude was used to diff go-github's structs against GitHub's OpenAPI description and to draft the edits; the field was verified against the live API, and the change was applied, tested (`script/fmt.sh`, `script/test.sh`, `script/lint.sh`) and reviewed locally by me.

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -19430,6 +19430,14 @@ func (h *HookDelivery) GetStatusCode() int {
 	return *h.StatusCode
 }
 
+// GetThrottledAt returns the ThrottledAt field if it's non-nil, zero value otherwise.
+func (h *HookDelivery) GetThrottledAt() Timestamp {
+	if h == nil || h.ThrottledAt == nil {
+		return Timestamp{}
+	}
+	return *h.ThrottledAt
+}
+
 // GetHeaders returns the Headers map if it's non-nil, an empty map otherwise.
 func (h *HookRequest) GetHeaders() map[string]string {
 	if h == nil || h.Headers == nil {
--- a/github/github-accessors_test.go
+++ b/github/github-accessors_test.go
@@ -24414,6 +24414,17 @@ func TestHookDelivery_GetStatusCode(tt *testing.T) {
 	h.GetStatusCode()
 }
 
+func TestHookDelivery_GetThrottledAt(tt *testing.T) {
+	tt.Parallel()
+	var zeroValue Timestamp
+	h := &HookDelivery{ThrottledAt: &zeroValue}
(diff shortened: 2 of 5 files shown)


gmlewis (approved):
Thank you, @MarcHuang168!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

PR #4519, opened 2026-09-04
refactor!: Rename `DependencyGraphSnapshot` to `CreateDependencyGraphSnapshotRequest` and pass by value

JamBalaya56562:
Continues the request-body-by-value work in #3644 with `DependencyGraphService.CreateSnapshot`, whose body is the dependency submission snapshot document.

### Why this one

The [snapshot schema](https://docs.github.com/rest/dependency-graph/dependency-submission?apiVersion=2022-11-28#create-a-snapshot-of-dependencies-for-a-repository) marks six of its eight properties as required, and its nested objects have required fields of their own, yet the Go types modelled all of them as optional pointers. This is the request-body type that the `check-schema-fields` tool proposed in #4375 flagged first, so this PR fixes the real finding rather than only flipping a parameter:

| Schema | Required | Before | After |
|---|---|---|---|
| `snapshot` | `version` | `Version int` | unchanged |
| | `sha`, `ref` | `*string,omitempty` | `string` |
| | `job`, `detector` | `*DependencyGraphSnapshotJob,omitempty` / `*DependencyGraphSnapshotDetector,omitempty` | value structs |
| | `scanned` | `*Timestamp,omitempty` | `Timestamp` |
| | `metadata`, `manifests` (optional) | maps with `omitempty` | unchanged |
| `job` | `id`, `correlator` | `*string,omitempty` | `string` (`html_url` stays optional) |
| `detector` | `name`, `version`, `url` | `*string,omitempty` | `string` |
| `manifest` | `name` | `*string,omitempty` | `string` (`file`, `metadata`, `resolved` stay optional) |

Value-typed nested structs follow `CreateHostedRunnerRequest.Image HostedRunnerImage` (#4371). `Scanned Timestamp` is the first required timestamp in an exported request type; it uses a bare `json:"scanned"` tag since the API requires it on every submission.

### Commits

1. **`refactor!:` `Sha` → `SHA`** — the field was on the `structfield` allowlist as a TODO; since the type is changing incompatibly anyway, this clears that entry.
2. **`refactor!:` rename + pass by value** — `DependencyGraphSnapshot` becomes `CreateDependencyGraphSnapshotRequest` (the `Create…Request` convention used by the rest of the series; the `DependencyGraphSnapshot*` sub-types keep their names as parts of the snapshot document), `CreateSnapshot` takes it by value, the required fields above become non-pointer, and `DependencyGraphSnapshot` is removed from the `paramcheck` allowlist. The docs operation name ("Create a snapshot…") already matches the method verb, so no method rename.
3. **`docs:` `//meta:schema` annotation** — a plain comment until #4375 lands, after which `check-schema-fields` validates this body automatically. Running the tool from that branch against this PR reports `Found 0 schema field issues` / `Checked 1 annotations on 1 annotated structs` — before this change the same type produced five required-but-pointer findings.

BREAKING CHANGE: `DependencyGraphSnapshot` is renamed to `CreateDependencyGraphSnapshotRequest`, its `Sha` field to `SHA`, `DependencyGraphService.CreateSnapshot` takes it by value, and the fields the API requires (`SHA`, `Ref`, `Job`, `Detector`, `Scanned`; `DependencyGraphSnapshotJob.ID` and `Correlator`; `DependencyGraphSnapshotDetector.Name`, `Version`, and `URL`; `DependencyGraphSnapshotManifest.Name`) are no longer pointers.

Verified with `go build ./...`, `gofmt`, `go vet -tags integration ./test/integration/`, the full `./github/` test suite (`CreateSnapshot` at 100% coverage), generator idempotency, and `custom-gcl` (0 issues with the allowlist entries removed). No usages outside the `github` package.

Updates #3644.

cc @jvm986

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -216,7 +216,6 @@ linters:
             - CustomDeploymentProtectionRuleRequest
             - CustomProperty
             - DependabotAlertState
-            - DependencyGraphSnapshot
             - EncryptedSecret
             - EnterpriseSecurityAnalysisSettings
             - Hook
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -304,7 +304,6 @@ linters:
             - CommitsSearchResult.Commits
             - CommitsSearchResult.Total
             - CreateOrgInvitationOptions.TeamID # TODO: TeamIDs
-            - DependencyGraphSnapshot.Sha # TODO: SHA
             - Discussion.DiscussionCategory # TODO: Category ?
             - EditOwner.OwnerInfo
             - EnterpriseLicensedUsers.GithubComSamlNameID # TODO: GithubComSAMLNameID
--- a/github/dependency_graph_snapshots.go
+++ b/github/dependency_graph_snapshots.go
@@ -77,7 +77,7 @@ type DependencyGraphSnapshotManifest struct {
 // GitHub API docs: https://docs.github.com/rest/dependency-graph/dependency-submission?apiVersion=2022-11-28#create-a-snapshot-of-dependencies-for-a-repository
 type DependencyGraphSnapshot struct {
 	Version   int                                         `json:"version"`
-	Sha       *string                                     `json:"sha,omitempty"`
+	SHA       *string                                     `json:"sha,omitempty"`
 	Ref       *string                                     `json:"ref,omitempty"`
 	Job       *DependencyGraphSnapshotJob                 `json:"job,omitempty"`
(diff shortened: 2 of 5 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #4523, opened 2026-09-05
feat: Add repository ruleset code quality support

### LINKED ISSUE #4520: Repository ruleset rule type `code_quality` is not modelled
### Summary

The repository ruleset rule type `code_quality` is not modelled by `RepositoryRule` / the ruleset rule types, so it cannot be read or written through this library. It corresponds to the **"Require code quality results"** checkbox on a branch ruleset, which reached general availability on 2026-07-20.

### API shape

`GET /repos/{owner}/{repo}/rulesets/{id}` returns it as a rule alongside the existing types:

```json
{
  "id": 16352232,
  "name": "Project branch",
  "target": "branch",
  "enforcement": "active",
  "rules": [
    { "type": "code_quality", "parameters": { "severity": "errors" } },
    { "type": "copilot_code_review", "parameters": { "review_on_push": true, "review_draft_pull_requests": false } }
  ]
}
```

`severity` is the lowest severity level at which code quality results block a merge. I have only observed `"errors"` on the wire, so the full accepted set is worth confirming against the docs before fixing an enum — note the existing `code_scanning` rule's `security_alerts_threshold` already uses `"errors"` / `"errors_and_warnings"` style values in `rules.go`, which may or may not be the same vocabulary here.

### Why it matters beyond a missing field

`PUT /repos/{owner}/{repo}/rulesets/{id}` replaces the entire `rules` array. Because the rule cannot be represented, any consumer that round-trips a ruleset — read, modify one unrelated attribute, write back — will silently drop `code_quality` and remove a merge gate, with nothing in the request or response to indicate it happened.

That is not hypothetical. `integrations/terraform-provider-github` builds its `github_repository_ruleset` resource on this library, and a ruleset carrying this rule imports cleanly, plans cleanly, and would lose the rule on the first apply that touched any other attribute of the resource. We ended up dropping the affected ruleset from Terraform management entirely rather than hold that risk. Modelling the type here unblocks the provider side (there is no provider-side workaround, since there is nothing to serialise).

### Environment

- go-github: absent as of current `master` — `github/rules.go` models `copilot_code_review`, `required_code_scanning`, `required_status_checks` et al., and contains no occurrence of `code_quality` or `CodeQuality`
- Observed against: GitHub.com, private repository in an organisation on the Team plan

### Willing to contribute

Happy to open a PR adding the rule type if a maintainer confirms the preferred shape — in particular whether `severity` should be a plain `string` or a typed constant set.

tvost2:
## Summary

- model the GitHub `code_quality` repository ruleset rule
- preserve its severity when marshaling and unmarshaling rulesets
- add regression coverage for round-tripping the rule

Fixes #4520

## Tests

- `go test ./github -count=1`

### DIFF
--- a/github/github-accessors.go
+++ b/github/github-accessors.go
@@ -3934,6 +3934,14 @@ func (b *BranchRules) GetBranchNamePattern() []*PatternBranchRule {
 	return b.BranchNamePattern
 }
 
+// GetCodeQuality returns the CodeQuality slice if it's non-nil, nil otherwise.
+func (b *BranchRules) GetCodeQuality() []*CodeQualityBranchRule {
+	if b == nil || b.CodeQuality == nil {
+		return nil
+	}
+	return b.CodeQuality
+}
+
 // GetCodeScanning returns the CodeScanning slice if it's non-nil, nil otherwise.
 func (b *BranchRules) GetCodeScanning() []*CodeScanningBranchRule {
 	if b == nil || b.CodeScanning == nil {
@@ -5334,6 +5342,14 @@ func (c *CodeQLDatabase) GetURL() string {
 	return *c.URL
 }
 
+// GetParameters returns the Parameters field.
+func (c *CodeQualityBranchRule) GetParameters() CodeQualityRuleParameters {
+	if c == nil {
+		return CodeQualityRuleParameters{}
+	}
+	return c.Parameters
+}
+
 // GetCreatedAt returns the CreatedAt field if it's non-nil, zero value otherwise.
 func (c *CodeQualityFinding) GetCreatedAt() Timestamp {
 	if c == nil || c.CreatedAt == nil {
@@ -5492,6 +5508,14 @@ func (c *CodeQualityFindingRule) GetTitle() string {
 		return ""
 	}
 	return c.Title
+}
+
+// GetSeverity returns the Severity field.
+func (c *CodeQualityRuleParameters) GetSeverity() CodeQualitySeverity {
+	if c == nil {
+		return ""
+	}
+	return c.Severity
 }
 
 // GetLanguages returns the Languages slice if it's non-nil, nil otherwise.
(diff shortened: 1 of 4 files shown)


gmlewis:
@tvost2 - please run step 4 of CONTRIBUTING.md and push the changes to this PR.

tvost2:
Addressed the Codecov coverage comment by adding marshal/unmarshal regression coverage for `BranchRules.CodeQuality`.

Validation:
- `go test ./... -count=1` passes.
- `gofmt` applied.
- `git diff --check` passes.
- `script/fmt.sh` and `script/lint.sh` could not complete in this Windows environment: the shell wrapper stalled after installing golangci-lint, and the direct linter invocation reports the repository plugin `structfield` is unavailable locally.

gmlewis:
> * `script/fmt.sh` and `script/lint.sh` could not complete in this Windows environment: the shell wrapper stalled after installing golangci-lint, and the direct linter invocation reports the repository plugin `structfield` is unavailable locally.

Hmm... was this in a WSL or `git bash` window? If not, could you please try running them in one of these windows?
I don't have a Windows box available, and we have not had this report before, so I'm just guessing here.

tvost2:
Updated the PR after the maintainer feedback:

- ran generation in Git Bash and committed the generated `CodeQuality` accessors/tests;
- added invalid-parameter regressions for repository and branch rules to cover the remaining Codecov misses;
- focused ruleset tests pass and `git diff --check` passes.

The official Git Bash wrapper reaches the repository scripts, but this Windows checkout cannot execute the downloaded `custom-gcl` path. `script/generate.sh --check` now gets past the generated Go files and reports only the local toolchain changing `go.sum` during tidy; no dependency files were included in this PR.

gmlewis:
@tvost2 - note that issue #4520 is assigned to someone else anyway so we will wait for their response. 
As covered in CONTRIBUTING.md, please sign up to address open issues before posting a PR to prevent application of effort.

PR #4530, opened 2026-09-08
fix: treat 202 Accepted as success in MergeAsync

rootkiller6788:
MergeAsync trips over the one response this endpoint actually returns: GitHub answers 202 Accepted once the merge request is enqueued, but Client.Do turns every 202 into an *AcceptedError. So MergeAsync errored out with a nil result on the happy path, and callers never got the uuid they need to pass to GetMergeAsyncResult.

The fix decodes the 202 payload into the result and returns it without an error, using the same approach as UploadSarif. I also made the existing MergeAsync test write a real 202 instead of the default 200 - the mock status was why this never surfaced. The test fails without the code change and passes with it.

### DIFF
--- a/github/pulls.go
+++ b/github/pulls.go
@@ -8,6 +8,7 @@ package github
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 )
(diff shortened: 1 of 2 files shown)


alexandear (changes requested):
Please create an issue first.

PR #4534, opened 2026-09-10
fix: Escape branch names when listing rules

nickita-khylkouski:
hey, ListRulesForBranch puts the branch name straight into the URL, so a valid name like fix/#123-login gets truncated at #. literal % characters can also change the name or make URL parsing fail. this uses url.PathEscape(branch), like the other branch methods and #2948. added regression tests in repos_rules_test.go for ordinary names, slashes, #, and literal %, with and without pagination. BREAKING CHANGE: callers that already escape names as a workaround need to pass the raw branch name instead. this also applies to ListRulesForBranchIter. AI-assisted: ChatGPT helped investigate and draft the fix/tests. the isolated regression checks fail before the fix and pass after; full Go 1.26 repository checks are still pending.

### DIFF
--- a/github/repos_rules.go
+++ b/github/repos_rules.go
@@ -9,15 +9,18 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"net/url"
 )
 
 // ListRulesForBranch gets all the repository rules that apply to the specified branch.
+//
+// Note: the branch name is URL path escaped for you. See: https://pkg.go.dev/net/url#PathEscape .
 //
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @nickita-khylkouski!
LGTM.

Note that I removed the "!" from "fix!:" since you are not technically breaking the API, although you will be breaking users you already escape the branch name... so we will leave the "BREAKING CHANGE" in the description, but we won't call it out as a breaking API change. I hope that makes sense.

alexandear (approved):

gmlewis:
Thank you, @alexandear!
Merging.

PR #4535, opened 2026-09-10
chore: Remove alexandear from REVIEWERS

alexandear:
Removes me from REVIEWERS as I'm stepping back from reviewing.

Added in #3588.


gmlewis (approved):
Thank you, @alexandear, for your excellent reviews. Your efforts are greatly appreciated.

PR #4536, opened 2026-09-10
fix: Escape environment names in URL paths across 5 files

prasanna8585:
A sibling gap to the branch-name escaping fix in ListRulesForBranch (#4534): every method that takes a GitHub "environment name" and embeds it into a request path via fmt.Sprintf did so without url.PathEscape, unlike every branch-name usage in repos.go, which already escapes consistently.

GitHub environment names, like branch names, are free-form strings set by the repository owner (e.g. "staging", "team/staging") with no character restriction comparable to the one GitHub's own naming rules place on owner/repo segments. An environment name containing a `/` is silently reinterpreted as additional path segments rather than staying part of the environment-name segment, redirecting the request to a different, unintended path under the caller's own GitHub credentials.

Fixed 23 call sites across 5 files, all mapping to the same underlying REST resource shape (`/repos/{owner}/{repo}/environments/{environment_name}/...`), just with a different local parameter name per file (`name`, `environment`, or `env`):

  - github/repos_environments.go (3 sites: GetEnvironment, CreateUpdateEnvironment, DeleteEnvironment)
  - github/repos_deployment_branch_policies.go (5 sites)
  - github/repos_deployment_protection_rules.go (5 sites)
  - github/actions_secrets.go (5 sites)
  - github/actions_variables.go (5 sites)

One incidental fix needed along the way: ActionsService.GetEnvPublicKey in actions_secrets.go named its own local URL-path variable `url`, which would have shadowed the newly-imported net/url package within that function's scope. Renamed the local variable to `u`, matching the convention used by every other method in these files; no behavior change beyond that rename.

Verified: gofmt -l reports no issues on all 10 touched files (5 source, 5 test). Full `go build`/`go test` could not be run in this environment (this module's go.mod requires go >= 1.26, network egress to proxy.golang.org for module downloads is unavailable here), so this depends on CI to confirm compilation and test results; the change itself is a mechanical, well-precedented pattern (matching url.PathEscape usage already established for `branch` throughout repos.go and for `branch` in this same PR's own ListRulesForBranch fix), and every modified call site and its surrounding function was reviewed by hand for correct variable references.

Added one escape-path regression test per file (5 new tests total), following the same table-driven pattern PR #4534 established for ListRulesForBranch: each asserts both the escaped wire-format request path (`r.URL.EscapedPath()`) and the decoded path (`r.URL.Path`) for a plain environment name and one containing a `/`.

### DIFF
--- a/github/actions_secrets.go
+++ b/github/actions_secrets.go
@@ -85,6 +85,8 @@ func (s *ActionsService) GetOrgPublicKey(ctx context.Context, org string) (*Publ
 
 // GetEnvPublicKey gets a public key that should be used for secret encryption.
 //
+// Note: the environment name is URL path escaped for you. See: https://pkg.go.dev/net/url#PathEscape .
+//
 // GitHub API docs: https://docs.github.com/rest/actions/secrets?apiVersion=2022-11-28#get-an-environment-public-key
 //
 //meta:operation GET /repos/{owner}/{repo}/environments/{environment_name}/secrets/public-key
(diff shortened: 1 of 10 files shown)


gmlewis (changes requested):
All modified methods need to have their comments updated so that they explicitly state that path escaping is being performed for the user.

gmlewis (approved):
Thank you, @prasanna8585!
LGTM.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

PR #4540, opened 2026-09-10
refactor!: Pass the Codespaces body types by value and rename them to `...Request`

JamBalaya56562:
Updates #3644

This converts the last four `CodespacesService` body types, removing 8 entries from the `paramcheck` exception lists in `.golangci.yml` (4 from `body-allowed-pointer-types` and 4 from `body-allowed-wrong-names`).

Checking the four types against the OpenAPI descriptions turned up two problems that are fixed here in separate commits:

1. `POST /user/codespaces` takes a `oneOf` body: either `repository_id` (with an optional `ref`) or `pull_request`. `CodespaceCreateForUserOptions` declared both `RepositoryID int64 \`json:"repository_id"\`` and `PullRequest *CodespacePullRequestOptions \`json:"pull_request"\`` without an omit option, so every request sent `"repository_id": 0` and `"pull_request": null` together, no matter which branch the caller meant. `RepositoryID` is now `*int64` and both fields use `omitempty`, so only the branch the caller filled in is sent. That `oneOf` is also why these two fields stay pointers instead of becoming required value types.
2. The `PATCH /user/codespaces/{codespace_name}` schema has `machine`, `display_name` and `recent_folders`, but `DisplayName` was missing from `UpdateCodespaceOptions`.

A few notes on what is deliberately *not* changed:

- `CodespacesService.Publish` keeps its name. The docs page is titled "Create a repository from an unpublished codespace", but the operation id is `codespaces/publish-for-authenticated-user`, so the current verb already matches the operation.
- `CodespacePullRequestOptions` keeps its name. It is a nested field type rather than a body parameter type, so the `Options` suffix rule does not apply to it.
- `CreateCodespaceRequest` stays shared between `CreateInRepo` and `CreateFromPullRequest`. The only difference between the two schemas is that the pull request variant has no `ref`.

BREAKING CHANGE: `CreateCodespaceOptions`, `CodespaceCreateForUserOptions`, `UpdateCodespaceOptions` and `PublishCodespaceOptions` are renamed to `CreateCodespaceRequest`, `CreateCodespaceForUserRequest`, `UpdateCodespaceRequest` and `PublishCodespaceRequest`; `CodespacesService.CreateInRepo`, `CreateFromPullRequest`, `Create`, `Update` and `Publish` now take them by value instead of by pointer; `CreateCodespaceForUserRequest.RepositoryID` is now `*int64`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

github/codespaces.go
```diff
@@ -150,8 +150,10 @@ func (s *CodespacesService) List(ctx context.Context, opts *ListCodespacesOption
 	return codespaces, resp, nil
 }
 
-// CreateCodespaceOptions represents options for the creation of a codespace in a repository.
-type CreateCodespaceOptions struct {
+// CreateCodespaceRequest represents a request to create a codespace in a repository.
+//
+//meta:schema request POST /repos/{owner}/{repo}/codespaces
```
Not-Dhananjay-Mishra, line 155: Why are we adding `meta:schema request ...` to structs?

JamBalaya56562: Fair question. They are for the `check-schema-fields` command in #4375 (not merged yet): it checks an annotated struct's field optionality against the request body schema of the named operation in the OpenAPI descriptions, the same way `//meta:operation` ties a method to its operation. Running it on this branch is how the missing `display_name` and the `oneOf` problem here were found. #4519 added the first one on `CreateDependencyGraphSnapshotRequest`.

Until #4375 lands they are inert comments, so if you would rather not carry them in unrelated PRs before that, I'm happy to drop them from this PR.

Not-Dhananjay-Mishra: > Until #4375 lands they are inert comments, so if you would rather not carry them in unrelated PRs before that, I'm happy to drop them from this PR.
> 

Please go ahead and drop them from this PR.

JamBalaya56562: Done, dropped in ab89cb3.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Thank you, @JamBalaya56562 and @Not-Dhananjay-Mishra!
Merging.

PR #4541, opened 2026-09-10
fix: Allow removing runner group networks

austenstone:
Add `RemoveNetworkConfiguration` to organization and enterprise runner-group update requests, following [`UpdateTeamRequest.RemoveParentTeam`](github/teams.go). Setting it sends `"network_configuration_id": null` and takes precedence over a supplied ID. The value-receiver marshaler supports both values and pointers without mutation. Creates and responses are unchanged.

The Go 1.26.0 `encoding/json` repro shows that `*string` behaves identically with `omitempty` and `omitzero`: nil omits the field, `new("network-id")` sends a string, and `new("")` sends `""`. Neither tag produces JSON null. Empty-string behavior is preserved for compatibility, but empty-string detachment is unverified. Both PATCH schemas allow null: [organization](https://github.com/github/rest-api-description/blob/0a144e030abbe358000f0d32874e197b59a8f7ed/descriptions/api.github.com/api.github.com.json#L15939-L15943) and [enterprise](https://github.com/github/rest-api-description/blob/0a144e030abbe358000f0d32874e197b59a8f7ed/descriptions/ghec/ghec.json#L8154-L8158).

Regression tests cover omission, assignment, empty-string compatibility, explicit null, flag precedence, preservation of other fields, value/pointer marshaling, request immutability, and both native PATCH bodies. The null cases fail without the marshalers. On Go 1.26.0, runner-group race tests, all-module `script/test.sh` race tests, `script/fmt.sh`, `script/generate.sh`, `script/lint.sh`, and OpenAPI metadata validation pass. No live API writes were made for this revision.

This revises the earlier `omitzero` implementation and still needs maintainer agreement on explicit-null removal. Needed by integrations/terraform-provider-github#3274, which remains gated on a merged and released SDK fix.

Copilot assisted with implementation, tests, and this description.

### DIFF
--- a/github/actions_runner_groups.go
+++ b/github/actions_runner_groups.go
@@ -7,6 +7,7 @@ package github
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 )
 
@@ -59,6 +60,24 @@ type UpdateRunnerGroupRequest struct {
 	RestrictedToWorkflows    *bool    `json:"restricted_to_workflows,omitempty"`
 	SelectedWorkflows        []string `json:"selected_workflows,omitempty"`
 	NetworkConfigurationID   *string  `json:"network_configuration_id,omitempty"`
+
+	// If true, the network configuration is removed by sending null.
+	// This takes precedence over NetworkConfigurationID.
+	RemoveNetworkConfiguration bool `json:"-"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (r UpdateRunnerGroupRequest) MarshalJSON() ([]byte, error) {
+	type alias UpdateRunnerGroupRequest
+	if !r.RemoveNetworkConfiguration {
+		return json.Marshal(alias(r))
+	}
+	return json.Marshal(&struct {
+		alias
+		NetworkConfigurationID *string `json:"network_configuration_id"`
+	}{
+		alias: alias(r),
+	})
 }
 
 // SetRepoAccessRunnerGroupRequest represents a request to replace the list of repositories
--- a/github/actions_runner_groups_test.go
+++ b/github/actions_runner_groups_test.go
@@ -6,6 +6,7 @@
 package github
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"testing"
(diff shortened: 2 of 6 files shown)


github/actions_runner_groups.go
```diff
@@ -59,6 +60,24 @@ type UpdateRunnerGroupRequest struct {
 	RestrictedToWorkflows    *bool    `json:"restricted_to_workflows,omitempty"`
 	SelectedWorkflows        []string `json:"selected_workflows,omitempty"`
 	NetworkConfigurationID   *string  `json:"network_configuration_id,omitempty"`
+
+	// If true, the network configuration is removed by sending null.
+	// This takes precedence over NetworkConfigurationID.
+	RemoveNetworkConfiguration bool `json:"-"`
```
gmlewis, line 66: If I'm not mistaken, we can solve this in a different way without resorting to adding a new "hidden" field to the struct... 

We simply change line 62 above to this:

```go
NetworkConfigurationID   *string  `json:"network_configuration_id,omitzero"`
```

and then when the user doesn't initialize the field (it is `nil`), the field will be ignored.
However, if the user wants to REMOVE the network configuration, they would send `new("")`.

I think we should put this in CONTRIBUTING.md on line 418 right after this comment:

```
For optional boolean fields where you need to distinguish between `false`
and "not set", use `*bool` with `omitzero`.
```

gmlewis (changes requested):

github/actions_runner_groups.go
austenstone: Updated in `b0fc528f`: replaced the custom removal flag/marshaler with `*string` + `omitzero`, so nil omits the field and `new("")` sends the removal value. I also added the convention to `CONTRIBUTING.md`, regenerated accessors, and verified `go test ./github` plus `go vet ./github`.

gmlewis:
@austenstone - just like in another recent PR, we now have a CLA problem because an agent signed a commit that was pushed to this PR.

```
The following contributors were found for this pull request:

✅ https://github.com/google/go-github/commit/b0fc528fc916672eb5d41e81defdbb91c1aa4416 Author: @austenstone <aus******ne​@github.com>
❌ https://github.com/google/go-github/commit/b0fc528fc916672eb5d41e81defdbb91c1aa4416 Co-Author: @Copilot <223556219+Copilot​@users.noreply.github.com>
```

So now, you can either clean up the commits locall and force-push the changes to this PR to get rid of the bot's commits or you can close this PR and open a brand new one. Your choice, but we cannot continue this PR with the current CLA state, unfortunately.

CONTRIBUTING.md
```diff
@@ -413,8 +413,15 @@ type RepositoryRuleset struct {
...
-and "not set", use `*bool` with `omitzero`.
+For optional primitive fields where the zero value has API semantics, use a
+pointer with `omitempty`. A nil pointer omits the field, while a pointer to the
+zero value includes it, such as `false`, `0`, or `""`.
+
+Neither `omitempty` nor `omitzero` makes a pointer to a zero value encode as
+JSON `null`. When an update must distinguish omission, assignment, and explicit
+removal using `null`, follow `UpdateTeamRequest.RemoveParentTeam`: add a Go-only
+removal flag and a value-receiver `MarshalJSON` method. This supports marshaling
+both request values and pointers without changing the request.
```
gmlewis, line 424: How about this?

```suggestion
Optional pointer fields should use `omitempty`: a nil pointer omits the field,
while a pointer to a zero value such as `false`, `0`, or `""` includes it.
`omitzero` behaves identically for pointers, so prefer `omitempty`.

Neither tag can send JSON `null`, so an update that must distinguish omission,
assignment, and removal follows `UpdateTeamRequest.RemoveParentTeam` in
`github/teams.go`: a Go-only removal flag tagged `json:"-"`, plus a
value-receiver `MarshalJSON` that also serves `*T` call sites. `new("")` sends
an empty string, not `null`.
```

gmlewis (changes requested):
Thank you, @austenstone.
One wording suggestion, otherwise, LGTM.

CONTRIBUTING.md
austenstone: Applied your wording exactly in [4eb1ab63](https://github.com/google/go-github/commit/4eb1ab636964293403f3c9f59a86bc98998b0ec8). Thanks!

gmlewis (approved):
Thank you, @austenstone!
LGTM.

CONTRIBUTING.md
```diff
@@ -413,8 +413,15 @@ type RepositoryRuleset struct {
 }
 ```
 
-For optional boolean fields where you need to distinguish between `false`
-and "not set", use `*bool` with `omitzero`.
+Optional pointer fields should use `omitempty`: a nil pointer omits the field,
+while a pointer to a zero value such as `false`, `0`, or `""` includes it.
+`omitzero` behaves identically for pointers, so prefer `omitempty`.
```
stevehipwell, line 418:
```suggestion
```
This looks like a duplication of the opening heading text.

CONTRIBUTING.md
```diff
@@ -413,8 +413,15 @@ type RepositoryRuleset struct {
...
+
+Neither tag can send JSON `null`, so an update that must distinguish omission,
+assignment, and removal follows `UpdateTeamRequest.RemoveParentTeam` in
+`github/teams.go`: a Go-only removal flag tagged `json:"-"`, plus a
+value-receiver `MarshalJSON` that also serves `*T` call sites. `new("")` sends
+an empty string, not `null`.
```
stevehipwell, line 424:
```suggestion
If you need to differentiate between an unset pointer to a basic type and a `nil` value you can add an un-marshaled struct field to control this behaviour and provide a custom `MarshalJSON` implementation for the struct (see `UpdateTeamRequest.RemoveParentTeam` in
`github/teams.go`).
```

stevehipwell (changes requested):
The code changes look good but I don't think the contributing changes are quite right.

PR #4544, opened 2026-09-11
feat: Add `image_gen` to `HostedRunner` and `UpdateHostedRunnerRequest`

### LINKED ISSUE #4532: `image_gen` missing from `HostedRunner` and `UpdateHostedRunnerRequest`
### Summary

`image_gen` is supported by the GitHub-hosted runner API on create, update, and read, but go-github only models it on create. As a result the field can be set but never read back. This seems to be a regression of functionality from https://github.com/google/go-github/pull/3973

In `github/actions_hosted_runners.go` on `main`:

| Type | Line | Has `image_gen`? |
|---|---|---|
| `HostedRunner` (response type) | 30 | no |
| `CreateHostedRunnerRequest` | 96 | yes — ``ImageGen *bool `json:"image_gen,omitempty"` `` |
| `UpdateHostedRunnerRequest` | 107 | no |

`HostedRunner` is the response type for the hosted-runner endpoints; returned by `ActionsService.GetHostedRunner`, `CreateHostedRunner`, `UpdateHostedRunner` and `DeleteHostedRunner`, by the corresponding `EnterpriseService` methods, and as the element type of `HostedRunners` from `ListHostedRunners`. There is currently no way to observe a runner pool's image-generation setting through the library.

### Requested change

1. Add ``ImageGen *bool `json:"image_gen,omitempty"` `` to `HostedRunner`. This is the more important of the two: without it, a caller cannot read the current value at all.
2. Add ``ImageGen *bool `json:"image_gen,omitempty"` `` to `UpdateHostedRunnerRequest`, since the PATCH body accepts the field.

Both are additive field additions to existing structs and are not breaking API changes. They do require `go generate ./...` to add the corresponding accessors to `github-accessors.go`.

### Evidence from the OpenAPI description

From [github/rest-api-description](https://github.com/github/rest-api-description), `descriptions/ghec/ghec.yaml`. All three declarations quoted below are identical in `descriptions/api.github.com/api.github.com.yaml`.

`components.schemas["actions-hosted-runner"]`, the response schema referenced by `GET`, `POST` and `PATCH` on the hosted-runner endpoints, declares:

```yaml
image_gen:
  type: boolean
  description: Whether custom image generation is enabled for the hosted runners.
```

Request bodies for the same endpoints declare it too.

`POST /orgs/{org}/actions/hosted-runners`:

```yaml
image_gen:
  description: Whether this runner should be used to generate custom images.
  type: boolean
  default: false
```

`PATCH /orgs/{org}/actions/hosted-runners/{hosted_runner_id}`:

```yaml
image_gen:
  description: Whether to enable image generation for this runner pool. When enabled, the runner pool is used to build and publish custom runner images.
  type: boolean
```

Docs: <https://docs.github.com/rest/actions/hosted-runners>

### Confirmed against a live response

I verified this against `api.github.com` on 2026-09-09, using an Enterprise Cloud organization. `image_gen` is present in every hosted-runner response shape:

```console
$ gh api orgs/ORG/actions/hosted-runners -X POST --input create.json
{"id":213,"name":"tf-imagegen-probe","image_details":{...},"platform":"linux-x64",
 "runner_group_id":1,"maximum_runners":1,"machine_size_details":{...},
 "public_ip_enabled":false,"public_ips":[],"last_active_on":"...",
 "status":"Provisioning","image_gen":false}

$ gh api orgs/ORG/actions/hosted-runners/213                 # GET   -> "image_gen":false
$ gh api orgs/ORG/actions/hosted-runners/213 -X PATCH \
      --input '{"maximum_runners":2}'                        # PATCH -> "image_gen":false
$ gh api orgs/ORG/actions/hosted-runners                     # LIST  -> "image_gen":false
```

So `POST`, `GET`, `PATCH` and `ListHostedRunners` all return the field, and it is absent from neither. 

One limitation on this evidence: the API validates `image_gen` on input rather than ignoring it.

Creating a runner with `"image_gen": true` in that organization returns:

```
422 Custom image generation is disabled for this organization by your enterprise administrator.
```

So a pointer (`*bool`) could be the right choice: it keeps an absent field distinguishable from an explicit `false` for any deployment or API version that does not return it.

### Why this matters

The Terraform provider for GitHub ([integrations/terraform-provider-github](https://github.com/integrations/terraform-provider-github)) exposes a `github_actions_hosted_runner` resource. It can send `image_gen` on create via `CreateHostedRunnerRequest`, but has nothing to read it back from, so `terraform import` of an existing runner silently records the attribute as `false`. Because the attribute forces replacement, a subsequent apply that sets it to `true` destroys and recreates the runner pool rather than recognising it was already enabled.

### Related, out of scope for this issue

The PATCH body also declares `image_source` (`enum: [github, partner, custom]`), which `UpdateHostedRunnerRequest` does not model either. Not included in the ask above; noting it in case it is worth a separate change.

breken-ai:
Fixes #4532

## What

The GitHub-hosted runner API supports `image_gen` on create, update, and read (it was added for create in #3973), but go-github only models it on `CreateHostedRunnerRequest`. The field can be set at create time but never read back or patched: `HostedRunner` (the response type for `GetHostedRunner`, `CreateHostedRunner`, `UpdateHostedRunner`, `DeleteHostedRunner`, `ListHostedRunners`, and the `EnterpriseService` equivalents) and `UpdateHostedRunnerRequest` both lack it. As the issue notes, this breaks terraform-provider-github round-trips: an import records `false`, and an apply with `true` destroys and recreates the pool.

Verified against current `master` (2433615): `github/actions_hosted_runners.go` has `ImageGen` only on `CreateHostedRunnerRequest` (:103); `HostedRunner` (:30) and `UpdateHostedRunnerRequest` (:107) do not.

## Change

- `HostedRunner`: add `ImageGen *bool \`json:"image_gen,omitempty"\``
- `UpdateHostedRunnerRequest`: add `ImageGen *bool \`json:"image_gen,omitempty"\``
- `script/generate.sh` run per CONTRIBUTING - `github-accessors.go` and `github-accessors_test.go` regenerated (accessor + accessor test for both new fields; no hand-written surface beyond the two field lines)
- New round-trip marshal tests in the repo's standard `testJSONMarshal` style: `TestHostedRunner_Marshal` and `TestUpdateHostedRunnerRequest_Marshal`, each covering the empty struct and a fully populated struct including `image_gen`

## Testing

- `TestHostedRunner_Marshal`, `TestUpdateHostedRunnerRequest_Marshal`: PASS
- Existing hosted-runner tests re-run against the change (`TestActionsService_ListHostedRunners`, `TestActionsService_GetHostedRunner`, `TestActionsService_UpdateHostedRunner`, `TestActionsService_ListHostedRunnersIter`): PASS
- `gofmt -l github/`: clean
- Disclosure: the full `github` package test binary OOMs at compile time in this memory-constrained sandbox (2 GB), so the complete suite was not run locally; the targeted tests above all pass and CI will cover the rest.

> Built by breken, your AI support engineer - breken.ai - this one's on us.

### DIFF
--- a/github/actions_hosted_runners.go
+++ b/github/actions_hosted_runners.go
@@ -38,6 +38,7 @@ type HostedRunner struct {
 	MaximumRunners     *int64                   `json:"maximum_runners,omitempty"`
 	PublicIPEnabled    *bool                    `json:"public_ip_enabled,omitempty"`
 	PublicIPs          []*HostedRunnerPublicIP  `json:"public_ips,omitempty"`
+	ImageGen           *bool                    `json:"image_gen,omitempty"`
 	LastActiveOn       *Timestamp               `json:"last_active_on,omitempty"`
 }
 
@@ -112,6 +113,7 @@ type UpdateHostedRunnerRequest struct {
 	Size           *string `json:"size,omitempty"`
 	ImageID        *string `json:"image_id,omitempty"`
 	ImageVersion   *string `json:"image_version,omitempty"`
+	ImageGen       *bool   `json:"image_gen,omitempty"`
 }
 
 // validateCreateHostedRunnerRequest validates the provided CreateHostedRunnerRequest to ensure
--- a/github/actions_hosted_runners_test.go
+++ b/github/actions_hosted_runners_test.go
@@ -1159,3 +1159,75 @@ func TestActionsService_DeleteHostedRunnerCustomImageVersion(t *testing.T) {
 		return client.Actions.DeleteHostedRunnerCustomImageVersion(ctx, "o", 1, "1.0.0")
 	})
 }
+
+func TestHostedRunner_Marshal(t *testing.T) {
+	t.Parallel()
+	testJSONMarshal(t, &HostedRunner{}, "{}")
+
+	u := &HostedRunner{
+		ID:            new(int64(5)),
+		Name:          new("My hosted ubuntu runner"),
+		RunnerGroupID: new(int64(2)),
+		Platform:      new("linux-x64"),
+		ImageDetails: &HostedRunnerImageDetail{
+			ID:          new("ubuntu-20.04"),
+			SizeGB:      new(int64(86)),
+			DisplayName: new("20.04"),
+			Source:      new("github"),
+			Version:     new("latest"),
+		},
+		MachineSizeDetails: &HostedRunnerMachineSpec{ID: "4-core", CPUCores: 4, MemoryGB: 16, StorageGB: 150},
+		Status:             new("Ready"),
+		MaximumRunners:     new(int64(10)),
+		PublicIPEnabled:    new(true),
+		PublicIPs:          []*HostedRunnerPublicIP{{Enabled: true, Prefix: "20.80.208.150", Length: 31}},
+		ImageGen:           new(true),
+		LastActiveOn:       &Timestamp{referenceTime},
+	}
+
+	want := `{
+		"id": 5,
+		"name": "My hosted ubuntu runner",
+		"runner_group_id": 2,
+		"platform": "linux-x64",
+		"image_details": {"id": "ubuntu-20.04", "size_gb": 86, "display_name": "20.04", "source": "github", "version": "latest"},
+		"machine_size_details": {"id": "4-core", "cpu_cores": 4, "memory_gb": 16, "storage_gb": 150},
+		"status": "Ready",
+		"maximum_runners": 10,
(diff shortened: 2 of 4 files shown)


gmlewis:
@breken-ai - please solve the issue reported here: https://github.com/google/go-github/pull/4544/checks?check_run_id=103155993223
and then we can proceed with this PR.

Not-Dhananjay-Mishra:
@gmlewis Looks like this PR can be closed 🙂

PR #4545, opened 2026-09-11
fix: Escape package names in UsersService methods

pucedoteth:
`UsersService.GetPackage` escapes `packageName`, but the other seven package methods interpolate it raw, so a name containing a slash produces a malformed path. Container and scoped npm package names routinely contain one.

`OrganizationsService` already escapes it in all seven equivalent methods and documents it, so this just brings the user methods in line: 12 call sites escaped, plus the same doc note.

The existing tests missed this because only the `GetPackage` test uses a slashed name (`hello/hello_docker`); the rest use `hello_docker`. The new test uses a slashed name against every affected method — all ten sub-cases fail without the change and pass with it.

AI assistance: Claude Code found the inconsistency, made the edits, and drafted the test and this description. I reviewed the diff and ran `script/fmt.sh`, `script/generate.sh` (no drift), `script/lint.sh` and `script/test.sh`.

### DIFF
--- a/github/users_packages.go
+++ b/github/users_packages.go
@@ -49,6 +49,8 @@ func (s *UsersService) ListPackages(ctx context.Context, user string, opts *Pack
 // GetPackage gets a package by name for a user. Passing the empty string for "user" will
 // get the package for the authenticated user.
 //
+// Note that packageName is escaped for the URL path so that you don't need to.
+//
 // GitHub API docs: https://docs.github.com/rest/packages/packages?apiVersion=2022-11-28#get-a-package-for-a-user
 //
 // GitHub API docs: https://docs.github.com/rest/packages/packages?apiVersion=2022-11-28#get-a-package-for-the-authenticated-user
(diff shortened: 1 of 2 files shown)


gmlewis:
@pucedoteth - for the CLA issue, chances are high that Anthropic will never sign the CLA and the following contributors were found for this pull request:

✅ https://github.com/google/go-github/commit/f02b3f20a67d31d0f06d988e6ae145331e8e2df9 Author: @pucedoteth <119044801+pucedoteth​@users.noreply.github.com>
❌ https://github.com/google/go-github/commit/f02b3f20a67d31d0f06d988e6ae145331e8e2df9 Co-Author: <no****y​@anthropic.com>

So I recommend that you CLOSE this PR and create a new PR where YOU are the only committer.

pucedoteth:
Thanks @gmlewis — closing in favour of #4546, which has the same change with me as the only committer. The AI disclosure is in that PR's description instead, per CONTRIBUTING.

PR #4546, opened 2026-09-11
fix: Escape package names in UsersService methods

pucedoteth:
`UsersService.GetPackage` escapes `packageName`, but the other seven package methods interpolate it raw, so a name containing a slash produces a malformed path. Container and scoped npm package names routinely contain one.

`OrganizationsService` already escapes it in all seven equivalent methods and documents it, so this just brings the user methods in line: 12 call sites escaped, plus the same doc note.

The existing tests missed this because only the `GetPackage` test uses a slashed name (`hello/hello_docker`); the rest use `hello_docker`. The new test uses a slashed name against every affected method — all ten sub-cases fail without the change and pass with it.

AI assistance: Claude Code found the inconsistency, made the edits, and drafted the test and this description. I reviewed the diff and ran `script/fmt.sh`, `script/generate.sh` (no drift), `script/lint.sh` and `script/test.sh`.

### DIFF
--- a/github/users_packages.go
+++ b/github/users_packages.go
@@ -49,7 +49,7 @@ func (s *UsersService) ListPackages(ctx context.Context, user string, opts *Pack
 // GetPackage gets a package by name for a user. Passing the empty string for "user" will
 // get the package for the authenticated user.
 //
-// Note that packageName is escaped for the URL path so that you don't need to.
+// Note: the package name is URL path escaped for you. See: https://pkg.go.dev/net/url#PathEscape .
 //
 // GitHub API docs: https://docs.github.com/rest/packages/packages?apiVersion=2022-11-28#get-a-package-for-a-user
 //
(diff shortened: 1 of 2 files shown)


gmlewis (approved):
Thank you, @pucedoteth!
LGTM.

github/users_packages.go
```diff
@@ -49,6 +49,8 @@ func (s *UsersService) ListPackages(ctx context.Context, user string, opts *Pack
 // GetPackage gets a package by name for a user. Passing the empty string for "user" will
 // get the package for the authenticated user.
 //
+// Note that packageName is escaped for the URL path so that you don't need to.
```
Not-Dhananjay-Mishra, line 52:
```suggestion
// Note: the packageName is URL path escaped for you. See: https://pkg.go.dev/net/url#PathEscap .
```
For consistency with rest of the repo. 
https://github.com/google/go-github/blob/2433615db90de9d8d7546e44591561aef8bf3bbc/github/repos_rules.go#L15-L22

Same applies to other occurrences in this PR.

pucedoteth:
Good catch, thanks — applied to all 8 occurrences in `5d36ebb`.

I'd matched `orgs_packages.go`, the sibling file, but you're right that it's the outlier: repo-wide the `// Note: the X name is URL path escaped for you.` form appears 55 times across 7 files, against 7 in that one file.

Two small deviations from the literal suggestion, both to match the source you linked:

- The suggested URL ends `#PathEscap`; I used `#PathEscape` so the link resolves.
- I used "the package name" rather than "the packageName", matching "the branch name" and "the environment name" in the existing notes.

The result is byte-identical to the `repos_rules.go` note apart from the noun.

Not-Dhananjay-Mishra (approved):

PR #4548, opened 2026-09-12
refactor!: Pass the secret scanning body types by value and rename them to `...Request`

JamBalaya56562:
Updates #3644

This converts the two remaining `SecretScanningService` body types, removing 4 entries from the `paramcheck` exception lists in `.golangci.yml` (2 from `body-allowed-pointer-types` and 2 from `body-allowed-wrong-names`).

Checking the types against the OpenAPI descriptions turned up two gaps that are fixed in separate commits:

1. The request schema for `PATCH /repos/{owner}/{repo}/secret-scanning/alerts/{alert_number}` has five properties (`state`, `resolution`, `resolution_comment`, `assignee`, `validity`) with `anyOf: [{required: [state]}, {required: [assignee]}, {required: [validity]}]`. `SecretScanningAlertUpdateOptions` only had the first three and declared `State string \`json:"state"\``, so an assignee-only or validity-only update was not possible. `State` is now `*string` with `omitempty`, and `Assignee` and `Validity` are added. As elsewhere in this package (e.g. `IssueRequest.Assignee`), `omitempty` cannot express the schema's "set to `null` to clear" case; that is unchanged here.
2. The secret scanning alert response schemas have `assigned_to`, `closure_request_comment`, `closure_request_reviewer`, `closure_request_reviewer_comment`, `provider`, `provider_slug` and `metadata`, none of which were on `SecretScanningAlert`. They are added as optional fields, with a new `SecretScanningAlertMetadata` type for the `metadata` entries.

`UpdateSecretScanningPatternConfigsRequest` stays shared between `UpdatePatternConfigsForEnterprise` and `UpdatePatternConfigsForOrg`; the two request schemas are identical. The pattern configuration tests now also assert the request body, which they did not before.

BREAKING CHANGE: `SecretScanningAlertUpdateOptions` and `SecretScanningPatternConfigsUpdateOptions` are renamed to `UpdateSecretScanningAlertRequest` and `UpdateSecretScanningPatternConfigsRequest`; `SecretScanningService.UpdateAlert`, `UpdatePatternConfigsForEnterprise` and `UpdatePatternConfigsForOrg` now take them by value instead of by pointer; `UpdateSecretScanningAlertRequest.State` is now `*string`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -236,8 +236,6 @@ linters:
             - RepositoryCreateForkOptions
             - RequiredStatusChecksRequest
             - SCIMUserAttributes
-            - SecretScanningAlertUpdateOptions
-            - SecretScanningPatternConfigsUpdateOptions
             - SourceImportAuthor
             - Subscription
             - TeamAddTeamMembershipOptions
@@ -265,8 +263,6 @@ linters:
             - RepositoryAddCollaboratorOptions
             - RepositoryContentFileOptions
             - RepositoryCreateForkOptions
-            - SecretScanningAlertUpdateOptions
-            - SecretScanningPatternConfigsUpdateOptions
             - TeamAddTeamMembershipOptions
             - TeamAddTeamRepoOptions
             - TeamProjectOptions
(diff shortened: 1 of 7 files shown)


gmlewis (approved):
Thank you, @JamBalaya56562!
LGTM.

PR #4553, opened 2026-09-14
fix: Re-add ldapdb for ghes team creation

stevehipwell:
This PR re-adds the `LDAPDN` field to the `CreateTeamRequest` struct for usage by gitHub Enterprise Server. This Field was incorrectly removed in #4446.


gmlewis (approved):
Thank you, @stevehipwell!
LGTM.
Merging.

PR #4556, opened 2026-09-15
fix: Reject an upload URL whose host differs from the configured upload host

sushant-me:
`UploadReleaseAssetFromRelease` takes `release.UploadURL` from the server's response, and `NewUploadRequest` lets an absolute URL replace the client's configured upload host entirely (`url.URL.Parse` discards the receiver for an absolute reference). A response naming a different host therefore received the artifact **and** the caller's `Authorization` header, while the caller's own `WithURLs`/`WithEnterpriseURLs` upload configuration was silently ignored.

```go
release := &RepositoryRelease{UploadURL: "https://example.invalid/upload{?name,label}"}
// previously: POST https://example.invalid/upload?name=n.txt  (with the bearer token)
// now:        error: upload URL host "example.invalid" does not match the
//             client's configured upload host "uploads.github.com"
```

The check runs only for *absolute* values; relative URLs — what GitHub actually returns, and the case the existing tests cover — are untouched. It compares against `c.uploadURL` rather than `c.baseURL`, because `uploads.github.com` and `api.github.com` differ by design.

This does not overlap with #4363/#4364/#4366: those scope the token for a foreign origin but would still POST the artifact body there. Rejecting the request keeps both on the configured host.

`TestRepositoriesService_UploadReleaseAssetFromRelease_ForeignHostIsRejected` fails on `master` (`expected an error for an upload URL naming a foreign host, got nil`) and passes with this change. `go test ./github/` and `script/lint.sh` both pass.

**Assisted contribution:** an AI tool drafted the change, the test, and this description. I reviewed the rendered diff, ran the full `github` package test suite and `script/lint.sh`, and verified the new test fails without the fix.


sushant-me:
@googlebot I signed it!

gmlewis:
@sushant-me - can you please increase the CodeCov results for this critical piece of code?

sushant-me:
@gmlewis done - added `TestRepositoriesService_UploadReleaseAssetFromRelease_MalformedUploadURL`, which exercises the `url.Parse` error return (a URL with an ASCII control character makes net/url refuse it). That was the uncovered line in `repos_releases.go`, so patch coverage should now be 100%. `go test ./github/` and `./script/lint.sh` both pass locally.

gmlewis (approved):
Thank you, @sushant-me!
LGTM.

sushant-me:
Thanks for the LGTM, @gmlewis! Everything is green CLA, codecov patch/project, and the full test matrix and the branch is current with master, so this is ready for a second LGTM whenever someone has bandwidth. Happy to make any adjustments if something needs changing.

github/repos_releases.go
```diff
@@ -504,6 +505,25 @@ func (s *RepositoriesService) UploadReleaseAssetFromRelease(
...
+		// configured with; that is uploads.github.com rather than api.github.com in
+		// the default configuration, which is why the comparison is against
+		// uploadURL and not baseURL.
+		u, err := url.Parse(uploadURL)
+		if err != nil {
+			return nil, nil, err
+		}
+		if !strings.EqualFold(u.Host, s.client.uploadURL.Host) {
+			return nil, nil, fmt.Errorf(
+				"upload URL host %q does not match the client's configured upload host %q",
```
Not-Dhananjay-Mishra:
```suggestion
				"upload URL host %v does not match the client's configured upload host %v",
```

github/repos_releases_test.go
```diff
@@ -850,6 +851,53 @@ func TestRepositoriesService_UploadReleaseAssetFromRelease_AbsoluteTemplate(t *t
...
+	release := &RepositoryRelease{UploadURL: evil.URL + "/upload{?name,label}"}
+	ctx := t.Context()
+	_, _, err := client.Repositories.UploadReleaseAssetFromRelease(
+		ctx, release, &UploadOptions{Name: "n.txt"}, reader, size,
+	)
+	if err == nil {
+		t.Fatal("expected an error for an upload URL naming a foreign host, got nil")
+	}
+	if leaked != 0 {
+		t.Fatalf("upload reached the foreign host %d time(s); the token and body must never be sent there", leaked)
```
Not-Dhananjay-Mishra:
```suggestion
		t.Fatalf("upload reached the foreign host %v time(s); the token and body must never be sent there", leaked)
```

sushant-me:
@Not-Dhananjay-Mishra thanks for the review — both suggestions are applied in 8590220: `%v` for the two host verbs in the error, and `%v` for the leak counter in the test failure.

Re-verified locally: `gofmt -l` clean, `go build ./...` ok, and all nine `TestRepositoriesService_UploadReleaseAssetFromRelease` subtests pass. CI is re-running now. Happy to change anything else.

Not-Dhananjay-Mishra (approved):
LGTM

gmlewis:
Merging.

PR #4560, opened 2026-09-16
feat: Support explicit custom property values

### LINKED ISSUE #4559: Support require_explicit_values on custom properties
## Problem

GitHub's organization custom-property API returns and accepts the optional `require_explicit_values` boolean, but `CustomProperty` cannot read or send it. Clients therefore cannot preserve or manage whether repositories must set an explicit property value.

GitHub documents the field on the organization custom-property schema and create/update endpoints: https://docs.github.com/en/rest/orgs/custom-properties#create-or-update-a-custom-property-for-an-organization

## Proposed change

Add an optional boolean field to `CustomProperty` and cover JSON decoding plus the batch and single-property request bodies with unit tests.

I plan to submit a focused patch.

devantler:
## Summary

`CustomProperty` could not represent GitHub’s optional `require_explicit_values` setting, preventing clients such as the Terraform provider from preserving or managing it.

This adds the optional field, keeps omitted and explicit `false` values distinct, and covers response decoding plus batch and single-property request bodies.

Closes https://github.com/google/go-github/issues/4559

## Assisted contribution

GitHub Copilot substantially assisted with the code, tests, and prose. I reviewed, understood, and tested every change.


devantler:
CLA is signed.

CONTRIBUTING.md
```diff
@@ -414,7 +414,8 @@ type RepositoryRuleset struct {
 ```
 
 For optional boolean fields where you need to distinguish between `false`
-and "not set", use `*bool` with `omitzero`.
+and "not set", use `*bool` with `omitempty`. A non-nil pointer preserves an
+explicit `false`, while a nil pointer is omitted.
```
gmlewis: No, please do not change this or AGENTS.md.

According to: https://go.dev/doc/go1.24#encodingjsonpkgencodingjson

```
When marshaling, a struct field with the new omitzero option in the struct field tag will be omitted if its value is zero. If the field type has an IsZero() bool method, that will be used to determine whether the value is zero. Otherwise, the value is zero if it is [the zero value for its type](https://go.dev/ref/spec#The_zero_value). The omitzero field tag is clearer and less error-prone than omitempty when the intent is to omit zero values. In particular, unlike omitempty, omitzero omits zero-valued [time.Time](https://go.dev/pkg/time#Time) values, which is a common source of friction.

If both omitempty and omitzero are specified, the field will be omitted if the value is either empty or zero (or both).
```

So a `*bool` with `omitzero` is perfectly valid. If the pointer is nil, the value is not included. However, if the value is `new(false)` or `new(true)`, the value is included.

gmlewis (changes requested):

CONTRIBUTING.md
devantler: Done. I restored `AGENTS.md` and `CONTRIBUTING.md` unchanged and switched `RequireExplicitValues` back to `omitzero`. The repository's `structfield` linter currently rejects `omitzero` on pointer primitives, so I added a field-specific exception for `CustomProperty.RequireExplicitValues`; the full format, generation, test, and lint checks now pass.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app) (GPT 5.6 Sol Extra High)_

gmlewis: > The repository's `structfield` linter currently rejects `omitzero` on pointer primitives, so I added a field-specific exception for `CustomProperty.RequireExplicitValues`; the full format, generation, test, and lint checks now pass.

OK, that's really weird... we spent a lot of time working on that linter and I _thought_ we got it right but maybe it needs more investigation... sorry about that! I'll try to look into it this week.

gmlewis:
According to the CLA checker:

```
The following contributors were found for this pull request:

   ✅ https://github.com/google/go-github/commit/dac265e8c161e10b58740ed4a426eccf1a8cecee Author: @devantler <niko********amm​@icloud.com>
   ❌ https://github.com/google/go-github/commit/dac265e8c161e10b58740ed4a426eccf1a8cecee Co-Author: @Copilot <223556219+Copilot​@users.noreply.github.com>
```

OK, so to resolve this problem, you are either going to need to recreate the commits locally so that ONLY YOU performed all the commits, and then FORCE-PUSH to this PR to eliminate the Copilot commits, or you are going to have to close this PR and open a brand new one where only you are the committer.

Sorry for the hassle, but we can't move forward with this PR until the CLA issue is resolved, per Google Open Source rules.

CONTRIBUTING.md
devantler: No worries. It is totally fine with me if you want to take over and polish it a bit before merge in this PR/branch.

Just tag me if you need me to do something :-)

devantler:
Replaced by https://github.com/google/go-github/pull/4561 with a newly created sole-author commit, as requested, so the CLA check can evaluate only the signed contributor. No force-push was used.

PR #4562, opened 2026-09-17
fix: Reject a metrics download URL whose host differs from the client's

prasanna8585:
## Problem

`DownloadCopilotMetrics` and the six current `Download*Metrics` methods (via the shared `fetchMetricsReport`) take a download URL that callers are documented to read out of a prior `Get*MetricsReport` response's `DownloadLinks` field, not one they construct themselves. Both built the outgoing request directly from that URL and sent it through the client's own credentialed `http.Client`, with no check that the URL's host matched the client's configured host.

That client's auth transport (`WithAuthToken`'s `RoundTripper`) attaches the caller's `Authorization` header to every request it sends, regardless of destination host. A report response naming a foreign host — from a compromised/malicious GitHub Enterprise Server instance, or any position able to influence that response — would have had the caller's bearer token, and the downloaded report body, sent to that host.

This is the same vulnerability class this repo already fixed twice this week: `bareDoUntilFound`'s cross-host check on 301 redirects, and `UploadReleaseAssetFromRelease`'s host check on the release object's `UploadURL` field. Neither mitigation covers this path, since both functions here build and send their request independently of those.

## Fix

Parse the download URL and compare its host against `client.baseURL.Host` before building the request, in both `fetchMetricsReport` and `DownloadCopilotMetrics`, following the same pattern already established by the `UploadURL` fix (`url.Parse` + `strings.EqualFold` + a descriptive error).

## Testing

- `gofmt -l` clean on both changed files.
- This sandbox couldn't fetch the toolchain the module's `go.mod` requires (`go >= 1.26.0`, network-restricted), so I verified the exact fix logic (URL parsing, host comparison) in a standalone, dependency-free Go program: confirmed the leak is real pre-fix (a simulated auth header reaches an attacker-controlled server), confirmed the fix blocks it with zero header leakage, and confirmed a legitimate same-host request still succeeds.
- Added `TestCopilotService_fetchMetricsReport_ForeignHostIsRejected`, `TestCopilotService_DownloadCopilotMetrics_ForeignHostIsRejected`, and `TestCopilotService_fetchMetricsReport_MalformedDownloadURL` to `copilot_test.go`, matching this file's existing `httptest.NewServer` + `Transport`-wrapping pattern. These are syntactically valid (`gofmt`-checked) but I couldn't run them against the toolchain for the reason above — please run in CI as part of review.
- Confirmed every existing test in this file already constructs download URLs as `client.baseURL.String() + path`, so this shouldn't change any existing test's behavior.


gmlewis:
Thanks for the PR. The underlying observation is right — fetchMetricsReport and DownloadCopilotMetrics do send the caller's token to whatever host is named, since they bypass bareDoUntilFound's guard entirely, and it's fair to note that.

I don't want to land it as written, for three reasons:

1. It doesn't close the hole it describes. Both functions dispatch through s.client.client / BareDo, and credentials are injected in the RoundTripper (github/github.go:612-615) — so they're re-added on every redirected hop. A download URL on the configured host that 302s to another host still leaks the token with this patch applied. Guarding only the caller's URL string doesn't address that, and the new tests don't cover it.
2. The repo already has the right mechanism for download paths. bareDoUntilFound refuses cross-host 301s, and DownloadReleaseAsset fetches redirect targets with a caller-supplied uncredentialed client. If we want a guard here, it should route through bareDoUntilFound and surface the redirect target the way DownloadReleaseAsset does, rather than adding a fourth pre-flight string check.
3. The boundary is under active discussion in #4366, including the allowed-origin set (baseURL + uploadURL) and comparison granularity. This patch settles both unilaterally as host-only against baseURL — which doesn't even match #4556's comparison against uploadURL.Host, and is the same predicate divergence flagged in review on #4363.

Also worth noting on the tests: url.Parse rejecting \x7f is already covered at copilot_test.go:4114 with "\n" (and failed before your change too), and the leakedAuth assertions can't fire once the function returns before sending. And the "fixed twice this week" framing is off — the redirect check is #4171 from April.

My suggestion is to hold this until #4366 settles the policy, then implement it once at the injection point.

gmlewis:
See: https://github.com/google/go-github/issues/4366#issuecomment-5717105620

prasanna8585:
Thanks for the detailed review, and for catching the redirect-hop gap specifically that's a real miss on my part. I'd traced the RoundTripper injection at github.go:612-615 for the report and still didn't carry it through to what happens after a redirect, which is exactly the case that matters here. Fair catch.

Agreed on all three points:

- A same-host URL that 302s elsewhere still leaks the token with this patch, since the check only validates the caller-supplied string, not the redirect target. The tests don't cover that because the patch doesn't handle it.
- bareDoUntilFound and DownloadReleaseAsset's uncredentialed-redirect-following pattern is the right mechanism for this, and I should have routed through it instead of adding a separate check.
- I didn't realize #4366 was actively deciding the allowed-origin set, and picking baseURL-only here inconsistent with #4556's uploadURL.Host comparison would've unilaterally pre-empted that. That's on me for not checking open issues first.

Also noted: the \x7f test does duplicate copilot_test.go:4114, and the "fixed twice this week" framing was wrong #4171 is from April, not this week. Thanks for the correction.

Happy to hold this per your suggestion. I'll watch #4366 and come back with a single implementation at the injection point once the origin-set policy is settled, matching whatever bareDoUntilFound/DownloadReleaseAsset end up doing. Closing/holding this PR in the meantime let me know if you'd rather I close it now or leave it open as a reference until #4366 lands.

gmlewis:
> I'll watch #4366 and come back with a single implementation at the injection point once the origin-set policy is settled, matching whatever bareDoUntilFound/DownloadReleaseAsset end up doing. Closing/holding this PR in the meantime let me know if you'd rather I close it now or leave it open as a reference until #4366 lands.

Please take a look at #4564 and make sure you are happy with that impementation, as I believe it should solve the problem across the entire client library.

prasanna8585:
Reviewed #4564 in detail it closes what we reported here, including the redirect-hop gap you caught us on, and does it more thoroughly than this patch would have (verified the new fetchMetricsReport comment and TestCopilotService_DownloadMetrics_ForeignHostGetsNoCredentials directly).

Given #4564 supersedes this and lists it in Closes:, go ahead and close #4562 now rather than holding it open no need to keep it as a live reference once #4564 lands. Thanks for looping us in on the broader fix.

Issue #4023, opened 2026-02-19
Extend CONTRIBUTING.md with common code guidelines


alexandear:
During reviews, I often find myself repeating the same comments and explanations. Many of these points have already been discussed numerous times. To help new contributors better understand the codebase and reduce the review burden, I propose extending the CONTRIBUTING.md document with a concise section on style and code guidelines.

The following topics could be documented:

1. Naming conventions
   - File names.
   - Receiver names.
   - Request option types.
   - Request body types.
   - Response types.
   - Method and variable naming.
   - Common variable names such as `owner`, `repo`, `org`.
   - Naming for enterprise & organization scoped, see #3761.
   - Common structs such as `ListOptions`.

2. JSON tags for structs that represent a request body:

   - When using `omitempty`, the field type should be a pointer.
   - Use `omitzero` for structs and slices where appropriate.

3. URL tags for structs that represent query parameters:

   - When using `omitempty`, the field type should be non-pointer.

4. Pagination:

   - Proper use of `ListOptions` and `ListCursorOptions`.

5. Common types:

   - For example, `ID` should consistently use `int64`.

6. Generation patterns.

Some of these rules could eventually be enforced by linters (we already have `sliceofpointers`, `structfield`), but documenting them first would make expectations explicit and help contributors avoid common mistakes.

Additionally, we could analyze the repository to identify recurring review comments and common mistakes. Based on that analysis, we could expand `CONTRIBUTING.md` to address frequently misunderstood patterns and conventions.

This would make expectations clearer for contributors and help maintain consistency across the project.

gmlewis:
Sounds good to me.

gmlewis:
Fixed by #4339.

Issue #3336, opened 2024-10-25
RemoveReviewers API: `omitempty` on required `reviewers` field causes spurious 422s


nagl-resourcely:
# Summary
Calling `RemoveReviewers` when you want to only remove teams elicits a 422, due to how `omitempty` prevents the golang client from serializing the empty `reviewers` slice. This field is required on the server, so the golang client needed to send `reviewers: []`.

# Workaround
Work around this by exploiting delete idempotency =) Just pass it a bogus username alongside real team slugs.

gmlewis:
I'm confused by this report.

I'm looking at this endpoint:
https://github.com/google/go-github/blob/792bd15b871a8a6aea57cca226ef7a9e85e2955a/github/pulls_reviewers.go#L73-L86

and am not seeing the slice you are talking about.

I do, however, see these slices:
https://github.com/google/go-github/blob/792bd15b871a8a6aea57cca226ef7a9e85e2955a/github/pulls_reviewers.go#L13-L24

which are used with the `RequestReviewers` method.

Can you please clarify your request?
And while I'm here, would you like to submit a PR to resolve the issue?

gmlewis:
EDIT: Ah!  I see it now.
OK, gotcha.

Would you like to create a PR to fix this issue, @nagl-resourcely ?

nagl-resourcely:
Heh, that probably would have been faster than typing up the issue. Sure - I'll send one after the workday ends.

PR #4576, opened 2026-09-18
feat: Check request body fields against OpenAPI

gmlewis:
Checking new PRs against the official GitHub v3 API for required fields in body parameters is extremely time-consuming, monotonous, and error-prone. This new tool helps contributors and maintainers alike by automating the mechanically-verifiable parts.

This PR adds `tools/schemafields`, run by `script/check-schema-fields.sh` and wired into `script/lint.sh` and the `linter` workflow. It checks each request body struct field against the request body schema of the operations that send it, mapping structs to operations via the `//meta:operation` annotation and the by-value `body` parameter that `paramcheck` requires, so new endpoints are covered with no new annotation.

No breaking API changes are made in this PR, but an upcoming PR will `-fix` the problems found.

AI assistance was used to write this PR.

Supersedes #4375 (closed), which required a `//meta:schema` annotation on every struct.
Relates to #3644 and #4023, and to the `required-but-omittable` bugs in #3336, #3226 and #2909.


tools/schemafields/check.go
```diff
@@ -0,0 +1,426 @@
...
+		case !slices.Contains(e.entries, line):
+			e.entries = append(e.entries, line)
+		}
+	}
+	return e, nil
+}
+
+// suppress reports whether key is grandfathered.
+func (e *exceptions) suppress(key string) bool {
+	return e != nil && slices.Contains(e.entries, key)
```
Not-Dhananjay-Mishra: Can we replace checking `slices.Contains` repeatedly with a map initialized once and then check against the map?
I guess we can also remove `slices.Contains` from `obsolete` too and use map.

tools/schemafields/scan.go
```diff
@@ -0,0 +1,371 @@
...
+	_, ok := e.(*ast.StarExpr)
+	return ok
+}
+
+// structInfo describes a struct declared in the scanned package.
+type structInfo struct {
+	name   string
+	file   string
+	line   int
+	fields []*fieldInfo
```
Not-Dhananjay-Mishra, line 95: I guess `structInfo.fields` is only accessed through `field()` method.
```suggestion
	fields map[string]*fieldInfo
}

// field returns the field with the given JSON name, if the struct has one.
func (s *structInfo) field(jsonName string) *fieldInfo {
	return s.fields[jsonName]
}
```

At L311
```go
si.fields = append(si.fields, fi)
```
can we replaced with 
```go
si.fields[fi.jsonName] = fi
```

tools/schemafields/check.go
```diff
@@ -0,0 +1,426 @@
...
+			c.checkRequired(f, si, len(checkable), nullable)
+		case required == 0:
+			c.checkOptional(f, si)
+		default:
+			// Required by some operations and optional in others: conditional
+			// requiredness that one Go type cannot express. Leave it to
+			// author judgement.
+			c.stats.fieldsConditional++
+		}
+	}
```
Not-Dhananjay-Mishra: We can get rid of second loop.
```suggestion
func (c *checker) checkStruct(si *structInfo, uses []*structUse) {
	for _, f := range si.fields {
		c.stats.fieldsChecked++

		required := 0
		checkable := 0
		nullable := false
		inSchema := false
		for _, u := range uses {
			meta, ok := u.schema.props[f.jsonName]
			if !ok {
				continue
			}
			inSchema = true
			if meta.readOnly {
				// A readOnly property is never sent, so optionality does not
				// apply to it.
				continue
			}
			if u.schema.required[f.jsonName] {
				required++
			}
			nullable = nullable || meta.nullable
			checkable++
		}

		if !inSchema {
			c.add(&diagnostic{
				sev: sevWarn, rule: ruleNotInSchema, file: f.file, line: f.line,
				owner: si.name, field: f.goName, info: f,
				message: fmt.Sprintf("Go field is not in the request body schema of any of its %v operation(s)", len(uses)),
			})
			continue
		}
		if checkable == 0 {
			continue // Present in the schema, but readOnly in every operation.
		}

		switch required {
		case checkable:
			c.checkRequired(f, si, checkable, nullable)
		case 0:
			c.checkOptional(f, si)
		default:
			// Required by some operations and optional in others: conditional
			// requiredness that one Go type cannot express. Leave it to
			// author judgement.
			c.stats.fieldsConditional++
		}
	}
```

tools/schemafields/exceptions.txt
Not-Dhananjay-Mishra, line 1: Can we use `.yml`/`.yaml` file instead of `.txt`

tools/schemafields/scan.go
gmlewis: `structInfo.fields` is also used by `checkStruct`.
Also, turning it into a map would lose the declaration order which would mess up the stable sort.
But most critically, it would silently drop a field when two fields of one struct share a JSON name, so that field would never be checked at all.
So I didn't make this change.

tools/schemafields/exceptions.txt
gmlewis: The file is a flat one-key-per-line list plus a comment header that the tool preserves verbatim when it rewrites the file. YAML buys nothing here but costs a parser dependency (the module currently requires only go-cmp) and comment round-tripping (yaml.v3 drops comments, so -fix/-write-exceptions would lose the header). If the format ever needs per-entry metadata, we can change it. But for now, txt is the better choice, IMO.

gmlewis:
@Not-Dhananjay-Mishra - thank you for the review! I addressed two of your comments, and pushed back on the other two :smile:. PTAL.

tools/schemafields/check.go
```diff
@@ -0,0 +1,431 @@
...
+			message: fmt.Sprintf("%q is optional in the schema, but the pointer has no omitempty, so a nil value is sent as null", f.jsonName),
+		})
+	case f.isStruct:
+		// CONTRIBUTING.md prefers omitzero over a pointer for structs, and it is the
+		// only omit option that works on a struct value.
+		c.add(&diagnostic{
+			sev: sevWarn, rule: ruleOptionalValueType, file: f.file, line: f.line,
+			owner: si.name, field: f.goName, info: f, action: &fixAction{addOmit: "omitzero"},
+			message: fmt.Sprintf("%q is optional in the schema, but the Go field is a value type, so it is always sent: add omitzero", f.jsonName),
+		})
```
Not-Dhananjay-Mishra: I think our `structfield` linter will not allow this kind of case, where a non-pointer struct uses `omitzero`.
```go
type SomeRequest struct {
	OptionalStruct            Struct         `json:"optional_struct,omitzero"`
}
```

gmlewis: I think the reason that all tests pass is because our linters are guarded by the "github" package name.

Not-Dhananjay-Mishra: > I think the reason that all tests pass is because our linters are guarded by the "github" package name.

linters are not running inside `testdata/...` so that why linter didn't caught it.

tools/schemafields/testdata/golden/TestFix/github/runner_groups.go
```go
// StructTypeRequest has an optional property of a struct value type.
type StructTypeRequest struct {
	// Inner is optional and CONTRIBUTING.md asks for omitzero on structs.
	Inner InnerConfig `json:"inner,omitzero"`
}
```
I moved it outside temporary to test it, and linters start complaning.
```cmd
    tools/schemafields/runner_groups.go:23:2: change the "ResponseOnly" field type to "*string" in the struct "CreateRunnerGroupRequest" because its json tag uses "omitempty" (structfield)
        ResponseOnly string `json:"response_only,omitempty"`
        ^
    tools/schemafields/runner_groups.go:47:2: change the "Inner" field type to "*InnerConfig" in the struct "StructTypeRequest" (structfield)
        Inner InnerConfig `json:"inner,omitzero"`
```

I believe that we should suggest `%q is optional in the schema, but the Go field is a value type, so it is always sent: use pointer with omitempty`

gmlewis: You are absolutely right, @Not-Dhananjay-Mishra!

I believe this is now fixed and I removed the 19 false-positives that were caused by this error.

PTAL.

script/lint.sh
```diff
@@ -96,6 +96,14 @@ if [ -n "$CHECK_GITHUB_OPENAPI" ]; then
...
     fail
   fi
+
+  print_header "Checking request body fields against the OpenAPI schemas"
+  if script/check-schema-fields.sh; then
+    printf "${GREEN}✔ request body fields match the OpenAPI schemas${NC}\n"
+  else
+    printf "${RED}✘ request body fields disagree with the OpenAPI schemas${NC}\n"
+    fail
+  fi
```
Not-Dhananjay-Mishra: When running `script/lint.sh` locally, this doesn't get triggered because it requires the `CHECK_GITHUB_OPENAPI=1` environment variable. Is this intentional?

CONTRIBUTING.md
```diff
@@ -652,6 +660,43 @@ Its subcommands are:
 - `unused` - lists operations from `openapi_operations.yaml` that are not mapped
   from any methods.
 
+### tools/schemafields
+
+The `tools/schemafields` package checks Go request body struct field
+optionality against GitHub's OpenAPI request body schemas. It is run by
+`script/check-schema-fields.sh`, which `script/lint.sh` and the `linter`
+workflow call, so you rarely need to run it directly.
```
Not-Dhananjay-Mishra, line 668: I feel that this statement

> It is run by
> `script/check-schema-fields.sh`, which `script/lint.sh` and the `linter`
> workflow call, so you rarely need to run it directly.

 feels inconsistent since contributors need to set the environment variable first then only they can run `script/check-schema-fields.sh` via `script/lint.sh`.

Not-Dhananjay-Mishra (approved):
Just one small doubt. Otherwise, LGTM 🫠

script/lint.sh
gmlewis: Good call! I've removed it.

CONTRIBUTING.md
gmlewis: Agreed. Updated.

gmlewis:
Updated. Thank you, @Not-Dhananjay-Mishra!

PR #4578, opened 2026-09-19
docs: Fix go-githubauth GitHub App example

jferrl:
## Summary

The `go-githubauth` example does not compile. Copied out of the README and
built against go-github v92 and go-githubauth v1.8.0:

    ./main.go:7:2:   "strconv" imported and not used
    ./main.go:17:61: int does not satisfy githubauth.Identifier
                     (int missing in ~int64 | ~string)
    ./main.go:30:2:  declared and not used: client

`NewApplicationTokenSource` takes `~int64 | ~string`. The untyped constant
`1112` defaults to `int`, so the call never type-checks.

## Changes

- Pass the App's client ID (a `string`), which GitHub now recommends over the
  legacy app ID. The prose above the example notes both forms are accepted.
- Read the installation ID with `strconv.ParseInt`, so the existing `strconv`
  import is used.
- Fix the misindented closing brace.
- Add a trailing `// Use client...` comment, matching the ghinstallation
  example above.

`declared and not used: client` is left as is — the ghinstallation example
has the same shape, so it reads as the convention for these snippets.


gmlewis (approved):
Thank you, @jferrl!
LGTM.
Merging.

PR #4585, opened 2026-09-21
fix: Honor HTTP client timeouts in redirects

### LINKED ISSUE #4584: http.Client.Timeout is ignored by redirect-aware requests
## Bug

Requests made through the redirect-aware endpoint helpers do not honor the timeout configured on the client's `http.Client`.

`roundTripWithOptionalFollowRedirect` calls the configured transport directly. That preserves the ability to inspect 301/302 responses, but bypasses `http.Client.Do`, which is where `http.Client.Timeout` is applied. This affects helpers such as workflow log downloads, artifact downloads, archive links, and other callers of that helper when `WithHTTPClient` or `WithTimeout` is used.

## Reproduction

Use a custom `RoundTripper` that waits for `req.Context().Done()` and a client timeout of 10 ms. On current `master`, the helper invokes the transport with the caller context and waits for the transport's full delay instead of returning a deadline error. The same request through the client's no-redirect `http.Client` returns `context deadline exceeded` at the configured timeout.

## Expected behavior

Redirect-aware requests should preserve the existing no-follow behavior for the first response while honoring the configured HTTP client timeout.

sb123sb123:
Fixes #4584

## Cause

`roundTripWithOptionalFollowRedirect` called the configured transport's `RoundTrip` method directly. That preserves access to 301/302 responses, but bypasses `http.Client.Do`, so the configured `http.Client.Timeout` was never applied to these redirect-aware requests.

## Fix

Use the existing no-redirect `http.Client` for the request. It shares the configured transport and timeout and returns the first redirect response because its `CheckRedirect` returns `http.ErrUseLastResponse`; the existing redirect and host checks remain unchanged.

## Tests

- `go test -modfile=go.test.mod ./github -run 'Test(RoundTripWithOptionalFollowRedirect|ActionsService_DownloadArtifact|ActionsService_GetWorkflowJobLogs|ActionsService_GetWorkflowRun.*Logs|RepositoriesService_GetArchiveLink|RepositoriesService_GetBranch)' -count=1`
- `go test -modfile=go.test.mod ./... -count=1`
- `go test -v -tags=integration -run='^$' -modfile=go.test.mod ./test/integration`
- `go vet -modfile=go.test.mod ./...`
- `gofmt -d github/github.go github/github_timeout_test.go`
- `git diff --check`

The temporary `go.test.mod` used an uncommitted local replacement for `github.com/google/go-querystring` because this Windows host could not reach `proxy.golang.org`; it was removed after each run.

## Limitations

The required repository-wide race run was attempted, but this host has `CGO_ENABLED=0` and no available C compiler. The all-module non-race script, custom linter bootstrap, generated-file check, schema check, and OpenAPI validation were also attempted; external module/TLS downloads and the custom-linter installer were unavailable on this host. The logs are retained under `/g/OSS-PR-200/work/evidence`.

AI assistance was used to investigate, reproduce, implement, test, and prepare this PR. I reviewed the complete diff and the reported test results; the change is intentionally limited to the timeout behavior and its regression test.


gmlewis (changes requested):
Thanks for the diagnosis - the missing timeout needs to be addressed, but not in the manner that you are proposing, because this implementation causes 3 more issues:

- Transport errors are now wrapped in *url.Error, so err == mySentinel, error types, and error strings all change.
- Do attaches the client's http.CookieJar: these requests now send jar cookies and write response Set-Cookie values into the shared jar.
- Do parses Location before our checkRedirectHost gets to run, so an unparseable Location now returns (nil, err) instead of the redirect response and callers lose the *Response.

I believe the safer approach is to revert the changes you made and add the following BEFORE line 2381:

```go
if t := c.client.Timeout; t > 0 {
    var cancel context.CancelFunc
    ctx, cancel = context.WithTimeout(ctx, t)
    defer cancel()
}
```

Also, please move your new test to `github/github_test.go` instead of making a new file.

And finally, please do not use `time.Sleep` in any unit tests - look at the other examples and take advantage of `synctest` and other testutils to make the unit test independent of time, for when these tests are run on extremely slow runners.

Once you fix the CLA issues, we can move forward with this PR.

PR #4586, opened 2026-09-21
feat: Add required field on combined status responses

jasonmcintosh:
This enables on status checks the ability to filter the list of status's to those that are REQUIRED to pass or not, allowing one to merge for example if all required checks pass, and others are not passed.  The field is documented on github api's and ONLY on the CombinedStatus response types.  

image

https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#get-the-combined-status-for-a-specific-reference

With out this, you can't tell whether a status check is easily with the SDK without querying the required APIs which requires more administrative level access.


gmlewis:
Please run step 4 of CONTRIBUTING.md and push the changes to this PR.

gmlewis (approved):
Thank you, @jasonmcintosh!
LGTM.

github/repos_statuses_test.go
```diff
@@ -142,3 +142,32 @@ func TestRepositoriesService_GetCombinedStatus(t *testing.T) {
 		return resp, err
 	})
 }
+
+func TestRepositoriesService_GetCombinedStatus_requiredField(t *testing.T) {
```
Not-Dhananjay-Mishra: Can we modify existing `TestRepositoriesService_GetCombinedStatus` test and include `{"id":1,"required":true},{"id":2,"required":false}` rather than creating a new test?

